Issue #5, opened 2019-09-06
Upstream cilium GetMapInfo


lmb:
Cilium has a cool function that parses map attributes from `/proc`: https://github.com/cilium/cilium/blob/12fe19d26afd005aa0e0e4c312f81b52af919b82/pkg/bpf/map_linux.go#L282

We should incorporate this into LoadPinnedMap, to make it work on kernels < 4.14.

lmb:
@joestringer do you think you could take a stab at this?

joestringer:
Sure, it's all just a matter of finding the time ;)

whl739:
And `GetProgInfoByFD` ? (https://github.com/cilium/cilium/blob/master/pkg/bpf/prog_linux.go#L106)

lmb:
That's what LoadPinned(Map|Program) use right now. This issue is about adding a fall back for older kernels that don't have the function.

whl739:
Got it. 
But how can i use APIs like GetMapInfo or GetProgInfo, they are private in `syscalls.go`(for example, `bpfGetMapInfoByFD`).

joestringer:
@whl739 Does `LoadPinnedMap()` / `LoadPinnedProgram()` do what you want? I believe that these are effectively the higher level API for the same thing.

whl739:
@joestringer you are right, these APIs are enough if i only want to get the map/prog.
But if i want to get more info about map/prog, such as `type bpfMapInfo struct` and `type bpfProgInfo struct` defined in `syscalls.go`,  i didn't find any API can do that, did i missing something?

lmb:
You can use https://godoc.org/github.com/cilium/ebpf#Map.ABI to access some of the information from the syscalls. Please create a separate issue explaining what you want to achieve if that isn't enough for your use case.

whl739:
Thanks for your kind help. I'll create a new issue.

PR #7, opened 2019-09-23
Upstream PerfReader from newtools/ebpf

### LINKED ISSUE #3: Upstream PerfReader from newtools/ebpf
See https://godoc.org/github.com/newtools/ebpf#PerfReader

I want to make some changes to the API however:
- Get rid of the Results and Errors channel and make the API synchronous
- Maybe move into a sub-package

lmb:
This is part two of the upstreaming effort, which moves the code to interact with perf ring buffers into the repository. I took some inspiration from [pkg/bpf/perf_linux.go](https://github.com/cilium/cilium/blob/fdf5191739015e63fd8378d75246268ed46f7090/pkg/bpf/perf_linux.go) and changed our implementation:
* There is now a `Read` function, which returns a single `*Record` and may block. Previously you had to read from a channel.
* `Close` interrupts `Read`

There are still large differences between the two though:
* A single `Reader` vs. multiple `PerfEvent`
* `Read` waits on an epoll fd for ready rings, instead of polling all rings
* `Read` returns `*Record, error` instead of using callbacks
* `Enable`, `Disable` are missing
* A lot less internals of Perf are exposed

@arthurfabre can you take a look and tell me whether `xdpcap` can get away without `FlushAndClose` with the new API?

@joestringer it seems like you've worked on your perf reader before, can you take a look and tell me what's missing / what would need to change?

Current API:
```
package perf // import "github.com/cilium/ebpf/perf"

Package perf allows interacting with Linux perf_events.

BPF allows submitting custom perf_events to a ring-buffer set up by
userspace. This is very useful to push things like packet samples from BPF
to a daemon running in user space.

FUNCTIONS

func IsClosed(err error) bool
    IsClosed returns true if the error occurred because a Reader was closed.


TYPES

type Reader struct {
	// Has unexported fields.
}
    Reader allows reading bpf_perf_event_output from user space.

func NewReader(opts ReaderOptions) (pr *Reader, err error)
    NewReader creates a new reader with the given options.

func (pr *Reader) Close() error
    Close frees resources used by the reader.

    It interrupts calls to Read.

    Calls to perf_event_output from eBPF programs will return ENOENT after
    calling this method.

func (pr *Reader) Read() (*Record, error)
    Read the next record from the perf ring buffer.

    The function blocks until there are at least Watermark bytes in one of the
    per CPU buffers.

    Records from buffers below the Watermark are not returned.

    Calling Close interrupts the function.

func (pr *Reader) ReadRecord(rec *Record) error
    ReadRecord reads into a pre-allocated Record.

    Record.RawSample will be re-allocated if the buffer is too small to hold the
    full sample.

    See Read for further information.

type ReaderOptions struct {
	// A map of type PerfEventArray.
	Map *ebpf.Map
	// Controls the size of the per CPU buffer in bytes.
	PerCPUBuffer int
	// The number of written bytes required in any per CPU buffer before
	Watermark int
}
    ReaderOptions control the behaviour of the user space reader.

type Record struct {
	// The data submitted via bpf_perf_event_output.
	// They are padded with 0 to have a 64-bit alignment.
	// If you are using variable length samples you need to take
	// this into account.
	RawSample []byte

	// The number of samples which could not be output, since
	// the ring buffer was full.
	LostSamples uint64
}
    Record contains either a sample or a counter of the number of lost samples.

``` 

Fixes #3


perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+// NewReader creates a new reader with the given options.
+func NewReader(opts ReaderOptions) (pr *Reader, err error) {
+	if opts.PerCPUBuffer < 1 {
+		return nil, errors.New("PerCPUBuffer must be larger than 0")
+	}
+
+	// We can't create a ring for CPUs that aren't online, so use only the online (of possible) CPUs
+	nCPU, err := internal.OnlineCPUs()
+	if err != nil {
+		return nil, errors.Wrap(err, "sampled perf event")
```
arthurfabre: Not sure this error message makes sense?

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+	case unix.ESRCH:
+		return -1, errors.WithMessage(unix.ESRCH, "pid does not exist")
+	default:
+		return -1, err
+	}
+}
+
+func addToEpoll(epollfd, fd int, data, pad int32) error {
+	// The representation of EpollEvent isn't entirely accurate.
+	// According to man 2 epoll_wait, Fd and Pad are really a union,
```
arthurfabre: Are `Fd` and `Pad` really a union? They're just `epoll_data`, which is 64bits no?

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+			if temp, ok := err.(temporaryError); ok && temp.Temporary() {
+				continue
+			}
+
+			if err != nil {
+				return nil, err
+			}
+
+			for _, event := range pr.epollEvents[:nEvents] {
+				if event.Pad == wokenByClose {
```
arthurfabre: Do we need to use `Pad` and `Fd` this way? I find it confusing to follow that `event.Fd` has the CPU number. What if we put actual fds in `Fd`, and have a `map[fd]ring` and check `event.Fd == closeFd`?

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+					return nil, errors.New("closed")
```
arthurfabre: Should this use `os.ErrClosed` so we can check for it easily?

arthurfabre:
I'm not sure this is enough to emulate `FlushAndClose()` entirely, `Read()` will still block if the data is below the watermark no (ie no epoll event)?

How do callers read all the already generated events, but no new ones? Even if they ensure no new events are generated (eg detach the program, or stop calling `prog.Test()`), Read() would end up blocking at some point. I guess they can `Close()` after a timeout.

I don't think this matters too much for `xdpcap`, but it makes testing easier if we can be reasonably certain that we get all the events.

perf/perf.go
lmb: Yeah, that comment is confusing. I meant to say that `Pad` is usable like everything else.

perf/perf.go
lmb: True, but in a way that is also obfuscating what is going on. There is no requirement for event to contain the fd, its just that `unix` named it so, which is why I added that hard to understand comment about the union further up.

There is also a bit of premature optimization here, because lookup in small maps are "slow":  https://github.com/golang/go/issues/19495 By using the CPU here we can zero index into a slice instead of using a map. Not sure if it really makes a difference.

I'll have a think whether I can keep using event like this but make it less confusing.

perf/perf.go
lmb: Ooh, TIL. Yes, that makes sense.

lmb:
Have a look at the rewritten tests: if you know how many record you expect its easy to just `Read` in a loop that many times.

> How do callers read all the already generated events, but no new ones?

Yeah, that's exactly the problem. But I'm not sure whether an application would really want to do that though. The (non-testing related) use cases we have don't seem to require it, and I can't come up with a good example.

If we need this, we could fix it by saying that after calling Close, Read will return all remaining records, then ErrClosed. But that is still full of race conditions, and it means that the code to clean up resources becomes more complicated. We'd also lose the guarantee that once Close returns, Read will always return an error.

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+	closeOnce sync.Once
+}
+
+// ReaderOptions control the behaviour of the user
+// space reader.
+type ReaderOptions struct {
+	// A map of type PerfEventArray.
+	Map *ebpf.Map
+	// Controls the size of the per CPU buffer in bytes.
+	PerCPUBuffer int
```
joestringer: For performance reasons we presumably would want to align this with page size, is there any benefit to allowing arbitrary byte-sized buffers?

From Cilium I'm expecting to basically do a `pageSize := os.Getpagesize()` then pass some multiple of that in here. Not a big deal by any means, but making it easier for users of the library to do the right thing might guide us to just expose a "number of pages" thing here.

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+}
+
+func readRawSample(rd io.Reader) ([]byte, error) {
+	var size uint32
+	if err := binary.Read(rd, internal.NativeEndian, &size); err != nil {
+		return nil, errors.Wrap(err, "can't read sample size")
+	}
+
+	data := make([]byte, int(size))
+	_, err := io.ReadFull(rd, data)
```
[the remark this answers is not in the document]
joestringer: I realise there may actually be some ugly CGo hacks behind Cilium's ability to retain / pass pointers into the ringbuffer here, not sure exactly what it looks like for a variable-length golang reference to be passed back out to the caller, in a safe manner, without allocations, when the sample structure isn't actually known by the library...

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+type perfEventHeader struct {
```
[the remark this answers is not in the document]
lmb: Good idea.

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+	"os"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
lmb: Yes, I made sure that the whole package doesn't export errors to this end. Instead we have `IsNotFound` etc. which we can migrate to the new package. I haven't figured out how to support Go 1.12 etc. though.

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+	return ring, nil
+}
+
+func createPerfEvent(cpu, watermark int) (int, error) {
+	attr := unix.PerfEventAttr{
+		Type:        unix.PERF_TYPE_SOFTWARE,
+		Config:      unix.PERF_COUNT_SW_BPF_OUTPUT,
+		Bits:        unix.PerfBitWatermark,
+		Sample_type: unix.PERF_SAMPLE_RAW,
+		Wakeup:      uint32(watermark),
```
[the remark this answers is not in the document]
lmb: Setting it to 0 makes it fall back to the kernel behaviour, which uses bufsize / 2 as the watermark. The idea is that you can batch reading from the ring buffer by delaying the epoll notification until the buffer is at least half (or a value of your choosing) full. This ends up saving CPU at the cost of latency.

We set it to 1 if we want immediate feedback (so `xdpcap`) and to bufsize / 2 for our DoS mitigation packet sampler, where we want to save CPU. We've not benchmarked whether this really makes a difference.

We could probably get rid of this for now, and just hardcode watermark 1 or whatever.

perf/perf.go
```diff
@@ -0,0 +1,520 @@
...
+)
+
+// NewReader creates a new reader with the given options.
+func NewReader(opts ReaderOptions) (pr *Reader, err error) {
+	if opts.PerCPUBuffer < 1 {
+		return nil, errors.New("PerCPUBuffer must be larger than 0")
+	}
+
+	// We can't create a ring for CPUs that aren't online, so use only the online (of possible) CPUs
+	nCPU, err := internal.OnlineCPUs()
```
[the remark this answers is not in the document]
lmb: I just read the documentation for `NumCPU`, and it seems to me like that value can be influenced by things like CPU affinity for the userspace proess. We really want online CPUs in the system, since we don't control what CPU an eBPF program runs on.

perf/perf.go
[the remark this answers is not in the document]
lmb: > These above two lines I'm wondering about from a high level, it's an enforced memory allocation/copy of the data from the ringbuffer, for every single sample.

I kind of figured you would raise this :) It was the reason I re-wrote the Reader to not use a buffered channel. How about something like this:

`ReadRecord(*Record) error`

You can then pre-allocate Record.RawSample to your liking, and afterwards unmarshal into your structure.

I didn't add it to this PR since it begs the question what happens when RawSample is to small. Return an error? Discard and re-allocate?

> Is it worth reconsidering the Read() API to use a model that allows arbitrary user code to interact with the memory-mapped buffer directly?

See my preferred solution above.

Accessing the ring buffer as []byte doesn't work because of records wrapping around the buffer. You need an `io.Reader` to abstract that away. Once you have a reader you kind of need to do a copy to get a linear buffer anyway, which I suspect is why your current code does a copy as well. Once you have the linear buffer you can unmarshal from it.

perf/perf.go
lmb: I think it rounds up to the nearest page multiple. Using number of pages makes the buffer size hard to understand IMO, since page sizes are variable between arches / kernel configs.

perf/perf.go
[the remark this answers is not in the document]
lmb: I've added these.

perf/perf.go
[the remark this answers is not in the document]
lmb: I've added a `ReadRecord` to the PR, please take a look. Usage would work like this:

```go
var rec Record
for {
  reader.ReadRecord(&rec)
  // ...
}
```

This will successively grow the buffer to the maximum sample you receive.

perf/perf.go
lmb: I added `IsClosed` instead, to the same effect.

perf/perf.go
lmb: I changed it around a little bit: `event.Fd` now really does contain the fd and we compare against `closeFd`. CPU has moved to event.Pad, and there is a small accessor function `cpuForEvent` which makes the code easier to read.

lmb:
I've pushed a version with comments addressed as far as possible.

@joestringer does `ReadRecord` satisfy you? There are other bits of the reader that still allocate (reading the headers and such) but optimizing that away should be trivial, and can be done in a follow up I think.

Should we keep Watermark, or chuck it? I think at the very least we should make `Watermark: 0` behave like `Watermark: 1`.

perf/perf.go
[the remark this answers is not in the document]
joestringer: (Do we need to support Go 1.12?)

perf/perf.go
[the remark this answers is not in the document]
joestringer: I think it's reasonable to specify via bytes. Easier to actually manage your application's memory footprint when the units are the standard units.

perf/perf.go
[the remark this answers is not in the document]
joestringer: When I think a bit more, I don't know if it's worth adding such an API at this point. If you're happy with this new version you came up with, and the testing / corner cases are all handled, then we can move forward with it, but otherwise maybe we can defer it until we get a strong argument for why we need to do it. It's certainly a bit more ugly from a go abstraction point of view.

(LMK, I can take a closer look, didn't get a chance today)

perf/perf.go
joestringer: I think it's reasonable to specify via bytes. Easier to actually manage your application's memory footprint when the units are the standard units.

perf/perf.go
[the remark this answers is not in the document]
joestringer: Ah, I see. By all means if you have regular use cases for configuring this another way, then we can support that. I'm mainly coming from the perspective of making the default simple and stupid and to work more-or-less the way you'd expect it to. I think that as-is, the user *must* think about how many bytes of events to look for, instead of defaulting to process one event at a time unless you provide a different option.

joestringer:
Didn't get much chance to keep rebasing and poking/prodding the PR today, hopefully more cycles later this week.

`Watermark: 0` behaving like `Watermark: 1` sounds like a reasonable default.

perf/perf.go
```diff
@@ -0,0 +1,543 @@
...
+		return errClosed
+	}
+
+	for {
+		if len(pr.epollRings) == 0 {
+			nEvents, err := unix.EpollWait(pr.epollFd, pr.epollEvents, -1)
+
+			// Handle EINTR
+			if temp, ok := err.(temporaryError); ok && temp.Temporary() {
+				continue
```
joestringer: Why does this need to continue, rather than breaking the loop?

perf/perf.go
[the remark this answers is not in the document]
lmb: I think we should support whatever the Go project supports, which is the last two releases.

perf/perf.go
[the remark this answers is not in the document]
lmb: Yeah, I'll change this.

perf/perf.go
lmb: Because we need to retry the syscall, see https://github.com/golang/go/issues/20400

perf/perf.go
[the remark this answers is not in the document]
lmb: OK, removed ReadRecord again. Let's tackle reducing allocations in a follow up.

lmb:
Made Watermark: 0 read events without batching and removed `ReadRecord`.

perf/perf.go
joestringer: Ah, so the interruption here is intended to be ignored and we just continue with the dump?

I think a comment with the intent and a link to the above golang thread would be useful in the code.

perf/perf.go
```diff
@@ -0,0 +1,529 @@
...
+	case perfRecordLost:
+		lost, err := readLostRecords(rd)
+		return &Record{LostSamples: lost}, err
+
+	case perfRecordSample:
+		sample, err := readRawSample(rd)
+		return &Record{RawSample: sample}, err
+
+	default:
+		return nil, errors.Errorf("unknown event type %d", header.Type)
```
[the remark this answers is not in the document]
joestringer: I wondered how we want user libraries to handle the variety of events that we might generate here as well. It might be kinda neat to have users do something like:

```
event, err := reader.Read()
if err {
  ...
  continue
}
switch event.(type) {
case SampleRecord:
...
case LostRecord:
...
case UnknownRecord:
...
}
```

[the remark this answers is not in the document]
joestringer: What do you folks do to handle the different types of events that might be generated here?

perf/perf.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+
+	switch header.Type {
+	case perfRecordLost:
+		lost, err := readLostRecords(rd)
+		return &Record{LostSamples: lost}, err
+
+	case perfRecordSample:
+		sample, err := readRawSample(rd)
+		return &Record{RawSample: sample}, err
```
joestringer: This will also allocate a structure, right? Because it's returning a pointer to the stack. Then inside, the `sample` was already allocated as well (the slice itself *might* be passed by value, but the `make(...)` call obviously needs to allocate enough memory for the data inside the slice).

internal/cpu.go
```diff
@@ -0,0 +1,64 @@
...
+
+var onlineCPU struct {
+	once sync.Once
+	err  error
+	num  int
+}
+
+// OnlineCPUs returns the number of currently online CPUs
+// Logical CPU numbers must be of the form 0-n
+func OnlineCPUs() (int, error) {
```
joestringer: Oh, and this is currently structured in the code to be hidden from users, but in Cilium we actually report the number of CPUs that we configure perf readers for, mainly I think just as a basic sanity check to ensure that the implementation is working as intended. At the moment we don't have any way to introspect the number used by the perf library, unless we add a dependency on this `github.com/cilium/ebpf/internal`.

Not sure what the best path forward on this is. In some sense, if the library works correctly on all platforms, everything is fine. But if it doesn't work properly, by misconfiguring the number of CPUs for instance, then I'm not sure how we would introspect this. Just using the actual BPF map state for the events map?

perf/perf.go
lmb: Done!

perf/perf.go
```diff
@@ -0,0 +1,529 @@
...
+		return nil, errClosed
+	}
+
+	defer ring.reader.writeTail()
+	return readRecord(ring.reader)
+}
+
+// Record contains either a sample or a counter of the
+// number of lost samples.
+type Record struct {
```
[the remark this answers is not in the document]
lmb: Good idea, I'll add it.

perf/perf.go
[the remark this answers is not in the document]
lmb: As far as I can tell, because of
```
		Type:        unix.PERF_TYPE_SOFTWARE,
		Config:      unix.PERF_COUNT_SW_BPF_OUTPUT,
```
We can only receive raw events from the ring. I think we've never seen the unknown record being triggered. Do you have a different set up?

perf/perf.go
lmb: Yes, right now reading a sample heap allocates twice, since returning the `*Record` makes it escape.

internal/cpu.go
lmb: We have counters for failed `perf_event_output` that are exported from the eBPF, which covers this failure case. Would that work for you as well?

Alternatively we could add `Reader.NumCPU() int` or something.

perf/perf.go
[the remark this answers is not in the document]
lmb: Re: your idea with `SampleRecord`, etc. It forces us to return an interface, which will make it impossible to avoid allocations.

So far my hope is that we can just add more members to `Record` if we need to. Not great, admittedly. The other option is to introduce completely distinct Record types, and add `Read` methods for them. Also not great.

perf/perf.go
```diff
@@ -0,0 +1,529 @@
...
+	ring.fd = -1
+	ring.mmap = nil
+}
+
+func (ring *perfEventRing) LoadHead() {
+	ring.reader.loadHead()
+}
+
+// NB: Has to be preceded by a call to LoadHead.
+func (ring *perfEventRing) ReadRecord() (*Record, error) {
```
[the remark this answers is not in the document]
lmb: Done!

perf/perf.go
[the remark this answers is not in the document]
lmb: Fixed

perf/perf.go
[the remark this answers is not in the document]
joestringer: I was wondering how allocations might fall into this, good point. There's not many record types anyway so it's not a big deal for that interface idea, we can drop that.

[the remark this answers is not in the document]
joestringer: > We can only receive raw events from the ring. I think we've never seen the unknown record being triggered. Do you have a different set up?

I've never seen it personally but there are rumours in the office that these were seen last year some time ;)

My main concern is to avoid logging the same error message every time we receive an event like this and causing disk issues due to this. So long as we have a way to tell that a given error is the same as the one we saw before, then we will have a reasonable means to ratelimit logging. Whether that's by having a static error variable that we can check against or a function that we can use to check "is it X type of error?" or we present unknown errors like this as a `Record{Unknown: true}` (or some variation) I don't have a strong opinion.

Currently as returned right now it looks like a brand new error pointer each time and there's no sense of "identity" for this error, so no way to tell from a library user whether this is the kind of error it is. Maybe we can just add a new error class for this error type.

internal/cpu.go
joestringer: The simplest to satisfy the API we expose right now is just to get `Reader.NumCPU()`, as we literally just print the number of CPUs configured with perf events. 

We have this other problem though, which is that our C template defines the perf event map using `__NR_CPUS__` to configure the size of the map.. then from the userspace we open this map. Right now I'm doing a  `ebpf.LoadPinnedMapExplicit(path, eventsMapTemplate)` where the `eventsMapTemplate` is configured to have `MaxEntries: uint32(runtime.NumCPU())`, and hope that this matches (which it seems to at least on my laptop). That said, we generate our C code templates so we can always change the `__NR_CPUS__` definition to instead be written as a constant value from our golang code, and define it also using `runtime.NumCPU()`. OK, so that problem is basically solved...

So let's say we have `Reader.NumCPU()`. We still need to define the map's `MaxEntries` somehow, and this should be configured properly before we pass that map into `perf.NewReader()`. So this means it's actually a bit weird for the Reader to go and determine the number of CPUs because it will use its own method for this which is not guaranteed to be the same as the method that the calling library uses for defining the map size. How do you expect library users to determine the size of the map and what kind of assumptions is the `PerfReader()` making here based on the map size?

> We have counters for failed `perf_event_output` that are exported from the eBPF, which covers this failure case. Would that work for you as well?

This is if the BPF code attempts to output but there's no ring fd defined, right? That's also interesting, how is this exposed from the kernel? I can't recall seeing somewhere that we expose this right now but it sounds like something that could be handy to know about when weird corner case bugs come up.

perf/perf.go
joestringer: Just to dig in and understand, bear with me ;)

What's the benefit to allocating and returning a pointer here? Could we pass back the struct by value? We'd still be allocating the actual buffer for the message, but this extra minor metadata (currently only the lost events counter and the slice header) wouldn't require an allocation.

If we knew for sure that this wouldn't be extended, I'd even consider passing back the `[]byte` structure directly. The only benefit that the `Record` structure provides right now is being a generic wrapper that supports passing back extended metadata in future. I think that's actually tidier than the alternative, but I'm also inclined to ask how we can minimize the expensive operations here ;)

lmb: Yeah, returning `Record` by value here makes sense. Good idea!

internal/cpu.go
lmb: > We have this other problem though, which is that our C template defines the perf event map using __NR_CPUS__ to configure the size of the map.

I made this quick test:

```
$ cat cpu.go
package main

import (
        "fmt"
        "runtime"
)

func main() {
        fmt.Println("CPUs:", runtime.NumCPU())
}
$ go run cpu.go
CPUs: 8
$ taskset -c 0 go run cpu.go
CPUs: 1
```

Using `NumCPU` is a bug I think.

> How do you expect library users to determine the size of the map [...]

The library forces PerfEventArrays to have MaxElems == 0, and then plugs in the correct value at creation time: https://github.com/cilium/ebpf/blob/master/map.go#L100 That has the same effect as NR_CPUS but is a bit more genera, since code that creates maps directly also works. To support migration we can remove the requirement of MaxElems == 0.

There is actually another problem: the map is created with MaxElems = possible CPUs, because some systems can hotplug CPUs (@arthurfabre has one). The eBPF will then happily run on those new CPUs.  However, the perf rings are created for online CPUs, since it's impossible to create perf rings for offline CPUs. I'll create a bug for this and come back to it later. Most likely we should just declare that we don't support hotplug right now.

> [...] what kind of assumptions is the PerfReader() making here based on the map size?

Currently, it assumes that the map size == online cpus. I think you have a point, it should be possible to change that to use the size of the map instead. That also circumvents the need for `Reader.NumCPU` since you can print map.ABI().MaxElems.

> This is if the BPF code attempts to output but there's no ring fd defined, right? That's also interesting, how is this exposed from the kernel?

Yes, we just bump a counter in a per CPU array whenever `perf_event_output` returns < 0. When collecting metrics we sum the per CPU counters into a single value, and expose that via Prometheus. This gives us very cheap metrics.

perf/perf.go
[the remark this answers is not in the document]
lmb: I've added `IsUknownEvent(error)`.

internal/cpu.go
joestringer: > Using `NumCPU` is a bug I think.

Great example, agree it doesn't make sense to allocate map size based on the golang process CPU affinity.

> The library forces PerfEventArrays to have MaxElems == 0, and then plugs in the correct value at creation time: https://github.com/cilium/ebpf/blob/master/map.go#L100 [...]

For the `NewMap()` case, but not the `LoadPinnedMap()` / `LoadPinnedMapExplicit()` cases. In those cases, whoever defines the map still needs to know about and configure the map size, but then there's a possibility for this to be out-of-sync with the Reader that is opened on top of the map. If they're in sync, everything is fine. If they somehow determine the `MaxEntries` / number of CPUs differently then opening the reader will consistently fail.

I guess where I'm going with this is, should we consider making the `PossibleCPUs` and/or `OnlineCPUs` part of the library API which users could then use to configure their maps for the pinned case? This would be useful not only for perf event reader maps, but also PerCPU maps as well.

> Most likely we should just declare that we don't support hotplug right now.

:+1: 

> Yes, we just bump a counter in a per CPU array [...]

Ah, makes sense. So there's not really a generic way to integrate this into the library, for example as statistics that could be fetched via some perf API. Anyway, sounds useful, thanks for the explanation.

joestringer: > I guess where I'm going with this is, should we consider making the PossibleCPUs and/or OnlineCPUs part of the library API which users could then use to configure their maps for the pinned case? This would be useful not only for perf event reader maps, but also PerCPU maps as well.

I was wondering if I'm thinking too narrowly on this, why not create the map from Cilium agent side? However, we still have the upgrade case where we would presumably prefer to leave the existing map in place, as-is, and not recreate it every time we upgrade or restart Cilium. So then we still need the ability to open the map from a pinned location on the filesystem, and (given our support for <4.13 kernels) will need to provide the map size in the ABI for the `LoadPinnedMapExplicit()` call.

lmb: Once we have #5 `LoadPinnedMap` will work on earlier kernels, so you won't need to use the explicit version. Which means you don't need to know the number of CPUs.

perf/ring.go
```diff
@@ -0,0 +1,182 @@
...
+	case unix.EINVAL:
+		return -1, errors.WithMessage(unix.EINVAL, "the specified event is invalid, most likely because a configuration parameter is invalid (i.e. too high, too low, etc)")
+	case unix.EMFILE:
+		return -1, errors.WithMessage(unix.EMFILE, "this process has reached its limits for number of open events that it may have")
+	case unix.ENODEV:
+		return -1, errors.WithMessage(unix.ENODEV, "this processor architecture does not support this event type")
+	case unix.ENOENT:
+		return -1, errors.WithMessage(unix.ENOENT, "the type setting is not valid")
+	case unix.ENOSPC:
+		return -1, errors.WithMessage(unix.ENOSPC, "the hardware limit for breakpoints)capacity has been reached")
```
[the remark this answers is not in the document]
lmb: Copy pasta problem!

lmb:
For some reason I can't reply to your last comment.

> I guess where I'm going with this is, should we consider making the PossibleCPUs and/or OnlineCPUs part of the library API which users could then use to configure their maps for the pinned case? This would be useful not only for perf event reader maps, but also PerCPU maps as well.

I'm happy to go that way if it turns out to be useful. I just prefer to start out with a minimal API, and then see where it goes. It seems that for the purpose of perf events we can get away without exposing them, so I'd like to punt on that for now. Especially because the functions deliberately don't support certain CPU configurations.

perf/ring.go
```diff
@@ -0,0 +1,182 @@
...
+		return fd, nil
+	}
+
+	switch err {
+	case unix.E2BIG:
+		return -1, errors.WithMessage(unix.E2BIG, "perf_event_attr size is incorrect,check size field for what the correct size should be")
+	case unix.EACCES:
+		return -1, errors.WithMessage(unix.EACCES, "insufficient capabilities to create this event")
+	case unix.EBADFD:
+		return -1, errors.WithMessage(unix.EBADFD, "group_fd is invalid")
```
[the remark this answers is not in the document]
lmb: The error messages are from a previous incarnation of the code, I never bothered to look at them in detail to be honest. I'll just remove the weird bits.

lmb:
I've removed the (partially) misleading error messages and made the API changes we've discussed so far.

perf/reader_test.go
```diff
@@ -0,0 +1,374 @@
...
+	// To generate a lost sample perf record:
+	//
+	// 1. Fill the perf ring buffer almost completely, with the output_large program.
+	//    The buffer is sized in number of pages, which are architecture dependant.
+	//
+	// 2. Write an extra event that doesn't fit in the space remaining.
+	//
+	// 3. Write a smaller event that does fit, with output_single program.
+	//    Lost sample records are generated opportunistically, when the kernel
+	//    is writing an event and realizes that there were events lost previously.
```
joestringer: This is awesome.

I was wondering about another case which is basically the writing of the events over the barrier of the ringbuffer, where you would expect the first half of the event to be at the end of the buffer and the last half is at the start of the buffer.

You might imagine writing a test similar to this one, where first you almost fill the buffer, then consume all of the events, then write one more event across the end (/start) to loop back around and make sure this final event is correctly read out.

Anyway, that can be part of a subsquent follow-up.

lmb: Yeah, good idea. There is a test for the wrap around behaviour of the ring reader, `TestRingBufferReader`. Have a look whether we need more.

PR #8, opened 2019-09-24
Use BPF_OBJ_NAME_LEN and BPF_TAG_SIZE from golang.org/x/sys/unix

tklauser:
Update golang.org/x/sys dependency to get these consts.


lmb:
Thanks!

PR #11, opened 2019-10-02
map: bound retries when iterating

### LINKED ISSUE #9: MapIterator doesn't terminate for heavily modified map
MapIterator.Next uses BPF_MAP_GET_NEXT_KEY to iterate a map. However, there is no guarantee that a lookup makes forward progress through a map. This is a problem with the kernel implementation, for which a couple of solutions were discussed at Linux Plumbers Conf '19.

We need a workaround until the solution is available. The Cilium project imposes a max # of iterations to protect against this behaviour. What is the right number for that? Maybe just use MaxElems as a hard limit? I'd like to avoid introducing user visible knobs, if possible.

FWIW, it's possible to do the following:

```go
for i := 0; entries.Next(&key, &value) && i < 1000; i ++ {
    // ...
}
```

But it requires every user to know about the problem.

cc @joestringer

lmb:
MapIterator.Next uses BPF_MAP_GET_NEXT_KEY to iterate a map. However,
there is no guarantee that a lookup makes forward progress through a
map. This is a problem with the kernel API, for which there is
currently no fix.

The problem manifests with heavily modified hashmaps, where the
race between MAP_GET_NEXT_KEY and MAP_LOOKUP_ELEM can lead to the
iterator code retrying an unbounded number of times.

Limit the iterator to MaxElems calls.

Fixes #9

### DIFF
--- a/map.go
+++ b/map.go
@@ -481,23 +481,32 @@ func (m *Map) MarshalBinary() ([]byte, error) {
 //
 // See Map.Iterate.
 type MapIterator struct {
-	target    *Map
-	prevKey   interface{}
-	prevBytes []byte
-	done      bool
-	err       error
+	target            *Map
+	prevKey           interface{}
+	prevBytes         []byte
+	count, maxEntries uint32
+	done              bool
+	err               error
 }
 
 func newMapIterator(target *Map) *MapIterator {
 	return &MapIterator{
-		target:    target,
-		prevBytes: make([]byte, int(target.abi.KeySize)),
+		target:     target,
+		maxEntries: target.abi.MaxEntries,
+		prevBytes:  make([]byte, int(target.abi.KeySize)),
 	}
 }
 
+var errIterationAborted = errors.New("iteration aborted")
+
 // Next decodes the next key and value.
 //
-// Returns false if there are no more entries.
+// Iterating a hash map from which keys are being deleted is not
+// safe. You may see the same key multiple times. Iteration may
+// also abort with an error, see IsIterationAborted.
+//
+// Returns false if there are no more entries. You must check
+// the result of Err afterwards.
 //
 // See Map.Get for further caveats around valueOut.
 func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
@@ -505,7 +514,7 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 		return false
 	}
 
-	for {
+	for ; mi.count < mi.maxEntries; mi.count++ {
 		var nextBytes []byte
 		nextBytes, mi.err = mi.target.NextKeyBytes(mi.prevKey)
 		if mi.err != nil {
@@ -529,11 +538,11 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 			// Even though the key should be valid, we couldn't look up
 			// its value. If we're iterating a hash map this is probably
 			// because a concurrent delete removed the value before we
-			// could get it. If we're iterating one of the fd maps like
+			// could get it. This means that the next call to NextKeyBytes
+			// is very likely to restart iteration.
+			// If we're iterating one of the fd maps like
 			// ProgramArray it means that a given slot doesn't have
-			// a valid fd associated.
-			// In either case there isn't much we can do, so just
-			// continue to the next key.
+			// a valid fd associated. It's OK to continue to the next slot.
 			continue
 		}
 		if mi.err != nil {
@@ -543,6 +552,9 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 		mi.err = unmarshalBytes(keyOut, nextBytes)
 		return mi.err == nil
 	}
+
+	mi.err = errIterationAborted
+	return false
 }
 
 // Err returns any encountered error.
@@ -556,4 +568,11 @@ func (mi *MapIterator) Err() error {
 // key doesn't exist.
 func IsNotExist(err error) bool {
 	return errors.Cause(err) == unix.ENOENT
+}
+
+// IsIterationAborted returns true if the iteration was aborted.
+//
+// This occurs when keys are deleted from a hash map during iteration.
+func IsIterationAborted(err error) bool {
+	return errors.Cause(err) == errIterationAborted
 }
--- a/map_test.go
+++ b/map_test.go
@@ -290,7 +290,15 @@ func TestIterateEmptyMap(t *testing.T) {
 }
 
 func TestMapIterate(t *testing.T) {
-	hash := createHash()
+	hash, err := NewMap(&MapSpec{
+		Type:       Hash,
+		KeySize:    5,
+		ValueSize:  4,
+		MaxEntries: 2,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
 	defer hash.Close()
 
 	if err := hash.Put("hello", uint32(21)); err != nil {
@@ -316,6 +324,9 @@ func TestMapIterate(t *testing.T) {
 
 	sort.Strings(keys)
 
+	if n := len(keys); n != 2 {
+		t.Fatal("Expected to get 2 keys, have", n)
+	}
 	if keys[0] != "hello" {
 		t.Error("Expected index 0 to be hello, got", keys[0])
 	}


joestringer:
Yep, complete restart. That's why we keep track of at least the last entry we saw so we can restart from there rather than defaulting to the kernel behaviour of starting again from the beginning of the map. (Note this can still fail; again, we have metrics to track this condition)

`Completed() bool` would solve for this specific case, but in Cilium we really have a whole bunch of metrics that we track which it'd be nice to figure out how to model:

https://github.com/cilium/cilium/blob/72a7c97e91b63217e7f06e865dbd86379d26df42/pkg/bpf/map_linux.go#L620
https://github.com/cilium/cilium/blob/72a7c97e91b63217e7f06e865dbd86379d26df42/pkg/bpf/stats_linux.go#L24

What do we do with these? Well, to a large extent, just export them to the user. If we don't track these weird cases, we'll never know if they get hit. If it triggers some alert or someone notices a metric increase on a dashboard somewhere, then they can let us know that "the bad number went up" and we can consider whether our map handling logic needs to be improved for a future release.

joestringer:
Some of the dump stats logic above could be composed on top of the core map functionality provided by this logic for sure - for example the dump start/finish time. If we can figure out a way to do all of those stats, then that would provide a way to keep the implementation here simple.

On the flip side, if this sounds potentially interesting/useful to application developers using this library and we can come up with a clean and efficient way to track/report the stats, then having it inside would be nice.

Quick idea off the top of my head, we could do something like this: Represent errors like dump interruption, key fallback with known errors which can be checked using something like a `IsRetry(error)` function; Simplest usage of the dump functions would just stop when interrupted. More advanced users would check the error against the retry function and potentially continue/restart. If the individual errors can be differentiated, then from Cilium logic we can always layer the stats on top by switching on the error type, somewhat similar to the "unknown message" error we did in the perf reader implementation.

lmb:
> That's why we keep track of at least the last entry we saw so we can restart from there rather than defaulting to the kernel behaviour of starting again from the beginning of the map. 

Ah, I need to take a look at that.

> Cilium we really have a whole bunch of metrics that we track which it'd be nice to figure out how to model.

You have the metrics because the kernel behaviour is so clumsy and error prone. I hope that we can fix it at some point, so I'd like the metrics to be as non-intrusive as possible.

> If it triggers some alert or someone notices a metric increase on a dashboard somewhere, then they can let us know that "the bad number went up" and we can consider whether our map handling logic needs to be improved for a future release.

Would it be OK to be more generic, by bumping a counter "iteration failed" and logging the error if it's not one of the known ones? I imagine you'd only ever look at this if it happens often, and then the log messages are there. That way we wouldn't have to distinguish as many errors and are free to modify them as new issues arise or are fixed.

> Quick idea off the top of my head, we could do something like this: Represent errors like dump interruption, key fallback with known errors which can be checked using something like a IsRetry(error) function; Simplest usage of the dump functions would just stop when interrupted. More advanced users would check the error against the retry function and potentially continue/restart.

This is a really interesting idea, thanks! Need to think about this for a bit.

lmb:
```go
// currentKey was returned by GetNextKey() so we know it existed in the map, but it may have been
// deleted by a concurrent map operation. If currentKey is no longer in the map, nextKey will be
// the first key in the map again. Use the nextKey only if we still find currentKey in the Lookup()
// after the GetNextKey() call, this way we know nextKey is NOT the first key in the map.
nextKeyValid := GetNextKeyFromPointers(m.fd, bpfNextKeyPtr, bpfNextKeySize)
err := LookupElementFromPointers(m.fd, bpfCurrentKeyPtr, bpfCurrentKeySize)
```

Consider the following:

```
A                                      B
                                       DELETE_ELEM("foo")
GET_NEXT_KEY("foo") = "bar"
                                       UPDATE_ELEM("foo")
LOOKUP_ELEM("foo")
```

The above race can't be detected by the code. It also breaks if this happens multiple times in a row. Fixing race conditions with more race conditions doesn't work IMO, so I think the best we can do is to detect such a case, and as you suggested, return an error. The caller can then decide to restart, or log an error or increment a metric.

lmb:
I've followed your suggestion and made `Err()` return an error if iteration doesn't finish. I've also looked at the stats you collect. What do you think about the following?

* Started, Finished: can be done by caller
* Lookup: `for i := 0; MapIterator.Next(); i++`? Alternatively, expose `MapIterator.count`.
* LookupFailed: no equivalent, we abort and expose the error via `Err()`
* PrevKeyUnavailable: not used?
* KeyFallback: no equivalent, I believe this can't be detected reliably
* MaxEntries: `Map.ABI().MaxEntries`
* Interrupted, Completed: via `MapIterator.Err()`

joestringer:
> The above race can't be detected by the code. It also breaks if this happens multiple times in a row. Fixing race conditions with more race conditions doesn't work IMO, so I think the best we can do is to detect such a case, and as you suggested, return an error. The caller can then decide to restart, or log an error or increment a metric.

Correct, it can't be detected by the code. However:

* The hash for a given key is consistent. This is the basis of locating the element when you look for it.
* Maps are not being resized during usage
* [Bucket selection is based directly on the number of buckets and the hash](https://github.com/torvalds/linux/blob/8e0d0ad206f08506c893326ca7c9c3d9cc042cef/kernel/bpf/hashtab.c#L412)
* [New elements are added to the head of the list](https://github.com/torvalds/linux/blob/8e0d0ad206f08506c893326ca7c9c3d9cc042cef/kernel/bpf/hashtab.c#L896)

What I'm expecting with the above set of factors is that if the element is deleted then added, worst case you would end up continuing the dump from the beginning of the same bucket and potentially dump some elements from that bucket twice. This is no worse than when the dump restarts from the beginning of the map, because when starting from the beginning of the map not only do you [not know that you've restarted the dump](https://github.com/torvalds/linux/blob/8e0d0ad206f08506c893326ca7c9c3d9cc042cef/kernel/bpf/hashtab.c#L617-L620), furthermore you will also dump the elements at the start of the map twice (or more!) if the dump is restarted from the beginning.

With regards to this race condition skipping elements, there is never a guarantee that you can dump an exact snapshot of the map anyway; elements added earlier in the map from the iterator after you iterated past will not be dumped. This also applies for if the current key was deleted and added and there are other elements that were added earlier in the bucket list during the dump.

> It also breaks if this happens multiple times in a row.

Can you expand on this? What kind of breakage do you have in mind?

I still think it's valuable to attempt continuing the dump from the current bucket rather than just iterating and dumping the beginning of the map over and over until the dump iterates through maxEntries.

joestringer:
> Started, Finished: can be done by caller
:+1: 

> Lookup: for i := 0; MapIterator.Next(); i++? Alternatively, expose MapIterator.count.
:+1: 

> LookupFailed: no equivalent, we abort and expose the error via Err()

See above discussion.

> PrevKeyUnavailable: not used?

Apparently not. I'm not sure whether there was supposed to be some sublety between this and `LookupFailed` but looks like we can ignore this.

> KeyFallback: no equivalent, I believe this can't be detected reliably

For map usage where deletions can occur concurrently, as long as the particularly racy condition you describe above doesn't occur, this can still happen when the current key was deleted. Unfortunately if we don't expose this, we can't know how valuable it really is for a given map usage pattern.

> MaxEntries: Map.ABI().MaxEntries

:+1: 

> Interrupted, Completed: via MapIterator.Err()

I would distinguish the meaning of `Interrupted` vs. Aborted.
* Interrupted: In Cilium this means that we looked for the current element and couldn't find it, then looked for the previous key and also couldn't find that, so we're pretty sure that the next key represents the beginning of the map again.
* Aborted: The `errIterationAborted = errors.New("iteration aborted")` here which says that we iterated `MaxEntries` times over the map and decided to stop the dump so it doesn't run forever. (Equivalent to `Lookup` == `MaxEntries`).

joestringer:
I thought through the case you described again and I see where my previous logic went wrong, and ugh  it's a doozie. I was missing basically that there's no atomicity between the `lookup` and the `get_next_key` which means that you can't say for sure that the state of the map at the time of getting the next key included the current key, so while it appears that a successful continuation of the dump is occurring, it is instead starting from the beginning of the map. This could only be further guaranteed by doing something like providing a creation timestamp hint in an element which could potentially make sense for specific use cases but is way out of scope for a standard dump implementation in this library.

I still think that it's valuable to at least attempt to continue from some point in the map, even if you don't know for sure that you will continue from that point, because it increases the likelihood of making progress and dumping further in the map rather than restarting to the beginning and dumping the beginning multiple times. It doesn't fundamentally change the API in that there are a few guarantees we cannot make:

* We cannot guarantee an entire snapshot of the hashmap is made unless the map is guaranteed to be static
* We cannot guarantee that we will dump each element in the map once
  * A particular element may be dumped twice if concurrent deletions occur
  * A particular element may be missed if it is inserted during dump

lmb:
> Can you expand on this? What kind of breakage do you have in mind?

I meant the case where LOOKUP returns ENOENT, we do NEXT_KEY(prevKey) but prevKey also doesn't exist.

> I would distinguish the meaning of Interrupted vs. Aborted

I'd rather not do that. Yes, we can detect one case where we likely restarted, but we can't detect all of them. It's a bit like the race detector, which is always correct when it finds a race, but has no guarantee that it finds all of them. This is really difficult to explain, and even harder to react to in code. IMO generating errIterationAborted is the best we can do. See also my rant below :P

> I still think that it's valuable to at least attempt to continue from some point in the map, even if you don't know for sure that you will continue from that point, because it increases the likelihood of making progress and dumping further in the map rather than restarting to the beginning and dumping the beginning multiple times.

Definitely, but how can we achieve that when it's not possible to reliably detect a reset? I've pushed a commit with the best I can come up with, and I don't like it very much. The biggest problem is that we can't just apply this band-aid to all map types: ProgramArrays for example will legitimately return ENOENT. So the code has to discern between these different types, which means looking at the kernel to figure out which map type does what. When we add a new map we need to check how it behaves, etc. It's a lot of faff for maybe working around a problem sometimes. The only way to really fix this is in the kernel, and I'm not sure we could do even that without breaking user space.

joestringer:
> > It's valuable to at least attempt to continue from some point in the map

> Definitely, but how can we achieve that when it's not possible to reliably detect a reset?

Best effort.

One use case from a Cilium perspective is that the BPF programs use a hashmap/LRU for handling connection tracking entries, then we have a garbage collector in userspace (even with LRU). The general usage pattern is:
1. We look up the CT map.
1. We look up policy. If a packet passes policy, we create an entry in the CT map that corresponds to the packet's connection (if it doesn't already exist).
1. If we change policy for a connection that was previously allowed to now deny that traffic, then the first outbound packet hits the BPF program and we delete the entry from the map.
1. No other logic would (re-)create the same entry unless the policy changes again (then, see step 2).

For this use case, the race condition we're concerned about in this discussion is not possible due to the way we use the map. However, dumps could still be interrupted by a datapath deletion. So for this case, if we're able to retain N previous keys that we've dumped (and be careful enough on the GC side not to delete those N until they're no longer part of the dump state), then we effectively have N chances to walk back our GC dump and try to find a nearby entry that we can use to continue the dump. We're typically more likely to dump the entire map rather than dumping all the entries at the start of the map over and again. In practice the Cilium code sets N=1 right now and tracks the difference between an iteration failure/`KeyFallback` and fallback iteration failure/`Interrupted`.

What I think this use case demonstrates is that there are cases where the probability of the race condition you mention (and corresponding failure to detect a reset) approaches or reaches zero. A fallback design like what Cilium currently uses means that we are able to handle general cases of large maps with relatively rare deletion of entries that are being iterated. If we had more aggressive element deletion but the same properties around "add", then we may consider increasing N. But it works in practice. And how do we know? Because we count `KeyFallback` (which is OK) and `Interrupted` (which suggests we need to increase N).

A few elements of this which affect the probability of dump interruption / failure:
* Size of the map
* Frequency of deletion; ie likelihood that the element being iterated is deleted
* Whether deleted elements are likely to be recreated and how quickly

Coming back to use cases, do you have a case in mind where your race condition is likely to be executed? Are you thinking that the `B` thread in your example above is executed from BPF? Why would the BPF code delete an element then update it, why not just update it? 

> The biggest problem is that we can't just apply this band-aid to all map types: ProgramArrays for example will legitimately return ENOENT. So the code has to discern between these different types, which means looking at the kernel to figure out which map type does what. When we add a new map we need to check how it behaves, etc. It's a lot of faff for maybe working around a problem sometimes.

Right, you can't do it for all map types. As long as we are pretty clear about the kinds of guarantees we are definitely providing vs. attempted logic, I think we can start with naive implementations and as usage demands improvements, we can drive better implementations from the usage.

> The only way to really fix this is in the kernel, and I'm not sure we could do even that without breaking user space.

Yeah, and as we saw at LPC this year there are ongoing efforts to try to improve dump use cases so I think the onus is on us to figure out what usage patterns we care about, and involve ourselves in those kernel discussions.

map.go
```diff
@@ -517,32 +527,41 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
-			// In either case there isn't much we can do, so just
-			// continue to the next key.
+			if mi.target.abi.Type == Hash {
+				// XXX: We have to audit all map types for their behaviour
+
+				// Don't use nextBytes in NextKeyBytes, since it will
+				// probably reset iteration. Instead, re-try the next key
+				// lookup. If we're lucky mi.prevKey still exists, if we're
+				// unlucky we re-start from the top. There is no way to tell.
+				continue
```
joestringer: Is this trying to protect against the following?

```
next = GetNextKey(prev)
                                       delete(next)
lookup(prev)
```

This is definitely a case worth handling.

The next case I'm interested in is what if `prev` disappears during the arbitrary library user code execution. While for this `next` case, we're mostly hedging against fairly tight race conditions or the process being swapped out in the middle of `GetNextKey` vs `Lookup`, there's a relatively small amount of code involved here - a couple of syscalls and maybe some marshaling. However once we return from this function to the user library, it's entirely up to them.

There's two cases I'm concerned about here:
1. Long delay because there's actually a lot of processing that the user wants to do; then a BPF deletion occurs, or
1. The caller actually assumes that it can delete the entry that is returned through the parameters. 

Case 1 is closely tied to the ongoing discussions so I won't go into any more depth here.

Case 2 says that if this behaviour remains in the long-term implementation, we should at least state this expectation in the API.

lmb: > Is this trying to protect against the following?

Hmm, not sure. The code does `next = GetNextKey(prev)` and then `lookup(next)`. Since `next` becomes `prev` in the next iteration, we know that iteration will reset iff we get ENOENT and the type of the map is hash. Is that what you mean?

> There's two variations I'm concerned about here:

The problem for (1) is that we can't even say that there is a reliable way to dump. Except retrying until you don't get errAborted.

Ugh, I've not thought about (2), that's even worse than the race conditions. The hash map semantics really are garbage. Yeah, we need to document that.

lmb:
> A fallback design like what Cilium currently uses means that we are able to handle general cases of large maps with relatively rare deletion of entries that are being iterated. If we had more aggressive element deletion but the same properties around "add", then we may consider increasing N. But it works in practice. And how do we know? Because we count KeyFallback (which is OK) and Interrupted (which suggests we need to increase N).

I think this is a really good point: users need to know about this problem and design their map modifications around this fact. Then they need an API to control that N, and a way to figure out what happened. From a usability POV this is terrible, because there are a gajillion caveats. Iterating a hash map should be simple and safe by default, not only if you design your system just right. But I don't think the library can provide that with what the kernel gives us.

> I hope you don't mind the slide off into map dump design discussion since there's definitely a bunch of attributes I see in the current implementation we use in Cilium which I'd like to be able to bring into this library as well.

Not at all, this is really useful and valuable. Thanks for taking the time to write such thoughtful reviews. You've convinced me that we need a way to achieve this for power users. What that is going to look like I don't know yet.

I'll remove the fallback behaviour for now, and try to pull our discussion into coherent documentation on this issue.

Issue #10, opened 2019-09-24
Split implementations into foo.go, foo_linux.go


joestringer:
It can be useful to split Linux-specific code into separate files, so that large programs which are written against the library can be compiled and unit-tested on other platforms, with stub implementations for the linux-specific portions. Otherwise as soon as some file in the user project imports the BPF library, it will cause compilation failures for developers on other platforms.

We have a few developers on Cilium who develop natively on OSX which this affects from time to time. Example commit which shifts the code around to achieve this:

https://github.com/cilium/cilium/commit/1d28c269b6c54de1b442a044a9bc76accb0e9058#diff-c2be647780ac9b1d4632bfb1b99a7c0a

lmb:
Yeah annoying, I used to hit that when I was still using macOS as well. Maybe we can get away with something less invasive though. It should be enough to pull all uses of `unix` into a separate file, which we can then stub out for macOS.

PR #12, opened 2019-10-03
use build tags

florianl:

This PR addresses the issue described in https://github.com/cilium/ebpf/issues/10.

It introduces only one new error variable `errNonLinux`. This error is returned, if unix specific functions are called.

### DIFF
--- a/internal/unix/types_linux.go
+++ b/internal/unix/types_linux.go
@@ -1,0 +1,126 @@
+// +build linux
+
+package unix
+
+import (
+	"syscall"
+
+	linux "golang.org/x/sys/unix"
+)
+
+const (
+	ENOENT                   = linux.ENOENT
+	EAGAIN                   = linux.EAGAIN
+	ENOSPC                   = linux.ENOSPC
+	EINVAL                   = linux.EINVAL
+	EPOLLIN                  = linux.EPOLLIN
+	BPF_OBJ_NAME_LEN         = linux.BPF_OBJ_NAME_LEN
+	BPF_TAG_SIZE             = linux.BPF_TAG_SIZE
+	SYS_BPF                  = linux.SYS_BPF
+	F_DUPFD_CLOEXEC          = linux.F_DUPFD_CLOEXEC
+	EPOLL_CTL_ADD            = linux.EPOLL_CTL_ADD
+	EPOLL_CLOEXEC            = linux.EPOLL_CLOEXEC
+	O_CLOEXEC                = linux.O_CLOEXEC
+	O_NONBLOCK               = linux.O_NONBLOCK
+	PROT_READ                = linux.PROT_READ
+	PROT_WRITE               = linux.PROT_WRITE
+	MAP_SHARED               = linux.MAP_SHARED
+	PERF_TYPE_SOFTWARE       = linux.PERF_TYPE_SOFTWARE
+	PERF_COUNT_SW_BPF_OUTPUT = linux.PERF_COUNT_SW_BPF_OUTPUT
+	PerfBitWatermark         = linux.PerfBitWatermark
+	PERF_SAMPLE_RAW          = linux.PERF_SAMPLE_RAW
+	PERF_FLAG_FD_CLOEXEC     = linux.PERF_FLAG_FD_CLOEXEC
+)
+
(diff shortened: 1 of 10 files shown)


lmb:
Hi Florian, thanks for your contribution! Have you tried the idea that I outlined in https://github.com/cilium/ebpf/issues/10#issuecomment-534993145 ?

Splitting the implementation into different files is fairly invasive and difficult to maintain. Hence my idea to make an `internal/unix` package which re-exports the constants / functions we need. That would reduce the changes a lot.

florianl:
Hi Lorenz,
thanks for your feedback.

I'll update the PR so it will have an `internal/unix` package as you suggested. So far, I had the feeling, that the different files are less invasive. But this is only a subjective personal view.

florianl:
Hi Lorenz,
I've updated the PR according to your suggestions.
And I totally agree, this way, the change is way less invasive.

internal/unix/types_other.go
```diff
@@ -0,0 +1,57 @@
...
+	Fsid    [2]int32
+	Namelen int64
+	Frsize  int64
+	Flags   int64
+	Spare   [4]int64
+}
+
+// Syscall is a wrapper
+func Syscall(trap, a1, a2, a3 uintptr) (r1, r2, err uintptr) {
+	return 0, 0, 0
```
lmb: Let's make sure this returns an error:

```suggestion
	return 0, 0, -1
```

lmb:
I just tried the following:

```
$ GOOS=darwin go build ./...
# github.com/cilium/ebpf
./map.go:558:27: invalid operation: "github.com/pkg/errors".Cause(err) == unix.ENOENT (mismatched types error and int)
./prog.go:119:33: invalid operation: "github.com/pkg/errors".Cause(err) == unix.ENOSPC (mismatched types error and int)
./prog.go:128:34: invalid operation: "github.com/pkg/errors".Cause(nerr) == unix.ENOSPC (mismatched types error and int)
./prog.go:299:28: invalid operation: "github.com/pkg/errors".Cause(err) == unix.EINVAL (mismatched types error and int)
./syscalls.go:192:10: invalid operation: err == unix.EAGAIN (mismatched types error and int)
./syscalls.go:392:7: cannot use errNo (type uintptr) as type error in assignment:
        uintptr does not implement error (missing Error method)
```

I think making the errnos of type syscall.Errno should be enough.

internal/unix/types_other.go
florianl: i've tried this, but get the following error:
`constant -1 overflows uintptr`

florianl: is it ok, to return `1` instead?
according to `include/asm/errno.h` `1` indicates `Operation not permitted`.

florianl: i've updated the PR and return now `syscall.Errno(1)`.

internal/unix/types_linux.go
```diff
@@ -0,0 +1,126 @@
...
+	MAP_SHARED               = linux.MAP_SHARED
+	PERF_TYPE_SOFTWARE       = linux.PERF_TYPE_SOFTWARE
+	PERF_COUNT_SW_BPF_OUTPUT = linux.PERF_COUNT_SW_BPF_OUTPUT
+	PerfBitWatermark         = linux.PerfBitWatermark
+	PERF_SAMPLE_RAW          = linux.PERF_SAMPLE_RAW
+	PERF_FLAG_FD_CLOEXEC     = linux.PERF_FLAG_FD_CLOEXEC
+)
+
+// Statfs_t is a wrapper
+type Statfs_t linux.Statfs_t
```
lmb: If you make these types an alias you should be able to skip the casts + creating a new `[]EpollEvent`:

```suggestion
type Statfs_t = linux.Statfs_t
```
etc.

See https://blog.golang.org/go1.9 (the bit on type aliases)

internal/unix/types_other.go
lmb: Yeah, that sounds good. I was looking at `man syscall` which says it will return -1, sorry.

internal/unix/types_linux.go
florianl: done in the updated PR. Thanks for the hint to type aliases!

PR #15, opened 2019-10-11
map, program: allow creation from raw FD

lmb:
BPF maps and programs can be sent around via SCM_RIGHTS, so users should
be able to interact with them.

### DIFF
--- a/map.go
+++ b/map.go
@@ -55,6 +55,23 @@ type Map struct {
 	fullValueSize int
 }
 
+// NewMapFromFD creates a map from a raw fd.
+//
+// You should not use fd after calling this function.
+func NewMapFromFD(fd int) (*Map, error) {
+	if fd < 0 {
+		return nil, errors.New("invalid fd")
+	}
+	bpfFd := newBPFFD(uint32(fd))
+
+	abi, err := newMapABIFromFd(bpfFd)
+	if err != nil {
+		bpfFd.forget()
+		return nil, err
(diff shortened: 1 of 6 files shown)


map.go
```diff
@@ -55,6 +55,31 @@ type Map struct {
 	fullValueSize int
 }
 
+// NewMapFromFD creates a map from a raw fd.
+//
+// name must be a valid BPF object name.
+//
+// You should not use fd after calling this function.
```
joestringer: Question for semantics: Which of the following would you expect when writing against this interface?

```
fd = MyCodeThatOpensAnFdForMap(...);
map, err := bpf.NewMapFromFD(fd, ...)
if err != nil {
    syscall.Close(fd)
    log("Failed to open map: %s", err)
}
```
Or
```
fd = MyCodeThatOpensAnFdForMap(...);
map, err := bpf.NewMapFromFD(fd, ...)
if err != nil {
    // Fd was already closed by the NewMapFromFD() function
    log("Failed to open map: %s", err)
}
```

I would expect that the FD is still "ours" in the error case, and the fd is now "stolen" in the success case. Particularly if there ends up being any other logic in between the fd open (`defer()`s, function layers, etc).

lmb: Fixed the code to follow the first style.

lmb:
I removed the "name" argument, since we should be able to infer it from the fd itself.

PR #19, opened 2019-10-17
fix compilation on 32-bit targets

### LINKED ISSUE #18: Can't be compiled for 32-bit ARM
https://github.com/opencontainers/runc/pull/2145#issuecomment-543161459

```
CGO_ENABLED=1 GOARCH=arm GOARM=6 CC=arm-linux-gnueabi-gcc go build -buildmode=pie  -ldflags "-X main.gitCommit="bf9519326d3dcc4a78f3cddbc54ac7a78a0aa948" -X main.version=1.0.0-rc9+dev " -tags "seccomp apparmor selinux ambient" -o runc-armel .
# github.com/opencontainers/runc/vendor/github.com/cilium/ebpf
vendor/github.com/cilium/ebpf/syscalls.go:285:17: constant 3405662737 overflows int32
Makefile:125: recipe for target 'localcross' failed
```

This line does not compile for `GOARCH=arm GOARM=6`
https://github.com/cilium/ebpf/blob/60d34524c5b34180514b1aae150a53bc41026d81/syscalls.go#L285

lmb:
The magic number for bpffs doesn't fit in Statfs_t.Type on 32bit arch,
which leads to a compilation error. Looking at the kernel source, the
thing to do is to treat the magic number as unsigned.

Fixes #18

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -19,9 +19,11 @@ blocks:
           - sudo apt-get -y update
           - sudo apt-get install -y qemu-system-x86
       jobs:
-      - name: Test build on non-Linux
+      - name: Test building on other OS and arch
         commands:
           - GOOS=darwin go build ./...
+          - GOARCH=arm GOARM=6 go build ./...
+          - GOARCH=arm64 go build ./...
       - name: Test on 5.0.13
         commands:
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
thanks for the quick fix 👍

PR #21, opened 2019-10-17
syscalls: fix pointer padding on 32bit big endian arches

### LINKED ISSUE #17: ptr_32_le and ptr_32_be are identical
https://github.com/cilium/ebpf/blob/60d34524c5b34180514b1aae150a53bc41026d81/ptr_32_le.go
https://github.com/cilium/ebpf/blob/60d34524c5b34180514b1aae150a53bc41026d81/ptr_32_be.go

I don't have big-endian machine, but maybe _be.go is wrong?

lmb:
Given a 32 bit big endian pointer {01, 02, 03, 04}, it's 64 bit
representation should be {00, 00, 00, 00, 01, 02, 03, 04}. Therefore
the current order of syscallPtr is wrong.

Fixes #17

### DIFF
--- a/ptr_32_be.go
+++ b/ptr_32_be.go
@@ -9,6 +9,6 @@ import (
 // ptr wraps an unsafe.Pointer to be 64bit to
 // conform to the syscall specification.
 type syscallPtr struct {
-	ptr unsafe.Pointer
 	pad uint32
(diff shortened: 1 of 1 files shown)


lmb:
Unfortunately there is no way for me to test this, so it's a bit of a best effort fix.

Issue #22, opened 2019-10-18
add support for BPF_PROG_ATTACH


lmb:
See #20. We already have attach types in the library, but are missing the syscall wrapper for it.

PR #25, opened 2019-10-24
Add prog Attach/Detach

florianl:

This PR adds the syscall wrappers to attach/detach a program as mentioned in https://github.com/cilium/ebpf/issues/22.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -413,6 +413,16 @@ func (p *Program) MarshalBinary() ([]byte, error) {
 	return buf, nil
 }
 
+// Attach a Program to a container object fd
+func (p *Program) Attach(fd uint32, bfpProgType ProgramType, flags uint32) error {
+	return bpfProgAttach(p.fd, fd, uint32(bfpProgType), flags)
+}
+
+// Detach a Program from a container object fd
+func (p *Program) Detach(fd uint32, bfpProgType ProgramType, flags uint32) error {
+	return bpfProgDetach(p.fd, fd, uint32(bfpProgType), flags)
+}
+
 // LoadPinnedProgram loads a Program from a BPF file.
 //
 // Requires at least Linux 4.13, use LoadPinnedProgramExplicit on
--- a/prog.go
+++ b/prog.go
@@ -425,10 +425,10 @@ func (p *Program) Attach(fd int, attachType AttachType, attachFlags AttachFlags)
 	}
 
 	attr := bpfProgAlterAttr{
-		targetFd:   uint32(fd),
-		alterBpfFd: pfd,
-		alterType:  uint32(attachType),
-		alterFlags: uint32(attachFlags),
+		targetFd:    uint32(fd),
+		attachBpfFd: pfd,
+		attachType:  uint32(attachType),
+		attachFlags: uint32(attachFlags),
 	}
 
 	return bpfProgAlter(_ProgAttach, &attr)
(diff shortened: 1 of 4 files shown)


prog.go
```diff
@@ -413,6 +413,16 @@ func (p *Program) MarshalBinary() ([]byte, error) {
 	return buf, nil
 }
 
+// Attach a Program to a container object fd
+func (p *Program) Attach(fd uint32, bfpProgType ProgramType, flags uint32) error {
```
lmb: `fd` should be `int`, and then we should check for `fd < 0`, to be consistent with other fds in the API. I think `flags` could be `type AttachFlags uint32`. We don't need to define all the flags now, but we leave this option open to us. User code can do `foo = ebpf.AttachFlags(0xbar)` in the meantime.

I'm a bit fuzzy what `bpfProgType` is for. Shouldn't it be `AttachType`?

prog.go
```diff
@@ -413,6 +413,16 @@ func (p *Program) MarshalBinary() ([]byte, error) {
...
+	return bpfProgAttach(p.fd, fd, uint32(bfpProgType), flags)
+}
+
+// Detach a Program from a container object fd
+func (p *Program) Detach(fd uint32, bfpProgType ProgramType, flags uint32) error {
```
lmb: Same comments as above.

syscalls.go
```diff
@@ -205,6 +212,50 @@ func bpfProgLoad(attr *bpfProgLoadAttr) (*bpfFD, error) {
...
+	attr := bpfProgAlterAttr{
+		targetFd:   targetFd,
+		alterBpfFd: fd,
+		alterType:  alterType,
+		alterFlags: alterFlags,
+	}
+
+	for {
+		_, err := bpfCall(_ProgAttach, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
+		if err == unix.EAGAIN {
```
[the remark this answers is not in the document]
florianl: I also don't know of a case, where attach/detach calls can be interrupted. As I was not sure, I did it this way.
But I'll simplify it, as you suggest.

syscalls.go
```diff
@@ -205,6 +212,50 @@ func bpfProgLoad(attr *bpfProgLoadAttr) (*bpfFD, error) {
 	}
 }
 
+func bpfProgAttach(m *bpfFD, targetFd, alterType, alterFlags uint32) error {
```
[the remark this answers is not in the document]
florianl: I was not sure how to call it. `bpfProgAlter(cmd, *bpfProgAlterAttr)` sounds good for me and I will update it.

prog.go
florianl: You are correct, it should be `AttachType` - sorry for the mistake!

florianl:
In the updated PR I've addressed the issues as suggested.

For the test, I'm still trying to find a solution. Most use cases of BPF_PROG_ATTACH I'm aware of, are in combination with cgroups. But I'm not sure if it is good to add such a dependency just for tests.

florianl:
The just added test case I heavily inspired by https://blog.cloudflare.com/sockmap-tcp-splicing-of-the-future/.

syscalls.go
```diff
@@ -173,6 +173,13 @@ type bpfProgTestRunAttr struct {
 	duration    uint32
 }
 
+type bpfProgAlterAttr struct {
+	targetFd   uint32
+	alterBpfFd uint32
```
AkihiroSuda: s/alter/attacj/g ?
https://github.com/torvalds/linux/blob/v5.3/include/uapi/linux/bpf.h#L423-L427

florianl: as the same struct is used for attaching and detaching, I've changes it to alter. But I will revert it, to keep it the way it is in the kernel.

prog.go
```diff
@@ -414,13 +414,45 @@ func (p *Program) MarshalBinary() ([]byte, error) {
 }
 
 // Attach a Program to a container object fd
-func (p *Program) Attach(fd uint32, bfpProgType ProgramType, flags uint32) error {
-	return bpfProgAttach(p.fd, fd, uint32(bfpProgType), flags)
+func (p *Program) Attach(fd int, attachType AttachType, attachFlags AttachFlags) error {
```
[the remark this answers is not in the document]
florianl: `type` is an identifier in the language. so, I wanted to avoid confusions here. any other suggestions for naming it?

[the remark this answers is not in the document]
AkihiroSuda: `typ` seems the common convention to avoid conflict with `type`: https://golang.org/pkg/go/types/#IsInterface

lmb:
Thanks!

PR #29, opened 2019-11-05
Add support for pausing/resuming perf readers

joestringer:
This implementation is inspired by the upstream Cilium commit
[ce6244109bd0](http://github.com/cilium/cilium/commit/ce6244109bd0) ("cilium: add Mute/Unmute function for perf RB") by Daniel
Borkmann, to support enabling and disabling perf ringbuffer
notifications at runtime.

On the way, I fixed running the perf unit tests via `sudo go test ./...`.

### DIFF
--- a/internal/unix/types_linux.go
+++ b/internal/unix/types_linux.go
@@ -30,6 +30,7 @@ const (
 	PerfBitWatermark         = linux.PerfBitWatermark
 	PERF_SAMPLE_RAW          = linux.PERF_SAMPLE_RAW
 	PERF_FLAG_FD_CLOEXEC     = linux.PERF_FLAG_FD_CLOEXEC
+	RLIM_INFINITY            = linux.RLIM_INFINITY
 )
 
 // Statfs_t is a wrapper
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -205,10 +205,6 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 		}
 		rings = append(rings, ring)
 
-		if err := array.Put(uint32(i), uint32(ring.fd)); err != nil {
-			return nil, errors.Wrapf(err, "could't put event fd for CPU %d", i)
-		}
-
 		if err := addToEpoll(epollFd, ring.fd, len(rings)-1); err != nil {
 			return nil, err
 		}
@@ -237,6 +233,9 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 		epollEvents: make([]unix.EpollEvent, len(rings)+1),
 		epollRings:  make([]*perfEventRing, 0, len(rings)),
 		closeFd:     closeFd,
+	}
+	if err = pr.Resume(); err != nil {
+		return nil, err
 	}
 	runtime.SetFinalizer(pr, (*Reader).Close)
 	return pr, nil
@@ -339,6 +338,38 @@ func (pr *Reader) Read() (Record, error) {
 
 		return record, err
 	}
+}
+
+// Pause stops all notifications from this perf reader.
+//
+// Subsequent calls to Read will block until a call to Resume.
+func (pr *Reader) Pause() error {
+	pr.mu.Lock()
+	defer pr.mu.Unlock()
+
+	for i := 0; i < len(pr.rings); i++ {
+		if err := pr.array.Delete(uint32(i)); err != nil && !ebpf.IsNotExist(err) {
+			return errors.Wrapf(err, "could't delete event fd for CPU %d", i)
+		}
+	}
+
+	return nil
+}
+
+// Resume allows this perf reader to emit notifications.
+//
+// Subsequent calls to Read will block until the next event notification.
+func (pr *Reader) Resume() error {
+	pr.mu.Lock()
+	defer pr.mu.Unlock()
+
+	for i := 0; i < len(pr.rings); i++ {
+		if err := pr.array.Put(uint32(i), uint32(pr.rings[i].fd)); err != nil {
+			return errors.Wrapf(err, "could't put event fd for CPU %d", i)
+		}
+	}
+
+	return nil
 }
 
 type temporaryError interface {
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -133,6 +133,9 @@ func readRawSample(rd io.Reader) ([]byte, error) {
 // Reader allows reading bpf_perf_event_output
 // from user space.
 type Reader struct {
+	// mu protects read/write access to the Reader structure with the
+	// exception of 'pauseFds', which is protected by 'pauseMu'.
+	// If locking both 'mu' and 'pauseMu', 'mu' must be locked first.
 	mu sync.Mutex
 
 	// Closing a PERF_EVENT_ARRAY removes all event fds
@@ -147,6 +150,12 @@ type Reader struct {
 	closeFd int
 	// Ensure we only close once
 	closeOnce sync.Once
+
+	// pauseFds are a copy of the fds in 'rings', protected by 'pauseMu'.
+	// These allow Pause/Resume to be executed independently of any ongoing
+	// Read calls, which would otherwise need to be interrupted.
+	pauseMu  sync.Mutex
+	pauseFds []int
 }
 
 // ReaderOptions control the behaviour of the user
@@ -179,9 +188,10 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 	}
 
 	var (
-		fds   = []int{epollFd}
-		nCPU  = int(array.ABI().MaxEntries)
-		rings = make([]*perfEventRing, 0, nCPU)
+		fds      = []int{epollFd}
+		nCPU     = int(array.ABI().MaxEntries)
+		rings    = make([]*perfEventRing, 0, nCPU)
+		pauseFds = make([]int, 0, nCPU)
 	)
 
 	defer func() {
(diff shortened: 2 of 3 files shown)


internal/unix/types_linux.go
```diff
@@ -30,6 +30,7 @@ const (
 	PerfBitWatermark         = linux.PerfBitWatermark
 	PERF_SAMPLE_RAW          = linux.PERF_SAMPLE_RAW
 	PERF_FLAG_FD_CLOEXEC     = linux.PERF_FLAG_FD_CLOEXEC
+	RLIM_INFINITY            = linux.RLIM_INFINITY
```
joestringer: I think it makes sense to swap the whole codebase over to `unix.RLIM_INFINITY` to get rid of `math` dependency, but I didn't want to do it in this PR since it's kinda unrelated. I can follow up on that later.

perf/reader.go
```diff
@@ -341,6 +340,30 @@ func (pr *Reader) Read() (Record, error) {
 	}
 }
 
+// Mute stops all notifications from this perf reader.
+//
+// Subsequent calls to Read will block until a call to Unmute.
+func (pr *Reader) Mute() error {
+	for i := 0; i < len(pr.rings); i++ {
```
joestringer: Regards handling failures, I could see arguments for either:
* Collecting all failures and returning a slice of errors, or
* Walking back through the successful operations here and undoing them (which could also technically fail...)

In Cilium we ignore the errors in `Mute()` and return one back like this in `Unmute()`. I did one extra enhancement in this version where I check against `ENOENT`, which also has the nice property of making `Mute()` idempotent without any additional fields to hold mute/unmute state.

Happy to take direction here.

perf/reader_test.go
```diff
@@ -287,6 +302,82 @@ func TestReadRecord(t *testing.T) {
...
+	select {
+	case err, ok := <-errChan:
+		// Failure: Mute was unsuccessful.
+		if ok {
+			t.Fatal(err)
+		} else {
+			t.Fatal("received notification on muted reader")
+		}
+	case <-time.After(readTimeout):
+		// Success
```
joestringer: We did actually already ensure that the `prog.Test()` run failed to write the perf event, so I could see an argument that we don't need to wait around to ensure that the event never arrived here. Downside of current implementation is that it will always take 250ms if the test is passing.

internal/unix/types_linux.go
lmb: Yeah, that's fine

perf/reader.go
```diff
@@ -341,6 +340,30 @@ func (pr *Reader) Read() (Record, error) {
...
+		if err := pr.array.Delete(uint32(i)); err != nil && errors.Cause(err) != unix.ENOENT {
```
lmb: Use `IsNotFound` here?

perf/reader.go
lmb: Slices of error are always difficult to use IMO, I think you made the right call here. Undoing also doesn't seem like it would help much. What does Cilium currently do if Mute / Unmute fail? Does it keep on running or does it bail out?

perf/reader_test.go
```diff
@@ -298,6 +302,82 @@ func TestReadRecord(t *testing.T) {
...
+	errChan := make(chan error)
+	go func() {
+		// Read one notification then send any errors and exit.
+		_, err := rd.Read()
+		errChan <- err
+		close(errChan)
+	}()
+	ret, _, err = prog.Test(make([]byte, 14))
+	if err == nil && ret == 0 {
+		t.Fatal("Unexpectedly rote sample while muted")
```
lmb: Nit: typo `rote`

perf/reader_test.go
```diff
@@ -298,6 +302,82 @@ func TestReadRecord(t *testing.T) {
...
+	}
+	if _, err := rd.Read(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Mute. No notification should trigger.
+	if err = rd.Mute(); err != nil {
+		t.Fatal(err)
+	}
+	errChan := make(chan error)
```
lmb: Nit: `make(chan error, 1)`?

perf/reader_test.go
```diff
@@ -298,6 +302,82 @@ func TestReadRecord(t *testing.T) {
...
+	go func() {
+		// Read one notification then send any errors and exit.
+		_, err := rd.Read()
+		errChan <- err
+		close(errChan)
```
lmb: I think it should be OK to skip the close, then we have to handle one less case in the select statement below.

perf/reader_test.go
lmb: True. I think it's nice that we're testing the whole thing. Can you throw in a `t.Parallel()` at the start? 250ms isn't a huge deal I think.

joestringer:
> The naming of Mute / Unmute: it seems weird to be able to do this to a Reader. Maybe Pause and Resume? This is a nit really.

I was thinking the same thing, but no ideas came to mind about better alternatives. I like `Pause` and `Resume`.

> Could we use PERF_EVENT_IOC_ENABLE with PERF_IOC_FLAG_GROUP to work around the multiple errors issue? See http://www.man7.org/linux/man-pages/man2/perf_event_open.2.html under perf_event ioctl calls.

I don't think that this helps, we actually had support for this in Cilium as well but it operates on a per-ring basis, which is the same as the fds for the map. So at least at a kernel interface level there's the same tradeoff. I suspect that in practice neither approach is likely to generate an error for these paths anyway though...  ¯\_(ツ)_/¯

Another option I noticed while exploring this was `PERF_EVENT_IOC_PAUSE_OUTPUT` (v4.7+), but again practically I'm not sure that there's any difference.

lmb:
Yeah, I had misunderstood the purpose of `group_fd` in `perf_event_open`. It's used for to group events on the same cpu, not across multiple like we do.

`PERF_EVENT_IOC_PAUSE_OUTPUT` is interesting:
```
The discarded samples are considered
lost, and cause a PERF_RECORD_LOST sample to be generated when
possible.  An overflow signal may still be triggered by the
discarded sample even though the ring-buffer remains empty.
```

Should `Pause` generate lost samples or not? I guess that will determine whether to use the ioctl or just removing entries from the map.

joestringer:
> Should `Pause` generate lost samples or not? I guess that will determine whether to use the ioctl or just removing entries from the map.

The `lost` samples to me are a sign that the current structure of your event handling (page size, work based on each event, etc.) is not able to keep up, and you may not intend that. Pausing the event output says that you're deliberately choosing not to handle the events, so counting lost events in that case means that the lost events counter now counts both expected and unexpected event discards.

Looking on the kernel side, if we use `PERF_EVENT_IOC_PAUSE_OUTPUT` then it will end up returning `-ENOSPC` from the `bpf_perf_event_output()` helper, meaning it is indistinguishable from actually running out of space on the buffer:

https://github.com/torvalds/linux/blob/26bc672134241a080a83b2ab9aa8abede8d30e1c/kernel/events/ring_buffer.c#L173-L176

If you contrast this with the "fds removal" approach, that will return `-ENOENT`:

https://github.com/torvalds/linux/blob/26bc672134241a080a83b2ab9aa8abede8d30e1c/kernel/trace/bpf_trace.c#L436-L438

This gives the BPF side a clearer signal that the events buffer is not currently listening, so the BPF code can be written to handle this particular case differently from generic event drops. Plus, there's a bit less actual code being executed if we delete the fds from the map---kernel-side the bpf helper first does the map lookup, then proceeds towards the generic event handling code.

lmb:
SGTM, can you push your latest changes?

joestringer:
Weird, not sure why github didn't pick it up before. I re-pushed.

perf/reader.go
```diff
@@ -341,6 +340,30 @@ func (pr *Reader) Read() (Record, error) {
 	}
 }
 
+// Pause stops all notifications from this perf reader.
+//
+// Subsequent calls to Read will block until a call to Resume.
+func (pr *Reader) Pause() error {
```
lmb: Sorry that I missed this earlier, but Pause and Resume have to take the mutex, since Close will modify the values.

joestringer: Turned out that this was a bit more involved than I initially thought, as the attempt to lock the mutex from a `Pause()`/`Resume()` call will block as long as `Read()` is blocking. So I needed to come up with an interrupt mechanism. Please take another look at this.

lmb: Is Github acting up again? I can't see the interrupt mechanism in your PR.

joestringer: I ended up hacking away on it for a while but didn't come up with a solution that works reliably yet. I'll mention you directly on the main PR thread below when I post it, and make sure it's available for you to check out.

lmb: Got it. I stared at the code for a bit as well, maybe the easiest solution is to just duplicate the raw fds into a separate array?

joestringer: So you mean have a distinct copy / distinct locks for the pause/resume? That may solve the issues I'm hitting, let me experiment..

joestringer:
@lorenz Thanks for the suggestion, I've pushed the changes out as a brand new patch. PTAL.

perf/reader.go
```diff
@@ -261,10 +276,12 @@ func (pr *Reader) Close() error {
 			return
 		}
 
-		// Acquire the lock. This ensures that Read
-		// isn't running.
+		// Acquire the locks. This ensures that Read, Pause and Resume
+		// aren't running.
 		pr.mu.Lock()
```
lmb: Nit: it's easier to see the unlock order is correct if you do this:
```
pr.mu.Lock()
defer pr.mu.Unlock()

pr.pauseMu.Lock()
defer pr.pauseMu.Unlock()
```

perf/reader.go
```diff
@@ -340,14 +358,17 @@ func (pr *Reader) Read() (Record, error) {
...
 //
 // Subsequent calls to Read will block until a call to Resume.
 func (pr *Reader) Pause() error {
-	pr.mu.Lock()
-	defer pr.mu.Unlock()
+	pr.pauseMu.Lock()
+	defer pr.pauseMu.Unlock()
 
-	for i := 0; i < len(pr.rings); i++ {
+	for i := 0; i < len(pr.pauseFds); i++ {
```
lmb: Check for `pauseFds == nil` and return ErrClosed here an in Resume?

joestringer:
I addressed your feedback and merged. Thanks for the reviews!

PR #34, opened 2019-11-13
btf: allow annotating Programs with BTF

lmb:
The BPF Type Format provides metadata about programs, maps and
the kernel itself. It allows for better introspection at run-time
and enables some advanced features like spin locks in map values.

Add support for reading BTF from ELF files, and passing file and
line information to the kernel. Map related BTF functionality is
not implemented.


lmb:
The test is currently failing on 4.19. I suspect this is because the BTF is too "new" for the kernel to understand. I'll take a look in the meantime.

lmb:
I was right, the error was due to a BTF_KIND_FUNC which 4.19 does not understand. I fixed this by changing the feature detection, which means BTF support currently requires a kernel >= 5.1.

lmb:
I've been working on this more, I need to make it reviewable again. Closing the PR for now.

PR #36, opened 2019-12-02
Run unit tests on 4.9 kernels

### LINKED ISSUE #32: Run CI on 4.9 kernel
Cilium requires at least Linux 4.9, which is why we added #5 We should run CI on 4.9 as well to make sure we don't regress. There are some tests which will fail, since the feature that is being tested is not supported on the old kernel. For now we probably need to skip those tests.

I've already built a 4.9 kernel: https://github.com/newtools/ci-kernels/blob/master/linux-4.9.198.bz

lmb:
Cilium proper requires at least Linux 4.9, so we need to support it in turn. This means we should run tests on a 4.9 kernel.

However, some features we are testing are not available on such old kernels. Add some helpers which allow us to skip tests if they are for an unsupported feature. This is somewhat dangerous, because if our feature test is buggy we end up skipping our tests, rather than reporting a failure. Reduce the risk of this by doing a sanity check against the currently running kernel: if the feature should be available, but the test indicates it is not we error out.

As a nice side effect, users of the library will get nicer error messages if they try to use some unsupported features.

Fixes #32


lmb:
This pulls out one commit from #35.

abi_test.go
```diff
@@ -1,6 +1,7 @@
 package ebpf
 
 import (
+	"github.com/cilium/ebpf/internal/testutils"
```
arthurfabre: NIT: This doesn't match the order / grouping of the rest of the imports

internal/feature.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (ufe *UnsupportedFeatureError) Error() string {
+	return fmt.Sprintf("%s not supported (requires >= %s)", ufe.Name, ufe.MinimumVersion)
+}
+
+// FeatureTest wraps a function so that it is run at most once.
+//
+// name should identify the tested feature, while version must be in the
+// form Major.Minor[.Patch].
```
arthurfabre: Why is `Patch` optional? Is there an issue with declaring `5.4` as `5.4.0`?

lmb: No issue, it's just a convenience. 99% of the time it'll be "Major.Minor".

abi_test.go
lmb: Fixed, thanks.

PR #40, opened 2019-12-17
map: add LookupAndDelete

florianl:

This PR implements LookupAndDelete() for maps.

### DIFF
--- a/map.go
+++ b/map.go
@@ -235,7 +235,7 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
 	if valueBytes == nil {
 		return nil
 	}
-	return m.valueProcessing(valueOut, valueBytes)
+	return m.unmarshalValue(valueOut, valueBytes)
 }
 
 // LookupAndDelete retrieves and deletes a value from a Map.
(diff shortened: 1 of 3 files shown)


map.go
```diff
@@ -235,7 +235,29 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
...
+		return err
+	}
 
+	if valueBytes == nil {
+		return nil
+	}
+	return m.valueProcessing(valueOut, valueBytes)
+}
+
+func (m *Map) valueProcessing(valueOut interface{}, valueBytes []byte) error {
```
florianl: I'm not sure about naming conventions here. As there needs to be the same post processing for values returned from the map for `Lookup()` and `LookupAndDelete()` and I wanted to avoid duplicating code.

lmb: How about `unmarshalValue`?

map.go
```diff
@@ -235,7 +235,29 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
...
 	if m.abi.Type.hasPerCPUValue() {
```
lmb: Is it valid to call LookupAndDelete on a per-CPU map?

map.go
```diff
@@ -235,7 +235,29 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
...
+// LookupAndDelete retrieves and deletes a value from a Map.
+//
+// Calls Close() on valueOut if it is of type **Map or **Program,
+// and *valueOut is not nil.
+//
+// Returns an error if the key doesn't exist, see IsNotExist.
+func (m *Map) LookupAndDelete(key, valueOut interface{}) error {
+	valuePtr, valueBytes := makeBuffer(valueOut, m.fullValueSize)
+
+	if err := m.lookup(key, valuePtr); err != nil {
```
lmb: Shouldn't this be `lookupAndDelete`?

map.go
```diff
@@ -299,6 +321,16 @@ func (m *Map) lookup(key interface{}, valueOut internal.Pointer) error {
 	return errors.WithMessage(err, "lookup failed")
 }
 
+func (m *Map) lookupAndDelete(key interface{}, valueOut internal.Pointer) error {
```
lmb: Is there a reason this code isn't just in `LookupAndDelete`?

map_test.go
```diff
@@ -67,6 +67,27 @@ func TestMap(t *testing.T) {
...
+
+	var v uint32
+	if err := m.LookupAndDelete(uint32(0), &v); err != nil {
+		t.Fatal("Can't lookup 0:", err)
+	}
+	if v != 42 {
+		t.Error("Want value 42, got", v)
+	}
+
+	if err := m.Lookup(uint32(0), &v); err != nil {
```
lmb: Check that `IsNotExist(err)`?

map.go
florianl: i've renamed it to `unmarshalValue`.

map.go
florianl: :see_no_evil: 
yes - you're right

map.go
florianl: I wanted to follow the existing structure `lookup` does.
So, I've moved this code into `LookupAndDelete`

map.go
florianl: As far as I can tell, it's valid. But maybe there is something, I don't know.

florianl:
I'm closing this PR for now and will have a closer look at `bpfMapLookupAndDelete()` on the kernel side.

PR #41, opened 2020-01-03
Push/pop/peek

florianl:
This PR adds the bpfMapLookupAndDelete functionality - which only is implemented for stacks and queues on kernel side [1].

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/bpf/syscall.c#n1219

### DIFF
--- a/map.go
+++ b/map.go
@@ -274,6 +274,22 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
 	}
 }
 
+// LookupAndDelete retrieves and deletes a value from a Map.
+func (m *Map) LookupAndDelete(key, valueOut interface{}) error {
+	valuePtr, valueBytes := makeBuffer(valueOut, m.fullValueSize)
+
+	keyPtr, err := marshalPtr(key, int(m.abi.KeySize))
+	if err != nil {
+		return errors.WithMessage(err, "can't marshal key")
+	}
+
+	if err := bpfMapLookupAndDelete(m.fd, keyPtr, valuePtr); err != nil {
+		return errors.WithMessage(err, "lookup and delete and delete failed")
+	}
+
+	return unmarshalBytes(valueOut, valueBytes)
+}
+
 // LookupBytes gets a value from Map.
 //
 // Returns a nil value if a key doesn't exist.
--- a/map_test.go
+++ b/map_test.go
@@ -144,6 +144,63 @@ func createArray(t *testing.T) *Map {
 	return m
 }
 
+func TestMapQueue(t *testing.T) {
+
+	testutils.SkipOnOldKernel(t, "4.20", "map type queue")
+
+	m, err := NewMap(&MapSpec{
+		Type:       Queue,
+		ValueSize:  4,
+		MaxEntries: 2,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer m.Close()
+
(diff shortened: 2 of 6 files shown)


map_test.go
```diff
@@ -144,6 +144,61 @@ func createArray(t *testing.T) *Map {
 	return m
 }
 
+func TestMapQueue(t *testing.T) {
+
+	testutils.SkipOnOldKernel(t, "4.20", "map type queue")
```
florianl: the map type queue and stack was added in [1]. So this feature is not available for older kernel.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/include/uapi/linux/bpf.h?id=f1a2e44a3aeccb3ff18d3ccc0b0203e70b95bd92

lmb: Sounds good, thanks for checking.

florianl:
Reading your points, I agree with you to keep the API more stable. So I will update the PR accordingly.

map_test.go
```diff
@@ -144,6 +144,61 @@ func createArray(t *testing.T) *Map {
...
+		t.Error("Want value 42, got", v)
+	}
+
+	if err := m.Lookup(nil, &v); err != nil {
+		t.Fatal("Can't lookup element:", err)
+	}
+	if v != 4242 {
+		t.Error("Want value 4242, got", v)
+	}
+	if err := m.LookupAndDelete(nil, &v); err != nil {
```
lmb: You need to reset v here.

```suggestion
	v = 0
	if err := m.LookupAndDelete(nil, &v); err != nil {
```

marshalers.go
```diff
@@ -14,6 +14,10 @@ import (
 )
 
 func marshalPtr(data interface{}, length int) (internal.Pointer, error) {
+	if length == 0 {
```
lmb: I think I'd prefer to check data being nil, since this is what the user passes. Maybe something like the following?

```
	if data == nil {
		if length != 0 {
			return internal.Pointer{}, errors.New("...")
```

Currently doing `Lookup(&garbage, &v)` doesn't return an error.

map.go
```diff
@@ -274,6 +274,20 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
 	}
 }
 
+// LookupAndDelete retrieves and deletes a value from a Map.
+func (m *Map) LookupAndDelete(key, valueOut interface{}) error {
+	valueBytes := make([]byte, m.fullValueSize)
```
lmb: You need to use makeBuffer here. The function currently doesn't change `valueOut`.

map_test.go
```diff
@@ -144,6 +144,62 @@ func createArray(t *testing.T) *Map {
...
+
+	var v uint32
+	if err := m.Lookup(nil, &v); err != nil {
+		t.Fatal("Can't lookup element:", err)
+	}
+	if v != 42 {
+		t.Error("Want value 42, got", v)
+	}
+
+	if err := m.LookupAndDelete(nil, &v); err != nil {
```
lmb:
```suggestion
	v = 0
	if err := m.LookupAndDelete(nil, &v); err != nil {
```

florianl:
Thanks for your review!
i've squashed the commits :)

lmb:
Thanks!

PR #44, opened 2020-01-21
Add ExistingMaps to CollectionOptions to provide possibility of map reuse when creating a collection, and slight change to PerfEventArray keysize valuesize validation

hamburgerbun:
This adds an optional field, ExistingMaps, to the CollectionOptions struct such that if a user wants to reuse an existing BPF map when creating a new collection, the user can add Map objects to ExistingMaps, and if there is a MapSpec at collection creation time with the same key as a Map in ExistingMaps, the existing map will be used instead of creating a new map. 

The use case that brought this on was for loading eBPF programs that each shared 4 "global" maps and 2 "local" maps per each interface involved. This change allows for using LoadPinnedMap to get the global maps and use them in a collection so that it only creates the local maps and reuses the global maps. 

The 2nd fix is simply allowing perf event array to not be set to 0 for keysize and valuesize since it is being hardcoded to 4 anyways; if the fields are already set to 4, we just skip the validation block. The issue is while using 0/0 works here, usage or 0/0 in other places like iproute2 and tc start throwing errors.

### DIFF
--- a/map.go
+++ b/map.go
@@ -20,7 +20,6 @@ type MapSpec struct {
 	ValueSize  uint32
 	MaxEntries uint32
 	Flags      uint32
-	Pinning    uint32
 
 	// InnerMap is used as a template for ArrayOfMaps and HashOfMaps
 	InnerMap *MapSpec
--- a/map.go
+++ b/map.go
@@ -20,6 +20,7 @@ type MapSpec struct {
 	ValueSize  uint32
 	MaxEntries uint32
 	Flags      uint32
+	Pinning    uint32
 
 	// InnerMap is used as a template for ArrayOfMaps and HashOfMaps
 	InnerMap *MapSpec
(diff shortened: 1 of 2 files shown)


lmb:
Hi, and thanks for contributing!

Regarding ExistingMaps, we currently have the following in our code base:
```Go
// rewriteMaps calls rewrites all programs in the spec to use the map
func rewriteMaps(spec *ebpf.CollectionSpec, maps map[string]*ebpf.Map) error {
	for symbol, m := range maps {
		// have we seen a program that uses this symbol / map
		seen := false
		fd := m.FD()
		for progName, progSpec := range spec.Programs {
			err := progSpec.Instructions.RewriteMapPtr(symbol, fd)

			switch {
			case err == nil:
				seen = true

			case asm.IsUnreferencedSymbol(err):
				// Not all programs need to use the map

			default:
				return errors.Wrapf(err, "program %s:", progName)
			}
		}

		if !seen {
			return errors.Errorf("map %s not referenced by any programs", symbol)
		}

		// Prevent NewCollection from trying to create bogus maps
		delete(spec.Maps, symbol)
	}

	return nil
}
```

Would this work for your use case as well?

For the PerfEventArray, you want to create arrays where KeySize, ValueSize is 4, but MaxEntries is 0?

map.go
```diff
@@ -129,22 +130,25 @@ func createMap(spec *MapSpec, inner *internal.FD, handle *btf.Handle) (*Map, err
...
-		if spec.ValueSize != 0 {
-			return nil, errors.Errorf("ValueSize must be zero for perf event array")
-		}
-		if spec.MaxEntries == 0 {
-			n, err := internal.OnlineCPUs()
-			if err != nil {
-				return nil, errors.Wrap(err, "perf event array")
+		if spec.KeySize != 4 || spec.ValueSize != 4 {
+			// do not validate them if they're already correct
+			if spec.KeySize != 0 {
```
lmb: I think doing the following might make the code a bit simpler:
```suggestion
			if spec.KeySize != 0 || spec.KeySize != 4 {
```

hamburgerbun: Hey, thanks for taking a look.

yeah i think rewriteMaps would do the trick for me without needing to add this hacky spec field if it's exposed for me to call. 

w.r.t. that keysize valuesize validation:
i need to look at the changes i made again but what it boiled down to was one of my fellow devs would do some of his quick-and-dirty testing via tc and iproute2 calls to set tc-bpf actions and xdp programs, but a reference to PerfEventArray with keysize and valuesize set to 0 makes whatever those two do for bpf loading throw errors concerning the map. i don't think it's a huge deal, it was more a matter of convenience for him since primarily we want to be loading using this stuff anyways.

anyways. if i'm able to call rewriteMaps, then this PR can just be turned down, the 0 or 4 thing was just a strange curiosity to me since it's gonna be set to 4 anyways.

lmb:
I've filed two PRs based on your feedback, please take a look. Thank you for raising this issue :)

PR #47, opened 2020-01-27
collection: allow rewriting individual maps

lmb:
Given a CollectionSpec, it's sometimes useful to use existing Maps instead
of having NewCollection spec create them from scratch. For example,
multiple programs might share the same map. Another use case is persistet
state that needs to be kept between two versions of an eBPF program.

Add a helper RewriteMaps which allows to achieve this.


lmb:
cc @hamburgerbun

PR #48, opened 2020-01-27
map: allow creating PerfEventArray with key or value size four

lmb:
libbpf based tooling expects map definitions for PerfEventArrays to
have a key and value size of four, while this package has traditionally
required a size of zero. Accept key or value size of four to be
compatible with libbpf.


lmb:
cc @hamburgerbun

PR #46, opened 2020-01-25
migrate error handling from github.com/pkg/errors to golang.org/x/xerrors

florianl:
Signed-off-by: Lehner Florian <dev@der-flo.net>

As mentioned in #38, this PR migrates error handling from github.com/pkg/errors to golang.org/x/xerrors

PR #49, opened 2020-01-27
migrate error handling

florianl:

As mentioned in https://github.com/cilium/ebpf/pull/46, this PR resolved the merge conflict and applied the suggestions from @lmb.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -12,7 +12,7 @@ import (
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/btf"
 
-	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 )
 
 type elfCode struct {
@@ -32,7 +32,10 @@ func LoadCollectionSpec(file string) (*CollectionSpec, error) {
 	defer f.Close()
 
 	spec, err := LoadCollectionSpecFromReader(f)
-	return spec, errors.Wrapf(err, "file %s", file)
+	if err != nil {
+		return &CollectionSpec{}, xerrors.Errorf("file %s: %w", file, err)
+	}
+	return spec, nil
 }
 
 // LoadCollectionSpecFromReader parses an ELF file into a CollectionSpec.
@@ -45,7 +48,7 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 
 	symbols, err := f.Symbols()
 	if err != nil {
-		return nil, errors.Wrap(err, "load symbols")
+		return nil, xerrors.Errorf("load symbols: %v", err)
 	}
 
 	ec := &elfCode{f, symbols, symbolsPerSection(symbols), "", 0}
@@ -71,13 +74,13 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 			btfMaps[elf.SectionIndex(i)] = sec
 		case sec.Type == elf.SHT_REL:
 			if int(sec.Info) >= len(ec.Sections) {
-				return nil, errors.Errorf("found relocation section %v for missing section %v", i, sec.Info)
+				return nil, xerrors.Errorf("found relocation section %v for missing section %v", i, sec.Info)
 			}
 
 			// Store relocations under the section index of the target
 			idx := elf.SectionIndex(sec.Info)
 			if relSections[idx] != nil {
-				return nil, errors.Errorf("section %d has multiple relocation sections", sec.Info)
+				return nil, xerrors.Errorf("section %d has multiple relocation sections", sec.Info)
 			}
 			relSections[idx] = sec
 		case sec.Type == elf.SHT_PROGBITS && (sec.Flags&elf.SHF_EXECINSTR) != 0 && sec.Size > 0:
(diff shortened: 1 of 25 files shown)


elf_reader.go
```diff
@@ -32,7 +32,10 @@ func LoadCollectionSpec(file string) (*CollectionSpec, error) {
 	defer f.Close()
 
 	spec, err := LoadCollectionSpecFromReader(f)
-	return spec, errors.Wrapf(err, "file %s", file)
+	if err != nil {
+		return &CollectionSpec{}, xerrors.Errorf("file %s: %w", file, err)
```
lmb:
```suggestion
		return nil, xerrors.Errorf("file %s: %w", file, err)
```

elf_reader.go
```diff
@@ -137,8 +140,10 @@ func loadVersion(sec *elf.Section, bo binary.ByteOrder) (uint32, error) {
 	}
 
 	var version uint32
-	err := binary.Read(sec.Open(), bo, &version)
-	return version, errors.Wrapf(err, "section %s", sec.Name)
+	if err := binary.Read(sec.Open(), bo, &version); err != nil {
+		return version, xerrors.Errorf("section %s: %v", sec.Name, err)
```
lmb:
```suggestion
		return 0, xerrors.Errorf("section %s: %v", sec.Name, err)
```

internal/btf/btf.go
```diff
@@ -454,8 +454,8 @@ func ProgramLineInfos(s *Program) (recordSize uint32, bytes []byte, err error) {
 // IsNotSupported returns true if the error indicates that the kernel
 // doesn't support BTF.
 func IsNotSupported(err error) bool {
-	ufe, ok := errors.Cause(err).(*internal.UnsupportedFeatureError)
-	return ok && ufe.Name == "BTF"
+	var ufe *internal.UnsupportedFeatureError
+	return xerrors.As(err, &ufe) && (err).(*internal.UnsupportedFeatureError).Name == "BTF"
```
lmb: If `As` returns true, you can use `ufe` like so:
```suggestion
	return xerrors.As(err, &ufe) && ufe.Name == "BTF"
```

internal/testutils/feature.go
```diff
@@ -48,13 +48,13 @@ func CheckFeatureTest(t *testing.T, fn func() error) {
 }
 
 func SkipIfNotSupported(tb testing.TB, err error) {
-	ufe, ok := errors.Cause(err).(*internal.UnsupportedFeatureError)
-	if !ok {
-		return
+	for err != nil {
```
lmb: I think you've re-implemented `errors.As` here, is there a particular reason for that?

My suggestion would be:
```
	var ufe *internal.UnsupportedFeatureError
	if xerrors.As(err, &ufe) {
		checkKernelVersion(tb, ufe)
		tb.Skip(ufe.Error())
	}
```

perf/reader.go
```diff
@@ -423,6 +429,5 @@ func (uev *unknownEventError) Error() string {
 // IsUnknownEvent returns true if the error occured
 // because an unknown event was submitted to the perf event ring.
 func IsUnknownEvent(err error) bool {
-	_, ok := errors.Cause(err).(*unknownEventError)
-	return ok
+	return xerrors.Is(err, (err).(*unknownEventError))
```
lmb: I think I mentioned this before, this will cause a panic if err is not an unknownEventError. Have a look at https://golang.org/ref/spec#Type_assertions

```suggestion
	var uee *unknownEventError
	return xerrors.As(err, &uee)
```

prog.go
```diff
@@ -140,8 +140,10 @@ func newProgramWithBTF(spec *ProgramSpec, btf *btf.Handle, opts ProgramOptions)
 		_, logErr := bpfProgLoad(attr)
 		err = internal.ErrorWithLog(err, logBuf, logErr)
 	}
-
-	return nil, errors.Wrap(err, "can't load program")
+	if err != nil {
+		return nil, xerrors.Errorf("can't load program: %w", err)
```
lmb: We already know that err != nil, so this if isn't needed. Just `return nil, xerrors.Errorf("can't load program: %w", err)`

prog.go
```diff
@@ -515,6 +526,5 @@ func SanitizeName(name string, replacement rune) string {
 // IsNotSupported returns true if an error occurred because
 // the kernel does not have support for a specific feature.
 func IsNotSupported(err error) bool {
-	_, notSupported := errors.Cause(err).(*internal.UnsupportedFeatureError)
-	return notSupported
+	return xerrors.Is(err, (err).(*internal.UnsupportedFeatureError))
```
lmb: Same as before.

syscalls.go
```diff
@@ -329,19 +332,27 @@ func bpfGetObjectInfoByFD(fd *internal.FD, info unsafe.Pointer, size uintptr) er
...
+	}
+	return nil
 }
 
 func bpfGetProgInfoByFD(fd *internal.FD) (*bpfProgInfo, error) {
 	var info bpfProgInfo
-	err := bpfGetObjectInfoByFD(fd, unsafe.Pointer(&info), unsafe.Sizeof(info))
-	return &info, errors.Wrap(err, "can't get program info")
+	if err := bpfGetObjectInfoByFD(fd, unsafe.Pointer(&info), unsafe.Sizeof(info)); err != nil {
+		return &info, xerrors.Errorf("can't get program info: %w", err)
```
lmb:
```suggestion
		return nil, xerrors.Errorf("can't get program info: %w", err)
```

syscalls.go
```diff
@@ -329,19 +332,27 @@ func bpfGetObjectInfoByFD(fd *internal.FD, info unsafe.Pointer, size uintptr) er
...
+	}
+	return &info, nil
 }
 
 func bpfGetMapInfoByFD(fd *internal.FD) (*bpfMapInfo, error) {
 	var info bpfMapInfo
 	err := bpfGetObjectInfoByFD(fd, unsafe.Pointer(&info), unsafe.Sizeof(info))
-	return &info, errors.Wrap(err, "can't get map info")
+	if err != nil {
+		return &bpfMapInfo{}, xerrors.Errorf("can't get map info: %w", err)
```
lmb:
```suggestion
		return nil, xerrors.Errorf("can't get map info: %w", err)
```

prog.go
florianl: will update this.

internal/testutils/feature.go
florianl: will update this according to your comment.

prog.go
florianl: `errors.Wrap()` returns [1] nil, if `err == nil`. Otherwise it returns the annotated error. So, this can not be simply replaced by `xerrors.Errorf()`.

[1] https://godoc.org/github.com/pkg/errors#Wrap

lmb: There is a check for `if err == nil` further up, so if we end up at this point `err != nil` always.

florianl: correct :see_no_evil: this not see this before. will adapt it.

lmb:
Thanks for making the fixes :) I've merged your commits (and fixed merge conflicts by doing the following):

```bash
# transplant your 4 commits on top of current master
git rebase --onto master HEAD~4 florianl-xerrors2
# mark your last 3 commits as fixup to merge them into a single one
git rebase -i HEAD~4
```

florianl:
thanks for taking care and your review!

PR #52, opened 2020-01-28
internal: zero terminate pointers to a string

### LINKED ISSUE #51: when pinning map, possible to fail pinning or have extra chars on end of filename because of non-null terminated string
this happened more commonly with maps with longer filenames for some reason, but for example, if i were to pin something like
/sys/fs/bpf/aaaaa/cachedConnectionTbl
there was a low chance that the pin would fail due to "No such file or directory" randomly, or a somewhat higher chance the file would be created but it'd end up something like
/sys/fs/bpf/aaaaa/cachedConnectionTbl[&%&%&% 
like there would be random characters attached to the end. I'm guessing it's because in bpfPinObject we directly pass in the pointer to the first character of the filename to pin the bpf object at without checking if it's null terminated, so there's a chance of stuff being slapped onto the end. 

In my case, passing in a string to <Map>.Pin(fileName) with null terminator purposely added to the end seems to fix the issue.

lmb:
The kernel expects any strings to be zero terminated. The pointer conversion
code so far hasn't ensured that this is the case. Unfortunately it's not
possible to test for this reliably, since the behaviour seems to depend on
how the Go runtime copies strings into byte buffers.

Fixes #51

### DIFF
--- a/internal/ptr.go
+++ b/internal/ptr.go
@@ -22,5 +22,9 @@ func NewStringPointer(str string) Pointer {
 		return Pointer{}
 	}
 
-	return Pointer{ptr: unsafe.Pointer(&[]byte(str)[0])}
+	// The kernel expects strings to be zero terminated
+	buf := make([]byte, len(str)+1)
+	copy(buf, str)
+
+	return Pointer{ptr: unsafe.Pointer(&buf[0])}
 }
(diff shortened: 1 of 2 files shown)


lmb:
Turns out that this problem only exists for paths that are exactly a power of two long. Thanks for reporting this fun bug!

PR #57, opened 2020-02-05
use sentinel errors instead of functions

lmb:
With the new errors.Is and errors.As functions from Go 1.13 we can remove
the various Is* functions without a loss in flexibility. Instead, we export
sentinel errors and take care to always return them in a wrapped form.

Users that are not yet on Go 1.13 can use the golang.org/x/xerrors package.

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -136,7 +136,7 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (col
 		var handle *btf.Handle
 		if mapSpec.BTF != nil {
 			handle, err = loadBTF(btf.MapSpec(mapSpec.BTF))
-			if err != nil && !btf.IsNotSupported(err) {
+			if err != nil && !xerrors.Is(err, btf.ErrNotSupported) {
 				return nil, err
 			}
 		}
@@ -176,7 +176,7 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (col
 		var handle *btf.Handle
 		if progSpec.BTF != nil {
 			handle, err = loadBTF(btf.ProgramSpec(progSpec.BTF))
-			if err != nil && !btf.IsNotSupported(err) {
+			if err != nil && !xerrors.Is(err, btf.ErrNotSupported) {
 				return nil, err
 			}
 		}
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -18,6 +18,11 @@ import (
 
 const btfMagic = 0xeB9F
 
+// Errors returned by BTF functions.
+var (
+	ErrNotSupported = internal.ErrNotSupported
+)
+
 // Spec represents decoded BTF.
 type Spec struct {
 	rawTypes  []rawType
@@ -307,8 +312,7 @@ type Handle struct {
 
 // NewHandle loads BTF into the kernel.
 //
-// Returns an error if BTF is not supported, which can
-// be checked by IsNotSupported.
+// Returns ErrNotSupported if BTF is not supported.
 func NewHandle(spec *Spec) (*Handle, error) {
 	if err := haveBTF(); err != nil {
 		return nil, err
@@ -449,13 +453,6 @@ func ProgramLineInfos(s *Program) (recordSize uint32, bytes []byte, err error) {
 	}
 
 	return s.lineInfos.recordSize, bytes, nil
-}
-
-// IsNotSupported returns true if the error indicates that the kernel
-// doesn't support BTF.
-func IsNotSupported(err error) bool {
-	var ufe *internal.UnsupportedFeatureError
-	return xerrors.As(err, &ufe) && ufe.Name == "BTF"
 }
 
 type bpfLoadBTFAttr struct {
--- a/map.go
+++ b/map.go
@@ -5,9 +5,14 @@ import (
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/btf"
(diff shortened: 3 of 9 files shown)


internal/btf/btf.go
```diff
@@ -307,8 +312,7 @@ type Handle struct {
 
 // NewHandle loads BTF into the kernel.
 //
-// Returns an error if BTF is not supported, which can
-// be checked by IsNotSupported.
+// Returns ErrNotSupported.
```
tklauser: This sounds like `NewHandle` returns `ErrNotSupported` unconditionally. Maybe change this to `Returns ErrNotSupported if BTF is not supported.`

syscalls.go
```diff
@@ -277,7 +277,19 @@ func bpfMapGetNextKey(m *internal.FD, key, nextKeyOut internal.Pointer) error {
...
+func wrapMapError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if err == unix.ENOENT {
+		return ErrKeyNotExist
+	}
+
+	return xerrors.New(err.Error())
```
florianl: can we directly return `err` instead of wrapping its string in a new error?

internal/btf/btf.go
```diff
@@ -451,13 +455,6 @@ func ProgramLineInfos(s *Program) (recordSize uint32, bytes []byte, err error) {
 	return s.lineInfos.recordSize, bytes, nil
 }
 
-// IsNotSupported returns true if the error indicates that the kernel
-// doesn't support BTF.
-func IsNotSupported(err error) bool {
-	var ufe *internal.UnsupportedFeatureError
-	return xerrors.As(err, &ufe) && ufe.Name == "BTF"
```
florianl: i'm not quite sure, why `ufe.Name == "BTF"` was introduced. Please correct me, but I think, it should be fine without this second check.

syscalls.go
```diff
@@ -277,7 +277,19 @@ func bpfMapGetNextKey(m *internal.FD, key, nextKeyOut internal.Pointer) error {
...
-	return err
+	return wrapMapError(err)
+}
+
+func wrapMapError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if err == unix.ENOENT {
```
florianl: maybe, we can use `xerrors.Is(err, unix.ENOENT)` here?

collection.go
```diff
@@ -136,7 +136,7 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (col
 		var handle *btf.Handle
 		if mapSpec.BTF != nil {
 			handle, err = loadBTF(btf.MapSpec(mapSpec.BTF))
-			if err != nil && !btf.IsNotSupported(err) {
+			if err != nil && !xerrors.Is(err, btf.ErrNotSupported) {
```
florianl: as `ErrNotSupported` is defined in `prog.go`, we can use it directly.

collection.go
```diff
@@ -176,7 +176,7 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (col
 		var handle *btf.Handle
 		if progSpec.BTF != nil {
 			handle, err = loadBTF(btf.ProgramSpec(progSpec.BTF))
-			if err != nil && !btf.IsNotSupported(err) {
+			if err != nil && !xerrors.Is(err, btf.ErrNotSupported) {
```
florianl: as `ErrNotSupported` is defined in `prog.go`, we can use it directly.

map.go
```diff
@@ -621,26 +634,15 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 		return mi.err == nil
 	}
 
-	mi.err = errIterationAborted
+	mi.err = xerrors.Errorf("%w", ErrIterationAborted)
```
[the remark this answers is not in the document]
lmb: Good question, I'm following the last example from https://blog.golang.org/go1.13-errors#TOC_5.

Without the wrapping, users might do `err == ErrIterationAborted` instead of using `errors.Is`.

syscalls.go
lmb: See previous comment

internal/btf/btf.go
lmb: I was trying to be extra cautious, it doesn't make a difference for the code at the moment.

syscalls.go
lmb: Good point!

collection.go
lmb: True, but I wanted to keep them separate if we need to change the semantics later. My idea was that btf.ErrNotSupported at some point may be != ErrNotSupported. Maybe it's overkill, not sure.

florianl: sounds good to me.

syscalls.go
```diff
@@ -277,7 +277,19 @@ func bpfMapGetNextKey(m *internal.FD, key, nextKeyOut internal.Pointer) error {
 		value: nextKeyOut,
 	}
 	_, err = internal.BPF(_MapGetNextKey, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
-	return err
+	return wrapMapError(err)
+}
+
+func wrapMapError(err error) error {
```
florianl: what do you think about extending this call by adding a message and returning the error with this additional message?
e.g.:
```
	_, err = internal.BPF(_MapGetNextKey, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
	return wrapMapError(err, "bpfMapGetNextKey failed")
```

PR #58, opened 2020-02-05
internal: consider hotplugged CPUs in PossibleCPUs and OnlineCPUs

tklauser:
With CPU hotplugging present, /sys/devices/system/cpu/online
may contain not only ranges such as "0-3" but a mixture of CPU ranges
and individual CPUs, e.g. "0,3-7" (meaning CPU 0 and 3-7). Consider this
fact in the implementation of parseCPUs and also don't cache the results
in OnlineCPUs, as CPUs might appear and disappear during runtime.

See Documentation/admin-guide/cputopology.rst in the Linux kernel source
tree for more information.

Implementation based on github.com/tklauser/numcpus

### DIFF
--- a/internal/cpu.go
+++ b/internal/cpu.go
@@ -1,8 +1,9 @@
 package internal
 
 import (
-	"fmt"
-	"os"
+	"io/ioutil"
+	"strconv"
+	"strings"
 	"sync"
 
 	"golang.org/x/xerrors"
@@ -14,8 +15,7 @@ var sysCPU struct {
 	num  int
 }
 
-// PossibleCPUs returns the max number of CPUs a system may possibly have
-// Logical CPU numbers must be of the form 0-n
+// PossibleCPUs returns the max number of CPUs a system may possibly have.
 func PossibleCPUs() (int, error) {
 	sysCPU.once.Do(func() {
 		sysCPU.num, sysCPU.err = parseCPUs("/sys/devices/system/cpu/possible")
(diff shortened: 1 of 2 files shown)


internal/cpu.go
```diff
@@ -24,41 +24,39 @@ func PossibleCPUs() (int, error) {
...
-	defer file.Close()
 
-	var low, high int
-	n, _ := fmt.Fscanf(file, "%d-%d", &low, &high)
-	if n < 1 || low != 0 {
-		return 0, xerrors.Errorf("%s has unknown format: %v", path, err)
+	cpus := strings.Trim(string(buf), "\n ")
+	n := int(0)
+	for _, cpuRange := range strings.Split(cpus, ",") {
+		if len(cpuRange) == 0 {
```
lmb: Is an empty group legitimate? If not we should bail out here.

tklauser:
Thanks for your review.

> Thanks for the PR! When we wrote the code to parse CPUs we chose to ignore hotplugging, since it adds complications:
> 
>     * OnlineCPUs: for PerfEventArrays use the highest online CPU to control MaxElements. However, perf ring buffers can't be created for offline CPUs, so either the code in `perf` needs to know about which CPUs are online, or learn how to ignore errors due to trying to initialize a perf ring for an offline CPU. There is also a race condition here: what if online CPUs change between calls to OnlineCPUs?
> 
>     * PossibleCPUs: is it even possible to have ranges here? If yes there are more problems with per-CPU values.

No, I don't think it is. It's always `0-n` afaict. I changed the commit back to only get PossibleCPUs once (as before) but forgot to adjust the commit/PR message.
 
> We never hotplug CPUs, so we didn't bother dealing with these. If you have a use case for this, I'd suggest the following:

I currently don't have a use case for this, it just stood out to me while reading the code that this might be an issue. But I didn't consider the implications this has wrt. perf ring buffers. So I think I'll close this PR for now and resurrect if there ever comes the need to consider hotplugged CPUs.

PR #60, opened 2020-02-10
Add *GetNextID

florianl:

This PR implements the *GetNextID functions for ebpf maps and programs.

### DIFF
--- a/map.go
+++ b/map.go
@@ -17,6 +17,9 @@ var (
 	ErrIterationAborted = xerrors.New("iteration aborted")
 )
 
+// MapID represents the unique ID of an eBPF map
+type MapID uint32
+
 // MapSpec defines a Map.
 type MapSpec struct {
 	// Name is passed to the kernel as a debug aid. Must only contain
@@ -764,4 +767,12 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 // Returns ErrIterationAborted if it wasn't possible to do a full iteration.
 func (mi *MapIterator) Err() error {
 	return mi.err
+}
+
+// MapGetNextID returns the ID of the next eBPF map.
+//
+// Returns ErrNotExist, if there is no next eBPF map.
+func MapGetNextID(startID MapID) (MapID, error) {
+	id, err := objGetNextID(_MapGetNextID, uint32(startID))
+	return MapID(id), err
 }
--- a/syscalls.go
+++ b/syscalls.go
@@ -11,6 +11,11 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// Generic errors returned by BPF syscalls.
+var (
+	ErrNotExist = xerrors.New("requested object does not exit")
+)
+
 // bpfObjName is a null-terminated string made up of
 // 'A-Za-z0-9_' characters.
 type bpfObjName [unix.BPF_OBJ_NAME_LEN]byte
@@ -148,6 +153,12 @@ type bpfGetFDByIDAttr struct {
 
 type bpfMapFreezeAttr struct {
 	mapFd uint32
+}
+
+type bpfObjGetNextIDAttr struct {
+	startID   uint32
+	nextID    uint32
+	openFlags uint32
 }
 
 func bpfProgLoad(attr *bpfProgLoadAttr) (*internal.FD, error) {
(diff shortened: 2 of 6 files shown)


types.go
```diff
@@ -115,6 +115,7 @@ const (
 	_TaskFDQuery
 	_MapLookupAndDeleteElem
 	_MapFreeze
+	_BTFGetNextID
```
florianl: At the moment, the BTF part is no implemented. Calling `objGetNextID()` from `syscalls.go` within `internal/btf/*.go` would introduce an import cycle.

lmb:
Sorry for the long wait!

florianl:
thanks for the feedback! I will address it and update the PR.

syscalls.go
```diff
@@ -301,6 +307,17 @@ func bpfMapGetNextKey(m *internal.FD, key, nextKeyOut internal.Pointer) error {
...
 
+func objGetNextID(cmd int, start uint32, next *uint32) error {
+	attr := bpfObjGetNextID{
+		start: start,
+	}
+	_, err := internal.BPF(cmd, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
+	if err == nil {
+		*next = attr.next
+	}
+	return wrapMapError(err)
```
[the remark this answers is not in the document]
florianl: I've removed the wrapper and directly return the error.

syscalls.go
```diff
@@ -150,6 +150,12 @@ type bpfMapFreezeAttr struct {
 	mapFd uint32
 }
 
+type bpfObjGetNextID struct {
+	start uint32
+	next  uint32
```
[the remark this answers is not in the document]
florianl: I've renamed the variables accordingly.

florianl:
> How about adding two new types, and make the functions take and return them?

At the moment, I don't see the advantage, of introducing extra types likes `MapID` and `ProgramID`.

> What happens when there is no next ID? I think we should return a wrapped ErrNotExist or something.

In the case, where there is no next map or program id, the syscall returns `unix.ENOENT`, so `ErrNotExist` sounds good to me.

> Should Map and Program get ID()?

Sounds also, like a good idea. I will check, how to do it.

>  If I remember correctly there currently isn't a way to go from an ID to a program. Should that be added?

Yes, currently there is no way to get from an ID to a map or program. So in a next step, `BPF_PROG_GET_FD_BY_ID` and `BPF_MAP_GET_FD_BY_ID` need to be implemented so one can go further. Sounds this reasonable to you?

types.go
florianl: I didn't move the function `objGetNextID` into `internal/` as I wanted to avoid making it public.

florianl:
In the last commit, I've added the suggested error `ErrNotExist`.

Regarding the next steps and use of the ID:
The command `bpftool prog list` calls `show_prog` [1], which then triggers the syscall for `BPF_OBJ_GET_INFO_BY_FD`. This call path is the same for maps, programs and btf objects. From this returned attribute, one can then return the ID, if needed. This syscall is already implemented by `bpfGetObjectInfoByFD` (and is using `internal.FD` as argument) but not public available.

[1] https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/prog.c#L410-L428
[2] https://github.com/torvalds/linux/blob/master/tools/lib/bpf/bpf.c#L719-L734

map.go
```diff
@@ -13,6 +13,7 @@ import (
 
 // Errors returned by Map and MapIterator methods.
 var (
+	ErrNotExist         = xerrors.New("requested object does not exit")
```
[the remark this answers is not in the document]
florianl: Addressed this in the latest update.

map.go
```diff
@@ -765,3 +768,11 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 func (mi *MapIterator) Err() error {
 	return mi.err
 }
+
+// MapGetNextID returns the ID of the next eBPF map.
+//
+// // Returns ErrNotExist, if there is no next eBPF map.
```
lmb: Nit: double comment.

syscalls.go
```diff
@@ -301,6 +312,25 @@ func bpfMapGetNextKey(m *internal.FD, key, nextKeyOut internal.Pointer) error {
...
+	_, err := internal.BPF(cmd, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
+	return attr.nextID, wrapObjError(err)
+}
+
+func wrapObjError(err error) error {
+	if err == nil {
+		return nil
+	}
+	if xerrors.Is(err, unix.ENOENT) {
+		return ErrNotExist
```
lmb: This needs to be wrapped.

```suggestion
		return xerrors.Errorf("%w", ErrNotExist)
```

map.go
florianl: I will fix this.

syscalls.go
florianl: Following `wrapMapError()`, where we also directly return `ErrKeyNotExist`, I did not wrap this error. Also, just like `ErrKeyNotExist` `ErrNotExist` becomes part of the API. Do you want to avoid this? If we wrap this error here, I also need to adapt the comments for `MapGetNextID` and `ProgramGetNextID`.

lmb: Yeah, I want to avoid unwrapped errors, to be able to add or remove detail later on. All (indirect) callers of wrapMapError end up returning a wrapped error, so there never should be a bare ErrKeyNotExist. The same should apply for ErrNotExist.

Like I mentioned, the wrap* functions are really confusing and probably need to go. But that can happen later.

PR #61, opened 2020-02-14
btf: wrap errors correctly

tklauser:
Two `xerrors.Errorf` calls with the `%w` verb are missing the error to wrap.
Add it.


lmb:
Thanks!

PR #63, opened 2020-02-17
Add New(Map|Program)FromID

florianl:

This PR add the functions to get the FD of a map or program by its ID.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -419,18 +419,7 @@ func unmarshalProgram(buf []byte) (*Program, error) {
 	// Looking up an entry in a nested map or prog array returns an id,
 	// not an fd.
 	id := internal.NativeEndian.Uint32(buf)
-	fd, err := bpfGetProgramFDByID(id)
-	if err != nil {
-		return nil, err
-	}
-
-	name, abi, err := newProgramABIFromFd(fd)
-	if err != nil {
-		_ = fd.Close()
-		return nil, err
-	}
-
-	return newProgram(fd, name, abi), nil
+	return NewProgramFromID(ProgramID(id))
 }
 
 // MarshalBinary implements BinaryMarshaler.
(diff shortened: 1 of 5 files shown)


florianl:
Thanks for your feedback and I've addressed them.

syscalls.go
```diff
@@ -469,26 +469,10 @@ var objNameAllowsDot = internal.FeatureTest("dot in object names", "5.2", func()
...
-	ptr, err := internal.BPF(_MapGetFDByID, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
-	if err != nil {
-		return nil, xerrors.Errorf("can't get fd for map id %d: %w", id, err)
-	}
-	return internal.NewFD(uint32(ptr)), nil
-}
-
-func bpfGetProgramFDByID(id uint32) (*internal.FD, error) {
-	// available from 4.13
+func objGetFDByID(cmd int, fd uint32) (*internal.FD, error) {
```
tklauser: Why is `id` renamed to `fd` here? I think this should remain as `id`.

prog.go
```diff
@@ -419,7 +419,7 @@ func unmarshalProgram(buf []byte) (*Program, error) {
 	// Looking up an entry in a nested map or prog array returns an id,
 	// not an fd.
 	id := internal.NativeEndian.Uint32(buf)
-	fd, err := bpfGetProgramFDByID(id)
+	fd, err := objGetFDByID(_ProgGetFDByID, id)
 	if err != nil {
```
lmb: You can replace the rest of this function with a call to NewProgramFromFD, same for unmarshalMap.

prog_test.go
```diff
@@ -360,6 +360,41 @@ func TestProgramGetNextID(t *testing.T) {
...
+	var next ProgramID
+
+	if next, err = ProgramGetNextID(ProgramID(0)); err != nil {
+		t.Fatal("Can't get next ID:", err)
+	}
+	if next == ProgramID(0) {
+		t.Fatal("Expected next ID other than 0")
+	}
+
+	if _, err = NewProgramFromID(next); err != nil {
```
lmb: How much work is it to add `Program.ID() (ProgramID, error)` and similar for `Map`? Then this test could be a lot simpler:
```
id, err := prog.ID()
NewProgramFromID(id)
```

prog_test.go
```diff
@@ -360,6 +360,41 @@ func TestProgramGetNextID(t *testing.T) {
...
+		t.Fatalf("Can't get FD for program ID %d: %v", uint32(next), err)
+	}
+
+	// As there can be multiple programs, we use max(uint32) as ProgramID to trigger an expected error.
+	_, err = NewProgramFromID(ProgramID(4294967295))
```
lmb: Why not `math.MaxUint32`?

syscalls.go
lmb: Agree with Tobias, also please keep the `bpf` prefix:

```suggestion
func bpfObjGetFDByID(cmd int, id uint32) (*internal.FD, error) {
```

prog_test.go
florianl: I didn't want to add another dependency. But you're right, it's easier with `math.MaxUint32`.

syscalls.go
florianl: renamed it

prog_test.go
florianl: I've added the implementation for `ID()` and updated the tests.

syscalls.go
florianl: sorry for the typo. fixed it.

prog.go
florianl: updated this functions as well.

PR #67, opened 2020-02-21
btf: add String() for btfKind

florianl:

This PR started by fixing a simple error message and resulted then in adding the missing `String()` for `btfKind`.

### DIFF
--- a/internal/btf/btf_types.go
+++ b/internal/btf/btf_types.go
@@ -2,6 +2,7 @@ package btf
 
 import (
 	"encoding/binary"
+	"fmt"
 	"io"
 
 	"golang.org/x/xerrors"
@@ -59,6 +60,45 @@ type btfType struct {
 	 * "type" is a type_id referring to another type.
 	 */
 	SizeType uint32
+}
+
+func (k btfKind) String() string {
+	switch k {
+	case kindUnknown:
+		return "Unknown"
+	case kindInt:
+		return "Integer"
+	case kindPointer:
+		return "Pointer"
+	case kindArray:
+		return "Array"
+	case kindStruct:
+		return "Struct"
+	case kindUnion:
+		return "Union"
+	case kindEnum:
+		return "Enumeration"
+	case kindForward:
+		return "Forward"
+	case kindTypedef:
+		return "Typedef"
+	case kindVolatile:
+		return "Volatile"
+	case kindConst:
+		return "Const"
+	case kindRestrict:
+		return "Restrict"
+	case kindFunc:
+		return "Function"
+	case kindFuncProto:
+		return "Function Proto"
+	case kindVar:
+		return "Variable"
+	case kindDatasec:
+		return "Section"
+	default:
+		return fmt.Sprintf("Unknown (%d)", k)
+	}
 }
 
 func mask(len uint32) uint32 {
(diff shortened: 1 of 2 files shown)


example_sock_elf_test.go
```diff
@@ -77,8 +77,6 @@ var program = [...]byte{
 	0010, 0000, 0000, 0000, 0000, 0000, 0000, 0000, 0030, 0000, 0000, 0000, 0000, 0000, 0000, 0000,
 }
 
-const sockexPin = "/sys/fs/bpf/sockex1"
```
tklauser: This change looks unrelated. Maybe move into a separate commit?

internal/btf/btf_types.go
```diff
@@ -61,6 +62,45 @@ type btfType struct {
...
+	case kindPointer:
+		return "Pointer"
+	case kindArray:
+		return "Array"
+	case kindStruct:
+		return "Struct"
+	case kindUnion:
+		return "Union"
+	case kindEnum:
+		return "Enum"
```
tklauser: I'd use `Enumeration` here in order to stay consistent with the comment in `linux/btf.h`.
```suggestion
		return "Enumeration"
```

internal/btf/btf_types.go
```diff
@@ -61,6 +62,45 @@ type btfType struct {
...
+	case kindTypedef:
+		return "Typedef"
+	case kindVolatile:
+		return "Volatile"
+	case kindConst:
+		return "Const"
+	case kindRestrict:
+		return "Restrict"
+	case kindFunc:
+		return "Func"
```
tklauser: Same here, better stay consistent with `linux/btf.h`
```suggestion
		return "Function"
```

internal/btf/btf_types.go
```diff
@@ -61,6 +62,45 @@ type btfType struct {
...
+	case kindFuncProto:
+		return "FuncProto"
```
tklauser: Same here:
```suggestion
		return "Function Proto"
```

internal/btf/btf_types.go
```diff
@@ -61,6 +62,45 @@ type btfType struct {
...
+	case kindVar:
+		return "Var"
```
tklauser:
```suggestion
		return "Variable"
```

internal/btf/btf_types.go
```diff
@@ -61,6 +62,45 @@ type btfType struct {
...
+	case kindDatasec:
+		return "Datasec"
```
tklauser: `linux/btf.h` calls this just `Section`
```suggestion
		return "Section"
```

example_sock_elf_test.go
florianl: I'v removed it from this PR.

florianl:
I've applied the requested changes.

internal/btf/btf_types.go
```diff
@@ -61,6 +62,45 @@ type btfType struct {
 	SizeType uint32
 }
 
+func (k btfKind) String() string {
+	switch k {
+	case kindUnknown:
+		return "Unknown"
+	case kindInt:
+		return "Int"
```
tklauser: Sorry, I missed one in my previous review
```suggestion
		return "Integer"
```

Issue #26, opened 2019-10-29
map, prog: allow access to bpfMapInfo, bpfProgInfo


whl739:
Currently after i load a map/prog from `LoadPinnedMap()` / `LoadPinnedProgram()`, i can use  [`Map.ABI()`](https://godoc.org/github.com/cilium/ebpf#Map.ABI) / [`Program.ABI()`](https://godoc.org/github.com/cilium/ebpf#ProgramABI) to get informations of the map/prog, `Map.ABI`'s informations are enough for me, but `Program.ABI ` has only `ProgramType`, why not just use `bpfMapInfo`/`bpfProgInfo` as an atrribute of `Map`/`Program`, and add an API to get it?

lmb:
Which parts of the structs do you want to access? Maybe it makes sense to add them to ProgramABI.

lmb:
This is now available in `v0.3.0`.

PR #69, opened 2020-02-25
Add ID, Name and OwnerProgType to MapABI

tklauser:
Updates #26

### DIFF
--- a/abi.go
+++ b/abi.go
@@ -15,20 +15,26 @@ import (
 
 // MapABI are the attributes of a Map which are available across all supported kernels.
 type MapABI struct {
-	Type       MapType
-	KeySize    uint32
-	ValueSize  uint32
-	MaxEntries uint32
-	Flags      uint32
+	Type          MapType
+	ID            MapID
+	KeySize       uint32
+	ValueSize     uint32
+	MaxEntries    uint32
+	Flags         uint32
+	Name          string
+	OwnerProgType ProgramType
 }
 
 func newMapABIFromSpec(spec *MapSpec) *MapABI {
 	return &MapABI{
 		spec.Type,
+		0,
 		spec.KeySize,
 		spec.ValueSize,
 		spec.MaxEntries,
 		spec.Flags,
+		spec.Name,
+		UnspecifiedProgram,
 	}
 }
 
@@ -44,25 +50,42 @@ func newMapABIFromFd(fd *internal.FD) (string, *MapABI, error) {
 
 	return "", &MapABI{
 		MapType(info.mapType),
+		MapID(info.id),
 		info.keySize,
 		info.valueSize,
 		info.maxEntries,
 		info.flags,
+		internal.CString(info.mapName[:]),
+		UnspecifiedProgram,
 	}, nil
 }
 
 func newMapABIFromProc(fd *internal.FD) (*MapABI, error) {
 	var abi MapABI
-	err := scanFdInfo(fd, map[string]interface{}{
+	fields := map[string]interface{}{
 		"map_type":    &abi.Type,
 		"key_size":    &abi.KeySize,
 		"value_size":  &abi.ValueSize,
 		"max_entries": &abi.MaxEntries,
 		"map_flags":   &abi.Flags,
-	})
+		"map_id":      &abi.ID,
+	}
+	if abi.Type == ProgramArray {
+		fields["owner_prog_type"] = &abi.OwnerProgType
+	}
+
+	err := scanFdInfo(fd, fields)
+	if xerrors.Is(err, errMissingFields) {
+		return nil, &internal.UnsupportedFeatureError{
+			Name:           "reading map ABI from /proc/self/fdinfo",
+			MinimumVersion: internal.Version{4, 5, 0},
+		}
+	}
+
 	if err != nil {
 		return nil, err
 	}
+
 	return &abi, nil
 }
 
@@ -71,6 +94,8 @@ func (abi *MapABI) Equal(other *MapABI) bool {
 	switch {
 	case abi.Type != other.Type:
 		return false
+	case abi.ID != other.ID:
+		return false
 	case abi.KeySize != other.KeySize:
 		return false
 	case abi.ValueSize != other.ValueSize:
@@ -78,6 +103,10 @@ func (abi *MapABI) Equal(other *MapABI) bool {
 	case abi.MaxEntries != other.MaxEntries:
 		return false
 	case abi.Flags != other.Flags:
+		return false
+	case abi.Name != other.Name:
+		return false
+	case abi.OwnerProgType != other.OwnerProgType:
 		return false
 	default:
 		return true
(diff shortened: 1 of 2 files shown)


abi.go
```diff
@@ -44,25 +50,42 @@ func newMapABIFromFd(fd *internal.FD) (string, *MapABI, error) {
...
+	}
+	if abi.Type == ProgramArray {
+		fields["owner_prog_type"] = &abi.OwnerProgType
+	}
+
+	err := scanFdInfo(fd, fields)
+	if xerrors.Is(err, errMissingFields) {
+		return nil, &internal.UnsupportedFeatureError{
+			Name:           "reading map ABI from /proc/self/fdinfo",
+			MinimumVersion: internal.Version{4, 5, 0},
```
lmb: Not all fields are available from 4.5, see the failing test on 4.9.

abi.go
```diff
@@ -44,25 +50,42 @@ func newMapABIFromFd(fd *internal.FD) (string, *MapABI, error) {
...
 		"map_type":    &abi.Type,
 		"key_size":    &abi.KeySize,
 		"value_size":  &abi.ValueSize,
 		"max_entries": &abi.MaxEntries,
 		"map_flags":   &abi.Flags,
-	})
+		"map_id":      &abi.ID,
+	}
+	if abi.Type == ProgramArray {
+		fields["owner_prog_type"] = &abi.OwnerProgType
```
lmb: So only ProgramArray will have this populated? Do other map types not honour this field?

abi.go
```diff
@@ -44,25 +50,42 @@ func newMapABIFromFd(fd *internal.FD) (string, *MapABI, error) {
 
 	return "", &MapABI{
 		MapType(info.mapType),
+		MapID(info.id),
 		info.keySize,
 		info.valueSize,
 		info.maxEntries,
 		info.flags,
+		internal.CString(info.mapName[:]),
+		UnspecifiedProgram,
```
lmb: I think this is problematic: newMapABIFromFd and newMapABIFromProc should produce the same result for a given fd. Otherwise the behaviour of MapABI will change subtly on different kernels. Maybe we should add that as a test.

abi.go
```diff
@@ -15,20 +15,26 @@ import (
 
 // MapABI are the attributes of a Map which are available across all supported kernels.
 type MapABI struct {
-	Type       MapType
-	KeySize    uint32
-	ValueSize  uint32
-	MaxEntries uint32
-	Flags      uint32
+	Type          MapType
+	ID            MapID
```
lmb: IMO name, ID aren't part of the ABI. They don't affect the behaviour of the map after all. Not sure about OwnerProgType, what is it used for?

abi.go
tklauser: Exactly, see here https://github.com/torvalds/linux/blob/63623fd44972d1ed2bfb6e0fb631dfcf547fd1e7/kernel/bpf/syscall.c#L542-L568

abi.go
tklauser: Is there a way to extract fields based on kernel version?

abi.go
tklauser: I don't think there currently is a way to get `OwnerProgType` from `newMapABIFromFd` unfortunately.

abi.go
tklauser: So where would you propose to expose them? Given the godoc comment of `MapABI` ("MapABI are the attributes of a Map which are available across all supported kernels.") I'd actually consider all of the fields part of the ABI, no? What would be your definition of the ABI?

tklauser:
> I think modifying ABI will be difficult, because we're bound by kernel 4.9. What do you want to use the fields for?

In the end I'd like to use this package in https://github.com/cilium/bpf-map to replace the use of the `github.com/cilium/cilium/pkg/bpf`. This is a first step towards that.

What would be your suggestion to make these fields from `bpfMapInfo` accessible other than exposing them via `MapABI`?

abi.go
tklauser: It seems the `map_id` field was only added in Kernel 4.18: torvalds/linux@4316b40914ecd

abi.go
lmb: Yeah, the documentation is terrible, sorry. I've been struggling to describe it better.

In my mind, the ABI tells you whether a map is compatible with your code. Say, if you expect an LPM map, you can verify that you have been passed one. Same for the key and value sizes, etc. For this reason, I want to keep MapABI the same across all the supported kernels (currently >= 4.9).

ID and Name don't change the behaviour, from the point of view what operations you can perform on the map. Hence I wouldn't consider them to be part of the ABI. At the same time this seems to confuse a lot of people interacting with the library for the first time, so who knows.

lmb:
> What would be your suggestion to make these fields from bpfMapInfo accessible other than exposing them via MapABI?

I think the complication with *ABI is that they make strong guarantees: they'll be consistent across kernels, etc. We need an interface that makes little consistency guarantees across kernel versions but exposes all available metadata. If MapABI is the intersection of all supported kernels, this other interface is the union of them. In practice that means some API that allows to check if a piece of metadata is present (ID for example) and what it's value is. In the best case you'll get types out of the interface as well, like MapID, etc. The absolutely simplest version of this would be something like:

```go
type MapFoobar struct {
	Type          *MapType
	ID            *MapID
	KeySize       *uint32
	ValueSize     *uint32
	MaxEntries    *uint32
	Flags         *uint32
	Name          *string
	OwnerProgType *ProgramType
}
```

Maybe that's enough, or even more useful than MapABI? What do you think?

abi.go
tklauser: I see, thanks for explaining. That makes sense to me but it would probably be good to document this somewhere?

tklauser:
My initial interpretation of `MapABI` was that it is exactly this union of kernel features, with the unsupported features just having their value as `nil` for kernels which don't provide that particular feature.

But given your explanation of `MapABI` above and the fact that in some cases, the `nil` value mighgt be a valid value for the the kernel to return also if it supports the feature, this `MapFoobar` approach with pointer members probably makes more sense. But then again, it might be a bit cumbersome to work with since each member needs to be checked for `!= nil` before access. Though I don't have a better proposal for it ATM. I'll think about this a bit more and will try fleshing out a proposal if I find some time. Thanks.

lmb:
Yeah, I think the union might make more sense really.

We could do the following as a compromise:
```Go
type MapFoobar struct {
	Type          MapType
	KeySize       uint32
	ValueSize     uint32
	MaxEntries    uint32
	Flags         uint32
	ID            *MapID
	Name          *string
	OwnerProgType *ProgramType
}
```

But then there is no way for us to migrate from e.g. `*ProgramType` to `ProgramType` without API breakage whenever we bump the minimum kernel version. We also can't add support for older kernels easily. Maybe that's ok though? Don't know.

lmb:
I'll close this for now, feel free to open a new PR any time though.

Issue #24, opened 2019-10-22
CPU profiling breaks BPF_PROG_TEST_RUN


lmb:
It seems like Go CPU profiling uses signals to periodically interrupt the process and sample the program counter at that point. This has the side effect of aborting in-progress calls to BPF_PROG_TEST_RUN with EINTR.

The simplest fix is to simply retry on EINTR. However this will skew benchmarks, since in fact we did more work.

lmb:
This is going to become more problematic with Go 1.14, since it uses signals for it's asynchronous preemption.

PR #70, opened 2020-02-27
program: retry BPF_PROG_TEST_RUN on EINTR

lmb:
Testing a BPF program can be interrupted if a signal arrives for a
userspace thread while being stuck in the BPF syscall.
BPF_PROG_TEST_RUN returns EINTR in that case, and we are required
to retry.

This has an unfortunate side-effect: benchmark results will be
skewed, because we might have to do more work than was originally
asked of us. The per-run calculation performed by the Go test
harness are therefore unreliable. Document this until we get a
better idea how to fix this.

Updates #24

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -302,7 +302,7 @@ func (p *Program) Close() error {
 //
 // This function requires at least Linux 4.12.
 func (p *Program) Test(in []byte) (uint32, []byte, error) {
-	ret, out, _, err := p.testRun(in, 1)
+	ret, out, _, err := p.testRun(in, 1, nil)
 	if err != nil {
 		return ret, nil, xerrors.Errorf("can't test program: %w", err)
 	}
@@ -312,12 +312,16 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 // Benchmark runs the Program with the given input for a number of times
 // and returns the time taken per iteration.
 //
-// The returned value is the return value of the last execution of
-// the program.
+// Returns the result of the last execution of the program and the time per
+// run or an error. reset is called whenever the benchmark syscall is
+// interrupted, and should be set to testing.B.ResetTimer or similar.
+//
+// Note: profiling a call to this function will skew it's results, see
+// https://github.com/cilium/ebpf/issues/24
 //
 // This function requires at least Linux 4.12.
-func (p *Program) Benchmark(in []byte, repeat int) (uint32, time.Duration, error) {
-	ret, _, total, err := p.testRun(in, repeat)
+func (p *Program) Benchmark(in []byte, repeat int, reset func()) (uint32, time.Duration, error) {
+	ret, _, total, err := p.testRun(in, repeat, reset)
 	if err != nil {
 		return ret, total, xerrors.Errorf("can't benchmark program: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


lmb:
> Another option would be to add an additional method, e.g. called BenchmarkWithB which takes a testing.(*B) on which it could then call ResetTimer itself on EINTR and retry.

That's kind of what I was going for in the beginning: `Benchmark(in []byte, repeat int, reset func())` (passing in testing.B is problematic, because pulling in testing creates command line flags). But then that changes the API, and I don't really know how much of a problem this is. So I figured I'd just fix the issue, and see if somebody complains.

tklauser:
> > Another option would be to add an additional method, e.g. called BenchmarkWithB which takes a testing.(*B) on which it could then call ResetTimer itself on EINTR and retry.
> 
> That's kind of what I was going for in the beginning: `Benchmark(in []byte, repeat int, reset func())` (passing in testing.B is problematic, because pulling in testing creates command line flags). But then that changes the API, and I don't really know how much of a problem this is. So I figured I'd just fix the issue, and see if somebody complains.

Right, we don't want to pass in `testing.B`.

Maybe just add a second `BenchmarkWithReset` function that takes a `reset func()` and mark `Benchmark` as deprecated suggesting to use `BenchmarkWithReset`?

prog_test.go
```diff
@@ -99,6 +103,74 @@ func TestProgramBenchmark(t *testing.T) {
...
 
+func TestProgramTestRunInterrupt(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.0", "EINTR from BPF_PROG_TEST_RUN")
+
+	prog := createSocketFilter(t)
+	defer prog.Close()
+
+	var (
+		tgid    = unix.Getpid()
+		tidChan = make(chan int, 1)
```
florianl: maybe, we should also close `tidChan`?

lmb: What would the benefit be? It will be garbage collected regardless, no?

lmb:
> Maybe just add a second BenchmarkWithReset function that takes a reset func() and mark Benchmark as deprecated suggesting to use BenchmarkWithReset?

Then we should just bite the bullet, and break the API. We've not made a stable release after all.

tklauser:
sgtm

lmb:
Updated the PR

prog.go
```diff
@@ -312,12 +312,20 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
...
+//
+// Note: profiling a call to this function will skew it's results, see
+// https://github.com/cilium/ebpf/issues/24
 //
 // This function requires at least Linux 4.12.
-func (p *Program) Benchmark(in []byte, repeat int) (uint32, time.Duration, error) {
-	ret, _, total, err := p.testRun(in, repeat)
+func (p *Program) Benchmark(in []byte, repeat int, reset func()) (uint32, time.Duration, error) {
+	if reset == nil {
+		return 0, 0, xerrors.New("reset may not be nil")
```
tklauser: May we could just set reset to an empty func here instead of erroring out?

```suggestion
		reset = func() {}
```

lmb: Fixed, testRun already accepts a nil reset.

PR #71, opened 2020-02-28
remove unused and duplicate code

florianl:

The last use of the constants was removed with commit aaa563ef6a5d4578ae252c89d230713a17b61dfe. And as there is `0xeB9F` already defined, it doesn't need to be defined twice.


lmb:
Thanks!

Issue #37, opened 2019-12-03
Document supported Go and Linux versions


lmb:
For Go, I believe we should only explicitly support the last two Go versions, in accordance with https://golang.org/doc/devel/release.html#policy This means that currently Go 1.12 and 1.13 is supported. It would also be nice to test against these on CI.

For Linux, Cilium's minimum requirement is 4.9. Cloudflare is always on either the ultimate or penultimate LTS release. I think we should only support mainline LTS kernels (so testing on CI).

PR #74, opened 2020-02-29
run tests on current supported go versions

florianl:

This PR runs the go tests on all upstream stable go versions. So far, there is no documentation, which describes the necessary go version for this repository - see issue https://github.com/cilium/ebpf/issues/37. Maybe, this PR can be an idea for a future discussion.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -24,10 +24,10 @@ blocks:
           - GOARCH=arm64 go build ./...
       - name: Test on 5.4.5
         commands:
-          - timeout -s KILL 90s ./run-tests.sh 5.4.5
+          - timeout -s KILL 180s ./run-tests.sh 5.4.5
       - name: Test on 4.19.81
         commands:
-          - timeout -s KILL 90s ./run-tests.sh 4.19.81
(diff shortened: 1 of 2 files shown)


lmb:
I think I'd prefer something along the lines of https://github.com/cilium/ebpf/commit/a935bf9d2b279e28c1218ce8a770261ef9151ec2

* run-test.sh is a bit simpler, since it doesn't have to download go binaries. my hunch is that we only need to test old versions on CI, not on developer laptops
* the sem-version thing is probably quicker than downloading the binaries manually
* using the semaphore matrix thingie means we get more parallelism and nice results in the UI: https://ebpf.semaphoreci.com/workflows/360d416e-a44c-46d8-840c-15f272aef2af

We should also add to the readme that the library follows the Go support model: https://golang.org/doc/devel/release.html#policy

Finally, this should mention #37 

What do you think?

florianl:
Reading https://github.com/cilium/ebpf/commit/a935bf9d2b279e28c1218ce8a770261ef9151ec2 is way simpler, than this PR. So I will close this PR and you will merge your changes from https://github.com/cilium/ebpf/tree/go-versions?

PR #77, opened 2020-03-07
fix errors

florianl:

This PR fixes two error messages.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -167,23 +167,23 @@ func (ec *elfCode) loadPrograms(progSections map[elf.SectionIndex]*elf.Section, 
 		libs  []*ProgramSpec
 	)
 
-	for idx, prog := range progSections {
+	for idx, sec := range progSections {
 		syms := ec.symbolsPerSection[idx]
 		if len(syms) == 0 {
-			return nil, xerrors.Errorf("section %v: missing symbols", prog.Name)
+			return nil, xerrors.Errorf("section %v: missing symbols", sec.Name)
 		}
 
 		funcSym, ok := syms[0]
 		if !ok {
-			return nil, xerrors.Errorf("section %v: no label at start", prog.Name)
+			return nil, xerrors.Errorf("section %v: no label at start", sec.Name)
 		}
 
-		insns, length, err := ec.loadInstructions(prog, syms, relocations[idx])
+		insns, length, err := ec.loadInstructions(sec, syms, relocations[idx])
 		if err != nil {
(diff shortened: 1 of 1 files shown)


elf_reader.go
```diff
@@ -197,7 +197,7 @@ func (ec *elfCode) loadPrograms(progSections map[elf.SectionIndex]*elf.Section,
 		if btf != nil {
 			spec.BTF, err = btf.Program(prog.Name, length)
 			if err != nil {
-				return nil, xerrors.Errorf("BTF for section %s (program %s): %w", prog.Name, funcSym, err)
```
lmb: Hm, not sure about this one. Do you have an example?

florianl: `funcSym` is of type `elf.Symbol` and this struct does not have a `String()` function. In addition, the `elf.Symbol` references to the BTF section and not `prog.Name`. So I switches these two arguments as well.

lmb: >In addition, the elf.Symbol references to the BTF section and not prog.Name.

Can you give me an example of the incorrect output? I find it difficult to wrap my head around this. Fair point re funcSym, however I think that prog.Name should be first, since it's the section we are trying to find BTF for.

florianl: a current example output looks like:

```
BTF for section global_fn2 (program .text): <nil>

BTF for section global_fn3 (program other): <nil>

BTF for section xdp_prog (program xdp): <nil>

BTF for section no_relocation (program socket): <nil>

BTF for section global_fn2 (program .text): <nil>

BTF for section global_fn3 (program other): <nil>

BTF for section xdp_prog (program xdp): <nil>

BTF for section no_relocation (program socket): <nil>
```
so, I will change the order as `.text` should be the section and not the program.

florianl: and here is the output after the most recent change:
```
BTF for section xdp (program xdp_prog): <nil>

BTF for section socket (program no_relocation): <nil>

BTF for section .text (program global_fn2): <nil>

BTF for section other (program global_fn3): <nil>

BTF for section .text (program global_fn2): <nil>

BTF for section other (program global_fn3): <nil>

BTF for section xdp (program xdp_prog): <nil>

BTF for section socket (program no_relocation): <nil>
```

lmb: Ah now I get it. It's confusing because `prog` in the code is the ELF section, and `spec` is actually prog. Maybe it makes sense to rename `prog` to `sec` or `section`, and then use `prog.Name` instead of `funcSym.Name`?

florianl: yes, the name confused me as well. I've applied your suggestion and named it `sec` instead of `prog`

PR #78, opened 2020-03-10
perf: fix breaking error for creating over perf ring from a loaded pr…

iAklis:
Keep same behavior with libbpf.

The ABI `max_entries` of  in **BPF_MAP_TYPE_PERF_EVENT_ARRAY** elf is more like a guidance than an exact value.

I'm glad someone reviewed the code for me.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -198,6 +198,15 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 		pauseFds = make([]int, 0, nCPU)
 	)
 
+	onlineCPU, err := internal.OnlineCPUs()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


lmb:
Thank your for the pull request! I'd prefer the approach I mentioned in: https://github.com/cilium/ebpf/pull/58#pullrequestreview-353738435

> Change the perf ring code in perf to try and create a ring for each slot in the PerfEventArray, and ignore errors due to offline CPUs.

That way the fix for your issue also partially fixes the issue with hotplugged CPUs.

According to `man 2 perf_event_open`, the syscall fails with EINVAL if the CPU doesn't exist. Problem: there are a bunch of other reasons why that is returned. Maybe we can ignore EINVAL errors, and then at the end check that we created at least a single event ring.

Are you up for that? If not, I'll try and get to this sometime in the next days.

PR #79, opened 2020-03-10
Don't ignore kernelVersion

florianl:

If specified in `ProgramSpec` so far `kenelVersion` was just ignored. This attribute is around since 4.1  [2541517c32be](https://github.com/torvalds/linux/commit/2541517c32be).

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -202,6 +202,7 @@ func convertProgramSpec(spec *ProgramSpec, handle *btf.Handle) (*bpfProgLoadAttr
 		insCount:           insCount,
 		instructions:       internal.NewSlicePointer(bytecode),
 		license:            internal.NewStringPointer(spec.License),
+		kernelVersion:      spec.KernelVersion,
 	}
 
 	if haveObjName() == nil {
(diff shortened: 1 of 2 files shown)


florianl:
True, this field was declared deprecated with commit [6c4fc209fcf9d27efbaa48368773e4d2bfbd59aa](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6c4fc209fcf9d27efbaa48368773e4d2bfbd59aa). As kernel version 4.9 and 4.19 are supported at the moment, this attribute is checked for BPF_PROG_TYPE_KPROBE.
I don't have a test in mind, but will think about it.

florianl:
I've added a test. Let me know, what you think about it.

florianl:
0 is already a bogus kernel version. The test fails on 4.9 and 4.19, if you comment out `	testutils.SkipOnOldKernel(...)`

florianl:
I've updated the test to use 42 instead of 0 as the kernel version.

lmb:
Sorry for the late reply, `0` would have been fine. Thanks again!

PR #80, opened 2020-03-14
Fix perf breaking

iAklis:
I found the error is not EINVAL but ENODEV, this could fix my code. Could you help me review? @lmb

Alternative solution #78

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -198,6 +198,15 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 		pauseFds = make([]int, 0, nCPU)
 	)
 
+	onlineCPU, err := internal.OnlineCPUs()
+	if err != nil {
+		return nil, xerrors.Errorf("perf event array: %w", err)
+	}
+
+	if nCPU > onlineCPU {
+		nCPU = onlineCPU
+	}
+
 	defer func() {
 		if err != nil {
 			for _, fd := range fds {
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"runtime"
 	"sync"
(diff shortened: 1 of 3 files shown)


perf/reader.go
```diff
@@ -215,6 +216,9 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 	for i := 0; i < nCPU; i++ {
 		ring, err := newPerfEventRing(i, perCPUBuffer, opts.Watermark)
 		if err != nil {
+			if xerrors.Is(err, syscall.ENODEV) {
```
lmb: The code expects you to append to `rings` for each CPU, and right now you skip this because of the continue. I need to look into how to handle that.

perf/reader_test.go
```diff
@@ -289,6 +290,18 @@ func TestCreatePerfEvent(t *testing.T) {
 		t.Fatal("Can't create perf event:", err)
 	}
 	unix.Close(fd)
+
+	onlineCPU, err := internal.OnlineCPUs()
+	if err != nil {
+		t.Fatal("perf event array: ", err)
+	}
+	fd, err = createPerfEvent(onlineCPU, 1)
+	if !xerrors.Is(err, syscall.ENODEV) {
```
lmb: If you check the Linux sources you'll find that ENODEV is only returned if:
- onlineCPU is [less than nr_cpu_ids](https://elixir.bootlin.com/linux/v5.5.9/source/kernel/events/core.c#L10685)
- and the [CPU is offline](https://elixir.bootlin.com/linux/v5.5.9/source/kernel/events/core.c#L11472)

I think you're violating the first condition, and getting EINVAL therefore.

lmb:
According to @iAklis he can use max_elems: 0, hence I'll close this.

PR #88, opened 2020-04-21
perf: Fix mmap buffer size calculation

arthurfabre:
The mmap buffer is sized in bytes, but needs to be page aligned.
This was correctly derived.

Additionally, with perf_event_open, according to man 2 perf_event_open:

> The mmap size should be 1+2^n pages

This wasn't correctly enforced, and caused the mmap call to fail with
EINVAL for some buffer sizes. For example 8193 bytes would be
incorrectly rounded to 3 pages instead of 4.
Always round up to the nearest power of two number of pages, and add
some tests.

Also allow `watermark == buffer` as this as a source of confusion, 
see https://github.com/cloudflare/xdpcap/issues/39#issuecomment-616382309

### DIFF
--- a/perf/ring.go
+++ b/perf/ring.go
@@ -2,6 +2,7 @@ package perf
 
 import (
 	"io"
+	"math"
 	"os"
 	"runtime"
 	"sync/atomic"
@@ -26,12 +27,6 @@ func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error) 
 		return nil, xerrors.New("watermark must be smaller than perCPUBuffer")
 	}
 
-	// Round to nearest page boundary and allocate
-	// an extra page for meta data
-	pageSize := os.Getpagesize()
-	nPages := (perCPUBuffer + pageSize - 1) / pageSize
-	size := (1 + nPages) * pageSize
-
 	fd, err := createPerfEvent(cpu, watermark)
 	if err != nil {
 		return nil, err
@@ -42,10 +37,10 @@ func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error) 
 		return nil, err
 	}
 
-	mmap, err := unix.Mmap(fd, 0, size, unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED)
+	mmap, err := unix.Mmap(fd, 0, perfBufferSize(perCPUBuffer), unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED)
 	if err != nil {
 		unix.Close(fd)
-		return nil, err
+		return nil, xerrors.Errorf("can't mmap: %v", err)
 	}
 
 	// This relies on the fact that we allocate an extra metadata page,
@@ -63,6 +58,22 @@ func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error) 
 	runtime.SetFinalizer(ring, (*perfEventRing).Close)
 
 	return ring, nil
+}
+
+// mmapBufferSize returns a valid mmap buffer size for use with perf_event_open (1+2^n pages)
+func perfBufferSize(perCPUBuffer int) int {
+	pageSize := os.Getpagesize()
+
+	// Smallest whole number of pages
+	nPages := (perCPUBuffer + pageSize - 1) / pageSize
+
+	// Round up to nearest power of two number of pages
+	nPages = int(math.Pow(2, math.Ceil(math.Log2(float64(nPages)))))
+
+	// Add one for metadata
+	nPages += 1
+
+	return nPages * pageSize
 }
 
 func (ring *perfEventRing) Close() {
--- a/perf/ring_test.go
+++ b/perf/ring_test.go
@@ -3,6 +3,7 @@ package perf
 import (
 	"bytes"
 	"io"
+	"os"
 	"testing"
 
 	"github.com/cilium/ebpf/internal/unix"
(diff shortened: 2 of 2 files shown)


perf/ring.go
```diff
@@ -65,6 +60,22 @@ func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error)
 	return ring, nil
 }
 
+// mmapBufferSize returns a valid mmap buffer size for use with perf_event_open (1+2^n pages)
+func mmapBufferSize(perCPUBuffer int) int {
```
lmb, line 64: Nit:

```suggestion
func perfBufferPages(perCPUBuffer int) int {
```

perf/ring_test.go
```diff
@@ -62,3 +64,44 @@ func makeRing(size, offset int) *ringReader {
...
+
+		size := len(ring.ringReader.ring)
+
+		// Ring size should be at least as big as buffer
+		if size < buffer {
+			t.Fatalf("ring size %d smaller than buffer %d", size, buffer)
+		}
+
+		// Ring size should be of the form 2^n pages (meta page has already been removed)
+		pages := float64(size) / float64(os.Getpagesize())
```
lmb, line 83: Why not `size % os.Getpagesize() != 0`? I find the math.Trunc a bit hard to follow.

perf/ring_test.go
```diff
@@ -62,3 +64,44 @@ func makeRing(size, offset int) *ringReader {
 
 	return newRingReader(&meta, ring)
 }
+
+func TestPerfEventRing(t *testing.T) {
```
lmb, line 68: I kind of expected you to test mmapBufferSize here, because you made it a separate function. That might save you some boilerplate?

arthurfabre: It does, but testing `mmapBufferSize` is hard because it depends on `os.Getpagesize()`, which varies. It also doesn't check that the sizes are actually accepted by the kernel, I figured testing the kernel is happy was more important. But I'm happy to change it if you like :)

perf/ring_test.go
arthurfabre: Nice, that is clearer! Fixed :)

perf/ring.go
arthurfabre: I changed it to `perfBufferSize()`, is that ok? I thought `perfBufferPages()` sounded like it returned a number of pages.

perf/ring.go
```diff
@@ -22,16 +23,10 @@ type perfEventRing struct {
 }
 
 func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error) {
-	if watermark >= perCPUBuffer {
+	if watermark > perCPUBuffer {
```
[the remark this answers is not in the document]
arthurfabre: Good point, I'll have a look.

arthurfabre:
> Thanks for taking this on :) While reading your test I had the following idea: what if we just reject buffer sizes that aren't a power of two already? That's a change in API, but we can probably audit the users of the perf reader for this. Downside: users have another hoop to jump through, upside: watermark and buffer size are now consistent.

Thanks for having a look :) Buffer sizes have to be a power of two number of pages, I don't get how requiring power of two buffer sizes helps? Or would users just specify a number of pages, and we would multiply it by the page size?

perf/ring.go
lmb: Ack

perf/ring_test.go
lmb: Good point. Maybe add `pageSize int` as an argument? Always testing the whole hog seems nice on the surface, but it means that making changes to newPerfEvent ripple out to the tests a lot.

lmb:
> Buffer sizes have to be a power of two number of pages, I don't get how requiring power of two buffer sizes helps? Or would users just specify a number of pages, and we would multiply it by the page size?

Ok, no that was a terrible idea of mine. Disregard!

perf/ring.go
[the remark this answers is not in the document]
arthurfabre: You're right, `watermark == buffer` will never trigger: https://elixir.bootlin.com/linux/latest/source/kernel/events/ring_buffer.c#L229

Removed :)

perf/ring_test.go
arthurfabre: I feel we should actually test the kernel accepts these combinations, and this seems like the best place to do it.
But if you'd rather just have a test for `perfBufferSize()`, I'll change it.

perf/ring.go
```diff
@@ -42,10 +37,10 @@ func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error)
 		return nil, err
 	}
 
-	mmap, err := unix.Mmap(fd, 0, size, unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED)
+	mmap, err := unix.Mmap(fd, 0, perfBufferSize(perCPUBuffer), unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED)
 	if err != nil {
 		unix.Close(fd)
-		return nil, err
+		return nil, xerrors.Errorf("can't mmap: %w", err)
```
lmb, line 43: Please don't wrap unless you want the caller to have access to the error. I think `%v` is the correct thing here.

lmb:
Can you fix the wrapping and rebase on master please?

perf/ring.go
arthurfabre: Good catch, fixed :)

arthurfabre:
Forced pushed again to re-run the builds, AFAICT https://ebpf.semaphoreci.com/jobs/5af77b40-756d-47b2-88a2-c1a11dbab61a seems to have just timed out. I diffed the output with a successful build on the same kernel (https://ebpf.semaphoreci.com/jobs/c59893bb-e772-4727-b8f7-899c2eb0de2c), and it doesn't seem like any tests were blocked or deadlocked, every `== RUN` had a matching `--- PASS`.

PR #90, opened 2020-04-28
The newtools ci-kernels project is running into LFS bandwidth

nathanjsweet:
limits, because it is not on a paid account. This retrieves the
kernels from cilium, rather than newtools, so the bandwidth
limit is not hit.


lmb:
404?

PR #94, opened 2020-05-11
Friendly map create errors, document rlimit and perf ring garbage

ti-mo:
Hi all,

While migrating a project off of gobpf, I encountered some issues loading an ELF binary, most of which have been addressed below. There's another part that deals with overwriting the binary's `version` section with the running kernel's `LINUX_VERSION_CODE`, but that I'd like to address in another PR.

The removal of BTF requirements in data sections and the rlimit changes might need some extra test coverage, but please let me know what you think before I get started on that.

I've added a bit more detailed info about the perf event buffer underrun in [the commit that introduced it](https://github.com/ti-mo/ebpf/commit/2cc7e4625e0baab30bfd8be9d5a5c4e6d8012dd2):
> https://elixir.bootlin.com/linux/v5.5.13/source/kernel/events/core.c#L6611 raw->size is used to size the buffer, but frag->size is used to fill it.

> https://elixir.bootlin.com/linux/v5.5.13/source/kernel/events/core.c#L6827
> While preparing the sample output, raw->size is set to the sum of all
> fragments' sizes minus 4 bytes. However, header->size doesn't include this
> 4-byte deduction.

I think I might've stumbled across a kernel bug, but looking at bcc and gobpf, it seems like many userspace libraries already depend on that behaviour by now and discard those bytes by default. Raw perf events are 64-bit aligned, but the test case wasn't.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -524,7 +524,7 @@ func uintFromBTF(typ btf.Type) (uint32, error) {
 
 func (ec *elfCode) loadDataSections(maps map[string]*MapSpec, dataSections map[elf.SectionIndex]*elf.Section, spec *btf.Spec) error {
 	if spec == nil {
-		return xerrors.New("data sections require BTF")
+		return xerrors.New("data sections require BTF, make sure all consts are marked as static")
 	}
 
 	for _, sec := range dataSections {
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -59,9 +59,8 @@ type Record struct {
 	CPU int
 
 	// The data submitted via bpf_perf_event_output.
-	// They are padded with 0 to have a 64-bit alignment.
-	// If you are using variable length samples you need to take
-	// this into account.
+	// Due to a kernel bug, this can contain between 0 and 7 bytes of trailing
+	// garbage from the ring depending on the input sample's length.
 	RawSample []byte
 
 	// The number of samples which could not be output, since
@@ -317,9 +316,11 @@ func (pr *Reader) Close() error {
 // Read the next record from the perf ring buffer.
 //
 // The function blocks until there are at least Watermark bytes in one
-// of the per CPU buffers.
+// of the per CPU buffers. Records from buffers below the Watermark
+// are not returned.
 //
-// Records from buffers below the Watermark are not returned.
+// Records can contain between 0 and 7 bytes of trailing garbage from the ring
+// depending on the input sample's length.
 //
 // Calling Close interrupts the function.
 func (pr *Reader) Read() (Record, error) {
--- a/perf/reader_test.go
+++ b/perf/reader_test.go
@@ -21,7 +21,7 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	err := unix.Setrlimit(8, &unix.Rlimit{
+	err := unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{
 		Cur: unix.RLIM_INFINITY,
 		Max: unix.RLIM_INFINITY,
 	})
--- a/syscalls.go
+++ b/syscalls.go
@@ -1,6 +1,7 @@
 package ebpf
 
 import (
+	"os"
 	"path/filepath"
 	"unsafe"
 
@@ -13,7 +14,7 @@ import (
 
 // Generic errors returned by BPF syscalls.
 var (
-	ErrNotExist = xerrors.New("requested object does not exit")
+	ErrNotExist = xerrors.New("requested object does not exist")
 )
 
 // bpfObjName is a null-terminated string made up of
@@ -185,6 +186,10 @@ func bpfProgAlter(cmd int, attr *bpfProgAlterAttr) error {
 
 func bpfMapCreate(attr *bpfMapCreateAttr) (*internal.FD, error) {
 	fd, err := internal.BPF(_MapCreate, unsafe.Pointer(attr), unsafe.Sizeof(*attr))
+	if xerrors.Is(err, os.ErrPermission) {
+		return nil, xerrors.New("permission denied or insufficient rlimit to lock memory for map")
+	}
+
 	if err != nil {
 		return nil, err
 	}
--- a/map.go
+++ b/map.go
@@ -107,6 +107,10 @@ func NewMapFromFD(fd int) (*Map, error) {
 //
 // Creating a map for the first time will perform feature detection
 // by creating small, temporary maps.
+//
+// The caller is responsible for ensuring the process' rlimit is set
+// sufficiently high for locking memory during map creation. This can be done
+// by calling unix.Setrlimit with unix.RLIMIT_MEMLOCK prior to calling NewMap.
 func NewMap(spec *MapSpec) (*Map, error) {
 	if spec.BTF == nil {
 		return newMapWithBTF(spec, nil)


elf_reader.go
```diff
@@ -554,8 +556,11 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec, dataSections map[e
 
 		switch sec.Name {
 		case ".rodata":
-			mapSpec.Flags = unix.BPF_F_RDONLY_PROG
-			mapSpec.Freeze = true
+			// Only freeze and mark read-only if the kernel supports it.
+			if haveMapMutabilityModifiers() == nil {
```
ti-mo, line 560: Putting both `Flags` and `Freeze` assignments behind `haveMapMutabilityModifiers()` causes the following test to fail:

```
TestLoadCollectionSpec/loader-clang-9-el.elf: elf_reader_test.go:97: program xdp_prog: can't load program: invalid argument: unrecognized bpf_ld_imm64 insn
```

`bpf_ld_imm64` should be used to load constants only, and the verifier in 4.19 seems to trip over loads from sections that are not frozen. Since the instruction was [introduced in 3.18](https://github.com/torvalds/linux/commit/02ab695bb37ee9ad515df0d0790d5977505dd04a), and `BPF_F_RDONLY_PROG` was introduced along with `BPF_MAP_FREEZE` in 5.2, what would be the correct behaviour here to support loading this program?

tklauser: /cc @lmb @joestringer @borkmann @pchaigno @qmonnet

qmonnet: Looks like the `LoadMapValue` generated by clang-9 are only supported since kernel 5.2 ([d8eca5bbb2be](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d8eca5bbb2be) ("bpf: implement lookup-free direct value access for maps")). The test is skipped (so doesn't fail) on 4.9. Not sure how the instruction appeared by moving the assignments behind the guard, though? Do you know if these `LoadMapValue` instructions appeared before in the programs produced for the tests on 4.19 before the change?

perf/reader.go
```diff
@@ -15,6 +15,15 @@ import (
 	"golang.org/x/xerrors"
 )
 
+const (
+	// `perf_raw_record` contains a trailing `u32 size`, whose length is
+	// included in `perf_event_header.size`, the size of the record's buffer.
+	// When reading raw perf samples from the ring, this many bytes need to be
+	// popped from the end of the data to avoid returning garbage once the ring
+	// has wrapped around.
+	perfRawRecordTailLength = 4
```
lmb, line 24: readRawSample uses the value of `u32 size`, not `perf_event_header.size`. From `man perf_event_open`:

```
size, data[size]
                      If PERF_SAMPLE_RAW is enabled, then a 32-bit value indicating size is included
                      followed by an array of 8-bit values of length size.  The  values  are  padded
                      with 0 to have 64-bit alignment.
```

What size is the sample you are submitting? Are you sure this isn't padding for the purpose of alignment?

Maybe you can add a test case that demonstrates the issue.

perf/reader_test.go
```diff
@@ -57,7 +57,7 @@ func TestPerfReader(t *testing.T) {
 		t.Fatal("Can't read samples:", err)
 	}
 
-	want := []byte{1, 2, 3, 4, 4, 0, 0, 0, 0, 0, 0, 0}
+	want := []byte{1, 2, 3, 4, 4, 0, 0, 0}
```
lmb, line 60: If you change `mustOutputSamplesProg(t, 5)` above to `mustOutputSamplesProg(t, 4)` you should see only []byte{1,2,3,4}. Your change will break this test.

syscalls.go
```diff
@@ -185,6 +186,18 @@ func bpfProgAlter(cmd int, attr *bpfProgAlterAttr) error {
 
 func bpfMapCreate(attr *bpfMapCreateAttr) (*internal.FD, error) {
 	fd, err := internal.BPF(_MapCreate, unsafe.Pointer(attr), unsafe.Sizeof(*attr))
+	if xerrors.Is(err, os.ErrPermission) {
+		// An EPERM being returned from bpf(2) could mean two things:
+		// - the user has no permissions to call bpf()
+		// - the user's rlimit is set too low to lock memory for bpf maps
+		// Since we cannot determine the exact amount of memory required,
+		// set the user's rlimit to unlimited to be able to create the map.
+		unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{Cur: unix.RLIM_INFINITY, Max: unix.RLIM_INFINITY})
```
lmb, line 195: We actually set a somewhat sensible rlimit in our application, to prevent it from mlocking all memory into place. This isn't a safe thing to do as a library.

elf_reader.go
lmb: You can't use data sections on 4.19. Setting flags here means that the check in createMap returns an internal.UnsupportedFeatureError, which the test suite knows to skip.

syscalls.go
ti-mo: I agree, map creations that exceed the rlimit should return a clearer error message (afaik this throws a generic 'insufficient memory' error) and/or the fact that rlimit needs to be taken into account by the user should be better documented on the exposed API. I'll think of a better solution. :+1:

elf_reader.go
```diff
@@ -523,14 +523,16 @@ func uintFromBTF(typ btf.Type) (uint32, error) {
 }
 
 func (ec *elfCode) loadDataSections(maps map[string]*MapSpec, dataSections map[elf.SectionIndex]*elf.Section, spec *btf.Spec) error {
-	if spec == nil {
-		return xerrors.New("data sections require BTF")
```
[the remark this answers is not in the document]
ti-mo: Okay, I figured out what about my program causes LLVM to emit an `.rodata` section. My program contains `const int ready_val = 0x90;`, as someone would naively write in C.

```
~ readelf -a build/bpf/acct/5.5.10.o
...
Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
...
  [11] .rodata           PROGBITS         0000000000000000  00001890
       0000000000000004  0000000000000000   A       0     0     4
```

Dumping the contents with `sec.Data()` in `loadDataSections()`, .rodata contains: `[144 0 0 0]`, or hex 0x90.

I'm not quite sure why LLVM needs to put this into a data section, because it's only ever used in comparisons like follows:
```
11:       55 01 14 01 90 00 00 00 if r1 != 144 goto +276 <LBB0_51>
```

This might be something that needs to be addressed in the BPF program itself, but without the nil check on `*btf.Spec` in `loadDataSections`, a call to `spec.Datasec()` will obviously panic. I'm not sure what the right path forward is here. We could just silently skip loading data sections if there's no BTF spec available, but that might lead to hard-to-debug surprises down the road.

perf/reader_test.go
ti-mo: Nice, great catch. This behaviour is related to your comment above https://github.com/cilium/ebpf/pull/94/files#r424357585.

elf_reader.go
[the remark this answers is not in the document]
lmb: The compiler can't eliminate `ready_val` because it's not marked as `static` :(

I guess you didn't want a data section, but got one due to missing `static` and thought the error to load is due to missing BTF?

syscalls.go
lmb: Yeah, having a better error would be nice, thanks for taking a look. I remember that the kernel returned something generic when it ran into the rlimit, maybe that has changed?

perf/reader.go
ti-mo: Indeed, now that part of the documentation makes more sense to me, but it's still not quite correct. I've taken another look at [`perf_output_sample()`](https://elixir.bootlin.com/linux/v5.5.13/source/kernel/events/core.c#L6611) and [`perf_prepare_sample()`](https://elixir.bootlin.com/linux/v5.5.13/source/kernel/events/core.c#L6827). Still not sure if the buffer sizing logic (in the kernel) is correct. The change I suggested here indeed breaks on payloads of 4 bytes and smaller; where popping a flat 4 bytes off the end of the array will just yield an empty slice.

The samples I'm submitting are over 100 bytes, so this didn't surface in my testing. Here, the last 4 bytes of the payload were zeroed until the 40-50th event, when they would start containing garbage, likely when the ring has  wrapped at least once.
```
 	want := []byte{1, 2, 3, 4, 4, 0, 0, 0, 0, 0, 0, 0}
                                               ^ these  ^
```

I've analyzed the [segment in `perf_prepare_sample()`](https://elixir.bootlin.com/linux/v5.5.13/source/kernel/events/core.c#L6827) and run the calculations by hand.

When sending a sample with `mustOutputSamplesProg(t, 4)`:
```
fragsize = 4

sum = 4
size = 8 (4 + sizeof32 = 8) aligned to sizeof64

raw->size = 4 (size - sizeof32)
frag->pad = 4 (raw->size - sum)

header->size += 8 (size)
```
Test output: `TestPerfReader: reader_test.go:62: [1 2 3 4]`, 4 bytes.

With `mustOutputSamplesProg(t, 5)`:
```
fragsize = 5

sum = 5
size = 16   (5 + sizeof32 = 9) aligned to sizeof64

raw->size = 12 (size - sizeof(32))
frag->pad = 7 (raw->size - sum)

header->size += 16 (size)
```
Test output: `TestPerfReader: reader_test.go:62: [1 2 3 4 4 0 0 0 0 0 0 0]`, 12 bytes.

In `perf_output_sample()`: `perf_output_put(handle, raw->size);`

The length of the **buffer** read from the ring seems to be of length `raw->size`, but its contents are of length `raw->size - sizeof(u32)`, with a floor of 4 bytes.

Does clamping the sample size to a minimum of 4 bytes sound reasonable @lmb ?

lmb: I don't think that works, because there is no way to recover the value of `pad` from the information stored in the ring buffer. Any sample between 5 and 5+7 bytes will have the same length in the buffer, but we can't tell where it ends. This is a shortcoming in the perf ring API.

The way you solve this is to include a length field in your raw sample if it is variable length. If it's fixed length you just ignore the tail.

Not sure where the garbage you mention comes from. I kind of assumed that the perf buffer zeroes out pad bytes.

ti-mo: Hmm, from reading that `man perf_event_open` snippet, I'm under the impression that the perf ring API is supposed to deliver buffers to userspace that are 64-bit aligned, except the alignment is incorrect. These are opaque buffers that the API itself knows nothing about. If the user wants to use those buffers to encode payloads with type-length-value and ship them to userspace, that's perfectly possible, but at least the buffers themselves should be 64-bit aligned.  That's not happening correctly at the moment, I would expect both `fragsize` 4 and 5 to yield 8 bytes. `pad` should just be used by the kernel internally to determine how many zeroes to append to the buffer so the alignment is correct; not sure why userspace would need access to that.

The initial idea of popping those 4 bytes came from here: https://github.com/iovisor/gobpf/blob/25bfcb50b85f18195f09cf4e6d9d525e985b4889/elf/perf.go#L208. I noticed the difference while migrating from gobpf, because there, whatever I wrote to the ring within BPF came out (64-bit aligned) in userspace, as expected. Same in BCC (Python).

I'll see if I can reproduce the garbage reads.

elf_reader.go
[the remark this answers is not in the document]
ti-mo: Yup! I'd like to extend the error message here to at least include the section name: `found data section %s, but the ELF binary did not contain a required BTF spec`.

Could we maybe add an extra message in case the offending section is `.rodata` to hint at consts with missing `static`? I have a feeling this might become a common issue for users. :)

elf_reader.go
ti-mo: Found the root cause in thread https://github.com/cilium/ebpf/pull/94/files#r424351006, thanks!

perf/reader.go
lmb: But then the unconditional size - 4 doesn't make sense either, no? If you submit 10 bytes it would cut off the last 2 bytes from the sample.

ti-mo: > But then the unconditional size - 4 doesn't make sense either, no?

Oh, of course. :) You are 100% right. That assumption only holds when sending structs to userspace, which 99% (if not more) of tracepoint/kprobe examples in bcc are doing. Structs will always be binpacked and aligned to 64 bits, which will always yield 4 extra bytes due to that kernel bug.

Looks like this cannot/should not be addressed in this library, but IMO it should be documented on the API, as promising 64-bit alignment is misleading. It ends up being more like '64-bit aligned with a possible extra 32 bits of uninitialized memory'.

Sorry for the annoying format, but I have a repro that proves reads return garbage after wrapping around the ring's page boundary for the first time. Add this to `reader_test.go`:

```go
func TestPerfReaderGarbage(t *testing.T) {
	sampleLen := 64

	prog, events := mustOutputSamplesProg(t, sampleLen)
	defer prog.Close()
	defer events.Close()

	rd, err := NewReader(events, 4096)
	if err != nil {
		t.Fatal(err)
	}
	defer rd.Close()

	b := 0
	for i := 0; i < 100; i++ {
		if _, _, err := prog.Test(make([]byte, 14)); err != nil {
			t.Fatal(err)
		}

		record, err := rd.Read()
		if err != nil {
			t.Fatal("Can't read samples:", err)
		}

		// Approximate the total amount of bytes read from the ring.
		// 4 bytes added to account for padding alignment between the
		// end of the sample (68 bytes) and the next header.
		b += binary.Size(perfEventHeader{}) + len(record.RawSample) + 4

		// The sampleLen'th byte should be padding.
		fmt.Println("iter", i, "after", b, "bytes:", record.RawSample)
		if record.RawSample[sampleLen] != 0 {
			t.Fail()
		}
	}
}
```

Returns the following output:

```
iter 49 after 4000 bytes: ...  4 3 2 1 0 0 0 0]
iter 50 after 4080 bytes: ...  4 3 2 1 0 0 0 0]
iter 51 after 4160 bytes: ...  4 3 2 1 1 2 3 4]
iter 52 after 4240 bytes: ...  4 3 2 1 1 2 3 4]
```

Thanks a lot for the reviews! Let me know what you think wrt. adding this as documentation to the API.

lmb: Interesting, is that kernel behaviour or a bug in the library?

Re documentation, I'm all for it. I was trying to do that with https://github.com/cilium/ebpf/blob/master/perf/reader.go#L62-L64 but reading that again it's not really very useful. It should mention that there are between 1-7 bytes of trailing garbage, and that you need your own size if you have variable length. Maybe you can add an example to the package?

ti-mo: The library just calls `io.ReadFull()` on buffers obtained from epoll, so there's not much that can go wrong there. I think in `perf_prepare_sample()`, instead of:

```
			size = round_up(sum + sizeof(u32), sizeof(u64));
			raw->size = size - sizeof(u32);
			frag->pad = raw->size - sum;
```

it should be something along the lines of:

```
			size = round_up(sum + sizeof(u32), sizeof(u64));
			raw->size = round_up(sum, sizeof(64));
			frag->pad = raw->size - sum;
```

I'll play around with it a bit. Looks like there's been a similar observation on the bcc bug tracker: https://github.com/iovisor/bcc/issues/2432.

ti-mo: I'm going to stop chasing this; to me it's clear that the buffer size calculation is not working as documented. There are probably multiple generations of perf tools that already depend on this behaviour, so not sure if this makes sense to fix upstream. I played around with a couple of builds, and this is what ends up correctly aligning the header and the `data` buffer: 

```c
void perf_prepare_sample(
...
			/* align the size of the data portion first */
			raw->size = round_up(sum, sizeof(u64));
			/* amount of padding required at end of data portion */
			frag->pad = raw->size - sum;
			/* calculate size of the sample on the ring buffer:
			   | size u32 | + | data[] aligned to u64 | */
			size = raw->size + sizeof(u32);
```

Lastly, the reason for the garbage reads from the ring is because the padding is never actually zeroed.. `__output_skip` is a no-op function.

```c
void perf_output_sample(
...
			if (frag->pad)
				__output_skip(handle, NULL, frag->pad);
```

In summary, I'd like to modify the documentation of `Record.RawSample` to reflect that it will:
- have a length of `(length of submitted sample + 32 bits), aligned to 64 bits, minus 32 bits`, and
- contain between 0 and 7 bytes of trailing garbage from the ring depending on the input sample's length.

Does this look correct? We should probably also add a hint to `Reader.Read()` about this.

A few examples:
- in: 4 bytes, out: 4 bytes
- in: 5 bytes, out: 12 bytes, 7 uninitialized
- in: 12 bytes, out: 12 bytes
- in: 13 bytes, out: 20 bytes, 7 uninitialized
- etc.

lmb: > I'm going to stop chasing this; to me it's clear that the buffer size calculation is not working as documented. There are probably multiple generations of perf tools that already depend on this behaviour, so not sure if this makes sense to fix upstream.

I think that's sensible, yes.

> Lastly, the reason for the garbage reads from the ring is because the padding is never actually zeroed.. `__output_skip` is a no-op function.
> 
> ```c
> void perf_output_sample(
> ...
> 			if (frag->pad)
> 				__output_skip(handle, NULL, frag->pad);
> ```

This might be fix-worthy upstream.

> In summary, I'd like to modify the documentation of `Record.RawSample` to reflect that it will:
> 
>     * have a length of `(length of submitted sample + 32 bits), aligned to 64 bits, minus 32 bits`, and
> 
>     * contain between 0 and 7 bytes of trailing garbage from the ring depending on the input sample's length.

I think you can drop the first bullet point, seems like a perf buffer implementation detail to me. The second point explains the issue more concisely.

> Does this look correct? We should probably also add a hint to `Reader.Read()` about this.

SGTM.

ti-mo: > This might be fix-worthy upstream.

Good idea, submitted this: https://lore.kernel.org/patchwork/patch/1244339.

> I think you can drop the first bullet point, seems like a perf buffer implementation detail to me. The second point explains the issue more concisely.

Agreed, will address all feedback soon. Thanks again for the reviews!

run-tests.sh
```diff
@@ -16,7 +16,7 @@ if [[ "${1:-}" = "--in-vm" ]]; then
   export GOCACHE=/run/go-cache
 
   echo Running tests...
-  /usr/local/bin/go test -coverprofile="$1/coverage.txt" -covermode=atomic -v ./...
+  go test -coverprofile="$1/coverage.txt" -covermode=atomic -v ./...
```
ti-mo, line 19: @lmb Not sure if this was an absolute path for a reason, but my Go executable is in another place.

ti-mo: Great, looks like `go` is not in Semaphore's path, I'll revert this.

ti-mo:
Hi @lmb, is this good to go? Any other feedback?

syscalls.go
```diff
@@ -185,6 +186,10 @@ func bpfProgAlter(cmd int, attr *bpfProgAlterAttr) error {
 
 func bpfMapCreate(attr *bpfMapCreateAttr) (*internal.FD, error) {
 	fd, err := internal.BPF(_MapCreate, unsafe.Pointer(attr), unsafe.Sizeof(*attr))
+	if xerrors.Is(err, os.ErrPermission) {
+		return nil, xerrors.Errorf("permission denied or insufficient rlimit to lock memory for map '%s'", attr.mapName)
```
lmb, line 190: mapName may be truncated or empty. I think callers will already include the name if possible, so skip the format here?

perf/reader_test.go
```diff
@@ -390,6 +391,54 @@ func TestPause(t *testing.T) {
...
+		if err != nil {
+			t.Fatal("Can't read samples:", err)
+		}
+
+		// Maintain a slice of zeroes the length of the padding section.
+		nb := make([]byte, len(record.RawSample)-sl)
+
+		// Compare the padding section to the empty slice.
+		if !reflect.DeepEqual(record.RawSample[sl:], nb) {
+			t.Skipf("%d non-zero bytes found in padding section of perf sample %d: %v", len(nb), i, record.RawSample[sl:])
```
lmb, line 436: Hmm, this isn't really a test. What do we gain by including it?

syscalls.go
ti-mo: Ah, the call to `newMapWithBTF` already contains the format, got it.

perf/reader_test.go
ti-mo: I think it's a valuable reproducer, and I hope this behaviour gets fixed in later kernels. As soon as that happens, this test will come up as passed, but on current kernels, it's not an error, hence the skip.

lmb: I agree it's a good reproducer, but it since it doesn't really have an error case there isn't much point to running it in CI, etc. Even when the bug is fixed the test doesn't increase the quality of the lib / protect us against regressions.

Re maintenance: the reproducer will fail on systems that have a non 4k page size, since in that case the perf ring will be larger than the 4k you ask for. So to really make this work you'd have to try and emulate what the test for lost perf samples does. If you take a look you'll see it's quite involved and I'd rather avoid that.

ti-mo: Alright, I've removed it then. It wouldn't protect against regressions, but it would serve as a canary against making assumptions about the buffer length, contents or alignment.

> the reproducer will fail on systems that have a non 4k page size

8k seems to be the lowest common denominator, which would be easy to account for by bumping the cycle count across the board.

PR #95, opened 2020-05-16
Get system specific tracepoint id

florianl:

This PR reads the system specific tracepoint id instead of using the hard coded id.


mythi:
Thanks!

Issue #83, opened 2020-03-23
support attaching to tracepoints


mythi:
`cilium/ebpf` does not seem to provide functions to enable tracepoints with BPFs.  Would you be OK to accept changes that help to do that?

lmb:
Hi! We had a similar discussion on Slack a while ago. Here is the thread for posterity:

```
Grant Seltzer Feb 22nd at 1:04 AM
Hi everyone, I'm excited about this package but from reading documentation I don't get the impression that I can load a uprobe using it yet, is that correct?

Joe Stringer  1 month ago
Correct. At this stage this library is not intending to take over all BPF use case like tracing, it's more focused on networking use cases where the BPF programs are managed by long-running daemons.

Grant Seltzer  27 days ago
Would there be interest in that feature if I were to try to implement it? I think having static go binaries that can do eBPF tracing would be widely appreciated

Lorenz  27 days ago
Do you have a rough idea what features you want? Contributions are always welcome of course, the main questions for me are who will maintain that code, and how much additional API surface does it expose?

Lorenz  27 days ago
Especially allowing uprobes on old kernels will be difficult i think

pbnj  26 days ago
Grant, have you seen this project: https://github.com/iovisor/gobpf

Grant Seltzer  26 days ago
@pbnj Yes, I'm currently using that for a project i'm working on: https://github.com/grantseltzer/weaver

Grant Seltzer  26 days ago
I could be misunderstanding the function and purpose of this package, but the iovisor gobpf package requires dynamic linking/CGO. If there were a purely go alternative then I could make my program which instruments ebpf into a static binary, right?

Lorenz  24 days ago
yes, with the caveat that you need to ship your compiled ebpf somehow

Lorenz  24 days ago
since you're using uprobes, do you rely on CO-RE / BTF at all?

Lorenz  24 days ago
you can try the following, and see how it goes: turn the gobpf C into something you can compile with vanilla clang, and try to load it with the library. if you put the function in a uprobe section the library should recognize it: https://github.com/cilium/ebpf/blob/master/elf_reader.go#L542 (edited) 
elf_reader.go:542

        "uretprobe/":      Kprobe,

Lorenz  24 days ago
at that point you can take a look at adding a function that attaches the ebpf to the user function, not sure how that works
```

[...]

PR #96, opened 2020-05-20
Simplify attaching to raw tracepoints

buglloc:
First of al,l thanks for the great project! :+1: 

I use `github.com/cilium/ebpf` to monitor various system events through raw tracepoints. This PR helps to use it (not needed to copy-paste BPF syscall wrapper and other).

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -494,6 +494,23 @@ func (p *Program) Detach(fd int, typ AttachType, flags AttachFlags) error {
 	return bpfProgAlter(_ProgDetach, &attr)
 }
 
+// Attach a Program to raw tracepoint.
+//
+// Requires at least Linux 4.17.
+func (p *Program) AttachRawTracepoint(tpName string) error {
+	fd, err := p.fd.Value()
+	if err != nil {
+		return xerrors.Errorf("failed to get program FD: %w", err)
+	}
+
+	attr := bpfRawTracepointOpenAttr{
+		name: internal.NewStringPointer(tpName),
+		fd:   fd,
+	}
+
+	return bpfRawTracepointOpen(&attr)
+}
+
 // LoadPinnedProgram loads a Program from a BPF file.
 //
 // Requires at least Linux 4.11.
--- a/prog_test.go
+++ b/prog_test.go
@@ -425,6 +425,29 @@ func TestProgramAlter(t *testing.T) {
 	}
 }
 
+func TestProgramAttachRawTracepoint(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "4.17", "BPF_RAW_TRACEPOINT Api")
+
+	var err error
+	var prog *Program
+	prog, err = NewProgram(&ProgramSpec{
+		Type: RawTracepoint,
+		Instructions: asm.Instructions{
+			asm.LoadImm(asm.R0, 0, asm.DWord),
+			asm.Return(),
+		},
+		License: "MIT",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	if err := prog.AttachRawTracepoint("cgroup_mkdir"); err != nil {
+		t.Fatal(err)
+	}
+}
+
 func TestHaveProgTestRun(t *testing.T) {
 	testutils.CheckFeatureTest(t, haveProgTestRun)
 }
--- a/syscalls.go
+++ b/syscalls.go
@@ -140,6 +140,11 @@ type bpfProgAlterAttr struct {
 	attachFlags uint32
 }
 
+type bpfRawTracepointOpenAttr struct {
+	name internal.Pointer
+	fd   uint32
+}
+
 type bpfObjGetInfoByFDAttr struct {
 	fd      uint32
 	infoLen uint32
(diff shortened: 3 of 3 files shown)


lmb:
Glad it's useful for you :)

Take a look at #83 where we discussed a similar proposal. The gist: there are many attachment points of BPF, and supporting them all sounds like a burden.

OTOH I see that this is only a small amount of code, and it would make the library more useful out of the box. Maybe we can put it in a subpackage? That way it could still access the internal stuff. @tklauser @florianl @joestringer thoughts?

This would raise another question: do we add a separate subpackage for #83 ? Or should that be unified with raw tracepoints (what is the difference)? Plus, can we move Program.Attach, Program.Detach to a subpackage as well? It seems way less general than I remembered, especially now that bpf_link is supposed to come and be the new thing.

florianl:
I think we should not mix it with https://github.com/cilium/ebpf/issues/83. As far as I know tracepoints are different to raw tracepoints from eBPF point of view. To attach eBPF programs to regular tracepoints you have to use `ioctl` while there is a own struct [1] for raw tracepoints in the `union bpf_attr`. And I agree that the handling of `ioctl` is out of scope of this package.
But attaching a eBPF program to a raw tracepoint should be part of this package or a subpackage, I think.
With bpf link on the horizon, I think, it should coexit with existing attach/detach methods.

[1] https://github.com/torvalds/linux/blob/444565650a5fe9c63ddf153e6198e31705dedeb2/include/uapi/linux/bpf.h#L546-L549

lmb:
I see. Hooray for consistent and easy to understand naming...

@buglloc do you feel like moving the function into a subpackage? `raw_tracepoints` maybe? With some documentation that the package may vanish at any point in time :P

buglloc:
> @buglloc do you feel like moving the function into a subpackage? raw_tracepoints maybe? With some documentation that the package may vanish at any point in time :P

Yessir!

lmb:
I'm currently looking into #102, and I think that will in some way involve adding support for raw tracepoints (not entirely sure yet). Just thought I'd give you a heads up so that you don't end up wasting your time.

raw_tracepoints/attach.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"golang.org/x/xerrors"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+// Attach a Program fd to a raw tracepoint.
+//
+// Requires at least Linux 4.17.
+func AttachFD(fd int, tpName string) error {
```
florianl, line 12: Can you update the comment and use the commonly used style?
```
// AttachFD ...
```

raw_tracepoints/attach_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal/testutils"
+	"github.com/cilium/ebpf/raw_tracepoints"
+)
+
+func TestProgramAttachRawTracepoint(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "4.17", "BPF_RAW_TRACEPOINT Api")
```
florianl, line 13: Can we capitalize `API`?

raw_tracepoints/syscalls.go
```diff
@@ -0,0 +1,21 @@
+package raw_tracepoints
+
+import (
+	"unsafe"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+const (
+	rawTracepointOpen = 17
```
florianl, line 10: To avoid import loops, we can not use `_RawTracepointOpen` from [1]. But keeping the same name schema would be great.

[1] https://github.com/cilium/ebpf/blob/master/types.go#L112

buglloc:
Sorry for the long delay, have a lot of work :(

Fixed previous issues and moved raw tracepoint to the `link` subpackage (as far as i'm concerned, it's more appropriate for now)

link/syscalls.go
```diff
@@ -134,3 +134,17 @@ func bpfIterCreate(attr *bpfIterCreateAttr) (*internal.FD, error) {
 	}
 	return nil, err
 }
+
+type bpfRawTracepointOpenAttr struct {
+	name internal.Pointer
+	fd   uint32
+	pad  uint32
```
florianl, line 141: `pad` is not part of the `BPF_RAW_TRACEPOINT` `bpf_attr`. Can we use `_` instead?

buglloc: Oh, sure. my bad :)

buglloc:
Thanks! :)

tklauser:
Thanks for your contribution @buglloc :rocket:

link/raw_tracepoint.go
```diff
@@ -0,0 +1,54 @@
...
+var _ Link = (*progAttachRawTracepoint)(nil)
+
+func (rt *progAttachRawTracepoint) isLink() {}
+
+func (rt *progAttachRawTracepoint) Close() error {
+	return rt.fd.Close()
+}
+
+func (rt *progAttachRawTracepoint) Update(_ *ebpf.Program) error {
+	return fmt.Errorf("can't update raw_tracepoint: %w", ErrNotSupported)
```
lmb, line 49: Is there no way to support `Update`? I would like progAttachRawTracepoint to pass `testLink()` which currently fails due to this.

buglloc: Honestly, I wasn't sure if it was useful. Implemented it in #122 (and uses `testLink` in tests)

PR #97, opened 2020-05-21
Correctly handle BPF and BTF endianness

lmb:
There are still endianness bugs in the library. Even though we can now parse and load ELFs in host endianness correctly, trying to parse a big-endian ELF on a little-endian system will give bogus results. This PR does two things:

* Make it possible to parse BPF and BTF that is not in native endianness
* Disallow loading BPF and BTF that is not in native endianness into the kernel

This means users can inspect ELFs of arbitrary endianness, while protecting them from hard to diagnose endianness problems should they try to load their programs.


asm/instruction.go
```diff
@@ -40,10 +39,12 @@ func (ins *Instruction) Unmarshal(r io.Reader, bo binary.ByteOrder) (uint64, err
 	}
 
 	ins.OpCode = bi.OpCode
-	ins.Dst = bi.Registers.Dst()
-	ins.Src = bi.Registers.Src()
 	ins.Offset = bi.Offset
 	ins.Constant = int64(bi.Constant)
+	ins.Dst, ins.Src, err = bi.Registers.Unmarshal(bo)
+	if err != nil {
+		return 0, xerrors.Errorf("unmarshal registers: %s", err)
```
tklauser, line 46: Wrap the `error` value with `%w`. Also maybe adjust the error message a bit:

```suggestion
		return 0, xerrors.Errorf("can't unmarshal registers: %w", err)
```

asm/instruction.go
```diff
@@ -76,9 +77,14 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 		cons = int32(uint32(ins.Constant))
 	}
 
+	regs, err := newBPFRegisters(ins.Dst, ins.Src, bo)
+	if err != nil {
+		return 0, xerrors.Errorf("can't marshal registers: %s", err)
```
tklauser, line 82: Same here:

```suggestion
		return 0, xerrors.Errorf("can't marshal registers: %w", err)
```

lmb:
Thanks for the review. Good point re adding "can't". I'm not sure wrt error wrapping: is there an error we need to expose to the user here?

tklauser:
> Thanks for the review. Good point re adding "can't". I'm not sure wrt error wrapping: is there an error we need to expose to the user here?

Personally, I usually consider it good practice to wrap underlying errors, especially in libraries. That way interested users can easily get them, while it doesn't really hurt users not using them. Also, we seem to wrap underlying errors in other cases in this package, so we would stay consistent. That being said, currently there seems to be only one possible underlying error, so it might be overkill in this case.

lmb:
Right, that's what I used to do as well. I changed my approach after reading https://blog.golang.org/go1.13-errors#TOC_5. I'm not entirely happy with the result of that, it requires lots of attention when making changes like this. For now I'll stick to not wrapping though.

PR #98, opened 2020-05-21
Expose EEXIST

florianl:

If you update a hash map with the flag `UpdateNoExist` you expect an error if the key already exists. This PR exposes this error and let the user of the package handle it properly.

### DIFF
--- a/map.go
+++ b/map.go
@@ -14,6 +14,7 @@ import (
 // Errors returned by Map and MapIterator methods.
 var (
 	ErrKeyNotExist      = xerrors.New("key does not exist")
+	ErrKeyExist         = xerrors.New("key does exist")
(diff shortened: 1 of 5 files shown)


map.go
```diff
@@ -14,6 +14,7 @@ import (
 // Errors returned by Map and MapIterator methods.
 var (
 	ErrKeyNotExist      = xerrors.New("key does not exist")
+	ErrKeyExist         = xerrors.New("key does exist")
```
tklauser, line 17: I'd suggest to slightly rephrase the error message:

```suggestion
	ErrKeyExist         = xerrors.New("key already exists")
```

florianl:
Thanks for the review. I also added a test.

PR #100, opened 2020-05-28
asm: support printing unsupported opcode

mmisono:
For example:
```
var ins asm.Instruction
ins.OpCode = 6
fmt.Printf("%v\n", ins)
// Result: OpCode(0x6)
```

Previously this causes a stack overflow error.


lmb:
Thanks for the fix! I'm surprised that the %x verb falls back to op.String().

PR #105, opened 2020-06-09
Add support for attaching to cgroups and bpf_iter

lmb:
bpf_link appeared in 5.7 and there is a push to migrate existing BPF hooks to using this infrastructure. For example, bpf_iter is only useable via bpf_link. Since I need to experiment with bpf_iter I decided to add link support. As you can tell this ended up being a larger PR than I anticipated. The reasons are as follows:
* Based on recent feedback and issues opened by users I think that it's important to add some sort of attachment functionality to the library. Without it there is a really steep curve to use it.
* However, some attachments need to use different mechanisms on different kernel versions (looking at you, cgroups). I think users shouldn't have to care about that, if possible.
* To figure out what the API would look like I went ahead and implemented version independent cgroup support.

In terms of API, this PR adds a new subpackage `link`. It contains low-level APIs to do attaching (RawAttachProgram, RawLink, etc.) as well as higher level APIs that are nicer to use. The higher level APIs all conform to the `Link` interface. This is the output of `go doc`:

```
package link // import "github.com/cilium/ebpf/link"


VARIABLES

FUNCTIONS

    RawAttachProgram is a low level wrapper around BPF_PROG_ATTACH.

    You should use one of the higher level abstractions available in this
    package if possible.

    RawDetachProgram is a low level wrapper around BPF_PROG_DETACH.

    You should use one of the higher level abstractions available in this
    package if possible.


TYPES

type CgroupOptions struct {
	// Path to a cgroupv2 folder.
	Path string
	// One of the AttachCgroup* constants
	Attach ebpf.AttachType
	// Program must be of type CGroup*, and the attach type must match Attach.
	Program *ebpf.Program
}

type Iter struct {
	// Has unexported fields.
}
    Iter represents an attached bpf_iter.

func AttachIter(opts IterOptions) (*Iter, error)
    AttachIter attaches a BPF seq_file iterator.

func LoadPinnedIter(fileName string) (*Iter, error)
    LoadPinnedIter loads a pinned iterator from a bpffs.

func (it *Iter) Close() error
    Close implements Link.

func (it *Iter) Open() (io.ReadCloser, error)
    Open creates a new instance of the iterator.

    Reading from the returned reader triggers the BPF program.

func (it *Iter) Pin(fileName string) error
    Pin implements Link.

func (it *Iter) Update(new *ebpf.Program) error
    Update implements Link.

type IterOptions struct {
	// Program must be of type Tracing with attach type
	// AttachTraceIter. The kind of iterator to attach to is
	// determined at load time via the AttachTo field.
	//
	// AttachTo requires the kernel to include BTF of itself,
	// and it to be compiled with a recent pahole (>= 1.16).
	Program *ebpf.Program
}

type Link interface {
	// Replace the current program with a new program.
	//
	// Passing a nil program is an error.
	Update(*ebpf.Program) error

	// Persist a link by pinning it into a bpffs.
	//
	// May return an error wrapping ErrNotSupported.
	Pin(string) error

	// Close frees resources.
	//
	// The link will be broken unless it has been pinned. A link
	// may continue past the lifetime of the process if Close is
	// not called.
	Close() error

	// Has unexported methods.
}
    Link represents a Program attached to a BPF hook.

func AttachCgroup(opts CgroupOptions) (Link, error)
    AttachCgroup links a BPF program to a cgroup.

func LoadPinnedCgroup(fileName string) (Link, error)
    LoadPinnedCgroup loads a pinned cgroup from a bpffs.

type RawAttachProgramOptions struct {
	// File descriptor to attach to. This differs for each attach type.
	Target int
	// Program to attach.
	Program *ebpf.Program
	// Program to replace (cgroups).
	Replace *ebpf.Program
	// Attach must match the attach type of Program (and Replace).
	Attach ebpf.AttachType
	// Flags control the attach behaviour. This differs for each attach type.
	Flags uint32
}

type RawDetachProgramOptions struct {
	Target  int
	Program *ebpf.Program
	Attach  ebpf.AttachType
}

type RawLink struct {
	// Has unexported fields.
}
    RawLink is the low-level API to bpf_link.

    You should consider using the higher level interfaces in this package
    instead.

func AttachRawLink(opts RawLinkOptions) (*RawLink, error)
    AttachRawLink creates a raw link.

func LoadPinnedRawLink(fileName string) (*RawLink, error)
    LoadPinnedRawLink loads a persisted link from a bpffs.

func (l *RawLink) Close() error
    Close breaks the link.

    Use Pin if you want to make the link persistent.

func (l *RawLink) Pin(fileName string) error
    Pin persists a link past the lifetime of the process.

    Calling Close on a pinned Link will not break the link until the pin is
    removed.

func (l *RawLink) Update(new *ebpf.Program) error
    Update implements Link.

func (l *RawLink) UpdateArgs(opts RawLinkUpdateOptions) error
    UpdateArgs updates a link based on args.

type RawLinkOptions struct {
	// File descriptor to attach to. This differs for each attach type.
	Target int
	// Program to attach.
	Program *ebpf.Program
	// Attach must match the attach type of Program.
	Attach ebpf.AttachType
}
    RawLinkOptions control the creation of a raw link.

type RawLinkUpdateOptions struct {
	New   *ebpf.Program
	Old   *ebpf.Program
	Flags uint32
}
    RawLinkUpdateOptions control the behaviour of RawLink.UpdateArgs.
```

This PR will break users of Program.Attach and Program.Detach, but they can migrate to link.RawAttachProgram, etc.


internal/btf/btf.go
```diff
@@ -108,42 +111,82 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
...
 		strings:   rawStrings,
-		funcInfos: funcInfos,
-		lineInfos: lineInfos,
-		byteOrder: file.ByteOrder,
+		byteOrder: bo,
 	}, nil
 }
 
+var (
+	kernelSpec   *Spec
```
florianl, line 156: Instead of two plain variables, I would prefer something like `type featureTest struct {}`.

PR #107, opened 2020-06-12
hasPerCPUValue function add MapType LRUCPUHash

yeze:
Hi lmb,
I have added a test for LRUCPUHash, LRUCPUHash is PerCPU map that store a distinct value for each CPU, so from userspace we need to see it in the form of an array


lmb:
Hi yeze,

Can you please add a test that exercises LRUCPUHash? As far as I understand it that map type doesn't fit the `hasPerCPUValue` in that from userspace you still only see a single value, not an array of values.

PR #108, opened 2020-06-12
Add support for raw bpf_link

lmb:
bpf_link appeared in 5.7 and there is a push to migrate existing BPF hooks to using this infrastructure. For example, bpf_iter is only useable via bpf_link. This PR adds a new subpackage link. It contains low-level APIs to do attaching. I'll submit the higher-level APIs based on `Link` in a follow up to this PR.

```
package link // import "github.com/cilium/ebpf/link"

type Link interface{ ... }
type RawAttachProgramOptions struct{ ... }
type RawDetachProgramOptions struct{ ... }
type RawLink struct{ ... }
    func AttachRawLink(opts RawLinkOptions) (*RawLink, error)
    func LoadPinnedRawLink(fileName string) (*RawLink, error)
type RawLinkOptions struct{ ... }
type RawLinkUpdateOptions struct{ ... }
```

### DIFF
--- a/link/syscalls.go
+++ b/link/syscalls.go
@@ -1,0 +1,28 @@
+package link
+
+import (
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+)
+
+var haveProgAttach = internal.FeatureTest("BPF_PROG_ATTACH", "4.10", func() (bool, error) {
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type:       ebpf.CGroupSKB,
+		AttachType: ebpf.AttachCGroupInetIngress,
+		License:    "MIT",
+		Instructions: asm.Instructions{
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+	})
+	if err != nil {
+		return false, nil
+	}
+
+	// BPF_PROG_ATTACH was introduced at the same time as CGgroupSKB,
+	// so being able to load the program is enough to infer that we
+	// have the syscall.
+	prog.Close()
+	return true, nil
+})
--- a/link/syscalls.go
+++ b/link/syscalls.go
@@ -1,9 +1,14 @@
 package link
 
 import (
+	"unsafe"
+
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/unix"
+
+	"golang.org/x/xerrors"
 )
 
 var haveProgAttach = internal.FeatureTest("BPF_PROG_ATTACH", "4.10", func() (bool, error) {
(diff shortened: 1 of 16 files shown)


prog.go
```diff
@@ -454,53 +454,11 @@ func (p *Program) MarshalBinary() ([]byte, error) {
 	return buf, nil
 }
 
-// Attach a Program to a container object fd
-func (p *Program) Attach(fd int, typ AttachType, flags AttachFlags) error {
```
florianl, line 458: Instead of removing the complete function, could we keep the signature and use it as wrapper for `RawAttachProgram`? Same for `func (p *Program) Detach(fd int, typ AttachType, flags AttachFlags) error`. Maybe also add a comment that for newer kernel 5.7 link should be the prefer option.
As not everyone is already using 5.7 this might break a lot of stuff :/

lmb: Yeah, that is a good point. We could document these as deprecated, and then remove these functions at a later time. In general I don't want to keep these around forever, so at some point we'll have to introduce breakage.

lmb: Turns out this is impossible, since it creates an import cycle:

```
can't load package: import cycle not allowed
package github.com/cilium/ebpf
	imports github.com/cilium/ebpf/link
	imports github.com/cilium/ebpf
```

I don't think I can fix this.

tklauser: In that case (i.e. removing the functions) I'd strongly suggest to tag a release before this PR is merged.

lmb: I've moved stuff around, and just kept the old functions pretty much as they are. This means we don't break the API anymore.

link/syscalls.go
```diff
@@ -26,3 +31,55 @@ var haveProgAttach = internal.FeatureTest("BPF_PROG_ATTACH", "4.10", func() (boo
...
+type bpfLinkCreateAttr struct {
+	progFd     uint32
+	targetFd   uint32
+	attachType ebpf.AttachType
+	flags      uint32
+}
+
+func bpfLinkCreate(attr *bpfLinkCreateAttr) (*internal.FD, error) {
+	ptr, err := internal.BPF(internal.BPF_LINK_CREATE, unsafe.Pointer(attr), unsafe.Sizeof(*attr))
+	if err == nil {
```
jrfastab, line 44: seems a bit odd to `if good_case {return good} return bad` instead of `if  bad {bad} good;` but just a nit take it or leave it.

tklauser:
Not sure about the CI tests failing with Go 1.14/Kernel 4.19 (https://ebpf.semaphoreci.com/workflows/f9fafea7-370c-4a02-b588-262dd87076d5) but I assume it's a flake?

lmb:
@jrfastab the follow up to this includes actual usage of the API: https://github.com/cilium/ebpf/tree/iter-and-cgroup (top two commits)

link/syscalls.go
lmb: Good point, fixed

kolyshkin:
> the follow up to this includes actual usage of the API: https://github.com/cilium/ebpf/tree/iter-and-cgroup (top two commits)

This link is dead; I assume you meant https://github.com/cilium/ebpf/pull/110

PR #110, opened 2020-06-16
bpf_iter and cgroup support

lmb:
Use the low-level APIs to construct high-level `Link`s for attaching to cgroups and bpf_iter. Especially the cgroup support is interesting, because it falls back to BPF_PROG_ATTACH after trying bpf_link on older kernel.

New API:

```
+type CgroupOptions struct{ ... }
+type Iter struct{ ... }
+    func AttachIter(opts IterOptions) (*Iter, error)
+    func LoadPinnedIter(fileName string) (*Iter, error)
+type IterOptions struct{ ... }
 type Link interface{ ... }
+    func AttachCgroup(opts CgroupOptions) (Link, error)
+    func LoadPinnedCgroup(fileName string) (Link, error)
 type RawAttachProgramOptions struct{ ... }
 type RawDetachProgramOptions struct{ ... }
 type RawLink struct{ ... }
```

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -183,12 +183,13 @@ func (ec *elfCode) loadPrograms(progSections map[elf.SectionIndex]*elf.Section, 
 			return nil, xerrors.Errorf("program %s: can't unmarshal instructions: %w", funcSym.Name, err)
 		}
 
-		progType, attachType := getProgType(sec.Name)
+		progType, attachType, attachTo := getProgType(sec.Name)
 
 		spec := &ProgramSpec{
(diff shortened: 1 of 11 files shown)


lmb:
I'm not sure about the `LoadPinned*` methods. I think it might be possible to write a `LoadPinnedLink` function which uses BPF_GET_OBJ_INFO (or whatever that is called) to determine what kind of Link we are dealing with, which would make these obsolete.

lmb:
Yes, if we can make them conform to `Link` and if we don't need new dependencies.

PR #111, opened 2020-06-21
Fixed marshalBytes length check.

takehaya:
## Why
When I passed the structure to RewriteConstants, it said the following
```
marshal hoge.consts: hoge.Constants doesn't marshal to 28 bytes
```

You don't appear to have taken alignment into account.

case example.
* buffsize(cons) == sizeof(cons): https://play.golang.org/p/5-na-qFSF9R
* buffsize(cons) != sizeof(cons): https://play.golang.org/p/7pvZxwJZrOP

## What
Alignment considerations and check for errors fix.

Thank you :)

### DIFF
--- a/marshalers.go
+++ b/marshalers.go
@@ -55,7 +55,7 @@ func marshalBytes(data interface{}, length int) (buf []byte, err error) {
 		return nil, err
 	}
 
-	if len(buf) != length {
+	if int(reflect.TypeOf(data).Size()) != length && len(buf) != length {
 		return nil, xerrors.Errorf("%T doesn't marshal to %d bytes", data, length)
 	}
 	return buf, nil


lmb:
I don't think this will work. The C code expects a type that is X bytes, and you've provided one that is Y bytes when marshalled using `encoding/binary`. By changing the test you end up either truncating your Go value, or reading undefined bytes into your C value.

The library itself doesn't account for alignment or padding, it always uses the packed binary representation. You can change your C type to have a packed attribute or re-order items so that you don't have holes due to alignment. If that doesn't work you can implement https://golang.org/pkg/encoding/#BinaryMarshaler and account for the alignment you need.

takehaya:
Hi lmb. Thanks for the reply! :)
hmm...I want to fix this.

Let's take a look at the structure of c in pahole command
```
struct constants {                                                                                                                                                                             
        __u16                      a;                 /*     0     2 */                                                                                                                     
        /* XXX 2 bytes hole, try to pack */                                                                                                                                                    
        __u32                      b;                 /*     4     4 */                                                                                                                     
        __u32                      c;                /*     8     4 */                                                                                                                     
        __u16                      d;                 /*    12     2 */                                                                                                                     
        /* XXX 2 bytes hole, try to pack */                                                                                                                                                    
        __u32                      e;                 /*    16     4 */                                                                                                                     
        __u32                      f;                /*    20     4 */                                                                                                                     
        __u16                      g;                  /*    24     2 */                                                                                                                     
        __u16                      h;                  /*    26     2 */                                                                                                                     
        /* size: 28, cachelines: 1, members: 8 */                                                                                                                                              
        /* sum members: 24, holes: 2, sum holes: 4 */                                                                                                                                          
        /* last cacheline: 28 bytes */                                                                                                                                                         
};   
```
Following the previous example 

It looks like you're comparing the sum of the members to the size of the structure.
`len(buf) != length`-> `24 != 28`
This looks like a problem.
Is it possible to bring sum members(e.g 24) from the BTF for comparison?

lmb:
Let's pretend we have a structure like this:

```
struct constants {
        __u16                      a;                 /*     0     2 */                                                                                                                     
        /* XXX 2 bytes hole, try to pack */                                                                                                                                                    
        __u32                      b;                 /*     4     4 */ 
}
```

Let's look at the memory layout of this structure in C (with holes due to alignment) and Go (without holes):
```
| 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
|   a   |       |       b       |    (this is how C will arrange it in memory)
|   a   |       b       |            (this is how Go will arrange it in memory)
```

As you can see, the two don't agree on where `b` should be. So just changing the `length` check will create a bug: your C code will only see bytes 3 and 4 from the Go code. Bytes 6 and 7 will be unpredictable garbage.

The library could fix this by trying to account for alignment differences using BTF. However, I don't want to do that: the code will be complex and slow.

I think the right approach is to either change your definition like so:
```
struct constants {
       __u32                      b;     
        __u16                      a;                                                                                                                                         
}
```

Or to implement a custom encoding.BinaryMarshaler that introduces the holes in the right places.

I hope this explanation is clearer than the one before. Please open an issue if you think the library should handle this differently, then we can discuss there. I'll close this PR since it doesn't do what you want.

takehaya:
Thank you for the clear explanation :bulb: 
I understood. I'd like to rewrite the definition once and develop it! : )

PR #113, opened 2020-06-25
Improve libbpf compatibility

lmb:
Run the libbpf integration test on CI, and fix any issues that prevent this.

The next step is to try loading them into the kernel as well, but that requires additional fudging.


lmb:
I'll go ahead with merging this tomorrow if nobody objects. I think the contents are uncontroversial since they don't add API, and CI is green.

Issue #38, opened 2019-12-03
Migrate to xerrors


lmb:
We should migrate from github.com/pkg/errors to golang.org/x/xerrors. This probably means removing the various Is* functions in favour of sentinel errors. I'm also not entirely sure how to deal with `internal.UnsupportedFeatureError` yet. Possibly wrap it around a sentinel error?

By default, we should not wrap "foreign" errors originating from outside the package. Any functions calling other package functions should however do so. See https://blog.golang.org/go1.13-errors#TOC_3.4.

PR #116, opened 2020-07-02
Drop support for go 1.12, and remove golang.org/x/xerrors

thaJeztah:
Now that Go1.12 reached EOL, it should be possible to drop support for it, and to remove the transitional golang.org/x/xerrors
dependency.

This also fixes a minor bug, detected by Go 1.13+;

    # github.com/cilium/ebpf
    ./elf_reader.go:211: Errorf call needs 2 args but has 3 args
    FAIL    github.com/cilium/ebpf [build failed]

relates to https://github.com/cilium/ebpf/issues/37
relates to https://github.com/cilium/ebpf/issues/38

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"debug/elf"
 	"encoding/binary"
+	"errors"
+	"fmt"
 	"io"
 	"math"
 	"os"
@@ -13,8 +15,6 @@ import (
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/btf"
 	"github.com/cilium/ebpf/internal/unix"
-
-	"golang.org/x/xerrors"
 )
 
 type elfCode struct {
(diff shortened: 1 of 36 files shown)


elf_reader.go
```diff
@@ -207,8 +207,8 @@ func (ec *elfCode) loadPrograms(progSections map[elf.SectionIndex]*elf.Section,
 
 		if btfSpec != nil {
 			spec.BTF, err = btfSpec.Program(sec.Name, length)
-			if err != nil && !xerrors.Is(err, btf.ErrNoExtendedInfo) {
-				return nil, xerrors.Errorf("program %s: %w", sec.Name, funcSym.Name, err)
+			if err != nil && !errors.Is(err, btf.ErrNoExtendedInfo) {
+				return nil, fmt.Errorf("program %s %s: %w", sec.Name, funcSym.Name, err)
```
thaJeztah, line 211: This was missing a `%s` placeholder; not sure if this is the right "format" for the error, so suggestions welcome

thaJeztah:
@lmb PTAL

elf_reader.go
lmb, line 211: Oops! I think we should drop sec.Name instead:

```suggestion
				return nil, fmt.Errorf("program %s: %w", funcSym.Name, err)
```

thaJeztah:
Oh, thanks! Missed your earlier comment, but see you pushed a commit to fix 👍 ❤️

PR #120, opened 2020-07-31
Require vmlinux BTF only if necessary

buglloc:
Many good old BPF programs (i.e. `kprobe`, `tracepoint`, `raw_tracepoint` and so on) don't require a vmlinux BTF to be loaded (at least not yet). But `github.com/cilium/ebpf` started requiring it and broke loading programs from ELF on old kernels (or w/o `CONFIG_DEBUG_INFO_BTF`).

This PR fix this by loading kernel spec only if it required by program/attach type.


lmb:
LGTM, thanks for the fix!

PR #122, opened 2020-08-04
Support raw_tracepoint link update

buglloc:
afaik `BPF_LINK_UPDATE` doesn't support raw tracepoint updates (see [bpf_raw_tp_link_lops](https://github.com/torvalds/linux/blob/c0842fbc1b18c7a044e6ff3e8fa78bfa822c7d1a/kernel/bpf/syscall.c#L2656-L2661)). So we have only one way to update a raw tracepoint link - open it for a new program and close the previous fd.

### DIFF
--- a/link/raw_tracepoint.go
+++ b/link/raw_tracepoint.go
@@ -18,23 +18,17 @@ type RawTracepointOptions struct {
 //
 // Requires at least Linux 4.17.
 func AttachRawTracepoint(opts RawTracepointOptions) (Link, error) {
-	if opts.Program.FD() < 0 {
-		return nil, fmt.Errorf("invalid program: %w", internal.ErrClosedFd)
-	}
-
-	fd, err := bpfRawTracepointOpen(&bpfRawTracepointOpenAttr{
-		name: internal.NewStringPointer(opts.Name),
-		fd:   uint32(opts.Program.FD()),
-	})
-	if err != nil {
+	link := progAttachRawTracepoint{tpName: opts.Name}
+	if err := link.Update(opts.Program); err != nil {
 		return nil, err
 	}
 
-	return &progAttachRawTracepoint{fd: fd}, nil
+	return &link, nil
 }
 
 type progAttachRawTracepoint struct {
-	fd *internal.FD
+	tpName string
+	fd     *internal.FD
 }
 
 var _ Link = (*progAttachRawTracepoint)(nil)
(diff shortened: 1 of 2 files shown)


lmb:
You mentioned in the other PR that you don't see the use of update for raw tp, can you explain that a bit more? I don't really know much about how one would use these attach points in practice. Maybe the right thing to do is documenting that Update may fail as well and adjusting the tests appropriately.

link/raw_tracepoint.go
```diff
@@ -18,35 +18,47 @@ type RawTracepointOptions struct {
...
 }
 
-func (rt *progAttachRawTracepoint) Update(_ *ebpf.Program) error {
-	return fmt.Errorf("can't update raw_tracepoint: %w", ErrNotSupported)
+func (rt *progAttachRawTracepoint) Update(prog *ebpf.Program) error {
+	if prog.FD() < 0 {
+		return fmt.Errorf("invalid program: %w", internal.ErrClosedFd)
+	}
+
+	fd, err := bpfRawTracepointOpen(&bpfRawTracepointOpenAttr{
```
lmb, line 51: A couple of questions:
* If two raw tp programs are attached to the same tpName, do both get executed?
* Is there some limit on how many raw TPs can be active on a given function?

buglloc:
> You mentioned in the other PR that you don't see the use of update for raw tp, can you explain that a bit more? I don't really know much about how one would use these attach points in practice. Maybe the right thing to do is documenting that Update may fail as well and adjusting the tests appropriately.

From my point of view, a typical application using kprobe/tracepoint/raw_tracepoint looks like this:
  - attach probes
  - loop {reading ring buffer or map}
  - die

Examples:
  - https://github.com/draios/sysdig
  - https://github.com/gravitational/teleport
  - https://github.com/iovisor/bpftrace

That's why it's not clear for when in practice I would want to update a BPF program attached to raw_tp on the fly. Especially when we cannot do this in atomic way.


On the other hand, I can also miss something, and since the implementation is pretty simple - maybe let it be.

link/raw_tracepoint.go
buglloc: > If two raw tp programs are attached to the same tpName, do both get executed?

Yep, both would be executed

> Is there some limit on how many raw TPs can be active on a given function?

I don't know any limits for raw_tp

lmb:
> On the other hand, I can also miss something, and since the implementation is pretty simple - maybe let it be.

No, I think you're right. The fact that the kernel doesn't support this either is a good indication for this. The best solution is to document that Update can fail and then adjust the test.

PR #123, opened 2020-08-05
elf: recognize BPF_PROG_TYPE_SK_LOOKUP

lmb:
Add support for the new BPF socket lookup program and attach type.

### DIFF
--- a/types.go
+++ b/types.go
@@ -96,60 +96,37 @@ type ProgramType uint32
 
 // eBPF program types
 const (
-	// Unrecognized program type
 	UnspecifiedProgram ProgramType = iota
-	// SocketFilter socket or seccomp filter
 	SocketFilter
-	// Kprobe program
 	Kprobe
-	// SchedCLS traffic control shaper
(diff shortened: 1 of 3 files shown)


lmb:
cc @jsitnicki can you take a look?

types.go
```diff
@@ -190,6 +167,15 @@ const (
 	AttachModifyReturn
 	AttachLSMMac
 	AttachTraceIter
+	AttachCgroupInet4GetPeername
+	AttachCgroupInet6GetPeername
+	AttachCgroupInet4GetSockname
+	AttachCgroupInet6GetSockname
+	AttachXDPDevmap
```
jsitnicki, line 174: I think 'dev map' is two words. Like 'CPU map'.

jsitnicki:
Checked it against `include/uapi/linux/bpf.h` in bpf-next. LGTM.

types.go
lmb: Done, thanks.

PR #125, opened 2020-08-12
CI: use go 1.15

lmb:


lmb:
Seems like 1.15 isn't available yet.

PR #127, opened 2020-08-12
CI: verify that generated files are up to date

lmb:


lmb:
This currently doesn't work because there is no way to pin our `stringer` dependecy without introducing it as a dependency to users of the library. golang/go#36460 might fix this.

Issue #33, opened 2019-11-13
The difference between MapSpec and MapABI is unclear


lmb:
See #30, #26 and probably others. The difference between MapABI and MapSpec isn't entirely obvious. Either we get rid of MapABI somehow, or we document the differences.

One possibility is to make the fields in MapABI public in Map, so users could access Map.Type, etc. This creates the opportunity for race conditions however, so I'm not a big fan.

Maybe there is also a better name for MapABI (and friends).

nathanjsweet:
What about making MapABI an interface of which MapSpec can be an implementor?

lmb:
Thanks for reminding me to give this some more thought.

Users try to use MapABI and end up frustrated, because it doesn't expose the information they want. MapABI doesn't expose them because the library has to be consistent across the supported kernels, which includes 4.9. On 4.9 we can't get more information that what is in MapABI currently.

I think the solution to this is to introduce something like `MapInfo`: this encapsulates metadata about a loaded map from the kernel. This isn't the same as a MapSpec, which contains metadata needed to create a map. On recent kernels it's probably possible to create a MapSpec from a MapInfo, but on older ones this isn't true.

The biggest hurdle to `MapInfo` is how to expose "optional" metadata: things that are not available on all supported kernels. One idea was to just use pointers for those:
```
type MapInfo struct{
    KeySize uint32
    ...
    SomeOtherField *uint32
}
```
Once MapInfo exists we could unexport or remove MapABI.

PR #129, opened 2020-08-26
Rework ABI info Info structs, add some optional fields

lmb:
This adds the Tag field to ProgramABI, which is useful to check whether two BPF programs are in fact the same bytecode.


lmb:
cc @jsitnicki

lmb:
`/proc` doesn't have the prog tag on 4.9, so I'll need to rework this a bit.

jsitnicki:
LGTM. Will give it another look when you have the workaround for 4.9 ready.

lmb:
This is ready for review now. I ended up tackling #33, which has been a long standing source of confusion. The downside is that this is an API breaking change, so I plan to tag a release before merging this so that users don't get broken immediately.

@nathanjsweet @tklauser both of you have done some thinking on this before, so I'd really appreciate your input.

lmb:
This PR is now good to go, please take a look if you have some time.

tklauser (commented):
Very nice change, I think this makes the API of this package a lot easier to use.

One thing that worries me a bit are the pointer members in `type.MapInfo`/`type ProgramInfo` and that people might forget to check them for `nil` before accessing. This might work fine on the (newer) kernel they're testing their code on but would then fail e.g. when running their code on an older kernel where the field won't be set.

I'm not really sure this is a valid concern and something that could happen that often. But maybe we could think about unexporting the pointer members and adding convenience getters/setters for them instead?

info.go
```diff
@@ -61,6 +62,10 @@ func newMapInfoFromProc(fd *internal.FD) (*MapInfo, error) {
 // The pointer fields are not supported across all kernels, and may be nil.
 type ProgramInfo struct {
 	Type ProgramType
+	// Truncated hash of the BPF bytecode.
+	Tag *string
+	// Name as supplied by user space at load time.
+	Name *string
```
jsitnicki, line 75: Can this be just a `string` that is empty when attribute is not available?

jsitnicki:
> One thing that worries me a bit are the pointer members in `type.MapInfo`/`type ProgramInfo` and that people might forget to check them for `nil` before accessing.

You could go with Null Object Pattern and return `MapInfo` with `Type` set to `MAP_PROG_TYPE_UNSPEC` instead.

info.go
lmb: Yeah, then it's not possible to distinguish unnamed vs. no names supported. But maybe that distinction is useless in the first place.

We could do the same thing for tag, and define `""` to mean "tag is not supported" but then the naive comparison `a.Tag == b.Tag` does the wrong thing (maybe?) on older kernels where Tag is not available.

jsitnicki: The user has to be mindful either way. Whether its to check for `nil` pointer or for empty `string`.

lmb:
> But maybe we could think about unexporting the pointer members and adding convenience getters/setters for them instead?

These are the signatures I could come up with:
* `func Tag(default string) string`: if there is no tag, return the default. Probably not very useful, since there isn't always a natural default, and it doesn't allow determining whether the value was present or not.
* `func Tag() (string, error)`: this is overkill, since there would only ever be a single error.
* `func Tag() (string, bool)`: this is actually quite nice, and allows ignoring the bool via `tag, _ := Tag()`. The user is forced to deal with the fact that tag might not be present.

I think I'll give the last option a try. It has the added benefit that we can avoid a bunch of allocations just for the sake of putting things behind a pointer.

lmb:
Thank you both for your feedback, it was really valuable!

- For this PR, I've turned both Tag and Name into a plain `string`. It turns out that all kernels where we support getting `Info` (aka 4.10 and later) Tag is available. And for Name distinguishing between "has no name" and "kernel doesn't support names" doesn't seem useful.
- In the future, we should use the `func Thing() (type, bool)` for information that isn't always available.

tklauser (approved):
Thanks, that looks really nice now.

And I agree that `func Thing() (type, bool)` for information that isn't always available is a good solution for future changes.

PR #130, opened 2020-08-27
Specify numa node during map creation

flier:


map.go
```diff
@@ -184,12 +190,17 @@ func createMap(spec *MapSpec, inner *internal.FD, handle *btf.Handle) (*Map, err
 		}
 	}
 
+	if spec.NumaNode > 0 {
```
lmb, line 193: Looking at the kernel source, I think that node 0 is actually valid. So I think we can't "automagically" set this flag.

map.go
```diff
@@ -31,6 +34,9 @@ type MapSpec struct {
 	MaxEntries uint32
 	Flags      uint32
 
+	// Specify numa node during map creation
```
lmb, line 37: Please mention BPF_F_NUMA_NODE.

map.go
```diff
@@ -10,6 +10,9 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
 
+// Specify numa node during map creation
+const NumaMode uint32 = unix.BPF_F_NUMA_NODE
```
lmb, line 14: I'd prefer to not export the flags specifically. There is a ton of them, and maintaining them is an extra burden. Just using unix.BPF_F_NUMA_NODE directly should be fine I think.

lmb (changes requested):
Thanks for contributing! I left a few comments, please let me know what you think.

map.go
flier: Agreed, I think we can add a method to explicitly set the NUMA node
```go
// Specify numa node during map creation
func (ms *MapSpec) WithNumaNode(node uint32) {
	ms.Flags |= unix.BPF_F_NUMA_NODE
	ms.NumaNode = node
}
```

map.go
flier: In most cases the user does not need to set a NUMA node, even if there are multiple sockets that will be assigned to node 0 if not specified.

Anyway, we can remove it if you don't feel up to it.

map.go
flier:
```go
	// Specify numa node during map creation
	// (effective only if BPF_F_NUMA_NODE is set)
	NumaNode uint32
```

map.go
```diff
@@ -44,6 +48,13 @@ type MapSpec struct {
 	BTF *btf.Map
 }
 
+// Specify numa node during map creation
+func (ms *MapSpec) WithNumaNode(node uint32) *MapSpec {
```
lmb, line 52: I'm a bit torn about this. It adds a function for a very seldomly used feature. Plus, the function signature is very confusing: it suggests that the original MapSpec isn't modified, which isn't true.

I'd prefer it if you removed the function, and the only thing we keep is the NumaNode field + documentation.

lmb (changes requested):

map.go
flier: If we remove the function, we must re-expose the `unix.BPF_F_NUMA_NODE` flag, which is in an internal package, or apply the flag automatically when `NumaNode` is set.

lmb:
Why do we need to export the flag? You can just use unix.BPF_F_NUMA_NODE from your code and all is well.

flier:
> Why do we need to export the flag? You can just use unix.BPF_F_NUMA_NODE from your code and all is well.

As I mentioned, the `unix.BPF_F_NUMA_NODE` flag is in an internal package.

lmb:
Have a look at https://godoc.org/golang.org/x/sys/unix, which is what the internal package re-exports. `internal/unix` exists just so that the library typechecks on non-Linux platforms.

flier:
> Have a look at https://godoc.org/golang.org/x/sys/unix, which is what the internal package re-exports. `internal/unix` exists just so that the library typechecks on non-Linux platforms.

Understood, I will add a comment if you're okay with using external flags directly.

lmb:
Yes, thanks!

lmb (approved):
Thanks, and sorry for the miscommunication!

Issue #114, opened 2020-06-30
support compile once, run everywhere (CO-RE)


lmb:
libbpf supports modifying BPF bytecode to adjust for differences in kernel structure layouts. This is extremely useful for tracing programs that want to introspect kernel state, since you can now compile them against one kernel version but run them against another one.

Constituent parts:
- [x] Field based relocations (existence, size, offset)
- [x] Field signedness
- [x] Type based relocations
- [x] Enum based relocations
- [x] #384

brycekahle:
👋  I'm going to give implementing simple CO-RE relocations a try. If anyone has any suggestions, I'd love to hear them!

lmb:
Hi Bryce, great! Let me know if I can help you in any way. If you want we can use this issue to discuss issues / potential designs.

lmb:
@brycekahle thanks for the inital set of PRs! Sorry for the delay in reviewing them, we had a long weekend here.

brycekahle:
OK. I have relocations working in my branch, but I want to add some automated tests. Do you have any thoughts on how to approach testing? There are many examples in `libbpf`, but they all rely on global variables which are not currently supported in this library. Perhaps using a map instead for the same type of in/out data transfer?

lmb:
With global variables you mean read-write ones? There is support for constants with CollectionSpec.RewriteConstants.

Using a map is probably the most straight forward way.

ti-mo:
@lmb with bitfield relo's done, what's left here? What is meant by `Field signedness`?

lmb:
Signedness was actually implemented by Jussi as part of the bitfield work, so this is complete! 🎉

PR #131, opened 2020-08-28
Inflate BTF enum values

brycekahle:
Inflates BTF enum values for easier access later since the link between `Spec.rawTypes` and `Spec.types` is difficult to navigate.

### DIFF
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -513,7 +513,7 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
 			for i, btfVal := range rawvals {
 				name, err := rawStrings.LookupName(btfVal.NameOff)
 				if err != nil {
-					return nil, fmt.Errorf("can't get name for enum value %d: %w", i, err)
+					return nil, fmt.Errorf("can't get name for enum value %d: %s", i, err)
 				}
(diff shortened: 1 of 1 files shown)


brycekahle:
This is in support of #114

internal/btf/types.go
```diff
@@ -497,7 +508,19 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
 			typ = &Union{id, name, raw.Size(), members}
 
 		case kindEnum:
-			typ = &Enum{id, name}
+			rawvals := raw.data.([]btfEnum)
+			vals := make([]EnumValue, 0, len(rawvals))
+			for i, btfVal := range rawvals {
+				name, err := rawStrings.LookupName(btfVal.NameOff)
+				if err != nil {
+					return nil, fmt.Errorf("can't get name for enum value %d: %w", i, err)
```
lmb, line 516: Nit: `%s` is probably fine here?
```suggestion
					return nil, fmt.Errorf("can't get name for enum value %d: %s", i, err)
```

lmb (changes requested):
One nit, otherwise LGTM. Thanks!

brycekahle:
Updated!

PR #132, opened 2020-08-28
Add BTF Int encoding, offset, and bits fields

brycekahle:
In support of #114

### DIFF
--- a/internal/btf/btf_types.go
+++ b/internal/btf/btf_types.go
@@ -144,11 +144,6 @@ func (bt *btfType) SetLinkage(linkage btfFuncLinkage) {
 	bt.setInfo(uint32(linkage), btfTypeVlenMask, btfTypeVlenShift)
 }
 
-func (bt *btfType) IntEncoding() (IntEncoding, uint32, byte) {
-	// TODO: Panic here if wrong kind?
-	return IntEncoding((bt.Info & 0x0f000000) >> 24), (bt.Info & 0x00ff0000) >> 16, byte(bt.Info & 0x000000ff)
-}
-
 func (bt *btfType) Type() TypeID {
 	// TODO: Panic here if wrong kind?
 	return TypeID(bt.SizeType)
@@ -261,4 +256,8 @@ func readTypes(r io.Reader, bo binary.ByteOrder) ([]rawType, error) {
 
 		types = append(types, rawType{header, data})
 	}
+}
+
+func intEncoding(raw uint32) (IntEncoding, uint32, byte) {
+	return IntEncoding((raw & 0x0f000000) >> 24), (raw & 0x00ff0000) >> 16, byte(raw & 0x000000ff)
 }
--- a/internal/btf/btf_types.go
+++ b/internal/btf/btf_types.go
@@ -144,6 +144,11 @@ func (bt *btfType) SetLinkage(linkage btfFuncLinkage) {
 	bt.setInfo(uint32(linkage), btfTypeVlenMask, btfTypeVlenShift)
 }
 
+func (bt *btfType) IntEncoding() (IntEncoding, uint32, byte) {
+	// TODO: Panic here if wrong kind?
+	return IntEncoding((bt.Info & 0x0f000000) >> 24), (bt.Info & 0x00ff0000) >> 16, byte(bt.Info & 0x000000ff)
+}
+
 func (bt *btfType) Type() TypeID {
 	// TODO: Panic here if wrong kind?
 	return TypeID(bt.SizeType)
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -46,9 +46,9 @@ func (v *Void) walk(*copyStack) {}
 type IntEncoding byte
 
 const (
-	Signed IntEncoding = 1 << 0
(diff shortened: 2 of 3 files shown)


internal/btf/types.go
```diff
@@ -43,13 +43,24 @@ func (v *Void) size() uint32    { return 0 }
 func (v *Void) copy() Type      { return (*Void)(nil) }
 func (v *Void) walk(*copyStack) {}
 
+type IntEncoding byte
+
+const (
```
lmb, line 48: Nit:
```
const (
	Signed IntEncoding = 1 << iota
	Char
	Bool
)
```

internal/btf/types.go
```diff
@@ -466,7 +477,7 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
 
 		switch raw.Kind() {
 		case kindInt:
-			typ = &Int{id, name, raw.Size()}
+			typ = &Int{id, name, raw.Size(), IntEncoding((raw.Info & 0x0f000000) >> 24), (raw.Info & 0x00ff0000) >> 16, byte(raw.Info & 0x000000ff)}
```
lmb, line 480: Nit: can you turn the bitmasking into `decodeIntInfo() (IntEncoding, uint32, byte)` or something similar?

internal/btf/types.go
```diff
@@ -43,13 +43,24 @@ func (v *Void) size() uint32    { return 0 }
...
 // Int is an integer of a given length.
 type Int struct {
 	TypeID
 	Name
 
 	// The size of the integer in bytes.
-	Size uint32
+	Size     uint32
+	Encoding IntEncoding
+	Offset   uint32
```
lmb, line 62: What does `Offset` mean here?

lmb (changes requested):

internal/btf/types.go
brycekahle: From https://www.kernel.org/doc/html/latest/bpf/btf.html
> The BTF_INT_OFFSET() specifies the starting bit offset to calculate values for this int. For example, a bitfield struct member has:

> btf member bit offset 100 from the start of the structure,
btf member pointing to an int type,
the int type has BTF_INT_OFFSET() = 2 and BTF_INT_BITS() = 4
Then in the struct memory layout, this member will occupy 4 bits starting from bits 100 + 2 = 102.

> Alternatively, the bitfield struct member can be the following to access the same bits as the above:

> btf member bit offset 102,
btf member pointing to an int type,
the int type has BTF_INT_OFFSET() = 0 and BTF_INT_BITS() = 4
The original intention of BTF_INT_OFFSET() is to provide flexibility of bitfield encoding. Currently, both llvm and pahole generate BTF_INT_OFFSET() = 0 for all int types.

internal/btf/types.go
brycekahle: Fixed.

internal/btf/types.go
brycekahle: Done.

internal/btf/types.go
```diff
@@ -466,7 +477,8 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
 
 		switch raw.Kind() {
 		case kindInt:
-			typ = &Int{id, name, raw.Size()}
+			encoding, offset, bits := raw.IntEncoding()
```
lmb, line 480: I think this is incorrect: you're decoding BTF_INT_ENCODING, etc. from btf_type.info instead of the `u32` that follows btf_type. This has to use `raw.data.(uint32)` instead.

internal/btf/btf_types.go
```diff
@@ -144,6 +144,11 @@ func (bt *btfType) SetLinkage(linkage btfFuncLinkage) {
 	bt.setInfo(uint32(linkage), btfTypeVlenMask, btfTypeVlenShift)
 }
 
+func (bt *btfType) IntEncoding() (IntEncoding, uint32, byte) {
```
lmb, line 147: As per my other comment, this function shouldn't be on btfType. Just make it a free function.

internal/btf/types.go
lmb: Thanks! Can you add a comment that Offset is in bits, not bytes?

internal/btf/types.go
lmb: I think it'd be good if you could add a small test for this. You should be able to extend the test that parses `vmlinux-btf.gz` to call `FindType` to find a bitset (`struct iphdr` comes to mind https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/ip.h#L86).

lmb (changes requested):

internal/btf/types.go
brycekahle: Good catch! Fixed and added the test requested.

internal/btf/btf_types.go
brycekahle: Updated.

internal/btf/types.go
brycekahle: Added a comment.

lmb (approved):

PR #133, opened 2020-08-28
Add kind_flag support for Struct, Union, and Fwd

brycekahle:
This supports bitfield types for Struct & Union. kind_flag indicates whether Fwd is for a struct or union.

In support of #114

### DIFF
--- a/internal/btf/btf_types.go
+++ b/internal/btf/btf_types.go
@@ -139,7 +139,7 @@ func (bt *btfType) SetVlen(vlen int) {
 }
 
 func (bt *btfType) KindFlag() bool {
-	return int(bt.info(btfTypeKindFlagMask, btfTypeKindFlagShift)) == 1
+	return bt.info(btfTypeKindFlagMask, btfTypeKindFlagShift) == 1
 }
 
 func (bt *btfType) Linkage() btfFuncLinkage {
--- a/internal/btf/btf_types.go
+++ b/internal/btf/btf_types.go
@@ -40,10 +40,12 @@ const (
 )
 
 const (
-	btfTypeKindShift = 24
-	btfTypeKindLen   = 4
-	btfTypeVlenShift = 0
-	btfTypeVlenMask  = 16
+	btfTypeKindShift     = 24
+	btfTypeKindLen       = 4
+	btfTypeVlenShift     = 0
+	btfTypeVlenMask      = 16
+	btfTypeKindFlagShift = 31
+	btfTypeKindFlagMask  = 1
 )
 
 // btfType is equivalent to struct btf_type in Documentation/bpf/btf.rst.
@@ -134,6 +136,10 @@ func (bt *btfType) Vlen() int {
 
 func (bt *btfType) SetVlen(vlen int) {
 	bt.setInfo(uint32(vlen), btfTypeVlenMask, btfTypeVlenShift)
+}
+
+func (bt *btfType) KindFlag() bool {
+	return int(bt.info(btfTypeKindFlagMask, btfTypeKindFlagShift)) == 1
 }
 
 func (bt *btfType) Linkage() btfFuncLinkage {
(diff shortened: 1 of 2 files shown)


internal/btf/btf_types.go
```diff
@@ -136,6 +138,10 @@ func (bt *btfType) SetVlen(vlen int) {
 	bt.setInfo(uint32(vlen), btfTypeVlenMask, btfTypeVlenShift)
 }
 
+func (bt *btfType) KindFlag() bool {
+	return int(bt.info(btfTypeKindFlagMask, btfTypeKindFlagShift)) == 1
```
lmb, line 142:
```suggestion
	return bt.info(btfTypeKindFlagMask, btfTypeKindFlagShift) == 1
```

internal/btf/types.go
```diff
@@ -90,8 +90,9 @@ type Struct struct {
 	TypeID
 	Name
 	// The size of the struct including padding, in bytes
-	Size    uint32
-	Members []Member
+	Size     uint32
+	Members  []Member
+	KindFlag bool
```
lmb, line 95: I think we shouldn't expose KindFlag to users of the package, it's a detail of the BTF encoding. Can you perform the normalization for this at inflation stage?

Basically, `Member` would gain `BitfieldSize` and `BitOffset`.

lmb (changes requested):

internal/btf/types.go
```diff
@@ -159,6 +216,7 @@ func (e *Enum) copy() Type {
 type Fwd struct {
 	TypeID
 	Name
+	KindFlag bool
```
lmb, line 219: I forgot to address this: from https://github.com/torvalds/linux/commit/9d5f9f701b1891466fb3dbb1806ad97716f95cc3#diff-fa650a64fdd3968396883d2fe8215ff3 it seems that this determines whether the forward declaration is for a struct or union. I think a dedicated enum like IntEncoding would be nicer here than the boolean. If you don't need this for your patches feel free to remove KindFlag and punt on this as well.

brycekahle: I haven't needed it thus far, so I'll remove it for now.

internal/btf/types.go
brycekahle: Updated.

internal/btf/types.go
```diff
@@ -133,13 +137,28 @@ func (u *Union) copy() Type {
...
+
 // Member is part of a Struct or Union.
 //
 // It is not a valid Type.
 type Member struct {
 	Name
-	Type   Type
-	Offset uint32
+	Type         Type
+	Offset       uint32
```
lmb, line 159: Can you document what units these are in? I assume bits?

internal/btf/types.go
```diff
@@ -133,13 +137,28 @@ func (u *Union) copy() Type {
...
+	BitfieldSize uint32
+	BitOffset    uint32
```
lmb, line 161: This should go into Offset instead.

internal/btf/types.go
```diff
@@ -436,10 +455,15 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
...
 			}
-			members = append(members, Member{
+			m := Member{
 				Name:   name,
 				Offset: btfMember.Offset,
-			})
+			}
+			if kindFlag {
+				m.BitfieldSize = btfMember.Offset >> 24
+				m.BitOffset = m.Offset & 0xffffff
```
lmb, line 464:
```suggestion
				m.Offset &= 0xffffff
```

lmb (changes requested):

internal/btf/types.go
brycekahle: Fixed.

internal/btf/types.go
brycekahle: Added.

lmb (approved):

PR #134, opened 2020-08-28
Inflate BTF FuncProto params

brycekahle:
In support of #114

### DIFF
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -537,9 +537,11 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
 		case kindFuncProto:
 			rawparams := raw.data.([]btfParam)
 			params := make([]FuncParam, 0, len(rawparams))
-			for _, param := range rawparams {
-				// OK for func params to not have a name
-				name, _ := rawStrings.LookupName(param.NameOff)
+			for i, param := range rawparams {
+				name, err := rawStrings.LookupName(param.NameOff)
+				if err != nil {
+					return nil, fmt.Errorf("can't get name for func proto parameter %d: %s", i, err)
+				}
 				params = append(params, FuncParam{
 					Name: name,
 				})
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -233,13 +233,20 @@ func (f *Func) copy() Type {
 type FuncProto struct {
 	TypeID
 	Return Type
-	// Parameters not supported yet
+	Params []FuncParam
 }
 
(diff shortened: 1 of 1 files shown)


internal/btf/types.go
```diff
@@ -528,7 +535,20 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (namedTypes map
 			typ = fn
 
 		case kindFuncProto:
-			fp := &FuncProto{id, nil}
+			rawparams := raw.data.([]btfParam)
+			params := make([]FuncParam, 0, len(rawparams))
+			for _, param := range rawparams {
+				// OK for func params to not have a name
+				name, _ := rawStrings.LookupName(param.NameOff)
```
lmb, line 542: From `Documentation/bpf/btf.rst`:
```rst
If a BTF_KIND_FUNC_PROTO type is referred by a BTF_KIND_FUNC type, then
``btf_param.name_off`` must point to a valid C identifier except for the
possible last argument representing the variable argument. The btf_param.type
refers to parameter type.

If the function has variable arguments, the last parameter is encoded with
``name_off = 0`` and ``type = 0``.
```

We should only try to resolve a name if NameOff != 0. From the description it sounds a bit like the rule doesn't apply if FUNC_PROTO is not referred to by a FUNC, can you tell me more about this?

lmb (changes requested):

internal/btf/types.go
brycekahle: From the tests, it looks like attempting to resolve `0` is actually OK, and results in `""`. I can reinstate the error handling. https://github.com/cilium/ebpf/blob/4048cd641690f2f82d57b130467a3fdfe4967a06/internal/btf/strings_test.go#L25

brycekahle: Updated.

lmb: Yes, offset 0 is guaranteed to be the empty string by the spec.

internal/btf/types.go
```diff
@@ -233,15 +233,22 @@ func (f *Func) copy() Type {
 type FuncProto struct {
 	TypeID
 	Return Type
-	// Parameters not supported yet
+	Params []FuncParam
 }
 
 func (fp *FuncProto) walk(cs *copyStack) { cs.push(&fp.Return) }
```
lmb, line 239: I think this needs to push all fp.Params.Type instances as well.

brycekahle: Updated.

lmb (approved):

PR #136, opened 2020-09-04
Extend kernel BTF search to other likely locations

brycekahle:
Supports #114

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -12,10 +12,12 @@ import (
 	"os"
 	"reflect"
 	"sync"
+	"syscall"
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/unix"
+	linux "golang.org/x/sys/unix"
 )
 
 const btfMagic = 0xeB9F
@@ -59,29 +61,9 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 	}
 	defer file.Close()
 
-	var (
-		btfSection    *elf.Section
-		btfExtSection *elf.Section
-		sectionSizes  = make(map[string]uint32)
-	)
-
-	for _, sec := range file.Sections {
-		switch sec.Name {
-		case ".BTF":
-			btfSection = sec
-		case ".BTF.ext":
-			btfExtSection = sec
-		default:
-			if sec.Type != elf.SHT_PROGBITS && sec.Type != elf.SHT_NOBITS {
-				break
-			}
-
-			if sec.Size > math.MaxUint32 {
-				return nil, fmt.Errorf("section %s exceeds maximum size", sec.Name)
-			}
-
-			sectionSizes[sec.Name] = uint32(sec.Size)
-		}
+	btfSection, btfExtSection, sectionSizes, err := findBtfSections(file)
+	if err != nil {
+		return nil, err
 	}
 
 	if btfSection == nil {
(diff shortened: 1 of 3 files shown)


internal/btf/btf.go
```diff
@@ -12,10 +12,12 @@ import (
 	"os"
 	"reflect"
 	"sync"
+	"syscall"
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/unix"
+	linux "golang.org/x/sys/unix"
```
lmb, line 20: You have the wrong import here.

internal/btf/btf.go
```diff
@@ -176,16 +200,52 @@ func LoadKernelSpec() (*Spec, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("can't read kernel BTF: %s", err)
+		return nil, err
+	}
+
+	locations := []struct {
+		pathFmt string
+		raw     bool
+	}{
+		/* try canonical vmlinux BTF through sysfs first */
```
lmb, line 212: This list is from libbpf? Can you add a link to the source?

internal/btf/btf.go
```diff
@@ -176,16 +200,52 @@ func LoadKernelSpec() (*Spec, error) {
...
+		{"/usr/lib/modules/%s/kernel/vmlinux", false},
+		{"/usr/lib/debug/boot/vmlinux-%s", false},
+		{"/usr/lib/debug/boot/vmlinux-%s.debug", false},
+		{"/usr/lib/debug/lib/modules/%s/vmlinux", false},
+	}
+
+	var spec *Spec
+	for _, loc := range locations {
+		path := fmt.Sprintf(loc.pathFmt, release)
+		if err := syscall.Access(path, linux.R_OK); err != nil {
```
lmb, line 227: This seems strange, what's the point of doing this?

internal/btf/btf.go
```diff
@@ -176,16 +200,52 @@ func LoadKernelSpec() (*Spec, error) {
...
+
+		if err != nil {
+			continue
+		}
+		return spec, nil
 	}
-	defer fh.Close()
 
-	return loadNakedSpec(fh, internal.NativeEndian, nil, nil)
+	return nil, ErrNotFound
```
lmb, line 248: Why ErrNotFound, not `ErrNotSupported` like in the doc of `LoadKernelSpec`? Also, please wrap the error so that callers are forced to use `errors.Is` instead of just comparing.

internal/btf/btf.go
```diff
@@ -176,16 +200,52 @@ func LoadKernelSpec() (*Spec, error) {
...
 
 func loadKernelSpec() (*Spec, error) {
-	fh, err := os.Open("/sys/kernel/btf/vmlinux")
-	if os.IsNotExist(err) {
-		return nil, fmt.Errorf("can't open kernel BTF at /sys/kernel/btf/vmlinux: %w", ErrNotFound)
-	}
+	release, err := unix.KernelRelease()
 	if err != nil {
-		return nil, fmt.Errorf("can't read kernel BTF: %s", err)
+		return nil, err
```
lmb, line 205: Please add a `fmt.Errorf` that explains whats going wrong.

lmb (changes requested):

internal/btf/btf.go
brycekahle: This function was returning `ErrNotFound` previously, so I maintained that. I can change it.

internal/btf/btf.go
brycekahle: I was mirroring what `libbpf` does, but it looks like `access` checks for access using the real UID/GID rather than effective. That doesn't seem relevant here. I'll remove it. https://www.man7.org/linux/man-pages/man2/access.2.html

internal/btf/btf.go
brycekahle: Added

internal/btf/btf.go
brycekahle: Added

lmb (approved):

internal/btf/btf.go
```diff
@@ -176,16 +198,50 @@ func LoadKernelSpec() (*Spec, error) {
...
+
+		if err != nil {
+			continue
+		}
+		return spec, nil
 	}
-	defer fh.Close()
 
-	return loadNakedSpec(fh, internal.NativeEndian, nil, nil)
+	return nil, fmt.Errorf("can't find kernel BTF: %w", ErrNotSupported)
```
lmb, line 244: Sorry, you were right. ErrNotFound is the correct return value here.

```suggestion
	return nil, fmt.Errorf("BTF for kernel version %s: %w", release, ErrNotFound)
```

PR #137, opened 2020-09-04
Fix BTF map loading to use pointer target types and sizes

brycekahle:
For BTF based map loading, the `key` and `value` members of a map definition should always be pointers. This is because the `__type` macro is defined as `#define __type(name, val) typeof(val) *name`. Surprisingly, the kernel expects the type IDs passed to the `bpf` syscall, to be the "target" types and not the pointer types themselves. The `key_size` and `value_size` values should also be based on the "target" types.

I was receiving verifier errors without these changes.

Supports #114

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -466,35 +466,19 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
 		return nil, fmt.Errorf("can't get BTF: %w", err)
 	}
 
-	var btfKeyType, btfValueType btf.Type = (*btf.Void)(nil), (*btf.Void)(nil)
-	var keySize, valueSize uint32
 	keyType := btf.MapKey(btfMap)
-	if pk, isPtr := keyType.(*btf.Pointer); !isPtr {
-		if _, isVoid := keyType.(*btf.Void); !isVoid {
-			return nil, fmt.Errorf("key type is not a pointer: %T", keyType)
-		}
-	} else {
-		size, err := btf.Sizeof(pk.Target)
-		if err != nil {
-			return nil, fmt.Errorf("can't get size of BTF key: %w", err)
-		}
-		keySize = uint32(size)
-		btfKeyType = pk.Target
+	size, err := btf.Sizeof(keyType)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


map.go
```diff
@@ -47,6 +47,9 @@ type MapSpec struct {
 
 	// The BTF associated with this map.
 	BTF *btf.Map
+
+	btfKeyType   btf.Type
```
lmb, line 51: MapSpec shouldn't have unexported types, because a user may create one themselves.

elf_reader.go
```diff
@@ -466,19 +466,35 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
 		return nil, fmt.Errorf("can't get BTF: %w", err)
 	}
 
+	var btfKeyType, btfValueType btf.Type = (*btf.Void)(nil), (*btf.Void)(nil)
+	var keySize, valueSize uint32
 	keyType := btf.MapKey(btfMap)
-	size, err := btf.Sizeof(keyType)
-	if err != nil {
-		return nil, fmt.Errorf("can't get size of BTF key: %w", err)
+	if pk, isPtr := keyType.(*btf.Pointer); !isPtr {
```
lmb, line 472: `btf.Spec.Map` should unwrap the pointer for Key and Value. Then you don't have to change the code here, and can also avoid having to extend MapSpec.

lmb (changes requested):
Thanks for spotting this. How come the existing BTF map tests don't catch this?

There is a CI failure on 5.4, can you take a look?

map.go
brycekahle: OK. Fixed.

elf_reader.go
brycekahle: Fixed.

brycekahle:
@lmb CI failure went away with code changes. Should be good to go now.

brycekahle:
> How come the existing BTF map tests don't catch this?

The only BTF map related testing I see is `btf_map` and `btf_map2` in `testdata/loader.c` for clang >= 9. There are maps defined, but they are never used in program code. That may be why it wasn't caught. The verifier would complain about invalid offsets if `key_size` or `value_size` were wrong, and there was program code trying to use them.

lmb:
Thanks!

PR #138, opened 2020-09-07
small BTF tweaks

lmb:
I was experimenting with writing an `Equal` function for btf.Type. That didn't go far, but these small changes from that work still seem useful.


lmb:
cc @brycekahle the `qualifier` interface might be useful to your CO-RE series.

PR #143, opened 2020-10-01
program: move jump and call fixup out of asm package

lmb:
Jumps and calls may have to be fixed up, if their target isn't known
at compile time. This happens when a user uses the asm package to write
"inline" BPF programs, and when using function calls in C derived BPF.

The jump and call targets have to be expressed in units of raw BPF
instructions. Since some asm.Instructions have to be encoded in two
BPF instructions there is no 1:1 mapping. Figuring out the raw
instruction offset therefore involves iterating all instructions
and keeping a running count.

Due to this the fixup currently happens during instruction marshalling.
This is a bit of a hack, but so far allowed to keep the details of raw
instruction offsets contained to the asm package.

With BPF CO-RE this doesn't work too well anymore: the fixup we have to do
is also based on raw BPF instruction offsets. Cramming this into
instruction marshalling seems like the wrong thing to do, and is
probably not possible due to import cycles between the asm and btf
package.

Introduce a RawInstructionOffset type and an InstructionIterator to
encapsulate the raw offset handling, and move jump and call fixup to
a function in the main package.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -12,6 +12,14 @@ import (
 // InstructionSize is the size of a BPF instruction in bytes
 const InstructionSize = 8
 
+// RawInstructionOffset is an offset in units of raw BPF instructions.
+type RawInstructionOffset uint64
+
+// Bytes returns the offset of an instruction in bytes.
+func (rio RawInstructionOffset) Bytes() uint64 {
+	return uint64(rio) * InstructionSize
+}
+
 // Instruction is a single eBPF instruction.
 type Instruction struct {
 	OpCode    OpCode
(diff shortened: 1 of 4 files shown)


lmb:
cc @brycekahle please take a look, I think this should make it easier to apply CO-RE at load time. `RawInstructionOffset.Bytes()` should give you what you need to match relocations to instructions.

asm/instruction.go
```diff
@@ -370,65 +363,67 @@ func (insns Instructions) Format(f fmt.State, c rune) {
...
+// Modifying the instruction slice will lead to undefined behaviour.
+func (insns Instructions) Iterate() *InstructionIterator {
+	return &InstructionIterator{insns: insns}
+}
+
+// InstructionIterator iterates over a BPF program.
+type InstructionIterator struct {
+	insns Instructions
+	// The instruction in question.
+	Ins *Instruction
```
lmb, line 405: We could also embed `*Instruction` instead, which would make referring to Instruction.OpCode etc. nicer.

brycekahle (approved):

PR #144, opened 2020-10-01
map, program: deprecate ABI()

lmb:
Add explicit getters for fields that are currently exported via
ABI() and mark the functions as deprecated. They will be removed
in a follow up PR.


lmb:
@arthurfabre I know that you use MapABI.Equal for `xdpcap`, can you take a look at this PR? The follow up is going to remove the ABI structs and ABI.Equal completely, I hope that isn't too onerous for xdpcap.

PR #145, opened 2020-10-01
btf: fix LoadKernelSpec

lmb:
The current code fails to find any kernel spec because the call to
Sprintf may contain too many or too few arguments. The result then
contains %!s(MISSING), etc. which of course triggers an error when
trying to open the path.

Another problem is that the function tries other filenames if reading
the BTF itself fails for some reason. In that case it will most likely
return ErrNotFound, and the user is none the wiser: we found a BTF
but couldn't load it.

Change the code so that we return any error from the BTF loading itself.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -137,7 +137,7 @@ func findBtfSections(file *elf.File) (*elf.Section, *elf.Section, map[string]uin
 	return btfSection, btfExtSection, sectionSizes, nil
 }
 
-func loadNakedSpecFromReader(rd io.ReaderAt, sectionSizes map[string]uint32, variableOffsets map[variable]uint32) (*Spec, error) {
+func loadSpecFromVmlinux(rd io.ReaderAt) (*Spec, error) {
 	file, err := elf.NewFile(rd)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


lmb:
cc @brycekahle this is why CI fails. Now I'm mystified why it worked on the other kernels :shrug:

internal/btf/btf.go
```diff
@@ -203,42 +203,35 @@ func loadKernelSpec() (*Spec, error) {
...
-			spec, err = loadNakedSpec(fh, internal.NativeEndian, nil, nil)
-		} else {
-			spec, err = loadNakedSpecFromReader(fh, nil, nil)
-		}
 
-		if err != nil {
-			continue
-		}
-		return spec, nil
+		return loadSpecFromVmlinux(fh)
```
brycekahle, line 234: There isn't a guarantee that these files will have a `.BTF` section. Should we keep searching if that is the case?

lmb: I think that's unlikely to help: most systems will only have a single path where vmlinux is located (since it's a single package) so trying other location is unlikely to help. Trying other paths that also has the downside that we return ErrNotFound instead of an error describing why we can't use the existing BTF. That is really confusing to users (and to me when I started to look into the issue).

brycekahle (approved):

PR #147, opened 2020-10-03
make lsm attachment work

djwmarks2020:
BPF LSM requires that the kernel has CONFIG_BPF_LSM enabled, which is required for the BTF for LSM hooks to be present, and that the bpf security module is active, without which the attached programs will never be called.

Tested on an Ubuntu Groovy system with a modified kernel config.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math"
 	"strings"
 	"time"
@@ -619,6 +620,17 @@ func resolveBTFType(name string, progType ProgramType, attachType AttachType) (b
 
 	target := match{progType, attachType}
 	switch target {
+	case match{LSM, AttachLSMMac}:
+		if err := haveBpfLSM(); err != nil {
+			return nil, err
+		}
+
+		var target btf.Func
+		if err := findKernelType("bpf_lsm_"+name, &target); err != nil {
+			return nil, fmt.Errorf("can't resolve BTF for LSM hook %s: %w", name, err)
+		}
+
+		return &target, nil
 	case match{Tracing, AttachTraceIter}:
 		var target btf.Func
 		if err := findKernelType("bpf_iter_"+name, &target); err != nil {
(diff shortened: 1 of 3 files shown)


djwmarks2020:
BPF LSM requires that the kernel has CONFIG_BPF_LSM enabled, which is required for the BTF for LSM hooks to be present, and that the bpf security module is active, without which the attached programs will never be called.

Tested on an Ubuntu Groovy system with a modified kernel config.

prog.go
```diff
@@ -631,3 +643,18 @@ func resolveBTFType(name string, progType ProgramType, attachType AttachType) (b
 		return nil, nil
 	}
 }
+
+var haveBpfLSM = internal.FeatureTest("LSM", "5.8", func() (bool, error) {
+	lsm, err := ioutil.ReadFile("/sys/kernel/security/lsm")
+	if err != nil {
+		return false, fmt.Errorf("LSM unavailable: %w", ErrNotSupported)
```
lmb, line 650: The feature test should only return an error if it can't determine whether the feature is supported or not. Is it possible to have LSM enabled but this file not present? Or can we take the absence of the file to mean LSM is not enabled?

prog.go
```diff
@@ -631,3 +643,18 @@ func resolveBTFType(name string, progType ProgramType, attachType AttachType) (b
...
+	}
+
+	for _, module := range strings.Split(string(lsm), ",") {
+		if module == "bpf" {
+			return true, nil
+		}
+	}
+
+	return false, fmt.Errorf("BPF LSM not loaded: %w", ErrNotSupported)
```
lmb, line 659: I think at this point we know that the feature is not available.

```suggestion
	return false, nil
```

prog_test.go
```diff
@@ -479,6 +479,24 @@ func TestProgramRejectIncorrectByteOrder(t *testing.T) {
...
 
+func TestProgramTypeLSM(t *testing.T) {
+	prog, err := NewProgram(&ProgramSpec{
+		AttachTo:   "task_getpgid",
+		AttachType: AttachLSMMac,
+		Instructions: asm.Instructions{
+			asm.LoadImm(asm.R0, 0, asm.DWord),
+			asm.Return(),
+		},
+		License: "GPL",
```
lmb, line 490: Would you be OK to make this MIT, like the rest of the library?

lmb (changes requested):
Thanks for your PR! Two nits, otherwise it looks good to me :+1:

prog_test.go
```diff
@@ -479,6 +479,24 @@ func TestProgramRejectIncorrectByteOrder(t *testing.T) {
...
+		Type:    LSM,
+	})
+	testutils.SkipIfNotSupported(t, err)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
```
tklauser, line 497: Nit: I don't think we need the `defer` here as the function is going to return anyway:

```suggestion
	prog.Close()
```

lmb:
ping @djwmarks2020, do you still want to work on this PR?

lmb:
Closing this in favour of #156

PR #152, opened 2020-11-05
map: support pinning

lmb:
Add support for the "pinning" attribute in BTF-style map definitions,
following the libbpf implementation. If pinning is requested via
LIBBPF_PIN_BY_NAME, a newly created map is pinned into a BPF FS by
NewMapWithOptions. Any subsequent calls to NewMapWithOptions with
the same map name will reuse the existing map. Reuse fails if the
pinned map is deemed to be incompatible with the passed MapSpec.
This again follows libbpf semantics and compares type, key size,
value size, max elements and flags. In the future we could also
introduce BTF based checks.

The implementation deviates from libbpf by requiring that an
explicit path to pin into is provided via MapOptions. Hopefully
this will encourage users to use their own subdirectories instead
of polluting the global mount point.


florianl (approved):
looks good to me :+1:

PR #158, opened 2020-11-11
link: add AttachSkLookup

lmb:
Add a wrapper for SkLookup programs that allows retrieving information about the link.

### DIFF
--- a/link/link.go
+++ b/link/link.go
@@ -2,6 +2,7 @@ package link
 
 import (
 	"fmt"
+	"unsafe"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal"
@@ -32,6 +33,9 @@ type Link interface {
 	isLink()
 }
 
+// ID uniquely identifies a BPF link.
+type ID uint32
+
 // RawLinkOptions control the creation of a raw link.
 type RawLinkOptions struct {
 	// File descriptor to attach to. This differs for each attach type.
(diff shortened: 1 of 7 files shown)


jsitnicki:
Looks like nothing in `link/sk_lookup.go` is specific to BPF sk_lookup. BPF flow_dissector would be attached and loaded from pinned link the same way. Have you considered orienting it around BPF link type (`BPF_LINK_TYPE_NETNS`) instead of BPF program type?

lmb:
Good point, I'll do that.

lmb:
Fixed, PTAL.

link/link.go
```diff
@@ -159,3 +190,25 @@ func (l *RawLink) UpdateArgs(opts RawLinkUpdateOptions) error {
...
+	id      uint32
+	prog_id uint32
+}
+
+// Info returns metadata about the link.
+func (l *RawLink) Info() (*RawLinkInfo, error) {
+	var info bpfLinkInfo
+	err := internal.BPFObjGetInfoByFD(l.fd, unsafe.Pointer(&info), unsafe.Sizeof(info))
+	if err != nil {
+		return nil, fmt.Errorf("sk_lookup info: %s", err)
```
jsitnicki, line 206: Error message needs updating.

link/syscalls.go
```diff
@@ -10,6 +10,22 @@ import (
...
+//
+// Equivalent to enum bpf_link_type.
+const (
+	UnspecifiedType Type = iota
+	RawTracepointType
+	TracingType
+	CgroupType
+	IterType
+	NetNsType
+	XDPType
```
jsitnicki, line 26: Is that name unique enough? Why not `…LinkType`?

jsitnicki (approved):

link/link.go
lmb: Aye

link/syscalls.go
lmb: Because it would read as `link.XDPLinkType` which seems a bit redundant.

jsitnicki: Makes sense. Didn't notice it's in its own package.

PR #160, opened 2020-11-12
btf: handle type name flavours

lmb:
CO-RE has the concept of type "flavours", which are used to deal
with changes in kernel data structures. Basically, anything after
three underscores in a type name is ignored for the purpose of
finding a candidate type in the kernel's BTF.

Accomodate this by using the stripped name as the key to the
named type map, and adjusting FindType accordingly.

See https://facebookmicrosites.github.io/bpf/blog/2020/02/19/bpf-portability-and-co-re.html#dealing-with-kernel-version-and-configuration-differences

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -31,7 +31,7 @@ var (
 type Spec struct {
 	rawTypes  []rawType
 	strings   stringTable
-	types     map[string][]Type
+	types     map[string][]namedType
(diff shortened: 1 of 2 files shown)


lmb:
cc @brycekahle I can't seem to add you as a reviewer.

brycekahle (approved):

nathanjsweet (approved):

PR #161, opened 2020-11-13
cmd/bpf2go: use temporary module for tests

lmb:
Create a temporary module instead of using a temporary subdirectory to
test compilation. This avoids dirtying the repository and avoids
interacting with the 9pfs root mount in the test VM.


florianl (approved):
works also fine for me :+1:

PR #162, opened 2020-11-13
cmd/bpf2go: output ELF .o next to the .go file

lmb:
Go 1.16 will add file embedding via magic comments, which we can use to
store the .o file instead of stuffing it into a global. Move the
compiled object file next to the .go to simplify this in the future.


florianl (approved):
works fine for me :+1: 
Thanks!

PR #165, opened 2020-11-17
btf: add go-fuzz targets

lmb:
Add two targets for go-fuzz, for further experimentation. A short run already found an OOM condition, so I'm sure there is lots more to discover.


tklauser (approved):
Nice!

PR #166, opened 2020-11-18
btf: avoid Type copy in FuncProto.walk

lmb:
FuncProto.walk currently takes a copy of FuncParam.Type since the
for loop implicitly makes a copy of the FuncParam value:

    for _, m := range fp.Params {

m is a copy of the original param, so &m.Type is the address of a
temporary variable. Instead, iterate params using the index and
take the correct address. Also extend the tests to cover this case.


tklauser (approved):
Nice catch

PR #169, opened 2020-11-19
better error messages when RLIMIT_MEMLOCK is too low

lmb:
map: make RLIMIT_MEMLOCK error more specific
    
    The current check for os.ErrPermission will also return true for EACCES,
    which shouldn't trigger the memlock message. It also discards the
    underlying error, which might break feature tests, etc.
    
    Move the check and make it specific to EPERM.

program: try to be helpful when RLIMIT_MEMLOCK is too low
    
    One of the most puzzling errors from BPF is EPERM when trying to load
    a perfectly legitimate program into the kernel. Most of the time this
    means that RLIMIT_MEMLOCK is too low.
    
    Detect this case by checking whether the error from BPF_PROG_LOAD is
    EPERM. Since EPERM may indicate other problems as well, only emit the
    error mentioning rlimit if the verifier buffer is empty.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -163,8 +163,12 @@ func newProgramWithBTF(spec *ProgramSpec, btf *btf.Handle, opts ProgramOptions) 
 		_, logErr = bpfProgLoad(attr)
 	}
 
+	if errors.Is(logErr, unix.EPERM) && logBuf[0] == 0 {
+		return nil, fmt.Errorf("load program: RLIMIT_MEMLOCK may be too low: %w", logErr)
+	}
+
 	err = internal.ErrorWithLog(err, logBuf, logErr)
-	return nil, fmt.Errorf("can't load program: %w", err)
+	return nil, fmt.Errorf("load program: %w", err)
 }
 
 // NewProgramFromFD creates a program from a raw fd.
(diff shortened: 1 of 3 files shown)


lmb:
The error is generated here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/bpf/syscall.c?id=c2e7554e1b85935d962127efa3c2a76483b0b3b6#n1640

tklauser (approved):

prog.go
```diff
@@ -163,8 +163,12 @@ func newProgramWithBTF(spec *ProgramSpec, btf *btf.Handle, opts ProgramOptions)
 		_, logErr = bpfProgLoad(attr)
 	}
 
+	if errors.Is(logErr, unix.EPERM) && logBuf[0] == 0 {
```
ti-mo, line 166: Could we add `only emit the error mentioning rlimit if the verifier buffer is empty` as a comment here for posterity?

ti-mo (approved):
Nice QoL improvement, thanks!

prog.go
lmb: Done!

PR #172, opened 2020-11-20
Simplify feature tests

lmb:
syscalls: simplify nested map feature test
    
    We can detect support for nested maps using a single instead of three
    syscalls.

internal: simplify FeatureTest
    
    FeatureTest requires a function that returns `(bool, error)`. The
    semantics of when to return what is a bit complicated, and difficult
    for contributors to grasp. In essence, we need to represent three
    states:
    
    * we know that the feature is available
    * we know that the feature is not available
    * we don't know either
    
    Switch FeatureTest to a `func() error`. I think this makes it easier
    to write accurate tests.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -718,7 +718,7 @@ func marshalBTF(types interface{}, strings []byte, bo binary.ByteOrder) []byte {
 	return buf.Bytes()
 }
 
-var haveBTF = internal.FeatureTest("BTF", "5.1", func() (bool, error) {
+var haveBTF = internal.FeatureTest("BTF", "5.1", func() error {
 	var (
 		types struct {
 			Integer btfType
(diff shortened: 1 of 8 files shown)


florianl (approved):
Just a minor comment on other OS handling.

Thanks for simplifying the feature tests!

internal/feature.go
```diff
@@ -42,10 +42,10 @@ type featureTest struct {
 //
 // The return values have the following semantics:
 //
+//   err == ErrNotSupported: the feature is not available
```
ti-mo, line 45: Ah, this is the crucial part. :+1: Couldn't deduct from the commit messages which solution you went with eventually.

internal/btf/btf.go
```diff
@@ -718,7 +718,7 @@ func marshalBTF(types interface{}, strings []byte, bo binary.ByteOrder) []byte {
 	return buf.Bytes()
 }
 
-var haveBTF = internal.FeatureTest("BTF", "5.1", func() (bool, error) {
+var haveBTF = internal.FeatureTest("BTF", "5.1", func() error {
```
ti-mo, line 721: Now that these no longer return booleans, the `have___` prefix no longer feels intuitive. Having trouble coming up with an alternative convention, though. Don't immediately remember a similar implementation to draw inspiration from. 

Perhaps `feature___` would work better? `featureBTF`, `featureFuncLinkage`, `featureProgAttach`, etc. That doesn't really convey `nil` as being a 'positive' return value either.

Perhaps `test___`? `testBTF`, `testFuncLinkage`, `testProgAttach`. `if testBTF() == nil { }` looks good, but the term `test` already carries other meaning(s).

lmb: Maybe `check`? I like `feature...` personally.

internal/feature.go
lmb: Ah, my bad, thanks!

internal/unix/types_other.go
```diff
@@ -20,6 +20,7 @@ const (
 	EPERM                    = syscall.EPERM
 	ESRCH                    = syscall.ESRCH
 	ENODEV                   = syscall.ENODEV
+	EBADF                    = 0
```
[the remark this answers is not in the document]
lmb: Ah, my bad, thanks!

internal/btf/btf.go
lmb: I've tried `feature...` and have decided against it for two reasons. First, it doesn't read clearer than the `have` version unfortunately:
```
	if featureObjName() == nil {
...
	}
```

Second, it means renaming a bunch of tests in addition to the variables, at which point it seems like it's not worth it.

lmb: I just re-read your comment and realised that you pointed out exactly this problem, oh well...

lmb:
Merging this for now, I'd still like a better name as @ti-mo pointed out.

Issue #112, opened 2020-06-23
Populate inner maps from BTF .values


takehaya:
Hi. thanks for the great project! 👍

I've been developing using the library since newtools/ebpf. 

I tried to switch to cilium/ebpf and found that the support for innermap was not complete realize.

For example, This function doesn't seem to take into account the fact that it is innermap.
https://github.com/cilium/ebpf/blob/7acf5cc039f43cc55e927f1c4b2fd161535aad26/elf_reader.go#L371

Do you plan to support innermap in the future?
If no one else is planning to do it, I'd like to help with that myself:)

lmb:
Hi takehaya, I'm glad you find it useful :)

We use nested maps in our internal code base, so the library supports it in general. I think what you're running into is that you can't populate Map.InnerMap from within the ELF.

If you'd like to contribute this feature you could start by figuring out how to specify inner maps with libbpf. Since we try to be compatible with that it'd be good to check what their approach is. Once you know that I can point you at the right places in the code if needed.

takehaya:
Hi Imb, thanks for replying!
Thanks for the great advice!

okay, let's look at the libbpf implementation.
I'll write about it when I understand it:)

ti-mo:
There is one example of map-in-map autopopulation in the kernel tree (`map_perf_test_kern.c`):

```c
struct inner_lru {
	__uint(type, BPF_MAP_TYPE_LRU_HASH);
	__type(key, u32);
	__type(value, long);
	__uint(max_entries, MAX_ENTRIES);
	__uint(map_flags, BPF_F_NUMA_NODE);
	__uint(numa_node, 0);
} inner_lru_hash_map SEC(".maps");

struct {
	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
	__uint(max_entries, MAX_NR_CPUS);
	__uint(key_size, sizeof(u32));
	__array(values, struct inner_lru); /* use inner_lru as inner map */
} array_of_lru_hashs SEC(".maps") = {
	/* statically initialize the first element */
	.values = { &inner_lru_hash_map },
};
```

Note that in this case, the type of the inner map (`inner_lru`) and its structure variable (`inner_lru_hash_map`) have named references. I'll look into how this could be written using anonymous structs, I'm not sure if clang supports re-declaring an identical anonymous struct inside the `.values` assignment, but we might need to take this case into account.

Issue #155, opened 2020-11-09
Support .values in BTF map definition


lmb:
Specifying a `values` field allows defining nested maps in BTF. Data may also be initialized.

`m_array_of_maps` specifies both `value` and `values`, that may be a bug?

```C
// tools/testing/selftests/bpf/progs/test_ringbuf_multi.c
struct ringbuf_map {
	__uint(type, BPF_MAP_TYPE_RINGBUF);
	__uint(max_entries, 1 << 12);
} ringbuf1 SEC(".maps"),
  ringbuf2 SEC(".maps");

struct {
	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
	__uint(max_entries, 4);
	__type(key, int);
	__array(values, struct ringbuf_map);
} ringbuf_arr SEC(".maps") = {
	.values = {
		[0] = &ringbuf1,
		[2] = &ringbuf2,
	},
};

// tools/testing/selftests/bpf/progs/map_ptr_kern.c
struct {
	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
	__uint(max_entries, MAX_ENTRIES);
	__type(key, __u32);
	__type(value, __u32);
	__array(values, struct {
		__uint(type, BPF_MAP_TYPE_ARRAY);
		__uint(max_entries, 1);
		__type(key, __u32);
		__type(value, __u32);
	});
} m_array_of_maps SEC(".maps") = {
	.values = { (void *)&inner_map, 0, 0, 0, 0, 0, 0, 0, 0 },
};
```

lmb:
Duplicate of #112

PR #174, opened 2020-11-24
Support for loading inner map from elf

takehaya:
hi. maintainer ✋ 
Sorry for the delay in tackling it :bow:
I would like to give you feedback if you would like

# what
* Read values, key, values in mapSpecFromBTF
* add arrayFromBTF(use __array macro)
* add btfResolveSize(function to extract the size of btf called btfResolveSize)

# related issue
https://github.com/cilium/ebpf/issues/112

# tasks
* [ ] write test code.(I'm not sure how to write... 🤔 
  * The code below can work
  * https://github.com/takehaya/test_ebpf_map
* ~~btf.Enum, btf.Pointer... this implementation of these sizes.~~
  * ~~Working with another PR↑~~
  * In the implementation of this PR matter we found no need, so we'll cancel

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -124,7 +124,6 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 			return nil, fmt.Errorf("load BTF maps: %w", err)
 		}
 	}
-
 	if len(dataSections) > 0 {
 		for idx := range dataSections {
 			if !referencedSections[idx] {
@@ -138,7 +137,6 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 			return nil, fmt.Errorf("load data sections: %w", err)
 		}
 	}
-
 	progs, err := ec.loadPrograms(progSections, relocations, btfSpec)
 	if err != nil {
 		return nil, fmt.Errorf("load programs: %w", err)
@@ -442,14 +440,12 @@ func (ec *elfCode) loadBTFMaps(maps map[string]*MapSpec, mapSections map[elf.Sec
 		if len(syms) == 0 {
 			return fmt.Errorf("section %v: no symbols", sec.Name)
 		}
-
 		for _, sym := range syms {
 			name := sym.Name
 			if maps[name] != nil {
 				return fmt.Errorf("section %v: map %v already exists", sec.Name, sym)
 			}
-
-			mapSpec, err := mapSpecFromBTF(spec, name)
+			mapSpec, err := mapSpecFromBTF(spec, name, maps, false)
 			if err != nil {
 				return fmt.Errorf("map %v: %w", name, err)
 			}
@@ -461,7 +457,7 @@ func (ec *elfCode) loadBTFMaps(maps map[string]*MapSpec, mapSections map[elf.Sec
 	return nil
 }
 
-func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
+func mapSpecFromBTF(spec *btf.Spec, name string, maps map[string]*MapSpec, isInnermap bool) (*MapSpec, error) {
 	btfMap, btfMapMembers, err := spec.Map(name)
 	if err != nil {
 		return nil, fmt.Errorf("can't get BTF: %w", err)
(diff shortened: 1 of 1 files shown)


elf_reader.go
```diff
@@ -479,76 +475,118 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
-			valueSize, err = uintFromBTF(member.Type)
+			valueSize, err := uintFromBTF(member.Type)
 			if err != nil {
 				return nil, fmt.Errorf("can't get BTF value size: %w", err)
 			}
+			mapSpec.ValueSize = valueSize
 
 		case "pinning":
+			if isInnermap {
+				return nil, fmt.Errorf("inner def can't be pinned")
```
lmb, line 532: Good idea, but can you move this check to `newMapWithBTF` in map.go instead? Then it applies to "manually" created MapSpec as well.

elf_reader.go
```diff
@@ -479,76 +475,118 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+			if isInnermap {
+				return nil, fmt.Errorf("multi-level inner maps not supported")
+			}
+
+			innerMapStruct, err := arrayFromBTF(member.Type)
+			if err != nil {
+				return nil, fmt.Errorf("can't get BTF map values: %w", err)
+			}
+
+			if mapSpec.Type != ArrayOfMaps && mapSpec.Type != HashOfMaps {
```
lmb, line 552: This check assumes that "type" comes before "values" in the type definition, which isn't guaranteed. I think we can skip the check or move it to after the loop.

elf_reader.go
```diff
@@ -479,76 +475,118 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+				return nil, fmt.Errorf("should be map-in-map")
+			}
+
+			inner, err := mapSpecFromBTF(spec, string(innerMapStruct.Name), maps, true)
```
lmb, line 556: I think this call doesn't work: the inner map can be defined using an anonymous struct, so there won't be a name you can use to look it up.

Please double check, but I think `innerMapStruct` already contains the map definition. You probably need to split `btf.Spec.Map` into two functions: one that uses `name` to find `mapStruct` and one that turns `mapStruct` into `btf.Map`. I think this function `mapSpecFromBTF` might also have to be split to make things work.

lmb (changes requested):
Hi! No worries at all, after all this isn't a race :)

I left some detailed comments on your changes, please take a look. Can you please also add tests that cover the new functionality? You can add some nested maps to loader.c and make sure they are parsed and loaded correctly.

From `test_btf_map_in_map.c` in the kernel sources it seems like the loader is also expected to fill in the nested map references at load time, like so:

```
= {
...
	.values = { (void *)&inner_map1, 0, (void *)&inner_map2 },
}
```

I think it's OK to not support this for a start, but it would be good if the library could return an error if a user tries to specify `.values` (or any other fields really).

elf_reader.go
```diff
@@ -567,6 +605,84 @@ func uintFromBTF(typ btf.Type) (uint32, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("not a pointer to struct: %v", typ)
+	}
+
+	return btfstruct, nil
+}
+
+const MAX_RESOLVE_DEPTH = 32
+
+func btfResolveSize(typ btf.Type) (uint32, error) {
```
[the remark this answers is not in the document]
takehaya: oh!
I did not notice btf.Sizeof.
Sizeof. I will fix it to use 🙆‍♂️

elf_reader.go
```diff
@@ -124,7 +124,6 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 			return nil, fmt.Errorf("load BTF maps: %w", err)
 		}
 	}
-
```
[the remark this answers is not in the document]
takehaya: fixed

elf_reader.go
takehaya: Let's check again in detail.
But now a simple print appears to be working because the name is displayed.
```
	if isInnermap {
		fmt.Println("check inner ******")
		fmt.Println(name)
	}
```
Can you elaborate on what kind of case it is not working?

elf_reader.go
takehaya: I'll consider it.
However, newMapWithBTF does not have a recursive definition, so I can't think of a reason to have it. (isInnermap is the flag I used to view the map recursively.
Could you please explain it to me?

takehaya (commented):
Thank you for your review!
I asked some questions so please reply :)

elf_reader.go
```diff
@@ -479,76 +475,118 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
 	if err != nil {
 		return nil, fmt.Errorf("can't get size of BTF value: %w", err)
 	}
-	valueSize := uint32(size)
 
-	var (
-		mapType, flags, maxEntries uint32
```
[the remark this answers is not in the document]
lmb: Yes please. I think it's fine to have different styles when writing new code, but changing existing code for the sake of personal preference just makes the diffs harder to read.

(Famous last words, I've probably done this myself as well)

elf_reader.go
lmb: For example (from #155):

```
struct {
	__uint(max_entries, MAX_ENTRIES);
	__type(key, __u32);
	__type(value, __u32);
	__array(values, struct {
		__uint(type, BPF_MAP_TYPE_ARRAY);
		__uint(max_entries, 1);
		__type(key, __u32);
		__type(value, __u32);
	});
}
```

Here, `values` has a type that is an anonymous struct. `innerMapStruct.Name` is therefore `""`.

takehaya: !!!!!
I know! i will fix it 🙆‍♀️

elf_reader.go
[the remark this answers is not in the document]
takehaya: okay. i will roll back:)

elf_reader.go
lmb: `newMapWithBTF` deals with nested maps here: https://github.com/cilium/ebpf/blob/e8edbe10927def099a96acddf0676ddb38b11ccf/map.go#L211-L223

You can add the check for pinning after the `spec.InnerMap == nil` check.

lmb:
@takehaya are you still working on this? Otherwise I'll close this PR for now.

takehaya:
@lmb I'm sorry for the lack of progress...
I'm working on my university thesis, so I can't work on it.
Can you wait a little while longer for me to finish it? 🙇 
(I will be able to work on it in February)

lmb:
Of course, take your time!

ti-mo:
Hi @takehaya, are you going to be working on this in the coming days/weeks? There's a feature I'm working on (#214) that is closely related to what you're doing here.

ti-mo:
@takehaya We're tackling this in https://github.com/cilium/ebpf/pull/218. I've preserved most of your initial commit and built on top of it. Thanks!

PR #176, opened 2020-11-25
btf: add String method to Type

lmb:
Give types a useful printable representation:

    === RUN   TestType/*btf.Void
        types_test.go:121: void#0
    === RUN   TestType/*btf.Int
        types_test.go:121: uint16#0[bits=3]
    === RUN   TestType/*btf.Pointer
        types_test.go:121: pointer#0[target=#0]
    === RUN   TestType/*btf.Array
        types_test.go:121: array#0[type=#0 n=0]
    === RUN   TestType/*btf.Struct
        types_test.go:121: struct#0[""]
    === RUN   TestType/*btf.Union
        types_test.go:121: union#0[""]
    === RUN   TestType/*btf.Enum
        types_test.go:121: enum#0[""]
    === RUN   TestType/*btf.Fwd
        types_test.go:121: fwd#0[struct "thunk"]
    === RUN   TestType/*btf.Typedef
        types_test.go:121: typedef#0["" #0]
    === RUN   TestType/*btf.Volatile
        types_test.go:121: volatile#0[#0]
    === RUN   TestType/*btf.Const
        types_test.go:121: const#0[#0]
    === RUN   TestType/*btf.Restrict
        types_test.go:121: restrict#0[#0]
    === RUN   TestType/*btf.Func
        types_test.go:121: func#0["foo" proto=#0]
    === RUN   TestType/*btf.FuncProto
        types_test.go:121: proto#0["bar"=#0, return=#0]
    === RUN   TestType/*btf.Var
        types_test.go:121: var#0[""]
    === RUN   TestType/*btf.Datasec
        types_test.go:121: section#0[""]

The rough format is

    <type>#<type id>[<other information>].

It is intentionally different from the in-kernel BTF log output:

    [type id] <type> <name> <other>

This is intentional, since we don't know where and how a user
might embed a stringified type. A more compact format seems
better here.

We can always add a dedicated function that prints a []Type in
kernel style if we find that necessary.

### DIFF
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -21,6 +21,8 @@ func (tid TypeID) ID() TypeID {
 type Type interface {
 	ID() TypeID
 
+	String() string
+
 	// Make a copy of the type, without copying Type members.
 	copy() Type
 
@@ -50,6 +52,7 @@ func (n Name) name() string {
 type Void struct{}
 
 func (v *Void) ID() TypeID      { return 0 }
+func (v *Void) String() string  { return "void#0" }
 func (v *Void) size() uint32    { return 0 }
 func (v *Void) copy() Type      { return (*Void)(nil) }
 func (v *Void) walk(*copyStack) {}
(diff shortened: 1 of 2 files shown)


tklauser (approved):
LGTM, agree that we can always add a dedicated function that in kernel style.

PR #177, opened 2020-11-25
CO-RE: add type / field compatibility logic

lmb:
btf: port CO-RE relocation compatibility logic
    
    coreAreTypesCompatible will be used to figure out if a particular target
    (aka kernel) type can be substituted for a local type. coreAreMembersCompatible
    will be used to figure out if we can relocate a load from a local type with
    a load from a target type.
    
    This code is based on Bryce Kahle's CO-RE work and derived from libbpf
    under a BSD 2-Clause license.
    
Updates #114

btf: use a deque instead of a stack
    
    For some of the CO-RE algorithms it's useful to get the first *Type
    from the stack, rather than the last one. Implement a deque based on
    a resizing ring buffer and use it instead of copyStack.

### DIFF
--- a/internal/btf/core.go
+++ b/internal/btf/core.go
@@ -1,0 +1,178 @@
+package btf
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
+// Code in this file is derived from libbpf, which is available under a BSD
+// 2-Clause license.
+
+/* The comment below is from bpf_core_types_are_compat in libbpf.c:
+ *
+ * Check local and target types for compatibility. This check is used for
+ * type-based CO-RE relocations and follow slightly different rules than
+ * field-based relocations. This function assumes that root types were already
+ * checked for name match. Beyond that initial root-level name check, names
+ * are completely ignored. Compatibility rules are as follows:
+ *   - any two STRUCTs/UNIONs/FWDs/ENUMs/INTs are considered compatible, but
+ *     kind should match for local and target types (i.e., STRUCT is not
+ *     compatible with UNION);
+ *   - for ENUMs, the size is ignored;
+ *   - for INT, size and signedness are ignored;
+ *   - for ARRAY, dimensionality is ignored, element types are checked for
+ *     compatibility recursively;
+ *   - CONST/VOLATILE/RESTRICT modifiers are ignored;
+ *   - TYPEDEFs/PTRs are compatible if types they pointing to are compatible;
+ *   - FUNC_PROTOs are compatible if they have compatible signature: same
+ *     number of input args and compatible return and argument types.
+ * These rules are not set in stone and probably will be adjusted as we get
+ * more experience with using BPF CO-RE relocations.
+ */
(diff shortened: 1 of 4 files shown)


internal/btf/core.go
```diff
@@ -0,0 +1,178 @@
...
+
+		default:
+			return false, fmt.Errorf("unsupported type %T", localType)
+		}
+	}
+
+	return false, errors.New("types are nested too deep")
+}
+
+func skipQualifierAndTypedef(typ Type) Type {
```
lmb, line 165: I think this is basically the same as your initial implementation @brycekahle, and it's much nicer this way. Sorry for making a bad call.

internal/btf/core.go
```diff
@@ -0,0 +1,178 @@
...
+
+	for ; l != nil && t != nil; l, t = localTs.shift(), targetTs.shift() {
+		if depth >= maxTypeDepth {
+			return false, errors.New("types are nested too deep")
+		}
+
+		localType = skipQualifierAndTypedef(*l)
+		targetType = skipQualifierAndTypedef(*t)
+
+		if reflect.TypeOf(localType) != reflect.TypeOf(targetType) {
```
brycekahle, line 48: If we are concerned about performance, we may want to preserve the `BTF_KIND_*` int value and compare those here instead. It seems `reflect.TypeOf` is about 10X the perf of an int comparison: https://play.golang.org/p/h-C4VY2-jdN

```
BenchmarkReflectTypeEqual-8     469669340                2.50 ns/op
BenchmarkIntEqual-8             1000000000               0.249 ns/op
```

internal/btf/core_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import "testing"
+
+func TestCoreAreTypesCompatible(t *testing.T) {
+	tests := []struct {
+		a, b       Type
+		compatible bool
+	}{
+		{&Void{}, &Void{}, true},
+		{&Struct{Name: "a"}, &Struct{Name: "b"}, true},
```
brycekahle, line 11: Perhaps add a couple more negative tests? For example, mismatched types (Struct vs Union).

brycekahle (approved):
Some minor comments, but overall looks good!

internal/btf/core.go
lmb: Interesting, seems like this is because comparing `TypeOf` ends up emitting a call to [runtime.ifaceeq](https://golang.org/src/runtime/alg.go?h=ifaceeq#L295). At this point I'm not terribly worried about the overhead to be honest, especially since we can't take a look what the whole CORE process will look like.

internal/btf/core_test.go
lmb: Struct and union is already there (is that what you were looking for?), but I'll add some negative tests for the qualifiers.

I also realised that anonymous fields aren't covered at the moment.

lmb:
Thanks for your review!

PR #180, opened 2020-11-26
elf: refuse initialized BTF maps

lmb:
BTF map definitions work by encoding map type, key size, etc. in a struct.
Most of the time, the declared map is never initialized, even though space
is allocated for it in the ELF. It is possible to intialize the map:

    struct {
        __uint(max_entries, 4);
        __type(key, int);
        __array(values, struct ringbuf_map);
    } ringbuf_arr SEC(".maps") = {
        .values = {
            [0] = &ringbuf1,
            [2] = &ringbuf2,
        },
    };

libbpf uses this for magic nested map initialization. Since we don't
support this at the moment, check that the contents of a BTF map are
zero, and bail out with internal.ErrNotSupported otherwise.


lmb:
@takehaya this is one change I mentioned in your nested map PR, please take a look.

takehaya:
Thank you for mentoring me!
I understand. So that means not supporting named structures
For example, here's what your mentioned
https://github.com/cilium/ebpf/pull/174#discussion_r531073956

lmb:
No, this is a different case again! In this example the declaration uses a named type: `__array(values, struct ringbuf_map);`. The key thing is this:
```
struct { ... } ringbuf_arr SEC(".maps") = {
   // the lines below "initialize" the map contents
    .values = {
        [0] = &ringbuf1,
        [2] = &ringbuf2,
    },
};
```

This PR prohibits such declarations.

Wangtt185 (approved):

takehaya:
Sorry for the slow response 🙏 

Oh. I was mistaken. Thanks for explaining.
I have a 2 of questions. 
1. I don't understand how to initialize it other than this way. How to do the initialization?

2. What is the difference between the following codes?
```
# Code is should be banned
struct { ... } ringbuf_arr SEC(".maps") = {
    .values = {
        [0] = &ringbuf1,
        [2] = &ringbuf2,
    },
};
```
```
# There is also this style of code in the kernel testing samples
struct { ... } ringbuf_arr SEC(".maps") = {
    .values = { (void *)&ringbuf1,(void *)&ringbuf2},
};
```

lmb:
For Q 1: If we don't support initialisation the user has to create inner maps manually in their own code. It's up to you whether you want to implement this functionality or not, from my POV both choices are OK. This PR doesn't mean I don't want initialization to happen, it just gives you a starting point if you do decide to tackle initialization.

For Q 2: have a look at https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html which explains the difference.

PR #183, opened 2020-12-07
elf: refuse references to strings

lmb:
Strings in BPF have to be stack allocated, so trying to reference
a string that is allocated in a dedicated section doesn't work. The
current error message is a bit confusing:

    load: .L.str: unsupported binding: STB_LOCAL

The most common case for this is a call to bpf_trace_printk:

    bpf_trace_printk("foo", sizeof("foo"));

Detect this case and return a better error.


lmb:
@anakryiko FYI another ELF corner case, maybe it's useful to libbpf. LMK if you don't want me to ping you.

florianl (approved):

anakryiko:
We currently get:

```
libbpf: prog 'handle_tp': bad map relo against '.L.str' in section '.rodata.str1.1'
```

But I think libbpf will support it properly very soon.

lmb:
Ah, good to know!

PR #187, opened 2020-12-16
nit: Give Feature Test RWMutex

nathanjsweet:
Refactor feature tests to use a read-write mutex
for the more common read case.

### DIFF
--- a/internal/feature.go
+++ b/internal/feature.go
@@ -32,7 +32,7 @@ func (ufe *UnsupportedFeatureError) Is(target error) bool {
 }
 
 type featureTest struct {
-	sync.Mutex
+	sync.RWMutex
 	successful bool
 	result     error
 }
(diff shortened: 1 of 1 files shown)


internal/feature.go
```diff
@@ -61,13 +67,11 @@ func FeatureTest(name, version string, fn FeatureTestFn) func() error {
 
 	ft := new(featureTest)
 	return func() error {
-		ft.Lock()
-		defer ft.Unlock()
-
-		if ft.successful {
+		if ft.isSuccessful() {
 			return ft.result
```
lmb, line 71: Can you move the locking in here? Like

```
ft.RLock()
if ft.successful {
  defer ft.RUnlock()
  return ft.result
}
ft.RUnlock()
```
etc.? Otherwise we need a comment to explain why it's safe to read ft.result without the lock being taken, which is kind of subtle.

lmb (commented):
Can you tell me more where you see contention on a feature test? Sounds interesting!

nathanjsweet:
> Can you tell me more where you see contention on a feature test? Sounds interesting!

Sure. I'm working on the batch APIs right now, and I wanted to call into the feature test every time they are used. Having those methods contend a write-lock every time they are accessed seems intuitively wrong to me. It *may* never actually be a problem, but in situations where this library is dealing with lots of programs and maps I could see it becoming one.

lmb:
Yeah, it probably doesn't hurt! Thanks for the context.

internal/feature.go
nathanjsweet: done.

internal/feature.go
```diff
@@ -61,13 +67,11 @@ func FeatureTest(name, version string, fn FeatureTestFn) func() error {
...
 		}
-
+		ft.Lock()
+		defer ft.Unlock()
```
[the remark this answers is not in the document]
nathanjsweet: done :)

lmb (approved):

PR #189, opened 2020-12-17
raw tracepoint cleanups

lmb:
link: check program type for raw tracepoints
    
    Return a more useful error message when passing a program with
    the wrong type. This is especially useful here because it's
    easy to mix up tracepoints and raw tracepoints.

link: allow Update to return ErrNotSupported
    
    Raw tracepoints don't allow updating the program, yet they use
    the bpf_link infrastructure. Allow Update to return ErrNotSupported
    for that reason.


buglloc (approved):
LGTM :+1:

PR #190, opened 2020-12-19
example: demonstrate tracepoint usage

florianl:

This PR adds an example on how a third party package can be used to attach an eBPF program to a tracepoint.


example_tracepoint_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/cilium/ebpf/asm"
+	ringbuffer "github.com/cilium/ebpf/perf"
+
+	"github.com/elastic/go-perf"
+)
+
+// This demonstrates how to attach an eBPF program to a tracepoint.
+// The program will be attached to the sys_enter_open syscall and print out the integer
+// 123 everytime the sycall is used.
+func main() {
```
lmb, line 22: Doesn't this have to be something like `func Example_quuxBar`?

example_tracepoint_test.go
```diff
@@ -0,0 +1,110 @@
...
+		// call FnPerfEventOutput
+		asm.FnPerfEventOutput.Call(),
+
+		// set exit code to 0
+		asm.Mov.Imm(asm.R0, 0),
+		asm.Return(),
+	}
+
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Name:         "sys_enter_open",
```
lmb, line 80: Maybe rename, to make it clear this is not related to the tracepoint?
```suggestion
		Name:         "trace_open",
```

example_tracepoint_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		panic(fmt.Errorf("failed to open perf event on tracepoint: %v", err))
+	}
+	defer openTracepoint.Close()
+
+	if err := openTracepoint.SetBPF(uint32(prog.FD())); err != nil {
+		panic(fmt.Errorf("failed to attach eBPF to tracepoint: %v", err))
+	}
+
+	<-ctx.Done()
```
lmb, line 109: Does the program ever print any output? Since the goroutine is locked to the thread, nothing else can run on it and invoke open, right?

example_tracepoint_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"os"
+	"runtime"
+	"time"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	ringbuffer "github.com/cilium/ebpf/perf"
+
+	"github.com/elastic/go-perf"
```
lmb, line 16: I guess this means the package will be included in the go.mod? Probably acceptable.

lmb (commented):
LGTM, should we remove the other example?

example_tracepoint_test.go
florianl: I was testing this program in a different context before :grin: and forgot to rename it.

example_tracepoint_test.go
florianl: yes. It adds `github.com/elastic/go-perf` with the `indirect` tag in `go.mod`. Added this as well now.

example_tracepoint_test.go
florianl: yes - right. It's the very same output/setup as in `example_program_test.go` before.

lmb:
Sorry for the long radio silence! I've discovered one unfortunate side-effect: `go mod vendor` includes `go-perf` now :( I need to figure out whether it's possible to avoid that.

florianl:
Good poind :+1 https://blog.golang.org/examples states that examples are executed as part of `go test`. Unless they are in package `xyz_test`. And as `xyz_test` is independent of package `xyz` it should not influence the modules files of `xyz`.
I will check this again.

florianl:
ok - I checked `go help test` and there it states `'Go test' recompiles each package along with any files with names matching the file pattern "*_test.go".`
So I guess there is only the option to close this PR in order not to introduce the new dependency. What do you think?

lmb:
I think that would be very sad :( Maybe https://github.com/golang/go/issues/36460 will help, do you mind if we keep this open until 1.16 lands and we try again, with an updated go.mod? The alternative would be to set up an examples repository.

florianl:
Keeping the PR open for the moment is fine for me :)

lmb:
@florianl I had an idea: maybe we can use a sub-Go-Module in an examples folder? That would have its own go.mod / go.sum so it wouldn't pollute the main module IIUC.

florianl:
Good idea :+1:
I have updated the PR and moved the example to its own sub module.

lmb:
One of the downsides of the nested module is that we have to manually bump the `cilium/ebpf` version every once in a while, it doesn't automatically follow the rest of the code in the repo. I'm not sure how much of a downside this is. On one hand the API should be fairly stable now, on the other side it's going to introduce some bit-rot. Anyways, I've pushed a commit to only use a single `examples` submodule so that we only have to do this step once if we decide to go forward.

Another downside is that invoking `go build ./...` in the project root now completely ignores the examples. I think we should at least `go build ./examples/...` on CI to make sure we don't commit completely broken code. Not sure what else we can do.

Finally, I think it should be convention that `examples` only ever contains `package main`. This means they are installable via `go get` but other projects can't depend on the examples package in any way. That leaves us a way to get rid of the nested module if it becomes feasible in the future.

WDYT?

lmb:
cc @ti-mo I think you had ideas for examples as well?

ti-mo:
> cc @ti-mo I think you had ideas for examples as well?

Sorry, missed this cc. This submodule thing actually looks like it could be a good approach for including (both userspace and kernelspace) samples in the lib without polluting the main `go.mod`! Haven't played with this kind of structure, but indeed we need to figure out how we're going to prevent these samples from rotting.

examples/tracepoint/main.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := gtp.Configure(ga); err != nil {
+		panic(fmt.Errorf("failed to configure tracepoint: %v", err))
+	}
+
+	// Restore original rlimit values.
+	if err := unix.Setrlimit(unix.RLIMIT_MEMLOCK, &originalLimit); err != nil {
+		panic(fmt.Errorf("failed to restore original rlimit: %v", err))
+	}
+
+	runtime.LockOSThread()
```
ti-mo, line 116: Since this is a sample, documenting why we need to lock an OS thread would be nice. Is this required?

Issue #142, opened 2020-09-28
Should we adjust the ETH_P_ALL definition in example_sock_elf_test.go?


DavadDi:
ETH_P_ALL(0x03) doesn't work in my virtual machine CentOS8, ByteOrder:binary.LittleEndian when I change ETH_P_ALL to 0x300, it works.

example_sock_elf_test.go 
```go
func openRawSock(index int) (int, error) {
	const ETH_P_ALL uint16 = 0x00<<8 | 0x03   // use syscall.ETH_P_ALL instead ？
	sock, err := syscall.Socket(syscall.AF_PACKET, syscall.SOCK_RAW|syscall.SOCK_NONBLOCK|syscall.SOCK_CLOEXEC, int(ETH_P_ALL))
        //...
}
```

Non-functioning `strace` logs：
```
socket(AF_PACKET, SOCK_RAW|SOCK_CLOEXEC|SOCK_NONBLOCK, 3 /*0x03*/) = 3
bind(3, {sa_family=AF_PACKET, sll_protocol=htons(0x300 /* ETH_P_??? */), sll_ifindex=if_nametoindex("lo"), sll_hatype=ARPHRD_NETROM, sll_pkttype=PACKET_HOST, sll_halen=0}, 20) = 0
```

A working `strace` log：
```
socket(AF_PACKET, SOCK_RAW|SOCK_CLOEXEC|SOCK_NONBLOCK, 768 /*0x300*/) = 3
bind(3, {sa_family=AF_PACKET, sll_protocol=htons(ETH_P_ALL), sll_ifindex=if_nametoindex("lo"), sll_hatype=ARPHRD_NETROM, sll_pkttype=PACKET_HOST, sll_halen=0}, 20) = 0
```

And I checked the same example In C `samples/bpf/sock_example.h`,  the code is as follows.
```c
 static inline int open_raw_sock(const char *name)
  {
         sock = socket(PF_PACKET, SOCK_RAW | SOCK_NONBLOCK | SOCK_CLOEXEC, htons(ETH_P_ALL));
         // htons(ETH_P_ALL)  htons can change ETH_P_ALL(0x03) to System BigEndian
         // ...
   }
```

Should we add some helper function to make this example work on both `ByteOrder:binary.LittleEndian` and `ByteOrder:binary.BigEndian`, like the `htons` function?

lmb:
Ah, interesting. Yeah, that would be nice! A `htons` wrapping `binary.BigEndian` shouldn't be too complicated.

PR #191, opened 2021-01-07
Fix little endian conversion of ETH_P_ALL in example

irozzo-1A:
Hi, congratulations on the cool project!

I was taking a look at the examples and (unless I'm missed something :-)), I think I found a bug. The logic for the network byte order conversion of  ETH_P_ALL does not seem to be good in `example_sock_elf_test.go`.

### DIFF
--- a/example_sock_elf_test.go
+++ b/example_sock_elf_test.go
@@ -150,7 +150,7 @@ func openRawSock(index int) (int, error) {
 	}
 	sll := syscall.SockaddrLinklayer{
 		Ifindex:  index,
-		Protocol: htons(ETH_P_ALL),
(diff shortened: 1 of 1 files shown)


lmb:
Thanks for your PR :) Do you think you could make this work on big endian as well? See #142 for context.

irozzo-1A:
> Thanks for your PR :) Do you think you could make this work on big endian as well? See #142 for context.

@lmb Sure, you can assign me to the issue, I'll take care.

irozzo-1A:
@lmb PTAL I think I addressed your comment

lmb (approved):
LGTM, thanks!

PR #192, opened 2021-01-11
Kprobe compile 

spyder-ian:
Work in progress to add the ability to compile kprobe type bpf programs.

Compiling to object code with clang get's hung up on the assembly macros in the kernel headers, but if we emit-llvm and then do a second pass with llc it seems to work.  This seems like a workaround for something in the compiler.   My bpf programs are working though.

### DIFF
--- a/cmd/bpf2go/compile.go
+++ b/cmd/bpf2go/compile.go
@@ -17,13 +17,19 @@ type compileArgs struct {
 	cc     string
 	cFlags []string
 	// Absolute working directory
-	dir string
+	dir    string
 	// Absolute input file name
 	source string
 	// Absolute output file name
-	dest string
+	dest   string
 	// Depfile will be written here if depName is not empty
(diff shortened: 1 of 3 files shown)


lmb:
Hi @spyder-ian, did you try compiling against `vmlinux.h` like I suggested on the slack? You can generate it using:

```
bpftool btf dump file /path/to/vmlinux format c
```

This requires your `vmlinux` to have BTF enabled via `CONFIG_DEBUG_INFO_BTF=y`.

I'd like to avoid supporting two phase compile if possible, since it means we'll have to find a way to specify flags for `llc` as well.

spyder-ian:
Two passes is ugly, I agree.     I will try with vmlinux.h to see if that works, I have not yet.

spyder-ian:
Closing this as I'm able to get it to work with vmlinux.h.  thank you!

PR #193, opened 2021-01-12
go.mod: bump github.com/google/go-cmp to v0.5.4

tklauser:
Changes since v0.5.2:
https://github.com/google/go-cmp/compare/v0.5.2...v0.5.4

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,6 @@ module github.com/cilium/ebpf
 go 1.14
 
 require (
-	github.com/google/go-cmp v0.5.2
+	github.com/google/go-cmp v0.5.4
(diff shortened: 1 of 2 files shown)


lmb (approved):
LGTM. Is there a bug in the current version? A bit of context in the commit would be nice :)

tklauser:
> LGTM. Is there a bug in the current version? A bit of context in the commit would be nice :)

Just a generic bump to keep the dependency up-to-date. No bugs specifically that I'd be aware we would hit. But the [v0.5.4 release notes](https://github.com/google/go-cmp/releases/tag/v0.5.4) mention the following which are certainly nice to have addressed:

> Fix non-determinism in diffing algorithm
> Impose verbosity limit when formatting map keys

PR #194, opened 2021-01-14
program: allow calculating the tag for a ProgramSpec

lmb:
The kernel exposes a tag for each loaded program. This is just the truncated
sha1 hash of the BPF bytecode. We can use this to check whether a loaded
program corresponds to a ProgramSpec, for example to make sure that the
latest version of a program is loaded.

Add a function to asm.Instructions to calculate the tag and a helper to
ProgramSpec to call the function using the native endianness.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -1,12 +1,16 @@
 package asm
 
 import (
+	"crypto/sha1"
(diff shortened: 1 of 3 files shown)


jsitnicki (approved):
LGTM

PR #195, opened 2021-01-18
bpf2go: rework the generated API

lmb:
cmd/bpf2go: rework the generated API
    
    After working with the API generated by bpf2go for a bit, I think
    there are some improvements to be made:
    
    * Fields have "Program" and "Map" prefixes to avoid name collisions,
      which leads to repetitive typing. From what I can tell most ELFs
      don't have name collisions, so we make the API less nice because
      of an uncommon problem.
    * The API forces loading all maps and programs, even if the user is
      only interested in all maps or programs or a subset.
    * There is too much code generated.
    * The generated code is undocumented.
    
    The first two points are addressed by splitting the "container"
    structs into smaller structs that only contain programs or maps.
    Using Go struct embedding we have a lot of flexibility how to
    compose the types. Embedding also means we remove the "Map" and
    "Program" field prefixes, since name collisions are handled by
    the Go compiler. Finally, reduce the generated API surface to
    two functions:
    
    * loadExampleSpec(): this only parses the embedded ELF and then
      gives the user access to a CollectionSpec. This is intended
      for complicated use cases that need custom logic during loading.
    * loadExampleObjects(): this parses the ELF and calls LoadAndAssign
      on the result. This is intended for simple / common use cases
      that we can cover by extending CollectionOptions.
    
    Finally, add a little bit of documentation to the generated functions.

collection: merge Assign and LoadAndAssign
    
    The two function are basically identical now that LoadAndAssign does
    lazy loading. Merge the two functions into one that can assign both
    *ProgramSpec and *Program, etc.

collection: support embedded structs in Assign
    
    Allow using embedded structs when assigning maps and programs from
    a collection. For example:
    
        type maps struct {
            Bar *ebpf.Map `ebpf:"bar"`
        }
    
        var objs struct {
            maps
            Prog *ebpf.Program `ebpf:"prog"`
        }
    
        collection.Assign(&objs)
    
    This is useful for separating out maps and programs, since they tend
    to have different lifetimes: programs are loaded, attached and then
    discarded. Maps stick around, since they are used to communicate with
    the program.


lmb:
@ti-mo @florianl feel free to remove yourselves if you don't have time to review the PR, I just wanted to make sure to include you in case you're using bpf2go somewhere.

lmb:
The PR fails on Go 1.14, it seems like the semantics of `reflect` have changed. Nothing obvious in the release notes though.

lmb:
It's probably this: https://github.com/golang/go/issues/32772#issue-460489872

florianl (approved):
Works for me and the changes are reasonable :+1:

ti-mo:
Currently not using bpf2go, so have too little context, but the synopsis makes sense. :+1:

lmb:
Thanks for taking a look :) I'm currently mulling over whether to remove `SectionRodata` and friends. There isn't a need for them anymore, so removing them for now would give us flexibility in the future..

PR #196, opened 2021-01-20
elf: better error message for non-static global variables

lmb:
It's easy to forget a static qualifier on a global variable:

    volatile int FOO;

Trying to load a program that references FOO fails like so:

    program filter: missing map FOO

This is due to how the library figures out whether you're tring to
do a direct load or are referencing a map. In any case, the error
is pretty confusing. Try to give a hint to our users.


florianl (approved):
works fine for me :+1:

lmb:
CI fails due to the following in [selftests](https://elixir.bootlin.com/linux/v5.4.91/source/tools/testing/selftests/bpf/progs/test_sysctl_loop2.c#L21):

```
const char tcp_mem_name[] = "net/ipv4/tcp_mem/very_very_very_very_long_pointless_string_to_stress_byte_loop";
static __attribute__((noinline)) int is_tcp_mem(struct bpf_sysctl *ctx)
{
```

Since this is the only such case in selftests it might be a bug in libbpf. I'll try and test against the newest kernel selftests and see whether we get more errors. Otherwise I'll send an email to bpf@

lmb:
These declarations seem to have proliferated, so this needs a proper fix :(

PR #198, opened 2021-01-21
Test on 5.10

lmb:
Test on the new stable 5.10 kernel. Cut down on CI times by not
testing the full Go version x Linux version matrix. Instead,
test the latest kernel on the old Go version only.


ti-mo:
Nice, were there specific interactions between older kernels and specific Go versions that needed to be tested?

ti-mo (approved):

lmb:
Not that I can remember, it just seemed nice to be exhaustive. The downside is that adding a new kernel version is doubly expensive.

PR #202, opened 2021-01-25
btf: delay loading and handle EPERM

lmb:
map, program: delay loading BTF as long as possible
    
    We currently try to load BTF into the kernel even when re-using a pinned
    map. Instead, load BTF only when it is really required. This also fixes
    "leaking" btf.Handle when directly creating a map with BTF via
    NewMapWithOptions.

btf: treat EPERM as unsupported during feature detection
    
    Loading an unprivileged socket filter that has been compiled with
    debug info enabled fails since the BTF feature test returns EPERM.
    However, the socket filter would work just fine without BTF enabled.
    Treat EPERM in the BTF feature test to mean not supported.
    
    This is a two edged sword: EPERM can be a transient condition, e.g.
    a process may drop capabilities after executing haveBTF. It's not
    clear to me whether this happens in practice. Since the library
    will return EPERM in that case we'll hopefully get a bug report.

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -13,6 +13,7 @@ import (
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/btf"
 	"github.com/cilium/ebpf/internal/testutils"
 	"github.com/cilium/ebpf/internal/unix"
 )
@@ -829,6 +830,11 @@ func TestMapPinning(t *testing.T) {
 	if err := m1.Put(uint32(0), uint32(42)); err != nil {
 		t.Fatal("Can't write value:", err)
 	}
+
+	// This is a terrible hack: if loading a pinned map tries to load BTF,
+	// it will get a nil *btf.Spec from this *btf.Map. This is turn will make
+	// btf.NewHandle fail.
+	spec.BTF = new(btf.Map)
 
 	m2, err := NewMapWithOptions(spec, MapOptions{PinPath: tmp})
 	if err != nil {
--- a/collection.go
+++ b/collection.go
@@ -269,6 +269,28 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (*Co
 	}, nil
 }
 
+type btfHandleCache map[*btf.Spec]*btf.Handle
+
+func (btfs btfHandleCache) load(spec *btf.Spec) (*btf.Handle, error) {
+	if btfs[spec] != nil {
+		return btfs[spec], nil
+	}
+
+	handle, err := btf.NewHandle(spec)
+	if err != nil {
+		return nil, err
+	}
+
+	btfs[spec] = handle
+	return handle, nil
+}
+
+func (btfs btfHandleCache) close() {
+	for _, handle := range btfs {
+		handle.Close()
+	}
+}
+
 func lazyLoadCollection(coll *CollectionSpec, opts *CollectionOptions) (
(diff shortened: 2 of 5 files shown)


jsitnicki:
Ideally as a user of `NewCollectionWithOptions`, `NewProgramWithOptions`, `NewMapWithOptions` I'd like to be able to decide if BTF failing to load should treated as fatal. Perhaps I can give my program more capabilities, but I just don't that is what is missing or switch to a different kernel build. Otherwise if I rely on BTF being loaded, I need to somehow check after loading programs and creating maps that it has been loaded.

map_test.go
```diff
@@ -830,6 +831,11 @@ func TestMapPinning(t *testing.T) {
 		t.Fatal("Can't write value:", err)
 	}
 
+	// This is a terrible hack: if loading a pinned map tries to load BTF,
+	// it will get a nil *btf.Spec from this *btf.Map. This is turn will make
+	// btf.NewHandle fail.
+	spec.BTF = new(btf.Map)
+
```
jsitnicki, line 838: Can you instead provide `btf.Spec` and check map info after opening the pinned map to confirm that it was not loaded?

lmb: Unfortunately not, let me try to explain with pseudo code. Right now, the library does this:

```
btf := loadBTFForMap()
if mapIsPinned()
    return loadPinnedMap() // this doesn't use btf at all
}
return createMap(btf)
```

Afterwards:

```
if mapIsPinned()
    return loadPinnedMap()
}
btf := loadBTFForMap()
return createMap(btf)
```

lmb:
Thanks for your review Jakub, I've created a new issue to discuss forcing BTF. In the meantime I've improved the error message when creating a map or program fails with unsupported BTF.

jsitnicki:
Okay, so I understood that unprivileged program load like below will no longer fail with these changes, but I will also not know that BTF was not loaded:
```
jkbs@toad ~/workspace/bpf/go-socket-filter $ cat load_filter.go
package main

import (
        "log"
)

//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang filter bpf/filter.bpf.c

func main() {
        specs, err := newFilterSpecs()
        if err != nil {
                log.Fatal(err)
        }

        objs, err := specs.Load(nil)
        if err != nil {
                log.Fatal(err)
        }
        defer objs.Close()
}
jkbs@toad ~/workspace/bpf/go-socket-filter $ go generate
Compiled /home/jkbs/workspace/bpf/go-socket-filter/filter_bpfel.o
Wrote /home/jkbs/workspace/bpf/go-socket-filter/filter_bpfel.go
Compiled /home/jkbs/workspace/bpf/go-socket-filter/filter_bpfeb.o
Wrote /home/jkbs/workspace/bpf/go-socket-filter/filter_bpfeb.go
jkbs@toad ~/workspace/bpf/go-socket-filter $ go build
jkbs@toad ~/workspace/bpf/go-socket-filter $ go run .
2021/01/28 16:24:38 field ProgramProg: detect support for BTF: operation not permitted
exit status 1
```

map_test.go
jsitnicki: Okay, I see, so you are trying to assert that BTF is not loaded by providing an invalid BTF object.

jsitnicki (approved):

lmb:
Yes, correct. However, if there was some other error, you'd now see something like: `field ProgramProg: load program without BTF: some error`.

PR #208, opened 2021-01-31
Get cpu count at runtime from system instead of MaxEntries

hariharan-a:
Similiar to cililum/cilium, getting CPU count from /sys/device/system…/cpu/possible instead of MaxEntries

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -5,8 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
+	"os"
 	"runtime"
+	"strings"
 	"sync"
 
 	"github.com/cilium/ebpf"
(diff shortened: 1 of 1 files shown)


florianl:
For some background, this PR originates from a discussion on Slack https://cilium.slack.com/archives/C4XCTGYEM/p1612019789126100

The suggested change sounds good to me. But I didn't test it yet.

lmb:
Hi @hariharan-a, thank you for contributing!

It's currently possible to have the library figure out the correct number of CPUs already, unfortunately it's poorly documented, see https://github.com/cilium/ebpf/commit/cbb0c6732be6c3f6d4493c35e2a0246d58d6a429 Basically, declare your perf event array with maxelems == 0 and the library does the right thing.

It used to be that the library forced users to set MaxElems to `0` for perf event maps, but then someone complained that their code with CPU == 8 or whatever didn't work out of the box. From my POV hardcoding the MaxElems is always wrong, but if we revert back to enforcing MaxElems to be possible CPUs we're back to square one. Maybe something to revisit, I've created #209.

Finally, it looks like you copied code from `cilium/cilium` pretty much verbatim. The two repos have a different license, MIT vs Apache 2 respectively. This means merging this code would come with an additional burden to carry the original copyrights, or get Martynas to re-license it under MIT. Since we already have the same functionality in place (see internal.PossibleCPU) I don't think we need to go through the trouble of doing so.

PR #211, opened 2021-02-03
map: distinguish between keys and values when marshaling

lmb:
Marshaling and unmarshaling map keys and values is quite complicated:

* Queues allow nil keys
* Some maps allow storing map and program fds
* But looking them up returns map and program IDs, not fds
* We allow passing unsafe.Pointer for "zero copy" lookups

The current implementation isn't careful enough about these
constraints. For example, Map implements BinaryMarshaler which
encodes the map fd as a host endian uint32. This representation
only makes sense when inserting into an ArrayOfMaps or similar.
Since we don't distinguish between keys and values, you can use
a *Map as a key and get the same behaviour, except that the
key ends up being the value of the file descriptor. This doesn't
make a lot of sense.

Add helpers on Map that should be used when marshaling or unmarshaling
keys and values. Re-using the same code everywhere also fixes
an inconsistency with LookupAndDelete, which currently doesn't
allow unsafe.Pointer.

### DIFF
--- a/map.go
+++ b/map.go
@@ -420,46 +420,7 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
 		return err
 	}
 
-	if valueBytes == nil {
-		return nil
-	}
-
-	if m.typ.hasPerCPUValue() {
-		return unmarshalPerCPUValue(valueOut, int(m.valueSize), valueBytes)
-	}
-
-	switch value := valueOut.(type) {
-	case **Map:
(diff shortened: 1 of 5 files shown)


lmb:
@nathanjsweet after this is merged your batch PR could use `Map.unmarshalKey`, etc. to keep the behaviour consistent with regular lookup, etc.

nathanjsweet (approved):

PR #213, opened 2021-02-10
Statistics

florianl:

    BPF can provide kernel internal statistics about how often a program was
    called and its runtime in ns.
    This PR adds the functionallity to enable the collection of these
    statistics and provides a way to fetch them.

### DIFF
--- a/info.go
+++ b/info.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/cilium/ebpf/internal"
 )
@@ -70,6 +71,14 @@ func (mi *MapInfo) ID() (MapID, bool) {
 	return mi.id, mi.id > 0
 }
 
+// programStats holds statistics of a program.
+type programStats struct {
+	// Total accumulated runtime of the program ins ns.
+	runtime time.Duration
+	// Total number of times the program was called.
+	runCount uint64
+}
+
 // ProgramInfo describes a program.
 type ProgramInfo struct {
 	Type ProgramType
@@ -78,6 +87,8 @@ type ProgramInfo struct {
 	Tag string
 	// Name as supplied by user space at load time.
 	Name string
+
+	stats *programStats
 }
 
 func newProgramInfoFromFd(fd *internal.FD) (*ProgramInfo, error) {
@@ -90,12 +101,16 @@ func newProgramInfoFromFd(fd *internal.FD) (*ProgramInfo, error) {
 	}
 
 	return &ProgramInfo{
-		ProgramType(info.prog_type),
-		ProgramID(info.id),
+		Type: ProgramType(info.prog_type),
+		id:   ProgramID(info.id),
 		// tag is available if the kernel supports BPF_PROG_GET_INFO_BY_FD.
-		hex.EncodeToString(info.tag[:]),
+		Tag: hex.EncodeToString(info.tag[:]),
 		// name is available from 4.15.
-		internal.CString(info.name[:]),
+		Name: internal.CString(info.name[:]),
+		stats: &programStats{
+			runtime:  time.Duration(info.run_time_ns),
+			runCount: info.run_cnt,
+		},
 	}, nil
 }
 
@@ -125,6 +140,28 @@ func newProgramInfoFromProc(fd *internal.FD) (*ProgramInfo, error) {
 // The bool return value indicates whether this optional field is available.
 func (pi *ProgramInfo) ID() (ProgramID, bool) {
 	return pi.id, pi.id > 0
+}
+
+// RunCount returns the total number of times the program was called.
+//
+// Can return 0 if the collection of statistics is not enabled. See EnableStats().
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) RunCount() (uint64, bool) {
+	if pi.stats != nil {
+		return pi.stats.runCount, true
+	}
+	return 0, false
+}
+
+// Runtime returns the total accumulated runtime of the program.
+//
+// Can return 0 if the collection of statistics is not enabled. See EnableStats().
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) Runtime() (time.Duration, bool) {
+	if pi.stats != nil {
+		return pi.stats.runtime, true
+	}
+	return time.Duration(0), false
 }
 
 func scanFdInfo(fd *internal.FD, fields map[string]interface{}) error {
@@ -181,4 +218,22 @@ func scanFdInfoReader(r io.Reader, fields map[string]interface{}) error {
 	}
 
 	return nil
+}
+
+// EnableStats starts the measuring of the runtime
+// and run counts of eBPF programs.
+//
+// Collecting statistics can have an impact on the performance.
+//
+// Requires at least 5.8.
+func EnableStats(which uint32) (io.Closer, error) {
+	attr := internal.BPFEnableStatsAttr{
+		StatsType: which,
+	}
+
+	fd, err := internal.BPFEnableStats(&attr)
+	if err != nil {
+		return nil, err
+	}
+	return fd, nil
 }
--- a/internal/syscall.go
+++ b/internal/syscall.go
@@ -91,6 +91,19 @@ func BPFProgDetach(attr *BPFProgDetachAttr) error {
 	return err
 }
 
+type BPFEnableStatsAttr struct {
+	StatsType uint32
+}
+
+func BPFEnableStats(attr *BPFEnableStatsAttr) (*FD, error) {
+	ptr, err := BPF(BPF_ENABLE_STATS, unsafe.Pointer(attr), unsafe.Sizeof(*attr))
+	if err != nil {
+		return nil, fmt.Errorf("enable stats: %w", err)
+	}
+	return NewFD(uint32(ptr)), nil
+
+}
+
 type bpfObjAttr struct {
 	fileName  Pointer
 	fd        uint32
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,5 @@ go 1.14
 
 require (
 	github.com/google/go-cmp v0.5.4
-	golang.org/x/sys v0.0.0-20200124204421-9fbb57f87de9
+	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c
 )
--- a/go.sum
+++ b/go.sum
@@ -2,5 +2,7 @@ github.com/google/go-cmp v0.5.4 h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 golang.org/x/sys v0.0.0-20200124204421-9fbb57f87de9 h1:1/DFK4b7JH8DmkqhUk48onnSfrPzImPoVxuomtbT2nk=
 golang.org/x/sys v0.0.0-20200124204421-9fbb57f87de9/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c h1:VwygUrnw9jn88c4u8GD3rZQbqrP/tgas88tPUbBxQrk=
+golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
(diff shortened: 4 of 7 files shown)


internal/syscall.go
```diff
@@ -91,6 +91,19 @@ func BPFProgDetach(attr *BPFProgDetachAttr) error {
...
+const (
+	BPF_STATS_RUN_TIME = iota
+)
+
+type BPFEnableStatsAttr struct {
+	StatsTyp uint32
+}
+
+func BPFEnableStats(attr *BPFEnableStatsAttr) error {
+	_, err := BPF(BPF_ENABLE_STATS, unsafe.Pointer(attr), unsafe.Sizeof(*attr))
```
nathanjsweet, line 103: I would like to see a feature test for this function. [See objectNames](https://github.com/cilium/ebpf/blob/master/syscalls.go#L399) as an example.

nathanjsweet (changes requested):
Could you flesh out the commit message a bit to explain what this is. I would also like to see an integration test as well.

florianl:
Sorry about being short worded in the commit message.
I've extended the commit message and added the requested feature test.

info.go
```diff
@@ -182,3 +182,19 @@ func scanFdInfoReader(r io.Reader, fields map[string]interface{}) error {
...
+// Requires at least 5.8.
+func EnableStats() error {
+	attr := internal.BPFEnableStatsAttr{
+		StatsTyp: internal.BPF_STATS_RUN_TIME,
+	}
+
+	return internal.BPFEnableStats(&attr)
+}
+
+var canEnableStats = internal.FeatureTest("can enable stats", "5.8", func() error {
```
nathanjsweet, line 198: Sorry. I didn't realize that there was no way to disable runtime stats after they were enabled. I guess this feature test is actually kind of useless. I recommend removing it.

nathanjsweet (changes requested):

info.go
```diff
@@ -182,3 +182,19 @@ func scanFdInfoReader(r io.Reader, fields map[string]interface{}) error {
 
 	return nil
 }
+
+// EnableStats starts the measuring of the runtime (in ns)
+// and run counts.
+//
+// Requires at least 5.8.
+func EnableStats() error {
```
florianl, line 190: At the moment only `BPF_STATS_RUN_TIME` is supported to be set as value in the attribute. Maybe (I don't know) the kernel will provide more `StatsTyp` at some point.
Do we want to change the API then or ask now the user to provide an argument to `EnableStats()`?

internal/syscall.go
```diff
@@ -91,6 +91,19 @@ func BPFProgDetach(attr *BPFProgDetachAttr) error {
 	return err
 }
 
+const (
+	BPF_STATS_RUN_TIME = iota
```
lmb, line 95: Move this to `internal/unix` instead? golang.org/x/sys/unix has the variable

info.go
lmb: I'd go with an extra parameter since the kernel API encourages extension. I wonder whether it should be `(which int)` or `(which StatsType)`.

info.go
```diff
@@ -182,3 +182,19 @@ func scanFdInfoReader(r io.Reader, fields map[string]interface{}) error {
 
 	return nil
 }
+
+// EnableStats starts the measuring of the runtime (in ns)
+// and run counts.
```
lmb, line 187: Document that this can't be turned off?

prog.go
```diff
@@ -613,6 +613,24 @@ func (p *Program) ID() (ProgramID, error) {
...
 }
 
+// ProgStats holds statistics of a program.
+type ProgStats struct {
+	RunTimeNs uint64
+	RunCnt    uint64
+}
+
+// Stats returns measured statistics of a the program.
+func (p *Program) Stats() (ProgStats, error) {
```
lmb, line 623: This should be on ProgramInfo, with a getter similar to `ProgramInfo.ID()`. Also document that this returns zero (?) values without EnableStats.

prog.go
```diff
@@ -613,6 +613,24 @@ func (p *Program) ID() (ProgramID, error) {
 	return ProgramID(info.id), nil
 }
 
+// ProgStats holds statistics of a program.
+type ProgStats struct {
```
lmb, line 617:
```suggestion
type ProgramStats struct {
```

info.go
lmb: Agree. Is it possible to return a wrapped ErrNotSupported based on the return value from BPFEnableStats?

For future reference, you need to return ErrNotSupported from feature tests in some form, otherwise there isn't much point to it.

lmb (changes requested):

prog.go
florianl: I decided against `func (i *ProgramInfo) Stats() (ProgramStats, error) ` as it requires more syscalls to get the same information.
E.g. one has to get the FD from `ProgramInfo` first before `bpfGetProgInfoByFD` can be used. And storing FD in `ProgramInfo` feels not correct. what do you think?

info.go
florianl: I added the extra argument for the function.

info.go
florianl: i have removed the feature test again.

internal/syscall.go
```diff
@@ -91,6 +91,15 @@ func BPFProgDetach(attr *BPFProgDetachAttr) error {
 	return err
 }
 
+type BPFEnableStatsAttr struct {
+	StatsTyp uint32
```
ti-mo, line 95: This can be `StatsType`

nathanjsweet (approved):
I'm happy, but I will defer to you and @lmb working out your differences on where the Stats function should live. I see the merits of both arguments.

info.go
```diff
@@ -182,3 +182,18 @@ func scanFdInfoReader(r io.Reader, fields map[string]interface{}) error {
 
 	return nil
 }
+
+// EnableStats starts the measuring of the runtime (in ns)
```
ti-mo, line 186: This reads a bit vaguely to me. I'm not familiar with this particular feature, so I immediately had the following questions:
- Which programs get their runtime stats collected? Surely not all BPF programs on the machine, that's why this syscall was introduced.
- How can the stats be queried?
- What am I supposed to pass to `which`?

prog.go
```diff
@@ -613,6 +613,26 @@ func (p *Program) ID() (ProgramID, error) {
 	return ProgramID(info.id), nil
 }
 
+// ProgramStats holds statistics of a program.
+//
+// Values can be 0 if collecting of statistics is not enabled.
```
ti-mo, line 618: This semantic should be documented on `(*Program).Stats()` instead.

prog.go
```diff
@@ -613,6 +613,26 @@ func (p *Program) ID() (ProgramID, error) {
...
+type ProgramStats struct {
```
ti-mo, line 619: Since this is an exported struct, IMO these fields should be properly capitalized and expanded as is the convention in Go: `RuntimeNS` and `RunCount`.

prog.go
```diff
@@ -613,6 +613,26 @@ func (p *Program) ID() (ProgramID, error) {
 	return ProgramID(info.id), nil
 }
 
+// ProgramStats holds statistics of a program.
```
ti-mo, line 616: Could you elaborate on the meaning of the values in the struct? e.g. is `RuntimeNS` the total accumulated runtime of the program? Or the latest invocation? Or the average?

ti-mo (changes requested):
Thanks! I would like to see at least a test or an example of how to use this, though.

info.go
florianl: > Which programs get their runtime stats collected? Surely not all BPF programs on the machine, that's why this syscall was introduced.

As far as I can tell, at the moment there is only the option to enable the colleciton of statistics. If enabled statistics for all executing eBPF programs are collected.
I can imagine, this statistic collection will be extendend and maybe more specific at some point in the future. 

> What am I supposed to pass to which?
At the moment, there is only a single valid value for `which` - `BPF_STATS_RUN_TIME`. `which` was introduced because of discussions in https://github.com/cilium/ebpf/pull/213#discussion_r574313592. 
As `BPF_STATS_RUN_TIME` is provided by golang.org/x/sys/unix I did not add an additional wrapper.

lmb:
I just found out that BPF_STATS_ENABLE returns a file descriptor, which disables statistics collection when it's closed!

prog.go
lmb: Take a look at newProgramInfoFromFd, which is called by Program.Info. There you can pull out the necessary info for ProgramStats and store it in an unexported field in ProgramInfo. The getter then allows the user to retrieve the `ProgramStats` while also telling them whether the feature is available in the first place (on old kernels we can't populate the info). All of this is still only a single syscall.

lmb (changes requested):
* Please take a look at my suggestion for ProgramInfo.Stats(): no extra syscall and we avoid cluttering Program.
* EnableStats needs to return an object that wraps an `internal.FD`, not sure what to call that, `StatsHandle`? 
* Please add a test that enables stats, runs a prog, checks the count, disables the stats.

ti-mo:
> I just found out that BPF_STATS_ENABLE returns a file descriptor, which disables statistics collection when it's closed!

I also suspected this from reading the kernel patches, this seemed the whole point of the syscall. Have you been able to play around with this? Does it only affect BPF programs the current process holds a reference to? What about pinned programs?

lmb:
I think it affects all programs, and is reetrant: multiple calls to enable stats can hold an fd at the same time, and collection is only stopped when the last one is closed. pinned programs are no different than regular programs, so those will get stats collected as well.

florianl:
Thanks for your feedback :pray: I didn't realize it was possible to disable the collecting of statistics.

I have updated the PR to use `ProgramInfo.Stats()`, added a wrapper function similar to `[context.CancelFunc](https://pkg.go.dev/context#CancelFunc)` to disable the collection of statistics.
And I added a test that fetches statistics before and after `EnableStats` and after disabling the statistics.

info.go
```diff
@@ -127,6 +130,28 @@ func (pi *ProgramInfo) ID() (ProgramID, bool) {
...
+
+// Stats returns measured statistics of a program.
+//
+// Values can be 0 if collecting of statistics is not enabled.
+func (pi *ProgramInfo) Stats() (ProgramStats, error) {
+	info, err := bpfGetProgInfoByFD(pi.fd)
+	if err != nil {
+		return ProgramStats{}, err
+	}
+	return ProgramStats{
```
lmb, line 149: Why not create this when doing the bpfGetProgInfoByFD here? https://github.com/cilium/ebpf/blob/7a313e22540feeda41df7b82432ab06c9ee50b78/info.go#L84
Instead of adding `*internal.FD` to `ProgramInfo` add `stats *ProgramStats`. Keeping the fd around like that without a `Close()` function is basically a leak.

This function can then be `Stats() (ProgramStats, bool)` with the following body:

```
if pi.stats != nil { return *pi.stats, true }
return ProgramStats{}, false
```

* A single syscall
* It's clear that ProgramStats may be absent (like on older kernels)
* No dangling fds

info.go
```diff
@@ -182,3 +207,24 @@ func scanFdInfoReader(r io.Reader, fields map[string]interface{}) error {
...
+// DisableStatsFunc stops collecting statistics of eBPF programs.
+type DisableStatsFunc func() error
+
+// EnableStats starts the measuring of the runtime (in ns)
+// and run counts of eBPF programs.
+//
+// Collecting statistics can have an impact on the performance.
+//
+// Requires at least 5.8.
+func EnableStats(which uint32) (DisableStatsFunc, error) {
```
lmb, line 220:
```suggestion
func EnableStats(which uint32) (io.Closer, error) {
```

IIRC internal.FD should already implement `Closer`. That way we have one less exported type, and returning an interface is more flexible than a function. If the API sprouts some additional usage of the FD we can extend the returned interface.

lmb (changes requested):
Tests LGTM, thanks! I think there is still some miscommunication around ProgramInfo, I left another comment, hopefully that is clearer. Let me know if it makes sense.

info.go
lmb: Something I don't have a strong opinion on, but that is worth thinking about: is the `ProgramStats` type necessary? We could  have two getters on `ProgramInfo` instead: `ProgramInfo.Runtime() (foo, bool)` and `ProgramInfo.RunCount(foo, bool)`.

info.go
```diff
@@ -127,6 +130,28 @@ func (pi *ProgramInfo) ID() (ProgramID, bool) {
 	return pi.id, pi.id > 0
 }
 
+// ProgramStats holds statistics of a program.
+type ProgramStats struct {
+	// Total accumulated runtime of the program ins ns.
+	RuntimeNS uint64
```
lmb, line 136:
```suggestion
	Runtime time.Duration
```

This is probably nicer?

info.go
florianl: Ok - I think I get your point.

The idea I had in mind was to be able to repeatly call `Stats()` and get the current information directly. While I understand your approach more like get `ProgramInfo` and from this get the statistics. With your approch getters are more usefull, I think.

florianl: I've chosen `ProgramInfo.RunCount()(uint64)` over `ProgramInfo.RunCount()(uint64, bool)`, as we can not decide from the available data, if stats are enabled or not. e.g. `0` is a valid return if the program was never called.
If one can not call `EnableStats()` without error, then the kernel doesn't provide this information. So it is more an indirect check via the enable function.

lmb (approved):

PR #216, opened 2021-02-15
program: treat EINTR during BPG_PROG_TEST_RUN as supported

### LINKED ISSUE #203: BPF_PROG_TEST_RUN feature test doesn't handle EINTR
```
=== RUN   TestCollectionSpecRewriteMaps
    TestCollectionSpecRewriteMaps: collection_test.go:150: can't test program: detect support for BPF_PROG_TEST_RUN: interrupted system call
```

lmb:
BPF_PROG_TEST_RUN returns EINTR if the test is interrupted by a
signal. Therefore we know that BPF_PROG_REST_RUN is supported if
we see EINTR.

Fixes #203

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -417,6 +417,10 @@ var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() e
 		// otherwise misdetect due to insufficient permissions.
 		return internal.ErrNotSupported
 	}
+	if errors.Is(err, unix.EINTR) {
+		// We know that PROG_TEST_RUN is supported if we get EINTR.
+		return nil
+	}
 	return err
 })
 


prog.go
```diff
@@ -417,6 +417,10 @@ var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() e
 		// otherwise misdetect due to insufficient permissions.
 		return internal.ErrNotSupported
 	}
+	if errors.Is(err, unix.EINTR) {
+		// Issue #203: we know that PROG_TEST_RUN is supported if we get EINTR.
```
nathanjsweet, line 421: Nit: are we referencing other GH issues in the code comments? I would prefer that the comment is just "We know that PROG_TEST_RUN is supported if we get EINTR". Unless the precedent has been set already, then I'm fine.

nathanjsweet (changes requested):

prog.go
lmb: Fixed

nathanjsweet (approved):

ti-mo (approved):

PR #218, opened 2021-02-16
Load inner map definitions from ELF

ti-mo:
This supersedes #174, as a bit of refactoring was necessary to fully support the use case. That's also why this was built on top of the initial `.values` work, to make sure `mapSpecFromBTF` was properly designed to be called recursively.

Should be safe against cyclic references as inner maps can only sit 1 level deep.

I still need to play around with this a bit to see what happens after the parsing stage. Do maps need to be named / can we actually support anonymous structs as inner map defs? Or do we already protect against nameless inners later on in the loading stage?

---

- [x] Add some unit tests
- [x] ELF testdata and examples

@lmb


elf_reader.go
```diff
@@ -541,28 +579,67 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
 			}
 
 		case "pinning":
+			if inner {
+				return nil, errors.New("inner maps can't be pinned")
```
lmb, line 583: I think it's better to check this at the loading stage in `newMapWithOptions`.

elf_reader.go
```diff
@@ -541,28 +579,67 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+					return nil, errors.New("outer map needs to be an array or a hash of maps")
+				}
+				if inner {
+					return nil, fmt.Errorf("multi-level inner maps not supported")
+				}
+
+				// Name the inner map after the struct that declared it (can be anonymous).
+				// Pass the BTF spec from the parent object as parent and child need to be
+				// created from the same BTF blob to be valid (on kernels that support BTF).
+				innerMapSpec, err = mapSpecFromBTF(string(t.Name), t.Members, true, spec)
```
lmb, line 617: I'd probably leave this unnamed. `MapSpec.InnerMap` is only ever used to create a temporary map that serves as a template to the kernel, so a name isn't very useful. There is also a subtlety here: is the name of the type really what the user would expect here?

elf_reader.go
```diff
@@ -520,8 +516,48 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+				return nil, fmt.Errorf("can't get size of BTF value: %w", err)
+			}
+
+			valueSize = uint32(size)
+
 		case "key_size":
-			if _, isVoid := keyType.(*btf.Void); !isVoid {
+			// Key needs to be nil and keySize needs to be 0 for key_size to be
+			// considered a valid member.
+			if key != nil || keySize != 0 {
```
lmb, line 560: Doesn't this always bail out if keySize != 0?

lmb: Oh I see, that is kind of confusing. This is to avoid the void type assertion?

elf_reader.go
```diff
@@ -541,28 +579,67 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+
+			switch t := valueType.(type) {
+			case *btf.Struct:
+				// The values member pointing to an array of structs means we're expecting
+				// a map-in-map declaration.
+				if MapType(mapType) != ArrayOfMaps && MapType(mapType) != HashOfMaps {
+					return nil, errors.New("outer map needs to be an array or a hash of maps")
+				}
+				if inner {
+					return nil, fmt.Errorf("multi-level inner maps not supported")
```
lmb, line 611:
```suggestion
					return nil, fmt.Errorf("nested inner maps not supported")
```

internal/btf/btf.go
```diff
@@ -442,47 +442,19 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
 // Map finds the BTF for a map.
 //
 // Returns an error if there is no BTF for the given name.
-func (s *Spec) Map(name string) (*Map, []Member, error) {
+func (s *Spec) Map(name string) (*Map, error) {
```
lmb, line 445: Maybe just return `(*Struct, error)` here? Another option would be to get rid of this function and move it to `loadBTFMaps`. Stuffing the extra values into Map just to retrieve them later seems indeed a bit weird.

lmb (commented):
> Do maps need to be named / can we actually support anonymous structs as inner map defs?

I think we should allow anonymous structs, if I remember correctly the kernel selftests do this. I'm actually a bit fuzzy on what non-anonymous definitions would look like. Do you have an example?

>  Or do we already protect against nameless inners later on in the loading stage?

We don't, but I also don't think we should. Map names are debug info basically, and the map we create from Spec.InnerMap is very short lived: https://github.com/cilium/ebpf/blob/62b22c2d72fa82f6600ec824f05834eb04e132ac/map.go#L210-L214

elf_reader.go
ti-mo: Yup, the type name is absolutely meaningless here. Thought I'd already removed it, thanks!

elf_reader.go
ti-mo: Ah, no, that's because `key, value` are no longer parsed or initialized in `(*Spec) Map()`, so we no longer need to make the distinction between `nil` and `*Void` here. We also cannot assume that e.g. `key` and `key_size` will appear in a specific order in the map definition.

I realize now that it should be impossible for `keySize` to be > 0 if `key` is nil, but I figured it wouldn't hurt to check both, just in case some behaviour is changed later on.

internal/btf/btf.go
ti-mo: Agreed, will give this some thought. This series of functions `(*Spec).{Map,Datasec,Program}()` could maybe use renaming to more accurately describe what they do. (e.g. `FindMap` or `FindMapByName`) That's probably best tackled in another PR.

ti-mo: @lmb I've changed this to `(*Spec) Map()` to return a `Struct` and moved the Var lookup into a new `(*Spec) VarByName()`.

FWIW, this is why I would argue that moving map construction completely into `internal/btf` could make sense. `Map()` now returns a `*Struct`, but i.e. `Program()` returns a 'finished' `*Program`, which is a bit incoherent. Ideally, `Map()` would return a fully-processed `*Map` with inner maps or `ProgramSpec` pointers in the PROG_ARRAY case.

Currently, we're leaking `internal` types into `ebpf`, which is not ideal. Easier said than fixed, but food for thought. :+1:

ti-mo:
@lmb 
> I'm actually a bit fuzzy on what non-anonymous definitions would look like. Do you have an example?

Sure, fairly simple:
```
struct inner_map {
	__uint(type, BPF_MAP_TYPE_HASH);
	__type(key, uint32_t);
	__type(value, int);
	__uint(max_entries, 1);
};

struct {
	__uint(max_entries, 1);
	__array(values, struct inner_map);
} outer_map __section(".maps");
```

`inner_map` can be declared as a type and re-used wherever needed.

elf_reader.go
lmb: I think I'd prefer just the `key` check, it took me a minute or two to understand why `keySize != 0` doesn't trigger in this case.

elf_reader.go
```diff
@@ -541,28 +579,67 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
 			}
 
 			pinType = PinType(pinning)
 
-		case "key", "value":
+		case "values":
+			// The 'values' field in BTF map definitions is used for declaring map
+			// value types that are references to other BPF objects, like other maps
+			// or programs. It is always expected to be an array of pointers.
+			// TODO: Check if 'values' is the last field in the map definition.
```
lmb, line 597: I think it'd be good to do this if it's not too complicated, to stay compatible with libbpf.

internal/btf/btf.go
```diff
@@ -439,50 +439,29 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
...
-		return nil, nil, err
+// This is a convenience wrapper that performs a lookup of a named Var
+// that is expected to contain a Struct. These Vars are implicitly created
+// by declaring a struct with a 'structure variable', as is common in eBPF C.
+// For example, `struct { ... } map_name ...` emits a global variable `map_name`
+// with the type of said struct (which can be of anonymous type).
+//
+// Returns an error if there is no BTF Var with the given name, or if the
+// Var is not a struct.
+func (s *Spec) Map(name string) (*Struct, error) {
```
lmb, line 453: You're right that returning Struct and Var here is a bit weird. Do we add an accessor function for every `Type` in existence? I'd rather not.

Why not kill this whole function + `VarByName` and put the code into `loadBTFMaps`? The way it looks your follow up work on tail calls will make this obsolete anyways.

lmb: P.S. `Spec.Datasec` is also a candidate for a function that is so short it doesn't do anything useful.

internal/btf/btf.go
```diff
@@ -599,8 +589,29 @@ func (h *Handle) FD() int {
 
 // Map is the BTF for a map.
 type Map struct {
+	name       string
```
lmb, line 592: I don't understand why you add these fields to `Map`?

ti-mo: Nice catch, this was a leftover! :+1:

internal/btf/btf.go
ti-mo: Done! Let's revisit if/when we make BTF types public and the style of necessary accessors and convenience methods is more clear.

elf_reader.go
ti-mo: Would there be a benefit to having both the ELF reader and `newMapWithOptions` check for this? Created https://github.com/cilium/ebpf/issues/237 to make sure we follow up.

ti-mo:
Marking as ready to get some feedback on the stuff in testdata. Will add some negative tests today to provide more coverage.

elf_reader.go
```diff
@@ -533,7 +533,21 @@ func (ec *elfCode) loadBTFMaps(maps map[string]*MapSpec) error {
 				return fmt.Errorf("section %v: map %v already exists", sec.Name, sym)
 			}
 
-			mapSpec, err := mapSpecFromBTF(ec.btf, name)
+			// A global Var is created by declaring a struct with a 'structure variable',
```
lmb, line 536: This can happen in a follow up, ultimately I think we should iterate the Datasec directly.

```
$ bpftool btf dump file testdata/loader-clang-11-el.elf
...
[12] VAR 'hash_map' type_id=11, linkage=global-alloc
[20] VAR 'hash_map2' type_id=19, linkage=global-alloc
[22] VAR 'btf_pin' type_id=21, linkage=global-alloc
[52] DATASEC '.maps' size=0 vlen=3
	type_id=12 offset=0 size=40
	type_id=20 offset=0 size=32
	type_id=22 offset=0 size=40
```

So instead of looping over sec.symbols try to `FindType(secName, &btf.Datasec)` and then iterate Datasec.Vars. That gives us access to VarSecinfo.Offset and VarSecinfo.Size which we'll need when trying to match relocations against the maps section.

Aside: kind of weird that `offset=0` in the dump, I wonder what's up with that.

elf_reader.go
```diff
@@ -610,28 +657,70 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+
 			pinning, err := uintFromBTF(member.Type)
 			if err != nil {
 				return nil, fmt.Errorf("can't get pinning: %w", err)
 			}
 
 			pinType = PinType(pinning)
 
-		case "key", "value":
+		case "values":
```
lmb, line 671: Is it valid to have both `values` and `value` or `value_size` specified?

elf_reader.go
```diff
@@ -610,28 +657,70 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+				// Inner maps are always unnamed, so provide an empty name value.
+				// Pass the BTF spec from the parent object, since both parent and
+				// child must be created from the same BTF blob (on kernels that support BTF).
+				innerMapSpec, err = mapSpecFromBTF("", t.Members, true, spec)
+				if err != nil {
+					return nil, fmt.Errorf("can't parse BTF map definition of inner map: %w", err)
+				}
+
+			default:
+				return nil, fmt.Errorf("unsupported value type '%T' in 'values' field", t)
```
lmb, line 704: The stringified representation of t is probably more informative than just the go type?
```suggestion
				return nil, fmt.Errorf("unsupported type %q in 'values' field", t)
```

internal/btf/btf.go
```diff
@@ -609,6 +564,25 @@ func (h *Handle) FD() int {
 type Map struct {
 	spec       *Spec
 	key, value Type
+	members    []Member
```
lmb, line 567: What is `members` used for?

elf_reader.go
```diff
@@ -545,31 +559,22 @@ func (ec *elfCode) loadBTFMaps(maps map[string]*MapSpec) error {
...
-	size, err = btf.Sizeof(valueType)
-	if err != nil {
-		return nil, fmt.Errorf("can't get size of BTF value: %w", err)
-	}
-	valueSize := uint32(size)
+// mapSpecFromBTF produces a MapSpec based mostly on a list of `btf.Member`s.
+// members should contain a list of struct fields extracted from the BTF type
+// used to declare the map. The name and spec arguments will be copied to the
+// resulting MapSpec, and inner must be true on any resursive invocations.
+func mapSpecFromBTF(name string, members []btf.Member, inner bool, spec *btf.Spec) (*MapSpec, error) {
```
lmb, line 566: Nit: maybe a bit clearer?
```suggestion
func mapSpecFromBTF(name string, def *btf.Struct, inner bool, spec *btf.Spec) (*MapSpec, error) {
```

elf_reader.go
```diff
@@ -610,28 +657,70 @@ func mapSpecFromBTF(spec *btf.Spec, name string) (*MapSpec, error) {
...
+				// The values member pointing to an array of structs means we're expecting
+				// a map-in-map declaration.
+				if MapType(mapType) != ArrayOfMaps && MapType(mapType) != HashOfMaps {
+					return nil, errors.New("outer map needs to be an array or a hash of maps")
+				}
+				if inner {
+					return nil, fmt.Errorf("nested inner maps are not supported")
+				}
+
+				// Inner maps are always unnamed, so provide an empty name value.
```
lmb, line 695: Just as an FYI, it's a bit more subtle than that. I think that inner maps which you look up from a nested map can have a name. But the "template" map that we create from MapSpec.InnerMap doesn't benefit from the name since it's ephemeral. If this function took a `*btf.Struct` you could just as well pass `Struct.name()` here. No need to change anything though.

lmb (commented):
Very nice!

elf_reader.go
ti-mo: > I think that inner maps which you look up from a nested map can have a name.
Interesting, do you have an example? Aren't these just plain (outer) map lookups? :thinking:

internal/btf/btf.go
ti-mo: Good catch, this one slipped through.

elf_reader.go
ti-mo: Good idea!

elf_reader.go
ti-mo: `value` is resolved into a `value_size`, so no issue there. libbpf enforces the following: `if (map->def.value_size && map->def.value_size != 4) {`. Will implement this check as well.

elf_reader.go
ti-mo: Yup, that's more useful, thanks!

elf_reader.go
lmb: Here is what happens for map in map:

```
template = create_map(ARRAY, ...)
maps = create_map(ARRAY_OF_MAPS, template, ...)
close(template)

somemap = create_map(ARRAY, "my_name", ...)
insert(maps, 0, somemap)

// later

somemap = lookup(maps, 0)
```

Maps aren't created on the fly, you still have to create them yourself and insert them.

ti-mo: Of course, now I understand.. libbpf names these `<outer_map>.inner`, will do the same here.

elf_reader.go
ti-mo: Turns out I was also missing a `valueSize = 4` assignment as well.

elf_reader.go
ti-mo: Spun this off into https://github.com/cilium/ebpf/issues/239.

ti-mo:
@lmb This is good for another round. Added a small test case in https://github.com/cilium/ebpf/pull/218/commits/60ae9af82e466bafee62d45af7869f6d30ea0370.

lmb (approved):
Very nice! Thank you and @takehaya for tackling this :)

takehaya:
Sorry for the late reaction:bow: (I couldn't work on the exam for univ graduation because it took a long time)
Thank you for doing a great job @ti-mo @lmb :+1:

Issue #214, opened 2021-02-12
Populate program arrays from BTF .values


ti-mo:
This is an issue to track ongoing work on 'statically' (from ELF) populating map contents using BTF map definitions.

This came up because Cilium makes extensive use tail calls that use `PROG_ARRAY` maps, and manually populating them from userspace would be tedious and error-prone. @lmb came up with the following syntax:

```c
struct {
	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
	__uint(key_size, sizeof(uint32_t));
	__uint(value_size, sizeof(uint32_t));
	__uint(max_entries, 2);
	__array(values, int (void *));
} prog_arr __section(".maps") = {
	.values = {
		[0] = &xdp_tail_call_1,
		[1] = &xdp_tail_call_2,
	},
};
```

(Note: clang expects anonymous function declarations to be in the format of `return_type (fn_params, ...)`. The `__array` macro just so happens to implicitly take out a pointer to this declaration, so this doesn't need to be done explicitly.)

There is one other example in the kernel's samples ([`map_perf_test_kern.c`](https://github.com/torvalds/linux/blob/master/samples/bpf/map_perf_test_kern.c)) that uses `.values` on an `ARRAY_OF_MAPS`, which we might be interested in using as well. Perhaps we could support `.values` for all map types over time?

lmb:
See also #174 which attempts to add a related feature for nested maps.

lmb:
@anakryiko do you have an opinion regarding this, provided that we can get clang to emit the right relocations, etc.?

anakryiko:
This is something like what I had in mind back when we were just inventing BTF-defined maps. I just never got around to implementing this, as no one really asked. It should be pretty easy now to add this on libbpf side, though.

ti-mo:
@anakryiko Great, thanks for the reply! Does the syntax in the issue description look like something libbpf would go with? With the snippet above, clang emits:
```
Relocation section '.rel.maps' at offset 0xe38 contains 2 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name
0000000000000038  0000002400000001 R_BPF_64_64            0000000000000000 xdp_tail_call_1
0000000000000040  0000002500000001 R_BPF_64_64            0000000000000010 xdp_tail_call_2
```
This looks like enough information to correlate map index with fn pointer. WDYT?

anakryiko:
[...]

PR #219, opened 2021-02-16
elf: allow non-static constants

### LINKED ISSUE #199: non-static global variables return error "missing map <var name>"
Originally #196, it turns out that kernel selftests now use global variables without a static qualifier:
```
const char tcp_mem_name[] = "net/ipv4/tcp_mem/very_very_very_very_long_pointless_string_to_stress_byte_loop";
// instead of
static const char tcp_mem_name[] = "net/ipv4/tcp_mem/very_very_very_very_long_pointless_string_to_stress_byte_loop";
```

This breaks our ELF relocation logic, since it assumes that a relocation for a GLOBAL OBJECT always references a map.

lmb:
elf: allow non-static constants
    
    The kernel selftests now use constants without a static qualifier,
    like so:
    
        const char tcp_mem_name[] = "net/ipv4/tcp_mem/very_very_...";
    
    Relax the relocation handling to allow non-static symbols as well.
    
    Fixes #199

elf: check symbol section instead of type and binding when relocating
    
    The relocation code currently looks at the type and binding of the
    symbol to decode what kind of relocation to perform. For example:
    
        struct bpf_map_def bar = { ... };   (1)
        static const int FOO = 2;           (2)
    
    The first case generates a GLOBAL relocation to an OBJECT, the second
    one a LOCAL relocation to a SECTION. In the first case we attempt to
    relocate against known maps, in the second one we try to find a data
    section. This logic breaks once we try to allow non-static globals:
    
        const int FOO = 2;                  (3)
    
    This third case also generates a GLOBAL relocation to an OBJECT.
    Effectively we can't tell a non-static global apart from a map load.
    To fix this we need look at the kind of section we are relocating
    against, e.g. a section with map definitions or data sections. This
    requires a bunch of refactoring to plumb this information through
    the ELF reader in the form of the elfSection struct.

elf: add a test for inline assembly relocations
    
    It's possible to emit a relocation using inline assembly which
    is useful when doing hacky constant substitution. Make sure we
    continue to accept these kinds of relocations.

testdata: add clang 10 and 11
    
    Add pre-compiled ELFs for clang 10 and 11. Clang 11 gets too smart
    and starts inlining function calls even though we've marked them
    noinline. Do a bit of work on the return value to prevent this.
    Unfortunately I don't have clang-7 on my machine anymore, so I've
    had to remove that one.


lmb:
@ti-mo: check out elfSection in this PR. I think that will make #214 easier because you can iterate elfSection.relocations.

PR #221, opened 2021-02-18
btf: add minimal CO-RE logic

lmb:
btf: add minimal CO-RE logic
    
    Add the minimal code required to iterate through CO-RE relocations
    and finding possible candidates. The code deliberately only supports
    the very simple local type id relocation to keep the change small.
    Any unsupported relocation will return ErrNotSupported.
    
    Hook up CO-RE parsing to program loading. This means that programs
    which use CO-RE will now fail with ErrNotSupported instead of
    mysteriously failing the verifier. We don't yet support applying
    relocations.
    
    Updates #114

testdata: factor out common Makefile bits
    
    Add a base.mk that includes common rules to build ELFs. Let's hope
    it doesn't get more complicated than it already is.

btf: add coreAccessor
    
    Accessors encode a path through a type in a string form. Encapsulate
    this as an array of integers since this is convenient to index into
    slices.

btf: add coreReloKind

return types indexed by index from inflateRawTypes

btf: include CO-RE relocations during linking
    
    It's not enough to parse CO-RE relocations into a Spec, we also have to
    include them when appending (aka linking) multiple Programs together.

btf: decode CO-RE relocations from .BTF.ext
    
    Introduce a dedicated type instead of re-using extInfo for CO-RE relocations.
    extInfo is convenient for line and function info since we treat both of those
    as opaque data apart from the instruction offset. The same isn't true for
    CO-RE relocations so it's worth to decode into a concrete type directly.

parse CO-RE .BTF.ext headers if present


lmb:
@brycekahle feel free to punt on this review if you don't have time. I figured that you're the best person for this code.

Issue #223, opened 2021-02-19
data section .data..percpu: missing size


mythi:
I'm experimenting with BPF LSM hooks and building my program (hook to `file_ioctl`) using `bpf2go` (clang 11).

With the default `bpf2go` cflags, the kernel refuses to load the program due to `unknown opcode 8d` (I debugged this with `bpftool prog`). [An old bug](https://bugzilla.redhat.com/show_bug.cgi?id=1618958) suggests `-O2` should always be used. 

However, when adding `-cflags "-O2"` to `bpf2go` params, the ELF loads OK with `bpftool prog` but my simple `cilium/ebpf` based Go prog errors with:
`resolve BTF for LSM hook file_ioctl: can't load kernel spec: data section .data..percpu: missing size`.

The error suggests my kernel is faulty but I don't see the error when building without `-O2`.

Any ideas where to look?

Kernel: 5.11, pahole 1.20.

mythi:
> The error suggests my kernel is faulty but I don't see the error when building without -O2.

After thinking this a bit I don't see how these are related. I submitted a PR for add -O2 to default bpf2go cflags and will look into internal/btf parsing test cases using my kernel next.

mythi:
> will look into internal/btf parsing test cases using my kernel next.

```
$ go test -v -run TestParseVmlinux
=== RUN   TestParseVmlinux
    btf_test.go:36: Can't load BTF: data section .data..percpu: missing size
--- FAIL: TestParseVmlinux (0.04s)
FAIL
exit status 1
FAIL	github.com/cilium/ebpf/internal/btf	0.046s
```

lmb:
Which kernel are you testing against? I probably need to update the shipped test vmlinux BTF and see what is going on.

mythi:
@lmb I have 5.11 but also pahole version matters.  AFAIK, pahole v1.18 started percpu values to BTF

mythi:
> I probably need to update the shipped test vmlinux BTF and see what is going on.

I can see if I can fix this. It's a good learning opportunity for BTF internals. I believe [this commit](https://git.kernel.org/pub/scm/devel/pahole/pahole.git/commit/?id=f3d9054ba8ff1df0fc44e507e3a01c0964cabd42) explains what needs to be done.

lmb:
[...]

PR #224, opened 2021-02-22
bpf2go: add -O2 to default cflags

mythi:
Currently, clang generates invalid opcodes when kernel helper functions
are used in BPF programs and they are compiled without optimizations.

This case is not detected built time so it's better to add safer default
cflags, i.e., with optimizations.

### DIFF
--- a/cmd/bpf2go/compile.go
+++ b/cmd/bpf2go/compile.go
@@ -27,7 +27,12 @@ type compileArgs struct {
 }
 
 func compile(args compileArgs) error {
-	cmd := exec.Command(args.cc, args.cFlags...)
+	// Default cflags that can be overriden by args.cFlags
+	overrideFlags := []string {
+		"-O2",
+	}
+
+	cmd := exec.Command(args.cc, append(overrideFlags, args.cFlags...)...)
 	cmd.Stderr = os.Stderr
 
 	inputDir := filepath.Dir(args.source)


cmd/bpf2go/compile.go
```diff
@@ -39,6 +39,7 @@ func compile(args compileArgs) error {
 	cmd.Args = append(cmd.Args,
 		"-c", args.source,
 		"-o", args.dest,
+		"-O2",
```
lmb, line 42: I think by putting it here it's not possible to specify e.g. `-O3` anymor. This later `-O2` will override the earlier flag. Can you put the "-O2" at the start of cmd.Args instead?

lmb (changes requested):

cmd/bpf2go/compile.go
mythi: Right, I'll fix it

mythi: v2 pushed

lmb (changes requested):
Seems like there was a miscommunication :) The current flags should stay as they are: we don't want users to be able to disable debug symbols for example. The new `-O2` you're proposing should be overridable however, and therefore go to the front of the argument list. Maybe you can include a comment to this effect.

mythi:
Sorry for the misunderstanding, I hope I got it right this time :)

lmb (approved):

ti-mo:
Linking to https://github.com/cilium/ebpf/issues/223 for posterity.

PR #227, opened 2021-02-23
Build testdata ELFs using Docker image (Buster)

ti-mo:
@lmb Best reviewed per commit. PTAL!

The `quay.io/cilium/ebpf-builder:0.1.0` image has already been pushed, so a `make docker all` in `testdata/` should work without issues.

Anything else that should be documented?


testdata/docker/Dockerfile.builder
```diff
@@ -0,0 +1,15 @@
...
+    apt-get -y install curl ca-certificates gnupg make
+
+COPY llvm.list /etc/apt/sources.list.d
+RUN curl -s https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+
+RUN apt-get update && \
+    apt-get -y install \
+    clang-7 llvm-7 \
+    clang-9 llvm-9 \
+    clang-11 llvm-11
```
lmb, line 15: `rm -rf /var/lib/apt/lists/*` together with `no-install...` that drops image size by half for me.

testdata/docker/Dockerfile.builder
```diff
@@ -0,0 +1,15 @@
+# This Dockerfile generates a build environment for generating ELFs
```
lmb, line 1: Why is this file `Dockerfiler.builder` and not just `Dockerfile`? This means I can't build just using `docker build .`.

testdata/docker/Makefile
```diff
@@ -0,0 +1,31 @@
...
+	docker build -f "${DIR}/Dockerfile.builder" "${DIR}" \
+		-t ${REPO}:${TAG} \
+		-t ${REPO}:${VERSION}
+
+push: ## Push the Docker image to the repo specified in REPO.
+	docker push	${REPO}:${TAG}
+	docker push ${REPO}:${VERSION}
+
+help: ## Display this help.
+	@grep -E '^[a-zA-Z_-]+:.*## .*$$' $(MAKEFILE_LIST) | \
```
lmb, line 27: I don't understand what this does, is it better than just looking at the Makefile?

testdata/Makefile
```diff
@@ -29,3 +41,42 @@ loader-%-eb.elf: loader.c
...
+GITDIR := $(shell git rev-parse --show-toplevel)
+UIDGID := $(shell stat -c '%u:%g' ${GITDIR})
+
+# Pop off the 'docker' goal so the remaining goals can be passed to
+# the Dockerized make.
+PARAMS := $(filter-out docker,$(MAKECMDGOALS))
+
+DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+REPO := $(shell cat ${DIR}/docker/REPO)
+VERSION := $(shell cat ${DIR}/docker/VERSION)
```
lmb, line 65: Why not `latest`?

testdata/docker/Dockerfile.builder
lmb: How about adding
```
ENTRYPOINT ["make"]
```

as per https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint ? It seems nice that the default invocation of the container invokes make.

testdata/Makefile
```diff
@@ -29,3 +41,42 @@ loader-%-eb.elf: loader.c
...
+
+else
+
+# If the first make goal contains `docker`, only declare Docker-related targets.
+
+.PHONY: docker docker-shell
+
+# Get the absolute path and uid/gid of the repository on disk.
+GITDIR := $(shell git rev-parse --show-toplevel)
+UIDGID := $(shell stat -c '%u:%g' ${GITDIR})
```
lmb, line 57: I think this should just stat `$(CURDIR)` so we can reuse the recipe elsewhere.

testdata/Makefile
```diff
@@ -29,3 +41,42 @@ loader-%-eb.elf: loader.c
...
+# the Dockerized make.
+PARAMS := $(filter-out docker,$(MAKECMDGOALS))
+
+DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+REPO := $(shell cat ${DIR}/docker/REPO)
+VERSION := $(shell cat ${DIR}/docker/VERSION)
+
+docker: ## Wrap the remaining arguments into a Dockerized 'make'.
+	docker run --rm --user "${UIDGID}" \
+		-v "${GITDIR}":/ebpf -w /ebpf/testdata \
```
lmb, line 69: Use `$(CURDIR)` again, something like:
```
-v "${CURDIR}":/work -w /work
```

That way the invocation

lmb (changes requested):
This is really nice, thank you for putting it together! Here are my thoughts:

* If docker is the way to build testdata then the default make target should just invoke docker to do the build. So `all` invokes docker which invokes `all-dockerized` or something. I don't think we should support running arbitrary commands in the docker image so `docker-shell` and the Makefile trickery for the `docker` target can go.
* I prefer the default target to actually build stuff, not give me help output. I think once our makefiles need a help function they are too complicated already.
* For #221 I split `testdata/Makefile` into `base.mk` and `Makefile` so that I can add more ELFs to `internal/btf/testdata`. It'd be great if I could just type `make` in both directories and get a dockerized build out of it. This is what my comments about using `$(CURDIR)` are about, assuming stuff about the path from the root to the repo to testdata break this.
* For image tagging, can we just do the simplest thing possible and always tag a `latest` image?
* Bonus points: it'd be nice if I could do `make IMAGE=$(docker build -q ./docker)` or similar and get a build that doesn't depend on the public image on quay. This is what my `Dockerfile` comments are about.

testdata/Makefile
```diff
@@ -1,21 +1,33 @@
-CLANG ?= clang
+# The development version of clang is distributed as the 'clang' binary,
+# while stable/released versions have a version number attached.
+# Pin the default clang to a stable version.
+CLANG ?= clang-11
```
lmb, line 4: Good idea!

testdata/docker/llvm.list
```diff
@@ -0,0 +1,13 @@
+# Taken from https://apt.llvm.org.
+
+deb http://apt.llvm.org/buster/ llvm-toolchain-buster main
```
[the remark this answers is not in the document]
ti-mo: Oh, by 'default clang package', do you mean the official Debian package repository? I'm not sure if we ever want to use that one. I left this line here to be able to just install nightly clang in an interactive shell. Remember, not everyone's running deb/ubuntu on their machines. :wink:

testdata/docker/Makefile
```diff
@@ -0,0 +1,31 @@
+# Makefile to build and push the `cilium/ebpf` llvm builder Docker image.
+DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
```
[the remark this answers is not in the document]
ti-mo: `${CURDIR}` is the equivalent of `pwd` and is misaligned when running `make -f testdata/Makefile` from the root of the repo.

testdata/docker/Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+REPO := $(shell cat ${DIR}/REPO)
```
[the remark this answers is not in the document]
ti-mo: This is a popular convention and avoids unnecessary diff on Makefiles. How strongly do you feel against it?

testdata/docker/Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+DATE := $(shell date +'%d%m%Y')
+TAG := ${VERSION}-${DATE}
+
+build: ## Build and tag the Docker image in REPO with VERSION.
+	docker build -f "${DIR}/Dockerfile.builder" "${DIR}" \
+		-t ${REPO}:${TAG} \
+		-t ${REPO}:${VERSION}
+
+push: ## Push the Docker image to the repo specified in REPO.
```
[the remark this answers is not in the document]
ti-mo: Good point, will take care of this!

testdata/docker/Dockerfile.builder
```diff
@@ -0,0 +1,15 @@
...
+FROM debian:buster
+
+RUN apt-get update && \
+    apt-get -y install curl ca-certificates gnupg make
+
+COPY llvm.list /etc/apt/sources.list.d
+RUN curl -s https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+
+RUN apt-get update && \
+    apt-get -y install \
```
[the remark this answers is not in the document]
ti-mo: Good catch! This skips:
```
  llvm-11-dev python3 libomp-11-dev llvm-7-dev python libomp-7-dev llvm-9-dev
  libomp-9-dev manpages manpages-dev
```

testdata/docker/Dockerfile.builder
```diff
@@ -0,0 +1,15 @@
+# This Dockerfile generates a build environment for generating ELFs
+# of testdata programs. Run `make build` in this directory to build it.
+FROM debian:buster
+
+RUN apt-get update && \
+    apt-get -y install curl ca-certificates gnupg make
+
+COPY llvm.list /etc/apt/sources.list.d
+RUN curl -s https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```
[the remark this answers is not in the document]
ti-mo: Done!

testdata/docker/Dockerfile.builder
ti-mo: Yup, `no-install-recommends` helps a ton here :+1:

testdata/docker/Dockerfile.builder
ti-mo: Good point re: Dockerfile name. Was originally done for clarity, but the README specifies that it's a builder image, so no longer matters. Done!

`make` entrypoint would be nice, but dropping into a shell then requires `--entrypoint /bin/bash` (and that flag still doesn't have a shorthand :angry:). If you feel strongly about this, we can change it.

testdata/docker/Makefile
ti-mo: In this case, probably not, since it's only build/push and there's a readme now. It's a convenience, nothing more, I include it by default.

testdata/Makefile
ti-mo: Having only `latest` would defeat the point of this change entirely, since it would make ELFs in historical commits non-reproducible. IMO having versioned containers is a hard requirement here. A possible next step could even be running `make` in testdata during CI to make sure the ELFs correspond to the C code.

testdata/Makefile
ti-mo: Could we align on the reusability requirements? Not clear to me wat needs to be done.

testdata/Makefile
ti-mo: Let's keep the conversation on `${CURDIR}` centered around a single thread

ti-mo:
Thanks for the in-depth review!

> If docker is the way to build testdata then the default make target should just invoke docker to do the build. So all invokes docker which invokes all-dockerized or something.

It's not trivial to make `make` wrap itself (cleanly) in another command like we're trying to do here, at least not when one target pulls in others. Providing a `$(DOCKER)` with the same semantics as e.g. `$(CC)` is not possible, since then `docker` would try to execute itself recursively inside the container, which doesn't make sense.

If the simplicity of the Makefile is key, we'll have to provide a `./run-docker.sh` that does the initial Dockerized invocation instead. I found a `docker` prefix target much more intuitive, so went with that instead.

> I don't think we should support running arbitrary commands in the docker image

Disagree, installing debs is not an option for me, so I will need to use a shell in this image to interact with different versions of clang during development. IMO having something like `docker-shell` as a convenience command is not too heavy of a burden.

> I prefer the default target to actually build stuff, not give me help output.

Ack, thought you might say this.

> I think once our makefiles need a help function they are too complicated already.

Disagree, I find it good form and makes the user's (so, us) lives easier. This is probably just personal preference.

> For #221 I split testdata/Makefile into base.mk and Makefile so that I can add more ELFs to internal/btf/testdata. It'd be great if I could just type make in both directories and get a dockerized build out of it. This is what my comments about using $(CURDIR) are about, assuming stuff about the path from the root to the repo to testdata break this.

Cool idea, this can probably be solved. Only necessary for `testdata/Makefile`, correct?

> Bonus points: it'd be nice if I could do make IMAGE=$(docker build -q ./docker) or similar and get a build that doesn't depend on the public image on quay. This is what my Dockerfile comments are about.

This is already possible with a `make build` in `testdata/docker`. It just builds the image locally and tags it with a qualified name so it can be referred to uniformly, whether it comes from Quay or not. It doesn't push the image anywhere. Note that this will cause the ELFs to drift if the version string is different in the output. I would suggest only using the prebuilt image as soon as this is merged.

testdata/docker/Dockerfile.builder
lmb: Nope, let's keep it as it is.

testdata/docker/Makefile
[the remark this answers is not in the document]
ti-mo: Also, this makes it possible to read repo and version info reliably from outside of the Makefile as well. `testdata/Makefile` relies on this for running the correct version of the image.

testdata/docker/llvm.list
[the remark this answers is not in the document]
lmb: Ok, makes sense.

testdata/docker/Makefile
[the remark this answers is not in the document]
lmb: Are Makefile diffs themselves bad? Complicated Makefiles are a bane, so IMO doing the stupid thing that leads to shorter, more straightforward Makefiles is the way to go.

testdata/docker/Makefile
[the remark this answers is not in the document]
aanm: @lmb @ti-mo we can setup automatic builders in quay.io to push the builds automatically for `Dockerfile.builder`. Does that sound reasonable?

testdata/docker/Dockerfile.builder
aanm: @ti-mo I was about to ask the same. Dockerfiles can be named anything as long they have the suffix Dockerfile, so "builder.Dockerfile" would be acceptable :)

testdata/docker/Makefile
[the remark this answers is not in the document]
ti-mo: We've concluded automated build/pushes are not the way to go at the moment.

ti-mo:
@lmb PTAL, I think this is more in the direction we want to go.

Makefile
```diff
@@ -0,0 +1,65 @@
...
+	$* $(CFLAGS) -mbig-endian -c $< -o $@
+
+%-el.elf: %.c
+	$(CLANG) $(CFLAGS) -mlittle-endian -c $< -o $@
+
+%-eb.elf : %.c
+	$(CLANG) $(CFLAGS) -mbig-endian -c $< -o $@
+
+# Usage: make KDIR=/path/to/foo vmlinux-btf.gz
+vmlinux-btf.gz: $(KDIR)/vmlinux
```
lmb, line 64:
```suggestion
internal/btf/testdata/vmlinux-btf.gz: $(KDIR)/vmlinux
```

testdata/README.md
```diff
@@ -0,0 +1,19 @@
+# eBPF ELF testdata
```
lmb, line 1: I guess this is outdated now?

testdata/docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# This Dockerfile generates a build environment for generating ELFs
+# of testdata programs. Run `make build` in this directory to build it.
+FROM debian:buster
+
+RUN apt-get update && \
```
lmb, line 5: Nit: merge with the apt get below for one less layer?

lmb (approved):

testdata/docker/Dockerfile
ti-mo: `gnupg` is required for `apt-key`, so not sure if possible.

testdata/README.md
ti-mo: Oops!

testdata/docker/Dockerfile
ti-mo: Yup, doesn't work. `apt-get update` doesn't complete if the key isn't installed yet. Minimizing for fewer layers is not very useful, by the way. The goal is to introduce as little data to the image per layer as possible, so removing apt cache after an apt-get in the same `RUN` statement is ideal.

ti-mo:
@lmb See 36e84a61beb393afffff91b44f71fe52aa236aa7, added a small fix for `objcopy`'s tempfiles and added an alias for `vmlinux-btf`. Also, `VMLINUX=/path/to/vmlinux` looks a bit clearer.

lmb (approved):

lmb:
Thanks for the nice work :)

PR #230, opened 2021-02-26
btf: handle vmlinux .data..percpu

mythi:
Pahole 1.18 adds vmlinux percpu variables (.data..percpu section)
into BTF with all the vars as BTF_KIND_VAR and the section info
as BTF_KIND_DATASEC. The section info has its size set by pahole.

Because we currently don't read ELF section info for vmlinux and
make that available in fixupDatasec(), the function fails to find
the size of .data..percpu section and errors.

However, fixupDatasec() is only needed for BTF_KIND_DATASEC with zero
size so we can handle the .data..percpu error by simply skipping all
non-zero BTF_KIND_DATASEC.

Also, update vmlinux-btf.bz testdata with .data..percpu BTF.

Closes: #223


lmb (approved):

lmb:
Thanks!

PR #231, opened 2021-02-26
docker: generate VERSION automatically

lmb:
When building ELFs we want to use a snapshot of clang versions at a
certain point in time. Instead of requiring to manually increment
VERSION just use the Unix epoch as the version. It's unlikely that
we will build multiple images on the same second and removes a step
from the process.


testdata/docker/VERSION
```diff
@@ -1 +1 @@
-0.1.0
+1614356656
```
lmb, line 1: I can't actually push, docker gives me `unauthorized: access to the requested resource is not authorized`. Can you take another look?

lmb: Nvm, I had to log in first :see_no_evil:

ti-mo:
Sure, this makes sense. :+1: Can we discuss clang target versions? I don't think we need to test against more than 3 versions currently. My idea was:

- clang7 (oldest version available in the repo)
- clang9 (BTF support landed)
- current (so, 12? I'm confused about what's the latest release, since https://releases.llvm.org/download.html stops at 11 :thinking:)

brycekahle:
clang 11.1.0 is current (for a couple more days 😄 ). The release schedule is on the right-hand side of https://llvm.org/
> 12.0.0:
>  Jan 26: release/12.x branch, 12.0.0-rc1
>  Feb 23: 12.0.0-rc2
>  Mar 2: 12.0.0-final

lmb:
Your suggestions make sense to me timo. bpf-next tends to need the latest and greatest clang so my idea was to include that in the image but not test against it. It bloats the image I guess, so if you'd prefer we can not install that. I'd like to include the repo nonetheless to make ad hoc installing easier.

ti-mo:
> It bloats the image I guess, so if you'd prefer we can not install that.

Yup, might be useful for keeping the weight down.

> I'd like to include the repo nonetheless to make ad hoc installing easier.

Yes! That's also where the `docker-shell` Makefile target comes in handy, since it gives you a root shell inside the container, allowing packages to be installed. The repo definitely needs to stay in `llvm.list`. :+1:

lmb:
Fixed. For some reason the clang-9 elfs changed with the new container image, not sure why.

Makefile
```diff
@@ -32,9 +32,9 @@ TARGETS := \
 # Build all ELF binaries using a Dockerized LLVM toolchain.
 docker-all:
 	docker run --rm --user "${UIDGID}" \
-		-v "${REPODIR}":/ebpf -w /ebpf \
+		-v "${REPODIR}":/ebpf -w /ebpf --env MAKEFLAGS \
 		"${IMAGE}:${VERSION}" \
-		make all
+		$(MAKE) all
```
ti-mo, line 37: Path to `make` on the host != path to `make` in the container, so I would probably leave this as is. Nice addition of `MAKEFLAGS` though, was thinking we should probably pass that down into the container. :+1:

ti-mo (changes requested):
`git diff --text` on your latest commit shows something like this:

```diff
- clang version 9.0.1-+20210228105911+c1a0a213378a-1~exp1~20210228220443.103
+ clang version 9.0.1-+20191211082622+c1a0a213378-1~exp1~20191211193200.102
```

Was this introduced on purpose?

lmb:
> ~Was this introduced on purpose?~ Didn't notice your latest comment. Could you investigate? Looks like it was built with an older version. There are some other diffs later on in the file as well.

I have no idea how that happened. The only thing I did was `make -C testdata/docker build push`. Maybe some caching problem? I've added `--no-cache` to the build target, maybe that will help. We shouldn't cache `apt install clang...` anyways.

lmb:
Well.

On master:
```
root@aaaa5c8be446:/ebpf# clang-9 --version
clang version 9.0.1-+20191211082622+c1a0a213378-1~exp1~20191211193200.102
```

On my branch:
```
root@1b895022f3d7:/ebpf# clang-9 --version
clang version 9.0.1-+20210301082731+c1a0a213378a-1~exp1~20210301193306.107 
```

There has been a new version released since you tagged the image on master, so the diff here is legitimate.

ti-mo:
I called `git diff --text HEAD..HEAD~1` with the refs in the wrong order. The version indeed went from 2019 -> 2021, not the other way around. :see_no_evil: Strange to see such a big jump, though. Anyway, all good, thanks!

ti-mo (approved):

Issue #226, opened 2021-02-23
"go mod download" dirties go.sum


lmb:
This might be a problem with the Go toolchain, but I've not found a relevant bug yet.

Since https://github.com/cilium/ebpf/commit/d124590cd85516492651e6407d47396cfab2c800 I get the following (using go 1.16):
```
$ go mod download
lorenz@antares:~/dev/ebpf (master *)
$ git diff
diff --git a/go.sum b/go.sum
index 7f94146..781a610 100644
--- a/go.sum
+++ b/go.sum
@@ -4,6 +4,7 @@ github.com/google/go-cmp v0.5.4 h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/kr/pretty v0.2.1 h1:Fmg33tUaq4/8ym9TJN1x7sLJnHVwhP33CNkpYV/7rwI=
 github.com/kr/pretty v0.2.1/go.mod h1:ipq/a2n7PKx3OHsz4KJII5eveXtPO4qwEXGdVfWzfnI=
+github.com/kr/pty v1.1.1 h1:VkoXIwSboBpnk99O/KFauAEILuNHv5DVFKZMBN/gUgw=
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 ```
 
 Running `go mod tidy` afterwards removes the entry again. This causes CI builds so (silently, due to another bug) fail.

ti-mo:
This might be related to https://github.com/golang/go/issues/43994, and seems to be happening across multiple Go projects.

lmb:
Good find, thanks! So we can probably just call `go mod tidy` instead of `go mod download`?

ti-mo:
I think `go mod tidy` should always be run right before committing `go.sum` to ensure it's in a minimal state. That's all I know. :sweat_smile:

aditighag:
[...]

PR #232, opened 2021-03-01
Use go mod tidy on CI

lmb:
cmd/bpf2go: regenerate example
    
    We missed regenerating the bpf2go example since the dirty repo check
    on CI wasn't working. Correct this.

CI: move dirty repo check from epilogue
    
    It seems like a failure in the epilogue doesn't fail the build, move
    the dirty repo check to the prologue again.

CI: use go mod tidy to prime cache
    
    According to https://github.com/golang/go/issues/43994 go mod tidy should
    be used to prime the module cache.
    
Updates #226


lmb:
@aditighag FYI I've made a quick PR to "stop the bleeding".

PR #238, opened 2021-03-03
cmd/bpf2go: set -mcpu=v1 by default

lmb:
Clang by default probes the kernel it is running on to determine whether
it can enable advanced features like 32 bit operations, which is kind of
similar to -march=native behaviour. This is a bad idea for ahead of time
compiled code, since there is no guarantee that the kernel we compile on
is older than the kernel we want to run on.

### DIFF
--- a/cmd/bpf2go/compile.go
+++ b/cmd/bpf2go/compile.go
@@ -28,8 +28,14 @@ type compileArgs struct {
 
 func compile(args compileArgs) error {
 	// Default cflags that can be overriden by args.cFlags
-	overrideFlags := []string {
+	overrideFlags := []string{
+		// Code needs to be optimized, otherwise the verifier will often fail
+		// to understand it.
 		"-O2",
+		// Clang defaults to mcpu=probe which checks the kernel that we are
+		// compiling on. This isn't appropriate for ahead of time
+		// compiled code so force the most compatible version.
+		"-mcpu=v1",
 	}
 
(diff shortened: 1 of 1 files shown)


ti-mo:
Nice, what's the drawback? Is there any known documentation on BPF's cpu types in particular?

lmb:
This is the best I could find with a quick search: https://github.com/llvm/llvm-project/blob/62ec4ac90738a5f2d209ed28c822223e58aaaeb7/llvm/lib/Target/BPF/BPFSubtarget.cpp#L43-L58

lmb:
We could also use `-mcpu=generic` which might be redefined at some point in the future to not be v1 anymore. Not sure when that would happen though.

ti-mo:
@lmb 

I'm getting this output from `./run-tests.sh 4.9`:

```
ok      github.com/cilium/ebpf/asm      0.008s  coverage: 5.5% of statements in ./...
=== RUN   TestLoadingSpec
--- PASS: TestLoadingSpec (0.00s)
=== RUN   TestCompile
clang-9: warning: argument unused during compilation: '-mcpu=v1' [-Wunused-command-line-argument]
--- PASS: TestCompile (0.48s)
=== RUN   TestReproducibleCompile
clang-9: warning: argument unused during compilation: '-mcpu=v1' [-Wunused-command-line-argument]
clang-9: warning: argument unused during compilation: '-mcpu=v1' [-Wunused-command-line-argument]
```

Is this expected?

lmb:
Thanks, opened #248 to fix this.

PR #241, opened 2021-03-05
docs: correct filename of contributing doc

twpayne:
Thanks @ti-mo for the spot. 🛁


ti-mo:
It was actually @mmat11 in https://github.com/cilium/ebpf/pull/236.

PR #243, opened 2021-03-05
run-tests.sh: prebuild test binary, single shared qemu dir

ti-mo:
This makes iterating on tests significantly faster on my machine since builds are no longer constrained to the 2vCPU/512MB environment on the VM.

Also removes some of the complexity of the multiple shared dirs and the setup required to get Go to build inside the VM without network access.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -8,30 +8,20 @@ set -o pipefail
 if [[ "${1:-}" = "--in-vm" ]]; then
   shift
 
-  mount -t bpf bpf /sys/fs/bpf
-  export CGO_ENABLED=0
-  export GOFLAGS=-mod=readonly
-  export GOPATH=/run/go-path
-  export GOPROXY=file:///run/go-path/pkg/mod/cache/download
-  export GOSUMDB=off
-  export GOCACHE=/run/go-cache
-
-  if [[ -d "/run/input/bpf" ]]; then
-    export KERNEL_SELFTESTS="/run/input/bpf"
-  fi
-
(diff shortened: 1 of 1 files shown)


run-tests.sh
```diff
@@ -61,31 +50,30 @@ fetch() {
...
-  mkdir "${input}/bpf"
-  tar --strip-components=4 -xjf "${tmp_dir}/${selftests}" -C "${input}/bpf"
+  mkdir "${scratch}/bpf"
+  tar --strip-components=4 -xjf "${tmp_dir}/${selftests}" -C "${scratch}/bpf"
 else
   echo "No selftests found, disabling"
 fi
 
+# Pre-build test binary on host, output into scratch directory shared with VM.
+CGO_ENABLED=0 go test -cover -c -o "${scratch}/ebpf.test"
```
lmb, line 60: Unfortunately this only runs the tests for the main library, not the nested packages like `asm`, etc. I think Go doesn't allow compiling multiple package into one binary unfortunately.

Options:
* Make the Go build cache work better. I tried doing this before without much success :(
* Compile all sub packages, invoke them manually from within the VM. Probably a PITA.

lmb (changes requested):

ti-mo:
Closing this. Building and executing tests for each package is trivial, but getting the cover profiles of each of those merged is another issue (there's https://github.com/wadey/gocovmerge). Will see if there are some other avenues to speed this up. Thanks for the review @lmb!

PR #245, opened 2021-03-08
info_test: BenchmarkStats for benching BPF runtime stats

ti-mo:
Initially written for trying to reproduce a CI flake of TestStats, this might be useful to have anyway.

### DIFF
--- a/info_test.go
+++ b/info_test.go
@@ -1,6 +1,8 @@
 package ebpf
 
 import (
+	"errors"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


florianl (approved):
works fine here :+1:

lmb:
How is this different that the test that you updated in #246? We don't run the benchmarks directly, so this will likely bitrot.

ti-mo:
> this will likely bitrot

Well, benches are built along with the test suite anyway, so not really an issue in practice.

Perhaps we should add benches as a CI step? Runs each bench for about 250-300ms (could even run shorter ones) and whole thing completes in seconds:
```
go test -exec sudo -run=^$ -bench=. -benchtime 250ms ./...
```

One of the existing benches even flakes on my machine:
```
BenchmarkMap/Update-16                     	--- FAIL: BenchmarkMap/Update-16
    map_test.go:1327: update failed: device or resource busy
BenchmarkMap/NextKey-16                    	 4632351	       257.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkMap/Delete-16                     	--- FAIL: BenchmarkMap/Delete-16
    map_test.go:1353: delete failed: device or resource busy
--- FAIL: BenchmarkMap
```
Will look into this :point_up: 

I've extracted the logic under test into an internal function that can be called in a bench, PTAL.

ti-mo:
The failures I posted above of `BenchmarkMap/Update` and `Delete` are likely related to running linux-ck. I've noticed some surprising behaviour running the go-perf test suite before, so it's likely the culprit. Can't reproduce this inside a VM running CFS.

However, I've found this thread (https://lore.kernel.org/bpf/CAEf4BzY0CfZ6mau4fRVmJokKZadjG0Mkx7+-1+UF5dEk613a3w@mail.gmail.com/T/) that does indicate update and delete can return EBUSY under lock contention since commit 20b6cc34ea74. For now, we'll let the caller handle these failures.

PR #247, opened 2021-03-09
Add Unpin for Program

aditighag:
This is a follow-up PR to #204 to add similar support for `Program`.

Signed-off-by: Aditi Ghag aditi@cilium.io

### DIFF
--- a/pinning.go
+++ b/pinning.go
@@ -1,0 +1,42 @@
+package ebpf
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+func pin(currentPath, newPath string, fd *internal.FD) error {
+	if newPath == "" {
+		return errors.New("given pinning path cannot be empty")
+	}
+	if currentPath == "" {
+		return internal.BPFObjPin(newPath, fd)
+	}
+	if currentPath == newPath {
+		return nil
+	}
+	var err error
+	// Object is now moved to the new pinning path.
+	if err = os.Rename(currentPath, newPath); err == nil {
+		return nil
+	}
+	if !os.IsNotExist(err) {
+		return fmt.Errorf("unable to move pinned object to new path %v: %w", newPath, err)
+	}
+	// Internal state not in sync with the file system so let's fix it.
+	return internal.BPFObjPin(newPath, fd)
+}
+
+func unpin(pinnedPath string) error {
+	if pinnedPath == "" {
+		return nil
+	}
+	err := os.Remove(pinnedPath)
+	if err == nil || os.IsNotExist(err) {
+		return nil
+	}
+	return err
+}
--- a/prog.go
+++ b/prog.go
@@ -103,9 +103,10 @@ type Program struct {
 	// otherwise it is empty.
 	VerifierLog string
 
-	fd   *internal.FD
-	name string
-	typ  ProgramType
+	fd         *internal.FD
+	name       string
+	pinnedPath string
+	typ        ProgramType
 }
 
 // NewProgram creates a new Program.
@@ -228,7 +229,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, btfs btfHandl
 
 	fd, err := bpfProgLoad(attr)
 	if err == nil {
-		return &Program{internal.CString(logBuf), fd, spec.Name, spec.Type}, nil
+		return &Program{internal.CString(logBuf), fd, spec.Name, "", spec.Type}, nil
 	}
 
 	logErr := err
@@ -287,7 +288,7 @@ func newProgramFromFD(fd *internal.FD) (*Program, error) {
 		return nil, fmt.Errorf("discover program type: %w", err)
 	}
 
-	return &Program{"", fd, "", info.Type}, nil
+	return &Program{"", fd, "", "", info.Type}, nil
 }
 
 func (p *Program) String() string {
@@ -338,17 +339,40 @@ func (p *Program) Clone() (*Program, error) {
 		return nil, fmt.Errorf("can't clone program: %w", err)
 	}
 
-	return &Program{p.VerifierLog, dup, p.name, p.typ}, nil
+	return &Program{p.VerifierLog, dup, p.name, "", p.typ}, nil
 }
 
-// Pin persists the Program past the lifetime of the process that created it
+// Pin persists the Program on the BPF virtual file system past the lifetime of
+// the process that created it
 //
 // This requires bpffs to be mounted above fileName. See https://docs.cilium.io/en/k8s-doc/admin/#admin-mount-bpffs
 func (p *Program) Pin(fileName string) error {
-	if err := internal.BPFObjPin(fileName, p.fd); err != nil {
-		return fmt.Errorf("can't pin program: %w", err)
+	if err := pin(p.pinnedPath, fileName, p.fd); err != nil {
+		return err
 	}
+	p.pinnedPath = fileName
 	return nil
+}
+
+// Unpin removes the persisted state for the Program from the BPF virtual filesystem.
+//
+// Failed calls to Unpin will not alter the state returned by IsPinned.
+//
+// Unpinning an unpinned Program returns nil.
+func (p *Program) Unpin() error {
+	if err := unpin(p.pinnedPath); err != nil {
+		return err
+	}
+	p.pinnedPath = ""
+	return nil
+}
+
+// IsPinned returns true if the Program has a non-empty pinned path.
+func (p *Program) IsPinned() bool {
+	if p.pinnedPath == "" {
+		return false
+	}
+	return true
 }
 
 // Close unloads the program from the kernel.
@@ -585,7 +609,7 @@ func LoadPinnedProgram(fileName string) (*Program, error) {
 		return nil, fmt.Errorf("info for %s: %w", fileName, err)
 	}
 
-	return &Program{"", fd, filepath.Base(fileName), info.Type}, nil
+	return &Program{"", fd, filepath.Base(fileName), "", info.Type}, nil
 }
 
 // SanitizeName replaces all invalid characters in name with replacement.
--- a/map.go
+++ b/map.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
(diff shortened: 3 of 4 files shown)


pinnable.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+		if err := os.Rename(currentPath, newPath); err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("unable to pin the object at new path %v: %w", newPath, err)
+			}
+		} else {
+			return nil
+		}
+	}
+	return internal.BPFObjPin(newPath, fd)
```
ti-mo, line 26: Executing the pin operation as a fall-through makes it difficult to figure out at first sight in which cases the pin syscall is executed. I would put this right after `if newpath == "" { ... }` at the top:
```go
if currentpath == "" {
  return internal.BPFObjPin(newPath, fd)
}
```

pinnable.go
```diff
@@ -0,0 +1,38 @@
...
+func pin(currentPath, newPath string, fd *internal.FD) error {
+	if newPath == "" {
+		return fmt.Errorf("pinned path cannot be empty")
+	}
+	if currentPath != "" {
+		if currentPath == newPath {
+			return nil
+		}
+		if err := os.Rename(currentPath, newPath); err != nil {
+			if !os.IsNotExist(err) {
```
ti-mo, line 19: I would put this at the root indentation level at the end of the function:

```go
if err := os.Rename(currentPath, newPath); err != nil && os.IsNotExist(err) {
  return internal.BPFObjPin(newPath, fd)
} else {
  return fmt.Errorf("unable to move object pin to new path %v: %w", newPath, err)
}
```

This removes a few layers of indentation and makes this easier to follow. Note: also edited the error message.

pinnable.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+func pin(currentPath, newPath string, fd *internal.FD) error {
+	if newPath == "" {
+		return fmt.Errorf("pinned path cannot be empty")
+	}
+	if currentPath != "" {
```
ti-mo, line 14: I would remove this condition and indentation block in favor of an early return with `currentPath == newPath { return nil }`, see other comments. I try to keep the happy path on the left as it tends to improve readability.

pinnable.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+	}
+	return internal.BPFObjPin(newPath, fd)
+}
+
+func unpin(pinnedPath string) error {
+	if pinnedPath == "" {
+		return nil
+	}
+	err := os.Remove(pinnedPath)
```
ti-mo, line 33: What about:
```go
if err := os.Remove(pinnedPath); err != nil && !os.IsNotExist(err) {
  return err
}
return nil
```

nathanjsweet (approved):

prog.go
```diff
@@ -228,7 +229,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, btfs btfHandl
 
 	fd, err := bpfProgLoad(attr)
 	if err == nil {
-		return &Program{internal.CString(logBuf), fd, spec.Name, spec.Type}, nil
+		return &Program{internal.CString(logBuf), fd, spec.Name, "", spec.Type}, nil
```
ti-mo, line 232: Not a comment on your code, but we should probably switch to field specifiers for initializing structs since there are now adjacent string fields. @lmb

prog.go
```diff
@@ -338,19 +339,39 @@ func (p *Program) Clone() (*Program, error) {
...
+	if err := pin(p.pinnedPath, fileName, p.fd); err != nil {
+		return err
+	}
+	p.pinnedPath = fileName
+	return nil
+}
+
+// Unpin removes the persisted state for the Program.
+//
+// Unpinning an un-pinned Program returns nil.
```
ti-mo, line 358:
```suggestion
// Unpinning a non-pinned Program returns nil.
```

prog.go
```diff
@@ -338,19 +339,39 @@ func (p *Program) Clone() (*Program, error) {
...
-	if err := internal.BPFObjPin(fileName, p.fd); err != nil {
-		return fmt.Errorf("can't pin program: %w", err)
+	if err := pin(p.pinnedPath, fileName, p.fd); err != nil {
+		return err
+	}
+	p.pinnedPath = fileName
+	return nil
+}
+
+// Unpin removes the persisted state for the Program.
```
ti-mo, line 356: Nit: perhaps we could explain the user here that the kernel will remove the program when the host process exits, and that the unpin operation is executed on the path specified in the latest call to `Pin()`. 'persistent state for the Program' sounds a bit vague to me.

prog.go
```diff
@@ -338,19 +339,39 @@ func (p *Program) Clone() (*Program, error) {
...
+//
+// Unpinning an un-pinned Program returns nil.
+func (p *Program) Unpin() error {
+	if err := unpin(p.pinnedPath); err != nil {
+		return err
 	}
+	p.pinnedPath = ""
```
ti-mo, line 363: We should document that unsuccessful calls to `Unpin()` will not modify a program's pinned state as visible by `IsPinned()`. Agree with the implementation here though. :+1: This will allow the caller to retry later if unpinning fails for whatever reason.

prog.go
```diff
@@ -672,3 +693,15 @@ func resolveBTFType(name string, progType ProgramType, attachType AttachType) (b
 		return nil, nil
 	}
 }
+
+func (p *Program) getFD() *internal.FD {
```
ti-mo, line 697: I believe these getters and setters here are now unused.

ti-mo (changes requested):
Thanks! Some nits regarding control flow, but nice overall :rocket:

pinnable.go
lmb: Channeling `go vet`: you can drop the `else` as well.

prog.go
lmb: Leaving them out is nice because it leads to a compile error when extending the struct. I don't feel very strongly about this. Let's do it in a different PR if you think it's possible.

prog.go
lmb: It's really easy to go into the weeds here, I prefer a high level explanation personally. How about `Unpin reverses the effect of Pin.`? Maybe too much meta.

ti-mo: Agreed, `Pin()` can be expanded a bit, so `Unpin()` can be kept short.

pinnable.go
aditighag: @ti-mo I don't think that code block works.

prog.go
aditighag: >perhaps we could explain the user here that the kernel will remove the program when the host process exits, and that the unpin operation is executed on the path specified in the latest call to Pin()

This is not really a behavior defined by the library. I've kept the description as-is.

prog.go
aditighag: Good point. Done.

aditighag:
Thanks for the review. I've addressed most of the review comments.

pinning.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/cilium/ebpf/internal"
+)
+
+func pin(currentPath, newPath string, fd *internal.FD) error {
+	var err error
+	if currentPath == "" {
+		return internal.BPFObjPin(newPath, fd)
+	}
+	if newPath == "" {
```
lmb, line 16: This check should be before `if currentPath == ""`?

lmb (changes requested):

pinnable.go
ti-mo: @aditighag Sorry, why would this not work?

Building on what @lmb suggested, this would be ideal:

```go
func pin( ... ) error {

...

	err := os.Rename(currentPath, newPath)
	if err != nil && os.IsNotExist(err) {
		return internal.BPFObjPin(newPath, fd)
	}
	
	return fmt.Errorf("unable to move object pin to new path %v: %w", newPath, err)
}
```

prog.go
ti-mo: IMO the library aids in educating the user about consequences and common pitfalls. My point is, 'removes the persisted state' is ambiguous and vague. There is no persisted state, as that would imply writing to disk.

aditighag:
Sorry, I missed the most straight forward case while addressing the review comments. :woman-facepalming: Pushing a new commit shortly.

pinnable.go
aditighag: >Sorry, why would this not work?

@ti-mo Did you try the multiple pin calls case?

aditighag: >Sorry, why would this not work?

Unless I'm missing something, your suggested code block doesn't cover the happy path (i.e. when `os.Rename succeeds`). Can you try running the `TestMapPinMultiple` unit test case?

prog.go
aditighag: >There is no persisted state, as that would imply writing to disk.

Umm... Why is that? Even though BPF `pinning` involves `FD` ref counting behind the scenes in the Linux kernel, there is a persistent state on the BPF virtual file system from user space perspective. 

Do you think it'll be helpful if we call out that the persistent state is on the BPF virtual file system?

aditighag:
Some of the suggested approaches didn't really cover all the possible cases. Based on the review feedback, the latest commit tries to avoid nested paths, and keeps the happy path aligned to the left. I've also added a couple of comments so that when the next time we see the code, it doesn't trip us.

pinnable.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+func pin(currentPath, newPath string, fd *internal.FD) error {
+	if newPath == "" {
+		return fmt.Errorf("pinned path cannot be empty")
```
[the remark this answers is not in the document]
aditighag: >'new pinning path cannot be empty'

"new path" sounds a bit weird to a user calling the public API if an object was being pinned for the very first time. I've changed this to "given path..." instead.

pinnable.go
ti-mo: Whoops, lost track of the goal. :sweat_smile: I like the current implementation! :+1:

prog.go
ti-mo: I think it's good like it is now :+1:

ti-mo (approved):
Looks good, thanks!

lmb (approved):

lmb:
Thank you :)

PR #249, opened 2021-03-09
map, program: allow passing flags when loading pinned object

lmb:
BPF_OBJ_GET accepts the BPF_F_WRONLY and BPF_F_RDONLY flags. Especially
BPF_F_RDONLY is useful, since it allows unprivileged read-only access
to BPF objects. Allow passing flags when loading pinned maps and
programs.

### DIFF
--- a/internal/pinning.go
+++ b/internal/pinning.go
@@ -1,23 +1,22 @@
-package ebpf
+package internal
 
 import (
 	"errors"
 	"fmt"
 	"os"
 
-	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/unix"
 )
 
-func pin(currentPath, newPath string, fd *internal.FD) error {
+func Pin(currentPath, newPath string, fd *FD) error {
 	if newPath == "" {
 		return errors.New("given pinning path cannot be empty")
 	}
-	if currentPath == "" {
-		return internal.BPFObjPin(newPath, fd)
-	}
 	if currentPath == newPath {
 		return nil
+	}
+	if currentPath == "" {
+		return BPFObjPin(newPath, fd)
 	}
 	var err error
 	// Renameat2 is used instead of os.Rename to disallow the new path replacing
@@ -30,10 +29,10 @@ func pin(currentPath, newPath string, fd *internal.FD) error {
 		return fmt.Errorf("unable to move pinned object to new path %v: %w", newPath, err)
 	}
 	// Internal state not in sync with the file system so let's fix it.
-	return internal.BPFObjPin(newPath, fd)
+	return BPFObjPin(newPath, fd)
 }
 
-func unpin(pinnedPath string) error {
+func Unpin(pinnedPath string) error {
 	if pinnedPath == "" {
 		return nil
 	}
--- a/link/link.go
+++ b/link/link.go
@@ -22,6 +22,11 @@ type Link interface {
 	// May return an error wrapping ErrNotSupported.
 	Pin(string) error
 
+	// Undo a previous call to Pin.
+	//
+	// May return an error wrapping ErrNotSupported.
+	Unpin() error
+
 	// Close frees resources.
 	//
 	// The link will be broken unless it has been pinned. A link
@@ -58,7 +63,8 @@ type RawLinkInfo struct {
 // You should consider using the higher level interfaces in this
 // package instead.
 type RawLink struct {
-	fd *internal.FD
+	fd         *internal.FD
+	pinnedPath string
 }
 
 // AttachRawLink creates a raw link.
@@ -86,22 +92,21 @@ func AttachRawLink(opts RawLinkOptions) (*RawLink, error) {
 		return nil, fmt.Errorf("can't create link: %s", err)
 	}
 
-	return &RawLink{fd}, nil
+	return &RawLink{fd, ""}, nil
 }
 
 // LoadPinnedRawLink loads a persisted link from a bpffs.
-func LoadPinnedRawLink(fileName string) (*RawLink, error) {
-	return loadPinnedRawLink(fileName, UnspecifiedType)
-}
-
-func loadPinnedRawLink(fileName string, typ Type) (*RawLink, error) {
-	fd, err := internal.BPFObjGet(fileName)
+//
+// Returns an error if the pinned link type doesn't match linkType. Pass
+// UnspecifiedType to disable this behaviour.
+func LoadPinnedRawLink(fileName string, linkType Type, opts *ebpf.LoadPinOptions) (*RawLink, error) {
+	fd, err := internal.BPFObjGet(fileName, opts.Marshal())
 	if err != nil {
-		return nil, fmt.Errorf("load pinned link: %s", err)
+		return nil, fmt.Errorf("load pinned link: %w", err)
 	}
 
-	link := &RawLink{fd}
-	if typ == UnspecifiedType {
+	link := &RawLink{fd, fileName}
+	if linkType == UnspecifiedType {
 		return link, nil
 	}
 
@@ -111,9 +116,9 @@ func loadPinnedRawLink(fileName string, typ Type) (*RawLink, error) {
 		return nil, fmt.Errorf("get pinned link info: %s", err)
 	}
 
-	if info.Type != typ {
+	if info.Type != linkType {
 		link.Close()
-		return nil, fmt.Errorf("link type %v doesn't match %v", info.Type, typ)
+		return nil, fmt.Errorf("link type %v doesn't match %v", info.Type, linkType)
 	}
 
 	return link, nil
@@ -142,13 +147,23 @@ func (l *RawLink) Close() error {
 // Calling Close on a pinned Link will not break the link
 // until the pin is removed.
 func (l *RawLink) Pin(fileName string) error {
-	if err := internal.BPFObjPin(fileName, l.fd); err != nil {
-		return fmt.Errorf("can't pin link: %s", err)
+	if err := internal.Pin(l.pinnedPath, fileName, l.fd); err != nil {
+		return err
 	}
+	l.pinnedPath = fileName
 	return nil
 }
 
-// Update implements Link.
+// Unpin implements the Link interface.
+func (l *RawLink) Unpin() error {
+	if err := internal.Unpin(l.pinnedPath); err != nil {
+		return err
+	}
+	l.pinnedPath = ""
+	return nil
+}
+
+// Update implements the Link interface.
 func (l *RawLink) Update(new *ebpf.Program) error {
 	return l.UpdateArgs(RawLinkUpdateOptions{
 		New: new,
--- a/link/link_test.go
+++ b/link/link_test.go
@@ -3,6 +3,7 @@ package link
 import (
 	"errors"
 	"io/ioutil"
+	"math"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -11,11 +12,11 @@ import (
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal/testutils"
+	"github.com/cilium/ebpf/internal/unix"
 )
 
 func TestRawLink(t *testing.T) {
-	cgroup, prog, cleanup := mustCgroupFixtures(t)
-	defer cleanup()
+	cgroup, prog := mustCgroupFixtures(t)
 
 	link, err := AttachRawLink(RawLinkOptions{
 		Target:  int(cgroup.Fd()),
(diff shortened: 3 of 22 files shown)


internal/pinning.go
```diff
@@ -1,23 +1,21 @@
...
 import (
 	"errors"
 	"fmt"
 	"os"
-
-	"github.com/cilium/ebpf/internal"
 )
 
-func pin(currentPath, newPath string, fd *internal.FD) error {
+func Pin(currentPath, newPath string, fd *FD) error {
```
aditighag, line 9: We don't want to export this API.

lmb: Since this in `internal` it's not reachable outside of the package, take a look at https://golang.org/doc/go1.4#internalpackages

aditighag: Ah! TIL - Thx!

map.go
```diff
@@ -972,9 +973,13 @@ func (m *Map) unmarshalValue(value interface{}, buf []byte) error {
 	return unmarshalBytes(value, buf)
 }
 
-// LoadPinnedMap load a Map from a BPF file.
-func LoadPinnedMap(fileName string) (*Map, error) {
-	fd, err := internal.BPFObjGet(fileName)
+// LoadPinnedMap loads a Map from a BPF file.
+func LoadPinnedMap(fileName string, opts *LoadPinOptions) (*Map, error) {
```
aditighag, line 986: AFAICS, the 2nd argument doesn't need to be a pointer. You can then get rid of the below `nil` check.

map.go
```diff
@@ -181,7 +181,8 @@ func newMapWithOptions(spec *MapSpec, opts MapOptions, btfs btfHandleCache) (*Ma
 			return nil, fmt.Errorf("pin by name: missing Name or PinPath")
 		}
 
-		m, err := LoadPinnedMap(filepath.Join(opts.PinPath, spec.Name))
+		path := filepath.Join(opts.PinPath, spec.Name)
+		m, err := LoadPinnedMap(path, nil)
```
aditighag: See comment @ L977 - pass `LoadPinOptions{}` here.

prog.go
```diff
@@ -594,8 +594,12 @@ func (p *Program) Detach(fd int, typ AttachType, flags AttachFlags) error {
 // LoadPinnedProgram loads a Program from a BPF file.
 //
 // Requires at least Linux 4.11.
-func LoadPinnedProgram(fileName string) (*Program, error) {
-	fd, err := internal.BPFObjGet(fileName)
+func LoadPinnedProgram(fileName string, opts *LoadPinOptions) (*Program, error) {
```
aditighag, line 597: Same as above.

prog_test.go
```diff
@@ -237,6 +237,28 @@ func TestProgramUnpin(t *testing.T) {
 	}
 }
 
+func TestProgramLoadPinnedWithFlags(t *testing.T) {
```
aditighag, line 240: It would be good to add one more test case that checks a legit case. Intercept calls to `BPFObjGet`, and verify that the flags bits are set correctly. One way is to introduce dependency injection by extracting the BPF* syscall wrappers to an interface.

types.go
```diff
@@ -202,6 +202,12 @@ const (
 	PinByName
 )
 
+// LoadPinOptions control how a pinned object is loaded.
+type LoadPinOptions struct {
+	// Flags may be BPF_F_RDONLY or BPF_F_WRONLY.
+	Flags uint32
```
aditighag, line 208: Why does the field need to be part of a struct?

lmb: I doesn't necessarily. It's in a struct because that allows extending the API in the future, without having to break compatibility.

map.go
lmb: The idea is that most callers won't want to set flags. With the current API that looks like:
```
LoadPinnedMap(path, nil)
```
Without a pointer we need this everywhere:
```
LoadPinnedMap(path, ebpf.LoadPinOptions{})
```

That's a lot more cumbersome for the common case.

The current approach of the library is to have two functions: `LoadPinnedMap()` and `LoadPinnedMapWithOptions(string, struct{})`. The upside is no breaking changes, the downside a lot of API bloat. After discussing with @ti-mo I think a single function with "optional" options via a pointer struct is the way forward.

prog_test.go
lmb: Yeah, I wanted to write such a test originally but I thought that it would require quite a bit of reworking the whole test suite. Writing down my rationale made me realise that I was wrong. I'll modify this to use BPF_F_RDONLY and then check that the loaded map can't be modified via Put. Thanks!

Other than that, I think we already know that the flag bits are set correctly:

1. Most tests don't set flags and therefore use 0 implicitly. We know this works.
2. Since the syscall with bogus flags fails we know that having non-zero flags is passed to the kernel.

Adding dependency injection might even weaken the second point, since there could be a bug in the syscall wrapper that we then miss.

lmb: So this worked well for maps, not so well for programs and links. Programs are already kind of immutable, so I don't know how to test the effect of BPF_F_RDONLY. For a link the kernel seems to ignore BPF_F_RDONLY, which seems dodgy to me. I've keep program and link tests as they are right now.

lmb: Flags are ignored for links and programs: https://elixir.bootlin.com/linux/latest/source/kernel/bpf/inode.c#L545

lmb: I sent a mail to bpf@ https://lore.kernel.org/bpf/CACAyw99e288cPoBuxTjt17YfMy8AHT72AmS1W83EexxvWKaP3w@mail.gmail.com/T/#u

aditighag: If a unit test fails, how can you tell if there is a bug in the library or kernel?

map.go
aditighag: I don't think this is a good trade off, where users are now expected to know that they need to pass nil as the 2nd argument. What's the more common case?

prog_test.go
lmb: I don't understand what you're asking, sorry. Can you rephrase it?

aditighag: If the unit test that checks a legit case fails, how do we know if it's because the correct flags weren't set, or there was a bug in the kernel?

lmb: Ah ok, you're saying that right now we don't get separate errors saying "the library is wrong!" and "the kernel is wrong!", which you would find useful. That's true! I think it's not enough of a benefit to start abstracting the syscalls away however. It implies writing tests for the raw syscalls that can tell us if the kernel is wrong, which would pretty much duplicate the tests for the higher level API (in an ideal world). It's better to assume that the error is in the library until proven otherwise.

map.go
lmb: For better or worse we already have APIs like this. Map.Batch* or CollectionSpec.LoadAndAssign already have an optional pointer, while other parts of the library have `WithOptions`, while creating a link takes a non-pointer struct. Right now there is no consistency, which I would like to fix. This API seems like a step in the right direction, but let me mull it over some more.

Can you explain more why you don't like the `nil` argument? Why is documenting that options can be nil not sufficient?

Maybe @ti-mo can weigh in as well, he seems to like the `*struct{}` style.

aditighag: It's not so much about whether the argument is `nil` or not. It's mainly about user experience, where I feel we are forcing them to pass an additional argument that they don't need in the most common case.

ti-mo: The `nil` config struct idea comes from https://github.com/mdlayher/netlink/blob/master/conn.go#L57, which has worked very well to keep the amount of exported functions down in packages with significant API surface. (like the netlink family of packages)

- Single, unambiguous way of getting something done (`FooBar()` vs. `FooBarWithOptions()`, why let the user make a choice?)
- Callers that want default behaviour pass `nil`
- Relatively painless for users to migrate if (when) we break API (`FooBar("x")` -> `FooBar("x", nil)`, `FooBarWithOptions("x", FooBarOptions{"y"})` -> `FooBar("x", &FooBarOptions{"y"})`
- Config struct can be extended without breaking API (:rocket:)
- It keeps godoc tidy and terse

Personally, I've always been used to this style, so `FooBarWithOptions()` looks strange (and more like Java) to me.

prog_test.go
ti-mo: Stubbing/mocking out syscalls sounds like a lot of work for arguably little gain. Bugs in the kernel are usually detected by testing against a battery of kernels (like this lib does), and we all know the kernel "does not break userland, period". :slightly_smiling_face:

When it comes to kernel interfaces, I've always found more value in integration testing rather than unit testing bits in/out, since a stub/mock would have to emulate all of the kernel's corner cases and idiosyncracies, too. I haven't found a noteworthy iteration speed difference, either. Well-written integration test suites can/should complete in seconds.

map.go
aditighag: Ok, thanks for the context @ti-mo. I still have a preference for adding a new API for the some of the reasons that you mentioned. 🙃 @lmb As the author of the PR, I'll leave the final decision to you. Marking this as resolved.

prog_test.go
aditighag: Ok, that's fair. @lmb Coming back to the original comment, what's the final testing plan to test legit cases?

link/link.go
```diff
@@ -142,9 +151,19 @@ func (l *RawLink) Close() error {
...
-	if err := internal.BPFObjPin(fileName, l.fd); err != nil {
-		return fmt.Errorf("can't pin link: %s", err)
+	if err := internal.Pin(l.pinnedPath, fileName, l.fd); err != nil {
+		return err
+	}
+	l.pinnedPath = fileName
+	return nil
+}
+
+// Unpin implements Link.
```
aditighag, line 161: Typo?

link/link_test.go
```diff
@@ -48,13 +49,43 @@ func TestRawLink(t *testing.T) {
...
+	}
+
+	path := filepath.Join(testutils.TempBPFFS(t), "link")
+	err = link.Pin(path)
+	testutils.SkipIfNotSupported(t, err)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// It seems like the kernel ignores BPF_F_RDONLY when updating a link,
```
aditighag, line 78: Should we add this mote to the public API as well?

aditighag (approved):
Barring couple of minor comments, lgtm.

types.go
```diff
@@ -202,6 +202,12 @@ const (
 	PinByName
 )
 
+// LoadPinOptions control how a pinned object is loaded.
+type LoadPinOptions struct {
```
ti-mo, line 206: Just thinking out loud:

Could this become a `LoadOptions` instead? `Flags uint32` is already quite generic and could be re-used across the public API surface to mean different things in different contexts.

Bitfields are difficult to use because they are a leaky abstraction (if you can call them an abstraction at all) and not very expressive or helpful to the user. If `LoadPinOption`'s only purpose is carrying rd/rw flags, what about the following:

```go
type LoadPinOptions struct {
  ReadOnly bool
  WriteOnly bool
}

func (opts LoadPinOptions) marshal() uint32 {
  var out uint32
  if opts.ReadOnly {
    out |= 1 << 0
  }
  if opts.WriteOnly {
    out |= 1 << 1
  }
  // .. etc ..
  return out
}
```
?

link/link.go
ti-mo: Unpin does indeed implement Link, but the comment should probably be made a bit more descriptive.

link/link_test.go
lmb: I hope that https://lore.kernel.org/bpf/20210326160501.46234-1-lmb@cloudflare.com/ gets somewhere, which would kind of make this obsolete. Keeping documentation up with kernel bugs is a lost cause I think.

types.go
lmb: Once we accept `LoadOptions` in more than one place it'll tie our hands as to what we can add to the struct. Basically, if we feed the type into two functions LoadOptions has to be the intersection of their two behaviours.

The booleans would be nicer, agreed. I'll play with it.

lmb: I added your suggestion, which is an improvement overall I think. I kept `LoadPinOptions.Flags` because it's important to provide an escape hatch for users. Without it my "brilliant" test strategy with `math.MaxUint32` also doesn't work.

I've had to resort to a bit of a hack to expose `LoadPinOptions`: the type is defined in `internal`, and users get access to it via `type LoadPinOptions = internal.LoadPinOptions` which is kind of dodgy. This is done to avoid a circular dependency: I want to define `LoadPinOptions` in `ebpf` but don't want to make `LoadPinFlags` public.

internal/pinning.go
```diff
@@ -0,0 +1,68 @@
...
+	// Request a read-only or write-only object. The default is a read-write
+	// object. Only one of the flags may be set.
+	ReadOnly  bool
+	WriteOnly bool
+
+	// Raw flags for the syscall. Other fields of this struct take precedence.
+	Flags uint32
+}
+
+func LoadPinFlags(opts *LoadPinOptions) uint32 {
```
ti-mo, line 55: Why not `(o *LoadPinOptions) Marshal() uint32`?

internal/pinning.go
```diff
@@ -0,0 +1,68 @@
...
+	if opts == nil {
+		return 0
+	}
+
+	flags := opts.Flags
+	if opts.ReadOnly {
```
ti-mo, line 61: Does the kernel accept both bits being flipped? If not, we probably shouldn't marshal an illegal value.

```go
	switch {
	case opts.ReadOnly:
 		flags |= unix.BPF_F_RDONLY
	case opts.WriteOnly:
 		flags |= unix.BPF_F_WRONLY
	}
```

link/link_test.go
```diff
@@ -48,13 +49,43 @@ func TestRawLink(t *testing.T) {
...
+	// so we can't test this.
+	_, err = LoadPinnedRawLink(path, UnspecifiedType, &ebpf.LoadPinOptions{
+		Flags: math.MaxUint32,
```
ti-mo, line 81: Regarding the exporting of `Flags` and the use of maxu32 in this PR in general: is there no other way of provoking an EINVAL? If not, why test it? I assume this is for hitting negative coverage, but not a good enough reason to warrant adding a `Flags` to `LoadPinOptions`.

internal/pinning.go
lmb: It'll reject it with EINVAL. Setting both should be an error, but IMO the kernel error is good enough for now.

link/link_test.go
lmb: There isn't another way to get that from the kernel. Originally I wanted to test that setting read-only really gives a read only map / link, but it turns out those two just ignore the R/O flag at the moment.

`Flags` is there because I think it's important that it's present. It saves us from having to implement every last bit of the BPF UAPI while allowing users flexibility.

internal/pinning.go
lmb: Then it's exported, but you're probably right.

PR #250, opened 2021-03-09
link/netns_test: don't panic on insufficient test permissions

ti-mo:
Move `panic(err)` into `ExampleAttachNetNs()` so the test suite doesn't panic by default when run without `-exec sudo`.


lmb (approved):

lmb:
Thanks! I hit this myself just earlier.

PR #253, opened 2021-03-10
ProgramSpec: add RewriteMaps()

florianl:
Sometimes it is useful to use existing Maps for a given ProgramSpec
instead of assigning new maps or rewrite all maps for all programs in a
CollectionSpec. For example, multiple programs might share the same map
but are loaded at different times.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -88,6 +88,25 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
 	return &cpy
 }
 
+// RewriteMaps replaces all references to specific maps.
+//
+// Use this function to use pre-existing maps before loading the
+// program.
+func (ps *ProgramSpec) RewriteMaps(maps map[string]*Map) error {
+	for symbol, m := range maps {
+		fd := m.FD()
+		if err := ps.Instructions.RewriteMapPtr(symbol, fd); err != nil {
+			if asm.IsUnreferencedSymbol(err) {
+				// Not all programs need to use the map
+				continue
+			}
+			return fmt.Errorf("program %s: %w", ps.Name, err)
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


prog.go
```diff
@@ -88,6 +88,25 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
 
+// RewriteMaps replaces all references to specific maps.
+//
+// Use this function to use pre-existing maps before loading the
+// program.
+func (ps *ProgramSpec) RewriteMaps(maps map[string]*Map) error {
+	for symbol, m := range maps {
+		fd := m.FD()
+		if err := ps.Instructions.RewriteMapPtr(symbol, fd); err != nil {
+			if asm.IsUnreferencedSymbol(err) {
```
lmb, line 99: So in this case, we're explicitly rewriting the maps for a single program. Shouldn't this be an error then?

lmb (commented):
The problem I see with adding `ProgramSpec.RewriteMaps` is that it begs to introduce `ProgramSpec.RewriteConstants` as well. However that isn't possible because constants are rewritten in the `.rodata` map which doesn't exist on a ProgramSpec. To put it another way, it seems to me like `ProgramSpec.RewriteMaps` solves a very specific problem which I don't understand well. It's also simple to implement the function outside of the library. So my instinct is to ask you to carry the function in your own code base.

However, my instinct is just as likely to be wrong, so please give me some more context. On the surface, you should be able to do the following:

```
spec = spec.Copy()
spec.RewriteMaps(...)
NewProgram(spec.Programs["foo"])
```

When discussing on slack you mentioned that this is too expensive. Can you give more detail? Is it the `Copy`? Something else is slow?

florianl:
I just realize one can already rewrite map references for a dedicated program spec, as all the needed components in `ProgramSpec` are available ... 
Sorry for the noise!

PR #257, opened 2021-03-18
program: fix LoadPinnedProgram not setting pinnedPath

lmb:
Currently, loading a program from BPFFS doesn't set it's pinned path
and IsPinned returns false.


aditighag (approved):
Oops! Thanks for fixing this.

aditighag:
@lorenz Is the plan to do a point release to include this fix, or override the version that was released couple of hours back?

lmb:
Overriding a version is pretty much impossible to due the Go module cache, but we can do a point release. I'd wait to see what other bugs appear and do a bugfix release some time next week.

PR #260, opened 2021-03-19
prog: add Flags to ProgSpec

mythi:
Linux 5.10 added sleepable BPF programs.

Quoting uapi/linux/bpf.h:

"If BPF_F_SLEEPABLE is used in BPF_PROG_LOAD command, the verifier will
restrict map and helper usage for such programs. Sleepable BPF programs can
only be attached to hooks where kernel execution context allows sleeping.
Such programs are allowed to use helpers that may sleep like
bpf_copy_from_user()."

The sleepable BPF programs are used, e.g., with LSM hooks and these
are also detected using "lsm.s" ELF section. This commit adds that
capability to elf_reader too.

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -272,11 +272,12 @@ func (ec *elfCode) loadPrograms() (map[string]*ProgramSpec, error) {
 			return nil, fmt.Errorf("program %s: %w", funcSym.Name, err)
 		}
 
-		progType, attachType, attachTo := getProgType(sec.Name)
+		progType, attachType, progFlags, attachTo := getProgType(sec.Name)
 
 		spec := &ProgramSpec{
 			Name:          funcSym.Name,
 			Type:          progType,
+			Flags:         progFlags,
 			AttachType:    attachType,
 			AttachTo:      attachTo,
 			License:       ec.license,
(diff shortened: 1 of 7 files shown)


elf_reader_test.go
```diff
@@ -329,7 +329,7 @@ func TestGetProgType(t *testing.T) {
 	}
 
 	for _, tc := range testcases {
-		pt, at, to := getProgType(tc.section)
+		pt, at, _, to := getProgType(tc.section)
```
lmb, line 332: Can you add a test case for this? Also assert that flags are 0 for regular sections.

elf_reader_test.go
```diff
@@ -329,7 +329,7 @@ func TestGetProgType(t *testing.T) {
...
 		if pt != tc.pt {
```
lmb, line 333: You can change this code to use `cmp.Diff` like elsewhere in the tests if you want.

elf_reader.go
```diff
@@ -834,56 +835,58 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		"sockops":               {SockOps, AttachCGroupSockOps, 0},
+		"sk_skb/stream_parser":  {SkSKB, AttachSkSKBStreamParser, 0},
+		"sk_skb/stream_verdict": {SkSKB, AttachSkSKBStreamParser, 0},
+		"sk_msg":                {SkMsg, AttachSkSKBStreamVerdict, 0},
+		"lirc_mode2":            {LircMode2, AttachLircMode2, 0},
+		"flow_dissector":        {FlowDissector, AttachFlowDissector, 0},
+		"iter/":                 {Tracing, AttachTraceIter, 0},
+		"sk_lookup/":            {SkLookup, AttachSkLookup, 0},
+		"lsm/":                  {LSM, AttachLSMMac, 0},
+		"lsm.s/":                {LSM, AttachLSMMac, unix.BPF_F_SLEEPABLE},
```
lmb, line 868: Can you add the following?

```
$ rg  'SEC_DEF\(".*?\.s/'
tools/lib/bpf/libbpf.c
8603:	SEC_DEF("fentry.s/", TRACING,
8608:	SEC_DEF("fmod_ret.s/", TRACING,
8613:	SEC_DEF("fexit.s/", TRACING,
8625:	SEC_DEF("lsm.s/", LSM,
```

prog_test.go
```diff
@@ -547,21 +547,47 @@ func TestProgramSpecTag(t *testing.T) {
...
+		attachFn    string
+		flags       uint32
+		expectedErr bool
+	}{
+		{
+			attachFn: "task_getpgid",
+		},
+		{
+			attachFn:    "task_setnice",
+			flags:       unix.BPF_F_SLEEPABLE,
```
lmb, line 560: These will have to be gated with `testutils.SkipOnOldKernel` since I'm guessing that BPF_F_SLEEPABLE was introduced at a different time than LSM programs. Maybe easier to have `sleepable bool` in the test definition?

lmb (changes requested):
Some nits, but it seems reasonable to me otherwise.

elf_reader.go
lmb: It's OK to not test them separately.

mythi:
> Some nits, but it seems reasonable to me otherwise.

@lmb OK! thanks for the review. I'll be OOO next week so I'll address the feedback once I'm back

elf_reader_test.go
mythi: done

prog_test.go
mythi: added a kernel version check against 5.10

elf_reader.go
mythi: added

mythi:
looks like CI is failing to create test env (apt failures)

elf_reader_test.go
```diff
@@ -314,32 +314,58 @@ func TestLibBPFCompat(t *testing.T) {
 	})
 }
 
+type elfSectionProgType struct {
```
ti-mo, line 317: I'd prefer to have this type declared in `TestGetProgType`, since it has little meaning outside of that scope. It can be named something generic like `testCase` to avoid it popping up in greps, and the uppercasing of the members is also not necessary as far as I can see.

prog.go
```diff
@@ -56,6 +56,7 @@ type ProgramSpec struct {
 	AttachTo     string
 	Instructions asm.Instructions
```
ti-mo, line 58: Could the purpose of Flags briefly be documented here?

ti-mo (changes requested):
Thank you!

I'm still on the fence whether we should already wrap this up into a `ProgFlags` ([like we're planning with `LoadPinOptions`](https://github.com/cilium/ebpf/pull/249/files#diff-b990161986c40e0867e71c60779bc98a8730a9eb0ca5866ae8189e126863c059R210)) so we don't have to break API when we inevitably add more flags later. I'm not a fan of exposing raw bitfields, it's not user-friendly. Also, the caller doesn't have access to `internal/unix`, so they would have to resort to the same dummy consts workaround for getting code completion on macOS.

mythi:
> I'm still on the fence whether we should already wrap this up into a `ProgFlags` 

@ti-mo I followed what is already available in `MapSpec` and thought `Flags` was justified for `ProgSpec` too.

> I'm not a fan of exposing raw bitfields, it's not user-friendly.

how this works for `MapSpec`? I can see `TestMapInfoFromProc` (info_test.go)  uses raw bitfields for `MapSpec` too.

prog.go
mythi: added brief descriptions for both map and prog `Flags`

elf_reader_test.go
mythi: fixed

lmb (approved):
Looks good to me.

ti-mo (approved):

PR #261, opened 2021-03-20
perf: use constants from sys/unix

florianl:
Replace the hard coded constants with the constants in sys/unix.

### DIFF
--- a/internal/unix/types_other.go
+++ b/internal/unix/types_other.go
@@ -48,6 +48,8 @@ const (
 	RLIM_INFINITY            = 0x7fffffffffffffff
 	RLIMIT_MEMLOCK           = 8
 	BPF_STATS_RUN_TIME       = 0
+	PERF_RECORD_LOST         = 2
+	PERF_RECORD_SAMPLE       = 9
 )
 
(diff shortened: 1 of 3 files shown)


internal/unix/types_other.go
```diff
@@ -48,6 +48,8 @@ const (
 	RLIM_INFINITY            = 0x7fffffffffffffff
 	RLIMIT_MEMLOCK           = 8
 	BPF_STATS_RUN_TIME       = 0
+	PERF_RECORD_LOST         = 2
```
ti-mo, line 51: Isn't `PERF_RECORD_*` Linux-specific? Also, isn't the perf API Linux-specific? I see the BSD's have PMC instead of perf. If so, why do we put the constants here behind build tags but not the perf package itself? For anything non-Linux, we should have stubs that don't refer to Linux symbols, which would allow us to remove unnecessary duplicates here.

Same goes for epoll constants etc.

cc @lmb

lmb: We can just use zero for all these constants.

lmb (approved):

PR #264, opened 2021-03-22
btf: add VarSecinfo.Var

lmb:
VarSecinfo.Type is guaranteed to be a *Var. Unfortunately changing
the field to be *Var breaks type traversal via typeDeque, etc. As
a stopgap, add a getter that wraps a type assertion. This is more
convenient than just documenting that VarSecinfo.Type is a *Var.

### DIFF
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -475,6 +475,13 @@ type VarSecinfo struct {
 	Size   uint32
 }
 
+// Var returns the variable described by the VarSecinfo.
+func (vsi VarSecinfo) Var() *Var {
+	// Type is guaranteed to be a *Var due to the fixup happening in
+	// inflateRawTypes.
+	return vsi.Type.(*Var)
+}
+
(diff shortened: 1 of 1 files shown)


lmb:
cc @mmat11

lmb:
I think this isn't the right approach, basically because it adds an API that isn't straightforward write a test for and its very easy to change the code in a way which destroys the invariant. 

@mmat11 for now I suggest just asserting `v, ok := ds.Vars[i].Type.(*Var)` in the ELF reader code and returning an error when `!ok`.

mmat11:
> I think this isn't the right approach, basically because it adds an API that isn't straightforward write a test for and its very easy to change the code in a way which destroys the invariant.
> 
> @mmat11 for now I suggest just asserting `v, ok := ds.Vars[i].Type.(*Var)` in the ELF reader code and returning an error when `!ok`.

👍🏻 sounds good to me

PR #265, opened 2021-03-25
K(ret)probe and Tracepoint support

ti-mo:
This is a continuation of https://github.com/cilium/ebpf/pull/186. Finally got a Round Tuit!

I believe this addresses one of the bigger shortcomings of the lib (as demonstrated by the verbose tracepoint/kprobe examples that were merged recently), and has been written about in more than a few blog posts over the years. (recent example: https://est357.github.io/posts/cilium_iovisor) There was an issue in the old newtools repo for this that was lost in the migration, so it kind of fell off my radar.

After reaching out to the more prominent/active users of the go-perf package (https://github.com/acln0/perf/issues/30), we've concluded that having this in the lib might be preferable, especially given the tiny subset of go-perf functionality we actually need for ebpf. Additionally, go-perf still doesn't provide the more thorny bits around handling tracefs kprobe_events, so we'd have to implement that here anyway, all while still relying on an external dependency. (whether it eventually makes it into the stdlib or not)

Putting this out here for a first review round as the first 90% is done, but I still need to convert the examples in order to evaluate if the package is fit for external consumption. Having some early feedback would be great. Getting the terminology right was the hardest part, I've found the body of existing libraries and documentation to be inconsistent.

@ungureanuvladvictor uprobe support should fit into this structure nicely, and will be added in a follow-up.

## To Do
- [x] Convert `examples/`
- [ ] Add Go `Example*`s in tests

cc fyi @lmb @mmat11 @florianl @buglloc

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -1,0 +1,295 @@
+package link
+
+import (
+	"crypto/rand"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+var (
+	kprobeEventsPath = filepath.Join(tracefsPath, "kprobe_events")
+)
+
+// Kprobe attaches the given eBPF program to a perf event that fires when the
+// given kernel symbol starts executing. See /proc/kallsyms for available
+// symbols. For example, printk():
+//
+//	Kprobe("printk")
+//
+// The resulting Link must be Closed during program shutdown to avoid leaking
+// system resources.
+func Kprobe(symbol string, prog *ebpf.Program) (Link, error) {
+	k, err := kprobe(symbol, prog, false)
+	if err != nil {
+		return nil, err
+	}
+
+	err = k.attach(prog)
+	if err != nil {
+		k.Close()
+		return nil, err
+	}
+
+	return k, nil
+}
+
+// Kretprobe attaches the given eBPF program to a perf event that fires right
+// before the given kernel symbol exits, with the function stack left intact.
+// See /proc/kallsyms for available symbols. For example, printk():
+//
+//	Kretprobe("printk")
+//
+// The resulting Link must be Closed during program shutdown to avoid leaking
+// system resources.
+func Kretprobe(symbol string, prog *ebpf.Program) (Link, error) {
+	k, err := kprobe(symbol, prog, true)
+	if err != nil {
+		return nil, err
+	}
+
+	err = k.attach(prog)
+	if err != nil {
+		k.Close()
+		return nil, err
+	}
+
+	return k, nil
+}
+
+// kprobe opens a perf event on the given symbol and attaches prog to it.
+// If ret is true, create a kretprobe.
+func kprobe(symbol string, prog *ebpf.Program, ret bool) (*perfEvent, error) {
+	if symbol == "" {
+		return nil, fmt.Errorf("symbol name cannot be empty: %w", errInvalidInput)
+	}
+	if prog == nil {
+		return nil, fmt.Errorf("prog cannot be nil: %w", errInvalidInput)
+	}
+	if !rgxTraceEvent.MatchString(symbol) {
+		return nil, fmt.Errorf("symbol '%s' must be alphanumeric or underscore: %w", symbol, errInvalidInput)
+	}
+	if prog.Type() != ebpf.Kprobe {
+		return nil, fmt.Errorf("eBPF program type %s is not a Kprobe: %w", prog.Type(), errInvalidInput)
+	}
+
+	// Use kprobe PMU if the kernel has it available.
+	tp, err := pmuKprobe(symbol, ret)
+	if err == nil {
+		return tp, nil
+	}
+	if err != nil && !errors.Is(err, ErrNotSupported) {
+		return nil, fmt.Errorf("creating perf_kprobe PMU: %w", err)
+	}
+
+	// Use tracefs if kprobe PMU is missing.
+	tp, err = tracefsKprobe(symbol, ret)
+	if err != nil {
+		return nil, fmt.Errorf("creating trace event '%s' in tracefs: %w", symbol, err)
+	}
+
+	return tp, nil
+}
+
+// pmuKprobe opens a perf event based on a Performance Monitoring Unit.
+// Requires at least 4.17 (e12f03d7031a "perf/core: Implement the
+// 'perf_kprobe' PMU").
+// Returns ErrNotSupported if the kernel doesn't support perf_kprobe PMU,
+// or os.ErrNotExist if the given symbol does not exist in the kernel.
+func pmuKprobe(symbol string, ret bool) (*perfEvent, error) {
+
+	// Getting the PMU type will fail if the kernel doesn't support
+	// the perf_kprobe PMU.
+	et, err := getPMUEventType("kprobe")
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a pointer to a NUL-terminated string for the kernel.
+	sp, err := unsafeStringPtr(symbol)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Parse the position of the bit from /sys/bus/event_source/devices/%s/format/retprobe.
+	config := 0
+	if ret {
+		config = 1
+	}
+
+	attr := unix.PerfEventAttr{
+		Type:   uint32(et),          // PMU event type read from sysfs
+		Ext1:   uint64(uintptr(sp)), // Kernel symbol to trace
+		Config: uint64(config),      // perf_kprobe PMU treats config as flags
+	}
+
+	fd, err := unix.PerfEventOpen(&attr, perfAllThreads, 0, -1, unix.PERF_FLAG_FD_CLOEXEC)
+
+	// Since commit 97c753e62e6c, ENOENT is correctly returned instead of EINVAL
+	// when trying to create a kretprobe for a missing symbol. Make sure ENOENT
+	// is returned to the caller.
+	if errors.Is(err, os.ErrNotExist) || errors.Is(err, unix.EINVAL) {
+		return nil, fmt.Errorf("symbol '%s' not found: %w", symbol, os.ErrNotExist)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("opening perf event: %w", err)
+	}
+
+	// Ensure the string pointer is not collected before PerfEventOpen returns.
+	runtime.KeepAlive(sp)
+
+	// Kernel has perf_kprobe PMU available, initialize perf event.
+	return &perfEvent{
+		fd:       internal.NewFD(uint32(fd)),
+		pmuID:    et,
+		name:     symbol,
+		ret:      ret,
+		progType: ebpf.Kprobe,
+	}, nil
+}
+
+// tracefsKprobe creates a trace event by writing an entry to <tracefs>/kprobe_events.
+// A new trace event group name is generated on every call to support creating
+// multiple trace events for the same kernel symbol. A perf event is then opened
+// on the newly-created trace event and returned to the caller.
+func tracefsKprobe(symbol string, ret bool) (*perfEvent, error) {
+
+	// Generate a random string for each trace event we attempt to create.
+	// This value is used as the 'group' token in tracefs to allow creating
+	// multiple kprobe trace events with the same name.
+	group, err := randomGroup("ebpf")
+	if err != nil {
+		return nil, fmt.Errorf("randomizing group name: %w", err)
+	}
+
+	// Before attempting to create a trace event through tracefs,
+	// check if an event with the same group and name already exists.
+	// Kernels 4.x and earlier don't return os.ErrExist on writing a duplicate
+	// entry, so we need to rely on reads for detecting uniqueness.
+	_, err = getTraceEventID(group, symbol)
+	if err == nil {
+		return nil, fmt.Errorf("trace event already exists: %s/%s", group, symbol)
+	}
+	// The read is expected to fail with ErrNotSupported due to a non-existing event.
+	if err != nil && !errors.Is(err, ErrNotSupported) {
+		return nil, fmt.Errorf("checking trace event %s/%s: %w", group, symbol, err)
+	}
+
+	// Create the kprobe trace event using tracefs.
+	if err := createTraceFSKprobeEvent(group, symbol, ret); err != nil {
+		return nil, fmt.Errorf("creating kprobe event on tracefs: %w", err)
+	}
+
+	// Get the newly-created trace event's id.
+	tid, err := getTraceEventID(group, symbol)
+	if err != nil {
+		return nil, fmt.Errorf("getting trace event id: %w", err)
+	}
+
+	// Kprobes are ephemeral tracepoints and share the same perf event type.
+	fd, err := openTracepointPerfEvent(tid)
+	if err != nil {
+		return nil, err
+	}
+
+	return &perfEvent{
+		fd:        fd,
+		group:     group,
+		name:      symbol,
+		ret:       ret,
+		tracefsID: tid,
+		progType:  ebpf.Kprobe, // kernel only allows attaching kprobe programs to kprobe events
+	}, nil
+}
+
+// createTraceFSKprobeEvent creates a new ephemeral trace event by writing to
+// <tracefs>/kprobe_events. Returns ErrNotSupported if symbol is not a valid
+// kernel symbol, or if it is not traceable with kprobes.
+func createTraceFSKprobeEvent(group, symbol string, ret bool) error {
+	// Open the kprobe_events file in tracefs.
+	f, err := os.OpenFile(kprobeEventsPath, os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		return fmt.Errorf("error opening kprobe_events: %w", err)
+	}
+	defer f.Close()
+
+	// The kprobe_events syntax is as follows (see Documentation/trace/kprobetrace.txt):
+	// p[:[GRP/]EVENT] [MOD:]SYM[+offs]|MEMADDR [FETCHARGS] : Set a probe
+	// r[MAXACTIVE][:[GRP/]EVENT] [MOD:]SYM[+0] [FETCHARGS] : Set a return probe
+	// -:[GRP/]EVENT                                        : Clear a probe
+	//
+	// Some examples:
+	// r:ebpf_1234/r_my_kretprobe nf_conntrack_destroy
+	// p:ebpf_5678/p_my_kprobe __x64_sys_execve
+	//
+	// Leaving the kretprobe's MAXACTIVE set to 0 (or absent) will make the
+	// kernel default to NR_CPUS. This is desired in most eBPF cases since
+	// subsampling or rate limiting logic can be more accurately implemented in
+	// the eBPF program itself. See Documentation/kprobes.txt for more details.
+	pe := fmt.Sprintf("%s:%s/%s %s", kprobePrefix(ret), group, symbol, symbol)
+	_, err = f.WriteString(pe)
+	// Writing to <tracefs>/kprobe_events will return ENOENT when the tracee
+	// kernel symbol does not exist (yet) in the kernel.
+	if os.IsNotExist(err) {
+		return fmt.Errorf("kernel symbol %s not found: %w", symbol, ErrNotSupported)
+	}
+	if err != nil {
+		return fmt.Errorf("writing '%s' to kprobe_events: %w", pe, err)
+	}
+
+	return nil
+}
+
+// closeTraceFSKprobeEvent removes the kprobe with the given group, symbol and kind
+// from <tracefs>/kprobe_events.
+func closeTraceFSKprobeEvent(group, symbol string) error {
+	f, err := os.OpenFile(kprobeEventsPath, os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		return fmt.Errorf("error opening kprobe_events: %w", err)
+	}
+	defer f.Close()
+
+	// See kprobe_events syntax above. Kprobe type does not need to be specified
+	// for removals.
+	pe := fmt.Sprintf("-:%s/%s", group, symbol)
+	if _, err = f.WriteString(pe); err != nil {
+		return fmt.Errorf("writing '%s' to kprobe_events: %w", pe, err)
+	}
+
+	return nil
+}
+
+// randomGroup generates a pseudorandom string for use as a tracefs group name.
+// Returns an error when the output string would exceed 63 characters (kernel
+// limitation), when rand.Read() fails or when prefix contains characters not
+// allowed by rgxTraceEvent.
+func randomGroup(prefix string) (string, error) {
+	if !rgxTraceEvent.MatchString(prefix) {
+		return "", fmt.Errorf("prefix '%s' must be alphanumeric or underscore: %w", prefix, errInvalidInput)
+	}
+
+	b := make([]byte, 8)
+	if _, err := rand.Read(b); err != nil {
+		return "", fmt.Errorf("reading random bytes: %w", err)
+	}
+
+	group := fmt.Sprintf("%s_%x", prefix, b)
+	if len(group) > 63 {
+		return "", fmt.Errorf("group name '%s' cannot be longer than 63 characters: %w", group, errInvalidInput)
+	}
+
+	return group, nil
+}
+
+func kprobePrefix(ret bool) string {
+	if ret {
+		return "r"
+	}
+	return "p"
+}
--- a/link/kprobe_test.go
+++ b/link/kprobe_test.go
@@ -1,0 +1,274 @@
+package link
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+var (
+	kprobeSpec = ebpf.ProgramSpec{
+		Type:    ebpf.Kprobe,
+		License: "MIT",
+		Instructions: asm.Instructions{
+			// set exit code to 0
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+	}
+)
+
+func TestKprobe(t *testing.T) {
+
+	//TODO: implement kprobe program version rewriting for pre-5.0 kernels.
+	// Requires at least 5.0 (6c4fc209fcf9 "bpf: remove useless version check for prog load")
+	testutils.SkipOnOldKernel(t, "5.0", "lifted version check for kprobes")
+
+	c := qt.New(t)
+
+	prog, err := ebpf.NewProgram(&kprobeSpec)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	k, err := Kprobe("printk", prog)
+	c.Assert(err, qt.IsNil)
+	defer k.Close()
+
+	testLink(t, k, testLinkOptions{
+		prog: prog,
+	})
+
+	k, err = Kprobe("bogus", prog)
+	c.Assert(errors.Is(err, os.ErrNotExist), qt.IsTrue)
+	if k != nil {
+		k.Close()
+	}
+}
+
+func TestKretprobe(t *testing.T) {
+
+	//TODO: implement kprobe program version rewriting for pre-5.0 kernels.
+	// Requires at least 5.0 (6c4fc209fcf9 "bpf: remove useless version check for prog load")
+	testutils.SkipOnOldKernel(t, "5.0", "lifted version check for kprobes")
+
+	c := qt.New(t)
+
+	prog, err := ebpf.NewProgram(&kprobeSpec)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	k, err := Kretprobe("printk", prog)
+	c.Assert(err, qt.IsNil)
+	defer k.Close()
+
+	testLink(t, k, testLinkOptions{
+		prog: prog,
+	})
+
+	k, err = Kretprobe("bogus", prog)
+	c.Assert(errors.Is(err, os.ErrNotExist), qt.IsTrue, qt.Commentf("got error: %s", err))
+	if k != nil {
+		k.Close()
+	}
+}
+
+func TestKprobeErrors(t *testing.T) {
+	c := qt.New(t)
+
+	// Invalid Kprobe incantations. Kretprobe uses the same code paths
+	// with a different ret flag.
+	_, err := Kprobe("", nil) // empty symbol
+	c.Assert(errors.Is(err, errInvalidInput), qt.IsTrue)
+
+	_, err = Kprobe("_", nil) // empty prog
+	c.Assert(errors.Is(err, errInvalidInput), qt.IsTrue)
+
+	_, err = Kprobe(".", &ebpf.Program{}) // illegal chars in symbol
+	c.Assert(errors.Is(err, errInvalidInput), qt.IsTrue)
+
+	_, err = Kprobe("foo", &ebpf.Program{}) // wrong prog type
+	c.Assert(errors.Is(err, errInvalidInput), qt.IsTrue)
+}
+
+// Test k(ret)probe creation using perf_kprobe PMU.
+func TestKprobeCreatePMU(t *testing.T) {
+
+	// Requires at least 4.17 (e12f03d7031a "perf/core: Implement the 'perf_kprobe' PMU")
+	testutils.SkipOnOldKernel(t, "4.17", "perf_kprobe PMU")
+
+	c := qt.New(t)
+
+	// kprobe happy path. printk is always present.
+	pk, err := pmuKprobe("printk", false)
+	c.Assert(err, qt.IsNil)
+	defer pk.Close()
+
+	c.Assert(pk.progType, qt.Equals, ebpf.Kprobe)
+
+	// kretprobe happy path.
+	pr, err := pmuKprobe("printk", true)
+	c.Assert(err, qt.IsNil)
+	defer pr.Close()
+
+	c.Assert(pr.progType, qt.Equals, ebpf.Kprobe)
+
+	// Expect os.ErrNotExist when specifying a non-existent kernel symbol
+	// on kernels 4.17 and up.
+	_, err = pmuKprobe("bogus", false)
+	c.Assert(errors.Is(err, os.ErrNotExist), qt.IsTrue, qt.Commentf("got error: %s", err))
+
+	// A kernel bug was fixed in 97c753e62e6c where EINVAL was returned instead
+	// of ENOENT, but only for kretprobes.
+	_, err = pmuKprobe("bogus", true)
+	c.Assert(errors.Is(err, os.ErrNotExist), qt.IsTrue, qt.Commentf("got error: %s", err))
+}
+
+// Test fallback behaviour on kernels without perf_kprobe PMU available.
+func TestKprobePMUUnavailable(t *testing.T) {
+	c := qt.New(t)
+
+	pk, err := pmuKprobe("printk", false)
+	if err == nil {
+		pk.Close()
+		t.Skipf("Kernel supports perf_kprobe PMU, not asserting error.")
+	}
+
+	// Only allow a PMU creation with a valid kernel symbol to fail with ErrNotSupported.
+	c.Assert(errors.Is(err, ErrNotSupported), qt.IsTrue, qt.Commentf("got error: %s", err))
+}
+
+func BenchmarkKprobeCreatePMU(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		pr, err := pmuKprobe("printk", false)
+		if err != nil {
+			b.Error("error creating perf_kprobe PMU:", err)
+		}
+
+		if err := pr.Close(); err != nil {
+			b.Error("error closing perf_kprobe PMU:", err)
+		}
+	}
+}
+
+// Test tracefs k(ret)probe creation on all kernel versions.
+// Create two trace events on the same symbol and ensure their IDs differ.
+func TestKprobeTraceFS(t *testing.T) {
+	c := qt.New(t)
+
+	symbol := "printk"
+
+	// Open and close tracefs kprobe, checking all errors.
+	te, err := tracefsKprobe(symbol, false)
+	c.Assert(err, qt.IsNil)
+	c.Assert(te.Close(), qt.IsNil)
+
+	// Create similar trace events, ensure their IDs differ.
+	te1, err := tracefsKprobe(symbol, false)
+	c.Assert(err, qt.IsNil)
+	defer te1.Close()
+	c.Assert(te1.progType, qt.Equals, ebpf.Kprobe)
+	c.Assert(te1.tracefsID, qt.Not(qt.Equals), 0)
+
+	te2, err := tracefsKprobe(symbol, false)
+	c.Assert(err, qt.IsNil)
+	defer te2.Close()
+	c.Assert(te2.progType, qt.Equals, ebpf.Kprobe)
+	c.Assert(te2.tracefsID, qt.Not(qt.Equals), 0)
+
+	c.Assert(te1.tracefsID, qt.Not(qt.CmpEquals()), te2.tracefsID)
+
+	// Write a kprobe event for a non-existing symbol.
+	err = createTraceFSKprobeEvent("syscalls", "bogus", false)
+	c.Assert(errors.Is(err, ErrNotSupported), qt.IsTrue)
+}
+
+func BenchmarkKprobeCreateTraceFS(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		// Include <tracefs>/kprobe_events operations in the benchmark loop
+		// because we create one per perf event.
+		pr, err := tracefsKprobe("printk", false)
+		if err != nil {
+			b.Error("error creating tracefs perf event:", err)
+		}
+
+		if err := pr.Close(); err != nil {
+			b.Error("error closing tracefs perf event:", err)
+		}
+	}
+}
+
+// Test k(ret)probe creation writing directly to <tracefs>/kprobe_events.
+// Only runs on 5.0 and over. Earlier versions ignored writes of duplicate
+// events, while 5.0 started returning -EEXIST when a kprobe event already
+// exists.
+func TestKprobeCreateTraceFS(t *testing.T) {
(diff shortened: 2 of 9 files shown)


florianl:
The problem I see at the moment with the proposed approach is, that it limits the power of eBPF on k(ret)probe and tracepoints. E.g. it can be useful to enable these eBPF programs on perf event just on certain CPU cores or for particular processes (e.g. a syscall monitor for a particular process). Or in case of a sampling approach to disable and enable it again.
As pointed out by @hodgesds in https://github.com/acln0/perf/issues/30 the perf system can become complex (and time consuming to maintain) quite fast. And so far I don't know of a high level API in Go that covers 90% of use cases. 
Where do you see the limits that should be supported by this proposal and use cases that should use `golang.org/x/sys/unix` directly? I think it is important to document these differences and make the user aware of it.

ti-mo:
Hi @florianl, thanks for taking a look! This is exactly the kind of feedback I'm looking for. The use cases you've given here are still incredibly narrow compared to the perf API surface. Could you think of any use cases that this implementation completely lacks?

> enable these eBPF programs on perf event just on certain CPU cores

Okay, this convinced me to add a `*LinkOpts` argument to `Tracepoint()`, `Kprobe()` etc. to allow the users to turn a few knobs. What you're describing is common in uprobes, correct? Could you give a more practical example where this is used?

> a sampling approach to disable and enable it again

This could be addressed by expanding the `Link` interface to have an `Enable()` and `Disable()` and implementing them on `perfEvent`. WDYT @lmb, would this make sense for other `Link` types?

> Where do you see the limits that should be supported by this proposal and use cases that should use golang.org/x/sys/unix directly?

I feel like users that know they need something radically different would've reached for `x/sys/unix` long ago. The perf_kprobe PMU stuff is done nearly identically in libbpf, and tracefs usage I've seen floating around is all 99% similar to what's in the proposal. (looking at the kernel code, the intersection of bpf+perf is but a tiny fraction of the perf API as a whole)

I think the limits of the package are immediately clear by looking at the API surface: create, attach and close tracepoints, k(ret)probes and uprobes (soon). As the community discovers more use cases, they will be exposed in the package and added to `examples/`. (as well as being presented in a few blog posts on ebpf.io)

>  As pointed out (...) the perf system can become complex (and time consuming to maintain) quite fast. And so far I don't know of a high level API in Go that covers 90% of use cases.

We advertise this as an eBPF API, not a perf API. :slightly_smiling_face: As I saw the same 2 syscalls and 3 ioctls enabling most use cases, I felt it would make sense to maintain them as a coherent set. I think this proposal walks the line between BCC's "load && attach all the things with a single call!" and users having to reach for `runtime.KeepAlive` and `unsafe` for simply attaching a kprobe. It hides the thorny bits and does the heavy lifting while still providing the freedom to choose when/how things are hooked together. It doesn't feel like magic, which is pretty much in line with the rest of the library.

hodgesds:
This [patch](https://git.kernel.org/pub/scm/linux/kernel/git/acme/linux.git/commit/?id=7fac83aaf2eecc9e) might be interesting in this space as well.

lmb (commented):
> the perf system can become complex (and time consuming to maintain) quite fast. And so far I don't know of a high level API in Go that covers 90% of use cases.

As you know I share your worries about the maintenance burden, but at this point I'm convinced that no kprobe support out of the box is a big barrier to adoption. I don't want to support random perf stuff, so we should absolutely focus on doing the smallest viable thing. Maybe asking for 90% is too much, how about 80% then? Sometimes we let the perfect get in the way of the okay, so what is the minimum feature set that is useful?

* Do we need support to run on a subset of CPU? I'd say no: the BPF can call `bpf_get_smp_processor_id` and do it's own filtering.
* Do we need to support turning it on and off again? I'd say no: you can update a map to bail out of your program early.
* Do we need to support a subset of processes? Maybe: BPF can call bpf_get_current_pid_tgid to do it's own filtering. Are there are permission implications?

> This could be addressed by expanding the Link interface to have an Enable() and Disable() and implementing them on perfEvent. WDYT @lmb, would this make sense for other Link types?

`Link` is meant to map as close as possible to `bpf_link`, but it's an interface so that we can backfill things like the Cgroupv2 attachment. So the answer is: if bpf_link has enable and disable we can add it. Otherwise we should do something else. I saw that there is BPF_LINK_DETACH now, maybe that is similar enough. See also the DIY approach with a map above.

@ti-mo this is a nit, but what's up with the empty newlines at the start of functions? I can't figure out what the system behind that is :D

link/kprobe.go
```diff
@@ -0,0 +1,233 @@
...
+		name:      symbol,
+		ret:       ret,
+		progType:  ebpf.Kprobe,
+		finalizer: nil, // no resources are created
+	}, nil
+}
+
+// tracefsKprobe returns a TraceEvent created by writing an entry to
+// <tracefs>/kprobe_events. On Close(), the entry is removed.
+// Repeated calls with the same arguments will yield TraceEvents without
```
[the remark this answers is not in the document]
ti-mo: Yes, still thinking about how to word this at the correct level of depth.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	group string
+	name  string
+
+	// PMU event ID. 0 when kernel doesn't have PMU.
+	pmuID uint64
+	// ID of the trace event on sysfs.
+	tracefsID uint64
+
+	// True for kretprobes/uretprobes.
+	ret bool
```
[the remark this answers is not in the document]
ti-mo: That's what I had initially, but I found it a bit overkill for a boolean state. It has no application outside of u/kprobes (so generalizing it to fnEntry/Exit is not very useful), and a simple flag did the job here. I'll think about it some more, but maybe we should keep it as it is.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if err := pe.Update(program); err != nil {
+		te.Close()
+		return nil, fmt.Errorf("attaching perf event: %w", err)
+	}
+
+	return pe, nil
+}
+
+func (te *traceEvent) perfEventOpenPMU() (int, error) {
```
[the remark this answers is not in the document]
ti-mo: Good point! I think I considered this, but a few lines later in `(*TraceEvent) Attach()`, we call `unix.IoctlSetInt()`, which takes an intptr. I'll play with it a bit more.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"strconv"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+	"golang.org/x/sys/unix"
```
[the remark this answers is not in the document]
ti-mo: My bad, wasn't aware. There's nothing ioctl-related in `internal/unix` yet, will add. :+1:

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, fmt.Errorf("enable perf event: %s", err)
+	}
+
+	pe := &perfEvent{
+		fd:       internal.NewFD(uint32(pfd)),
+		progType: te.progType,
+	}
+
+	if err := pe.Update(program); err != nil {
+		te.Close()
```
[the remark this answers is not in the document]
ti-mo: Damn, nice catch.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+
+// A traceEvent represents a trace event in the Linux kernel.
+type traceEvent struct {
+	// Group and name of the tracepoint/kprobe/uprobe.
+	group string
+	name  string
+
+	// PMU event ID. 0 when kernel doesn't have PMU.
+	pmuID uint64
+	// ID of the trace event on sysfs.
```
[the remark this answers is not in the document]
ti-mo: Yes, all trace events and PMU types are 1-indexed. Will document this equally. This should also read `trace event on tracefs` instead. Thanks!

link/kprobe.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return fmt.Errorf("writing '%s' to kprobe_events: %w", pe, err)
+	}
+
+	return nil
+}
+
+// closeTraceFSKprobeEvent removes the kprobe with the given group, symbol and kind
+// from <tracefs>/kprobe_events.
+func closeTraceFSKprobeEvent(group, symbol string, ret bool) error {
```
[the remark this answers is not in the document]
ti-mo: Yes, entries will pile up in `<tracefs>/kprobe_events`, which is exactly why the refcounted alternative (perf_kprobe PMU) was created. Not sure how many entries `kprobe_events` will hold, I'll give it a try.

link/kprobe.go
```diff
@@ -0,0 +1,233 @@
...
+	//
+	// Some examples:
+	// r:ebpf_1234/r_my_kretprobe nf_conntrack_destroy
+	// p:ebpf_5678/p_my_kprobe __x64_sys_execve
+	//
+	// Leaving the kretprobe's MAXACTIVE set to 0 (or absent) will make the
+	// kernel default to NR_CPUS. This is desired in most all eBPF cases since
+	// subsampling or rate limiting logic can be more accurately implemented in
+	// the eBPF program itself. See Documentation/kprobes.txt for more details.
+	pe := fmt.Sprintf("%s:%s/%s %s", kprobePrefix(ret), group, kprobeTraceEventName(symbol, ret), symbol)
```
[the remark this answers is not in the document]
ti-mo: Correct. `group` and `symbol` are checked against `rgxTraceEvent`, which is alphanum_-, so won't accept spaces.

link/kprobe.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Kprobe returns a TraceEvent that fires when the given kernel symbol starts
+// executing. See /proc/kallsyms for available symbols. For example, printk():
+//
+//	Kprobe("printk")
+//
+// As of kernel 4.15, the presence of the given symbol is only checked when
+// attaching a program. One TraceEvent can be re-used to attach an arbitrary
+// number of programs.
+func Kprobe(symbol string) (*traceEvent, error) {
```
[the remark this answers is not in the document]
ti-mo: I was thinking you might say the opposite. This is what I meant by `evaluate if the package is fit for external consumption`, haven't tried to consume it from outside the package. I wanted to prevent a user from instantiating a `TraceEvent{}` manually, so we'd need to add a few checks in methods on `TraceEvent`. Any tips?

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	PerfAllThreads = -1
+)
+
+// A traceEvent represents a trace event in the Linux kernel.
+type traceEvent struct {
+	// Group and name of the tracepoint/kprobe/uprobe.
+	group string
+	name  string
+
+	// PMU event ID. 0 when kernel doesn't have PMU.
```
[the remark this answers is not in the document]
ti-mo: Correct, will improve the docs.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return unsafe.Pointer(p), nil
+}
+
+// getTraceEventID reads a trace event's ID from tracefs given its group and name.
+// group and name must be alphanumeric (with _-), as required by the kernel.
+func getTraceEventID(group, name string) (uint64, error) {
+	if !rgxTraceEvent.MatchString(group) {
```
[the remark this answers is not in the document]
ti-mo: :facepalm: The regex should be `^[a-zA-Z0-9_-]+$`.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	if te.tracefsID != 0 {
+		// Open a perf event based on a tracefs trace event ID.
+		fd, err := te.perfEventOpenTraceFS()
+		if err != nil {
+			return nil, fmt.Errorf("open tracefs perf event: %w", err)
+		}
+		pfd = fd
+	}
+
+	if err := unix.IoctlSetInt(int(pfd), unix.PERF_EVENT_IOC_ENABLE, 0); err != nil {
```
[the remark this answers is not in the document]
ti-mo: Because `PERF_EVENT_IOC_ENABLE` and `PERF_EVENT_IOC_DISABLE`. The value is ignored. Yay ioctls! Does this need to be documented?

link/tracepoint.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// Tracepoint returns a trace event for the given tracepoint group and name.
+// See /sys/kernel/debug/tracing/events to find available tracepoints. The
+// top-level directory is group, the event's subdirectory is name. Example:
+//
+//	Tracepoint("syscalls", "sys_enter_fork")
+//
+// Multiple calls with the same group and name are unnecessary, as they can
+// be re-used to attach an arbitrary amount of eBPF programs.
```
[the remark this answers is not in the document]
ti-mo: No, a tracepoint is static and cannot be closed or removed in any way. I outlined this at the top of `trace_event.go` but forgot to mention it here. :+1:

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+		return fmt.Errorf("getting perf event fd: %w", err)
+	}
+
+	err = unix.IoctlSetInt(int(pfd), unix.PERF_EVENT_IOC_SET_BPF, prog.FD())
+	if err != nil {
+		return fmt.Errorf("setting perf event bpf program: %w", err)
+	}
+
+	// Store a reference to the attached program.
+	ev.prog = prog
```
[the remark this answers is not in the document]
ti-mo: True, we don't really need this. The program being closed or not doesn't really impact this particular logic, though. If the pointers match, an attach is skipped.

Now I think of it, `Attach()` should probably check if the program is still open/valid, perhaps through a `closed bool` exposed through a `(Program).Closed() bool`? Would that be useful?

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	tid := bytes.TrimSpace(data)
+	return strconv.ParseUint(string(tid), 10, 64)
+}
+
+// getPMUEventType reads a Performance Monitoring Unit's type (numeric identifier)
+// from /sys/bus/event_source/devices/<pmu>/type.
+func getPMUEventType(pmu string) (uint64, error) {
+	path := filepath.Join("/sys/bus/event_source/devices", pmu, "type")
```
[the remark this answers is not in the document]
ti-mo: We should check `pmu` against `rgxTraceEvent` as well :+1:.

link/kprobe_test.go
```diff
@@ -0,0 +1,191 @@
...
+	c.Assert(err, qt.IsNil)
+	// Create a kretprobe on the same symbol using the same group.
+	err = createTraceFSKprobeEvent(group, symbol, true)
+	c.Assert(err, qt.IsNil)
+
+	// Attempt to create an identical kprobe using tracefs,
+	// expect it to fail with os.ErrExist.
+	err = createTraceFSKprobeEvent(group, symbol, false)
+	c.Assert(err, qt.Not(qt.IsNil))
+	// Don't bail out here, the close(s) below needs to fire to avoid dangling
```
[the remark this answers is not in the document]
ti-mo: Ah, didn't occur to me that it would be possible, because we're reusing `err` here (which bites when deferring closures). I wanted to make sure those calls returned `nil`. Since it's a test suite, we can easily do `defer c.Assert(closeTraceFSKprobeEvent(group, symbol, false), qt.IsNil)` instead, which won't gobble up the errors. :100:

link/kprobe.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+var (
+	kprobeEventsPath = filepath.Join(tracefsPath, "kprobe_events")
+
+	// Generate a unique string for each runtime instance of the program.
+	// This value is used as the 'group' token in tracefs entries to make sure
+	// multiple instances of the program don't risk clobbering each others'
+	// kprobe_events entries.
+	traceGroup = mustRandomGroup("ebpf")
```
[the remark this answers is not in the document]
ti-mo: I'm not sure if reaching for a `sync.Once` is worth it, given the `ebpf` input string is baked into the library, and the only real source of panic would be `rand.Read()`. (is that what you're worried about?) I see this as no different than a `regexp.MustCompile` into a package var.

link/kprobe.go
```diff
@@ -0,0 +1,233 @@
...
+	// Get the newly-created trace event's id.
+	tid, err = getTraceEventID(group, kprobeTraceEventName(symbol, ret))
+	if err != nil {
+		return nil, fmt.Errorf("getting trace event id: %w", err)
+	}
+
+	// An entry was written to <tracefs>/kprobe_events in this call,
+	// so set a finalizer that undoes the operation when this TraceEvent
+	// is closed.
+	te.finalizer = func() error { return closeTraceFSKprobeEvent(group, symbol, ret) }
```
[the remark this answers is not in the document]
ti-mo: Correct, `a.Close()` kills `b`, but only on kernels 4.15 and lower, since the first call writes a global entry to tracefs. My rationale was that successive `Kprobe()` calls would get their `Close()`s deferred in order, which would automatically result in `a` being closed last, since `defer` is a stack. I'm struggling to document this trap without going to deep into the weeds, but maybe briefly mentioning PMU vs. kprobe_events would be useful. (as well as mentioning that on 4.15 and earlier, kprobe events are global)

The finalizer approach felt like the most straightforward to me. The alternative would be a refcounting pool that closes the kprobe event when the last handle is closed, _but only if the current process is the one that wrote the record initially_. Also consider that:
- the caller might want to avoid closing a kprobe_event handle on purpose after pinning a program and perf ring
- other processes can create these events as well, so any pre-existing handle should never be taken ownership over

link/trace_event.go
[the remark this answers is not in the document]
ti-mo: Added a small comment.

link/trace_event.go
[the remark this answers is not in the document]
ti-mo: I'm moving input validation to happen once, as early as possible, and adding path sanitizing once in a shared `uint64FromFile(...string)`.

link/kprobe.go
```diff
@@ -0,0 +1,231 @@
...
+	// Before attempting to create a trace event through tracefs,
+	// check if an event with the same name already exists.
+	tid, err := getTraceEventID(group, kprobeTraceEventName(symbol, ret))
+	if err != nil && !errors.Is(err, ErrNotSupported) {
+		return nil, fmt.Errorf("getting trace event id: %w", err)
+	}
+	if err == nil {
+		// Kprobe trace event already exists, return a TraceEvent to the caller.
+		// Since this kprobe event was created by a prior call to this function,
+		// or even by another tool or process, it should not be removed by Close().
```
mmat11, line 114: Correct me if I am wrong, another tool or process shouldn't manage probes in the same group/namespace of this instance, right?
If we start using this probe and then the other process deletes it, the probe will "stop working" but for the library everything is ok.

In theory the same thing can happen in the same instance:

`a, _ = Kprobe("__x64_sys_execve")`
`b, _ = Kprobe("__x64_sys_execve")`

What about instanciating probes with an identifier? Example:

`p:ebpf_1234/p_my_kprobe_123 __x64_sys_execve`
`p:ebpf_1234/p_my_kprobe_456 __x64_sys_execve`

This way we should always expect the finalizer to be present (and should then be safe to call it)

link/kprobe.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Kprobe trace event doesn't exist yet, create it using tracefs.
+	if err := createTraceFSKprobeEvent(group, symbol, ret); err != nil {
+		return nil, fmt.Errorf("creating kprobe event on tracefs: %w", err)
+	}
+
+	// Get the newly-created trace event's id.
+	tid, err = getTraceEventID(group, kprobeTraceEventName(symbol, ret))
+	if err != nil {
```
mmat11, line 128: probably we should call `closeTraceFSKprobeEvent` and cleanup before returning an error

link/trace_event.go
```diff
@@ -0,0 +1,299 @@
...
+// - perf event: An object instantiated based on an existing trace event or
+//   kernel symbol. Referred to by fd in userspace.
+//   Exactly one eBPF program can be attached to a perf event. Multiple perf
+//   events can be created from a single trace event. Closing a perf event
+//   stops any further invocations of the attached eBPF program.
+
+var (
+	tracefsPath = "/sys/kernel/debug/tracing"
+
+	// Trace event groups and names must adhere to this set of characters.
```
mmat11, line 42: IIRC names cannot start with a number

mmat11:
```
root@ubuntu-2gb-nbg1-1:~# echo 'p:test123/myprobe __x64_sys_execve' >> /sys/kernel/debug/tracing/kprobe_events

root@ubuntu-2gb-nbg1-1:~# cat /sys/kernel/debug/tracing/kprobe_events
p:test123/myprobe __x64_sys_execve

root@ubuntu-2gb-nbg1-1:~# echo 'p:test123/123myprobe2 __x64_sys_execve' >> /sys/kernel/debug/tracing/kprobe_events
-bash: echo: write error: Invalid argument
```

but we never do that because we always hardcode the prefix, I don't know if it's worth adapting the regex

ti-mo: Huh, interesting find! The prefix is only used for generating the group (the component before the slash), but the same rule seems to apply to both groups and names:

```
# echo 'p:1test/probe __x64_sys_execve' >> /sys/kernel/debug/tracing/kprobe_events
echo: write error: invalid argument
# echo 'p:test/1probe __x64_sys_execve' >> /sys/kernel/debug/tracing/kprobe_events
echo: write error: invalid argument
```

Tracepoint and kprobe groups, names and kernel symbols are matched against this regex. I can't remember why I allowed dashes to appear, as the kernel doesn't seem to allow that. There are no kernel symbols containing dashes either, nor are there any that start with a number. Will remove the dash and disallow leading numbers.

link/kprobe.go
ti-mo: True, I've given it some thought, and not taking ownership in some cases will always be brittle. Will generate a random group name for each creation instead, which means the finalizer system can be removed completely. This will allow `TraceEvent`s to be serialized using `gob` for tools that want to implement seamless restarts and will still allow those tools to clean up after themselves properly.

link/kprobe.go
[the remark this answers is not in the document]
ti-mo: Will remove the finalizer system, see https://github.com/cilium/ebpf/pull/265#discussion_r602519878.

link/kprobe.go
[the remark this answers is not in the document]
ti-mo: Will no longer generate a group as a side effect, these will be generated on every `Kprobe()` call instead.

link/kprobe.go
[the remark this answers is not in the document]
lmb: A user can always do `new(Type)` and shoot themselves in the foot. I don't consider that something to worry about.

link/trace_event.go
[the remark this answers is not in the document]
lmb: internal.FD has a Value method I think, if that isn't enough add what you need. Having the finalizer is a must imo, too easy to get wrong.

link/trace_event.go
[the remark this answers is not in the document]
lmb: > If the pointers match, an attach is skipped.

I can call `prog.Clone()` and invoke this function. Should it bail out early or not? IMO, in the absence of a strong reason to do this ww should drop this.

> Now I think of it, Attach() should probably check if the program is still open/valid, perhaps through a closed bool exposed through a (Program).Closed() bool? Would that be useful?

Using an object after calling `Close()` on it is in the same realm as `new(Type)` for me: not worth defending against.

link/kprobe.go
ti-mo: Hmm, not being able to read a newly-created trace event would be a freak error IMO. If this read fails, there's probably something unexpected going on in the kernel and we shouldn't try to manipulate state beyond that point.

mmat11: makes sense 👍🏻

link/kprobe.go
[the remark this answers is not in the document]
ti-mo: Thousands of `kprobe_events` entries can be created in seconds, though cleaning them up takes significantly longer. Having many entries in this file is not really an issue. I'll include some benchmarks with the proposal:

```
BenchmarkKprobeCreatePMU
BenchmarkKprobeCreatePMU-16        	      73	  86018197 ns/op
BenchmarkKprobeCreateTraceFS
BenchmarkKprobeCreateTraceFS-16    	      60	  94638102 ns/op
```

The tracefs benchmark includes 1 `kprobe_events` write for each perf event creation.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return fd, nil
+}
+
+// unsafeStringPtr returns an unsafe.Pointer to a NUL-terminated copy of str.
+func unsafeStringPtr(str string) (unsafe.Pointer, error) {
+	p, err := unix.BytePtrFromString(str)
```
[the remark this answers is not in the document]
ti-mo: Done, see https://github.com/cilium/ebpf/pull/268.

link/trace_event.go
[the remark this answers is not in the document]
ti-mo: If I understand the following kernel code correctly: https://elixir.bootlin.com/linux/v5.11.10/source/kernel/trace/bpf_trace.c#L1943, `PERF_EVENT_IOC_SET_BPF` doesn't look like it replaces an attached program at all. It adds it to a prog array owned by the perf event, so `Update()` will keep appending to it. Will have to play around with this a bit to see if that's actually what happens.

link/trace_event.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (ev *perfEvent) Update(prog *ebpf.Program) error {
+	if t := prog.Type(); t != ev.progType {
+		return fmt.Errorf("invalid program type (expected %s): %s", ev.progType, t)
+	}
+	if prog.FD() < 0 {
+		return fmt.Errorf("invalid program: %w", internal.ErrClosedFd)
+	}
+
+	// Return if the given prog is already attached.
```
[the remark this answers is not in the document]
ti-mo: See https://github.com/cilium/ebpf/pull/265#discussion_r603463235, I think it appends the prog to a prog array.

link/trace_event.go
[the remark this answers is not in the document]
lmb: I think your reading is correct, for a replace the second argument to bpf_prog_array_copy should be non-NULL.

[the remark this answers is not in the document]
ti-mo: I'll also put these findings in the code, but calling `PERF_EVENT_IOC_SET_BPF` multiple times on the same perf event has only been possible since 4.15 (e87c6bc3852b "bpf: permit multiple bpf attachments for a single perf event"). Before that, it returns EEXIST. There doesn't seem to be a way to remove a BPF prog from a perf event (tried `PERF_EVENT_IOC_RESET` and `_DISABLE`).

The `perf_event_detach_bpf_prog` kernel fn only seems to be called during perf event teardown, and I can't find a way to trigger it using a syscall/ioctl. I'll make `Update()` not implemented for perf events for now. For kernels after 4.15, we could support attaching multiple progs to a single perf event, but the current semantics of  `Link` don't make that very intuitive. Omitting it for now would be the safest choice.

link/kprobe.go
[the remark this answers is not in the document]
ti-mo: This is no longer an issue, unique TraceEvents are now created on every call to `tracefsKprobe`.

link/kprobe.go
```diff
@@ -0,0 +1,221 @@
...
+	// p[:[GRP/]EVENT] [MOD:]SYM[+offs]|MEMADDR [FETCHARGS] : Set a probe
+	// r[MAXACTIVE][:[GRP/]EVENT] [MOD:]SYM[+0] [FETCHARGS] : Set a return probe
+	// -:[GRP/]EVENT                                        : Clear a probe
+	//
+	// Some examples:
+	// r:ebpf_1234/r_my_kretprobe nf_conntrack_destroy
+	// p:ebpf_5678/p_my_kprobe __x64_sys_execve
+	//
+	// Leaving the kretprobe's MAXACTIVE set to 0 (or absent) will make the
+	// kernel default to NR_CPUS. This is desired in most all eBPF cases since
```
etherandrius, line 158: Nit: in most ~~all~~ eBPF cases

ti-mo:
I think this is ready for wider review. Updated examples are here https://github.com/ti-mo/ebpf/commit/7deb85b2c66a00abdef30f4874b060e80c5f42b9 and will follow after this has been merged. (go.mod constraints)

@kkourt @jrfastab What do you think about the API? Anything that should be added?

link/kprobe.go
```diff
@@ -0,0 +1,221 @@
...
+	kprobeEventsPath = filepath.Join(tracefsPath, "kprobe_events")
+)
+
+// Kprobe returns a TraceEvent that fires when the given kernel symbol starts
+// executing. See /proc/kallsyms for available symbols. For example, printk():
+//
+//	Kprobe("printk")
+//
+// As of kernel 4.17, the presence of the given symbol is only checked when
+// Attaching a program later. One TraceEvent can be reused to attach an
```
lmb, line 23: Can you document that `TraceEvent.Attach` can be called multiple times instead? (I'm not sure if that is true, but it's kind of implied by the API)

link/kprobe.go
```diff
@@ -0,0 +1,221 @@
...
+
+// tracefsKprobe returns a TraceEvent created by writing an entry to
+// <tracefs>/kprobe_events. On Close(), the entry is removed.
+// A new group identifier is generated on every call to ensure uniqueness.
+func tracefsKprobe(symbol string, ret bool) (*TraceEvent, error) {
+
+	// Generate a random string for each trace event we attempt to create.
+	// This value is used as the 'group' token in tracefs to allow creating
+	// multiple kprobe trace events with the same name.
+	group := mustRandomGroup("ebpf")
```
lmb, line 102: Return the error instead of panicking?

link/kprobe.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// Before attempting to create a trace event through tracefs,
+	// check if an event with the same group and name already exists.
+	// Kernels 4.x and earlier don't return os.ErrExist on writing a duplicate
+	// entry, so we need to rely on reads for detecting uniqueness.
+	_, err := getTraceEventID(group, symbol)
```
lmb, line 108: This means the code we have to generate a random group isn't random enough? In that case, can we make it more random and drop the check here? If not, should this retry to find a different group?

link/kprobe_test.go
```diff
@@ -0,0 +1,241 @@
...
+	testutils.SkipOnOldKernel(t, "5.0", "<tracefs>/kprobe_events doesn't reject duplicate events")
+
+	c := qt.New(t)
+
+	symbol := "printk"
+	pg := mustRandomGroup("ebpftest")
+	rg := mustRandomGroup("ebpftest")
+
+	// Tee up cleanups in case any of the Asserts abort the function.
+	defer func() { _ = closeTraceFSKprobeEvent(pg, symbol) }()
```
lmb, line 201: Just deferring them directly doesn't work?

link/link.go
```diff
@@ -9,6 +10,7 @@ import (
 )
 
 var ErrNotSupported = internal.ErrNotSupported
+var ErrInvalidInput = errors.New("invalid input")
```
lmb, line 13: Unexport? I'd also move it closer to where it's used.

link/trace_event.go
```diff
@@ -0,0 +1,322 @@
...
+//   events can be created from a single trace event. Closing a perf event
+//   stops any further invocations of the attached eBPF program.
+
+var (
+	tracefsPath = "/sys/kernel/debug/tracing"
+
+	// Trace event groups, names and kernel symbols must adhere to this set
+	// of characters. Non-empty, first character must not be a number, all
+	// characters must be alphanumeric or underscore.
+	rgxTraceEvent = regexp.MustCompile("^[a-zA-Z_][0-9a-zA-Z_]*$")
```
lmb, line 46: This allows empty strings, is that intentional?

link/trace_event.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+		ifd = fd
+
+	default:
+		return nil, errors.New("need a PMU type or tracefs ID")
+	}
+
+	kfd, err := ifd.Value()
+	if err != nil {
+		return nil, fmt.Errorf("getting perf event fd: %w", err)
```
lmb, line 125: This and the following return leak `ifd`.

Nit: I think it'd be ok to ignore the error here and just use kfd.

link/trace_event.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	et := bytes.TrimSpace(data)
+	return strconv.ParseUint(string(et), 10, 64)
+}
+
+// A perfEvent represents a perf event kernel object. Exactly one eBPF program
+// can be attached to it. It is created based on a tracefs trace event or a
+// Performance Monitoring Unit (PMU).
+type perfEvent struct {
```
lmb, line 285: Can you run this through `func testLink` please, that checks a bunch of Link invariants.

link/trace_event_test.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+		})
+	}
+}
+
+func TestTraceEventClose(t *testing.T) {
+	c := qt.New(t)
+
+	var te TraceEvent
+	c.Assert(te.Close(), qt.ErrorMatches, ".* empty trace event")
```
lmb, line 76: I'm not a fan of asserting error strings. Users shouldn't rely on them so we shouldn't test them. If there are some invariants to uphold it's nicer to use an unexported sentinel like you did for errInvalidInput. This way I'll change an error string somewhere, and a test blows up for no good reason.

link/trace_event_test.go
```diff
@@ -0,0 +1,80 @@
...
+				t.Errorf("expected string '%s' to %s regex match", tt.in, exp)
+			}
+		})
+	}
+}
+
+func TestTraceEventClose(t *testing.T) {
+	c := qt.New(t)
+
+	var te TraceEvent
```
lmb, line 75: What does this test? That stuff doesn't panic if you don't use a constructor for `TraceEvent`?

lmb (commented):
This looks really nice, thank you :) I have a couple of questions to understand the semantics better, but nothing major.

lmb:
> > the perf system can become complex (and time consuming to maintain) quite fast. And so far I don't know of a high level API in Go that covers 90% of use cases.
>  
>     * Do we need support to run on a subset of CPU? I'd say no: the BPF can call `bpf_get_smp_processor_id` and do it's own filtering.
> 
>     * Do we need to support turning it on and off again? I'd say no: you can update a map to bail out of your program early.
> 
>     * Do we need to support a subset of processes? Maybe: BPF can call bpf_get_current_pid_tgid to do it's own filtering. Are there are permission implications?

Maybe this comment got lost, what are we going to do about these?

lmb:
Another thing that occurred to me too late:

Current API:
```
tp, _ := Tracepoint(...)
link, _ := tp.Attach(prog)
```

What if we did this instead:
```
link, _ := Tracepoint(..., prog)
```

Pros:
* Shorter in the common case
* Works like the other links
* TraceEvent isn't exported
* "Simpler"

Cons:
* Can't attach multiple programs via a single TraceEvent. Does that ever happen though?

This might also make implementing `perfEvent.Update()` in the future simpler:
```
func (pe) Update(*Program) error {
    // create new TraceEvent in the background
    // attach program
    // close old TraceEvent, which removes the old bpf program (???)
}
```

link/kprobe_test.go
ti-mo: It does, but then we'd need to disable errcheck for those lines. This is fine.

link/trace_event.go
ti-mo: It doesn't, the asterisk only affects the second character class. There's even a test `{"empty string", "", true},` in `TestTraceEventRegex` The first character must always be alphanum_.

link/kprobe.go
ti-mo: There used to be a retry loop here, but I couldn't justify the complexity I needed to add to be able to test it properly. The value add wasn't clear, because this is never supposed to happen in the first place. Initially, my idea was to allow the caller to specify the group prefix (hardcoded to `ebpf` here), so I kept the random segment to a minimum. Will keep the check here in place, but increase the amount of randomness in the string.

TOCTOU, I know, but this is what we've got to work with on older kernels. Newer kernels will EEXIST when writing to the file anyway, so this is not an issue there. Lowest common denominator.

link/kprobe.go
ti-mo: Sure. Panicking in a library is bad form.

link/kprobe.go
```diff
@@ -0,0 +1,221 @@
...
+// executing. See /proc/kallsyms for available symbols. For example, printk():
+//
+//	Kprobe("printk")
+//
+// As of kernel 4.17, the presence of the given symbol is only checked when
+// Attaching a program later. One TraceEvent can be reused to attach an
+// arbitrary number of programs.
+//
+// Close() must be called on the resulting TraceEvent to clean up system state
+// on kernels before 4.17. On those kernels, Closing a TraceEvent created by
```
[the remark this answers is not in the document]
ti-mo: > On < 4.17 calling Close will undo any calls to TraceEvent.Attach, but on later ones it doesn't?

Partially correct. On 4.17 and earlier, closing a kprobe trace event closes all perf events derived from it. That's just how it works in the kernel.

> If at all possible the API should be consistent across kernel versions

I'd argue that this approach **is** consistent across kernel versions in that it automatically picks the most recent kernel API available. If used as prescribed, a program written against this API will work as expected across multiple kernel versions, without the caller needing to take any differences into account.

link/kprobe_test.go
lmb: Ah, the false positive strikes again.

link/kprobe.go
lmb: Sorry the comment was misleading: document it on the function `TraceEvent.Attach` instead of here and in Kretprobe.

link/trace_event.go
ti-mo: Good catch. fixed.

link/kprobe.go
```diff
@@ -0,0 +1,224 @@
...
+	pe := fmt.Sprintf("-:%s/%s", group, symbol)
+	if _, err = f.WriteString(pe); err != nil {
+		return fmt.Errorf("writing '%s' to kprobe_events: %w", pe, err)
+	}
+
+	return nil
+}
+
+// randomGroup generates a pseudorandom string for use as a tracefs group name.
+// Panics when the output string would exceed 63 characters (kernel limitation),
```
mmat11, line 198: `Panics` -> `Returns an error`

link/tracepoint.go
```diff
@@ -0,0 +1,41 @@
...
+//
+//	Tracepoint("syscalls", "sys_enter_fork")
+//
+// Note that attaching eBPF programs to syscalls (sys_enter_*/sys_exit_*) is
+// only possible as of kernel 4.14 (commit cf5f5ce).
+//
+// Multiple calls with the same group and name are unnecessary. Tracepoints
+// can be reused to Attach() an arbitrary amount of eBPF programs. Closing a
+// TraceEvent created by Tracepoint will not disconnect attached perf events.
+func Tracepoint(group, name string) (*TraceEvent, error) {
```
etherandrius, line 22: Can we create and expose a TraceEvent interface ? instead of a TraceEvent struct ? 

This would help for users writing unit tests.

link/kprobe_test.go
```diff
@@ -0,0 +1,249 @@
...
+			b.Error("error creating tracefs-backed perf event:", err)
+		}
+
+		if err := fd.Close(); err != nil {
+			b.Error("error closing tracefs perf event:", err)
+		}
+	}
+}
+
+// Test k(ret)probe creation writing durectly to <tracefs>/kprobe_events.
```
mmat11, line 187: typo `durectly`

link/tracepoint.go
ti-mo: Could you elaborate on which behaviours you would like to test as a consumer of the package?

etherandrius: I think for non-trivial BPF program deployments it would be important to test if attachments and close calls are correct. 

An example use case I am working on is creating and deleting uprobes for individual docker containers, which are constantly coming and going. My program needs to manage the lifecycle of these uprobes, Ideally I would be able to test the interaction with the `cilium/ebpf` library by just mocking interfaces, the alternative is introducing a shim between my code and the library. 
> The above example is irrelevant for kprobes, but it seems like the same API will be used for both program types.

Having said this returning an interface here is not a silver bullet for testing. My use case dynamically finds the binary that we need to attach too. Hence we need to dynamically call `link.Uprobe` (this is equivalent to dynamically figuring out symbol for `link.Kprobe`). `link.Uprobe` cannot be mocked and proper testing requires a shim.

Ideal solution for above is making `TraceEvent` an interface and making sure that the Uprobe API supports dynamic configuration of which binary file the uprobe is attached to.

ti-mo: > making sure that the TraceEvent API supports dynamic configuration of which binary file the uprobe is attached to

I don't think the kernel supports moving an existing perf event to another symbol. That would require creating a new perf event. We're considering to implement `Link.Update(prog)` for perf events that would do this as a convenience method, but that wouldn't cover your use case.

That said, I've factored out `TraceEvent` completely last night. The API will be `Kprobe(symbol, *prog) (Link, error)` instead of a split create/attach stage, will push this in a bit.

ti-mo:
`Tracepoint()`, `Kprobe()` and `Kretprobe()` now take group/name/symbol and `*ebpf.Program`, no longer burdening the caller with keeping track of `TraceEvent` objects to correctly manage their lifecycle. Closing a kprobe `TraceEvent` early on kernels pre-4.17 could've been a source of subtle bugs. The new API design maps better onto PMU perf events, which is the future anyway.

link/trace_event.go
ti-mo: As far as I can see, `testLink()` only calls `Pin()` and `Update()`, neither of which are currently supported by the type. Maybe we should do this if/when we implement `Update()` instead?

ti-mo: Done.

lmb: Thanks! I think it also asserts that the correct errors are returned, so there is still use I think.

link/tracepoint.go
lmb: My 2c: we should avoid returning an interface in our API if possible, aka `accept interfaces, return structs`. If we do return an interface (that is not from the stdlib) we should make it non-implementable for users, like `Link`. Otherwise our hands are tied in terms of API evolution.

@etherandrius does anything prevent you from defining your own interface and using that for stub out unit tests?

lmb (approved):

link/tracepoint.go
ti-mo: I didn't want to expand on this too much, but even if the return type is an interface (which `Link` is), the underlying return value is still a struct. `Link` is indeed non-implementable outside of the package, so if you want to stub out a Link in a consuming program, you'd still have to wrap it. I think that's acceptable.

PR #268, opened 2021-03-29
internal - use BytePtrFromString from stdlib in NewStringPointer()

ti-mo:
As requested in https://github.com/cilium/ebpf/pull/265#discussion_r602169513.

@lmb

### DIFF
--- a/internal/ptr.go
+++ b/internal/ptr.go
@@ -1,6 +1,10 @@
 package internal
 
-import "unsafe"
+import (
+	"unsafe"
+
+	"github.com/cilium/ebpf/internal/unix"
+)
 
 // NewPointer creates a 64-bit pointer from an unsafe Pointer.
 func NewPointer(ptr unsafe.Pointer) Pointer {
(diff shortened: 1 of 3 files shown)


internal/ptr.go
```diff
@@ -22,9 +26,10 @@ func NewStringPointer(str string) Pointer {
 		return Pointer{}
 	}
 
-	// The kernel expects strings to be zero terminated
-	buf := make([]byte, len(str)+1)
-	copy(buf, str)
+	p, err := unix.BytePtrFromString(str)
+	if err != nil {
+		return Pointer{}
```
lmb, line 31: Should we return the error here?

lmb (approved):

internal/ptr.go
ti-mo: Well, I tried that initially, but didn't like it. `NewStringPointer` is used in a few struct literal decls, and they'd have to be moved out for error checking. `unix.BytePtrFromString`'s only error is  `-EINVAL` if the input string contains a `\0`. I'm not sure if that's worth protecting against.

Issue #263, opened 2021-03-22
pinning an object for the first time can return EEXIST


lmb:
Assuming that `/foo/bar` is a pinned program on a bpffs:

```
prog := NewProgram()
prog.Pin("/foo/bar") // returns EEXIST
```

Compare this to what happens when the program is already pinned elsewhere:

```
prog := NewProgram()
prog.Pin("/foo/baz")
prog.Pin("/foo/bar") // returns nil
```

This is because we call os.Rename behind the scenes in the latter case. I'm not sure what the correct behaviour is but I think that the current behaviour is surprising and we should do something about it.

I realised the implementation has another corner case: os.Rename will return `EXDEV` if the old and new path don't reside on the same file system.

Some testcases which show the problem in detail:

```
func TestProgramPinExisting(t *testing.T) {
	prog := createSocketFilter(t)
	tmp := tempBPFFS(t)

	path := filepath.Join(tmp, "program")
	if err := prog.Pin(path); err != nil {
		t.Fatal(err)
	}

	prog2, err := prog.Clone()
	if err != nil {
		t.Fatal(err)
	}

	if err := prog2.Pin(path); err != nil {
		t.Fatal("Can't override existing pin:", err)
	}
}

func TestProgramPinRenameExisting(t *testing.T) {
	prog := createSocketFilter(t)
	tmp := tempBPFFS(t)

	path := filepath.Join(tmp, "program")
	if err := prog.Pin(path); err != nil {
		t.Fatal(err)
	}

	prog2, err := prog.Clone()
	if err != nil {
		t.Fatal(err)
	}

	if err := prog2.Pin(filepath.Join(tmp, "prog2")); err != nil {
		t.Fatal(err)
	}
	if err := prog2.Pin(path); err != nil {
		t.Fatal("Can't override existing pin:", err)
	}
}
```

lmb:
cc @aditighag

lmb:
I think we should document that moving a pin across filesystems is not supported, since I don't see a use case for it.

As for the EEXIST behaviour: being able to override a pin atomically is useful when upgrading pinned state (which is how I hit this problem). However, making this the default behaviour seems a bit risky, overwriting should be opt in. So for the purposes of this bug I think we should call `renameat2` with `RENAME_NOREPLACE` which would bring the second case in line with the first one.
[...]

PR #271, opened 2021-04-02
pinning: Disallow replace during pinning

aditighag:
The current implementation uses `os.Rename` which calls `rename` to overwrite an existing pinned path. If the new path already existed before the overwrite, it'll simply be replaced with this syscall. Hence, use the `renameat2` syscall instead that can disallow the replace behavior.

Note to reviewers: Do we want to expose an API that gives user the option to instead replace the new pinned path (i.e. call `renameat2 `with `RENAME_EXCHANGE`), which is the default behavior of the `rename`* syscalls?

Fixes: #263 
Suggested-by : @lmb

### DIFF
--- a/internal/unix/types_other.go
+++ b/internal/unix/types_other.go
@@ -54,6 +54,8 @@ const (
 	BPF_STATS_RUN_TIME       = 0
 	PERF_RECORD_LOST         = 2
 	PERF_RECORD_SAMPLE       = 9
+	AT_FDCWD                 = -0x2
+	RENAME_NOREPLACE         = 0x1
 )
 
 // Statfs_t is a wrapper
@@ -237,6 +239,11 @@ func Tgkill(tgid int, tid int, sig syscall.Signal) (err error) {
 // BytePtrFromString is a wrapper
 func BytePtrFromString(s string) (*byte, error) {
 	return nil, errNonLinux
+}
+
+// Renameat2 is a wrapper
+func Renameat2(olddirfd int, oldpath string, newdirfd int, newpath string, flags uint) error {
+	return errNonLinux
 }
 
 func KernelRelease() (string, error) {
--- a/map.go
+++ b/map.go
@@ -805,7 +805,8 @@ func (m *Map) Clone() (*Map, error) {
 // Pin persists the map on the BPF virtual file system past the lifetime of
 // the process that created it .
 //
-// Calling Pin on a previously pinned map will override the path.
+// Calling Pin on a previously pinned map will overwrite the path, except when
+// the new path already exists. Re-pinning across filesystems is not supported.
 // You can Clone a map to pin it to a different path.
 //
 // This requires bpffs to be mounted above fileName. See https://docs.cilium.io/en/k8s-doc/admin/#admin-mount-bpffs
--- a/map_test.go
+++ b/map_test.go
@@ -433,6 +433,30 @@ func TestMapPinWithEmptyPath(t *testing.T) {
 	c.Assert(err, qt.Not(qt.IsNil))
 }
 
+func TestMapPinFailReplace(t *testing.T) {
+	tmp := tempBPFFS(t)
+	c := qt.New(t)
+	spec := spec1.Copy()
+	spec2 := spec1.Copy()
+	spec2.Name = spec1.Name + "bar"
+
+	m, err := NewMapWithOptions(spec, MapOptions{PinPath: tmp})
+	if err != nil {
+		t.Fatal("Failed to create map:", err)
+	}
+	defer m.Close()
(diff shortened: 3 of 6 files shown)


map.go
```diff
@@ -805,7 +805,8 @@ func (m *Map) Clone() (*Map, error) {
 // Pin persists the map on the BPF virtual file system past the lifetime of
 // the process that created it .
 //
-// Calling Pin on a previously pinned map will override the path.
+// Calling Pin on a previously pinned map will overwrite the path, except when
+// the new path already exists. Pinning across filesystems is not supported.
```
ti-mo, line 809:
```suggestion
// the new path already exists. Re-pinning across filesystems is not supported.
```

pinning.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os"
 
 	"github.com/cilium/ebpf/internal"
+
+	"golang.org/x/sys/unix"
```
ti-mo, line 10: See CI failure: https://ebpf.semaphoreci.com/jobs/e430af84-9549-413d-94d0-5c1b0d093d9d

`internal/unix` should be imported here, with wrappers and constants added to `types_other.go` and `types_linux.go`.

pinning.go
```diff
@@ -20,7 +22,7 @@ func pin(currentPath, newPath string, fd *internal.FD) error {
 	}
 	var err error
 	// Object is now moved to the new pinning path.
```
ti-mo, line 24: I would document the reason for using `renameat2` here.

map_test.go
```diff
@@ -433,6 +433,32 @@ func TestMapPinWithEmptyPath(t *testing.T) {
...
+		t.Fatal("Failed to create map:", err)
+	}
+	defer m.Close()
+	m2, err := NewMapWithOptions(spec2, MapOptions{PinPath: tmp})
+	if err != nil {
+		t.Fatal("Failed to create map2:", err)
+	}
+	defer m2.Close()
+	pinned := m.IsPinned()
+	c.Assert(pinned, qt.Equals, true)
```
ti-mo, line 454:
```suggestion
	c.Assert(m.IsPinned(), qt.Equals, true)
```

map_test.go
```diff
@@ -433,6 +433,32 @@ func TestMapPinWithEmptyPath(t *testing.T) {
...
+	newPath := filepath.Join(tmp, spec2.Name)
+
+	err = m.Pin(newPath)
+	fmt.Printf("ERROR is %v", err)
```
ti-mo, line 458: Stray Printf() here.

map_test.go
```diff
@@ -433,6 +433,32 @@ func TestMapPinWithEmptyPath(t *testing.T) {
...
+	c.Assert(err, qt.Not(qt.IsNil))
```
ti-mo, line 459:
```suggestion
	c.Assert(m.Pin(newPath), qt.Not(qt.IsNil))
```

ti-mo (changes requested):
Neat, thank you! Some nits, and CI is failing. :sweat_smile:

map_test.go
```diff
@@ -433,6 +433,32 @@ func TestMapPinWithEmptyPath(t *testing.T) {
 	c.Assert(err, qt.Not(qt.IsNil))
 }
 
+func TestMapPinReplace(t *testing.T) {
```
ti-mo, line 436: Could you add a quick comment here describing which error case is being tested? It's not immediately clear what this code is supposed to do, and the name of the test isn't descriptive. (perhaps adding a `qt.Commentf()` to the final assertion might help)

pinning.go
aditighag: Ah! I only ran the relevant TestMap* and TestProg* tests locally. Will fix it! Thank you.

aditighag:
Thanks for the review. Agree with your comments; the switch to `renameat2` does warrant some explanation in the code as well.

ti-mo (approved):
:rocket:

internal/unix/types_other.go
```diff
@@ -54,6 +54,8 @@ const (
 	BPF_STATS_RUN_TIME       = 0
 	PERF_RECORD_LOST         = 2
 	PERF_RECORD_SAMPLE       = 9
+	AT_FDCWD                 = -0x2
```
lmb, line 57: This is fine, in the future you can just use `0`.

lmb (approved):

lmb:
> Note to reviewers: Do we want to expose an API that gives user the option to instead replace the new pinned path?

Yeah, this might make sense. The problem is: how do we have the same behaviour when the map isn't pinned already? BPF_OBJ_PIN refuses to override, so we'd have to pin to a temporary location and then rename. It all gets pretty complex.

PR #272, opened 2021-04-06
link: set a finalizer on perfEvent

lmb:
Try to clean up perfEvent even if it's leaked by setting a finalizer
on it.

### DIFF
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"unsafe"
@@ -161,6 +162,8 @@ func (pe *perfEvent) attach(prog *ebpf.Program) error {
 		return fmt.Errorf("enable perf event: %s", err)
 	}
 
+	// Close the perf event when its reference is lost to avoid leaking system resources.
+	runtime.SetFinalizer(pe, (*perfEvent).Close)
 	return nil
 }
 
--- a/link/tracepoint.go
+++ b/link/tracepoint.go
@@ -48,6 +48,7 @@ func Tracepoint(group, name string, prog *ebpf.Program) (Link, error) {
 	}
 
 	if err := pe.attach(prog); err != nil {
+		pe.Close()
 		return nil, err
 	}
 


link/tracepoint.go
```diff
@@ -48,6 +48,7 @@ func Tracepoint(group, name string, prog *ebpf.Program) (Link, error) {
 	}
 
 	if err := pe.attach(prog); err != nil {
+		pe.Close()
```
ti-mo, line 51: Nice catch!

link/perf_event.go
```diff
@@ -157,6 +158,7 @@ func (pe *perfEvent) attach(prog *ebpf.Program) error {
 		return fmt.Errorf("enable perf event: %s", err)
 	}
 
+	runtime.SetFinalizer(pe, (*perfEvent).Close)
```
ti-mo, line 161: This needs to be considered fired as soon as the caller loses the pe reference, so IMO it should be well-documented on the API. This might fall under `Failure to call Close() is an error`, since it would imply keeping a reference alive until shutdown (be it through `defer` or other means).

As a user though, I would find this finalizer behaviour quite surprising (not to mention difficult to debug), as it disconnects my probe when I lose the pe reference, which is not something I asked for. This approach might be a bit too clever, and I would consider probes mysteriously disconnecting to be the worse of two evils compared to system resources piling up (which will only happen on pre-4.17).

ti-mo (commented):
Not sure if this is the right approach.

Also just realized I forgot to add the following to `Tracepoint()`:

```go
// The resulting Link must be Closed during program shutdown to avoid leaking
// system resources.
```

Could we add that in this PR?

link/perf_event.go
lmb: > As a user though, I would find this finalizer behaviour quite surprising (not to mention difficult to debug), as it disconnects my probe when I lose the pe reference, which is not something I asked for. 

This already happens with "real" bpf_link: if you don't pin the link, leaking it will invoke internal.FD.Close via a finalizer. That in turn detaches the program. The user experience here is pretty poor, but I'm pointing this out since I believe that closing a leaked fd is the right thing to do.
> This approach might be a bit too clever [...]

I see your point, I was chasing a bug due to an interaction of this + prog arrays having silly semantics before. I'm not sure there is a really good choice.

This raises some questions:
* What happens when a PMU fd is closed? Is it similar to bpf_link, or does the event stay attached? If it's like bpf_link we already have this problem, so adding the finalizer doesn't make things worse (just consistently worse).
* Can leaking the tracefs group lead to other applications malfunctioning? This would be a bad outcome, and another reason to add a finalizer.

ti-mo: > What happens when a PMU fd is closed?

Some experimenting with the examples:

```
2021/04/06 12:09:37 __x64_sys_execve called 14 times
2021/04/06 12:09:38 __x64_sys_execve called 56 times
2021/04/06 12:09:38 closed kp
2021/04/06 12:09:39 __x64_sys_execve called 56 times
2021/04/06 12:09:40 __x64_sys_execve called 56 times
```

Closing a PMU removes the function hook and releases the underlying prog_array. I remember this behaviour being the same with tracefs-based probes. This is the canonical way of disconnecting/removing probes, so this behaviour is expected. I would expect uprobe and kprobe PMU to behave the same way.

(switching gears) For **uprobes**, we've had someone on Slack reporting that writing a close operation to `<tracefs>/uprobe_events` returns EBUSY if a child `perfEvent` is still open, so we need to make sure we unwind those in the correct order.

An argument in favor of setting a finalizer would be consistent behaviour across dev and production deployments. Omitting Close() would accumulate resources over time, with attach operations becoming incrementally slower. A disconnected probe due to a GC cycle (running every x milliseconds in a freshly-started Go program) should manifest much sooner, perhaps even during test runs.

> Can leaking the tracefs group lead to other applications malfunctioning?

What do you mean by this?

link/perf_event.go
```diff
@@ -157,6 +158,7 @@ func (pe *perfEvent) attach(prog *ebpf.Program) error {
 		return fmt.Errorf("enable perf event: %s", err)
 	}
```
ti-mo, line 160:
```suggestion
// Close the perf event when its reference is lost to avoid leaking system resources.
```

link/perf_event.go
lmb: > Closing a PMU removes the function hook and releases the underlying prog_array.

That tells me adding a finalizer that clears up the tracefs group doesn't make things any worse. WDYT?

> Can leaking the tracefs group lead to other applications malfunctioning?

Consider an application written with cilium/ebpf that leaks lots of tracefs based kprobe attachments. Run this on a system for a while. Will other applications trying to use tracefs kprobes suffer from slow downs or receive an error because of the leaky app? If the answer is yes, it is another argument in favour of adding a finalizer.

ti-mo: I was thinking this finalizer might prevent is pinning a kprobe program throughout a service restart. (more specifically, keeping the perf event open so events continue to be delivered) With tracefs, this would be possible if the perfEvent can be serialized to a tmpfs somewhere. Though I can't think of a way to extend the lifetime of a PMU perf event beyond the lifetime of the program, so this might be a moot point.

PR #274, opened 2021-04-07
Detect kernel version to attach kprobes to pre-5.0 kernels

ti-mo:
Before kernel 5.0, kprobes can only attached by providing the correct 'version' argument to BPF_PROG_LOAD. This argument must contain the value calculated during kernel compile time using the `KERNEL_VERSION` macro.

This implementation takes a slightly different approach to detecting the kernel version than other libraries, but I think it's more concise and readable, while also being less fragile. Most importantly, there's nothing explicitly distro-specific about it.

- check `/proc/version_signature` (only exists on Ubuntu kernels, but is the only source of truth that contains the patch level)
- read `uname -v`
- read `uname -r`

Against each source, run a regex that matches x.y(.z) that appear at the start of the string or right after a space. If a source yields a usable version number, return the version and halt the search. The result is cached in package `internal`, so only one lookup is performed over the lifetime of the program.

Best reviewed per commit, as some code was moved around.

@brycekahle thanks for the reminder that this was still needed and for pointing out that the patch level can simply be missing on some kernels.

cc @jrfastab


internal/version_test.go
```diff
@@ -29,4 +31,10 @@ func TestVersion(t *testing.T) {
 	if v200.Less(a) {
 		t.Error("2.0.0 should not be less than 1.2.1")
 	}
+
+	// Kernels 4.4 and 4.9 have a SUBLEVEL of over 255 and clamp it to 255.
+	// In our implementation, the other version segments are truncated.
+	if v := (Version{256, 256, 256}); v.Kernel() != 255 {
```
lmb, line 37: Nit: test a known good version like `5, 4, 1` or similar?

link/kprobe.go
```diff
@@ -234,10 +234,11 @@ func createTraceFSKprobeEvent(group, symbol string, ret bool) error {
...
 	_, err = f.WriteString(pe)
-	// Writing to <tracefs>/kprobe_events will return ENOENT when the tracee
-	// kernel symbol does not exist (yet) in the kernel.
-	if os.IsNotExist(err) {
-		return fmt.Errorf("kernel symbol %s not found: %w", symbol, ErrNotSupported)
+	// Since commit 97c753e62e6c, ENOENT is correctly returned instead of EINVAL
+	// when trying to create a kretprobe for a missing symbol. Make sure ENOENT
+	// is returned to the caller.
+	if errors.Is(err, os.ErrNotExist) || errors.Is(err, unix.EINVAL) {
+		return fmt.Errorf("kernel symbol %s not found: %w", symbol, os.ErrNotExist)
```
lmb, line 241: This makes me think that there is a bug in `kprobe()`: it checks for ErrNotSupported from `pmuKprobe()` but that function never returns ErrNotSupported?

lmb (approved):

internal/version_test.go
ti-mo: Sure, will do!

link/kprobe.go
ti-mo: `getPMUEventType()` called from `pmuKprobe()` returns `ErrNotSupported`.

PR #276, opened 2021-04-08
examples: update cilium/ebpf dependency to 0.5.0

ti-mo:


lmb (approved):

lmb:
Maybe we can get rid of this now that we don't need go-perf anymore?

ti-mo:
> Maybe we can get rid of this now that we don't need go-perf anymore?

What do you mean? The uprobe example still uses go-perf. Creating an issue for adding a uprobe link.

PR #278, opened 2021-04-09
Increase coverage from libbpf selftests

### LINKED ISSUE #254: Support passing a custom BTF file
While using the library I realized that the kernel which I have on my host (`5.4.0-1037-aws`) does not have BTF support enabled out of the box. Would be good the the library to support loading an out of band compiled BTF file if the kernel does not come with one. While talking with @lmb on Slack one suggestion is to plumb this through `CollectionOptions`.

lmb:
elf: increase coverage of libbpf integration test
    
    The libbpf integration tests currently only test the ELF loader, since
    we never attempt to load the collection into the kernel. This leaves
    out a good chunck of code, e.g. CO-RE relocation handling.
    
    Attempt to load kernel selftests into the kernel. Since we don't want
    to replicate the user space portion of the selftests we have to be
    quite lax in what we consider an error. The following cases lead
    to a skipped test:
    
    * ErrIsNotSupported indicates that we've hit a known deficiency in the
      library.
    * syscall.Errno indicates that we've made it all the way up to loading
      a map or program, but don't have the necessary set up code.
    
    All other errors are considered bugs.

program: allow specifying a custom target for CO-RE relocations
    
    We need a target BTF when calculating CO-RE relocations for a program.
    By default this is the BTF for the current kernel. This is too inflexible:
    
    Sometimes the kernel doesn't have BTF and so the BTF has to come from a
    separate file. The library might also be used in a container where the
    canonical paths aren't available.
    
    Furthermore, some programs actually relocate against themselves for
    testing purposes.
    
    Allow specifing an io.ReaderAt in ProgramOptions which is used as the
    target BTF for CO-RE relocations.
    
Fixes #254

internal: allow unwrapping VerifierError
    
    VerifierError hides the underlying syscall error from inspection.
    Implement Unwrap() so that callers can check for syscall.Errno.

program: allow empty License
    
    The kernel has a requirement to provide a license string, but the
    string itself can be empty. Allow empty License to be able to load
    a bunch of kernel selftests.

map: return ErrNotSupported for unknown pin types

linker: don't fix up instructions without a Reference
    
    The kernel selftests contain the follwing assembly in test_get_stack_rawtp_err.o:
    
        ;   while (1) {
        6:  05 00 ff ff 00 00 00 00 goto -1 <LBB0_1>
    
    which is generated for the following C snippet:
    
        while (1) {
            error++;
        }
    
    Basically, the infinite loop is turned into a jump to -1, aka the
    jump instruction itself. The linker currently interprets the -1 to
    mean "requires relocation" and so gives a somewhat cryptic error.
    
    Stop trying to fix up jumps and calls if they don't have a refernce.

### DIFF
--- a/elf_reader_test.go
+++ b/elf_reader_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 
 	"github.com/cilium/ebpf/internal"
@@ -119,7 +120,7 @@ func TestLoadCollectionSpec(t *testing.T) {
 		}),
 	)
 
-	testutils.TestFiles(t, "testdata/loader-*.elf", func(t *testing.T, file string) {
+	testutils.Files(t, testutils.Glob(t, "testdata/loader-*.elf"), func(t *testing.T, file string) {
 		have, err := LoadCollectionSpec(file)
 		if err != nil {
 			t.Fatal("Can't parse ELF:", err)
@@ -208,7 +209,7 @@ func TestCollectionSpecDetach(t *testing.T) {
 }
 
 func TestLoadInvalidMap(t *testing.T) {
-	testutils.TestFiles(t, "testdata/invalid_map-*.elf", func(t *testing.T, file string) {
+	testutils.Files(t, testutils.Glob(t, "testdata/invalid_map-*.elf"), func(t *testing.T, file string) {
 		_, err := LoadCollectionSpec(file)
 		t.Log(err)
 		if err == nil {
@@ -218,7 +219,7 @@ func TestLoadInvalidMap(t *testing.T) {
 }
 
 func TestLoadInvalidMapMissingSymbol(t *testing.T) {
-	testutils.TestFiles(t, "testdata/invalid_map_static-el.elf", func(t *testing.T, file string) {
+	testutils.Files(t, testutils.Glob(t, "testdata/invalid_map_static-el.elf"), func(t *testing.T, file string) {
 		_, err := LoadCollectionSpec(file)
 		t.Log(err)
 		if err == nil {
@@ -228,7 +229,7 @@ func TestLoadInvalidMapMissingSymbol(t *testing.T) {
 }
 
 func TestLoadInitializedBTFMap(t *testing.T) {
-	testutils.TestFiles(t, "testdata/initialized_btf_map-*.elf", func(t *testing.T, file string) {
+	testutils.Files(t, testutils.Glob(t, "testdata/initialized_btf_map-*.elf"), func(t *testing.T, file string) {
 		_, err := LoadCollectionSpec(file)
 		t.Log(err)
 		if !errors.Is(err, internal.ErrNotSupported) {
@@ -238,7 +239,7 @@ func TestLoadInitializedBTFMap(t *testing.T) {
 }
 
 func TestStringSection(t *testing.T) {
-	testutils.TestFiles(t, "testdata/strings-*.elf", func(t *testing.T, file string) {
+	testutils.Files(t, testutils.Glob(t, "testdata/strings-*.elf"), func(t *testing.T, file string) {
 		_, err := LoadCollectionSpec(file)
 		t.Log(err)
 		if !errors.Is(err, ErrNotSupported) {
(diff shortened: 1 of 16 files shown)


internal/btf/btf.go
```diff
@@ -670,11 +670,18 @@ func ProgramLineInfos(s *Program) (recordSize uint32, bytes []byte, err error) {
...
 // This is a free function instead of a method to hide it from users
 // of package ebpf.
-func ProgramRelocations(s *Program, target *Spec) (map[uint64]Relocation, error) {
+func ProgramRelocations(s *Program, target *Spec) (_ map[uint64]Relocation, err error) {
 	if len(s.coreRelos) == 0 {
 		return nil, nil
 	}
 
+	if target == nil {
+		target, err = LoadKernelSpec()
```
ti-mo, line 679: Nit: Does `target` get shadowed if we use `:=` here? It's probably worth declaring `var err error` over using a `_` named return value for readability's sake.

prog.go
```diff
@@ -41,6 +42,9 @@ type ProgramOptions struct {
 	// Controls the output buffer size for the verifier. Defaults to
 	// DefaultVerifierLogSize.
 	LogSize int
+	// The ELF file containing the BTF to relocate against for CO-RE. May be
+	// nil, which implies relocating against the kernel.
```
ti-mo, line 46: Is there an alternative description we could give instead of `relocating against the kernel`. Perhaps something that explicitly indicates that setting`TargetBTF` overrides the default kimg scanning routine, or that it might apply to containerized binaries? Taking some parts of the commit message would improve this. :+1:

ti-mo (approved):
Awesome! Some nits, and CI is failing. :grin:

prog.go
lmb: I reworded it a bit, hopefully it's better now.

lmb:
This is turning into a bigger saga, there is a memory leak somewhere that I need to track down.

prog.go
ti-mo: Perfect! :ok_hand:

ti-mo:
@lmb Would it make sense to split it up into smaller PRs?

lmb:
I moved #287 out of this PR, I'll update once it's merged.

internal/testutils/rlimit.go
```diff
@@ -0,0 +1,20 @@
+package testutils
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+func init() {
```
lmb, line 10: @ti-mo this changed since your last review. I generally don't like to use `init` but it seems appropriate here. What do you think?

lmb:
Sorry for all the noise, the tests seem somewhat stable now. I've not been able to get to the root cause of this problem, but I think it's a mixture of:
* Heavy allocation pressure from the BPF verifier, which repeatedly calls kmalloc / kfree
* Go compiling / running multiple tests at once, which exhausts CPU

My theory is that the kfree are actually deferred work which isn't getting done due to the go test being greedy with CPU. I don't have any real proof though.

elf_reader_test.go
```diff
@@ -290,29 +291,116 @@ func TestLibBPFCompat(t *testing.T) {
...
 		switch file {
-		case "test_ksyms.o":
-			// Issue #114
-			t.Skip("Kernel symbols not supported")
 		case "test_sk_assign.o":
-			t.Skip("Incompatible struct bpf_map_def")
-		case "test_ringbuf_multi.o", "map_ptr_kern.o", "test_btf_map_in_map.o":
-			// Issue #155
-			t.Skip("BTF .values not supported")
+			t.Skip("Skipping due to incompatible struct bpf_map_def")
```
ti-mo, line 328: Note: this incompatible struct is actually called `struct bpf_elf_map`. I've investigated ways of supporting this legacy struct before, and I still think it might be worthwhile to at least be able to parse this struct successfully and exposing the trailing bits to the caller, if not just supporting it natively out of the box.

ti-mo (approved):

elf_reader_test.go
ti-mo: On second thought, this might put on more code complexity than I initially assumed. We should fix this string, though. :point_up:

PR #281, opened 2021-04-11
Update README.md for how to run examples

nevermosby:
Hi all,

The commit is to provide the helpful guidelines on how to run the examples.

Signed-off-by: Wen-Quan Li <legendarilylwq@gmail.com>

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -4,3 +4,11 @@
 - [uprobe](uprobe/) - Like a kprobe, but for symbols in userspace binaries (e.g. `bash`).
 - [tracepoint](tracepoint/) - Attach a program to predetermined kernel tracepoints.
 - Add your use case(s) here!
+
+## How to run
+
+```bash
+cd ebpf/examples/
+go run -exec sudo [./kprobe,./uprobe,./tracepoint...]
+
+```


examples/README.md
```diff
@@ -4,3 +4,14 @@
 - [uprobe](uprobe/) - Like a kprobe, but for symbols in userspace binaries (e.g. `bash`).
 - [tracepoint](tracepoint/) - Attach a program to predetermined kernel tracepoints.
 - Add your use case(s) here!
+
+## How to run
+
+```bash
+git clone https://github.com/cilium/ebpf.git
+cd examples/[kprobe,uprobe,tracepoint...]
```
lmb, line 12: You can simplify this to:

```suggestion
cd ebpf/examples/
go run -exec sudo [kprobe,uprobe,tracepoint...]
```

No go build step needed, so you can skip the the bit about `go build main.go`.

ti-mo (commented):
Why don't we just add a `make examples` target to the main Makefile instead, and document that here? Not sure if this repo is the place to teach a user about `go build`... :thinking:

lmb:
> Why don't we just add a `make examples` target to the main Makefile instead, and document that here? Not sure if this repo is the place to teach a user about `go build`... thinking

I'd rather not bloat the Makefile more. Users are probably interested in individual examples anyways, how do you run those via the makefile?

Documenting the `go run -exec sudo` trick is worth it IMO.

ti-mo:
> Users are probably interested in individual examples anyways, how do you run those via the makefile?

Oh, I would just suggest `make examples` to build them, not run them. That would leave runnable binaries in the example dirs.

I hadn't used `go run` that way before, I thought it had to be called on a single file. So what about `go run -exec sudo .` in the individual example dirs then instead? That doesn't leave a binary behind in the directory.

`go build .` is functionally equivalent to `go build`, so I would omit the dot in case we go for that.

examples/README.md
```diff
@@ -4,3 +4,14 @@
 - [uprobe](uprobe/) - Like a kprobe, but for symbols in userspace binaries (e.g. `bash`).
 - [tracepoint](tracepoint/) - Attach a program to predetermined kernel tracepoints.
 - Add your use case(s) here!
+
+## How to run
+
+```bash
+git clone https://github.com/cilium/ebpf.git
```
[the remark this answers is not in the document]
nevermosby: Agreed, will remove it.

examples/README.md
```diff
@@ -4,3 +4,14 @@
...
+cd examples/[kprobe,uprobe,tracepoint...]
+go build . # You will get the executable file and just run it.
+
+# Do not run `go build main.go` directly for the examples of kprobe and uprobe, otherwise you will get errors like "undefined: KProbeExampleObjects".
```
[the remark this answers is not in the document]
nevermosby: It does make sense, I will remove it.

examples/README.md
nevermosby: I tried "go run -exec sudo kprobe" but got errors: `package kprobe is not in GOROOT (/snap/go/7360/src/kprobe)`. And I tried "go run -exec sudo ./kprobe" and it works. So I will document that instead.

nevermosby:
Just updated based on your suggestions, pls help review.

examples/README.md
ti-mo: Thanks!

PR #285, opened 2021-04-14
run-test.sh cleanups

lmb:
run-tests: run a custom command in the vm
    
    Let the user specify the command to execute in the VM. This is useful
    if you want to run only a subset of tests, for example.

run-tests: use go mod tidy
    
    go mod download for some reason modifies go.mod. According to an upstream
    Golang issue "go mod tidy" is the preferred command to prime the module
    cache.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -32,7 +32,7 @@ fi
 
 # Pull all dependencies, so that we can run tests without the
 # vm having network access.
-go mod download
+go mod tidy
 
 # Use sudo if /dev/kvm isn't accessible by the current user.
 sudo=""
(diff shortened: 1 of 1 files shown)


run-tests.sh
```diff
@@ -68,22 +70,28 @@ else
   echo "No selftests found, disabling"
 fi
 
+if (( $# > 0 )); then
```
ti-mo, line 73: This is a bit uncommon in Bash, what about `if [ "$#" -ne 0 ]; then` instead?

ti-mo (approved):
Nice!

run-tests.sh
lmb: That's what I had at some point in the past, but a colleague asked me to use `(())` instead. Some people say that `[` should be deprecated so :shrug: I'll just keep it as it is.

PR #291, opened 2021-04-21
map: define new ErrMapMismatch error

### LINKED ISSUE #289: Delete and recreate pinned maps in case of mismatch with the map's spec
To address https://github.com/cilium/cilium/issues/15629 I need to delete and recreate a pinned map in case there's a mismatch with the map's spec and the actual pinned map (type, k/v size, flags etc).

I see 2 approaches :
* have this logic in `cilium/ebpf` by introducing a new option in `MapOptions.LoadPinOptions`. If there's a mismatch and the option is set, `NewMapWithOptions` will delete the existing map and try to recreate it.
Code changes are small but I'm not sure if this logic would be too specialized/high level for the package.
* have this logic in `cilium/cilium`. The caller of `ebpf.NewMapWithOptions` would be responsible for detecting the mismatch, deleting the map and recreating it.
My concern here would be on _detecting_ the mismatch, as the best I could come up with is:
```go
newMap, err := ebpf.NewMapWithOptions(spec, opts)
if err != nil {
    if strings.HasPrefix(err.Error(), "use pinned map ") {
        // delete the map and call again NewMapWithOptions
```
which to me looks fragile :confused:

jibi:
This error will be returned by NewMapWithOptions whether the spec for a
pinned map is incompatible with the actual instance of the map, in order
to allow the caller to properly detect this mismatch.

Fixes #289

Signed-off-by: Gilberto Bertin <gilberto@isovalent.com>

### DIFF
--- a/map.go
+++ b/map.go
@@ -18,6 +18,7 @@ var (
 	ErrKeyNotExist      = errors.New("key does not exist")
 	ErrKeyExist         = errors.New("key already exists")
 	ErrIterationAborted = errors.New("iteration aborted")
+	ErrMapIncompatible  = errors.New("map's spec is incompatible with pinned map")
 )
 
 // MapOptions control loading a map into the kernel.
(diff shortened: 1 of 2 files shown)


map.go
```diff
@@ -20,6 +20,18 @@ var (
...
+type ErrMapIncompatible struct {
+	mapName string
+}
+
+func NewErrMapIncompatible(mapName string) ErrMapIncompatible {
+	return ErrMapIncompatible{mapName}
+}
+
+func (e ErrMapIncompatible) Error() string {
+	return fmt.Sprintf("incompatible map's spec for pinned map %s", e.mapName)
```
jibi, line 32: I slightly changed the error message to try make it more explicit (c.f. https://github.com/cilium/cilium/issues/15629)

map_test.go
```diff
@@ -1315,6 +1315,16 @@ func TestMapPinning(t *testing.T) {
...
 		t.Fatal("Pinning doesn't use pinned maps")
 	}
+
+	spec.KeySize = 8
+	m3, err := NewMapWithOptions(spec, MapOptions{PinPath: tmp})
+	if err == nil {
+		m3.Close()
+		t.Fatalf("Opening a pinned map with a mismatching spec did not fail")
+	}
+	if !errors.Is(err, NewErrMapIncompatible("test")) {
```
lmb, line 1325: If you keep the separate error type this has to become `errors.As` which isn't as ergonomic:

```
var *err ErrMapIncompatible
if errors.As(err, &err) {
   ...
}
```

map.go
```diff
@@ -202,7 +214,7 @@ func newMapWithOptions(spec *MapSpec, opts MapOptions, btfs btfHandleCache) (_ *
 		defer closeOnError(m)
 
 		if err := spec.checkCompatibility(m); err != nil {
-			return nil, fmt.Errorf("use pinned map %s: %s", spec.Name, err)
+			return nil, fmt.Errorf("%w: %s", NewErrMapIncompatible(spec.Name), err)
```
lmb, line 217: This doesn't work because `%w` has to be at the end of fmt.Errorf if I remember correctly. My suggestion, stick to the simple exported error `var ErrMapIncompatible = ...`.

```suggestion
			return nil, fmt.Errorf("use pinned map %s: %w", spec.Name, err)
```

In checkCompatibility you can then add additional information using the same trick `return fmt.Errorf("details of what is wrong %d: %w", ErrMapIncompatible)`.

lmb (changes requested):

lmb (approved):

lmb:
Can you add documentation to NewMapWithOptions that indicates it can return ErrMapIncompatible? Feel free to merge afterwards.

ti-mo:
@jibi thanks!

PR #294, opened 2021-04-23
CI: faster please

lmb:
CI: compile on the host, run in the VM
    
    Split run-tests.sh into three distinct stages:
    
    1. Prep work like downloading kernel and selftests
    2. Booting a VM for each package to be tested
    3. Executing the test binary for each package in the VM
    
    This allows us to move compiling from the VM to the host.
    Hat tip to timo for suggesting this.

CI: move go generate into lint section
    
    There is no need run go generate in the prologue. Also include modules,
    which call bpf2go now. Since they are in a separate package we need
    to change into the examples directory before we can generate.

CI: don't run bpf2go compile tests in the VM
    
    It turns out that TestRun in the bpf2go package is really expensive in
    the VM. This is most likely because the test invokes the Go toolchain
    to ensure that the generated code is valid. Move the test out of the
    VM and only run it in a single step on CI.


lmb:
This drops CI run time by about half I'd say. Without this it'll be hard to land #278.

.gitignore
```diff
@@ -11,3 +11,4 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+coverage.txt
```
ti-mo, line 14: Nit: this is called `coverage.out` in most places (like https://blog.golang.org/cover)

ti-mo (approved):
Nice, thanks! While we're at it, could we somehow retain `coverage.txt` (or `.out`)? I always add a temporary `cp` in there somewhere. It's fine if parallel executions of the script clobber each others' coverage outputs, it's for development only.

lmb:
coverage.out is retained now.

Issue #292, opened 2021-04-22
pmu uprobes are broken


lmb:
@ti-mo @lmb FYI pmu uprobes are broken; they were working before doing some refactoring. I suppose it's the offset not being passed correctly to `perf_event_open`:

```
$ sudo bpftool perf
pid 1056229  fd 3: prog_id 712  uretprobe  filename /bin/bash  offset 0
```
I am continuing to debug.

PS. currently I believe there aren't "e2e" tests, are these planned?

_Originally posted by @mmat11 in https://github.com/cilium/ebpf/issues/280#issuecomment-824230882_

ti-mo:
@ungureanuvladvictor pointed me to the fact that the uprobe example works on Ubuntu's build of Bash (including 21.04), and I found that Ubuntu ships with a statically-linked version of `readline`. It looks like they vendor a version of readline in the Bash tree, but it can optionally be linked against `libreadline.so`:
```
~ objdump -T /bin/bash
00000000000ad900 g    DF .text  0000000000000088  Base        readline
```

On my system (Arch), readline is provided by `libreadline.so`:
```
λ  ~  objdump -T /bin/bash | grep readline
0000000000000000      DF *UND*	0000000000000000              readline
```

The 0 offset as parsed by `debug/elf` is correct, this symbol is to be provided by a library and still needs to be relocated. This means we'll either need to find or implement a dynamic linker in Go, and allow the caller to pass in their own offsets for now. Will close this with a PR to disable creating Uprobes from symbols from SHN_UNDEF with offset 0.

PR #296, opened 2021-04-27
link: uprobe - mark shared library symbols as unsupported

ti-mo:
Help the user by returning an error when attempting to set a uprobe on a symbol provided by a shared object (library).

Bash's `readline` implementation can be a library call depending on the distribution. Use main instead.

Closes https://github.com/cilium/ebpf/issues/292, but needs a follow-up that allows the caller to specify an offset manually.

@lmb @mmat11

### DIFF
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -92,13 +92,14 @@ func (ex *Executable) symbol(symbol string) (*elf.Symbol, error) {
 
 // Uprobe attaches the given eBPF program to a perf event that fires when the
 // given symbol starts executing in the given Executable.
-// For example, /bin/bash::readline():
+// For example, /bin/bash::main():
 //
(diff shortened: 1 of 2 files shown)


link/uprobe.go
```diff
@@ -149,6 +149,13 @@ func (ex *Executable) uprobe(symbol string, prog *ebpf.Program, ret bool) (*perf
 		return nil, fmt.Errorf("symbol '%s' not found in '%s': %w", symbol, ex.path, err)
 	}
 
+	// Symbols with location 0 from section undef are shared library calls and
```
lmb, line 152: I think this is worth documenting on the Uprobe function as well.

lmb (approved):

link/uprobe.go
ti-mo: Good point, done :+1:

PR #300, opened 2021-05-10
elf: clamp PerfEventArray.MaxEntries to possible CPUs

### LINKED ISSUE #209: perf event arrays should be useable out of the box
Using a perf event array with a `perf.Reader` is error prone at the moment, since you need to declare your PERF_EVENT_ARRAY with max_entries == 0 so that the library figures out the # of possible CPUs. This has bitten at least @tqbf and @hariharan-a.

The library used to reject PERF_EVENT_ARRAY that didn't have max_entries == 0, but that broke for people trying to load kernel selftests.

Maybe we can find a behaviour that makes the common case work out of the box. We can check what libbpf does. If they don't have anything sensible either we can at least improve the documentation.

lmb:
PerfEventArray maps are somewhat peculiar, since there is no reasonable value
that can be used for max_entries without library support. This is because the
map is usually (but not always) sized according to the number of possible
CPUs on the system.

Follow the behaviour of libbpf and clamp MapSpec.MaxEntries to the number of
possible CPUs for maps defined in ELF files. This allows map definitions in
the ELF that have "nonsensical" numbers of entries:

    struct bpf_map_def perf_event_array __section("maps") = {
        .type        = BPF_MAP_TYPE_PERF_EVENT_ARRAY,
        .key_size    = sizeof(uint32_t),
        .value_size  = sizeof(uint32_t),
        .max_entries = 4096,
    };

The safest option is still to set max_entries = 0.

Fixes #209


map.go
```diff
@@ -88,6 +88,23 @@ func (ms *MapSpec) Copy() *MapSpec {
 	return &cpy
 }
 
+func (ms *MapSpec) clampPerfEventArraySize() error {
```
ti-mo, line 91: Would it be worth documenting here why clamping is necessary?

PR #301, opened 2021-05-11
CO-RE support

lmb:
This should make CO-RE useable. Bitfield relocations are not supported yet.

btf: add CO-RE field relocations
    
    Field relocations record the path through a nested structure. They
    are generated by the following macros:
    
        bpf_core_field_exists(foo->bar)
        bpf_core_field_size(foo->bar)
    
    BPF_FIELD_BYTE_OFFSET is currently not generated directly. To resolve
    such relocations we follow the accessor through the (possibly) nested
    array, struct or union.
    
    For arrays we use the accessor as the index into the array, after
    checking that the index is within bounds. The bounds check is not
    performed for  flexible arrays at the end of a structure or union:
    
        struct flex {
            int bar;
            char flex[];
        };
    
    For structs and unions we use the accessor to look up the field in
    the local type and then try to find a field with the same name in
    the target type. Anonymous unions and structs are ignored for the
    purpose of relocation:
    
        struct bar {
            int bar;
            union {
                int foo;
            };
        };
    
        // equivalent to
    
        struct bar {
            int bar;
            int foo;
        };
    
    Updates #114

btf: add CO-RE type relocations
    
    Updates #114

btf: add CO-RE enum relocations
    
    Support relocating enum values. The relocation happens by comparing the
    (essential) name of the local enum value with the name of the target
    enum. If the names match the target value is used. If no match is found,
    the relocation returns 0.
    
    Updates #114

btf: apply CO-RE fixups
    
    Applying a CO-RE fixup is fairly straighforward and is based on
    the instruction that is being fixed up. The most complicated bit
    is poisoning an instruction when there is no valid fixup. Since
    we have to poison double-wide instructions aka 64 bit constant
    loads we can't just take a single fixup and apply it to a single
    instruction. Instead we take a set of fixups and apply them to
    a list of instructions aka a whole BPF program.
    
    Updates #114

btf: core: calculate fixups per type, not per relocation
    
    Currently, we take each CO-RE relocation and try to apply it to each
    possible target type. This approach is simple but leads to a problem
    when relocating code like this:
    
        struct foo;
        core_reloc(foo.a);
        core_reloc(foo.b);
    
    against a target with types like this:
    
       struct foo___first { int a; }
       struct foo___second { int b; }
    
    With the current approach we will use foo___first to access foo.a,
    foo___second to access foo.b. Instead we should choose a single
    type that is constrained by all the relocations present.
    
    To implement this, take each possible target type and try to
    apply all relocations to it. The result is a list of fixups for each
    target, which is scored by counting the number of poisoned fixups
    per type. If a check for the existence of an enum value, type or
    field is negative we also increase the score. Finally, we choose
    the target type with the lowest score.
    
    This way all relocations for one local type ID use the same target
    type, while maximising the code paths that can be executed
    successfully (since we choose the type with the least poisinings).

btf: parse CO-RE accessors early

btf: return sentinel error instead of bool

btf: strip qualifiers before CO-RE relocating
    
    CO-RE relocations ignore all qualifiers like const, volatile, etc. as
    well as typedefs. So far we've dealt with this by sprinkling calls
    to skipQualifierAndTypedef throughout the code base. It's incredibly
    easy to forget such a call, so having good test coverage is important.
    Trying to add coverage just hugely bloats the test suite however.
    
    Instead, strip out all qualifiers and typedef before performing CO-RE
    relocation.


lmb:
Thank you @brycekahle for starting this work in #141!

Issue #266, opened 2021-03-25
aarch64 / arm64 support


shrkamat:
I see failure in running [kprobe example](https://github.com/cilium/ebpf/tree/master/examples/kprobe) for aarch64 kernel version `4.19.113`

```
GOOS=linux GOARCH=arm64 go build -o ebpf-kprobe

./ebpf-kprobe 
2021/03/25 22:03:40 error while loading objects: field KprobeExecve: program kprobe_execve: load program without BTF: invalid argument
```

looks like `BPF_PROG_LOAD` is [failing](https://gist.github.com/shrkamat/d24a81505833391f95cbe81f2c865c1a#file-epbf-kprobe-fail-L406)

lmb:
We don't have arm64 on CI, so the support is best effort. I think the real cause of the error is that loading BTF is bailing out for some reason. Without access to arm64 it'll be hard to debug.

ti-mo:
Thanks for reporting. I have some Pi4's laying around that I still need to set up, but we'll likely only be able to support this properly once we can build and test on arm64 CI kernels. There's some experimental arm64 support in virtme, but haven't tried it yet. Might be on the roadmap at some point, will leave this open to keep track.

lmb:
FWIW I suspect that this particular error is because `./ebpf-kprobe` was run without root privileged.

ti-mo:
Reopening as #304 was only kprobe-related.

ti-mo:
Finally had the chance to look at this. Have ALARM set up on a rpi4, and this is the test output: https://gist.github.com/ti-mo/033e7de02d58da9abc6a2bac91458291

Everything seems to work for the most part, but it seems like the following map/prog types just don't work on this kernel build:
```
--- FAIL: TestHaveMapType (0.10s)
    --- FAIL: TestHaveMapType/XSKMap (0.00s)
        map_test.go:64: Map type XSKMap isn't supported even though kernel is at least 4.18: not supported
    --- FAIL: TestHaveMapType/InodeStorage (0.00s)
        map_test.go:64: Map type InodeStorage isn't supported even though kernel is at least 5.10: not supported
    --- FAIL: TestHaveMapType/TaskStorage (0.00s)
        map_test.go:64: Map type TaskStorage isn't supported even though kernel is at least 5.11: not supported
--- FAIL: TestHaveProgType (0.00s)
    --- FAIL: TestHaveProgType/RawTracepointWritable (0.00s)
[...]

PR #304, opened 2021-05-14
arch: support arm64

nickolaev:
Adds basic developer support for ARM64, fixed the tests and added some instructions/notes.

Modify `kprobe` to be smart about syscall naming, depending on the platform.
For example, it will take `sys_getpid` and turn it into `__arm64_sys_getpid` when running on an ARM64 host.

Fixes https://github.com/cilium/ebpf/issues/266

Signed-off-by: Nikolay Nikolaev <nicknickolaev@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,12 +57,6 @@ environment to keep the build output stable.
 
 The toolchain image build files are kept in [testdata/docker/](testdata/docker/).
 
-### Regeneration Testdata on non-x86 platforms
-
-Before running `make`, ensure [Docker buildx](https://docs.docker.com/buildx/working-with-buildx/)
-is enabled. Additionally `QEMU user` and `binfmt` should be installed. On a Debian based distribution
-the command to add them is `apt install qemu-user-static binfmt-support`.
-
 ## License
 
 MIT
--- a/README.md
+++ b/README.md
@@ -57,6 +57,12 @@ environment to keep the build output stable.
 
 The toolchain image build files are kept in [testdata/docker/](testdata/docker/).
 
+### Regeneration Testdata on non-x86 platforms
+
+Before running `make`, ensure [Docker buildx](https://docs.docker.com/buildx/working-with-buildx/)
+is enabled. Additionally `QEMU user` and `binfmt` should be installed. On a Debian based distribution
+the command to add them is `apt install qemu-user-static binfmt-support`.
+
 ## License
 
 MIT
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -6,3 +6,5 @@ require (
 	github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2
 	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c
 )
+
+replace github.com/cilium/ebpf => ./..
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -6,5 +6,3 @@ require (
 	github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2
 	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c
 )
-
-replace github.com/cilium/ebpf => ./..
--- a/examples/kprobe/main.go
+++ b/examples/kprobe/main.go
@@ -1,7 +1,7 @@
 // +build linux
 
 // This program demonstrates attaching an eBPF program to a kernel symbol.
-// The eBPF program will be attached to the start of the __x64_sys_execve
+// The eBPF program will be attached to the start of the
 // kernel function and prints out the number of times it has been called
 // every second.
 package main
@@ -24,7 +24,7 @@ const mapKey uint32 = 0
 func main() {
 
 	// Name of the kernel function to trace.
-	fn := "__x64_sys_execve"
+	fn := "sys_execve"
 
 	// Subscribe to signals for terminating the program.
 	stopper := make(chan os.Signal, 1)
--- a/link/perf_event_test.go
+++ b/link/perf_event_test.go
@@ -27,7 +27,7 @@ func TestTraceEventTypePMU(t *testing.T) {
 func TestTraceEventID(t *testing.T) {
 	c := qt.New(t)
 
-	eid, err := getTraceEventID("syscalls", "sys_enter_fork")
+	eid, err := getTraceEventID("syscalls", "sys_enter_execve")
 	c.Assert(err, qt.IsNil)
 	c.Assert(eid, qt.Not(qt.Equals), 0)
 }
--- a/link/platform.go
+++ b/link/platform.go
@@ -8,11 +8,17 @@ import (
 func platformPrefix(symbol string) string {
 
 	prefix := runtime.GOARCH
+
+	// per https://github.com/golang/go/blob/master/src/go/build/syslist.go
 	switch prefix {
-	case "i386":
+	case "386":
 		prefix = "ia32"
-	case "amd64":
+	case "amd64", "amd64p32":
 		prefix = "x64"
+	case "arm64", "arm64be":
+		prefix = "arm64"
+	default:
+		return symbol
 	}
 
 	return fmt.Sprintf("__%s_%s", prefix, symbol)
--- a/link/platform.go
+++ b/link/platform.go
@@ -1,0 +1,19 @@
+package link
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func platformPrefix(symbol string) string {
+
+	prefix := runtime.GOARCH
+	switch prefix {
+	case "i386":
+		prefix = "ia32"
+	case "amd64":
+		prefix = "x64"
+	}
+
+	return fmt.Sprintf("__%s_%s", prefix, symbol)
+}
--- a/link/tracepoint_test.go
+++ b/link/tracepoint_test.go
@@ -79,7 +79,7 @@ func TestTracepointErrors(t *testing.T) {
 }
 
 func TestTraceGetEventID(t *testing.T) {
-	_, err := getTraceEventID("syscalls", "sys_enter_open")
+	_, err := getTraceEventID("syscalls", "sys_enter_openat")
 	if err != nil {
 		t.Fatal("Can't read trace event ID:", err)
 	}
--- a/examples/go.sum
+++ b/examples/go.sum
@@ -1,3 +1,5 @@
+github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2 h1:OmjZ22sFv73tXT+9ntlaZTujJp+4wSMmzX/fSNlq6ew=
+github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2/go.mod h1:4tRaxcgiL706VnOzHOdBlY8IEAIdxINsQBcU4xJJXRs=
 github.com/frankban/quicktest v1.11.3 h1:8sXhOn0uLys67V8EsXLc6eszDs8VXWxL3iRvebPhedY=
 github.com/frankban/quicktest v1.11.3/go.mod h1:wRf/ReqHper53s+kmmSZizM8NamnL3IM0I9ntUbOk+k=
 github.com/google/go-cmp v0.5.4 h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=
--- a/examples/go.sum
+++ b/examples/go.sum
@@ -1,5 +1,3 @@
-github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2 h1:OmjZ22sFv73tXT+9ntlaZTujJp+4wSMmzX/fSNlq6ew=
-github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2/go.mod h1:4tRaxcgiL706VnOzHOdBlY8IEAIdxINsQBcU4xJJXRs=
 github.com/frankban/quicktest v1.11.3 h1:8sXhOn0uLys67V8EsXLc6eszDs8VXWxL3iRvebPhedY=
 github.com/frankban/quicktest v1.11.3/go.mod h1:wRf/ReqHper53s+kmmSZizM8NamnL3IM0I9ntUbOk+k=
 github.com/google/go-cmp v0.5.4 h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=
--- a/examples/kprobe/bpf/kprobe_example.c
+++ b/examples/kprobe/bpf/kprobe_example.c
@@ -10,7 +10,7 @@ struct bpf_map_def SEC("maps") kprobe_map = {
     .max_entries = 1,
 };
 
-SEC("kprobe/__x64_sys_execve")
+SEC("kprobe/sys_execve")
 int kprobe_execve() {
     u32 key = 0;
     u64 initval = 1, *valp;
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -131,7 +131,10 @@ func kprobe(symbol string, prog *ebpf.Program, ret bool) (*perfEvent, error) {
 	}
 
 	// Use kprobe PMU if the kernel has it available.
-	tp, err := pmuKprobe(symbol, ret)
+	tp, err := pmuKprobe(platformPrefix(symbol), ret)
+	if errors.Is(err, os.ErrNotExist) {
+		tp, err = pmuKprobe(symbol, ret)
+	}
 	if err == nil {
 		return tp, nil
 	}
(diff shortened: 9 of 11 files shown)


Makefile
```diff
@@ -1,7 +1,14 @@
 # The development version of clang is distributed as the 'clang' binary,
 # while stable/released versions have a version number attached.
 # Pin the default clang to a stable version.
+GOARCH := $(shell go env GOARCH)
+ifneq ($(GOARCH),arm64)
 CLANG ?= clang-12
+else
+# Ubuntu Focal used for the ARM64 builder container does not provide
+# clang-12 out of the box
+CLANG ?= clang-11
```
lmb, line 10: Unfortunately we need clang-12 for upcoming CO-RE support, see #301. FWIW I don't think we need to build using arm64 clang, there should be no difference in output for bpfel / bpfeb targets on either platform.

examples/kprobe/fn_amd64.go
```diff
@@ -0,0 +1,4 @@
+package main
+
+// Name of the kernel function to trace.
+const fn = "__x64_sys_execve"
```
lmb, line 4: Checking https://golang.org/pkg/runtime/#pkg-constants GOARCH in the code should be simpler.

lmb (changes requested):
Thanks for making the PR :) Please see my comments, I think changing the ELF builder is not necessary. It's nice that we have working tests on arm64, but we can't claim to support arm64 until we have a working CI.

examples/kprobe/fn_amd64.go
nickolaev: OK, changed the approach. Let me know if that works.

link/platform.go
```diff
@@ -0,0 +1,29 @@
+package link
+
+import "runtime"
+
+func byArch(arch_map map[string]string) string {
```
lmb, line 5: Suggestion: just switch on runtime.GOARCH and prepend `__x64_`, etc. to a string supplied by the user.
```suggestion
func archSyscallName(name string) string {
```

```
archSyscallName("sys_getpid")
```

CONTRIBUTING.md
```diff
@@ -20,4 +20,32 @@ reason about the proposed changes.
...
+
+## Other platforms
+
+The main development platform that the project targets is `amd64`. However running
+the code and the examples should not be hard on any platform supported by `go` and
+`docker`. Here are some hints for `arm64`, adding other platforms should be similar.
+The suggested approach will always use `amd64` build tools, but allow to cross-build
+and use them with an emulation.
+
+#### Ubuntu `linux/arm64`
```
lmb, line 33: So what commands would you usually run when developing on linux/arm64 then? `go test ./...`? `make`?

examples/kprobe/main.go
```diff
@@ -50,7 +46,7 @@ func main() {
 	// pre-compiled program. Each time the kernel function enters, the program
 	// will increment the execution counter by 1. The read loop below polls this
 	// map value once per second.
-	kp, err := link.Kprobe(fn, objs.KprobeExecve)
+	kp, err := link.Kprobe(link.SysExecve(), objs.KprobeExecve)
```
lmb, line 49: I wonder whether we should move this fallback behaviour into Kprobe, Kretprobe, etc. proper. Otherwise our interface isn't very portable. That is, the user specifies "sys_execve" here, and we discover
* the correct arch prefix
* with a fallback to just the plain name

@ti-mo thoughts?

link/platform.go
```diff
@@ -0,0 +1,29 @@
...
+	value, ok := arch_map[runtime.GOARCH]
+	if !ok {
+		panic("Unsupported arch " + runtime.GOARCH)
+	}
+	return value
+}
+
+func SysGetpid() string {
```
lmb, line 13: I don't think this warrants a stand alone / exported function, see my comments about archSyscallName and maybe baking this into Kprobe, etc.

Makefile
```diff
@@ -31,15 +31,15 @@ TARGETS := \
 
 # Build all ELF binaries using a Dockerized LLVM toolchain.
 docker-all:
-	docker run --rm --user "${UIDGID}" \
+	docker run --platform linux/amd64 --rm --user "${UIDGID}" \
```
lmb, line 34: Unfortunately this breaks on my machine:
```
$ make
docker run --platform linux/amd64 --rm --user "1000:1000" \
	-v "/home/lorenz/dev/ebpf":/ebpf -w /ebpf --env MAKEFLAGS \
	--env CFLAGS="-fdebug-prefix-map=/ebpf=." \
	"quay.io/cilium/ebpf-builder:1619686610" \
	make all
"--platform" is only supported on a Docker daemon with experimental features enabled
make: *** [Makefile:34: docker-all] Error 1
```

CONTRIBUTING.md
ti-mo, line 33: So Linux == Ubuntu? :) 

If I understand correctly, this section is supposed to describe how to generate a Docker image containing an arm64-based LLVM toolchain. Can't tell for sure because there's no conclusion to the description. This would enable developers running arm64-based Linux distributions as a development environment (which might include Docker-on-M1), is that correct?

If so, I would remove all of the fluff here and make `testdata/docker` rely on buildx. A short mention can be included in the readme there about `qemu-user-static` and `binfmt-support` being required, although this is kind of expected for running a cross-compilation toolchain. `buildx` will likely fail and tell the user these tools are required.

Makefile
ti-mo: This change feels a little unnecessary. `docker run` should try to pull and run an image compatible with the host's architecture. (I assume this is the point of distributing arm64 LLVM builder images) Does this actually work on M1? Will it try to pull and run an x86 Docker image and execute it on an emulated x86 host?

If so, what's the point of building arm64 toolchain images? (I assume it's horribly slow for larger eBPF codebases)

link/perf_event_test.go
```diff
@@ -26,7 +26,7 @@ func TestTraceEventTypePMU(t *testing.T) {
 func TestTraceEventID(t *testing.T) {
 	c := qt.New(t)
 
-	eid, err := getTraceEventID("syscalls", "sys_enter_fork")
+	eid, err := getTraceEventID("syscalls", "sys_enter_execve")
```
ti-mo, line 29: Why?

link/tracepoint_test.go
```diff
@@ -66,7 +66,7 @@ func TestTracepointErrors(t *testing.T) {
 }
 
 func TestTraceGetEventID(t *testing.T) {
-	_, err := getTraceEventID("syscalls", "sys_enter_open")
+	_, err := getTraceEventID("syscalls", "sys_enter_openat")
```
ti-mo, line 69: Why?

examples/kprobe/main.go
ti-mo: Agreed, this needs to become a part of the API, not a hack in the test suite. That doesn't add much value to the user, as it makes it harder to write tools portable across architectures.

We also need to hash out what the user should expect, as in: do they specify `__x64_sys_execve` and rely on the lib to implicitly fix it up to make it portable? Or do they specify `sys_execve`, and the lib always tries with and without the arch prefix? There might be code out there already specifying `__x64_sys_execve` that will break if we change the rules.

I'd suggest the following:
- If we detect an arch prefix, fix it up to the prefix that belongs to the host. This should work in most cases.
- Then, first, do exactly what the user asked for (`sys_execve`). If the symbol already contains an arch prefix, we can bail out here.
- If that fails, try again with the arch prefix prepended (`__x64_sys_execve`). Syscall symbols are in the minority, so this should avoid the most back-and-forth.

Note: Using tracepoints should be preferred for probing most popular syscalls (execve was only chosen here because it's easy to trigger), but I'm not sure if tracepoints yield the same flexibility in terms of walking memory.

ti-mo (changes requested):
Thanks for the patch! We're seeing arm being used more and more frequently, but I think this needs some more work in order to become useful.

- Please remove all merge commits. `git fetch origin && git rebase origin/master` is the way to keep your branch up-to-date with changes that have landed on master.
- I don't feel like the additions to `CONTRIBUTING.md` are necessary. If we want to provide llvm builder images going forward, `testdata/docker` should output images for both arm64 and x86, which will automatically be picked on hosts/dev machines of those respective architectures.
- `link.K(ret)probe` should handle `__x64_` syscall fn prefixes transparently, but implementing this might be annoying given the difference between tracefs and PMU, see other comment on the code itself. Please reach out through Cilium Slack (`#libbpf-go`) if you need more input.

Marking this as draft in the mean time.

examples/kprobe/main.go
lmb: I think the can do something simpler. Assuming we get `__x64_sys_execve` as the argument:
* First try with the arch specific prefix added: `__x64___x64_sys_execve`. This will fail.
* Fall back to trying without the arch prefix (necessary pre 4.17 anyways): `__x64_sys_execve`. This will work on new kernels.

We shouldn't try to translate arch prefixes from one arch to another. This not any more or less broken than at the moment and a lot simpler.

Makefile
lmb: I had a quick look the other day, I'm not sure docker is smart enough unfortunately. Something about us having to publish a manifest. I didn't dig too deep and would love to proven wrong.

examples/kprobe/main.go
ti-mo: You might be right. The user might want to attach only to `__x64_sys_execve` and bail out on other architectures. We also wouldn't have to maintain a list of architecture prefixes to strip, reducing maintenance overhead.

For portable probes, the bare syscall symbol (`sys_execve`) must be specified. This is probably the cleanest API contract.

Makefile
nickolaev: > This change feels a little unnecessary. `docker run` should try to pull and run an image compatible with the host's architecture. (I assume this is the point of distributing arm64 LLVM builder images) Does this actually work on M1? Will it try to pull and run an x86 Docker image and execute it on an emulated x86 host?
> 
> If so, what's the point of building arm64 toolchain images? (I assume it's horribly slow for larger eBPF codebases)

I have tried this one on my M1 in both Docker Desktop for Mac and my Parallels Ubuntu 20.04 VM. Both work with this flag, and I would say is usable for the needs of the development within this project. At this point, I don't think there is any reason to build arm64 images at all.

link/perf_event_test.go
nickolaev: From what I can tell `sys_enter_fork` was not available on arm, so I thought that for the purposes of testing `sys_enter_execve` is more portable.

link/tracepoint_test.go
nickolaev: Same as above, just a more widely implemented syscall, that is available on other archs.

nickolaev:
@lmb and @ti-mo, I've tried to be creative here and figure a better way to handle syscall naming cross platforms and kernels. Let me know if that aligns with your ideas here.

internal/kernel.go
```diff
@@ -0,0 +1,53 @@
+package internal
```
lmb, line 1: check out internal/version.go

link/platform.go
```diff
@@ -0,0 +1,68 @@
...
+	"regexp"
+	"runtime"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+const (
+	// Since 4.17, syscalls symbols are generated with the `__x64_` prefix.
+	// https://github.com/torvalds/linux/commit/d5a00528b58cdb2c71206e18bd021e34c4eab878
+	LinuxVersionWithPrefix = 4<<16 + 17<<8 + 0
```
lmb, line 15: Version comparisons are brittle, since individual distributions freeze or change version numbers according to their own whims. It's always preferable to attempt an action and fall back to old behaviour if necessary.

link/platform.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var parseArchSyscall = regexp.MustCompile(`__(\w*)_sys_(\w*)`)
+var parseSyscall = regexp.MustCompile(`sys_(\w*)`)
+
+func ResolveSyscall(syscall string) (string, error) {
```
lmb, line 21: Why do you want to support this API? The user should give us the "platform independent" syscall name, then we don't need all of this code.

lmb (changes requested):
Please take a look at https://github.com/cilium/ebpf/pull/304#discussion_r636994287 which outlines a simpler API.

examples/go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	github.com/cilium/ebpf v0.5.1-0.20210428090942-9aea0e1002a2
 	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c
 )
+
+replace github.com/cilium/ebpf => ./..
```
lmb, line 10: I don't think this will work: https://github.com/golang/go/issues/30354 The way to go about this is to make the change to examples in a separate PR, after this change has been merged. Otherwise there is no way to update the dependency.

link/platform.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+	"runtime"
+)
+
+func platformPrefix(symbol string) string {
+
+	prefix := runtime.GOARCH
+	switch prefix {
```
lmb, line 11: Do you have a reference / link which prefixes the kernel uses? Your code currently assumes that GOARCH == kernel prefix in general, I'm not sure that will be true.

lmb (changes requested):

link/platform.go
ti-mo: > Your code currently assumes that GOARCH == kernel prefix in general, I'm not sure that will be true.

Out of curiosity, what do you mean by this? I fail to come up with an example. Perhaps when running the binary under `qemu-user-static`? What other facilities would the runtime expose to detect underlying OS?

link/platform.go
```diff
@@ -0,0 +1,19 @@
...
+	case "i386":
```
ti-mo, line 12: This is not a valid GOARCH value. All possible values are here: https://github.com/golang/go/blob/master/src/go/build/syslist.go.

link/platform.go
ti-mo: > Do you have a reference / link which prefixes the kernel uses?

Tricky, but from what I can find:

```
# in linux
λ  linux master ✗ egrep -R "define __SYSCALL(.*) = __" arch
arch/x86/entry/syscall_64.c:#define __SYSCALL_64(nr, sym) [nr] = __x64_##sym,
arch/x86/entry/syscall_32.c:#define __SYSCALL_I386(nr, sym) [nr] = __ia32_##sym,
arch/x86/entry/syscall_x32.c:#define __SYSCALL_X32(nr, sym) [nr] = __x32_##sym,
arch/x86/entry/syscall_x32.c:#define __SYSCALL_COMMON(nr, sym) [nr] = __x64_##sym,
arch/arm64/kernel/sys32.c:#define __SYSCALL(nr, sym)	[nr] = __arm64_##sym,
arch/arm64/kernel/sys.c:#define __SYSCALL(nr, sym)	[nr] = __arm64_##sym,
```

lmb: > Out of curiosity, what do you mean by this?

The code relies on the fact that on arm64 GOARCH == the prefix used by the kernel, and assumes that the same is true for other platforms. Nothing to do with changes in GOARCH due to different runtimes.

link/platform.go
```diff
@@ -0,0 +1,19 @@
...
+		prefix = "ia32"
+	case "amd64":
+		prefix = "x64"
+	}
```
ti-mo, line 16: This branch/PR is targeting arm64, but I don't see an arm64 entry here. :) Note that according to https://github.com/golang/go/blob/master/src/go/build/syslist.go there's also an `arm64be` entry, but the Go compiler doesn't seem to support that out of the box:
```
λ  ~/documents/go/sandbox  GOARCH=arm64be go build
cmd/go: unsupported GOOS/GOARCH pair linux/arm64be
```
so I think we can skip an arm64be entry.

link/platform.go
ti-mo: Oh, completely missed that. We should probably return empty by default and only emit known prefixes.

@iii-i (sorry for the random ping), could you chime in on what's common in s390(x)? I'm finding references to both `__s390_` and `__s390x_` in Linux.

nickolaev: I assume we can cover as much as possible of the options. I tried to cover x86 and arm64. We can panic on unknown platform and let these be added over time once users hit "unsupported platform".

link/platform.go
nickolaev: True, my eyes misled me on this. From what I recall I checked this gist -> https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63#a-list-of-valid-goarch-values

link/platform.go
nickolaev: Yes, my approach was to "patch" only for those platform that differ from GOARCH. Maybe it is better to make these translations more explicint in map?

examples/go.mod
nickolaev: It depends on what is `works`. For sure with this patch, I am able to run the kprobe example and am not able to do so without it.

link/platform.go
ti-mo: As this is library code, we cannot panic. The safer option would be to not return a prefix, so the given symbol is plumbed through as is. Of course, we'd want to ship this with most of the available architectures supported as you suggested..

README.md
```diff
@@ -57,6 +57,12 @@ environment to keep the build output stable.
 
 The toolchain image build files are kept in [testdata/docker/](testdata/docker/).
 
+### Regeneration Testdata on non-x86 platforms
```
ti-mo, line 60: Could this be moved to `testdata/docker/README.md` under the `Building` section?

link/platform.go
ti-mo: The structure is fine as you suggested, I would just add a default case with `return symbol`.

lmb (approved):

ti-mo:
@nickolaev Thanks for the work!

PR #306, opened 2021-05-18
program: better error when loading a single program

### LINKED ISSUE #210: bad file descriptor: fd -1 is not pointing to valid bpf_map
I am getting this error when calling `ebpf.NewProgramWithOptions`. What is interesting is that the problem only observed when ebpf' s .c code carries these two definitions, once I comment these out the function succeeds.
```
static volatile const uint32_t svc_proto = 0;
static volatile const uint32_t svc_port = 0;
```

Here is ebpf c code:
```
#include "common.h"

#include <linux/if_ether.h>
#include <linux/swab.h>
#include <linux/pkt_cls.h>
#include <bpf/bpf_helpers.h>
#include <arpa/inet.h>

static volatile const uint32_t svc_proto = 0;
static volatile const uint32_t svc_port = 0;

static __always_inline int global_fn(uint32_t arg)
{
    return arg;
}

__section("classifier_egress_drop") int egress_drop(struct __sk_buff *skb)
{
    void *data_end = (void *)(unsigned long long)skb->data_end;
    void *data = (void *)(unsigned long long)skb->data;
    struct ethhdr *eth_h = data;

    /* Drop mailformated packet	*/
    if (data + sizeof(struct ethhdr) > data_end)
        return TC_ACT_SHOT;

    // Interested only in IPv6 packet
    if (ntohs(eth_h->h_proto) != ETH_P_IPV6)
    {
        return TC_ACT_OK;
    }
    // Now it is safe to type cast to ipv6 header struct
    // struct ip6_hdr *ipv6_h = (struct ip6_hdr *)(data - sizeof(struct ethhdr));
    // struct tcphdr *th = tcp_hdr(skb)

    global_fn(svc_proto);
    global_fn(svc_port);

    return TC_ACT_OK;
}

struct bpf_map_def __section("maps") map1 = {
    .type = BPF_MAP_TYPE_ARRAY,
    .key_size = sizeof(unsigned int),
    .value_size = 4096,
    .max_entries = 1,
};
```

Portion of go code:
```
func main() {

	ctx, cancel := context.WithTimeout(context.Background(), 600*time.Second)
	defer cancel()

	// Load eBPF from an elf file
	coll, err := ebpf.LoadCollectionSpec("ebpf/drop")
	if err != nil {
		fmt.Fprintf(os.Stderr, "could not load collection from file: %v\n", err)
		return
	}
	 if _, ok := coll.Maps[".rodata"]; !ok {
	 	fmt.Fprintf(os.Stderr, "could not find .rodata section to set argument\n")
	 	return
	 }
	 if err := coll.RewriteConstants(map[string]interface{}{
	 	"svc_proto": uint32(6),
	 }); err != nil {
	 	fmt.Fprintf(os.Stderr, "could not find override arg with error: %+v\n", err)
	 	return
	 }
	 if err := coll.RewriteConstants(map[string]interface{}{
	 	"svc_port": uint32(2020),
	 }); err != nil {
	 	fmt.Fprintf(os.Stderr, "could not find override arg with error: %+v\n", err)
	 	return
	 }

	if _, ok := coll.Maps["map1"]; !ok {
		fmt.Fprintf(os.Stderr, "could not find map %q in loaded collection\n", "map1")
		return
	}
	_, err = ebpf.NewMap(coll.Maps["map1"])
	if err != nil {
		fmt.Fprintf(os.Stderr, "could not make new map: %v\n", err)
		return
	}
	// Load the eBPF program egress_drop
	p, ok := coll.Programs["egress_drop"]
	if !ok {
		fmt.Fprintf(os.Stderr, "ebpf/drop does not container egress_drop section\n")
		return
	}
	fmt.Printf("Program: %+v\n", *p)
	egressDrop, err := ebpf.NewProgramWithOptions(p,
		ebpf.ProgramOptions{
			LogLevel: 1,
			LogSize:  65536,
		})
	if err != nil {
		fmt.Fprintf(os.Stderr, "NewProgramWithOptions could not load program %+v\n with error: %+v\n", *p, err)
		return
	}
	defer egressDrop.Close()
}
```

It is possible that I am doing something incorrectly, nevertheless declaration of two constant should not brea things or at least the error should be more helpful.

lmb:
Sometimes a user assumes that a ProgramSpec from a CollectionSpec can
be loaded with NewProgram instead of using NewCollection. This isn't
true unfortunately, and leads to a hard to understand error issued
by the verifier:

    bad file descriptor: fd -1 is not pointing to valid bpf_map

We can catch an unresolved map reference when loading a program and
use that to give a better error message to callers of NewProgram.

Fixes #210

### DIFF
--- a/asm/load_store.go
+++ b/asm/load_store.go
@@ -111,7 +111,7 @@ func LoadMapPtr(dst Register, fd int) Instruction {
 		OpCode:   LoadImmOp(DWord),
 		Dst:      dst,
 		Src:      PseudoMapFD,
-		Constant: int64(fd),
+		Constant: int64(uint32(fd)),
 	}
 }
 
--- a/prog.go
+++ b/prog.go
@@ -20,6 +20,8 @@ import (
 // ErrNotSupported is returned whenever the kernel doesn't support a feature.
 var ErrNotSupported = internal.ErrNotSupported
 
+var errUnsatisfiedReference = errors.New("unsatisfied reference")
+
 // ProgramID represents the unique ID of an eBPF program.
 type ProgramID uint32
 
@@ -135,7 +137,11 @@ func NewProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 	handles := newHandleCache()
 	defer handles.close()
 
-	return newProgramWithOptions(spec, opts, handles)
+	prog, err := newProgramWithOptions(spec, opts, handles)
+	if errors.Is(err, errUnsatisfiedReference) {
+		return nil, fmt.Errorf("cannot load program without loading its whole collection: %w", err)
+	}
+	return prog, err
 }
 
 func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *handleCache) (*Program, error) {
--- a/testdata/loader-eb.elf
+++ b/testdata/loader-eb.elf
@@ -1,0 +1,1 @@
+loader-clang-12-eb.elf
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,8 @@ clean:
 	-$(RM) internal/btf/testdata/*.elf
 
 all: $(addsuffix -el.elf,$(TARGETS)) $(addsuffix -eb.elf,$(TARGETS))
+	ln -srf testdata/loader-$(CLANG)-el.elf testdata/loader-el.elf
+	ln -srf testdata/loader-$(CLANG)-eb.elf testdata/loader-eb.elf
 
 testdata/loader-%-el.elf: testdata/loader.c
 	$* $(CFLAGS) -mlittle-endian -c $< -o $@
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -138,8 +138,12 @@ func (ins *Instruction) RewriteMapPtr(fd int) error {
 	return nil
 }
 
-func (ins *Instruction) mapPtr() uint32 {
-	return uint32(uint64(ins.Constant) & math.MaxUint32)
+// MapPtr returns the map fd for this instruction.
+//
+// The result is undefined if the instruction is not a load from a map,
+// see IsLoadFromMap.
+func (ins *Instruction) MapPtr() int {
+	return int(int32(uint64(ins.Constant) & math.MaxUint32))
 }
 
 // RewriteMapOffset changes the offset of a direct load from a map.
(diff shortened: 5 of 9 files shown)


prog.go
```diff
@@ -20,6 +20,8 @@ import (
 // ErrNotSupported is returned whenever the kernel doesn't support a feature.
 var ErrNotSupported = internal.ErrNotSupported
 
+var errUnsatisfiedReference = errors.New("unsatisfied reference")
```
ti-mo, line 23: The error string is nice and short, but still somewhat opaque. How about `unsatisfied reference to map`? It's currently only used for maps, so could be `errUnsatisfiedMapReference`.

prog.go
```diff
@@ -135,7 +137,11 @@ func NewProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 	handles := newHandleCache()
 	defer handles.close()
 
-	return newProgramWithOptions(spec, opts, handles)
+	prog, err := newProgramWithOptions(spec, opts, handles)
+	if errors.Is(err, errUnsatisfiedReference) {
+		return nil, fmt.Errorf("possible attempt to load program instead of collection: %w", err)
```
ti-mo, line 142: This will come out as `possible attempt to load program instead of collection: unsatisfied reference`. I think we could drop the word 'possible' here: `cannot load program without loading its whole collection: unsatisfied reference to map`.

On the off chance the caller tries to load a bare prog with a missing map ref (no collection), there's still the 'unsatisfied reference to map' to help out as the more-specific error.

asm/load_store.go
```diff
@@ -111,7 +113,7 @@ func LoadMapPtr(dst Register, fd int) Instruction {
 		OpCode:   LoadImmOp(DWord),
 		Dst:      dst,
 		Src:      PseudoMapFD,
-		Constant: int64(fd),
+		Constant: int64(uint32(fd) & math.MaxUint32),
```
ti-mo, line 116: Just for my reference: is the mask required here? Won't the u32 conversion take care of truncating already?

testdata/loader-eb.elf
```diff
@@ -0,0 +1 @@
+loader-clang-12-eb.elf
```
ti-mo, line 1: Do we want to commit these to the repo?

ti-mo (commented):
Looks good! Sorry for the delay. Minor nit on the error string and left a few questions.

prog.go
lmb: The error is actually `possible attempt to load program instead of collection: map foobar: unsatisfied reference`, so I'd leave it as it is.

prog.go
lmb: sgtm.

asm/load_store.go
lmb: You're right, it's not necessary.

testdata/loader-eb.elf
lmb: I think it's useful to have a symlink to whatever the version compiled with the latest clang is. The other option is to hardcode `loader-clang-12-eb.elf` in the tests and change them when we upgrade clang version.

lmb: I don't mind either way, up to you.

prog.go
ti-mo: Okay, cool! That's perfect :+1:

testdata/loader-eb.elf
ti-mo: Nope, I prefer the symlink approach. :+1:

ti-mo (approved):
:rocket:

Issue #305, opened 2021-05-17
bpf2go: define __TARGET_ARCH_xxx based on GOARCH


lmb:
_Originally posted by @mmat11 in https://github.com/cilium/ebpf/issues/303#issuecomment-840850641_

Hi, I believe `bpf_tracing.h` needs additional env vars as written [here](https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/bpf_tracing.h#L5).
The following works for me:

```
/t/tcpconnlat *master> BPF_CFLAGS='-D__TARGET_ARCH_x86' go generate ./...
Compiled /tmp/tcpconnlat/tcpconnlat_bpfel.o
Wrote /tmp/tcpconnlat/tcpconnlat_bpfel.go
Compiled /tmp/tcpconnlat/tcpconnlat_bpfeb.o
Wrote /tmp/tcpconnlat/tcpconnlat_bpfeb.go
/t/tcpconnlat *master> git diff
diff --git a/main.go b/main.go
index b4fb0f2..9d685f7 100644
--- a/main.go
+++ b/main.go

-//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang-11 TCPConnLat ./bpf/tcpconnlat.bpf.c -- -I./headers
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cflags $BPF_CFLAGS -cc clang-12 TCPConnLat ./bpf/tcpconnlat.bpf.c -- -I./headers
```

Implementation: in bpf2go inspect `runtime.GOARCH` and map it to one of the suffixes in `bpf_tracing.h`. Add the appropriate define to the command line.

lmb:
I think to support this properly we have to do the following:

- Extend the `-target` flag to take any valid GOARCH in addition to `bpf, bpfel, bpfeb`
- If the `-target` is a GOARCH add the necessary `__TARGET_ARCH_xxx` define
- For `bpf`, `bpfel`, `bpfeb` targets define `__BPF_TARGET_MISSING`: "The eBPF is using target specific macros, please provide -target"
- Support multiple targets separated by comma `,`

PR #307, opened 2021-05-18
chore: bpf2go will automatically add BPF_CFLAGS='-D__TARGET_ARCH`

nickolaev:
An attempt to implement https://github.com/cilium/ebpf/issues/305


cmd/bpf2go/platform.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+import "runtime"
+
+func getTargetArch() string {
+	if runtime.GOOS != "linux" {
```
lmb, line 6: runtime.GOOS is the platform we're running on, not the target platform. This would effectively prevent cross compiling BPF for Linux on e.g. a MacOS machine. I think this should be removed.

cmd/bpf2go/platform.go
```diff
@@ -0,0 +1,26 @@
...
+		panic("Not supported platform " + runtime.GOOS)
+	}
+
+	switch runtime.GOARCH {
```
lmb, line 10: Similar comment as above: runtime.GOARCH is the platform we run on. Instead of only looking at `runtime.GOARCH` we should only use it as a fall back if $GOARCH isn't defined in the environment.

My suggestion: add `goarch string` as a parameter and call it like `getTargetArch(os.Getenv("GOARCH"))`.

cmd/bpf2go/compile.go
```diff
@@ -67,6 +67,9 @@ func compile(args compileArgs) error {
 		"-g",
 	)
 	cmd.Dir = args.dir
+	cmd.Env = append(cmd.Env,
+		"BPF_CFLAGS='-D"+getTargetArch()+"'",
```
lmb, line 71: You should pass `"-D"+getTargetArch()` directly to the compiler, not modify BPF_CFLAGS (which doesn't have any effect). I think the right place to add the flag is actually in main.go, where you can **prepend** it to `cFlags`. Prepending is important so that our users can override the define if they choose to do so.

cmd/bpf2go/platform.go
```diff
@@ -0,0 +1,26 @@
...
+	case "arm":
+		return "__TARGET_ARCH_arm"
+	case "arm64":
+		return "__TARGET_ARCH_arm64"
+	case "mipsle", "mips", "mips64", "mips64le":
+		return "__TARGET_ARCH_mips"
+	case "ppc64", "ppc64le":
+		return "__TARGET_ARCH_powerpc"
+	default:
+		panic("Not supported platform " + runtime.GOARCH)
```
lmb, line 24: In this case we should just not add a `-D` IMO.

lmb (changes requested):
Hey, thanks for tackling this! Can you add a test for this please? I think the best would be to add something like the following to minimal.c

```
#if !defined(__TARGET_ARCH_x86) && !defined(...) && ...
#error TARGET_ARCH is not defined
#endif
```

This will make the compiler error out if detection doesn't work.

lmb:
I've outlined what we should do here: https://github.com/cilium/ebpf/issues/305#issuecomment-895076995 It's not as straight forward as I had thought. I'll close this PR, sorry for leading you down the wrong path.

PR #315, opened 2021-05-28
link: return os.ErrNotExist when symbol doesn't exist

lmb:
Some of the tracing related links currently return ErrNotSupported
when trying to attach to a non-existant symbol. This is misleading.
Consistently return os.ErrNotExist in such cases.

### DIFF
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -207,8 +207,8 @@ func unsafeStringPtr(str string) (unsafe.Pointer, error) {
 // group and name must be alphanumeric or underscore, as required by the kernel.
 func getTraceEventID(group, name string) (uint64, error) {
 	tid, err := uint64FromFile(tracefsPath, "events", group, name, "id")
-	if errors.Is(err, ErrNotSupported) {
-		return 0, fmt.Errorf("trace event %s/%s: %w", group, name, ErrNotSupported)
+	if errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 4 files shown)


link/tracepoint_test.go
```diff
@@ -48,6 +48,19 @@ func TestTracepoint(t *testing.T) {
 	}
 }
 
+func TestNonexistantTracepoint(t *testing.T) {
```
ti-mo, line 51: Could this be `TestTracepointMissing` instead? I used `-run="Test(Kprobe|Kretprobe|Trace)"` frequently for this package.

link/perf_event.go
```diff
@@ -255,22 +257,13 @@ func openTracepointPerfEvent(tid uint64) (*internal.FD, error) {
...
 	p := filepath.Join(base, l)
 	if !strings.HasPrefix(p, base) {
 		return 0, fmt.Errorf("path '%s' attempts to escape base path '%s': %w", l, base, errInvalidInput)
 	}
 
 	data, err := ioutil.ReadFile(p)
-	if os.IsNotExist(err) {
-		// Only echo leaf path, the base path can be prepended at the call site
-		// if more verbosity is required.
-		return 0, fmt.Errorf("symbol %s: %w", l, ErrNotSupported)
```
ti-mo, line 272: Thanks, this makes sense. :+1: The first version of this function might've taken  `symbol`, but no longer has anything to do with it.

ti-mo (approved):
Thanks! Some minor nits.

PR #317, opened 2021-06-01
docs: add testing instructions to CONTRIBUTING.md

rgo3:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,23 @@ reason about the proposed changes.
 ## Running the tests
 
 Many of the tests require privileges to set resource limits and load eBPF code.
-The easiest way to obtain these is to run the tests with `sudo`:
+The easiest way to obtain these is to run the tests with `sudo`.
 
-    sudo go test ./...
+To test the current package with your local kernel you can simply run:
+```
+go test -exec sudo  ./...
+```
+
+To test the current package with a different kernel version you can use the [run-tests.sh](run-tests.sh) script.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,6 +18,25 @@ reason about the proposed changes.
 ## Running the tests
 
 Many of the tests require privileges to set resource limits and load eBPF code.
-The easiest way to obtain these is to run the tests with `sudo`:
+The easiest way to obtain these is to run the tests with `sudo`.
+
+To test the current package with your local kernel you can simply run:
+```bash
+sudo go test ./...
```
ti-mo, line 25: This should be replaced with `go test -exec sudo ./...`. Unlike `sudo go test ./...`, it won't execute the Go build as sudo, only the resulting `a.out` artifacts.

ti-mo (commented):
Thanks! Looping in @lmb because I'm not sure if we want to 'hardcode' 5.4 in the docs here, since it'll eventually go away.

rgo3:
Fair point with the version, I picked it because it was the example from the script comment. Should I change it to something like `./run-tests.sh <kernel-version>`?

ti-mo:
> Fair point with the version, I picked it because it was the example from the script comment. Should I change it to something like `./run-tests.sh <kernel-version>`?

Actually, 5.4 is LTS, so should be fine having it there. `<kernel-version>` doesn't tell the user which versions are available to test.

rgo3:
Ok, Ill change the sudo command example and push the changes then :).

ti-mo:
Thank you!

Issue #235, opened 2021-03-02
Expose kernel feature test API


ti-mo:
For https://github.com/cilium/cilium/issues/11014, we're looking for a safe and fast way to make decisions in the Cilium code around (BPF-related) kernel features. Currently, we rely on `bpftool` for some things, some Cilium-specific code for others, but we'd prefer if this would be abstracted and unified to make things clearer. We've informally discussed this topic before, but I'm looking to open the conversation here so we can gather some more data to make an informed decision about the way forward.

Some implementation/design ideas below:

### Errors

@lmb suggested we don't explicitly expose a 'features' API, but that the caller rely on errors being returned from particular library features. For example, a (hypothetical) `Map.IterateBatch()` would return a `NotSupportedError` when the kernel doesn't support map batch operations.

### Package Variables

Exporting current `have...` variables (e.g. `ebpf.HaveNestedMaps`). This would hugely inflate the package's API surface; not ideal.

### Features Struct

`ebpf.Features()` returning a struct containing `bool`s:

```go
type KernelFeatures {
  NestedMaps bool
  MapMutabilityModifiers bool
  etc ...
}
```

The struct can easily be appended to as new features are added, and will not cause breakage. Tests can be deprecated over time if they no longer prove useful.

Doesn't inflate the package API surface.

This option would allow Cilium to include a quick overview of supported kernel features in debug/status output. I remember someone internally voicing a need for this, though not sure who did.

---

Another conversation topic is whether or not we want to continue maintaining these feature tests going forward. What is the maintenance burden of the tests currently in the library? Once written, do they need to be extended or modified?

lmb:
Thanks for putting this together. In my mind your proposals fall into two categories:

Minimal (aka ErrNotSupported): provides a smooth user experience with minimal API surface. You start out writing your code in a naive way and abort on ErrNotSupported. You figure out a way to work around the issue and use `errors.Is(err, ErrNotSupported)` to switch to that other code path. However I suspect there are cases where this doesn't work well. Do you know of any in the Cilium code base?

Batteries included (aka package vars, features struct): probably even nicer for the user, since based on these tests we could return ErrNotSupported as well. This is a lot more uncertain for me: which feature tests do we accept, which ones do we reject? Do we have to maintain them forever? I don't want the library to become a dumping ground of feature tests that nobody uses anymore.

[...]

PR #320, opened 2021-06-08
link: wrap errors for Raw{At,De}tachProgram 

cyphar:
This allows callers to detect the underlying syscall error, which is
necessary for being able to implement safe fallbacks based on the error
during program loading and unloading. runc in particular needs this to
be able to implement BPF_F_REPLACE fallbacks correctly.

Ref: opencontainers/runc#3008
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/link/program.go
+++ b/link/program.go
@@ -43,7 +43,7 @@ func RawAttachProgram(opts RawAttachProgramOptions) error {
 	}
 
 	if err := internal.BPFProgAttach(&attr); err != nil {
-		return fmt.Errorf("can't attach program: %s", err)
+		return fmt.Errorf("can't attach program: %w", err)
 	}
 	return nil
 }
@@ -69,7 +69,7 @@ func RawDetachProgram(opts RawDetachProgramOptions) error {
 		AttachType:  uint32(opts.Attach),
 	}
 	if err := internal.BPFProgDetach(&attr); err != nil {
-		return fmt.Errorf("can't detach program: %s", err)
+		return fmt.Errorf("can't detach program: %w", err)
 	}
 
 	return nil


AkihiroSuda (changes requested):
I was about to open the same PR but my draft have the following changes 😛 

```diff
diff --git a/internal/elf.go b/internal/elf.go
index 54a4313..88326bb 100644
--- a/internal/elf.go
+++ b/internal/elf.go
@@ -24,7 +24,7 @@ func NewSafeELFFile(r io.ReaderAt) (safe *SafeELFFile, err error) {
                }
 
                safe = nil
-               err = fmt.Errorf("reading ELF file panicked: %s", r)
+               err = fmt.Errorf("reading ELF file panicked: %w", r)
        }()
 
        file, err := elf.NewFile(r)
@@ -44,7 +44,7 @@ func (se *SafeELFFile) Symbols() (syms []elf.Symbol, err error) {
                }
 
                syms = nil
-               err = fmt.Errorf("reading ELF symbols panicked: %s", r)
+               err = fmt.Errorf("reading ELF symbols panicked: %w", r)
        }()
 
        syms, err = se.File.Symbols()
@@ -60,7 +60,7 @@ func (se *SafeELFFile) DynamicSymbols() (syms []elf.Symbol, err error) {
                }
 
                syms = nil
-               err = fmt.Errorf("reading ELF dynamic symbols panicked: %s", r)
+               err = fmt.Errorf("reading ELF dynamic symbols panicked: %w", r)
        }()
 
        syms, err = se.File.DynamicSymbols()
```

AkihiroSuda (changes requested):
```diff
diff --git a/cmd/bpf2go/compile.go b/cmd/bpf2go/compile.go
index a7509e1..0f0b28f 100644
--- a/cmd/bpf2go/compile.go
+++ b/cmd/bpf2go/compile.go
@@ -91,7 +91,7 @@ func compile(args compileArgs) error {
        }
 
        if err := cmd.Start(); err != nil {
-               return fmt.Errorf("can't execute %s: %s", args.cc, err)
+               return fmt.Errorf("can't execute %s: %w", args.cc, err)
        }
 
        if depRd != nil {
@@ -104,7 +104,7 @@ func compile(args compileArgs) error {
        }
 
        if err := cmd.Wait(); err != nil {
-               return fmt.Errorf("%s: %s", args.cc, err)
+               return fmt.Errorf("%s: %w", args.cc, err)
        }
 
        return nil
```

cyphar:
Ah, I did a search-and-replace rather than chekcing for the cases where the format string was more complicated. I'll enable the `errorlint` linter and will fix all the errors then push again.

internal/btf/btf.go
```diff
@@ -152,7 +152,7 @@ func loadSpecFromVmlinux(rd io.ReaderAt) (*Spec, error) {
 
 	btfSection, _, _, err := findBtfSections(file)
 	if err != nil {
-		return nil, fmt.Errorf(".BTF ELF section: %s", err)
+		return nil, fmt.Errorf(".BTF ELF section: %w", err)
```
AkihiroSuda, line 155:
```diff
diff --git a/internal/btf/core.go b/internal/btf/core.go
index 7c888f6..a31abf2 100644
--- a/internal/btf/core.go
+++ b/internal/btf/core.go
@@ -473,7 +473,7 @@ func parseCoreAccessor(accessor string) (coreAccessor, error) {
                // 31 bits to avoid overflowing int on 32 bit platforms.
                index, err := strconv.ParseUint(part, 10, 31)
                if err != nil {
-                       return nil, fmt.Errorf("accessor index %q: %s", part, err)
+                       return nil, fmt.Errorf("accessor index %q: %w", part, err)
                }
 
                result = append(result, int(index))
@@ -559,7 +559,7 @@ func coreFindField(local Type, localAcc coreAccessor, target Type) (_, _ coreFie
                        if localMember.Name == "" {
                                _, ok := localMember.Type.(composite)
                                if !ok {
-                                       return coreField{}, coreField{}, fmt.Errorf("unnamed field with type %s: %s", localMember.Type, ErrNotSupported)
+                                       return coreField{}, coreField{}, fmt.Errorf("unnamed field with type %s: %w", localMember.Type, ErrNotSupported)
                                }
 
                                // This is an anonymous struct or union, ignore it.
diff --git a/internal/btf/ext_info.go b/internal/btf/ext_info.go
index beba1bc..893c8db 100644
--- a/internal/btf/ext_info.go
+++ b/internal/btf/ext_info.go
@@ -269,7 +269,7 @@ func parseExtInfoRelos(r io.Reader, bo binary.ByteOrder, strings stringTable) (m
 
                        accessor, err := parseCoreAccessor(accessorStr)
                        if err != nil {
-                               return nil, fmt.Errorf("accessor %q: %s", accessorStr, err)
+                               return nil, fmt.Errorf("accessor %q: %w", accessorStr, err)
                        }
 
                        relos = append(relos, coreRelo{
diff --git a/internal/btf/types.go b/internal/btf/types.go
index 9b7eca2..812d2a9 100644
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -766,7 +766,7 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (types []Type,
                        for i, btfVal := range rawvals {
                                name, err := rawStrings.LookupName(btfVal.NameOff)
                                if err != nil {
-                                       return nil, nil, fmt.Errorf("get name for enum value %d: %s", i, err)
+                                       return nil, nil, fmt.Errorf("get name for enum value %d: %w", i, err)
                                }
                                vals = append(vals, EnumValue{
                                        Name:  name,
@@ -813,7 +813,7 @@ func inflateRawTypes(rawTypes []rawType, rawStrings stringTable) (types []Type,
                        for i, param := range rawparams {
                                name, err := rawStrings.LookupName(param.NameOff)
                                if err != nil {
-                                       return nil, nil, fmt.Errorf("get name for func proto parameter %d: %s", i, err)
+                                       return nil, nil, fmt.Errorf("get name for func proto parameter %d: %w", i, err)
                                }
                                params = append(params, FuncParam{
                                        Name: name,
diff --git a/internal/elf.go b/internal/elf.go
index 54a4313..88326bb 100644
--- a/internal/elf.go
+++ b/internal/elf.go
@@ -24,7 +24,7 @@ func NewSafeELFFile(r io.ReaderAt) (safe *SafeELFFile, err error) {
                }
 
                safe = nil
-               err = fmt.Errorf("reading ELF file panicked: %s", r)
+               err = fmt.Errorf("reading ELF file panicked: %w", r)
        }()
 
        file, err := elf.NewFile(r)
@@ -44,7 +44,7 @@ func (se *SafeELFFile) Symbols() (syms []elf.Symbol, err error) {
                }
 
                syms = nil
-               err = fmt.Errorf("reading ELF symbols panicked: %s", r)
+               err = fmt.Errorf("reading ELF symbols panicked: %w", r)
        }()
 
        syms, err = se.File.Symbols()
@@ -60,7 +60,7 @@ func (se *SafeELFFile) DynamicSymbols() (syms []elf.Symbol, err error) {
                }
 
                syms = nil
-               err = fmt.Errorf("reading ELF dynamic symbols panicked: %s", r)
+               err = fmt.Errorf("reading ELF dynamic symbols panicked: %w", r)
        }()
 
        syms, err = se.File.DynamicSymbols()
```

cyphar:
Wait, @AkihiroSuda please stop sending reviews -- I'm fixing them all now. :sweat_smile:

AkihiroSuda:
@cyphar Sure (FYI: my draft is at https://github.com/AkihiroSuda/cilium-ebpf/commit/650e2970813ec411b1636aac0c01688db747960f)

AkihiroSuda (approved):

lmb:
Thanks for the PR, error wrapping is a bit muddled at the moment. I'm conflicted about this change: we're not going to make all wrapped errors part of our API contract. That is, if your code relies on unwrapping an error that is not explicitly documented by a function we may well break it in the future. Are you OK with that? @ti-mo do you have opinions on error wrapping vs. no error wrapping?

I also dug into the downstream issue you linked and I wonder whether you could use https://pkg.go.dev/github.com/cilium/ebpf/link#AttachCgroup instead? This should handle the cgroupv2 abstraction for you, including the bpf flags to use.

cyphar:
@lmb We can't use that API because it depends on giving a path to a cgroup, but we have various hardenings such that the dirfd we use is definitely safe to operate on. To be fair, we could use `/proc/self/fd/$fd` but I'd prefer to not have to go through `/proc` if there is a way to use the dirfd directly.

lmb:
> I'd prefer to not have to go through /proc if there is a way to use the dirfd directly.
> As far as I can tell, there doesn't appear to be a nice way for us to detect that the error was "BPF_F_REPLACE is unsupported".

Makes sense. What if we added an fd-only API? From your description it seems like something about BPF_F_REPLACE would still trip you up, but I'm missing context. How does your behaviour differ if F_REPLACE is not supported? Is the problem that the library forces BPF_F_REPLACE if F_ALLOW_MULTI is specified?

> I can reduce this PR to just wrap RawAttachProgram and RawDetachProgram if that would cause less issues for you.

Yes please, it'll make it easier to review and get you unblocked.

cyphar:
@lmb 

> How does your behaviour differ if F_REPLACE is not supported?

We have slightly strange fallback behaviour. If `BPF_F_REPLACE` is not supported, we attach the program then detach the old one (effective an inatomic `BPF_F_REPLACE`). This is okay for us because we're the only people managing the cgroup. I don't expect this fallback behaviour to be in the ebpf library, but being able to detect that "BPF_F_REPLACE isn't supported" was the error would allow us to fallback cleanly.

> Yes please, it'll make it easier to review and get you unblocked.

Sure, I'll do that tomorrow. Thanks.

florianl:
To me this reported issue sounds like a good example for feature probing (similar to the ideas in issue #235 ). Like @lmb I would appreciate it if not all errors would be part of the API.

cyphar:
@florianl It would be nice if we could just directly use the feature probing that this library already does internally (and if we could do that, this PR wouldn't be necessary), but I expect that design discussions on such a feature will probably take some time -- we need some kind of workaround for this fairly quickly so we can do a runc release.

But personally I think it's fairly reasonable to expect that the errors from a "raw program API" would allow you to get the underlying syscall error.

lmb (approved):

ti-mo:
A first proposal for feature probing is being worked on at the moment, but don't expect full coverage for all program and map types right out of the gate, that will take some time. Of course, if there is a strong use case for particular program types like `F_REPLACE`, those should be worked on first if it means we can positively impact (and gain adoption in) a project like runc.

I'm not sure where to go with this PR, though. On the surface, propagating errors all the way down to the user sounds reasonable. Do those errors then suddenly become part of the package API? Well, yes. We're not exporting sentinels, and we don't know which error(s) the kernel will give us. We could only document this like "returns error x when y" in godoc, but that doesn't make for a particularly strong API, as the kernel tends to change from under our feet.

If we want to make that work long-term, adding tests for specific error returns would be a base requirement. @cyphar this includes your most recent changes (wrapping `RawAttachProgram` and `RawDetachProgram` only). If we allow users to become reliant on certain error returns, we need to make sure that keeps working in the future. I get the urgency, but even these one-liners break with the current way of doing things, and we'll have to maintain them as an exception going forward.

Would it be possible to just rely on 'there was an error' for now to kick off your fallback behaviour, or is it really important to know the cause? If you're willing to implement a workaround, you could string-match errors for now until a feature probe API is available. This prevents your particular pain from potentially knock-on effects to other users of the library.

lmb:
Good points @ti-mo.

> Do those errors then suddenly become part of the package API? Well, yes. We're not exporting sentinels, and we don't know which error(s) the kernel will give us.

Like I said, if it's not documented and tested it's not part of the API from my POV. @cyphar I merged the PR so that you are unblocked, but I'd suggest adding a test and documentation if you don't want this to regress.

PR #321, opened 2021-06-09
MapType probe API

rgo3:
Currently only supports default bpfMapCreateAttr values, to showcase
basic functionality. Results are cached by a go map. Tests log durations
to see cache performance. Added some open questions and missing error
interface checks to the comments in the file.

### DIFF
--- a/features/map.go
+++ b/features/map.go
@@ -1,0 +1,182 @@
+package features
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+type MapCache struct {
+	sync.Mutex
+	mapTypes map[ebpf.MapType]error
+}
+
+var (
+	mapCache MapCache
+)
+
+func init() {
+	mapCache.mapTypes = make(map[ebpf.MapType]error)
+}
+
+func createMapTypeAttr(mt ebpf.MapType) *internal.BPFMapCreateAttr {
+	var (
+		keySize        uint32 = 4
+		valueSize      uint32 = 4
+		maxEntries     uint32 = 1
+		innerMapFd     uint32
+		flags          uint32
+		btfKeyTypeID   uint32
+		btfValueTypeID uint32
+		btfFd          uint32
+	)
+
+	// switch on map types to generate correct bpfMapCreateAttr
+	switch mt {
+	case ebpf.StackTrace:
+		// valueSize needs to be sizeof(uint64)
+		valueSize = 8
+	case ebpf.LPMTrie:
+		// keySize and valueSize need to be sizeof(struct{u32 + u8}) + 1 + padding = 8
+		// BPF_F_NO_PREALLOC needs to be set
+		// checked at allocation time for lpm_trie maps
+		keySize = 8
+		valueSize = 8
+		flags = unix.BPF_F_NO_PREALLOC
+	case ebpf.ArrayOfMaps, ebpf.HashOfMaps:
+		// assign invalid innerMapFd to pass validation check
+		// will return EBADF
+		innerMapFd = ^uint32(0)
+	case ebpf.CGroupStorage, ebpf.PerCPUCGroupStorage:
+		// keySize needs to be sizeof(struct{u32 + u64}) = 12 (+ padding = 16)
+		// by using unsafe.Sizeof(int) we are making sure that this works on 32bit and 64bit archs
+		// checked at allocation time
+		var align int
+		keySize = uint32(8 + unsafe.Sizeof(align))
+		maxEntries = 0
+	case ebpf.Queue, ebpf.Stack:
+		// keySize needs to be 0, see alloc_check for queue and stack maps
+		keySize = 0
+	case ebpf.RingBuf:
+		// keySize and valueSize need to be 0
+		// maxEntries needs to be power of 2 and PAGE_ALIGNED
+		// checked at allocation time
+		keySize = 0
+		valueSize = 0
+		maxEntries = uint32(os.Getpagesize())
+	case ebpf.SkStorage, ebpf.InodeStorage, ebpf.TaskStorage:
+		// maxEntries needs to be 0
+		// BPF_F_NO_PREALLOC needs to be set
+		// btf* fields need to be set
+		// see alloc_check for local_storage map types
+		maxEntries = 0
+		flags = unix.BPF_F_NO_PREALLOC
+		btfKeyTypeID = 1   // BTF_KIND_INT
+		btfValueTypeID = 3 // BTF_KIND_ARRAY
+		btfFd = ^uint32(0)
+	}
+
+	return &internal.BPFMapCreateAttr{
+		MapType:        uint32(mt),
+		KeySize:        keySize,
+		ValueSize:      valueSize,
+		MaxEntries:     maxEntries,
+		InnerMapFd:     innerMapFd,
+		Flags:          flags,
+		BTFKeyTypeID:   btfKeyTypeID,
+		BTFValueTypeID: btfValueTypeID,
+		BTFFd:          btfFd,
+	}
+
+}
+
+// HaveMapType probes the running kernel for the availability of the specified map type.
+// Return values have the following semantics:
+//
+//   err == nil: The feature is available.
+//   errors.Is(err, ebpf.ErrNotSupported): The feature is not available.
+//   err != nil: Any errors encountered during probe execution, wrapped.
+//
+// Note that the latter case may include false negatives, and that map creation may succeed
+// despite an error being returned. Some map types cannot reliably be probed and will also
+// return error. Only `nil` and `ebpf.ErrNotSupported` are conclusive.
+//
+// Probe results are cached and persist throughout any process capability changes.
+func HaveMapType(mt ebpf.MapType) error {
+	if err := validateMaptype(mt); err != nil {
+		return err
+	}
+
+	return haveMapType(mt)
+}
+
+func validateMaptype(mt ebpf.MapType) error {
+	if mt > mt.Max() {
+		return os.ErrInvalid
+	}
+
+	if mt == ebpf.StructOpts {
+		// A probe for a StructOpts map has vmlinux BTF requirements we currently cannot meet
+		// Once we figure out how to add a working probe in this package, we can remove this check
+		return errors.New("a probe for MapType StructOpts isn't implemented")
+	}
+
+	return nil
+}
+
+func haveMapType(mt ebpf.MapType) error {
+	mapCache.Lock()
+	defer mapCache.Unlock()
+	err, ok := mapCache.mapTypes[mt]
+	if ok {
+		return err
+	}
+
+	_, err = internal.BPFMapCreate(createMapTypeAttr(mt))
+
+	// For nested and storage map types we accept EBADF as indicator that these maps are supported
+	if errors.Is(err, unix.EBADF) {
+		if isMapOfMaps(mt) || isStorageMap(mt) {
+			err = nil
+		}
+	}
+
+	// EINVAL occurs when attempting to create a map with an unknown type.
+	// E2BIG occurs when BPFMapCreateAttr contains non-zero bytes past the end
+	// of the struct known by the running kernel, meaning the kernel is too old
+	// to support the given map type.
+	if errors.Is(err, unix.EINVAL) || errors.Is(err, unix.E2BIG) {
+		err = ebpf.ErrNotSupported
+	}
+
+	if err != nil && err != ebpf.ErrNotSupported {
+		err = fmt.Errorf("unexpected error during feature probe: %w", err)
+	}
+
+	mapCache.mapTypes[mt] = err
+
+	return err
+}
+
+func isMapOfMaps(mt ebpf.MapType) bool {
+	switch mt {
+	case ebpf.ArrayOfMaps, ebpf.HashOfMaps:
+		return true
+	}
+	return false
+}
+
+func isStorageMap(mt ebpf.MapType) bool {
+	switch mt {
+	case ebpf.SkStorage, ebpf.InodeStorage, ebpf.TaskStorage:
+		return true
+	}
+
+	return false
+}
--- a/features/map.go
+++ b/features/map.go
@@ -12,17 +12,17 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
 
-type MapCache struct {
-	sync.Mutex
-	mapTypes map[ebpf.MapType]error
+func init() {
+	mc.mapTypes = make(map[ebpf.MapType]error)
 }
 
 var (
-	mapCache MapCache
+	mc mapCache
 )
 
-func init() {
-	mapCache.mapTypes = make(map[ebpf.MapType]error)
+type mapCache struct {
+	sync.Mutex
+	mapTypes map[ebpf.MapType]error
 }
 
 func createMapTypeAttr(mt ebpf.MapType) *internal.BPFMapCreateAttr {
@@ -131,9 +131,9 @@ func validateMaptype(mt ebpf.MapType) error {
 }
 
 func haveMapType(mt ebpf.MapType) error {
-	mapCache.Lock()
-	defer mapCache.Unlock()
-	err, ok := mapCache.mapTypes[mt]
+	mc.Lock()
+	defer mc.Unlock()
+	err, ok := mc.mapTypes[mt]
 	if ok {
 		return err
 	}
@@ -163,7 +163,7 @@ func haveMapType(mt ebpf.MapType) error {
 		err = fmt.Errorf("unexpected error during feature probe: %w", err)
 	}
 
-	mapCache.mapTypes[mt] = err
+	mc.mapTypes[mt] = err
 
 	return err
 }
--- a/features/map.go
+++ b/features/map.go
@@ -140,22 +140,26 @@ func haveMapType(mt ebpf.MapType) error {
 
 	_, err = internal.BPFMapCreate(createMapTypeAttr(mt))
 
+	switch {
 	// For nested and storage map types we accept EBADF as indicator that these maps are supported
-	if errors.Is(err, unix.EBADF) {
+	case errors.Is(err, unix.EBADF):
 		if isMapOfMaps(mt) || isStorageMap(mt) {
 			err = nil
 		}
-	}
 
 	// EINVAL occurs when attempting to create a map with an unknown type.
 	// E2BIG occurs when BPFMapCreateAttr contains non-zero bytes past the end
 	// of the struct known by the running kernel, meaning the kernel is too old
 	// to support the given map type.
-	if errors.Is(err, unix.EINVAL) || errors.Is(err, unix.E2BIG) {
+	case errors.Is(err, unix.EINVAL), errors.Is(err, unix.E2BIG):
 		err = ebpf.ErrNotSupported
-	}
 
-	if err != nil && err != ebpf.ErrNotSupported {
+	// EPERM is kept as-is and is not converted or wrapped.
+	case errors.Is(err, unix.EPERM):
+		break
+
+	// Wrap unexpected errors.
+	case err != nil:
 		err = fmt.Errorf("unexpected error during feature probe: %w", err)
 	}
 
--- a/features/map_test.go
+++ b/features/map_test.go
@@ -1,0 +1,87 @@
+package features
+
+import (
+	"fmt"
+	"math"
+	"os"
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+var mapTypeMinVersion = map[ebpf.MapType]string{
+	ebpf.Hash:                "3.19",
+	ebpf.Array:               "3.19",
+	ebpf.ProgramArray:        "4.2",
+	ebpf.PerfEventArray:      "4.3",
+	ebpf.PerCPUHash:          "4.6",
+	ebpf.PerCPUArray:         "4.6",
+	ebpf.StackTrace:          "4.6",
+	ebpf.CGroupArray:         "4.8",
+	ebpf.LRUHash:             "4.10",
+	ebpf.LRUCPUHash:          "4.10",
+	ebpf.LPMTrie:             "4.11",
+	ebpf.ArrayOfMaps:         "4.12",
+	ebpf.HashOfMaps:          "4.12",
+	ebpf.DevMap:              "4.14",
+	ebpf.SockMap:             "4.14",
+	ebpf.CPUMap:              "4.15",
+	ebpf.XSKMap:              "4.18",
+	ebpf.SockHash:            "4.18",
+	ebpf.CGroupStorage:       "4.19",
+	ebpf.ReusePortSockArray:  "4.19",
+	ebpf.PerCPUCGroupStorage: "4.20",
+	ebpf.Queue:               "4.20",
+	ebpf.Stack:               "4.20",
+	ebpf.SkStorage:           "5.2",
+	ebpf.DevMapHash:          "5.4",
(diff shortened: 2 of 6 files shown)


probe.go
```diff
@@ -0,0 +1,62 @@
...
+	// Does this happen automatically once using process dies?
+	_, err := bpfMapCreate(attr)
+
+	// interpret kernel error as own error interface
+	// obviously needs more than just the err != nil check
+	if err != nil {
+		err = internal.ErrNotSupported
+	}
+
+	// store result to cash
```
kolyshkin, line 57: typo: `cache`

features/map.go
```diff
@@ -0,0 +1,118 @@
...
+	_, err := internal.BPFMapCreate(attr)
+
+	// For nested maps we accept EBADF as indicator that nested maps are supported
+	if isNestedMap(mt) && errors.Is(err, unix.EBADF) {
+		err = nil
+	}
+
+	// interpret kernel error as own error interface
+	// obviously needs more than just the err != nil check
+	if err != nil {
```
lmb, line 101: This needs a policy decision around EPERM: do we treat it as "feature not available" and cache it? Programs can change caps over their lifetime after all.

ti-mo: > Programs can change caps over their lifetime after all.

Good point. In the scenario of processes dropping capabilities over their lifetime, and assuming probing always takes place before map creation, I don't see this as a deal breaker. In this case, the cache would contain a false positive (since the probe was run with a larger cap set), and a subsequent call to create the map would still fail with a clear error. Since we wouldn't probe+cache anything as part of an import side effect, the user would have to make sure the proper capability set is active before executing bpf-related logic. Not surprising, but could lead to subtle bugs.

As a solution, we could record the active capability set of the process during every interaction with the cache, and flush if there are any differences. Not sure how expensive that is, though. I would keep this as a later improvement, we should probably aim for a naive-ish initial implementation.

lmb: Agreed, taking the capability set into account seems overkill, simpler is better. What we need is someone to come up with a policy and think through the implications though.

qmonnet (changes requested):
I'm not super familiar with the library, but I've worked on bpftool's probes and I've been following the work from Robin (as a GSoC co-mentor), so here are some comments on the PR.

In addition to the comments below, I would suggest to rework the content of the different commits (possibly squashing them all as the result is not that big), there is not point in adding probe.go to remove it in the following commit for example.

I see a test for the feature. I haven't looked at the CI so I don't know if it will integrate well with it. Ideally we need an environment with all features available so we can check we can detect them? And an environment where we know that a given set of features is missing, so we can check that there are not reported as supported? How can we validate the cache is working? It would also be interesting to be able to test with different capabilities for the process (although the policy on how to deal with `-EPERM` would need to be sorted out first).

rgo3:
Thanks for the review @qmonnet.

I always planned on squashing the commits at the end to have a nice single commit for this work, Timo and I just agreed I should add smaller commits for now documenting my progress and making the review easier while its a draft.

I'll start addressing the comments on Monday morning.

> I see a test for the feature. I haven't looked at the CI so I don't know if it will integrate well with it. Ideally we need an environment with all features available so we can check we can detect them? And an environment where we know that a given set of features is missing, so we can check that there are not reported as supported? 

For now the test was just a way for me to check if my probes are working. Generally I had similar thoughts, I was also wondering if this feature API, or at least the testing of it needs a minimum version for featuers like we have with the FeatureTest abstraction e.g. https://github.com/cilium/ebpf/blob/ca492085341e0e917f48ec30704d5054c5d42ca8/internal/feature.go#L16

> How can we validate the cache is working?

In the earlier commits I had took some time stamps in the tests to validate it, but generally isn't it ok to assume that a Go map works?

features/map_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	for _, tc := range testCases {
+		if skipTestCaseOnOldKernel(tc.minVersion) {
+			t.Logf("Skipped test for MapType %s: requires at least kernel version %s", tc.mapType.String(), tc.minVersion)
+			continue
+		}
+
+		err := HaveMapType(tc.mapType)
+		if err != nil {
+			t.Fatalf("MapType %s isn't supported even though kernel is at least %s", tc.mapType.String(), tc.minVersion)
```
qmonnet, line 76: Note (no action required): A few maps might also depend on the options set to compile the kernel, but looking at the current set, I don't expect this will be an issue in practice (`CONFIG_NET` and `CONFIG_PERF_EVENTS` should be set for most configs I think).

features/map_test.go
```diff
@@ -0,0 +1,80 @@
...
+		{ebpf.XSKMap, "4.18"},
+		{ebpf.SockHash, "4.18"},
+		{ebpf.CGroupStorage, "4.19"},
+		{ebpf.ReusePortSockArray, "4.19"},
+		{ebpf.PerCPUCGroupStorage, "4.20"},
+		{ebpf.Queue, "4.20"},
+		{ebpf.Stack, "4.20"},
+		{ebpf.SkStorage, "5.2"},
+		{ebpf.DevMapHash, "5.4"},
+		// {ebpf.StructOpts, "5.6"}, whats the best way to still test this path?
```
qmonnet, line 62: We could simply check that it returns the expected error?

features/map_test.go
```diff
@@ -0,0 +1,80 @@
+package features
+
+import (
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+)
+
+// similar as in testutils, skipTestCaseOnOldKernel helps to determine wether a testcase can be skipped
```
qmonnet, line 10:
```suggestion
// similar as in testutils, skipTestCaseOnOldKernel helps to determine whether a testcase can be skipped
```

features/map.go
```diff
@@ -0,0 +1,161 @@
...
+		BTFValueTypeID: btfValueTypeID,
+		BTFFd:          btfFd,
+	}
+
+}
+
+// HaveMapType allows probing the availability of a specified ebpf.MapType
+// It will call the syscall to create a dummy map of a given ebpf.MapType at most once
+// storing the result of the first call in a global cache.
+// This potentially can result in false results if the calling process changes its capabilities.
```
qmonnet, line 101: (Side note: Besides a change in capabilities, there's also the case where the kernel rejects the map creation because it hits the memory limit for the user. Retrying at a later time with a different number of objects loaded for the user might result in a different result. I think we can safely ignore that for now, a privileged user should be able to detect this case from the error returned anyway.)

qmonnet (approved):
Looks good to me, thanks for addressing all the feedback. Nice work!

features/map.go
```diff
@@ -0,0 +1,161 @@
...
+		InnerMapFd:     innerMapFd,
+		Flags:          flags,
+		BTFKeyTypeID:   btfKeyTypeID,
+		BTFValueTypeID: btfValueTypeID,
+		BTFFd:          btfFd,
+	}
+
+}
+
+// HaveMapType allows probing the availability of a specified ebpf.MapType
```
ti-mo, line 98: The API contract needs to be documented here. (largely mirroring `FeatureTestFn()`) My suggestion:

```
// HaveMapType probes the running kernel for the availability of the specified map type.
// Return values have the following semantics:
//
//   err == nil: The feature is available.
//   errors.Is(err, ebpf.ErrNotSupported): The feature is not available.
//   err != nil: Any errors encountered during probe execution, wrapped.
//
// Note that the latter case may include false negatives, and that map creation may succeed
// despite an error being returned. Some map types cannot reliably be probed and will also
// return error. Only `nil` and `ebpf.ErrNotSupported` are conclusive.
//
// Probe results are cached and persist throughout any process capability changes.
```

@lmb @qmonnet Any thoughts?

features/map.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// For nested and storage map types we accept EBADF as indicator that these maps are supported
+	if errors.Is(err, unix.EBADF) {
+		if isMapOfMaps(mt) || isStorageMap(mt) {
+			err = nil
+		}
+	}
+
+	if err != nil {
+		err = internal.ErrNotSupported
```
ti-mo, line 125: As discussed in https://github.com/cilium/ebpf/issues/235#issuecomment-876461428, I believe this still needs some extra logic.

- Any unexpected errors need to be wrapped: `fmt.Errorf("unexpected error during feature probe: %w", err)`
- Only expected kernel responses (EINVAL?) should be converted to `ErrNotSupported`.
- This returns a value from `internal` across the package boundary to the caller. `ebpf/prog.go` aliases `internal.ErrNotSupported` to `ebpf.ErrNotSupported`, so that symbol should be returned instead in case it's ever modified.

features/map_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	for _, tc := range testCases {
+		if skipTestCaseOnOldKernel(tc.minVersion) {
+			t.Logf("Skipped test for MapType %s: requires at least kernel version %s", tc.mapType.String(), tc.minVersion)
+			continue
+		}
+
+		err := HaveMapType(tc.mapType)
+		if err != nil {
```
ti-mo, line 75: This can be shortened to `if err := HaveMapType(tc.mapType); err != nil { ... }`.

features/map_test.go
ti-mo, line 76: `t.Fatalf()` should include the error string in case an unexpected error is encountered. (since this is likely to trip in CI)

features/map_test.go
```diff
@@ -0,0 +1,80 @@
...
+			t.Fatalf("MapType %s isn't supported even though kernel is at least %s", tc.mapType.String(), tc.minVersion)
+		}
+	}
+
```
ti-mo, line 79: This is missing negative tests, the `ErrNotSupported` case is never hit as far as I can see. Try creating a map with mt `math.MaxUint32`, which should fail on all kernel versions.

Tip: Go has built-in coverage reports, see: https://blog.golang.org/cover. (esp. the part about `go tool cover -html=coverage.out`) This should help you build coverage.

ti-mo (changes requested):
This is getting in great shape! The error translation needs to be finished up now that we've worked through the design decisions, and with a negative test case this is good to go. :ship:

features/map_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+)
+
+// similar as in testutils, skipTestCaseOnOldKernel helps to determine wether a testcase can be skipped
+// internal.SkipOnOldKernel would skip the entire test which isn't what I want.
+func skipTestCaseOnOldKernel(minVersion string) bool {
```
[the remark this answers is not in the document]
lmb: Why not `t.Run(subcase)` and then use `SkipOnOldKernel`?

features/map_test.go
rgo3: That test case would only improve the coverage slightly because its caught by the `validateMapType()` check. I could invert `TestHaveMapType` to test new features on old kernels where they don't exist yet to provoke a proper EINVAL and resulting ebpf.ErrNotSupported?

features/map.go
lmb: SGTM. We'll have to find a way to document this more generally than on a function comment though. We should make all feature probes respect this contract (and check for it).

types.go
```diff
@@ -89,6 +89,8 @@ const (
 	RingBuf
 	InodeStorage
 	TaskStorage
+	// MaxMapType - Bound enum of MapTypes, has to be last in enum.
+	MaxMapType
```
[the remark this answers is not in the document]
rgo3: How is the `features` package supposed to know the value if I don't export it? Should I rather add a "Getter" like function to the ebpf package to retrieve this value so its not exported as an actual MapType enum value?

features/map.go
rgo3: Maybe We could also document the API contract in the doc.go file for the entire subpackage?

features/map_test.go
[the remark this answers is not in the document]
rgo3: Ah I wanted to have something like @lmb suggests in the first place and wasn't aware I can do it like that. I just tried it and it definitely gives nicer test output + the skip messages are conform with the ones we already have.

features/map.go
ti-mo: For the testing part, we can work with isolated testing packages: declaring `package feature_test` in the header, but with `xxx_test.go` files still in the `feature/` directory. This makes it so the test can emulate an external consumer and has to refer to constants with their fully-qualified specifiers (e.g. `ebpf.ErrNotSupported`). This might help in validating the external API contract.

As for the function comment here, I think it makes sense to declare this documentation on the function itself as it plays nicest with code editors. Package docs often go unnoticed.

features/map.go
```diff
@@ -0,0 +1,161 @@
...
+	return err
+}
+
+func validateMaptype(mt ebpf.MapType) error {
+	if mt >= ebpf.MaxMapType {
+		return os.ErrInvalid
+	}
+
+	if mt == ebpf.StructOpts {
+		// maybe a sentinel like ErrNotImplemented would do the job here?
```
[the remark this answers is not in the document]
ti-mo: I've spun this out into an issue: https://github.com/cilium/ebpf/issues/344.

features/map.go
```diff
@@ -0,0 +1,173 @@
...
+		keySize = 16
+		valueSize = 8
+		maxEntries = 0
+	case ebpf.Queue:
+		fallthrough
+	case ebpf.Stack:
+		keySize = 0
+	case ebpf.StructOpts:
+		// we can't support StructOps probes currently as it will require a valid BTF fd
+		btfVmLinuxValueTypeID = 1
```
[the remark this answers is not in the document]
ti-mo: Thanks @qmonnet, that's useful! Lorenz suggested this might be possible. Which BTF blob are you using there? I've spun this out into https://github.com/cilium/ebpf/issues/344 for now.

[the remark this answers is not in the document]
qmonnet: :+1: I pasted the complete patch on the new issue.

features/map_test.go
ti-mo: Ah, good point. To make this more testable, the current `HaveMapType` could become an exported wrapper around `validateMapType()` and a new `haveMapType()`. This would allow probing to take place from tests without input validation.

>  I could invert TestHaveMapType to test new features on old kernels where they don't exist yet 

That would be the alternative, but as we remove old kernel versions over time, these tests will stop being executed at some point in the future if all kernels we test against are too new.

rgo3: Ok, I'll go with the wrapper approach you are suggesting. This will also fix the currently failing tests.

features/map.go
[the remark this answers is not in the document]
ti-mo: Thanks! One more thing, you mentioned:

> The `btfVmLinuxValueTypeID = 1` is wrong anyway

Why is this? Is this a hardcoded check in the kernel? According to https://www.kernel.org/doc/html/v5.9/bpf/btf.html#type-encoding, type ID 3 is `BTF_KIND_ARRAY`, at least as far as vmlinux is concerned. I think for eBPF programs, the compiler can (and does) assign type IDs 1 and up to any arbitrary type.

If this is really needs to be `BTF_KIND_ARRAY`, we should probably use constants for this.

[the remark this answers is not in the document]
qmonnet: My understanding of BTF is not perfect, but the value we want here is the BTF id corresponding to `struct bpf_struct_ops_tcp_congestion_ops` in kernel's BTF. The value I needed to have it work was `26843`, as readable from:

```
# bpftool btf dump id 1 | grep bpf_struct_ops_tcp_congestion_ops
[26843] STRUCT 'bpf_struct_ops_tcp_congestion_ops' size=256 vlen=3
```

I'm not confident at all that this can be a constant. My guess is that this is fixed for a given kernel BTF object, but ids would probably be numbered differently on a kernel with different definitions.

It doesn't _have_ to point to the BTF of the valid struct (I think?), but if we want to create the map we have to use a `type_id` from which the `st_ops` retrieved with `bpf_struct_ops_find_value()` passes the different checks in `bpf_struct_ops_map_alloc()` in kernel/bpf/bpf_struct_ops.c. I spent some time trying to forge something that would work, but eventually the solution I have in the patch above was way cleaner (and the only one I managed to make work).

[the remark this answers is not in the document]
ti-mo: > My understanding of BTF is not perfect, but the value we want here is the BTF id corresponding to struct bpf_struct_ops_tcp_congestion_ops in kernel's BTF. The value I needed to have it work was 26843

Yes, type IDs are dynamic and will differ between binaries. (The same is true for the size btw, which I see you've hardcoded to `value_size = 256` in your example in the other issue. The struct size can be pulled from the BTF too.)

> I spent some time trying to forge something that would work, but eventually the solution I have in the patch above was way cleaner (and the only one I managed to make work).

Great, thanks for the research! If successfully creating any StructOps map depends on vmlinux BTF being available, then I'm afraid there's no way around it. We can try and fiddle around a bit more, but we'll probably end up querying the BTF tree in this library as well, then.

features/map.go
```diff
@@ -0,0 +1,173 @@
...
+	if err, ok := mc.mapTypes[mt]; ok {
+		defer mc.mu.RUnlock()
+		return err
+	}
+	mc.mu.RUnlock()
+
+	attr := probeMapTypeAttr(mt)
+	_, err := internal.BPFMapCreate(attr)
+
+	// For nested and storage maps we accept EBADF as indicator that nested maps are supported
```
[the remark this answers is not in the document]
ti-mo: Yup, agree with the reasoning overall. It seems like we cannot create StructOps maps without vmlinux BTF available, hence we could already return ErrNotSupported if a valid BTF blob cannot be found.

> If it is present, then we can maybe open a FD on it

Maybe I didn't get this right, but is loading the kernel's own vmlinux BTF blob into the kernel necessary, though? Doesn't the kernel already have the ability to introspect its own types? All that's needed to determine `btfVmLinuxValueTypeID` is parsing the BTF blob in userspace and querying it for the `struct bpf_struct_ops_tcp_congestion_ops` type. After all, isn't that what `btf__find_by_name_kind(btf_vmlinux, "bpf_struct_ops_tcp_congestion_ops", ...` does in your example in #344? `libbpf_find_kernel_btf()` doesn't return an fd.

types.go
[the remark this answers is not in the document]
ti-mo: As discussed offline, one solution is to implement the following:
```go
func (_ MapType) Max() MapType {
  return maxMapType
}
```

It's slightly annoying to use since it requires access to a `MapType` variable, but enables things like `if mt >= mt.Max() { ... }`, which is quite elegant. I much prefer this over having an exported `MaxMapType` const. Any other suggestions @lmb?

features/map.go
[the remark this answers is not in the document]
qmonnet: Not sure I understand.

> Maybe I didn't get this right, but is loading the kernel's own vmlinux BTF blob into the kernel necessary (or even possible), though? Doesn't the kernel already have the ability to introspect its own types?

As far I understand the struct_ops maps rely on the availability of the vmlinux BTF in the kernel. So yes, the vmlinux BTF is loaded into the kernel somehow? But I don't think _we_ should attempt to load it again from user space. It should be possible to retrieve a file descriptor to that BTF object if it's present in the kernel, with `bpf_btf_get_fd_by_id()`? It still requires a way to find the id for kernel BTF, though :thinking:.

Then once we have this fd we could assign it to `attr->btf_fd`, this won't be valid and the map won't be created, but it should return `-EBADF` instead of `-EINVAL`, so we could conclude that struct_ops maps are supported.

features/map.go
```diff
@@ -0,0 +1,178 @@
...
+
+	_, err = internal.BPFMapCreate(createMapTypeAttr(mt))
+
+	// For nested and storage map types we accept EBADF as indicator that these maps are supported
+	if errors.Is(err, unix.EBADF) {
+		if isMapOfMaps(mt) || isStorageMap(mt) {
+			err = nil
+		}
+	}
+
```
ti-mo, line 149:
```suggestion
	// EINVAL occurs when attempting to create a map with an unknown type.
	// E2BIG occurs when BPFMapCreateAttr contains non-zero bytes past the end
	// of the struct known by the running kernel, meaning the kernel is too old
	// to support the given map type.
```

features/map.go
```diff
@@ -0,0 +1,178 @@
...
+	if errors.Is(err, unix.EINVAL) || errors.Is(err, unix.E2BIG) {
+		err = ebpf.ErrNotSupported
+	}
+
+	if err != nil && err != ebpf.ErrNotSupported {
+		err = fmt.Errorf("unexpected error during feature probe %w", err)
```
ti-mo, line 155:
```suggestion
		err = fmt.Errorf("unexpected error during feature probe: %w", err)
```

internal/unix/types_other.go
```diff
@@ -21,6 +21,7 @@ const (
 	ESRCH  = syscall.ESRCH
 	ENODEV = syscall.ENODEV
 	EBADF  = syscall.Errno(0)
+	E2BIG  = syscall.Errno(0x7)
```
ti-mo, line 24: Sorry for the confusion, we've recently decided to use 0 for future additions to non-Linux consts. This const doesn't exist on other platforms, so copying the value Linux uses might cause unintended side effects. (0x7 might be used by another const on a certain platform)

types.go
```diff
@@ -10,6 +10,11 @@ import (
 // that will be initialized in the kernel.
 type MapType uint32
 
+// Max returns the
```
ti-mo, line 13:
```suggestion
// Max returns the latest supported MapType.
```

types.go
```diff
@@ -10,6 +10,11 @@ import (
...
+func (_ MapType) Max() MapType {
+	return maxMapType
```
ti-mo, line 15: I've been thinking about this; we should probably make this return `maxMapType-1`, since the `maxMapType` value points past what's actually supported.

features/map_test.go
```diff
@@ -0,0 +1,87 @@
...
+	ebpf.SkStorage:           "5.2",
+	ebpf.DevMapHash:          "5.4",
+	ebpf.StructOpts:          "5.6", // will skip test for struct ops for now
+	ebpf.RingBuf:             "5.8",
+	ebpf.InodeStorage:        "5.10",
+	ebpf.TaskStorage:         "5.11",
+}
+
+func TestHaveMapType(t *testing.T) {
+	for mt := ebpf.UnspecifiedMap + 1; mt < mt.Max(); mt++ {
```
ti-mo, line 46: Following my comment on `MapType.Max()` below, remember to change this to `mt <= mt.Max()`.

features/map_test.go
```diff
@@ -0,0 +1,87 @@
...
+	ebpf.XSKMap:              "4.18",
+	ebpf.SockHash:            "4.18",
+	ebpf.CGroupStorage:       "4.19",
+	ebpf.ReusePortSockArray:  "4.19",
+	ebpf.PerCPUCGroupStorage: "4.20",
+	ebpf.Queue:               "4.20",
+	ebpf.Stack:               "4.20",
+	ebpf.SkStorage:           "5.2",
+	ebpf.DevMapHash:          "5.4",
+	ebpf.StructOpts:          "5.6", // will skip test for struct ops for now
```
ti-mo, line 39:
```suggestion
	ebpf.StructOpts:          "5.6", // requires vmlinux BTF, skip for now
```

features/map_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+	"math"
+	"os"
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+var testCases = map[ebpf.MapType]string{
```
ti-mo, line 13: This is a (testing) package-global variable, so the name `testCases` might be a bit too generic. Maybe `mapTypeKernels`?

ti-mo (changes requested):
Excellent! Just a few nits left!

features/map.go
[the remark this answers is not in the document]
ti-mo: You're definitely on to something, let's resume this in https://github.com/cilium/ebpf/issues/344!

ti-mo (approved):

PR #325, opened 2021-06-15
internal: export BPFMapCreate syscall wrapper

rgo3:
This commit allows the BPFMapCreate syscall wrapper and BPFMapCreateAttr
type to be used by other subpackages of the library while not publicly
exporting them. 

As discussed in #235 this will help me to move the code in #321 into its own subpackage.
To keep the PR small, I only addressed the function and type I need for now. 

### DIFF
--- a/syscalls.go
+++ b/syscalls.go
@@ -13,17 +13,6 @@ import (
 // Generic errors returned by BPF syscalls.
 var ErrNotExist = errors.New("requested object does not exist")
 
-// bpfObjName is a null-terminated string made up of
-// 'A-Za-z0-9_' characters.
-type bpfObjName [unix.BPF_OBJ_NAME_LEN]byte
-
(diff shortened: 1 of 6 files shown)


syscalls.go
```diff
@@ -13,13 +13,9 @@ import (
...
 var ErrNotExist = errors.New("requested object does not exist")
 
-// bpfObjName is a null-terminated string made up of
-// 'A-Za-z0-9_' characters.
-type bpfObjName [unix.BPF_OBJ_NAME_LEN]byte
-
 // newBPFObjName truncates the result if it is too long.
-func newBPFObjName(name string) bpfObjName {
-	var result bpfObjName
+func newBPFObjName(name string) internal.BPFObjName {
```
lmb, line 17: This should move to `internal` as well I guess.

lmb (approved):

PR #326, opened 2021-06-15
map: fix flaky test on Go 1.17beta1

lmb:
It seems like the output of os.TempDir is now sometimes exactly 32 bytes
long, which makes TestMapPinning fail with

    pin object /sys/fs/bpf/ebpf-test2297582189/: file exists

Fix this by padding to 64 bytes.

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -303,7 +303,7 @@ func TestMapPin(t *testing.T) {
 	// Issue 51: pad path out to a power of two, to avoid having a
 	// trailing zero at the end of the allocation which holds the string.
 	path := tmp + string(filepath.Separator)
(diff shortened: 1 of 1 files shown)


ti-mo:
I'm trying to understand https://github.com/cilium/ebpf/issues/51 and the resulting fix. Isn't this something that can/needs to be fixed in the lib itself? I assume we can't have a pin operation fail because its input path was exactly 32 bytes.. Shouldn't we transparently grow the input buffer transparently to the next-largest power of two if the input string is exactly (power of two)-1 long?

ti-mo (approved):
Snip

PR #327, opened 2021-06-15
cmd/bpf2go: use Go 1.16 embed to include BPF elfs

lmb:
Instead of emitting a (long) string literal use the embed package to
pull in the .o. Users may have to update their .gitignores.

This can only go in after 1.17 has been released and 1.15 is EOL.


ti-mo (approved):

cmd/bpf2go/main.go
```diff
@@ -173,23 +173,17 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
 
 		var tags []string
 		if targetTags := tagsByTarget[target]; targetTags != "" {
-			tags = append(tags, strings.Split(targetTags, " ")...)
```
lmb, line 176: Needs #383 to land

PR #329, opened 2021-06-17
Add feature test for BPF_F_MMAPABLE maps

christarazi:
Introduced in
https://github.com/torvalds/linux/commit/fc9702273e2edb90400a34b3be76f7b08fa3344b.

Signed-off-by: Chris Tarazi <chris@isovalent.com>

### DIFF
--- a/map.go
+++ b/map.go
@@ -316,6 +316,11 @@ func createMap(spec *MapSpec, inner *internal.FD, opts MapOptions, handles *hand
 			return nil, fmt.Errorf("map create: %w", err)
 		}
 	}
+	if spec.Flags&unix.BPF_F_MMAPABLE > 0 {
+		if err := haveMmapableMaps(); err != nil {
+			return nil, fmt.Errorf("map create: %w", err)
+		}
+	}
 
 	attr := internal.BPFMapCreateAttr{
 		MapType:    uint32(spec.Type),
--- a/syscalls.go
+++ b/syscalls.go
@@ -214,6 +214,22 @@ var haveMapMutabilityModifiers = internal.FeatureTest("read- and write-only maps
 	return nil
 })
 
+var haveMmapableMaps = internal.FeatureTest("mmapable maps", "5.5", func() error {
+	// This checks BPF_F_MMAPABLE, which appeared in 5.5 for array maps.
+	m, err := internal.BPFMapCreate(&internal.BPFMapCreateAttr{
+		MapType:    uint32(Array),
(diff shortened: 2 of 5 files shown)


nathanjsweet (approved):

aditighag (changes requested):
🚀 
The tests are failing on a non-unix platform. You would also have to add stubs to `internal/unix/types_other.go`.

christarazi:
Ah thanks @aditighag. I instinctively updated the library thinking that the `BPF_F_MMAPABLE` constant wasn't available, but didn't notice that the `darwin` builds were failing. Should be good now.

map.go
```diff
@@ -311,10 +311,15 @@ func createMap(spec *MapSpec, inner *internal.FD, opts MapOptions, handles *hand
 		}
 	}
 
-	if spec.Flags&(unix.BPF_F_RDONLY_PROG|unix.BPF_F_WRONLY_PROG) > 0 || spec.Freeze {
+	switch {
```
ti-mo, line 314: I don't think rdonly and mmapable flags are mutually exclusive. Selftest `test_mmap.c` contains a map with flags `BPF_F_MMAPABLE | BPF_F_RDONLY_PROG`, so we should probably always execute both checks here.

ti-mo (changes requested):
Thanks! Requested a small change.

@rgo3 FYI, here's an extra test.

map.go
christarazi: Ah, good catch, fixed!

lmb (approved):

syscalls.go
```diff
@@ -214,6 +214,22 @@ var haveMapMutabilityModifiers = internal.FeatureTest("read- and write-only maps
...
+var haveMmapableMaps = internal.FeatureTest("mmapable maps", "5.5", func() error {
+	// This checks BPF_F_MMAPABLE, which appeared in 5.5 for array maps.
+	m, err := internal.BPFMapCreate(&internal.BPFMapCreateAttr{
+		MapType:    uint32(Array),
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 1,
+		Flags:      unix.BPF_F_MMAPABLE,
+	})
+	if err != nil {
```
lmb, line 226: Could this return EPERM?

christarazi: From looking at the kernel code for all references to `BPF_F_MMAPABLE`, I didn't see such a return path.

PR #334, opened 2021-07-01
syscalls: do not conceal BPF_PROG_* syscall errors

cyphar:
The error wrapping code for BPF_PROG_* syscall-related errors would mask
the true source of all underlying syscall errors, which meant that you
couldn't detect several fairly important cases (such as -EACESS and
-EPERM). It seems that this behaviour wasn't intentional (prior to
commit de57e915ff34, the behaviour was to bubble up the syscall error)
and the similar wrapping of BPF_MAP_* errors did bubble up the syscall
error too.

This is needed for runc to be able to detect permission errors due to
SELinux labels blocking certain operations (mainly NewProgramFromID),
and unifies the behaviour for BPF_PROG_* and BPF_MAP_* syscalls.

It turns out that wrapMapError doesn't actually wrap the either error,
but @lmb said they will come up with a better long term solution, so
leave this alone for now.

Ref: opencontainers/runc#3055
Fixes: de57e915ff34 ("Add *GetNextID")
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/syscalls.go
+++ b/syscalls.go
@@ -357,10 +357,9 @@ func wrapObjError(err error) error {
 		return nil
 	}
 	if errors.Is(err, unix.ENOENT) {
-		return fmt.Errorf("%w", ErrNotExist)
+		err = ErrNotExist
 	}
-
-	return errors.New(err.Error())
+	return fmt.Errorf("%w", err)
 }
 
 func wrapMapError(err error) error {
--- a/syscalls_test.go
+++ b/syscalls_test.go
@@ -1,10 +1,12 @@
 package ebpf
 
 import (
+	"errors"
 	"strings"
 	"testing"
 
 	"github.com/cilium/ebpf/internal/testutils"
+	"golang.org/x/sys/unix"
 )
 
 func TestObjNameCharacters(t *testing.T) {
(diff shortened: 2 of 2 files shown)


syscalls.go
```diff
@@ -357,10 +357,9 @@ func wrapObjError(err error) error {
 		return nil
 	}
 	if errors.Is(err, unix.ENOENT) {
-		return fmt.Errorf("%w", ErrNotExist)
```
lmb, line 360: This is on purpose, it forces `errors.Is` rather than directly comparing the error.

syscalls.go
```diff
@@ -357,10 +357,9 @@ func wrapObjError(err error) error {
...
+		return ErrNotExist
 	}
-
-	return errors.New(err.Error())
+	return err
```
lmb, line 362: Same here, this should be wrapped.

syscalls.go
cyphar: Should `wrapMapError` be doing that as well then? I was just mirroring what `wrapMapError` does.

syscalls.go
cyphar: Ditto here.

syscalls.go
lmb: Yeah that is kind of bogus. I think in practice callers of wrapMapError add context and so provide the necessary wrapping, but really this function should wrap explicitly.

Seeing as you're adding more and more of these escape hatches I think I need to come up with a better way of dealing with syscall errors. For now, can you ensure that wrapObjErr does wrapping? I'll make a PR with a better long-term solution.

cyphar: Will do, thanks!

cyphar:
#336 fixes this in a much nicer way.

PR #335, opened 2021-07-01
link: support creating bpf_map_elem iterators

lmb:
We need to pass the target map fd in a separate struct to iterate
the elements of a map, e.g. for bpf_map_elem and sockmap iterators.

### DIFF
--- a/link/iter.go
+++ b/link/iter.go
@@ -3,8 +3,10 @@ package link
 import (
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


alban:
Thanks, it works for me!

I tested that the iterator successfully prints the content of my hash map. I also tested that the BPF program is able to selectively delete some entries of the hash map depending on the content of another map.

PR #336, opened 2021-07-01
improve error wrapping

lmb:
The library is inconsistent in how it wraps errors from syscalls. Sometimes it obscures the underlying error, sometimes it wraps it, sometimes it replaces it with a more specific error. This PR introduces a new approach:

- We always wrap errors returned by SYS_BPF. This means that err == unix.EFOO will never work.
- We always include the original errno when introducing a more specific error. This means that errors.Is(err, unix.EFOO) will always work.

This approach gives users access to syscall.Errno without painting us into a corner.

### DIFF
--- a/internal/syscall.go
+++ b/internal/syscall.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"runtime"
+	"syscall"
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal/unix"
@@ -61,7 +62,7 @@ func BPF(cmd BPFCmd, attr unsafe.Pointer, size uintptr) (uintptr, error) {
 
 	var err error
 	if errNo != 0 {
-		err = errNo
+		err = wrappedErrno{errNo}
 	}
 
 	return r1, err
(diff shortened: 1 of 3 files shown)


lmb:
@cyphar please take a look, I think this should solve your problem.

cyphar (approved):
This seems very reasonable to me, and means that even the `Err*`-wrapped errors can be compared against either the syscall value or their `Err*` equivalent. Thanks for this!

cyphar:
Is it possible to get a new release with this patch? (In the meantime, I'll update the runc PR to use HEAD.)

internal/syscall.go
```diff
@@ -226,3 +226,20 @@ type wrappedErrno struct {
...
 }
+
+type syscallError struct {
+	error
+	errno syscall.Errno
+}
+
+func SyscallError(err error, errno syscall.Errno) error {
+	return &syscallError{err, errno}
+}
```
kolyshkin, line 237: Why not export the type and use it directly? This is what golang stdlib does with types such as `os.PathError`.

kolyshkin: I mean, instead of e.g `return syscall.SyscallError(err, errno)` you can do `return &syscall.SyscallError{err, errno}` and avoid an unnecessary function call (not sure if golang will inline it). Since this is an internal type, you're not exporting it.

PR #341, opened 2021-07-10
Mark PerCPUCGroupStorage as per-cpu

yadutaf:
Currently, using "some_map->Lookup" on a PerCPUCGroupStorage only allows retrieving the first value of the per-cpu storage area. This PR adds the missing declaration.

### DIFF
--- a/internal/testutils/cgroup.go
+++ b/internal/testutils/cgroup.go
@@ -1,17 +1,59 @@
 package testutils
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
+	"strings"
+	"sync"
 	"testing"
 
 	"golang.org/x/sys/unix"
 )
 
+var cgroup2 = struct {
+	once sync.Once
+	path string
+	err  error
+}{}
+
+func cgroup2Path() (string, error) {
+	cgroup2.once.Do(func() {
+		mounts, err := ioutil.ReadFile("/proc/mounts")
+		if err != nil {
+			cgroup2.err = err
+			return
+		}
+
+		for _, line := range strings.Split(string(mounts), "\n") {
+			mount := strings.SplitN(line, " ", 3)
+			if mount[0] == "cgroup2" {
+				cgroup2.path = mount[1]
+				return
+			}
+
+			continue
+		}
+
+		cgroup2.err = errors.New("cgroup2 not mounted")
+	})
+
+	if cgroup2.err != nil {
+		return "", cgroup2.err
+	}
+
+	return cgroup2.path, nil
+}
+
 func CreateCgroup(tb testing.TB) *os.File {
 	tb.Helper()
 
-	cgdir, err := ioutil.TempDir("/sys/fs/cgroup/unified", "ebpf-link")
+	cg2, err := cgroup2Path()
+	if err != nil {
+		tb.Fatal("Can't locate cgroup2 mount:", err)
+	}
+
+	cgdir, err := ioutil.TempDir(cg2, "ebpf-link")
 	if err != nil {
 		tb.Fatal("Can't create cgroupv2:", err)
 	}
--- a/internal/testutils/cgroup.go
+++ b/internal/testutils/cgroup.go
@@ -1,0 +1,40 @@
+package testutils
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"golang.org/x/sys/unix"
+)
+
+func CreateCgroup(tb testing.TB) *os.File {
+	tb.Helper()
+
+	cgdir, err := ioutil.TempDir("/sys/fs/cgroup/unified", "ebpf-link")
+	if err != nil {
+		tb.Fatal("Can't create cgroupv2:", err)
+	}
+
+	cgroup, err := os.Open(cgdir)
+	if err != nil {
+		os.Remove(cgdir)
+		tb.Fatal(err)
+	}
+	tb.Cleanup(func() {
+		cgroup.Close()
+		os.Remove(cgdir)
+	})
+
+	return cgroup
+}
+
+func GetCgroupIno(t *testing.T, cgroup *os.File) uint64 {
+	cgroupStat := unix.Stat_t{}
+	err := unix.Fstat(int(cgroup.Fd()), &cgroupStat)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return cgroupStat.Ino
+}
--- a/link/link_test.go
+++ b/link/link_test.go
@@ -7,8 +7,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"strings"
-	"sync"
 	"testing"
 
 	"github.com/cilium/ebpf"
@@ -16,12 +14,6 @@ import (
 	"github.com/cilium/ebpf/internal/testutils"
 	"github.com/cilium/ebpf/internal/unix"
 )
-
-var cgroup2Mount = struct {
-	once sync.Once
-	path string
-	err  error
-}{}
 
 func TestRawLink(t *testing.T) {
 	cgroup, prog := mustCgroupFixtures(t)
@@ -98,26 +90,7 @@ func mustCgroupFixtures(t *testing.T) (*os.File, *ebpf.Program) {
 
 	testutils.SkipIfNotSupported(t, haveProgAttach())
 
-	prog := mustCgroupEgressProgram(t)
-	cgdir, err := ioutil.TempDir(cgroup2Path(t), "ebpf-link")
-	if err != nil {
-		prog.Close()
-		t.Fatal("Can't create cgroupv2:", err)
-	}
-
-	cgroup, err := os.Open(cgdir)
-	if err != nil {
-		prog.Close()
-		os.Remove(cgdir)
-		t.Fatal(err)
-	}
-	t.Cleanup(func() {
-		prog.Close()
-		cgroup.Close()
-		os.Remove(cgdir)
-	})
-
-	return cgroup, prog
+	return testutils.CreateCgroup(t), mustCgroupEgressProgram(t)
 }
 
 func mustCgroupEgressProgram(t *testing.T) *ebpf.Program {
@@ -135,34 +108,12 @@ func mustCgroupEgressProgram(t *testing.T) *ebpf.Program {
 	if err != nil {
 		t.Fatal(err)
 	}
-	return prog
-}
 
-func cgroup2Path(t *testing.T) string {
-	cgroup2Mount.once.Do(func() {
-		mounts, err := ioutil.ReadFile("/proc/mounts")
-		if err != nil {
-			cgroup2Mount.err = err
-			return
-		}
-		for _, line := range strings.Split(string(mounts), "\n") {
-			mount := strings.SplitN(line, " ", 3)
-			if mount[0] == "cgroup2" {
-				cgroup2Mount.path = mount[1]
-				return
-			}
-
-			continue
-		}
-
-		cgroup2Mount.err = errors.New("cgroup2 not mounted")
+	t.Cleanup(func() {
+		prog.Close()
 	})
 
-	if cgroup2Mount.err != nil {
-		t.Fatal(cgroup2Mount.err)
-	}
-
-	return cgroup2Mount.path
+	return prog
 }
 
 type testLinkOptions struct {
--- a/map_test.go
+++ b/map_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"unsafe"
 
+	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/btf"
 	"github.com/cilium/ebpf/internal/testutils"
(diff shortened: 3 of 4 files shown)


ti-mo:
Thanks!

@lmb Do we need extra tests for this?

lmb (changes requested):
Yes, `TestPerCPUMarshaling` needs updating.

yadutaf:
I pushed a proposed test in a dedicated test function. Adding per-cpu cgroup storage to the existing `TestPerCPUMarshaling` proved a bit difficult due to the way it works. According [to the documentation](https://www.kernel.org/doc/html/latest/bpf/map_cgroup_storage.html), the cgroup storage is only allocated when an eBPF program is attached to the target cgroup. Additionally, `MaxEntries` can not be set for this kind of storage.

Instead of stacking "ifs" in `TestPerCPUMarshaling`  and attaching a dummy program, I would propose to add a dedicated test which focuses on the unmarshaling step. It has the downside of not exercising the full `marshall->kernel->unmarshal` chain, but would it be acceptable to consider this covered by the existing tests ?

lmb:
> I pushed a proposed test in a dedicated test function.

Yeah, good idea.

> I would propose to add a dedicated test which focuses on the unmarshaling step.

I think that's the part I'm least worried about: it's already being exercised by the other per-CPU maps. The important bit is to check that the kernel interface matches what we expect, so do a real lookup from the map. Would it be enough to reuse `mustCgroupFixtures` and attach an empty program to that cgroup to make the lookup from the map feasible?

yadutaf:
> The important bit is to check that the kernel interface matches what we expect.

Totally makes sense.

> Would it be enough to reuse mustCgroupFixtures ?

I tried that, but the CGroup storage area is only allocated when attaching a program which uses this storage. Thus, the attached program must at least call `bpf_get_local_storage` with the allocated map.

I pushed a new revision with the tests moved to `map_test.go` to get access to the helpers without getting a circular dependency and a couple new helpers.

yadutaf:
Tests failed on Linux 5.4. I'm not sure of the real reason. I suspect this is related to the changes in the way CGroup local storage is allocated in kernel commit 7d9c3427894fe70d1347b4820476bf37736d2ff0 in 5.9. Honestly, this is really a wild guess.

Would it be acceptable to raise the minimum Linux version for this test to 5.9 and thus skip the 5.4 tests ?

link/cgroup_test.go
```diff
@@ -4,9 +4,18 @@ import (
 	"testing"
 
 	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/testutils"
+
+	"golang.org/x/sys/unix"
 )
 
+type BpfCgroupStorageKey struct {
```
lmb, line 13: Please unexport.

link/cgroup_test.go
```diff
@@ -75,3 +84,66 @@ func TestLinkCgroup(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var mapKey = &BpfCgroupStorageKey{
+		CgroupInodeId: cgroupStat.Ino,
+		AttachType:    ebpf.AttachCGroupInetEgress,
+	}
+
+	values := []uint64{1, 2}
```
lmb, line 132: This needs to have length numCPU, I think. Otherwise the test will fail on machines with more than two CPUs.
```suggestion
	values := make([]uint64, numCPU)
	copy(values, []uint64{1, 2})
```

link/link_test.go
```diff
@@ -91,6 +91,16 @@ func mustCgroupFixtures(t *testing.T) (*os.File, *ebpf.Program) {
 	testutils.SkipIfNotSupported(t, haveProgAttach())
 
 	prog := mustCgroupEgressProgram(t)
+	cgroup := mustCgroupFixturesWithProgram(t, prog)
+
+	return cgroup, prog
+}
+
+func mustCgroupFixturesWithProgram(t *testing.T, prog *ebpf.Program) *os.File {
```
lmb, line 99: I kind of expected this function to attach the program to the cgroup, but that doesn't seem to happen? Why pass program here?

link/link_test.go
```diff
@@ -91,6 +91,16 @@ func mustCgroupFixtures(t *testing.T) (*os.File, *ebpf.Program) {
...
+	t.Helper()
+
+	testutils.SkipIfNotSupported(t, haveProgAttach())
```
lmb, line 102: This skip doesn't make sense since you're splitting program creation out of the helper. Can you move this call into the functions that actually call `NewProgram`?

link/link_test.go
lmb: I think you can move this function into `internal/testutils/cgroup.go` and then keep `TestCgroupPerCPUStorageMarshaling` in `map_test.go`. That resolves your import cycle while keeping the test closer to the other per cpu tests.

link/cgroup_test.go
```diff
@@ -75,3 +84,66 @@ func TestLinkCgroup(t *testing.T) {
...
+	}
+
+	var retrieved []uint64
+	if err := arr.Lookup(mapKey, &retrieved); err != nil {
+		t.Fatalf("Can't retrieve cgroup %s storage: %s", cgroup.Name(), err)
+	}
+
+	for i, want := range []uint64{1, 2} {
+		if retrieved[i] == 0 {
+			t.Error("First item is 0")
```
lmb, line 144: Nit: "first" is incorrect here.

link/link_test.go
```diff
@@ -130,6 +140,27 @@ func mustCgroupEgressProgram(t *testing.T) *ebpf.Program {
...
+	t.Helper()
+
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type:       ebpf.CGroupSKB,
+		AttachType: ebpf.AttachCGroupInetEgress,
+		License:    "MIT",
+		Instructions: asm.Instructions{
+			asm.LoadMapPtr(asm.R1, arr.FD()),
+			asm.Mov.Imm(asm.R2, 0),
+			asm.FnGetLocalStorage.Call(),
```
lmb, line 153: I'm assuming that this will only accept a local storage map? If that's true it would make sense to  move creating of the array here.

lmb (commented):
> Would it be acceptable to raise the minimum Linux version for this test to 5.9 and thus skip the 5.4 tests ?

We can do that once we can explain the test failure, I'd say. Otherwise how do we know that this isn't a bug in the library?

FWIW it looks like on 5.4 we need an insert from the kernel side before user space can do a put? https://elixir.bootlin.com/linux/v5.4.131/source/kernel/bpf/local_storage.c#L147 If that's the case you can add a skip and document why skipping is ok.

link/cgroup_test.go
yadutaf: Sure. Could be worth promoting to the main library with proper helpers in the future though.

link/cgroup_test.go
yadutaf: This does not appear to be necessary. This test passes fine on my machine with 4 cpus.

link/link_test.go
yadutaf: I tried to follow your suggestion, but ended up with another go compiler import cycle. I'm not sure what to do / if this is worth the trouble. I'll push the (broken) WIP in a separate commit.

link/link_test.go
yadutaf: Absolutely. I'll clean this up.

lmb:
Please take a look at #348 I think you can keep all the map / prog creation bits in map_test.go

yadutaf:
Thanks for your suggestion. Following it solves 2 out of 3 import cycles. However, there is still an issue because `map_test.go` imports `github.com/cilium/ebpf/link` which, in turn, imports again `github.com/cilium/ebpf`. This import of the 'link' module is necessary to attach the dummy program to the test cgroup while the dummy program is needed to trigger the cgroup local storage allocation on kernel side:

```
go test -exec sudo  ./...
# github.com/cilium/ebpf
package github.com/cilium/ebpf (test)
	imports github.com/cilium/ebpf/link
	imports github.com/cilium/ebpf: import cycle not allowed in test
FAIL	github.com/cilium/ebpf [setup failed]
ok  	github.com/cilium/ebpf/asm	0.093s
ok  	github.com/cilium/ebpf/cmd/bpf2go	1.572s
ok  	github.com/cilium/ebpf/internal	0.070s
ok  	github.com/cilium/ebpf/internal/btf	0.720s
?   	github.com/cilium/ebpf/internal/testutils	[no test files]
?   	github.com/cilium/ebpf/internal/unix	[no test files]
ok  	github.com/cilium/ebpf/link	1.994s
ok  	github.com/cilium/ebpf/perf	0.275s
FAIL
```

I see at least 2 possible ways forward:

1. We can accept to move this test to `link_test.go` where we do not have this import cycle. I understand this is not the ideal place to exercise the (un)marshaling, but in a sense it also exercises a great deal of the link as well, including the intended side effect of cgroup local storage allocation.
2. Alternatively, we could spend some time trying to directly call the low level `internal.BPFProgAttach` which is not part of the link module.

yadutaf:
Well, moving to the internal calls proved to be much easier that I expected. This leaves us with the 5.4 test issue. I'm trying to dig in the kernel source code, but would clearly appreciate the help from a kernel side eBPF expert here...

lmb:
> However, there is still an issue because map_test.go imports github.com/cilium/ebpf/link which, in turn, imports again github.com/cilium/ebpf.

Ah, I didn't realize what was happening. Apologies.

> This leaves us with the 5.4 test issue. I'm trying to dig in the kernel source code, but would clearly appreciate the help from a kernel side eBPF expert here...

I took another look at the code, it's fine to add a skip for kernels that don't have https://github.com/torvalds/linux/commit/7d9c3427894fe70d1347b4820476bf37736d2ff0 so basically only run the test on v5.9. The UAPI shouldn't change between kernel versions, and it's not worth spending the time to understand why 5.4 is weird, IMO.

yadutaf:
Thanks! Test updated with the skip.

map_test.go
```diff
@@ -1033,6 +1040,93 @@ func TestPerCPUMarshaling(t *testing.T) {
...
+	if numCPU < 2 {
+		t.Skip("Test requires at least two CPUs")
+	}
+	testutils.SkipOnOldKernel(t, "5.9", "per-CPU CGoup storage with write from user space support")
+
+	cgroup := testutils.CreateCgroup(t)
+
+	arr, err := NewMap(&MapSpec{
+		Type:      PerCPUCGroupStorage,
+		KeySize:   16,
```
ti-mo, line 1057: If the map key is a `bpfCgroupStorageKey`, couldn't `unsafe.Sizeof()` be used instead? Then there's no magic constant.

map_test.go
```diff
@@ -31,6 +32,12 @@ var (
 	}
 )
 
+type bpfCgroupStorageKey struct {
```
ti-mo, line 35: Is this a necessity? Feels a bit out of place declaring this in the test suite. If this key is a requirement for being able to write to a per-CPU-Cgroup map, shouldn't this be exposed from the package instead?

ti-mo (commented):
A few nits, but I'm wondering about the bigger picture. Even after this patch, PerCPUCGroupStorage remains a bit difficult to use.

internal/testutils/cgroup.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+	"testing"
+
+	"golang.org/x/sys/unix"
+)
+
+func CreateCgroup(tb testing.TB) *os.File {
+	tb.Helper()
+
+	cgdir, err := ioutil.TempDir("/sys/fs/cgroup/unified", "ebpf-link")
```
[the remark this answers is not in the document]
lmb: There is a fix in https://github.com/cilium/ebpf/pull/288/commits/94eca08ea7513611523023cfd6db0a84ce7fe6be maybe we can ask Zvi to create a separate PR for it.

map_test.go
lmb: We can just move it to the test itself in that case. I'm against adding this as an exported API:

- Seems like newer kernels accept just uint64
- Adding key types for every map is going to get unwieldy
- Adds a custom exported type for a niche (I'm assuming?) use case

internal/testutils/cgroup.go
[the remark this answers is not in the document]
yadutaf: It would actually be helpful to export such a helper in the link package: This is boilerplate code that all users targeting cgroups will need.

Would it be acceptable to consider this as "not a regression" and "works when run with the test helper + in the CI" ?

map_test.go
yadutaf: Sure.

map_test.go
yadutaf: If I understand correctly (and that's a big "if"), when used, the attach type field allows multiple eBPF programs of different type to have dedicated cgroup local storage. On the other hand, without this field, the eBPF programs would be forced to share their storage area.

If so, I'd suggest a helper to create the key from a cgroup path and either a prog or a link. Would this be more acceptable ? If so, I may propose a patch in a further PR later in August (I have some holidays coming).

internal/testutils/cgroup.go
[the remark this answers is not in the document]
zeffron: Fixed the onetime error issue with 94eca08 and made #352. I'm happy to turn this into an export if that's desired, but left it as a unit test function right now to keep scope minimized.

lmb (approved):

internal/testutils/cgroup.go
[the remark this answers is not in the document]
ti-mo: I've just merged https://github.com/cilium/ebpf/pull/352, will give this a rebase.

map_test.go
ti-mo: Let's leave this as is for now. I do think the library should provide a concrete type for assembling composite map keys to work with sharded maps like these. We currently don't have the right interfaces for interacting with maps that would make this user-friendly. I have some ideas on how to improve this, but not to be discussed here.

ti-mo (approved):
Thanks!

lmb (approved):

Issue #343, opened 2021-07-13
btf: `BTF_KIND_FLOAT` support is missing


mauriciovasquezbernal:
https://github.com/torvalds/linux/commit/8fd886911a6a9 added support for a new `BTF_KIND_FLOAT` type. This library is missing this support and hence BTF related functionality is not working on Linux 5.13:

```
... load kernel spec: can't read types: type id 9468: unknown kind: Unknown
```

lmb:
This will need changes to:
* `readTypes`: add kindFloat switch case
* `inflateRawTypes`: add kindFloat switch case
* Add `type Float struct` to package `btf`

mauriciovasquezbernal:
I gave it a try in https://github.com/cilium/ebpf/pull/345 but I haven't been able to update `internal/btf/testdata/vmlinux-btf.gz` because I get so many errors when trying to compile that kernel using https://github.com/cilium/ci-kernels. 

I have the following diff:
```
$ git diff
diff --git make.sh make.sh
index c96f2b6..b705c44 100755
--- make.sh
+++ make.sh
@@ -30,7 +30,7 @@ fetch_and_configure() {
        echo "${src_dir}"
 }
 
-readonly kernel_versions=("4.9.266" "4.14.230" "4.19.187" "5.4.112" "5.10.30")
+readonly kernel_versions=("4.9.266" "4.14.230" "4.19.187" "5.4.112" "5.10.30" "5.13.1")
 for kernel_version in "${kernel_versions[@]}"; do
        series="$(echo "$kernel_version" | cut -d . -f 1-2)"
```

When I run `make.sh`, the compilation appears to work fine but then the following error happens:

```
...
WARNING: Symbol version dump "Module.symvers" is missing.
         Modules may not have dependencies or modversions.
         You may get many unresolved symbol warnings.
  MODPOST /home/mvb/kinvolk/ebpf/ci-kernels/build/linux-5.13.1/tools/testing/selftests/bpf/bpf_testmod/Module.symvers
WARNING: modpost: "__x86_indirect_alt_call_rax" [/home/mvb/kinvolk/ebpf/ci-kernels/build/linux-5.13.1/tools/testing/selftests/bpf/bpf_testmod/bpf_testmod.ko] undefined!
[...]

PR #345, opened 2021-07-14
bpf: Add support for BTF_KIND_FLOAT

mauriciovasquezbernal:
Fix BTF failure on Linux 5.13  as https://github.com/torvalds/linux/commit/8fd886911a6a9 added support for a new `BTF_KIND_FLOAT` type. 

Fixes https://github.com/cilium/ebpf/issues/343


lmb:
@mauriciovasquezbernal can you try rebasing on top of #346?

lmb (approved):

mauriciovasquezbernal:
@lmb done!

ti-mo:
Commenting out both `case kindFloat` doesn't make `TestParseVmlinux` fail, is that expected? Do we need a running 5.13 to exercise this path?

lmb:
I just did the following quick hack:

```
diff --git a/Makefile b/Makefile
index 5dd342c..5217627 100644
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # The development version of clang is distributed as the 'clang' binary,
 # while stable/released versions have a version number attached.
 # Pin the default clang to a stable version.
-CLANG ?= clang-12
+CLANG ?= clang-13
 CFLAGS := -target bpf -O2 -g -Wall -Werror $(CFLAGS)
 
 # Obtain an absolute path to the directory of the Makefile.
diff --git a/internal/btf/testdata/relocs-eb.elf b/internal/btf/testdata/relocs-eb.elf
index a55edc4..0183d49 100644
Binary files a/internal/btf/testdata/relocs-eb.elf and b/internal/btf/testdata/relocs-eb.elf differ
diff --git a/internal/btf/testdata/relocs-el.elf b/internal/btf/testdata/relocs-el.elf
index 826fb98..a08189c 100644
Binary files a/internal/btf/testdata/relocs-el.elf and b/internal/btf/testdata/relocs-el.elf differ
diff --git a/internal/btf/testdata/relocs.c b/internal/btf/testdata/relocs.c
index 24fae51..b32e04f 100644
--- a/internal/btf/testdata/relocs.c
+++ b/internal/btf/testdata/relocs.c
@@ -223,7 +223,7 @@ __section("socket_filter/fields") int fields() {
 
 struct ambiguous {
 	int _1;
-	char _2;
+	float _2;
 };
 
 struct ambiguous___flavour {

```

Which gives me:

```
$ make internal/btf/testdata/relocs-eb.elf
clang-13 -target bpf -O2 -g -Wall -Werror  -mbig-endian -c internal/btf/testdata/relocs.c -o internal/btf/testdata/relocs-eb.elf
$ ~/dev/bpf-next/tools/bpf/bpftool/bpftool btf dump file internal/btf/testdata/relocs-eb.elf | grep float
[36] FLOAT 'float' size=4
```

Problem: we need to at least include clang-13 in the docker builder. Then we need to either switch the default to clang-13 (I'd rather not...) or add a new file `internal/btf/testdata/float.c` and add `internal/btf/testdata/float-clang-13` to `TARGETS` in Makefile. Finally, add a test that loads `float-clang-13*.elf`.

Thoughts?

ti-mo:
@lmb Yes, we should add floats to testdata, no argument there. I assume clang12 doesn't emit BTF for floats yet?

I'd rather have our vmlinux BTF contain floats as well, since it's a much more complex blob. @mauriciovasquezbernal how/where did you come across this error?

@lmb Maybe our (your :sweat_smile:) pahole version used to generate the vmlinux BTF is out of date?

mauriciovasquezbernal:
> @mauriciovasquezbernal how/where did you come across this error?

@ti-mo I tried to load a BPF program that uses CO-RE in kernel `Linux mvb-desktop 5.13.0-051300-generic #202106272333 SMP Sun Jun 27 23:36:43 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`. 

Something like this fails in my computer without this patch: 

```
$ cat bpf/iterators.bpf.c
#include "vmlinux.h"

#include <bpf/bpf_helpers.h>
#include <bpf/bpf_core_read.h>

SEC("tracepoint/syscalls/sys_enter_execve")
int tracepoint__syscalls__sys_enter_execve(struct trace_event_raw_sys_enter* ctx)
{
	struct task_struct *task = (struct task_struct*)bpf_get_current_task();
	pid_t ppid = (pid_t)BPF_CORE_READ(task, real_parent, tgid);

	return 0;
}

char LICENSE[] SEC("license") = "Dual BSD/GPL";
```

```
$ cat main.go
// +build linux

package main

import (
	"log"
	"os"
	"os/signal"
	"syscall"

	"golang.org/x/sys/unix"
)

//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang-12 Iterators ./bpf/iterators.bpf.c

func main() {
	// Subscribe to signals for terminating the program.
	stopper := make(chan os.Signal, 1)
	signal.Notify(stopper, os.Interrupt, syscall.SIGTERM)

	// Increase the rlimit of the current process to provide sufficient space
	// for locking memory for the eBPF map.
	if err := unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{
		Cur: unix.RLIM_INFINITY,
		Max: unix.RLIM_INFINITY,
	}); err != nil {
		log.Fatalf("failed to set temporary rlimit: %v", err)
	}

	// Load pre-compiled programs and maps into the kernel.
	objs := IteratorsObjects{}
	if err := LoadIteratorsObjects(&objs, nil); err != nil {
		log.Fatalf("loading objects: %v", err)
	}
	defer objs.Close()
}
```

```
$ go generate && go build . && sudo ./iterators && echo $?
Compiled /home/mvb/kinvolk/ebpf/iterators/iterators_bpfel.o
Wrote /home/mvb/kinvolk/ebpf/iterators/iterators_bpfel.go
Compiled /home/mvb/kinvolk/ebpf/iterators/iterators_bpfeb.o
Wrote /home/mvb/kinvolk/ebpf/iterators/iterators_bpfeb.go
2021/07/16 07:35:54 loading objects: field TracepointSyscallsSysEnterExecve: program tracepoint__syscalls__sys_enter_execve: CO-RE relocations: can't read types: type id 9468: unknown kind: Unknown

```

I originally found it while trying to run an example from libbpf-boostrap, it was solved by updating libbpf and then @anakryiko [suggested](https://github.com/libbpf/libbpf-bootstrap/pull/36#issuecomment-879302951) `BTF_KIND_FLOAT` could be the reason.

ti-mo:
Thanks for the input!

lmb:
I fixed the vmlinux BTF thanks to timo's hint with pahole, and merged your change manually. Thanks!

PR #346, opened 2021-07-15
testdata: update vmlinux BTF to 5.13.2

lmb:


ti-mo:
With bpftool 5.13.0, I don't see a float type in this blob:
```
~ gunzip vmlinux-btf.gz && bpftool btf dump file vmlinux-btf | grep -i float
[18062] FUNC 'cgroup_parse_float' type_id=18061 linkage=static
~
```

lmb:
Good catch :hankey:

> Do we need to upgrade a compiler somewhere to emit this new type?

Nope. I think the problem is that x86 / arm64 kernel uapi doesn't expose float, but s390 does. Not sure what to do about this. Maybe just accept the other PR as it is?

mauriciovasquezbernal:
> I think the problem is that x86 / arm64 kernel uapi doesn't expose float,

:thinking: I'm running x86 and I think my kernel exposes some floats:

```
$ uname -a
Linux mvb-desktop 5.13.0-051300-generic #202106272333 SMP Sun Jun 27 23:36:43 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
$ ./bpftool --version
./bpftool v5.13.0
features: skeletons
$ ./bpftool btf dump file /sys/kernel/btf/vmlinux | grep -i float
[9468] FLOAT 'double' size=8
[24538] FUNC 'cgroup_parse_float' type_id=24537 linkage=static
[31908] FUNC 'btf_float_log' type_id=31907 linkage=static
[31910] FUNC 'btf_float_check_member' type_id=31909 linkage=static
[31912] FUNC 'btf_float_check_meta' type_id=31911 linkage=static
[103805] FLOAT 'float' size=4
```

ti-mo:
> I think the problem is that x86 / arm64 kernel uapi doesn't expose float

Yep, this seems rather unlikely. @lmb try upgrading your pahole? iirc kernel BTF is gcc DWARF -> pahole -> BTF.

Mauricio's kernel is mainline `kernel-ppa`, so this will start popping up in the wild.

lmb:
I was almost right ;P

```
$ ~/dev/bpf-next/tools/bpf/bpftool/bpftool btf dump file vmlinux-btf  | grep -i float
[9607] FLOAT 'double' size=8
[18062] FUNC 'cgroup_parse_float' type_id=18061 linkage=static
```

We should get ci-kernels into the docker builder thing as well...

PR #351, opened 2021-07-19
link/uprobe: handle ENOTSUPP from the kernel's prepare_uprobe()

ti-mo:
When running the uprobe tests, I noticed this was failing with my copy of Bash. Offset 0x12345 seems to contain a trap instruction and is rejected by prepare_uprobe(): https://elixir.bootlin.com/linux/v5.13.3/source/kernel/events/uprobes.c#L854

This patch changes the dummy offset to 0x1 and adds some error context to the slightly esoteric ENOTSUPP.

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -230,6 +230,11 @@ func pmuProbe(typ probeType, symbol, path string, offset uint64, ret bool) (*per
 	if errors.Is(err, os.ErrNotExist) || errors.Is(err, unix.EINVAL) {
 		return nil, fmt.Errorf("symbol '%s' not found: %w", symbol, os.ErrNotExist)
 	}
+	// Since at least commit cb9a19fe4aa51, ENOTSUPP is returned
+	// when attempting to set a uprobe on a trap instruction.
+	if errors.Is(err, unix.ENOTSUPP) {
+		return nil, fmt.Errorf("failed setting uprobe on offset %#x (possible trap insn): %w", offset, err)
+	}
 	if err != nil {
 		return nil, fmt.Errorf("opening perf event: %w", err)
 	}
--- a/link/uprobe_test.go
+++ b/link/uprobe_test.go
@@ -81,7 +81,7 @@ func TestUprobeExtWithOpts(t *testing.T) {
 
 	// This Uprobe is broken and will not work because the offset is not
 	// correct. This is expected since the offset is provided by the user.
-	up, err := bashEx.Uprobe("open", prog, &UprobeOptions{Offset: 0x12345})
+	up, err := bashEx.Uprobe("open", prog, &UprobeOptions{Offset: 0x1})
 	if err != nil {
 		t.Fatal(err)
 	}


link/kprobe.go
```diff
@@ -230,6 +230,11 @@ func pmuProbe(typ probeType, symbol, path string, offset uint64, ret bool) (*per
 	if errors.Is(err, os.ErrNotExist) || errors.Is(err, unix.EINVAL) {
 		return nil, fmt.Errorf("symbol '%s' not found: %w", symbol, os.ErrNotExist)
 	}
+	// Since commit cb9a19fe4aa51, ENOTSUPP is returned when attempting to set
+	// a uprobe on a trap instruction.
+	if errors.Is(err, unix.ENOTSUPP) {
+		return nil, fmt.Errorf("setting a uprobe on a trap insn (offset %#x) is not supported: %w", offset, err)
```
lmb, line 236: Since this is probably no the only reason for ENOTSUPP, make the language less strong?
```
failed to set uprobe on offset (possible trap insn): %w
```

link/kprobe.go
```diff
@@ -230,6 +230,11 @@ func pmuProbe(typ probeType, symbol, path string, offset uint64, ret bool) (*per
 	if errors.Is(err, os.ErrNotExist) || errors.Is(err, unix.EINVAL) {
 		return nil, fmt.Errorf("symbol '%s' not found: %w", symbol, os.ErrNotExist)
 	}
+	// Since commit cb9a19fe4aa51, ENOTSUPP is returned when attempting to set
```
mmat11, line 233: I think the specific commit is just a small refactor, the `is_swbp_insn(...)` check was already there so I'd remove the first part of the comment

link/kprobe.go
ti-mo: It's the only occurrence in `kernel/events/`, but I'll change it. This errno has a bit of an interesting history; it's [not part of UAPI and shouldn't be propagated to userspace](https://linux.kernel.narkive.com/dkX5RyfC/question-usage-of-enotsupp-error-code), yet over the years it's made its way into a bunch of drivers, and more recently `bpf/` started using it a bunch as well.

link/kprobe.go
ti-mo: I thought a sha would be more convenient than a path or a function name, since those tend to change over time. I'll specify 'at least' there. :)

link/uprobe_test.go
```diff
@@ -81,7 +81,7 @@ func TestUprobeExtWithOpts(t *testing.T) {
 
 	// This Uprobe is broken and will not work because the offset is not
 	// correct. This is expected since the offset is provided by the user.
-	up, err := bashEx.Uprobe("open", prog, &UprobeOptions{Offset: 0x12345})
```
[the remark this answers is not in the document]
ti-mo: Clearly high enough for this to happen in the wild. :stuck_out_tongue:

PR #354, opened 2021-07-23
elf_reader: validate .maps ELF data one BTF map definition at a time

ti-mo:
Currently, the whole .maps section is checked for non-zero bytes, rejecting the whole ELF if it contains even a single non-zero byte.

This patch results in nicer error output and doesn't reject the whole program, which currently prevents us from mixing valid and invalid BTF map definitions in a single testdata ELF.

An extra check is added at the end of the function to ensure all bytes in .maps are accounted for with BTF metadata.

Also documented loadBTFMaps() a bit to clarify the rationale.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


elf_reader.go
```diff
@@ -582,6 +598,13 @@ func (ec *elfCode) loadBTFMaps(maps map[string]*MapSpec) error {
 
 			maps[name] = mapSpec
 		}
+
+		// Drain the ELF section reader to make sure all bytes are accounted for
+		// with BTF metadata.
+		var b bytes.Buffer
+		if i, err := b.ReadFrom(rs); i > 0 || err != nil {
```
lmb, line 605: Nit: why not `io.Copy(ioutil.Discard, rs)`?

lmb (approved):

elf_reader.go
ti-mo: The subsequent error contains the unexpected bytes. On second thought, this might not be a good idea since it could be a large amount of data that would be printed to stdout in most cases.

ti-mo: Done, no longer dumping the bytes but just reporting `i` instead. (btw, all of `ioutil` is now deprecated, using `io.Discard` instead)

ti-mo: Looks like we still need ioutil for the previous stable Go. :shrug:

PR #355, opened 2021-07-25
features: add HaveProgType API

rgo3:
With this commit we add `HaveProgType(pt ebpf.ProgramType) error`,
which allows to probe for available BPF program types. As with the
HaveMapType API, results are cached and probes run at most once.

### DIFF
--- a/features/prog.go
+++ b/features/prog.go
@@ -1,0 +1,156 @@
+package features
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+func init() {
+	pc.progTypes = make(map[ebpf.ProgramType]error)
+}
+
+var (
+	pc progCache
+)
+
+type progCache struct {
+	sync.Mutex
+	progTypes map[ebpf.ProgramType]error
+}
+
+func createProgLoadAttr(pt ebpf.ProgramType) (*internal.BPFProgLoadAttr, error) {
+	var expectedAttachType ebpf.AttachType
+
+	insns := asm.Instructions{
+		asm.LoadImm(asm.R0, 0, asm.DWord),
+		asm.Return(),
+	}
+
+	buf := bytes.NewBuffer(make([]byte, 0, len(insns)*asm.InstructionSize))
+	if err := insns.Marshal(buf, internal.NativeEndian); err != nil {
+		return nil, err
+	}
+
+	bytecode := buf.Bytes()
+	instructions := internal.NewSlicePointer(bytecode)
+
+	// Some programs have expected attach types which are checked during the
+	// BPD_PROG_LOAD syscall.
+	switch pt {
+	case ebpf.CGroupSockAddr:
+		expectedAttachType = ebpf.AttachCGroupInet4Connect
+	case ebpf.CGroupSockopt:
+		expectedAttachType = ebpf.AttachCGroupGetsockopt
+	case ebpf.SkLookup:
+		expectedAttachType = ebpf.AttachSkLookup
+	default:
+		expectedAttachType = ebpf.AttachNone
+	}
+
+	// Kernels before 5.0 (6c4fc209fcf9 "bpf: remove useless version check for prog load")
+	// require the version field to be set to the value of the KERNEL_VERSION
+	// macro for kprobe-type programs.
+	v, err := internal.KernelVersion()
+	if err != nil {
+		return nil, fmt.Errorf("detecting kernel version: %w", err)
+	}
+	kv := v.Kernel()
+
+	return &internal.BPFProgLoadAttr{
+		ProgType:           uint32(pt),
+		Instructions:       instructions,
+		InsCount:           uint32(len(bytecode) / asm.InstructionSize),
+		ExpectedAttachType: uint32(expectedAttachType),
+		License:            internal.NewStringPointer("GPL"),
+		KernelVersion:      kv,
+	}, nil
+}
+
+// HaveProgType probes the running kernel for the availability of the specified program type.
+// Return values have the following semantics:
+//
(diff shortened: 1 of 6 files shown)


rgo3:
Keeping it as a draft for now, as I still have to investigate some of the failing probes.

features/prog.go
```diff
@@ -0,0 +1,157 @@
...
+
+type progCache struct {
+	sync.Mutex
+	progTypes map[ebpf.ProgramType]error
+}
+
+func createProgLoadAttr(pt ebpf.ProgramType) (*internal.BPFProgLoadAttr, error) {
+	var expectedAttachType ebpf.AttachType
+
+	insns := asm.Instructions{
```
ti-mo, line 32: The assembly of the probe program could be moved to `init()` (rather, a function called by `init()`), since it only needs to be assembled once. `instructions` and the parameter to `InsCount` could then be stored in `pc`, which saves some garbage.

features/prog_test.go
```diff
@@ -0,0 +1,85 @@
...
+	ebpf.RawTracepointWritable: "5.2",
+	ebpf.CGroupSockopt:         "5.3",
+	ebpf.Tracing:               "5.5",
+	ebpf.StructOps:             "5.6",
+	ebpf.Extension:             "5.6",
+	ebpf.LSM:                   "5.7",
+	ebpf.SkLookup:              "5.9",
+}
+
+func TestHaveProgType(t *testing.T) {
```
ti-mo, line 46: This could be run with `t.Parallel()`.

features/prog.go
rgo3: Since I probably also want to use that function to generate the load attribute when probing helpers, I imagined generating the instructions "dynamically" here according to the helper call thats supposed to be probed. So moving this to `init()` would just be reverted in that case.

features/prog_test.go
rgo3: Done. Had to make use of some IMO not so elegant variable copying as mentioned [here](https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721#how-to-solve-this), but that works.

features/prog.go
ti-mo: I was thinking we could use the pre-cooked program when probing program types and generating one on-the-fly when probing helper types. This might be a premature optimization, though. Let's wait and see what this will look like with helper probes added. :+1:

features/prog_test.go
ti-mo: This is common in Go. :slightly_smiling_face: Good catch!

features/prog_test.go
```diff
@@ -0,0 +1,88 @@
...
+			// we should make sure the test runs anyway and fails on old kernels
+			minVersion = "0.0"
+		}
+
+		feature := fmt.Sprintf("program type %s", pt.String())
+
+		t.Run(pt.String(), func(t *testing.T) {
+			t.Parallel()
+			if pt == ebpf.LircMode2 {
+				t.Skip("LircMode2 isn't supported by testing kernels")
```
ti-mo, line 63: I think we could use `testutils.SkipIfNotSupported()` here, so at least the probe runs, but we discard ErrNotSupported until/if/when we rebuild CI kernels with this kconfig enabled.

ti-mo (approved):
Looks good! Left one more nit for handling lircmode2.

features/prog_test.go
rgo3: `SkipIfNotSupported()` might be handy to create an easy test for the HaveProgHelper API. We could basically run a double for over progs and helpers and discard NotSupported results. While its not to accurate it would at least give good coverage and we could avoid creating a complicated "test case matrix".

features/prog.go
```diff
@@ -0,0 +1,157 @@
...
+		asm.LoadImm(asm.R0, 0, asm.DWord),
+		asm.Return(),
+	}
+
+	buf := bytes.NewBuffer(make([]byte, 0, len(insns)*asm.InstructionSize))
+	err := insns.Marshal(buf, internal.NativeEndian)
+	if err != nil {
```
qmonnet, line 39: Could be
```suggestion
	if err := insns.Marshal(buf, internal.NativeEndian); err != nil {
```
Here and below in the same function.

internal/syscall.go
```diff
@@ -68,6 +69,48 @@ func BPF(cmd BPFCmd, attr unsafe.Pointer, size uintptr) (uintptr, error) {
...
 
+type BPFProgLoadAttr struct {
+	ProgType           uint32
+	InsCount           uint32
+	Instructions       Pointer
+	License            Pointer
+	LogLevel           uint32
+	LogSize            uint32
+	LogBuf             Pointer
+	KernelVersion      uint32     // since 4.1  2541517c32be
```
qmonnet, line 80: Not sure if worth keeping the comments, since the latest fields don't have it and I doubt there's much motivation to keep the list up-to-date (but that's not my call to make).

qmonnet (approved):
Minor, non-blocking comments below. The PR looks good!

features/prog_test.go
ti-mo: Yup, we should only test a few known-good and known-bad prog/helper combinations, and combinatorially test all prog/helper types to catch errors that are not `nil`/`ErrNotSupported`. Though that won't alert us for cases like StructOps where probes always fail when no BTF is specified. That's probably something we need to accept.

features/prog.go
rgo3: I don't think I can apply this below with `v, err := internal.KernelVersion()` because then variable `v` is only available in the if scope I believe, so I would have to use an `else` which I'd like to avoid. I'll use your suggestion for the first occurrence though :)

internal/syscall.go
rgo3: If we choose to remove the comments I'd say we can do it in another PR, likewise if we'd chose to add some comments to keep the fields up-to-date.

features/prog.go
qmonnet: Correct I read too fast for the second one :+1:

internal/syscall.go
qmonnet: No objection

PR #356, opened 2021-07-29
Assorted patches in preparation of BTF map auto-population

ti-mo:
Best reviewed per commit. These are a few things I encountered while working on prog array auto-population from ELF/BTF.

Making separate PRs here to (hopefully) make reviewing a bit easier.


collection.go
```diff
@@ -49,6 +50,17 @@ func (cs *CollectionSpec) Copy() *CollectionSpec {
 	return &cpy
 }
 
+// ByteOrder returns the byte order of the Programs in the CollectionSpec.
+//
+// If the CollectionSpec doesn't contain any Programs, returns the host's
+// native endianness.
+func (cs *CollectionSpec) ByteOrder() binary.ByteOrder {
```
lmb, line 57: I think I'd prefer adding a field `ByteOrder binary.ByteOrder` to CollectionSpec instead. What do you think?

map.go
```diff
@@ -17,6 +17,7 @@ import (
 var (
 	ErrKeyNotExist      = errors.New("key does not exist")
 	ErrKeyExist         = errors.New("key already exists")
+	ErrMapSize          = errors.New("map too small for given key")
```
lmb, line 20: Express it the other way round: `key too big for map`? Why does this need to be exported?

syscalls.go
```diff
@@ -372,6 +372,10 @@ func wrapMapError(err error) error {
 		return internal.SyscallError(ErrNotSupported, unix.ENOTSUPP)
 	}
 
+	if errors.Is(err, unix.E2BIG) {
+		return internal.SyscallError(ErrMapSize, unix.E2BIG)
```
lmb, line 376: I think `fmt.Errorf("foo: %w")` would suffice here, unless you need ErrMapSize later.

collection.go
ti-mo: > adding a field `ByteOrder binary.ByteOrder` to CollectionSpec instead

I had it like this initially but it's a breaking API change for callers declaring `CollectionSpec` struct literals with implicit fields. (there were a few such spots in the codebase) It also needs to be included in `CollectionSpec.Copy()`, and it ultimately felt a bit fragile.

> the `internal.NativeEndian` fallback is weird

Ooops, this returns an internal value from an exported method. Unexporting is probably the way to go then. :+1:

map.go
ti-mo: True, makes sense. I provided the rationale in the commit message: out-of-bounds writes return a message like `update failed: argument list too long`, which I find a bit misleading. It may be easier for us to figure out what this means by being familiar with the API, but I can imagine this being rather confusing when it happens to a user for the first time.

I'll go with your suggestion below of just adding some string context. We don't really need to `ErrMapSize` here, just thought I'd stick with the existing convention.

collection.go
ti-mo: Looks like `internal.NativeEndian` is a package-global, not a const. I went with unexported `byteOrder binary.ByteOrder` instead. Knowing the byte order is required for parsing ELF map definitions as well, and this needs to work correctly for ELFs not containing any Programs as well.

lmb (approved):

PR #357, opened 2021-07-29
Auto-populate PROG_ARRAY maps from ELF+BTF

### LINKED ISSUE #214: Populate program arrays from BTF .values

ti-mo:
(This PR depends on https://github.com/cilium/ebpf/pull/356)

Closes #214.

Best reviewed per commit. These patches add support for statically declaring map contents of PROG_ARRAY maps from C code. clang9 and up is required, as is compiling with BTF enabled (`clang -g ...`).

Follow-up work:
- map-in-map auto-population
- program in `examples/`

cc @anakryiko


testdata/btf_map_init.c
```diff
@@ -0,0 +1,33 @@
+#include "common.h"
+
+#if __clang_major__ >= 9
+
+int __section("socket/1") tail_1() {
```
anakryiko, line 5: is the "socket/1" required to be able to use it in PROG_ARRAY? libbpf doesn't support "/N" approach and I haven't planned to do so specifically because of the array initialization syntax you are implementing in this PR. So just trying to make sure this is not a requirement.

testdata/btf_map_init.c
```diff
@@ -0,0 +1,33 @@
...
+	return 42;
+}
+
+// Tail call map (program array) initialized with program pointers.
+struct {
+	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+	__uint(key_size, sizeof(uint32_t));
+	__uint(max_entries, 2);
+	__array(values, int ());
```
anakryiko, line 14: this is nifty, I didn't realize `int()` can be used with typeof() to get a function prototype type recorded. Nice!

testdata/btf_map_init.c
```diff
@@ -0,0 +1,33 @@
...
+} prog_array_init __section(".maps") = {
+	.values = {
+    // Skip index 0 to exercise empty array slots.
+		[1] = &tail_1,
```
anakryiko, line 18: tailcalls do accept void *ctx, right? Might be a good idea to have a realistic example with the context

anakryiko (commented):
I only looked at BPF example. This looks great! We definitely need to complete libbpf's map initialization support with this feature.

testdata/btf_map_init.c
ti-mo: Hey, thanks for the feedback! The numeric identifier in there might be a bit misleading, there is no `/N` magic going on there, the lib works exactly like libbpf. This prefix just specifies the prog type. Initially, I had 2 tail calls in this test, and I needed unique section names for them, since we don't yet implement multiple progs per section.

Your comment prompted me to investigate why some libbpf `section_defs` end in `/`, and turns out the part after the slash specifies the object to attach the program to. :slightly_smiling_face:  I'll just rename this to `"socket"`.

testdata/btf_map_init.c
ti-mo: Yep! The stuff in `testdata/` are contrived examples meant to exercise specific loader code paths, most of the .c files have a disclaimer. :stuck_out_tongue: Forgot it here. We also have an `examples/` folder with more realistic examples, will add something there later.

PR #361, opened 2021-08-02
Export `CollectionSpec.ByteOrder`

ti-mo:
This is necessary to be able to access it from package `link`.

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -27,7 +27,9 @@ type CollectionSpec struct {
 	Maps     map[string]*MapSpec
 	Programs map[string]*ProgramSpec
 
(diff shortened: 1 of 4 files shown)


collection.go
```diff
@@ -30,6 +30,12 @@ type CollectionSpec struct {
 	byteOrder binary.ByteOrder
 }
 
+// ByteOrder returns the byte order of the ELF object
+// the CollectionSpec was created from.
+func (cs *CollectionSpec) ByteOrder() binary.ByteOrder {
+	return cs.byteOrder
```
lmb, line 36: Just export the `byteOrder` field?

PR #362, opened 2021-08-02
Support ELF map definitions larger than `bpf_map_def`

ti-mo:
Supersedes https://github.com/cilium/ebpf/pull/360, as its base branch was merged, closing the PR.

This would allow us to gradually migrate the CIlium datapath from iproute2 (tc) to this library.

This patch provides access to the raw byte contents, which the caller can then parse at will. It was decided not to include iproute2 map-in-map support out of the box since it was seldom used, and supporting it would add complexity for a feature that is essentially deprecated.


lmb (changes requested):
Timo and I spoke, and decided that the following is a better approach:
* Add a `[]byte` to MapSpec which contains the trailing bytes from the ELF
* Enforce that this `[]byte` is nil / of zero length when creating a map from the MapSpec

map.go
```diff
@@ -65,6 +66,11 @@ type MapSpec struct {
 	// InnerMap is used as a template for ArrayOfMaps and HashOfMaps
 	InnerMap *MapSpec
 
+	// Extra trailing bytes found in the ELF map definition when using structs
+	// larger than libbpf's bpf_map_def. Must be empty before instantiating
+	// the MapSpec into a Map.
+	Extra bytes.Buffer
```
lmb, line 72: Why not `[]byte`?

lmb: I'm trying to figure out what `MapSpec.Copy` should do with `Extra`. For consistency it would be nice to copy the field but doing that with a `Buffer` while keeping the original read position doesn't seem straightforward. `[]byte` doesn't have that problem and has a smaller API surface and you can get a reader by using `bytes.NewReader(spec.Extra)`.

ti-mo: Good point, I completely missed this initially. For our use case, re-initializing the Buffer on the destination and then `Write()`ing the remaining bytes should be sufficient. It won't work with `UnreadByte()` and friends, but oh well,

```go
	cpy.Extra = bytes.Buffer{}
	_, _ = cpy.Extra.Write(ms.Extra.Bytes())
```

Note that `cpy.Extra = bytes.NewBuffer(ms.Extra.Bytes())` retains `ms.Extra`'s backing array and is not safe to use.

ti-mo: We've decided to go with a `bytes.Reader` instead.

lmb (approved):

PR #363, opened 2021-08-02
Auto-populate tail-call and map-in-map maps from ELF/BTF

### LINKED ISSUE #112: Populate inner maps from BTF .values

### LINKED ISSUE #214: Populate program arrays from BTF .values

ti-mo:
Supersedes https://github.com/cilium/ebpf/pull/357 as its base branch was merged.

Closes #214
Closes #112

Best reviewed per commit. These patches add support for statically declaring map contents of PROG_ARRAY, HASH_OF_MAPS and ARRAY_OF_MAPS from C code. clang9 and up is required, as is compiling with BTF enabled (`clang -g ...`).

Follow-up work:
- program in `examples/`, after a merge


elf_reader.go
```diff
@@ -821,6 +843,62 @@ func resolveBTFArrayMacro(typ btf.Type) (btf.Type, error) {
...
+	for k, off := uint32(0), start; k < elems; k, off = k+1, off+align {
+		r, ok := es.relocations[uint64(off)]
+		if ok {
+			// Relocation exists for the current offset in the ELF section.
+			// Emit a value stub based on the type of relocation to be replaced by
+			// a real fd later in the pipeline before populating the map.
+			// Map keys are encoded in MapKV entries, so empty array indices are
+			// skipped here.
+			switch t := elf.ST_TYPE(r.Info); t {
+			case elf.STT_FUNC:
```
lmb, line 889: What if we check elfCode.sections[r.Section???].kind instead? The STT_FUNC / STT_OBJECT semantics are kind of brittle, the less we rely on it the better I think.

collection.go
```diff
@@ -325,6 +338,7 @@ func (hc handleCache) close() {
 func lazyLoadCollection(coll *CollectionSpec, opts *CollectionOptions) (
```
lmb, line 338: This function is such a turd :fearful:

collection.go
```diff
@@ -424,6 +438,44 @@ func lazyLoadCollection(coll *CollectionSpec, opts *CollectionOptions) (
...
+
+			mapSpec = mapSpec.Copy()
+
+			// Replace any object stubs with loaded objects.
+			for i, kv := range mapSpec.Contents {
+				switch v := kv.Value.(type) {
+				case programStub:
+					prog, ok := progs[string(v)]
+					if !ok {
+						return fmt.Errorf("program %s not assigned, but map %s requires it: %w", v, mapName, errUnsatisfiedReference)
```
lmb, line 457: I think this is a bit problematic, because it breaks lazy loading a bit: it's now not enough to request the maps and programs your code needs, you also need to add the transitive dependencies of maps. That isn't straight forward, except in the "load everything always" case.

Maybe its enough to call `loadMap` here? Same for `loadProgram`.

map.go
```diff
@@ -892,12 +894,21 @@ func (m *Map) Freeze() error {
 	return nil
 }
 
-func (m *Map) populate(contents []MapKV) error {
-	for _, kv := range contents {
+// populate populates the Map according to the Contents specified
+// in spec and freezes the Map if requested by spec.
+func (m *Map) populate(spec *MapSpec) error {
```
lmb, line 904: Maybe `finalize`? `populate` makes me forget that the map might be frozen after this.

testdata/btf_map_init.c
```diff
@@ -0,0 +1,57 @@
...
+int __section("socket/tail") tail_1() {
+	return 42;
+}
+
+// Tail call map (program array) initialized with program pointers.
+struct {
+	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+	__uint(key_size, sizeof(uint32_t));
+	__uint(max_entries, 2);
+	__array(values, int ());
```
lmb, line 16: Echoing Andrii's comment from the earlier PR: does this work for `int (void*)` as well?

lmb (changes requested):
Looks really good, I only have minor things. Thanks!

testdata/btf_map_init.c
ti-mo: Yes, ofc it works when the function signatures match. `T ()` seems to be agnostic as to what the function arguments are. The function signature isn't reflected in the resulting ELF, so from the loader side there's no difference.

elf_reader.go
```diff
@@ -821,6 +843,62 @@ func resolveBTFArrayMacro(typ btf.Type) (btf.Type, error) {
...
+func resolveBTFValuesContents(es *elfSection, vs *btf.VarSecinfo, member btf.Member) ([]MapKV, error) {
+	// The elements of a .values pointer array are not encoded in BTF.
+	// Instead, relocations are generated into each array index.
+	// However, it's possible to leave certain array indices empty, so all
+	// indices' offsets need to be checked for emitted relocations.
+
+	// The offset of the 'values' member within the _struct_ (in bits)
+	// is the starting point of the array. Convert to bytes. Add VarSecinfo
+	// offset to get the absolute position in the ELF blob.
+	start := (member.Offset / 8) + vs.Offset
```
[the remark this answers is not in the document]
ti-mo: Only `btf.Member.Offset` is in bits. Both `btf.VarSecinfo`'s `Offset` and `Size` are in bytes, hence the Member offset is converted to bytes.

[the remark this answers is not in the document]
ti-mo: Perhaps you mean renaming `btf.Member.Offset` to `OffsetBits`? That might help!

elf_reader.go
```diff
@@ -821,6 +843,62 @@ func resolveBTFArrayMacro(typ btf.Type) (btf.Type, error) {
 	return ptr.Target, nil
 }
 
+// resolveBTFValuesContents resolves relocations into ELF sections belonging
+// to btf.VarSecinfo's. This can be used on the 'values' member in BTF map
+// definitions to extract static declarations of map contents.
+func resolveBTFValuesContents(es *elfSection, vs *btf.VarSecinfo, member btf.Member) ([]MapKV, error) {
```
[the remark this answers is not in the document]
ti-mo: I think `mapSpecFromBTF` is due for a refactor, will try to include that. They should probably both be free functions.

collection.go
ti-mo: Okay, will give it a try. I'm concerned about refcounting traps since we won't be holding on to any references to inner maps/progs. Programs don't hold direct references to inner objects either, so they will be deallocated when the outer map key holding them is overwritten. It feels safer failing explicitly. If we go for the hands-off approach of automatically lazy-loading dependencies, this would need to be well-documented, but users tend to ignore documentation whenever they possibly can.

elf_reader.go
ti-mo: I did some more digging and found that relocations also have their 'source' section index attached, the section where the symbol itself lives. In our test programs, the section of relocations of type `STT_FUNC` points to wherever the function lives. Relocations of type `STT_OBJECT` point to the `.maps` section index.

```
λ  ebpf prog-array-autopop ✗ readelf -Sgt -x .maps -r testdata/btf_map_init-el.elf
There are 26 section headers, starting at offset 0x10f8:

Section Headers:
  [Nr] Name
       Type              Address          Offset            Link
       Size              EntSize          Info              Align
       Flags
...
  [ 3] socket/tail
       PROGBITS         0000000000000000  0000000000000040  0
       0000000000000010 0000000000000000  0                 8
       [0000000000000006]: ALLOC, EXEC
...
  [ 6] .maps
       PROGBITS         0000000000000000  0000000000000080  0
       0000000000000078 0000000000000000  0                 8
       [0000000000000003]: WRITE, ALLOC
```

Unfortunately, both the program and .maps sections are of type `PROGBITS`. `.maps` has different flags than `socket/tail` in this example, but going off of that doesn't feel any less brittle than using the relocation symbol type to begin with. Such seems to be the nature of ELF parsing.

Is this what you had in mind? Any other suggestions?

elf_reader.go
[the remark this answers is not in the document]
ti-mo: I've included a rename in this PR.

elf_reader.go
lmb: I mean using this info: https://github.com/cilium/ebpf/blob/master/elf_reader.go#L219-L231 The loader already classifies sections based on some heuristics. Instead of asking "is the symbol of the correct type" we'd check that the symbol's section has the correct kind. Writing this out, it's probably best to stick with what you have right now. We'll return an error if the map or program don't exist when creating the map.

collection.go
lmb: > I'm concerned about refcounting traps since we won't be holding on to any references to inner maps/progs.

Hmm you mean the behaviour that program arrays are removed when the user space handle goes away? Yeah, that is really annoying. I remember that there is a flag which restores a sane behaviour, could we force enable this for ProgramArrays?

elf_reader.go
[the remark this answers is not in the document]
lmb: Yeah, we can do that in a separate PR.

collection.go
ti-mo: Program arrays can be pinned, I don't think that's an issue. Let's say userspace creates a program and a program array, writes a reference to the program to prog array index 0, and then closes the fd because it was not requested by `LoadAndAssign`. The last reference to that program is the entry in index 0, and if that map key gets overwritten, the program gets deallocated. I also vaguely remember prog_arrays being write-only, so the program reference can't be retrieved from the prog_array.

> I remember that there is a flag which restores a sane behaviour

Sounds like something we should take into account, do you have more info?

lmb: >Program arrays can be pinned, I don't think that's an issue. Let's say userspace creates a program and a program array, writes a reference to the program to prog array index 0, and then closes the fd because it was not requested by LoadAndAssign.

I only half remembered the problem, sorry. In your scenario there is no problem when closing the program fd. The issue is that closing the fd for the program array (so, removing the last user space reference) will clear all slots of the program array even though it might still be referenced by the BPF. The result is missed tail calls. This is a really big footgun that has bitten us in production before. I thought that `BPF_F_PRESERVE_ELEMS` would fix this, but that flag only works for perf event arrays, not program arrays.

For map auto population this means that a program array that is populated by the library has to be explicitly requested / held on to by the user. Otherwise tail calls will fail mysteriously.

In the best case we could detect that a program array is only referenced by the BPF and return an error? Unfortunately that requires fudging lazyLoadCollection even more to distinguish between "map required by BPF" and "map required by user".

Otherwise we could ignore the problem.

> The last reference to that program is the entry in index 0, and if that map key gets overwritten, the program gets deallocated.

That's sane behaviour, don't you think?

ti-mo: >> The last reference to that program is the entry in index 0, and if that map key gets overwritten, the program gets deallocated.

> That's sane behaviour, don't you think?

Agreed, that would require an explicit action on behalf of the user, and is not something we can account for.

Re: losing prog_array references: this is something we should prevent IMO, but can be addressed outside of this PR since it would require a refactor of the lazy-loader. We should probably merge this one as-is and land a refactor before releasing 0.7.0.

Created https://github.com/cilium/ebpf/issues/366 for this.

collection.go
```diff
@@ -424,6 +438,46 @@ func lazyLoadCollection(coll *CollectionSpec, opts *CollectionOptions) (
 		return prog, nil
 	}
 
+	populateMaps = func() error {
+		for mapName, m := range maps {
+			// TODO: Store reference to MapSpec in Map to avoid this?
```
lmb, line 443: Nit: make your mind up about the TODO.

lmb (approved):

PR #364, opened 2021-08-04
Add kprobepin example showing how to re-use a pinned map in a BPF prog

ungureanuvladvictor:
Chatted with @ti-mo on Slack about adding an example on how to re-use a pinned map given this is something supported by the library. I figured out the current `kprobe` example is pretty good given it keeps a counter of how many time `sys_execve` got called.


examples/kprobepin/bpf/kprobe_pin_example.c
```diff
@@ -0,0 +1,26 @@
+#include "common.h"
+#include "bpf_helpers.h"
+
+char __license[] SEC("license") = "Dual MIT/GPL";
+
+struct bpf_map_def SEC("maps") kprobe_map = {
```
lmb, line 6: You can simplify this by using a BTF style map definition with `__uint(pinning, LIBBPF_PIN_BY_NAME)`. Then specify [MapOptions.PinPath](https://pkg.go.dev/github.com/cilium/ebpf#MapOptions) and the library will handle things for you.

examples/kprobepin/main.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Fatalf("failed to set temporary rlimit: %v", err)
+	}
+
+	var kProbeObj KProbePinExampleObjects
+	if err := LoadKProbePinExampleObjects(&kProbeObj, &ebpf.CollectionOptions{
+		Maps: ebpf.MapOptions{
+			PinPath: bpfFSPath,
+		},
+		Programs: ebpf.ProgramOptions{
+			LogLevel: 1,
```
lmb, line 52: Nit: this isn't strictly necessary and the default (0) is going to be better in most cases.

examples/kprobepin/main.go
```diff
@@ -0,0 +1,86 @@
...
+		Cur: unix.RLIM_INFINITY,
+		Max: unix.RLIM_INFINITY,
+	}); err != nil {
+		log.Fatalf("failed to set temporary rlimit: %v", err)
+	}
+
+	var kProbeObj KProbePinExampleObjects
+	if err := LoadKProbePinExampleObjects(&kProbeObj, &ebpf.CollectionOptions{
+		Maps: ebpf.MapOptions{
+			PinPath: bpfFSPath,
```
lmb, line 49: Can you add a comment that explains what this does? Plus, please use a subdirectory of `/sys/fs/bpf`: `filepath.Join(bpfFSPath, "execve_probe")`.

examples/kprobepin/main.go
```diff
@@ -0,0 +1,86 @@
...
+	// pre-compiled program. Each time the kernel function enters, the program
+	// will increment the execution counter by 1. The read loop below polls this
+	// map value once per second.
+	kp, err := link.Kprobe(fn, kProbeObj.KprobeExecve)
+	if err != nil {
+		log.Fatalf("opening kprobe: %s", err)
+	}
+	defer kp.Close()
+
+	// // Read loop reporting the total amount of times the kernel
```
lmb, line 68: Nit: double comment

examples/kprobepin/main.go
```diff
@@ -0,0 +1,86 @@
...
+	// Increase the rlimit of the current process to provide sufficient space
+	// for locking memory for the eBPF map.
+	if err := unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{
+		Cur: unix.RLIM_INFINITY,
+		Max: unix.RLIM_INFINITY,
+	}); err != nil {
+		log.Fatalf("failed to set temporary rlimit: %v", err)
+	}
+
+	var kProbeObj KProbePinExampleObjects
```
lmb, line 46: This should have a `defer kProbeObj.Close()` I think.

examples/kprobepin/main.go
ungureanuvladvictor: Changed it.

examples/kprobepin/main.go
ungureanuvladvictor: Done

examples/kprobepin/main.go
ungureanuvladvictor: Ended up calling `os.MkdirAll` which is a no-op if the path exists.

lmb (approved):

lmb:
Thanks!

ti-mo:
Very clean, thanks!

zhangzhiqiangcs:
Hello, all, I run into this kprobe demo with the problem that 
```
failed to create bpf fs subpath: mkdir /sys/fs/bpf/sys_execve: no such file or directory
```
but I have  /sys/fs/bpf/ directory and filesystem is `sysfs`.

Any ideas or documents can help with this situation?

Thanks for any help.

ti-mo:
@zhangzhiqiangcs `/sys/fs/bpf` needs to be of type bpffs, not sysfs. Your distribution's init system (I assume you're also on wsl2?) usually takes care of setting this up.

Otherwise: `mount -t bpf bpf /sys/fs/bpf`.

zhangzhiqiangcs:
@ti-mo Sorry to forget paste my os-release. My machine is Ubuntu 18.04 . 

I am not sure why `/sys/fs/bpf` is exists and is `sysfs` type.

Execute command `mount -t bpf bpf /sys/fs/bpf` really solved my problem. Now the example runs as expect! Thank you very much !

PR #365, opened 2021-08-04
link/kprobe: don't redeclare a new variable sp

jellor:
The statement sp, err := unsafeStringPtr() declares a new variable sp
which shadows the original, which may result in the new variable sp is
collected before PerfEventOpen() is invoked.


ti-mo:
Damn, good catch, thank you!

cc @mmat11

PR #367, opened 2021-08-06
feature: add call fixups for bpf_probe_read_kernel, etc

xh4n3:
I met the `invalid func unknown#113` error yesterday when deploying bpf programs to a 4.19 kernel, and found that libbpf have this bpf_probe_read_kernel auto replacement feature when it loads program.
https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L6009

So I added this feature to this project and have it tested.

```
$ go test -run TestHaveProbeReadKernel
PASS
ok  	github.com/cilium/ebpf	0.003s
```

Please let me know if any problem with this implementation 🙇  Thanks a lot!

### DIFF
--- a/linker.go
+++ b/linker.go
@@ -136,5 +136,25 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
 		}
 	}
 
+	// fixupBPFCalls replaces bpf_probe_read_{kernel,user}[_str] with bpf_probe_read[_str] on older kernels
+	// https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L6009
+	iter = insns.Iterate()
+	for iter.Next() {
+		ins := iter.Ins
+		if !ins.IsBuiltinCall() {
+			continue
+		}
+		switch asm.BuiltinFunc(ins.Constant) {
+		case asm.FnProbeReadKernel, asm.FnProbeReadUser:
+			if err := haveProbeReadKernel(); err != nil {
+				ins.Constant = int64(asm.FnProbeRead)
+			}
+		case asm.FnProbeReadKernelStr, asm.FnProbeReadUserStr:
+			if err := haveProbeReadKernel(); err != nil {
+				ins.Constant = int64(asm.FnProbeReadStr)
+			}
+		}
+	}
+
 	return nil
 }
--- a/syscalls.go
+++ b/syscalls.go
@@ -1,11 +1,13 @@
 package ebpf
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
 	"unsafe"
 
+	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/unix"
 )
@@ -425,5 +427,33 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
 	if err != nil {
 		return internal.ErrNotSupported
 	}
+	return nil
+})
+
+var haveProbeReadKernel = internal.FeatureTest("bpf_probe_read_kernel", "5.5", func() error {
+	insns := asm.Instructions{
+		asm.Mov.Reg(asm.R1, asm.R10),
+		asm.Add.Imm(asm.R1, -8),
+		asm.Mov.Imm(asm.R2, 8),
+		asm.Mov.Imm(asm.R3, 0),
+		asm.FnProbeReadKernel.Call(),
+		asm.Return(),
+	}
+	buf := bytes.NewBuffer(make([]byte, 0, len(insns)*asm.InstructionSize))
+	if err := insns.Marshal(buf, internal.NativeEndian); err != nil {
+		return err
+	}
+	bytecode := buf.Bytes()
+
+	fd, err := internal.BPFProgLoad(&internal.BPFProgLoadAttr{
+		ProgType:     uint32(Kprobe),
+		License:      internal.NewStringPointer("GPL"),
+		Instructions: internal.NewSlicePointer(bytecode),
+		InsCount:     uint32(len(bytecode) / asm.InstructionSize),
+	})
+	if err != nil {
+		return internal.ErrNotSupported
+	}
+	_ = fd.Close()
 	return nil
 })
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -181,6 +181,11 @@ func (ins *Instruction) IsFunctionCall() bool {
 	return ins.OpCode.JumpOp() == Call && ins.Src == PseudoCall
 }
 
+// IsBuiltinCall returns true if the instruction is a built-in call, i.e. BPF helper call.
+func (ins *Instruction) IsBuiltinCall() bool {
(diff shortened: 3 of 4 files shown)


linker.go
```diff
@@ -136,5 +136,22 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
 		}
 	}
 
+	// fixupBPFCalls replaces bpf_probe_read_{kernel,user}[_str] with bpf_probe_read[_str] on older kernels
+	// https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L6009
+	iter = insns.Iterate()
```
ti-mo, line 141: Why is this implemented in an additional loop?

syscalls.go
```diff
@@ -469,3 +471,40 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
 	}
 	return nil
 })
+
+var haveProbeReadKernel = internal.FeatureTest("bpf_probe_read_kernel", "5.5", func() error {
+	spec := ProgramSpec{
```
ti-mo, line 476: It doesn't look like this `ProgramSpec` is actually passed to anything, could it be removed?

ti-mo (changes requested):
Nice find! Left a few comments.

linker.go
xh4n3: IMHO, the above two for-loops are for handling symbol call relocation, and both have if-condition that will filter bpf_probe_read instructions out. Thought it would be cleaner for another for-loop.

Or we can implement here https://github.com/cilium/ebpf/blob/76832385809beed631f0a1b8e4f7c51b6d24fd89/linker.go#L93 , what do you think?

syscalls.go
xh4n3: Sure!

syscalls.go
```diff
@@ -469,3 +471,36 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
...
+		license:  internal.NewStringPointer("GPL"),
+	}
+	attr.instructions = internal.NewSlicePointer(bytecode)
+	attr.insCount = uint32(len(bytecode) / asm.InstructionSize)
+
+	fd, err := bpfProgLoad(attr)
+	if err != nil {
+		return internal.ErrNotSupported
+	}
+	defer func() {
```
lmb, line 502: Nit: no need to defer here?

linker.go
```diff
@@ -136,5 +136,22 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
...
+	for iter.Next() {
+		ins := iter.Ins
+		switch asm.BuiltinFunc(ins.Constant) {
```
lmb, line 144: This is only valid to do if `ins.IsFunctionCall` is true.

linker.go
lmb: Your current approach and putting it in the first loop are both OK. I wouldn't put it in the second loop.

lmb (changes requested):

linker.go
xh4n3: You're right, I will add an if-filter here.
However, `ins.IsFunctionCall()` returns false for the FnProbeReadKernel, because ins.Src equals to R0. Can I use a `ins.OpCode.JumpOp() == Call` instead?
https://github.com/cilium/ebpf/blob/master/asm/instruction.go#L177-L182

linker.go
xh4n3: Okay, let me mark this as resolved.

syscalls.go
xh4n3: Will fix it.

linker.go
xh4n3: Updated.

lmb: Good point! Can you add `IsBuiltinCall` which asserts `JumpOp == Call` and `ins.Src == R0` and use that?

xh4n3: Nice suggestion!
I just refer to libbpf and found they have additional ins.Dst == R0 check, so I added this as well.

lmb (changes requested):
LGTM, @ti-mo do you want another look?

lmb (approved):

syscalls.go
```diff
@@ -469,3 +471,34 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
...
+	insns := asm.Instructions{
+		asm.Mov.Reg(asm.R1, asm.R10),
+		asm.Add.Imm(asm.R1, -8),
+		asm.Mov.Imm(asm.R2, 8),
+		asm.Mov.Imm(asm.R3, 0),
+		asm.FnProbeReadKernel.Call(),
+		asm.Return(),
+	}
+	buf := bytes.NewBuffer(make([]byte, 0, len(insns)*asm.InstructionSize))
+	err := insns.Marshal(buf, internal.NativeEndian)
```
rgo3, line 485: nit:
```suggestion
	if err := insns.Marshal(buf, internal.NativeEndian); err != nil {
	    return err
	}
```

syscalls.go
```diff
@@ -469,3 +471,34 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
...
+	if err != nil {
+		return err
+	}
+	bytecode := buf.Bytes()
+
+	attr := &bpfProgLoadAttr{
```
rgo3, line 491: `bpfProgLoadAttr` and `bpfProgLoad()` were moved to internal by me in a PR #355 that just landed so you'll have to rebase.

syscalls.go
```diff
@@ -469,3 +471,34 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
...
+		progType: Kprobe,
+		license:  internal.NewStringPointer("GPL"),
+	}
+	attr.instructions = internal.NewSlicePointer(bytecode)
+	attr.insCount = uint32(len(bytecode) / asm.InstructionSize)
```
rgo3, line 496: nit: this could be moved directly into the struct creation above.

syscalls.go
```diff
@@ -469,3 +471,34 @@ var haveBatchAPI = internal.FeatureTest("map batch api", "5.6", func() error {
...
+
+	fd, err := bpfProgLoad(attr)
```
rgo3, line 498: I think you could do this:
```suggestion
	_, err := bpfProgLoad(attr)
```
to avoid the `fd.Close()` call beneath but I am not sure if its better to close it implicitly.

rgo3 (commented):
Just a couple of nits:

syscalls.go
lmb: An explicit close for fds is better, I think.

xh4n3: Agreed, I think setting fd to `_` will let go runtime to do a simple gc over fd object without calling a close, may lead to a leak.

syscalls.go
xh4n3: Updated.

syscalls.go
xh4n3: Rebased and re-tested.

ti-mo (approved):

PR #370, opened 2021-08-11
btf: prepare for making the package public

lmb:
I think this is the MVP to make `btf` accessible to users.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -491,7 +491,8 @@ func (s *Spec) FindType(name string, typ Type) error {
 
 // Handle is a reference to BTF loaded into the kernel.
 type Handle struct {
-	fd *internal.FD
+	spec *Spec
+	fd   *internal.FD
 }
 
 // NewHandle loads BTF into the kernel.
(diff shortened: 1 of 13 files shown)


internal/btf/btf.go
```diff
@@ -593,43 +581,8 @@ func (h *Handle) FD() int {
 
 // Map is the BTF for a map.
 type Map struct {
```
lmb, line 583: * We could move this type to `ebpf` and hide the fields instead. What would we call the type? `ebpf.MapBTF`?
* We could embed these fields in `MapSpec` directly.

brb: The first option sounds right to me. The type doesn't seem to have much to do with other types in the `btf` package. `ebpf.MapBTF` :+1:

brb (approved):
Thanks for working on this.

internal/btf/btf.go
ti-mo: Why would `btf.Map` be moved and `btf.Program` be kept in place?

lmb: I'd say we keep it as it is now then?

lmb:
@brb I pushed 4f00d3d3333b911474142732ece1e8dac29a7e52 which changes the type copy behaviour we talked about.

@ti-mo are you OK with proceeding with this?

Issue #366, opened 2021-08-05
Require explicit assignment of prog arrays in `LoadAndAssign`


ti-mo:
Continuation of https://github.com/cilium/ebpf/pull/363#discussion_r683304962.

In this library, losing the reference to an fd means closing it. Currently, forgetting to assign a prog array that's lazy-loaded in the background as a dependency will close the prog_array on the first GC, resulting in the prog_array being truncated and missed tail calls.

This issue is for failing `LoadAndAssign` if any prog_arrays are loaded but not assigned to the object.

PR #371, opened 2021-08-11
Deprecate `Collection.Assign()`, refactor `CollectionSpec.(LoadAnd)Assign()`

ti-mo:
Closes https://github.com/cilium/ebpf/issues/366.

Best reviewed per commit.

tl;dr
- deprecated Collection.Assign()
- `flattenStruct()` is now a discrete function in the form of `ebpfFields()`
- `assignValues()` now takes a validator function that can be used to reject the transaction before the struct's fields are actually `Set()`
- `interface{}` is now used instead of a `reflect.Value` for tracking assigned object uniqueness

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -211,6 +211,7 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 	defer loader.cleanup()
 
 	// Support assigning Programs and Maps, lazy-loading the required objects.
+	assignedMaps := make(map[string]bool)
 	getValue := func(typ reflect.Type, name string) (interface{}, error) {
 		switch typ {
 
@@ -218,6 +219,7 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 			return loader.loadProgram(name)
 
 		case reflect.TypeOf((*Map)(nil)):
+			assignedMaps[name] = true
 			return loader.loadMap(name)
 
 		default:
@@ -230,9 +232,22 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 		return err
 	}
 
-	// Finally, populate the requested maps.
+	// Populate the requested maps. Has a chance of lazy-loading other dependent maps.
 	if err := loader.populateMaps(); err != nil {
 		return err
+	}
+
+	// Evaluate the loader's objects after all (lazy)loading has taken place.
+	for n, m := range loader.maps {
+		switch m.typ {
+		case ProgramArray:
+			// Require all lazy-loaded ProgramArrays to be assigned to the given object.
+			// Without any references, they will be closed on the first GC and all tail
+			// calls into them will miss.
+			if !assignedMaps[n] {
+				return fmt.Errorf("ProgramArray %s must be assigned to prevent missed tail calls", n)
+			}
+		}
 	}
 
 	loader.finalize()
@@ -592,9 +607,12 @@ func ebpfFields(structVal reflect.Value, visited map[reflect.Type]bool) ([]struc
 }
 
 // assignValues attempts to populate all fields of 'to' tagged with 'ebpf'.
-// getValue is called for every such field of and must return the value
+//
+// getValue is called for every tagged field of 'to' and must return the value
 // to be assigned to the field with the given typ and name.
-func assignValues(to interface{}, getValue func(typ reflect.Type, name string) (interface{}, error)) error {
+func assignValues(to interface{},
+	getValue func(typ reflect.Type, name string) (interface{}, error)) error {
+
 	toValue := reflect.ValueOf(to)
 	if toValue.Type().Kind() != reflect.Ptr {
 		return fmt.Errorf("%T is not a pointer to struct", to)
--- a/collection.go
+++ b/collection.go
@@ -209,7 +209,7 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 	}
 
 	loader := newCollectionLoader(cs, opts)
-	defer loader.close()
+	defer loader.cleanup()
 
 	valueOf := func(typ reflect.Type, name string) (reflect.Value, error) {
 		switch typ {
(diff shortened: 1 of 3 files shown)


collection.go
```diff
@@ -159,26 +159,10 @@ func (cs *CollectionSpec) RewriteConstants(consts map[string]interface{}) error
 // Returns an error if any of the fields can't be found, or
 // if the same map or program is assigned multiple times.
 func (cs *CollectionSpec) Assign(to interface{}) error {
```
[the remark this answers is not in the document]
ti-mo: True, didn't consider that. Conversely, LoadAndAssign now also supports assigning Specs where it didn't before.

Now I understand the context a bit better, wouldn't it make sense to keep `Collection.Assign()` around and deprecate `CollectionSpec.Assign()` instead? That way, the caller can assign Programs and Maps from Collections, and Program(Spec)s and Map(Specs) from CollectionSpecs.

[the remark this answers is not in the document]
lmb: Here is why CollectionSpec.Assign exists:
```Go
// from bpf2go
type fooMapSpecs struct
type fooObjects struct

var specs fooMapSpecs
collSpec.Assign(&specs)

specs.FooBar.KeySize = 42

var objs fooObjects
collSpec.LoadAndAssign(&objs) // uses KeySize = 42
```

In my experience, I need access to fooMapSpecs before loading *Map and *Program to do some fudging. Your suggestion to get rid of CollectionSpec.Assign by folding it into CollectionSpec.LoadAndAssign reads kinda funny:

```Go
...
collSpec.LoadAndAssign(&specs, nil)
...
collSpec.LoadAndAssign(&objs, &Options{})
```

It's also awkward that LoadAndAssign needs options. The user has to understand how LoadAndAssign is implemented to figure out whether they should provide options or not.

I'd rather not bring back Collection.Assign since it seems stricly less useful than LoadAndAssign.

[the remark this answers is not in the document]
ti-mo: > I'd rather not bring back Collection.Assign since it seems stricly less useful than LoadAndAssign.

I think it's still useful in the sense that it allows the userspace program to continue with a custom struct instead of a Collection, but now I think about it, there might not be much point in doing so. If you forget to assign any Maps/Progs, those are silently closed on the next GC if the original Collection is no longer referenced. This could be done on purpose, of course, but in that case, LoadAndAssign would be more efficient since it doesn't create unneeded resources. Is this what you meant?

> It's also awkward that LoadAndAssign needs options. The user has to understand how LoadAndAssign is implemented to figure out whether they should provide options or not.

Isn't this already the case? :stuck_out_tongue: In general, I agree. `CollectionSpec.Assign()`'s purpose is clear to me now.

Assuming we're just deprecating `Collection.Assign()` I'll try to come up with a better solution for breaking up `assignValues()`, or at least for providing a way to get access to its internals (map `assignedTo`).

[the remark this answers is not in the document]
ti-mo: Okay, I've restored the `valueOf` structure and (`CollectAnd`)`Assign`'s original behaviour.

collection.go
```diff
@@ -594,39 +604,40 @@ func assignValues(to interface{}, valueOf func(reflect.Type, string) (reflect.Va
 		return fmt.Errorf("nil pointer to %T", to)
 	}
 
-	if err := flattenStruct(toValue.Elem()); err != nil {
-		return err
+	// Follow the pointer to the provided object and collect
+	// its fields that were tagged with 'ebpf'.
+	fields, err := ebpfFields(toValue.Elem(), nil)
+	if err != nil {
+		return fmt.Errorf("gathering field tags: %w", err)
```
lmb, line 637: Nit: `gather`?

collection.go
```diff
@@ -594,39 +604,40 @@ func assignValues(to interface{}, valueOf func(reflect.Type, string) (reflect.Va
...
-
-		value, err := valueOf(field.Type, name)
+		// Get the eBPF object referred to by the tag.
+		value, err := getValue(field.Type, tag)
 		if err != nil {
 			return fmt.Errorf("field %s: %w", field.Name, err)
 		}
 
+		// Use the interface value as a unique key, since they're comparable.
+		if af, ok := assigned[value]; ok {
```
lmb, line 657: I see why this works, but it makes the code kind of subtle. e.g. we now require that getValue always returns the same value for a given type, tag argument for this check to succeed. I think I prefer the explicit `{reflect.Type, name}` check.

collection.go
```diff
@@ -225,8 +225,24 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 		}
 	}
 
+	validate := func(assigned map[interface{}]structField) error {
```
lmb, line 230: I find closure arguments to functions always a bit unergonomic, since they end up not being very flexible. In this case it also exposes internal state via `assigned` which makes it harder to change the implementation later.

Why not have a `var assignedMaps map[string]bool` which you populate from the `reflect.TypeOf((*Map)(nil))` arm in `getValue`? Then you can do the same check without the need for a closure, after `assignValues` has succeeded.

collection.go
ti-mo: After offline discussion, since `assigned` doesn't technically need to be exported, this can be reverted.

collection.go
ti-mo: After discussing offline, assigning to the receiver struct before throwing errors is not really an issue, as all assigned objects are closed in case of an error. Will revert a few changes accordingly.

collection.go
```diff
@@ -604,29 +615,32 @@ func assignValues(to interface{}, valueOf func(reflect.Type, string) (reflect.Va
...
-		value, err := valueOf(field.Type, name)
+		// Get the eBPF object referred to by the tag.
+		value, err := getValue(field.Type, tag)
 		if err != nil {
 			return fmt.Errorf("field %s: %w", field.Name, err)
 		}
 
+		// Check if the eBPF object requested by the field was already assigned elsewhere.
+		e := elem{field.Type, tag}
+		if af := assigned[e]; af != "" {
```
lmb, line 654: Why did you move the check after `getValue`?

collection.go
```diff
@@ -230,11 +232,24 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
...
 	}
 
+	// Evaluate the loader's objects after all (lazy)loading has taken place.
+	for n, m := range loader.maps {
+		switch m.typ {
+		case ProgramArray:
+			// Require all lazy-loaded ProgramArrays to be assigned to the given object.
+			// Without any references, they will be closed on the first GC and all tail
+			// calls into them will miss.
+			if _, ok := assignedMaps[n]; !ok {
```
lmb, line 249: Nit: `if !assignedMaps[n] {` is enough.

lmb (approved):

collection.go
ti-mo: Reverted this, it was part of my initial patch and thought it would be useful to include the type of the value returned by `getValue()` in the error message. Was useful for troubleshooting during iteration, but doesn't add any real value.

PR #372, opened 2021-08-11
features: prog_test - enable LircMode2 test

ti-mo:
CONFIG_BPF_LIRC_MODE2 is now enabled in CI kernels, so unconditionally
enable this test.

@rgo3 

Follow-up to https://github.com/cilium/ci-kernels/issues/14.


rgo3:
@ti-mo An hour too late but I just realized that removing this check sort of breaks the experience when running tests with a local kernel. Although that was the case for other prog types before as well, I just didn't notice because my kernel only missed lirc_mode2.

ti-mo:
@rgo3 https://github.com/cilium/ebpf/pull/374 reverts this.

PR #375, opened 2021-08-13
features: add HaveProgramHelper API

rgo3:
`HaveProgHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error`
allows to probe the available BPF helpers to a given BPF program
type. Probe results are cached and run at most once.

Signed-off-by: Robin Gögge <r.goegge@gmail.com>

### DIFF
--- a/features/prog.go
+++ b/features/prog.go
@@ -15,7 +15,7 @@ import (
 )
 
 func init() {
-	pc.progTypes = make(map[ebpf.ProgramType]error)
+	pc.types = make(map[ebpf.ProgramType]error)
 	pc.helpers = make(map[ebpf.ProgramType]map[asm.BuiltinFunc]error)
 	allocHelperCache()
 }
@@ -31,8 +31,8 @@ var (
 )
 
 type progCache struct {
-	typeMu    sync.Mutex
-	progTypes map[ebpf.ProgramType]error
+	typeMu sync.Mutex
+	types  map[ebpf.ProgramType]error
 
 	helperMu sync.Mutex
 	helpers  map[ebpf.ProgramType]map[asm.BuiltinFunc]error
@@ -95,17 +95,22 @@ func createProgLoadAttr(pt ebpf.ProgramType, helper asm.BuiltinFunc) (*sys.ProgL
 
 // HaveProgType probes the running kernel for the availability of the specified program type.
 //
+// Deprecated: use HaveProgramType() instead.
+var HaveProgType = HaveProgramType
+
+// HaveProgramType probes the running kernel for the availability of the specified program type.
+//
 // See the package documentation for the meaning of the error return value.
-func HaveProgType(pt ebpf.ProgramType) error {
-	if err := validateProgType(pt); err != nil {
+func HaveProgramType(pt ebpf.ProgramType) error {
+	if err := validateProgramType(pt); err != nil {
 		return err
 	}
 
-	return haveProgType(pt)
+	return haveProgramType(pt)
 
 }
 
-func validateProgType(pt ebpf.ProgramType) error {
+func validateProgramType(pt ebpf.ProgramType) error {
 	if pt > pt.Max() {
 		return os.ErrInvalid
 	}
@@ -120,10 +125,10 @@ func validateProgType(pt ebpf.ProgramType) error {
 	return nil
 }
 
-func haveProgType(pt ebpf.ProgramType) error {
+func haveProgramType(pt ebpf.ProgramType) error {
 	pc.typeMu.Lock()
 	defer pc.typeMu.Unlock()
-	if err, ok := pc.progTypes[pt]; ok {
+	if err, ok := pc.types[pt]; ok {
 		return err
 	}
 
@@ -154,7 +159,7 @@ func haveProgType(pt ebpf.ProgramType) error {
 		fd.Close()
 	}
 
-	pc.progTypes[pt] = err
+	pc.types[pt] = err
 
 	return err
 }
@@ -173,7 +178,7 @@ func haveProgType(pt ebpf.ProgramType) error {
 //
 // Probe results are cached and persist throughout any process capability changes.
 func HaveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
-	if err := validateProgType(pt); err != nil {
+	if err := validateProgramType(pt); err != nil {
 		return err
 	}
 
--- a/features/prog.go
+++ b/features/prog.go
@@ -16,6 +16,14 @@ import (
 
 func init() {
 	pc.progTypes = make(map[ebpf.ProgramType]error)
+	pc.helpers = make(map[ebpf.ProgramType]map[asm.BuiltinFunc]error)
+	allocHelperCache()
+}
+
+func allocHelperCache() {
+	for pt := ebpf.UnspecifiedProgram + 1; pt <= pt.Max(); pt++ {
+		pc.helpers[pt] = make(map[asm.BuiltinFunc]error)
+	}
 }
 
 var (
@@ -23,17 +31,24 @@ var (
 )
 
 type progCache struct {
-	sync.Mutex
+	typeMu    sync.Mutex
 	progTypes map[ebpf.ProgramType]error
+
+	helperMu sync.Mutex
+	helpers  map[ebpf.ProgramType]map[asm.BuiltinFunc]error
 }
 
-func createProgLoadAttr(pt ebpf.ProgramType) (*sys.ProgLoadAttr, error) {
+func createProgLoadAttr(pt ebpf.ProgramType, helper asm.BuiltinFunc) (*sys.ProgLoadAttr, error) {
 	var expectedAttachType ebpf.AttachType
 	var progFlags uint32
 
 	insns := asm.Instructions{
 		asm.LoadImm(asm.R0, 0, asm.DWord),
 		asm.Return(),
+	}
+
+	if helper != asm.FnUnspec {
+		insns = append(asm.Instructions{helper.Call()}, insns...)
 	}
 
 	buf := bytes.NewBuffer(make([]byte, 0, insns.Size()))
@@ -106,14 +121,13 @@ func validateProgType(pt ebpf.ProgramType) error {
 }
 
 func haveProgType(pt ebpf.ProgramType) error {
-	pc.Lock()
-	defer pc.Unlock()
-	err, ok := pc.progTypes[pt]
-	if ok {
+	pc.typeMu.Lock()
+	defer pc.typeMu.Unlock()
+	if err, ok := pc.progTypes[pt]; ok {
 		return err
 	}
 
-	attr, err := createProgLoadAttr(pt)
+	attr, err := createProgLoadAttr(pt, asm.FnUnspec)
 	if err != nil {
 		return fmt.Errorf("couldn't create the program load attribute: %w", err)
 	}
@@ -124,7 +138,7 @@ func haveProgType(pt ebpf.ProgramType) error {
 	// EINVAL occurs when attempting to create a program with an unknown type.
 	// E2BIG occurs when ProgLoadAttr contains non-zero bytes past the end
 	// of the struct known by the running kernel, meaning the kernel is too old
-	// to support the given map type.
+	// to support the given prog type.
 	case errors.Is(err, unix.EINVAL), errors.Is(err, unix.E2BIG):
 		err = ebpf.ErrNotSupported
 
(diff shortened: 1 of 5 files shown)


rgo3:
Compared to bpftool I just check for EACCES and EINVAL for now instead of checking the [verifier log](https://github.com/torvalds/linux/blob/1746f4db513563bb22e0ba0c419d0c90912dfae1/tools/lib/bpf/libbpf_probes.c#L321-L323). Not sure if that suffices and if there is an (future) edge case that could break this and result in a false negative/positive. cc @ti-mo @qmonnet

qmonnet:
Did you compare the results with bpftool's (minus the ones for the few program types that aren't working properly on bpftool)? Do you get the same thing?

rgo3:
Initially I just checked a few helpers for equality randomly, but now did the full comparison and I actually come up short for like 5 helper calls regarding the `bpf_get_current_task_btf` helper. So I guess just checking for EINVAL isn't enough :(...

qmonnet:
So there's just the `bpf_get_current_task_btf()` that is not detected as available with your PR, is this correct? I suppose your kernel is >= 5.11? (Asking because it would be nice to be sure where this `EINVAL` comes from)

rgo3:
Yes, just `bpf_get_current_task_btf()`. I think the EINVAL comes from the verifier [here](https://github.com/torvalds/linux/blob/f8e6dfc64f6135d1b6c5215c14cd30b9b60a0008/kernel/bpf/verifier.c#L6155). I modified bpftool so it outputs the verifier log and the error message for `bpf_get_current_task_btf()` is: invalid return type 11 of func bpf_get_current_task_btf#158

qmonnet:
A few lines above then, because (again) there's no BTF. `invalid return type` is not the same as `unknown return type`. The `unknown return type` case would mean the helper would never pass the verifier. Anyway :).

Yeah, in that case it's not sure there's a way other than parsing the log. We could start making cases for helpers too, but this will be harder to maintain I'm afraid.

rgo3:
Oh yeah I selected the wrong line :see_no_evil:.
I and Timo talked about this yesterday, how likely are the verifier messages to change aka how stable is it to use them as an API?

qmonnet:
Tricky question. But then the error code from the verifier are hardly a stable API either :shrug:.

lmb:
Drive by observation: at least for error codes developers tend to try to preserve them. The same can't be said for the verifier log, and I think we shouldn't rely on it.

rgo3:
Hey! :wave: Apologies for letting this PR go stale a bit, but I would like to pick it up again so that I eventually can make use of it in cilium.

To sum up the current status: Just relying on error codes can result in false negatives, e.g. `bpf_get_current_task_btf()`. We could rely on verifier log output, but I agree with Lorenz that verifier output is more likely subject to change (But I am new, so that is just a feeling to be honest).

To get this PR merged I can think of two possible solutions:

* If we want to rely on error codes, we would have to make sure EINVAL is only ever returned when the helper isn't available. We can do this by adding exception cases to various helper probes, similar to what we have with the other probe APIs that were merged. As there are so many helpers, the maintenance burden of these exception cases might not actually be better than relying on the verifier output, because we have to cover more code paths of the verifier now.

* Go with verifier log, accepting the risk that the messages might change, but this API then only codes against a few specific code paths in the verifier reducing complexity of the probes.

I hope @ti-mo and @lmb can give their opinions on the two solutions I proposed or maybe come up with an even better option.

ti-mo:
> Is it an option to move forward with this PR keeping it the way it is currently, basically not trying to create a catch-em-all solution right from the start?

I'm okay with the initial implementation not being perfect, as long as we have a way of documenting what the errata are. Preferably in tests, with a GH issue to track the follow-up work. That way, if anyone has the need for a particular broken probe, they can grab the issue and go ahead with the implementation.

lmb:
What's going to happen to this?

rgo3:
I'd still like to get some version of this in, considering this would be the most useful when trying to integrate the features API into cilium (which was the original goal of my GSoC project). 

I got stuck looking for ways to create tests that nicely show what the errata are because of all the possible combinations between prog types and available helpers on respective kernel versions. I could add tests per prog type, start with the prog types I'd need to use this in cilium and add more over time? Like that I wouldn't have to test the full matrix of possibilities for now but only a smaller subset. If yes I'd try to navigate along this [list](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md) of versions and available helpers per prog type to incrementally add tests?

If that's not a clean enough solution to move forward with this PR, I might need to rethink how we could add helper probes to this API.

rgo3:
Alright, I think it's finally ready to graduate from its draft status.
There's still a test-case that I added that's failing, but that's related to the kernel config in CI:
```
--- FAIL: TestHaveProgHelper (0.04s)
    ...
    --- FAIL: TestHaveProgHelper/CGroupSockAddr/FnGetCgroupClassid (0.00s)
    ...
```
Locally this test works for me, as my kernel has this helper type compiled in.

Additionally there are two more test-cases I couldn't add because the lib doesn't support those particular builtins yet (just a matter of updating the BuiltinFunc enum I think). I've mentioned those in comments above the test-cases.

I've also built a small tool to compare the features API to bpftools feature probes. Currently just living on a [branch](https://github.com/rgo3/ebpf/tree/bpftoolcmp) of my fork of the repo. Running that tool yields following output:
```
❯ go run -exec sudo cmd/bpftoolcmp/main.go
Running bpftool -j feature ...

Comparing available program types ...

Comparing available helper functions ...
    False negative: API got different result than bpftool for: RawTracepoint/FnGetFuncIp
    False negative: API got different result than bpftool for: Syscall/FnSkStorageGet
    False negative: API got different result than bpftool for: Syscall/FnSkStorageDelete
    False negative: API got different result than bpftool for: Syscall/FnDPath
    False negative: API got different result than bpftool for: Syscall/FnGetFuncIp
    False negative: API got different result than bpftool for: TracePoint/FnGetFuncIp
    False negative: API got different result than bpftool for: PerfEvent/FnGetFuncIp
    False negative: API got different result than bpftool for: RawTracepointWritable/FnGetFuncIp
```
I haven't fully investigated why those false negatives exist, but I'd create an issue to track and investigate if we choose to accept  that these differ from bpftool currently. 
This output excludes false negatives for program types `LSM` , `Extension`, `StructOps` and `Tracing` because we can't even probe those prog types correctly.

cc @florianl

lmb:
Could `bpftoolcmp` just be part of the testsuite? We could skip tests if the binary doesn't exist.

rgo3:
So instead of having `bpftoolcmp` add a test like e.g. `TestBPFToolCompat`? And skip on missing binary and known exceptions?

florianl (approved):
Thanks for bringing this forward! :+1:
The PR as is works fine here.

For possible future work wrt `bpftoolcmp` I would also vote for having this in a test case like `TestBPFToolCompare`. Rather than calling an external tool that `exec.Cmd()` another external tool (`bpftool`).

rgo3:
Sorry, had to repush as I found some typos...

As for the `FnGetFuncIp` helper probes: I think these are actually false positives on bpftools/libbpfs end. The verifier has a [check](https://github.com/torvalds/linux/blob/f3f19f939c11925dadd3f4776f99f8c278a7017b/kernel/bpf/verifier.c#L6597) that only allows `Kprobe` and `Tracing` progs to call it. This API returns an error as `ENOTSUPP` is returned from the verifier. libbpf misses it because [checking](https://github.com/libbpf/libbpf/blob/87dff0a2c775c5943ca9233e69c81a25f2ed1a77/src/libbpf_probes.c#L414-L427) the verifier log output doesn't match in this case. cc @qmonnet

features/prog.go
```diff
@@ -145,6 +160,89 @@ func haveProgType(pt ebpf.ProgramType) error {
...
+	}
+
+	return nil
+}
+
+func haveProgHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
+	pc.helperMu.Lock()
+	defer pc.helperMu.Unlock()
+	err, ok := pc.progHelpers[pt][helper]
+	if ok {
```
ti-mo, line 200: To indicate the short lifespan of `err`:

```suggestion
	if err, ok := pc.progHelpers[pt][helper]; ok {
```

features/prog_test.go
```diff
@@ -92,3 +95,85 @@ func TestHaveProgTypeInvalid(t *testing.T) {
...
+	// Referecing linux kernel commits to track the kernel version required to pass these test-cases
+	// These test-cases are derived from libbpf's selftests and helper/prog combinations that are
+	// probed for in cilium/cilium
+	// Still missing since those helpers are not available in the lib yet, are:
+	// - Kprobe, GetBranchSnapshot
+	// - SchedCLS, SkbSetTstamp
+
+	testCases := []testCase{
+		{ebpf.Kprobe, asm.FnMapLookupElem, nil, "3.19"},                     // @d0003ec
+		{ebpf.Kprobe, asm.FnKtimeGetCoarseNs, ebpf.ErrNotSupported, "5.16"}, // @5e0bc30
```
ti-mo, line 116: Feel free to comment this one out until 5.17 testing is live in CI.

I've added a note to https://github.com/cilium/ebpf/pull/668 to make sure this gets uncommented before merging.

features/prog_test.go
```diff
@@ -92,3 +95,85 @@ func TestHaveProgTypeInvalid(t *testing.T) {
...
+
+	// Referecing linux kernel commits to track the kernel version required to pass these test-cases
+	// These test-cases are derived from libbpf's selftests and helper/prog combinations that are
+	// probed for in cilium/cilium
+	// Still missing since those helpers are not available in the lib yet, are:
+	// - Kprobe, GetBranchSnapshot
+	// - SchedCLS, SkbSetTstamp
+
+	testCases := []testCase{
+		{ebpf.Kprobe, asm.FnMapLookupElem, nil, "3.19"},                     // @d0003ec
```
ti-mo, line 115: Nit: the `@` prefix gets included in the selection if I double-click the word in my editor. :cry:

features/prog_test.go
```diff
@@ -92,3 +95,85 @@ func TestHaveProgTypeInvalid(t *testing.T) {
...
+
+func TestHaveProgHelper(t *testing.T) {
+	type testCase struct {
+		prog     ebpf.ProgramType
+		helper   asm.BuiltinFunc
+		expected error
+		version  string
+	}
+
+	// Referecing linux kernel commits to track the kernel version required to pass these test-cases
```
ti-mo, line 107:
```suggestion
	// Referencing linux kernel commits to track the kernel version required to pass these test cases
```

features/prog_test.go
```diff
@@ -92,3 +95,85 @@ func TestHaveProgTypeInvalid(t *testing.T) {
...
+	// These test-cases are derived from libbpf's selftests and helper/prog combinations that are
```
ti-mo, line 108:
```suggestion
	// These test cases are derived from libbpf's selftests and helper/prog combinations that are
```

features/prog_test.go
```diff
@@ -92,3 +95,85 @@ func TestHaveProgTypeInvalid(t *testing.T) {
...
+		{ebpf.SchedCLS, asm.FnRedirectNeigh, nil, "5.10"},                   // @b4ab314
+		{ebpf.SchedCLS, asm.FnSkbEcnSetCe, nil, "5.1"},                      // @f7c917b
+		{ebpf.SchedACT, asm.FnSkAssign, nil, "5.6"},                         // @cf7fbe6
+		{ebpf.SchedACT, asm.FnFibLookup, nil, "4.18"},                       // @87f5fc7
+		{ebpf.CGroupSockAddr, asm.FnGetCgroupClassid, nil, "5.10"},          // @b426ce8
+		{ebpf.CGroupSockAddr, asm.FnGetsockopt, nil, "4.15"},                // @cd86d1f
+		{ebpf.CGroupSockAddr, asm.FnSkLookupTcp, nil, "4.20"},               // @6acc9b4
+		{ebpf.CGroupSockAddr, asm.FnGetNetnsCookie, nil, "5.7"},             // @f318903
+		{ebpf.CGroupSock, asm.FnGetNetnsCookie, nil, "5.7"},                 // @f318903
+
```
ti-mo, line 133: Could we add one or two more expected `ErrNotSupported` combos from older (4.x) kernels? This would ensure our detection strategy is sound throughout the history. The current negative tests only run on 5.14 and above (modulo `TestHaveProgHelperUnsupported`, but that one might trip early user input checks in the syscall handler)

ti-mo (approved):
Love it.

> Could bpftoolcmp just be part of the testsuite? We could skip tests if the binary doesn't exist.

This can be done in a follow-up. :+1:

features/prog_test.go
rgo3: Commented two test cases out, which require CI kernel changes.

features/prog_test.go
rgo3: Added:
* `{ebpf.Kprobe, asm.FnSkbVlanPush, ebpf.ErrNotSupported, "4.3"}, // 4e10df9a60d9`
* `{ebpf.Kprobe, asm.FnFibLookup, ebpf.ErrNotSupported, "4.18"},  // 87f5fc7e48dd`

lmb:
Bit of a late bike shed: can you call the function HaveProgramHelper?

PR #378, opened 2021-08-19
changed slack channel link

mabis:
since the channel was renamed yesterday from #libbpf-go to #ebpf-go

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ be used in long running processes.
 
 The library is maintained by [Cloudflare](https://www.cloudflare.com) and
 [Cilium](https://www.cilium.io). Feel free to
-[join](https://cilium.herokuapp.com/) the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ be used in long running processes.
 The library is maintained by [Cloudflare](https://www.cloudflare.com) and
 [Cilium](https://www.cilium.io). Feel free to
 [join](https://cilium.herokuapp.com/) the
```
markpash, line 20: I think while we're here cilium.herokuapp.com could be replaced by ebpf.io/slack. While it seems to just be a redirect at the moment, it's probably what cilium wants to be here.

ti-mo: Yep, feel free to change the join link as well, thanks :+1:

mabis: done

PR #380, opened 2021-08-19
Add //go:build lines

tklauser:
Starting with Go 1.17, `//go:build` lines are preferred over `// +build`
lines, see https://golang.org/doc/go1.17#build-lines and
https://golang.org/design/draft-gobuild for details.

This change was generated by running Go 1.17 `gofmt -w .` which
automatically adds `//go:build` lines based on the existing `// +build`
lines.


tklauser:
Will need to change `bpf2go` to generate these lines as well: https://ebpf.semaphoreci.com/jobs/bf2b1b84-a4a4-42b4-b59a-e9e968e57c51

tklauser:
> Will need to change `bpf2go` to generate these lines as well: https://ebpf.semaphoreci.com/jobs/bf2b1b84-a4a4-42b4-b59a-e9e968e57c51

Pushed a second commit changing `bpf2go` to emit `//go:build` lines.

internal/ptr_64.go
```diff
@@ -1,5 +1,5 @@
-// +build !386,!amd64p32,!arm,!mipsle,!mips64p32le
-// +build !armbe,!mips,!mips64p32
+//go:build !386 && !amd64p32 && !arm && !mipsle && !mips64p32le && !armbe && !mips && !mips64p32
```
ti-mo, line 1: I've noticed fmt understands arch exclusions and added a few of its own. Wouldn't it be more concise to turn this into an allow list now?

ti-mo (commented):
Thanks! Also noticed this yesterday trying out 1.17. Left one comment about something I'd like us to investigate.

internal/ptr_64.go
tklauser: I don't think `gofmt` added any new arch exclustions, it just re-arranged the existing ones differently.

From my perspective it'd make sense to turn this into an allow list, as it would also mean we have to explicitly update the list whenever a new architecture is added to Go or gccgo (and ideally verify that `cilium/ebpf` works on that architecture).

/cc @lmb who initially added that code

tklauser: FWIW, this would be the current list of 64-bit architectures for the allow list:

```go
// +build alpha amd64 arm64 arm64be mips64 mips64le ppc64 ppc64le riscv64 s390x sparc64
```

ti-mo: Alright, that doesn't look much better. Looks like there was a proposal (https://github.com/golang/go/issues/33388) for 32/64-bit build tags, but that didn't make it. Let's leave this as-is.

PR #381, opened 2021-08-19
Rename 'StructOpts' to StructOpsMap

ti-mo:
This seems to have been a typo that slipped in unnoticed. The old 'StructOpts' has been kept around as a deprecated alias
to preserve backwards-compatibility and will be removed in a subsequent release.

@mythi @rgo3

### DIFF
--- a/types.go
+++ b/types.go
@@ -90,9 +90,9 @@ const (
 	SkStorage
 	// DevMapHash - Hash-based indexing scheme for references to network devices.
 	DevMapHash
-	// StructOps - This map holds a kernel struct with its function pointer implemented in a BPF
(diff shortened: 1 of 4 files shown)


types.go
```diff
@@ -90,9 +90,9 @@ const (
 	SkStorage
 	// DevMapHash - Hash-based indexing scheme for references to network devices.
 	DevMapHash
-	// StructOps - This map holds a kernel struct with its function pointer implemented in a BPF
+	// StructOpsMap - This map holds a kernel struct with its function pointer implemented in a BPF
 	// program.
-	StructOpts
```
mythi, line 95: @ti-mo sorry about this. looks like to comment is correct!

PR #383, opened 2021-08-23
bpf2go: revert explicit build constraint support

lmb:
all: re-generate bpf2go files using Go 1.17
    
    From Go 1.17 onwards go/format.Source will automatically convert
    old build tags into build constraints. This means we don't need
    explicit build constraint support in bpf2go. Instead, invoking
    bpf2go with a 1.17 toolchain is enough.

Revert "cmd/bpf2go: generate //go:build lines"
    
    This reverts commit 1ca0c4eea6690e708b741e66ee14b6a1966b995a.
    
    The behaviour for -tags is both subtle and underspecified. It is meant
    to be used like
    
        -tags "a b"
    
    which will then emit an additional //+build line into the generated source.
    The go toolchain interprets multiple lines by ANDing the conditions. The
    given example would only include the generated file in a build if the target
    specific tags and either a or b are satisfied.
    
    The change treats -tags as if it can only contain a single tag that is ORed
    with the target specific tags.

@tklauser I'd like to undo your change so that #369 applies cleanly again. I think we also should deprecate `-tags` in a follow up, it's difficult to support it across the two different syntaxes.


tklauser (approved):
Wasn't aware of the `-tags` intricacies, sorry about that. Given that we now default to Go 1.17 the change LGTM.

Issue #290, opened 2021-04-21
Automatically bump process resource limits


ti-mo:
Currently, users of the library need to explicitly call `unix.Setrlimit` to increase the process' maximum amount of memory available for memlocking eBPF maps. [A series of patches](https://patchwork.kernel.org/project/netdevbpf/list/?series=394251&state=*) recently landed in kernel v5.11 that switches bpf map memory accounting to be cgroup-based. As far as I understand, the consequence for userspace is that setrlimit simply no longer needs to be called.

Following this change, I would like us to investigate whether we can alleviate the user's burden of calling Setrlimit manually. The strongest argument against the library doing this automatically is the fact that it's not the library's responsibility of managing process limits, and that it adds a potential footgun. After all, these limits exist for a reason.

In practice, though, I strongly suspect that most (if not all) eBPF users set `RLIM_INFINITY` without thinking twice about it. I'm not sure if it's possible to determine a safe, high-enough value up front without without running the risk of runtime errors. For example, both the Cilium and Datadog agents set `RLIM_INFINITY`. Could I even conclude that RLIM_INFINITY is basically a requirement?

With the above in mind, I see a few options:
- Naively attempt to create a map and try to bump rlimit on EPERM. If both map creation and rlimit fail, there's no way map creation can continue, so the library can stop there. If rlimit is bumped successfully, retry map creation. If that also returns EPERM, the user really has insufficient permissions. We can do a few things to improve error reporting, like checking euid or capabilities. (libbpf does this) On kernels 5.11 and above, setrlimit would never be touched on the happy path unless the user lacks permissions.
- Provide a simple public function in the library to bump the rlimit. Currently, the caller needs to import `x/sys/unix` and basically make a manual syscall.

lmb:
Personally, a public function in the library seems like the best trade off. It's easy enough to add / maintain, and the callsite serves as a reminder in user code that rlimit is being touched. At some point we can deprecate it. Something as simple as `RemoveMemlockRlimit() error` or something should do.

PR #392, opened 2021-08-31
Utility function to remove memlock rlimit

folbricht:
This should implement the suggestion in https://github.com/cilium/ebpf/issues/290 if I understood it correctly.

Wasn't sure what package this would best live in, let me know if there's a better location/name.

Going to open another PR to update the examples once (if) this gets merged.

### DIFF
--- a/internal/testutils/rlimit.go
+++ b/internal/testutils/rlimit.go
@@ -10,11 +10,7 @@ import (
 func init() {
 	// Increase the memlock for all tests unconditionally. It's a great source of
 	// weird bugs, since different distros have different default limits.
-	err := unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{
-		Cur: unix.RLIM_INFINITY,
-		Max: unix.RLIM_INFINITY,
-	})
-	if err != nil {
+	if _, err := unix.RemoveMemlockRlimit(); err != nil {
 		fmt.Fprintln(os.Stderr, "WARNING: Failed to adjust rlimit, tests may fail")
 	}
 }
--- a/internal/unix/types_linux.go
+++ b/internal/unix/types_linux.go
@@ -5,6 +5,7 @@ package unix
 
 import (
 	"bytes"
+	"fmt"
 	"syscall"
 
 	linux "golang.org/x/sys/unix"
@@ -70,11 +71,6 @@ type Statfs_t = linux.Statfs_t
 
 // Rlimit is a wrapper
 type Rlimit = linux.Rlimit
-
-// Setrlimit is a wrapper
-func Setrlimit(resource int, rlim *Rlimit) (err error) {
-	return linux.Setrlimit(resource, rlim)
-}
 
 // Syscall is a wrapper
 func Syscall(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err syscall.Errno) {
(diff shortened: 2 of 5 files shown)


florianl:
Personal opinion, but I think it is not a good idea to set limits to infinity without the option to set them to their original values. Rlimits often need to be increased to load eBPF resources but should then be set again to the original value, I think.

lmb:
@florianl I think in practice everyone just bumps RLIMIT_MEMLOCK unconditionally. We could return the original memlock limit from the function, to enable users to reset the limit themselves. But then they can just call the syscall directly, I would think?

What I wonder: should this function be `RemoveMemlockLimitIfNecessary` (needs a better name), and only bump MEMLOCK if on a kernel that doesn't have memcg accounting? Caveat: I don't know how to detect whether a kernel has memcg accounting or not.

florianl:
Maybe my context is different, but I think it is good practice to only increase RLimit while loading eBPF resources.

Staying with `RLIM_INFINITY` might lead to resource exhaustion or resource scaling issues.

E.g. cilium/cilium also follows this practice:
https://github.com/cilium/cilium/blob/1695d9c59ac4e78b5a02a96e83a57ec07ddbaa7f/pkg/bpf/bpf_linux.go#L633-L644

lmb:
I don't think that is true in general for the cilium agent: https://github.com/cilium/cilium/blob/1695d9c59ac4e78b5a02a96e83a57ec07ddbaa7f/daemon/cmd/daemon.go#L340 I shared your hesitancy re. RLIMIT_MEMLOCK, but it's what we've been using in production for several years without adverse effects. Do know of cases where this caused a problem?

ti-mo:
Although still potentially racy, I think returning a function that can be used to reset the rlimit to its value before the infinity bump could make sense here. The caller remains in control, and they're free to ignore the function. This would be rather convenient, and maintains rlimit's footgun removal feature. :slightly_smiling_face:

florianl:
Monitoring `RLIMIT_MEMLOCK` can indicate unintended memory leaks and/or trigger spawning of instances. Depending how the environment is used and configured.
So if a package needs to increase it temporary, I think it would be nice to provide a way to set it to its original values afterwards.

folbricht:
Thanks, I updated it to return a function that resets the value as suggested. Any thoughts on a good name for it? Would you prefer `IncreaseRlimit()`?

lmb:
> Would you prefer IncreaseRlimit()?

Also nice, both work for me.

> What I wonder: should this function be RemoveMemlockLimitIfNecessary (needs a better name), and only bump MEMLOCK if on a kernel that doesn't have memcg accounting? Caveat: I don't know how to detect whether a kernel has memcg accounting or not.

Do you have thoughts on this?

folbricht:
I wasn't able to find a way to detect if memcg is available. Found https://lore.kernel.org/bpf/CAEf4BzYe1SDnWBzAP_U7NtUhu_cWa0EEMLv+d-q3YTFvP+y3og@mail.gmail.com/ which discussed the same issue but also didn't come up with a solution. The best alternative I can think of at the moment would be to simply document this in the doc string of the function, saying something along the way of "not necessary on kernels 5.11+ (?)"

process.go
```diff
@@ -0,0 +1,33 @@
+package ebpf
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
```
lmb, line 6: Please import `github.com/cilium/ebpf/internal/unix` instead, that contains dummy definitions so that working on the code base works on non-linux OS.

lmb (approved):

ti-mo (changes requested):
So sorry, thought of one more quick change after I clicked submit: could we also point the rlimit code in `internal/testutils/rlimit.go:init()` to this function instead? That way, we get test coverage for free. 

Will require moving this function to `internal/unix` and adding a small wrapper or alias in package ebpf.

process.go
folbricht: Good call, I had no idea there was one. Done.

folbricht:
Thank you

- Updated to use `github.com/cilium/ebpf/internal/unix`
- `internal/testutils/rlimit.go:init()` uses it too now

TODO (in a separate PR) is to update what's under `examples/`. There was another PR that requires updating the examples anyway.

lmb:
Works for me. @ti-mo?

ti-mo (approved):
I've squashed the commits. Also found a typo, removed `unix.Setrlimit` and moved the wrapper to `syscalls.go`, didn't really warrant its own file.

ti-mo:
I've also found [`prlimit(2)`](https://man7.org/linux/man-pages/man2/prlimit.2.html) on Linux that can perform this operation atomically, which would improve the safety of this call. This isn't exposed in the Go stdlib, though, but there's a playground snippet floating around. I'll create an issue.

PR #394, opened 2021-08-31
Add a script to update vendored libbpf headers for examples

markpash:
I've added a simple script that fetches a specific version of libbpf and extracts the desired headers. This is to have a quick way to get new versions of the headers and ensure consistency.


examples/headers/update.sh
```diff
@@ -0,0 +1,19 @@
...
+LIBBPF_VERSION=0.4.0
+
+# The headers we want
+headers=(
+    bpf_helper_defs.h
+    bpf_helpers.h
+)
+
+# Fetch libbpf release and store in /tmp
+curl -sL "https://github.com/libbpf/libbpf/archive/refs/tags/v${LIBBPF_VERSION}.tar.gz" -o /tmp/libbpf.tar.gz
```
lmb, line 13: Maybe write to `$(mktemp)` and add a `trap EXIT` to remove the file?

examples/headers/update.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Get each header from the libbpf tar.gz and place in the current directory
+for header in "${headers[@]}"
+do
+   tar xf /tmp/libbpf.tar.gz --strip-components=2  libbpf-"$LIBBPF_VERSION"/src/"$header"
```
lmb, line 18: This will probably get really slow once we get more headers, but it's probably fine for now.

examples/headers/update.sh
markpash: I think I've circumvented the need for a temp file in the latest push.

examples/headers/update.sh
markpash: I've condensed it all into one invocation of tar in the latest push.

lmb (approved):

lmb:
Thanks!

ti-mo (commented):
Thanks!

PR #395, opened 2021-09-01
program: don't retrieve map ids by default

lmb:
nr_map_ids was added to struct bpf_prog_info in v4.15. If we
specify a non-zero value for the field in kernels < v4.15 we'll get
an EINVAL from the syscall. This isn't covered by our testsuite since
we don't test against such a kernel, so strictly speaking we don't
have to support this. As per Hyrum's law it's probably better to
go the extra mile however.

### DIFF
--- a/info.go
+++ b/info.go
@@ -97,9 +97,7 @@ type ProgramInfo struct {
 }
 
 func newProgramInfoFromFd(fd *internal.FD) (*ProgramInfo, error) {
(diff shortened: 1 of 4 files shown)


lmb:
cc @arikachen

ti-mo (approved):
:rocket: Test still needs a SkipIfNotSupported for pre-4.10

PR #402, opened 2021-09-02
Add a tracepoint example and some other related stuff

hao-lee:


examples/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.o
```
lmb, line 1: If we gitignore the artefacts it's not possible to `go install` or `go run` the examples, which I think is kind of neat.

examples/README.md
```diff
@@ -12,3 +12,12 @@
 cd ebpf/examples/
 go run -exec sudo [./kprobe, ./uretprobe, ./tracepoint, ...]
 ```
+
+## How to compile
```
lmb, line 16: This kind of duplicates the section above, is it necessary?

examples/kprobe_percpu/main.go
```diff
@@ -0,0 +1,81 @@
+//go:build linux
+// +build linux
+
+// This program demonstrates attaching an eBPF program to a kernel symbol.
+// The eBPF program will be attached to the start of the sys_execve
```
lmb, line 5: Mention the use of a per CPU map instead?

lmb (commented):
Is adding a full example to demonstrate per-CPU maps the best way to go? There is a lot of code here that has nothing to do with per-CPU maps which the user has to ignore.

There is already an example of per CPU map usage here: https://pkg.go.dev/github.com/cilium/ebpf#example-Map-PerCPU How can we make that more discoverable? Maybe you saw it but it didn't help?

examples/.gitignore
hao-lee: > There is already an example of per CPU map usage here: https://pkg.go.dev/github.com/cilium/ebpf#example-Map-PerCPU How can we make that more discoverable? Maybe you saw it but it didn't help?

Yes. This example uses pure Go to manipulate the percpu map, but I think a C example is also necessary for users who want to put the main logic in C files.

hao-lee:
> There is already an example of per CPU map usage here: https://pkg.go.dev/github.com/cilium/ebpf#example-Map-PerCPU How can we make that more discoverable? Maybe you saw it but it didn't help?

Yes. This example uses pure Go to manipulate the percpu map, but I think a C example is also necessary for users who want to put the main logic in C files.

examples/README.md
hao-lee: Yeah, I will merge them.

examples/kprobe_percpu/main.go
hao-lee: 😅 I overlooked these lines.

examples/.gitignore
hao-lee: @lmb Hi, after thinking carefully, I still think we should ignore all these intermediate files and instruct users on how to compile and build step by step. After all, `go run -exec sudo` hides many details.

hao-lee: @lmb hello 😃

examples/kprobe_percpu/main.go
```diff
@@ -0,0 +1,81 @@
...
+	// Name of the kernel function to trace.
+	fn := "sys_execve"
+
+	// Subscribe to signals for terminating the program.
+	stopper := make(chan os.Signal, 1)
+	signal.Notify(stopper, os.Interrupt, syscall.SIGTERM)
+
+	// Increase the rlimit of the current process to provide sufficient space
+	// for locking memory for the eBPF map.
+	if err := unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{
```
lmb, line 36: Can you use the new `RemoveMemlockRlimit`?

examples/tracepoint_in_c/bpf/handler.c
```diff
@@ -0,0 +1,41 @@
...
+	u32 key = 0;
+	u64 initval = 1, *valp;
+
+	valp = bpf_map_lookup_elem(&counting_map, &key);
+	if (!valp) {
+		bpf_map_update_elem(&counting_map, &key, &initval, BPF_ANY);
+		return 0;
+	}
+	__sync_fetch_and_add(valp, 1);
+	bpf_printk("pfn=%lu order=%u gfp_flags=0x%x\n", info->pfn, info->order,
```
lmb, line 38: Can you remove this?

examples/kprobe_percpu/main.go
```diff
@@ -0,0 +1,81 @@
...
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/cilium/ebpf/link"
+	"golang.org/x/sys/unix"
+)
+
+//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang-11 KProbeExample ./bpf/kprobe_percpu_example.c -- -I../headers
```
lmb, line 21:
```suggestion
//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang-11 KProbeExample ./bpf/kprobe_percpu_example.c -- -nostdinc -I../headers
```

lmb: Same for the other occurrences

lmb (changes requested):
Some small stuff, but otherwise I think this is good to merge. The percpu example is more verbose than I'd like, but we can fix that in a follow up.

hao-lee:
😅 Thanks!

Issue #407, opened 2021-09-07
Implement `RemoveMemlockRlimit()` using `prlimit(2)`


ti-mo:
See https://github.com/cilium/ebpf/pull/392#issuecomment-914265729 for context.

`RemoveMemlockRlimit()` is currently implemented using separate Getrlimit/Setrlimit syscalls, which could potentially lead to unexpected behaviour when calling the returned function.

Linux supports `prlimit(2)`, which writes the previous rlimit value to a struct provided by userspace within a single syscall. The Go stdlib doesn't implement `prlimit(2)` for some reason, but here's a snippet that does: https://play.golang.org/p/j9xp57BDrk.

PR #408, opened 2021-09-07
internal/unix: implement RemoveMemlockRlimit() using prlimit(2)

ti-mo:
Closes https://github.com/cilium/ebpf/issues/407

This returns the original rlimit values overwritten during the syscall, making it safer to call concurrently.

### DIFF
--- a/internal/unix/types_linux.go
+++ b/internal/unix/types_linux.go
@@ -202,21 +202,17 @@ func KernelRelease() (string, error) {
 }
 
 func RemoveMemlockRlimit() (func() error, error) {
-	oldLimit := new(Rlimit)
-	if err := linux.Getrlimit(RLIMIT_MEMLOCK, oldLimit); err != nil {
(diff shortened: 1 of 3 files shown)


internal/unix/types_linux.go
```diff
@@ -202,22 +205,44 @@ func KernelRelease() (string, error) {
...
+	nl := unsafe.Pointer(&newLimit)
+
+	_, _, errno := linux.RawSyscall6(SYS_PRLIMIT64,
+		0, // pid 0 means own process
+		uintptr(resource),
+		uintptr(nl),
+		uintptr(ol),
+		0, 0)
+
+	runtime.KeepAlive(ol)
```
lmb, line 240: Do the `uintptr(unsafe.Pointer)` conversion and you don't need these, see point (4) https://pkg.go.dev/unsafe#Pointer

ti-mo: Ah, thanks! Was wondering if that was valid. :+1:

lmb (approved):

PR #410, opened 2021-09-08
ringbuf: avoid leaking epoll FD on invalid map size in NewReader

tklauser:
If the ringbuffer map size is zero or not a power of two,
ringbuf.NewReader would leak the epoll file descriptor it just created.

Fix this by moving the ringbuffer map size check before the call to
unix.EpollCreate1.

Fixes: af50a38dd5c8 ("ringbuf: add new package to read from BPF ring buffers")

### DIFF
--- a/ringbuf/reader.go
+++ b/ringbuf/reader.go
@@ -121,20 +121,21 @@ func NewReader(ringbufMap *ebpf.Map) (r *Reader, err error) {
 		return nil, fmt.Errorf("invalid Map type: %s", ringbufMap.Type())
 	}
 
+	maxEntries := int(ringbufMap.MaxEntries())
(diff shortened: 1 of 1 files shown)


lmb:
cc @mythi

tklauser:
Not sure about the CI failures, but they look unrelated to this change from a quick glance: https://ebpf.semaphoreci.com/workflows/2f295e5a-9589-4e73-ab6c-899d133f0aa3?pipeline_id=b4373766-1516-4e72-beb4-bb120f9d6e45

mythi (approved):
@tklauser thanks!

looks like I had the `defer` path added too late. Agree it makes more sense to check the input parameters first...

PR #411, opened 2021-09-08
CI: turn off verbose tests

lmb:
We're at a stage where it's difficult to figure out what went wrong on CI,
since there are so many tests. Turn off verbose output for now.


tklauser (approved):

lmb:
I still don't know why test binaries are getting killed. I kind of suspect qemu, so maybe the Ubuntu 20.04 bump will help.

PR #412, opened 2021-09-08
Add an example for the RINGBUF map type

### LINKED ISSUE #396: Add BPF ringbuffer example to examples/
Following comment https://github.com/cilium/ebpf/pull/318#discussion_r699770055, we should add a ringbuf example once #318  is merged.

cc @folbricht

pippolo84:
Add a new example under `/examples` to show the usage of the newly introduced RINGBUF map type.

FIxes #396


mythi (commented):
I tested it and also tried to `go generate` from scratch and everything seems to work OK:

```
$ sudo ./ringbuffer
2021/09/09 10:36:41 Waiting for events..
2021/09/09 10:36:44 pid: 338112	comm: bash
2021/09/09 10:36:48 pid: 338114	comm: bash
2021/09/09 10:37:02 pid: 338116	comm: bash
2021/09/09 10:37:07 pid: 338117	comm: bash
2021/09/09 10:37:20 pid: 338141	comm: waybar
2021/09/09 10:37:20 pid: 338142	comm: sh
^C2021/09/09 10:37:27 Received signal, exiting..
```

lmb:
LGTM also. I'm still a bit concerned that a lot of the examples are quite "verbose". Something to tackle later maybe.

lmb (approved):

hao-lee:
Do we really need these intermediate files like `*ebfel.go` and `*.o` ? I still think we should provide a guide for users to generate them manually.

mythi:
> Do we really need these intermediate files like `*ebfel.go` and `*.o` ?

With this, the users don't need to worry about getting clang etc. installed and all they need is the Go compiler. This improves the user experience. More advanced users are always OK to re-generated if they like.

hao-lee:
> With this, the users don't need to worry about getting clang etc. installed and all they need is the Go compiler. This improves the user experience. More advanced users are always OK to re-generated if they like.

However, what we provide are just examples. They will eventually install clang-related packages if they want to develop their own tools.

lmb:
@hao-lee I understand that you might prefer not to commit files, however as long as it is your personal preference I'd ask you to trust our judgement and not raise this issue multiple times. Feel free to start a thread in our slack if there is a specific problem that these files cause you.

hao-lee:
@lmb Got it. Thanks.

examples/ringbuffer/main.go
```diff
@@ -0,0 +1,102 @@
...
+	log.Println("Waiting for events..")
+
+	for {
+		record, err := rd.Read()
+		if err != nil {
+			if errors.Is(err, ringbuf.ErrClosed) {
+				log.Println("Received signal, exiting..")
+				return
+			}
+			log.Printf("reading from reader: %s", err)
```
lmb, line 90: Sorry, coming back to this: why not abort the program here and in the tracepoint example? And why not `continue` after logging the error if we want to keep going?

pippolo84: Don't know exactly all the possible errors that may arise, I think it'll be fine if we just print the error and then let the user terminate the program, if needed.
I agree that we should add a `continue`, since it would be pointless to try to extract an event from the ring buffer in case of error. I'll fix this in both examples.

pippolo84: Actually, I see that the uretprobe example has the same issue, I'll fix that, too.

lmb (approved):

lmb:
Thanks!

PR #414, opened 2021-09-09
map: detect map Name and PinPath separately

hao-lee:
libbpf uses `/sys/fs/bpf` as the default pinpath if the user doesn't specify
one. We'd better keep consistent with it.

```
static int build_map_pin_path(struct bpf_map *map, const char *path)
{
...
	if (!path)
		path = "/sys/fs/bpf";
...
}
```

Signed-off-by: Hao Lee <haolee@didiglobal.com>

### DIFF
--- a/map.go
+++ b/map.go
@@ -226,8 +226,12 @@ func newMapWithOptions(spec *MapSpec, opts MapOptions, handles *handleCache) (_ 
 
 	switch spec.Pinning {
 	case PinByName:
-		if spec.Name == "" || opts.PinPath == "" {
-			return nil, fmt.Errorf("pin by name: missing Name or PinPath")
+		if spec.Name == "" {
+			return nil, fmt.Errorf("pin by name: missing Name")
+		}
+
+		if opts.PinPath == "" {
+			return nil, fmt.Errorf("pin by name: missing MapOptions.PinPath")
 		}
 
 		path := filepath.Join(opts.PinPath, spec.Name)


lmb:
Hi @hao-lee,

I think this is not a safe default: `/sys/fs/bpf` is shared across the system, so using it as a default path will lead to conflicts if multiple projects using this library want to pin maps with the same name. For that reason, it's best to always use a subdirectory of `/sys/fs/bpf`.

If we want to tackle this in the library we need to come up with a safer default. Maybe it's possible to derive a subdirectory using `os.Args[0]`? This needs some thinking through of corner cases, etc. For example, what happens when using `go run file.go` like you can do in the examples?

lmb:
> I think maybe it's the responsibility of the eBPF program developer to make sure the map's name is unique.

From my experience this unfortunately doesn't happen.

> I think this behavior is too mandatory, especially for some beginners who only want to fire up a runnable eBPF program as simple as possible. 

Will beginners start by pinning maps?

Maybe we can improve the error message?

hao-lee:
> Maybe we can improve the error message?

Agree. The error message confused me.

hao-lee:
@lmb Hello, I separate error detections. This should be more reasonable.

map.go
```diff
@@ -226,8 +226,12 @@ func newMapWithOptions(spec *MapSpec, opts MapOptions, handles *handleCache) (_
...
 	switch spec.Pinning {
 	case PinByName:
-		if spec.Name == "" || opts.PinPath == "" {
-			return nil, fmt.Errorf("pin by name: missing Name or PinPath")
+		if spec.Name == "" {
+			return nil, fmt.Errorf("pin by name: missing Name")
+		}
+
+		if opts.PinPath == "" {
+			return nil, fmt.Errorf("pin by name: missing PinPath")
```
lmb, line 234: Maybe `MapOptions.PinPath`?

lmb (approved):

PR #417, opened 2021-09-14
reproducible builds for examples

lmb:
Use the existing container to make build of examples reproducible. This requires adding a go toolchain, so unfortunately the image gets quite a bit bigger.


Makefile
```diff
@@ -50,21 +58,25 @@ clean:
 	-$(RM) testdata/*.elf
 	-$(RM) internal/btf/testdata/*.elf
 
-all: $(addsuffix -el.elf,$(TARGETS)) $(addsuffix -eb.elf,$(TARGETS))
+all: $(addsuffix -el.elf,$(TARGETS)) $(addsuffix -eb.elf,$(TARGETS)) generate
 	ln -srf testdata/loader-$(CLANG)-el.elf testdata/loader-el.elf
 	ln -srf testdata/loader-$(CLANG)-eb.elf testdata/loader-eb.elf
 
+generate:
+	go generate ./cmd/bpf2go
```
ti-mo, line 66: Could we put the `export` statements that were added above here instead? That would make it a bit more explicit and easier to track down where they're used.

ti-mo (commented):
This is starting to look a bit like `make` spaghetti, but not sure what the alternative would be.

Maybe instead of having to control `-cc` explicitly, pick up `BPF_CLANG` using `os.Getenv()` in bpf2go directly (and document it as part of its UI) to reduce the plumbing required?

Makefile
lmb: We can't make them part of the recipe, if that's what you mean (since the recipe runs in a shell already).

The best we could do is:

```
export FOO

generate:
```

Do you prefer that?

lmb:
> Maybe instead of having to control -cc explicitly, pick up BPF_CLANG using os.Getenv() in bpf2go directly (and document it as part of its UI) to reduce the plumbing required?

I made that choice deliberately, because it means you can easily grep BPF_CLANG and see all occurences. If you feel strongly we can change it, but then the examples become harder for users to adapt since they need to understand that env thing, and nobody reads documentation.

PR #420, opened 2021-09-16
Add sync.Pool to reduce allocations for the default case in unmarshalBytes()

rockdaboot:
This change avoids the allocations for bytes.Reader.

### DIFF
--- a/marshalers.go
+++ b/marshalers.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	"sync"
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal"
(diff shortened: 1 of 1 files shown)


lmb:
I think this is a net positive:

```
name                      old time/op    new time/op    delta
Marshalling/reflection-8    1.02µs ± 1%    1.00µs ± 1%   -1.90%  (p=0.029 n=4+4)
Marshalling/custom-8         756ns ± 0%     757ns ± 1%     ~     (p=0.486 n=4+4)
Marshalling/unsafe-8         712ns ± 2%     711ns ± 1%     ~     (p=1.000 n=4+4)

name                      old alloc/op   new alloc/op   delta
Marshalling/reflection-8     96.0B ± 0%     48.0B ± 0%  -50.00%  (p=0.029 n=4+4)
Marshalling/custom-8         24.0B ± 0%     24.0B ± 0%     ~     (all equal)
Marshalling/unsafe-8         0.00B          0.00B          ~     (all equal)

name                      old allocs/op  new allocs/op  delta
Marshalling/reflection-8      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.029 n=4+4)
Marshalling/custom-8          1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Marshalling/unsafe-8          0.00           0.00          ~     (all equal)
```

lmb (approved):

PR #424, opened 2021-09-16
Fix btf.FindType to avoid copy

### LINKED ISSUE #416: btf.Spec.FindType copies type
In 65df3d1d33f1196ce65d44d6171248274d38ed17 I tried to change the semantics of btf.Spec.FindType so that it doesn't return a copy of a type, but the type itself. Invocation of FindType looks like this:

```
var iphdr Struct
err = spec.FindType("iphdr", &iphdr)
```

Note that we "allocate" a full `iphdr Struct` and pass a pointer to that. `FindType` then copies the result into the pre-allocated `iphdr`, performing a shallow copy. To express the semantics I'd like a call would have to look something like this:

```
var iphdr *Struct
err = spec.FindType("iphdr", &iphdr)
```

It's not possible to express the type we'd like to accept in `FindType` except via `interface{}` and reflection.

pippolo84:
Fix `btf.FindType` to return a pointer to the matching BTF type.
This will allow to get exactly the same type (same address) for multiple calls requesting the same type name.

Fixes #416

### DIFF
--- a/internal/btf/btf_test.go
+++ b/internal/btf/btf_test.go
@@ -29,7 +29,7 @@ func TestParseVmlinux(t *testing.T) {
 		t.Fatal("Can't load BTF:", err)
 	}
 
-	var iphdr Struct
+	var iphdr *Struct
 	err = spec.FindType("iphdr", &iphdr)
 	if err != nil {
 		t.Fatalf("unable to find `iphdr` struct: %s", err)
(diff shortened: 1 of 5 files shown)


internal/btf/btf_test.go
```diff
@@ -12,6 +12,49 @@ import (
...
+	if err != nil {
+		t.Fatal("Can't load BTF:", err)
+	}
+
+	// spec.FindType MUST fail if typ is not a non-nil **T, where T satisfies btf.Type.
+	for _, typ := range []interface{}{
+		nil,
+		Struct{},
+		&Struct{},
+	} {
```
lmb, line 38: Add something like `**int`?

lmb (changes requested):

lmb:
Looks good, thanks for the quick turn around!

internal/btf/btf_test.go
```diff
@@ -12,6 +12,49 @@ import (
...
+		if err := spec.FindType("iphdr", nil); err == nil {
```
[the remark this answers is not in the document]
pippolo84: That's what happen when coding at late night :sweat_smile:

internal/btf/btf_test.go
pippolo84: Yes, and now that I think about it, we should also check something like slices and maps as "strange" types.

lmb (approved):
Thanks!

PR #425, opened 2021-09-17
automatically generate syscall wrappers

lmb:
Sorry for the big PR. I've tried to split things up into smaller pieces, review by commit is the best.

The idea: instead of manually keeping struct definitions up to date, use BTF to generate type definitions. This requires code that knows how to turn a btf.Type into a Go equivalent. I later plan to use the same code to allow generating defintions for custom types from bpf2go. Once we're confident that the generated types are flexible enough I'd like to move `sys` out of internal: it's something users have asked for, and since we auto generate code the maintenance burden is low.

all: remove syscall wrappers
    
    As much as possible, remove syscall wrappers scattered all over the code
    base.

sys: generate BPF types using vmlinux BTF

sys: simplify FD

sys: simplify BPFObjPin
    
    Move the BPF fs check into the single caller of BPFObjPin. This makes
    it easier to remove the function later on.

sys: move syscall code into new package

btf: export Copy

btf: add GoFormatter
    
    GoFormatter converts BTF types into Go types. It uses the following
    rules:
    
    * Only struct, union, array, integer, enum types
    * Data sections are represented as a struct
    * Data sections skip static variables
    * Explicit padding between fields if necessary


lmb:
The last commit in the series introduces `BPF(attr Attr)` where `Attr` is an interface. The result is that syscall arguments now escape and are therefore heap allocated:

```
name                      old time/op    new time/op    delta
Marshalling/reflection-8    1.02µs ± 0%    1.10µs ± 1%    +7.57%  (p=0.029 n=4+4)
Marshalling/custom-8         794ns ± 1%     826ns ± 0%    +4.13%  (p=0.029 n=4+4)
Marshalling/unsafe-8         730ns ± 2%     775ns ± 1%    +6.20%  (p=0.029 n=4+4)

name                      old alloc/op   new alloc/op   delta
Marshalling/reflection-8     48.0B ± 0%     80.0B ± 0%   +66.67%  (p=0.029 n=4+4)
Marshalling/custom-8         24.0B ± 0%     56.0B ± 0%  +133.33%  (p=0.029 n=4+4)
Marshalling/unsafe-8         0.00B         32.00B ± 0%     +Inf%  (p=0.029 n=4+4)

name                      old allocs/op  new allocs/op  delta
Marshalling/reflection-8      2.00 ± 0%      3.00 ± 0%   +50.00%  (p=0.029 n=4+4)
Marshalling/custom-8          1.00 ± 0%      2.00 ± 0%  +100.00%  (p=0.029 n=4+4)
Marshalling/unsafe-8          0.00           1.00 ± 0%     +Inf%  (p=0.029 n=4+4)
```

I'll have to revisit the code.

lmb:
I ended up generating functions that wrap the raw sycall.

ti-mo:
Please allow me some time to review. Would also like a walkthrough of this at some point.

lmb:
There's no rush.

PR #428, opened 2021-09-19
Fix missing close of fds from BPF{ProgLoad,MapCreate}

### LINKED ISSUE #373: Always close fd's returned from `internal.BPFProgLoad`
There are a few places where `internal.BPFProgLoad`'s return value is discarded, relying on the GC to close FDs using their finalizers. Relying on this is not ideal, since it's possible to run Go programs with GC disabled.

In `newProgramWithOptions()`, this is only done when re-running the prog load with verifier output enabled, but it's still theoretically possible for a retry to succeed and yield an fd. We should at least handle the fd if one happens to be returned.

In `features/` this is also done in a few places, and it should definitely be handled correctly there. (@rgo3)

pippolo84:
Since it is possible to disable the GC, we cannot rely solely on the finalizers to close the file descriptors returned by BPFProgLoad or BPFMapCreate.
So, we always close them explicitly, if needed.

Fixes #373

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -307,7 +307,10 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		attr.LogSize = uint32(len(logBuf))
 		attr.LogBuf = internal.NewSlicePointer(logBuf)
 
-		_, logErr = internal.BPFProgLoad(attr)
+		fd, logErr = internal.BPFProgLoad(attr)
+		if logErr == nil {
+			fd.Close()
+		}
 	}
 
 	if errors.Is(logErr, unix.EPERM) && logBuf[0] == 0 {
--- a/features/map.go
+++ b/features/map.go
@@ -139,7 +139,7 @@ func haveMapType(mt ebpf.MapType) error {
 		return err
 	}
 
-	_, err = internal.BPFMapCreate(createMapTypeAttr(mt))
+	fd, err := internal.BPFMapCreate(createMapTypeAttr(mt))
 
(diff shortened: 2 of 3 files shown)


prog.go
```diff
@@ -307,7 +307,10 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		attr.LogSize = uint32(len(logBuf))
 		attr.LogBuf = internal.NewSlicePointer(logBuf)
 
-		_, logErr = internal.BPFProgLoad(attr)
```
lmb, line 310: I think this case is impossible, since we already know that executing the syscall with these arguments must fail. Maybe add a comment to that effect instead of adding the if?

pippolo84: In the related issue @ti-mo suggested that a valid fd may be theoretically returned.
Looking at the possible return values from `bpf` syscall, I can think of a transient error condition only if the first call returns `ENOMEM` and the next one succeeds thanks to the OOM freeing memory. Is this possible?
Anyway, even if it were, things would probably be really bad, so I am ok to leave just a comment, if I am not missing something.

lmb: Looking at it another way: if this case really could happen we would return `nil, nil` from this function which is also wrong.

lmb: In the freak situation where the second call returns `fd, nil` this code will end up closing the fd and then returning `nil, nil` again?

lmb: Sorry, I misread the code.

lmb (approved):

PR #430, opened 2021-09-21
tree-wide: drop deprecated io/ioutil

mythi:
Go 1.16 announced [1] the deprecation of io/ioutil with (almost)
drop in replacements in io and os packages. This change moves to
use those replacements.

Note: io/ioutil going away was NOT announced but dropping it is
trivial and allows to reduce imports.

[1] https://golang.org/doc/go1.16#ioutil

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

### DIFF
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"go/token"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
@@ -289,7 +288,7 @@ func (b2g *bpf2go) convert(tgt target, arches []string) (err error) {
 	}
 
 	depFileName := goFileName + ".d"
-	if err := ioutil.WriteFile(depFileName, depFile, 0666); err != nil {
+	if err := os.WriteFile(depFileName, depFile, 0666); err != nil {
 		return fmt.Errorf("can't write dependency file: %s", err)
 	}
 
--- a/asm/instruction_test.go
+++ b/asm/instruction_test.go
@@ -5,7 +5,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"math"
 	"testing"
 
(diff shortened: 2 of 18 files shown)


cmd/bpf2go/main.go
```diff
@@ -289,7 +288,7 @@ func (b2g *bpf2go) convert(tgt target, arches []string) (err error) {
 	}
 
 	depFileName := goFileName + ".d"
-	if err := ioutil.WriteFile(depFileName, depFile, 0666); err != nil {
+	if err := os.WriteFile(depFileName, depFile, 0600); err != nil {
```
lmb, line 291: Changing the mode here isn't correct, IMO. You lose the ability to specify what resulting mode you want using `umask`. It'd also be better if this was a separate commit, since it changes behaviour vs. a "mechanical" rename.

lmb (changes requested):

cmd/bpf2go/main.go
mythi: `0600` should not be here. it got here by mistake, will drop it.

ti-mo:
Thanks!

kolyshkin:
@ti-mo usually they mark something as deprecated before dropping it, so users have another release cycle to fix this. By "mark as deprecated" I mean a `Deprecated:` annotation that is recognized by a linter, that issues a warning if this ID is used.

Currently, ioutil is not marked as deprecated, and it looks like it won't be marked in Go 1.18 (judging by the current golang/go git repo state).

Also note that ioutil doc says
```
// As of Go 1.16, the same functionality is now provided
// by package io or package os, and those implementations
// should be preferred in new code.
```

but it does not say that the existing code should immediately be migrated.

In my particular case, we have to drop Go 1.15 compatibility from opencontainers/runc to accommodate cilium/ebpf 0.7.0. While we are well aware that Go 1.15 is unsupported, we try to be a bit more conservative when dropping compatibility with older releases, as many projects import some packages from runc, and some might be a tad late updating their dependencies (including Go version). Giving them some slack is good.

All this does not matter now as we've decided to drop Go 1.15 support anyway, but in general I think it would be nice to be slightly more conservative.

> based on a0a3347 I did not realize there was any issue.

Ah, I was not aware of PR #382 -- in view of this, it's totally OK to drop io/ioutil use. So what I said above should have been addressed to that PR, not here.

To reiterate -- everything is fine, no reverts needed, the above is just an explanation for my earlier comment.

PR #433, opened 2021-09-22
RFC: enable gosec linter

mythi:


elf_reader.go
```diff
@@ -592,7 +592,7 @@ func (ec *elfCode) loadBTFMaps(maps map[string]*MapSpec) error {
 				return fmt.Errorf("expected struct, got %s", v.Type)
 			}
 
-			mapSpec, err := mapSpecFromBTF(sec, &vs, mapStruct, ec.btf, name, false)
+			mapSpec, err := mapSpecFromBTF(sec, &ds.Vars[idx], mapStruct, ec.btf, name, false)
```
ti-mo, line 595: The memory aliasing warning is pretty convenient while coding IMO, but that's pretty much it. A bit too many exclusions for my liking.

mythi: > The memory aliasing warning is pretty convenient while coding IMO

All contributors may not run linters with checks like this so having CI error for them might be useful.

asm/instruction.go
```diff
@@ -1,7 +1,7 @@
 package asm
 
 import (
-	"crypto/sha1"
+	"crypto/sha1" // #nosec G505: SHA1 used by the kernel so we must follow
```
lmb, line 4: Why not exclude this rule then?

.golangci.yaml
```diff
@@ -26,4 +27,12 @@ linters:
     # - gocyclo
     # - prealloc
     # - maligned
-    # - gosec
+
+linters-settings:
+  gosec:
```
lmb, line 32: Here is the list of current gosec rules that seem interesting based on the short description:

```
    G103: Audit the use of unsafe block
    G104: Audit errors not checked
    G109: Potential Integer overflow made by strconv.Atoi result conversion to int16/32
    G110: Potential DoS vulnerability via decompression bomb
    G204: Audit use of command execution
    G301: Poor file permissions used when creating a directory
    G302: Poor file permissions used with chmod
    G303: Creating tempfile using a predictable path
    G304: File path provided as taint input
    G305: File traversal when extracting zip/tar archive
    G306: Poor file permissions used when writing to a new file
```

* Is there a way to read up on those rules? I can't find explanations on the website or the repo.
* Why is `G204` disabled? Seems like the most useful one?
* Why is `G601: Implicit memory aliasing of items from a range statement` a security bug? Why is it enabled?
* Why is `0660` better than `0666`? Assuming that umask is sensible.

lmb (commented):
Similar to the prealloc discussion, please give your reasons why you'd like to enable the gosec linter.

* Can you choose a list of rules to enable instead of which ones to disable? More than half of the rules don't really apply.
* Can you use excludes in .golangci.yaml instead of `#nosec` in the files? Then removing a linter that doesn't pull it's weight becomes easier.

PR #435, opened 2021-09-23
[internal/FeatureTest] avoid testing more than once the feature

nplanel:
Fix the issue when a featureTest() failed, it would re-evalutated each time


nplanel:
Sorry I miss readed the fallthrough here

PR #436, opened 2021-09-23
[ebpf/map] add map.Iterate() support for old kernel (< 4.4.132)

nplanel:
Safe map.Iterate() to use with old kernel by guessing the first key (randomly)

The idea is the following on old kernel (< 4.4.132) :
* we know that if 2 random key return the same values then it mean we found the 1st element of the map to iterate
* if one of the key match an element in the map, we try a 3rd random key to decide if we found the 1st element
* we fallback an error if all 3 random keys doesn't match
* the cost is 2/3 more map.lookup() calls on the first iteration

### DIFF
--- a/map.go
+++ b/map.go
@@ -5,9 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"path/filepath"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/btf"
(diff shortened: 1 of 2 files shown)


lmb:
Hi Nicholas, sorry that the first review took so long. Last week was busy with Linux Plumbers conference.

We don't support kernels < 4.9 at the moment, so this code will never be exercised on CI. So to be able to merge this we'd have to change this policy. We don't have to support kernels that old, so I don't want to shoulder that burden, tbh. Is that something DataDog is willing to invest time in? If yes, you would need to start by adding 4.4 kernels to CI and making sure tests are all working. Then we could tackle this PR.

map.go
```diff
@@ -160,6 +184,9 @@ type Map struct {
 	pinnedPath string
 	// Per CPU maps return values larger than the size in the spec
 	fullValueSize int
+	// randKey used for kernel < 4.4.132 that didn't support nil key
+	randKey []byte
```
lmb, line 188: Why are these part of `Map`?

map.go
```diff
@@ -1180,6 +1224,29 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 		return false
 	}
 
+	// guessing a random start key
+	if err := haveMapIterateFromNullKey(); mi.prevKey == nil && errors.Is(err, ErrNotSupported) {
```
lmb, line 1228: Can't you achieve the same by checking EFAULT and mi.prevKey == nil below?

map.go
nplanel: It was to keep the random context per map, but it's optional

map.go
nplanel: I moved the test logic in the error of the first NextKeyBytes() that will always failed on old kernel.
Let me know if you want/prefer to move local helper function to the MapIterator local members

nplanel:
> Hi Nicholas, sorry that the first review took so long. Last week was busy with Linux Plumbers conference.
> 
> We don't support kernels < 4.9 at the moment, so this code will never be exercised on CI. So to be able to merge this we'd have to change this policy. We don't have to support kernels that old, so I don't want to shoulder that burden, tbh. Is that something DataDog is willing to invest time in? If yes, you would need to start by adding 4.4 kernels to CI and making sure tests are all working. Then we could tackle this PR.

No problem for the delay Linux Plumbers was nice.
I will add support to the CI for the 4.4 kernels and skip() tests that can't work with

PR #437, opened 2021-09-23
avoid wrap exported error with no additional value

nplanel:


lmb:
Hi, wrapping these errors is on purpose: it forces callers to use `errors.Is` instead of a straight comparison. What is the problem you're trying to solve?

nplanel:
> Hi, wrapping these errors is on purpose: it forces callers to use `errors.Is` instead of a straight comparison. What is the problem you're trying to solve?

It was more for readability, but I see, you don't want the caller to do a direct comparaison other than against nil.

PR #438, opened 2021-09-24
make RemoveMemlockRlimit conditional on kernel version

lmb:
We only need to raise RLIMIT_MEMLOCK on old kernels. Check whether
the kernel uses memcg accounting by lowering the limit to zero and
creating a small map.

### DIFF
--- a/syscalls.go
+++ b/syscalls.go
@@ -17,16 +17,6 @@ import (
 // Deprecated: use os.ErrNotExist instead.
 var ErrNotExist = os.ErrNotExist
 
-// RemoveMemlockRlimit removes the limit on the amount of memory the current
-// process can lock into RAM. Returns a function that restores the limit to
-// its previous value.
-//
-// This is not required to load eBPF resources on kernel versions 5.11+
-// due to the introduction of cgroup-based memory accounting.
-func RemoveMemlockRlimit() (func() error, error) {
-	return unix.RemoveMemlockRlimit()
-}
-
 // invalidBPFObjNameChar returns true if char may not appear in
 // a BPF object name.
 func invalidBPFObjNameChar(char rune) bool {
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -20,6 +20,7 @@ blocks:
       prologue:
         commands:
           - checkout
+          - curl -sSfL http://security.ubuntu.com/ubuntu/pool/main/c/ca-certificates/ca-certificates_20210119~20.04.2_all.deb -o /tmp/ca-certificates.deb && sudo dpkg -i /tmp/ca-certificates.deb
           - sudo mkdir -p /usr/local/golang/1.17 && curl -fL "https://golang.org/dl/go1.17.linux-amd64.tar.gz" | sudo tar -xz -C /usr/local/golang/1.17
           - sem-version go 1.17
           - go install github.com/mattn/goveralls@latest
--- a/internal/testutils/rlimit.go
+++ b/internal/testutils/rlimit.go
@@ -4,13 +4,13 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/cilium/ebpf/internal/unix"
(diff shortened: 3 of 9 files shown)


florianl (approved):

syscalls.go
```diff
@@ -18,13 +18,22 @@ import (
...
-// due to the introduction of cgroup-based memory accounting.
-func RemoveMemlockRlimit() (func() error, error) {
-	return unix.RemoveMemlockRlimit()
+// due to the introduction of cgroup-based memory accounting. On such kernels
+// the function is a no-op.
+//
+// Since the function changes global per-process limits it should be invoked at
+// program start up, in main() or init().
+//
+// This function exists as a convenience and doesn't cover situations where
```
ti-mo, line 30: Use of a double negative here makes it hard to understand.

`... as convenience and should only be used when permanently raising RLIMIT_MEMLOCK to infinite is appropriate.`

When is it not appropriate? Should `prlimit(2)` be used in case it isn't?

internal/rlimit.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func RemoveMemlockRlimit() error {
+	var oldLimit unix.Rlimit
+	if err := unix.Prlimit(0, unix.RLIMIT_MEMLOCK, nil, &oldLimit); err != nil {
+		return fmt.Errorf("retrieve RLIMIT_MEMLOCK: %s", err)
+	}
+
+	// Reduce the limit to zero. This is always allowed.
+	newLimit := unix.Rlimit{Cur: 0, Max: oldLimit.Max}
```
ti-mo, line 17: I'm a bit worried about a bad interaction with package `features`. This might cause false negatives on feature probes, as those might be run during package `init()` as well, though not necessarily in parallel.

If we're keeping the 'zero' behaviour around (see comments below), calling this `zeroLimit` is likely more appropriate.

internal/rlimit.go
```diff
@@ -0,0 +1,54 @@
...
+	if err := unix.Prlimit(0, unix.RLIMIT_MEMLOCK, &newLimit, &oldLimit); err != nil {
+		return fmt.Errorf("lower RLIMIT_MEMLOCK: %s", err)
+	}
+
+	attr := BPFMapCreateAttr{
+		MapType:    2, /* Array */
+		KeySize:    4,
+		ValueSize:  4,
```
ti-mo, line 25: - What about using a more targetted value like `oldLimit.Max` here instead of dropping the limit to zero? Creating a map that's highly likely to be over the current limit should be sufficient and less dangerous. If we go down this path, `getrlimit(2)` should be sufficient, no need to mutate system state.
- What about exposing this as a feature probe instead?

ti-mo (changes requested):

internal/rlimit.go
lmb: I think the answer is: `features` should not run things on init. Where are we doing that?

internal/rlimit.go
lmb: Feature probe doesn't work since we need to change program global state.

Interesting idea re making a bigger map. The problem is that we might then run into memcg limits. When mcmg limit < memlock limit we would incorrectly deduce that the kernel uses memlimit accounting.

ti-mo: Would there be a more reliable way to determine if memcg accounting is enabled? We could make a similarly wrong deduction if the memcg limit is set low enough.

internal/rlimit.go
ti-mo: I mean a caller using feature probe output in an `init()` function anywhere in a program.

internal/rlimit.go
lmb: That's what I asked on the initial PR, and the answer was no unfortunately. If it turns out to be a problem we can remove the check and always bump the rlimit. This way round seems better for a start though.

ti-mo (approved):

PR #443, opened 2021-10-01
WIP : [CI] add kernel 4.4.131 support

nplanel:
I push this here for a first pass review

I disabled (ErrNotSupported) map.Pin() and map.UnPin() when the kernel can't read map info (keysize, ...) back from /proc/pid/fdinfo
Technically for these old kernel is possible to Pin (If I'm correct) but the caller need to give the map Spec, if it's need to be supported we need to have a persistent way to save map spec, may via shmget() ?

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -61,7 +61,7 @@ blocks:
       - name: Run unit tests
         matrix:
           - env_var: KERNEL_VERSION
-            values: ["5.10", "5.4", "4.19", "4.9"]
+            values: ["5.10", "5.4", "4.19", "4.9", "4.4.131"]
         commands:
           - timeout -s KILL 600s ./run-tests.sh $KERNEL_VERSION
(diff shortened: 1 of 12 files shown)


ti-mo:
Hi @nplanel, thank you for the PR, there is some really valuable stuff in here! Sorry it took a while to respond.

I'll work on an initial review on this PR, but I'd like for things to be split out into multiple PRs given the size of the patch, in order of dependency:
- feature probes
- non-atomic pin behaviour
- NULL-key map lookups: there's some overlap with cilium/cilium here that I want to validate and eventually land into the library -- be aware there might be extended discussion around this topic involving other Cilium contributors
- enabling CI for 4.4: IMO this can wait until the above is merged, but I understand it's included in this PR here to actually validate things

This way, the discussions will remain more focused and things can be merged and reviewed in parallel. We'll cut 0.7.1 as soon as all of this has landed, so you can do a clean import. 0.7.0 is set to release early next week. Thanks in advance for your understanding!

PR #448, opened 2021-10-12
Add bpf2go example using BPF CO-RE

markpash:
```
This example uses bpf2go to produce a portable Kprobe with CO-RE that
attaches to `__neigh_event_send` and reads out some information about
neighbour events over a perf event to userspace, where it's logged to
stdout.
```

I've included a vmlinux.h, but I think it might be a little bulky to carry in our tree, I'd appreciate some advice on what to do there.


lmb:

PR #471, opened 2021-10-29
features: fix HaveProgType(ebpf.Syscall)

### LINKED ISSUE #464: features: fix HaveProgType(ebpf.Syscall)
ebp.Syscall is a BPF program type introduced with kernel version 5.14, and the HaveProgType API probe for it fails on the respective kernel version. 

We will have to make sure that the `ProgLoadAttr` structure is built with the correct values before executing the probe or that we are interpreting the returned error correctly. In either case going through the BPF syscall code [here](https://github.com/torvalds/linux/blob/3906fe9bb7f1a2c8667ae54e967dc8690824f4ea/kernel/bpf/syscall.c#L4587) should help finding a solution.

rgo3:
BPF Syscall programs need to be sleepable which is enforced by the
verifier. So we need to set the correct program flags before executing
the feature probe.

Fixes #464.

Signed-off-by: Robin Gögge <r.goegge@gmail.com>

### DIFF
--- a/features/prog.go
+++ b/features/prog.go
@@ -29,6 +29,7 @@ type progCache struct {
 
 func createProgLoadAttr(pt ebpf.ProgramType) (*sys.ProgLoadAttr, error) {
 	var expectedAttachType ebpf.AttachType
+	var progFlags uint32
 
 	insns := asm.Instructions{
 		asm.LoadImm(asm.R0, 0, asm.DWord),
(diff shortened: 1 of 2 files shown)


florianl (approved):
works fine here :+1: 
sorry for adding the syscall prog type without thinking about the features tests :pray:

rgo3:
No worries, when we added those tests we hoped they would just automatically fail when new prog types are added, forcing people to ensure the feature probe for that type also works. But if no CI kernel for that feature exists yet its skipped and one can get away with not making sure the probes actually work. I think that's what happened here, even though you actually had thought about the feature tests and added the right kernel version to the test matrix.

PR #473, opened 2021-11-03
Allows user space to update/lookup map elements under bpf_spin_lock

mehrdadrad:
- Added FLock constant (BPF_F_LOCK)
- Added three new map methods to support lookup with FLock: 
    - LookupWithFlags
    - LookupBytesWithFlags
    - LookupAndDeleteWithFlags
- Added tests with test data.


lmb (commented):
This adds a lot of API for a pretty niche use case. Could you instead detect whether a map contains a spinlock at creation time and infer the need for BPF_F_LOCK? This will be somewhat complicated by `LoadPinnedMap` which should also work with spinlock maps. 

Alternatively we should investigate a single new method `LookupOpts(key, valueOut, opts)`. Opts could be:

```
struct Opts {
    Delete bool
    Lock bool
}
```

That would avoid so much new API as well.

mehrdadrad:
@lmb re: detect whether a map contains a spinlock 
I was thinking about it but why BPF_F_LOCK is an option which it [MUST pass from user space through BPF command](https://github.com/torvalds/linux/blob/master/include/uapi/linux/bpf.h#L129-L132)? why doesn't the kernel handle it by itself?
I just tried bpftool and looks we can detect a map with bpf_spin_lock through BTF:

```
[9] STRUCT 'hash_elem' size=8 vlen=2
        'cnt' type_id=2 bits_offset=0
        'lock' type_id=10 bits_offset=32
[10] STRUCT 'bpf_spin_lock' size=4 vlen=1
        'val' type_id=6 bits_offset=0
```
   
In meanwhile the `libbpf` added more APIs as blow if we follow them:
- int bpf_map_lookup_and_delete_elem_flags(int fd, const void *key, void *value, __u64 flags)
- int bpf_map_lookup_elem_flags(int fd, const void *key, void *value, __u64 flags)

Your alternative option can be like below:
LookupOpts should be support:
- BPF_MAP_LOOKUP_AND_DELETE_ELEM
- BPF_MAP_LOOKUP_ELEM


```
LookupOptions struct { 
   Flags MapUpdateFlags 
   Delete bool
}
LookupOpts(key, valueOut interface{}, opts LookupOptions) error
```
The last thing, we can add at least two of APIs which I already did same as libbpf.

Any thoughts?

lmb:
You're right, since it's possible to do a lookup without BPF_F_LOCK always adding it doesn't make sense. Let's scrap the BTF idea.

The best thing we can do (imo):
* LookupFlags
* LookupAndDeleteFlags

There should be a new type `LookupFlags uint64` though, with a constant `LookupLock`, in addition to `UpdateLock`.

map.go
```diff
@@ -497,7 +497,26 @@ func (m *Map) Info() (*MapInfo, error) {
...
+//
+// Passing FLock flag will look up the value of a spin-locked
+// map without returning the lock. This must be specified if the
+// elements contain a spinlock.
+//
+// Calls Close() on valueOut if it is of type **Map or **Program,
+// and *valueOut is not nil.
+//
+// Returns an error if the key doesn't exist, see ErrKeyNotExist.
+func (m *Map) LookupWithFlags(key, valueOut interface{}, flags MapUpdateFlags) error {
```
lmb, line 517:
```suggestion
func (m *Map) LookupWithFlags(key, valueOut interface{}, flags MapLookupFlags) error {
```

etc. for the other function.

map.go
```diff
@@ -508,42 +527,55 @@ func (m *Map) Lookup(key, valueOut interface{}) error {
...
-// LookupBytes gets a value from Map.
+// LookupBytesWithFlags gets a value from Map.
+//
+// Passing FLock flag will look up the value of a spin-locked
+// map without returning the lock. This must be specified if the
+// elements contain a spinlock.
 //
 // Returns a nil value if a key doesn't exist.
-func (m *Map) LookupBytes(key interface{}) ([]byte, error) {
+func (m *Map) LookupBytesWithFlags(key interface{}, flags MapUpdateFlags) ([]byte, error) {
```
lmb, line 566: Skip this, it's possible to do this via `LookupWithFlags`

map.go
```diff
@@ -573,6 +628,8 @@ const (
 	UpdateNoExist MapUpdateFlags = 1 << (iota - 1)
 	// UpdateExist updates an existing element.
 	UpdateExist
+	// Allows user space to update/lookup map elements under bpf_spin_lock.
+	FLock
```
lmb, line 632:
```suggestion
	UpdateLock
```

map.go
mehrdadrad: Addressed in c379cbfad81f51a1bc3eda17cbda2f674abe1c92

map.go
mehrdadrad: Addressed in 623ff45f69ae33eecd66282db41702eafba6bf66

map.go
mehrdadrad: Addressed in 623ff45f69ae33eecd66282db41702eafba6bf66

testdata/map_spin_lock.c
```diff
@@ -0,0 +1,23 @@
+/* This file excercises bpf_spin_lock. */
```
lmb, line 1: You need to add this file to TARGETS in Makefile and rebuild the testdata via `make` using the docker build image.

lmb:
Can you squash your changes please?

testdata/map_spin_lock.c
mehrdadrad: I just did but no idea why CI shows: generated files are not up to date

mehrdadrad: Fixed!

lmb (approved):

lmb:
Thanks!

PR #475, opened 2021-11-04
epoll: factor out poller

lmb:
The ringbuf package uses a copy of the perf epoll code. Move the code into
a shared package.

### DIFF
--- a/internal/epoll/poller.go
+++ b/internal/epoll/poller.go
@@ -1,0 +1,207 @@
+package epoll
+
+import (
+	"fmt"
+	"math"
+	"os"
+	"runtime"
+	"sync"
+
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+// Poller waits for readiness notifications from multiple file descriptors.
+//
+// The wait can be interrupted by calling Close.
+type Poller struct {
+	// mutexes protect the fields declared below them. If you need to
+	// acquire both at once you must lock epollMu before eventMu.
+	epollMu sync.Mutex
+	epollFd int
+
+	eventMu sync.Mutex
+	event   *eventFd
+}
+
+func New() (*Poller, error) {
+	epollFd, err := unix.EpollCreate1(unix.EPOLL_CLOEXEC)
+	if err != nil {
+		return nil, fmt.Errorf("create epoll fd: %v", err)
+	}
+
+	p := &Poller{epollFd: epollFd}
+	p.event, err = newEventFd()
+	if err != nil {
+		unix.Close(epollFd)
+		return nil, err
+	}
+
+	if err := p.Add(p.event.raw, 0); err != nil {
+		unix.Close(epollFd)
+		p.event.close()
+		return nil, fmt.Errorf("add eventfd: %w", err)
+	}
+
+	runtime.SetFinalizer(p, (*Poller).Close)
+	return p, nil
+}
+
+// Close the poller.
+//
+// Interrupts any calls to Wait. Multiple calls to Close are valid, but subsequent
+// calls will return os.ErrClosed.
+func (p *Poller) Close() error {
+	runtime.SetFinalizer(p, nil)
+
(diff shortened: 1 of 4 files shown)


perf/reader.go
```diff
@@ -222,34 +199,23 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
...
 	}
 
 	pr = &Reader{
-		array:   array,
-		rings:   rings,
-		epollFd: epollFd,
+		array:  array,
+		rings:  rings,
+		poller: poller,
 		// Allocate extra event for closeFd
```
mythi, line 216: This should probably say "extra event for event fd". I'm thinking would it be better to hide this extra "+1" `EpollEvent` from `Poller` users? In the ringbuf's `ePollEvents` the extra "+1" is not added but AFAICS it's currently needed.

internal/epoll/poller.go
```diff
@@ -0,0 +1,207 @@
...
+func (p *Poller) Wait(events []unix.EpollEvent) (int, error) {
+	p.epollMu.Lock()
+	defer p.epollMu.Unlock()
+
+	if p.epollFd == -1 {
+		return 0, fmt.Errorf("epoll wait: %w", os.ErrClosed)
+	}
+
+	for {
+		n, err := unix.EpollWait(p.epollFd, events, -1)
```
mythi, line 131: in my `ringbuf` PR  (https://github.com/cilium/ebpf/pull/318#discussion_r690745230) we discussed the possibility to set poll timeout value and do busyloop reading for `BPF_RB_NO_WAKEUP` ringbuf events.

`BPF_RB_NO_WAKEUP` handling should be improved but I wonder what would be the best way to do that: keep `Reader.Close()` logic via `Poller` and have 0 epolltimeout or skip `Poller` completely when setting up the reader to busyloop. What do you think?

mythi (commented):
Thanks @lmb 

I gave it try here and things work OK, I added a couple small nits and questions.

perf/reader.go
lmb: Yeah, good point. FWIW I think the extra event is not needed:
* We use the event fd to unblock Read(), so that Close can acquire the mutex
* Without the `+1` event we're not guaranteed to receive the close event, but we are guaranteed to be unblocked if any event is pending

internal/epoll/poller.go
lmb: `BPF_RB_NO_WAKEUP` seems orthogonal to busy looping to me, am I misunderstanding things?
* BPF_RB_NO_WAKEUP: control when epoll wakes. already works with the current code.
* busy looping: don't enter the poller in the first place.

Unverified assumption: busy looping should go via exposing an API to read from the ring directly?

perf/reader.go
mythi: Agree, makes sense.

internal/epoll/poller.go
mythi: > Unverified assumption: busy looping should go via exposing an API to read from the ring directly?

This is probably the best approach. I'll draft something for it. FWIW, `libbpf` seems to have `ring_buffer__consume()` for the same.

internal/epoll/poller.go
```diff
@@ -0,0 +1,207 @@
...
+	defer p.epollMu.Unlock()
+
+	if p.epollFd == -1 {
+		return fmt.Errorf("epoll add: %w", os.ErrClosed)
+	}
+
+	// The representation of EpollEvent isn't entirely accurate.
+	// Pad is fully useable, not just padding. Hence we stuff the
+	// CPU in there, which allows us to use a slice to access
+	// the correct perf ring.
```
mythi, line 104:
```suggestion
	// Pad is fully useable, not just padding. Hence we stuff the
	// id in there, which allows us to identify the event later (e.g.,
	// in case of perf events, which CPU sent it).
```

mythi (approved):
Looks nice, I don't have anything to add

Issue #479, opened 2021-11-09
Improve Makefile usability with `podman`


ti-mo:
Currently, in order to rebuild testdata using `podman`, make needs to be invoked as follows:
```
CONTAINER_ENGINE=podman CONTAINER_RUN_ARGS= make
```

This is a bit verbose and hard to remember. Perhaps the Makefile could run `command -v podman` and automatically use it when present on the machine?

Also, running rootless containers requires subuids and subgids to be allocated for the current user (see https://github.com/containers/podman/issues/3421#issuecomment-544455837).

```
~ cat /etc/subuid /etc/subgid
timo:100000:65536
timo:100000:65536
```

Since I didn't set it up before running podman for the first time, an extra `podman system migrate` was required to get containers to start properly.

Ideally, this would be documented as well.

syscll:
I can take a stab at this if you like? Should be fairly straightforward.

My only thoughts are what if a user has `podman` installed, but _still_ wants to use `docker`?

PR #480, opened 2021-11-09
Makefile: prefer podman for container engine

syscll:
Ref: https://github.com/cilium/ebpf/issues/479

Usage:
```
# If podman is found on the system, then it is preferred. Otherwise fall back to using docker.
# Passing the variable at runtime will always take precedence:
$ command -v podman
/bin/podman
$ CONTAINER_ENGINE=docker make container-all
docker run --rm --user "1000:985" \
...
$ make container-all
podman run --rm  \
...
# Override container run args
$ CONTAINER_RUN_ARGS='--user "1000:985"' make container-all
podman run --rm --user "1000:985" \
...
# Override both
$ CONTAINER_ENGINE=docker CONTAINER_RUN_ARGS='--name=cilium-ebpf' make container-all
docker run --rm --name=cilium-ebpf \
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ UIDGID := $(shell stat -c '%u:%g' ${REPODIR})
 # Prefer podman if installed, otherwise use docker.
 # Note: Setting the var at runtime will always override.
 CONTAINER_ENGINE ?= $(if $(shell command -v podman), podman, docker)
-CONTAINER_RUN_ARGS ?= --user "${UIDGID}"
+CONTAINER_RUN_ARGS ?= $(if $(filter ${CONTAINER_ENGINE}, podman),, --user "${UIDGID}")
 
 IMAGE := $(shell cat ${REPODIR}/testdata/docker/IMAGE)
 VERSION := $(shell cat ${REPODIR}/testdata/docker/VERSION)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,7 +9,9 @@ CFLAGS := -O2 -g -Wall -Werror $(CFLAGS)
 REPODIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 UIDGID := $(shell stat -c '%u:%g' ${REPODIR})
 
-CONTAINER_ENGINE ?= docker
+# Prefer podman if installed, otherwise use docker.
+# Note: Setting the var at runtime will always override.
+CONTAINER_ENGINE ?= $(if $(shell command -v podman), podman, docker)
 CONTAINER_RUN_ARGS ?= --user "${UIDGID}"
```
lmb, line 15: This means `podman` will use `--user` though, which is (somehow?) detrimental.

mmat11: correct, `CONTAINER_RUN_ARGS` must be empty when ran with `podman`

syscll: Noted. Will update!

syscll: Fixed.

mmat11: works for me

lmb (approved):

PR #481, opened 2021-11-10
cmd/bpf2go: strip DWARF info

### LINKED ISSUE #429: bpf2go: strip dwarf debug symbols from ELF
From https://nakryiko.com/posts/libbpf-bootstrap/#makefile we can strip dwarf symbols from the compiled .o via `llvm-strip -g` resulting in a smaller ELF.

- Would we need a command line option to override `llvm-strip`?
- How much of a difference does stripping make?

lmb:
The compiled ELF contain DWARF debug info, which isn't used by the
BPF toolchain. Strip the debug info by default. Passing -strip=""
disables stripping.

Fixes #429


cmd/bpf2go/main.go
```diff
@@ -70,6 +70,7 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
 
 	fs := flag.NewFlagSet("bpf2go", flag.ContinueOnError)
 	fs.StringVar(&b2g.cc, "cc", "clang", "`binary` used to compile C to BPF")
+	fs.StringVar(&b2g.strip, "strip", "llvm-strip", "`binary` used to strip DWARF from compiled BPF (pass empty string to disable)")
```
brycekahle, line 73: IMHO the empty string to disable is a bit strange and could be difficult to get right. I would prefer a separate `-no-strip` flag.

lmb: Good call, I added `no-strip`.

ti-mo (approved):
Thanks, looks good! Also looked into a way to get clang to do this automatically, but doesn't seem to be possible since we're not invoking the linker.

PR #484, opened 2021-11-12
Link: add XDP support

mehrdadrad:
This new API: AttachXDP links an XDP program to an interface through BPF_LINK_CREATE.

### DIFF
--- a/link/xdp.go
+++ b/link/xdp.go
@@ -1,0 +1,54 @@
+package link
+
+import (
+	"fmt"
+
+	"github.com/cilium/ebpf"
+)
+
+// XDPAttachFlags represents how XDP program will be attached to interface.
+type XDPAttachFlags uint32
+
+const (
+	// XDPGenericMode (SKB) links XDP BPF program for drivers which do
+	// not yet support native XDP.
+	XDPGenericMode XDPAttachFlags = 1 << (iota + 1)
+	// XDPDriverMode links XDP BPF program into the driver’s receive path.
+	XDPDriverMode
+	// XDPOffloadMode offloads the entire XDP BPF program into hardware.
+	XDPOffloadMode
+)
+
+type XDPOptions struct {
+	// Program must be an XDP BPF program.
+	Program *ebpf.Program
+
+	// Interface is the interface index to attach program to.
+	Interface int
+
+	// Flags is one of XDPAttachFlags (optional).
+	//
+	// Only one XDP mode should be set, without flag defaults
+	// to driver/generic mode (best effort).
+	Flags XDPAttachFlags
+}
+
+// AttachXDP links an XDP BPF program to an XDP hook.
+func AttachXDP(opts XDPOptions) (Link, error) {
+	if t := opts.Program.Type(); t != ebpf.XDP {
+		return nil, fmt.Errorf("invalid program type %s, expected XDP", t)
+	}
+
+	if opts.Interface < 1 {
+		return nil, fmt.Errorf("invalid interface index: %d", opts.Interface)
+	}
+
+	rawLink, err := AttachRawLink(RawLinkOptions{
+		Program: opts.Program,
+		Attach:  ebpf.AttachXDP,
+		Target:  opts.Interface,
+		Flags:   uint32(opts.Flags),
+	})
+
+	return rawLink, err
+}
--- a/link/xdp_test.go
+++ b/link/xdp_test.go
@@ -1,0 +1,42 @@
+package link
+
+import (
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+const IfIndexLO = 1
+
+func TestAttachXDP(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.9", "BPF_LINK_TYPE_XDP")
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type: ebpf.XDP,
+		Instructions: asm.Instructions{
+			asm.LoadImm(asm.R0, 2, asm.DWord),
+			asm.Return(),
+		},
+		License: "MIT",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	l, err := AttachXDP(XDPOptions{
+		Program:   prog,
+		Interface: IfIndexLO,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testLink(t, l, testLinkOptions{
+		prog: prog,
+		loadPinned: func(s string, opts *ebpf.LoadPinOptions) (Link, error) {
+			return LoadPinnedRawLink(s, XDPType, opts)
+		},
+	})
+}
--- a/link/link.go
+++ b/link/link.go
@@ -52,6 +52,8 @@ type RawLinkOptions struct {
 	Attach ebpf.AttachType
 	// BTF is the BTF of the attachment target.
 	BTF btf.TypeID
+	// Flags control the attach behaviour.
+	Flags uint32
 }
 
 // RawLinkInfo contains metadata on a link.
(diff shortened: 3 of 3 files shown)


link/xdp.go
```diff
@@ -0,0 +1,72 @@
...
+	AttachTypeXDPDriver
+	// AttachTypeXDPOffload offloads the entire XDP BPF program into hardware.
+	AttachTypeXDPOffload
+)
+
+type XDPOptions struct {
+	// Program must be an XDP BPF program.
+	Program *ebpf.Program
+	// IfName is the interface name to attach to.
+	IfName string
```
lmb, line 33: I think this should be `Interface int`. Less convenient, but `net.Interfaces()` is notorious for doing lots of netlink work.

Alternatively: `*net.Interface`

link/xdp.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil, err
+	}
+
+	var ifIndex *int
+	for _, iface := range ifaces {
+		if iface.Name == opts.IfName {
```
lmb, line 55: For future reference: https://pkg.go.dev/net#InterfaceByName

link/xdp.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// XDPAttachTypeFlags represents the XDP attach type.
+type XDPAttachTypeFlags uint32
+
+const (
+	// AttachTypeXDP tries to link as native XDP but in case
+	// the driver does not support native XDP, it will automatically
+	// fall back to generic XDP (default).
+	AttachTypeXDP XDPAttachTypeFlags = 0
```
lmb, line 18: Do we need the `0` flag? I'd expect people to just skip the flag.

link/xdp.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+	"net"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+// XDPAttachTypeFlags represents the XDP attach type.
+type XDPAttachTypeFlags uint32
```
lmb, line 12: The name is a bit cumbersome. `XDPMode`? `XDPAttachFlags`? `XDPAttachType`?

link/xdp.go
```diff
@@ -0,0 +1,72 @@
...
+
+// AttachXDP links an XDP BPF program to an XDP hook.
+func AttachXDP(opts XDPOptions) (Link, error) {
+	if t := opts.Program.Type(); t != ebpf.XDP {
+		return nil, fmt.Errorf("invalid program type %s, expected XDP", t)
+	}
+
+	if opts.Program.FD() < 0 {
+		return nil, fmt.Errorf("invalid program %w", sys.ErrClosedFd)
+	}
```
lmb, line 46: This is already checked by AttachRawLink.

link/xdp_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func TestAttachXDP(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.7", "BPF_LINK_TYPE_XDP")
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type: ebpf.XDP,
+		Instructions: asm.Instructions{
+			asm.LoadImm(asm.R0, 2, asm.DWord),
+			asm.Return(),
+		},
+		License: "GPL",
```
lmb, line 19: MIT?

lmb (changes requested):

link/xdp.go
mehrdadrad: Zero flag removed.

link/xdp.go
mehrdadrad: Updated to XDPAttachFlags.

link/xdp.go
mehrdadrad: Removed.

link/xdp_test.go
mehrdadrad: I think GPL for XDP without using any specific map. should I change it to MIT or Dual MIT/GPL?

link/xdp.go
mehrdadrad: I just wanted to propose supporting interface name and index together. I pushed the code, what do you think?
I think it's better to give options to developer then they can choose which one is better / works for them.
```go
link.AttachXDP(link.XDPOptions{
    Program: prog,
    Interface: "eth0",
})
```
or   
```go
link.AttachXDP(link.XDPOptions{
    Program: prog,
    Interface: 2,
})
```

link/xdp_test.go
lmb: I prefer MIT if there is no strong reason to use GPL (such as a helper call) since the repo is MIT as well.

link/xdp.go
```diff
@@ -0,0 +1,64 @@
...
+	// XDPOffloadMode offloads the entire XDP BPF program into hardware.
+	XDPOffloadMode
+)
+
+type XDPOptions struct {
+	// Program must be an XDP BPF program.
+	Program *ebpf.Program
+	// Interface is string name of interface or index (int)
+	// to attach program to.
+	Interface interface{}
```
lmb, line 28: I think `interface{}` isn't warranted in this case. If you want to be flexible I would add `InterfaceName` and `InterfaceIndex` and document that `InterfaceName` takes precedence over index if not empty.

link/xdp_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+const IfIndexLO = 1
+
+func TestAttachXDP(t *testing.T) {
```
lmb, line 13: Missing a call to testLink.

lmb (changes requested):

link/xdp_test.go
mehrdadrad: Updated.

link/xdp.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type XDPOptions struct {
+	// Program must be an XDP BPF program.
+	Program *ebpf.Program
+
+	// Interface is the interface index to attach program to.
+	Interface int
+
+	// Flags must match the attach flag of XDP.
```
lmb, line 29: Is that correct? I think you can load XDP without any flags and still force XDPDriverMode when attaching (at least that's possible via netlink).

link/xdp_test.go
```diff
@@ -0,0 +1,47 @@
...
+	testutils.SkipOnOldKernel(t, "5.7", "BPF_LINK_TYPE_XDP")
```
lmb, line 14: This appeared in 5.9 I think https://github.com/torvalds/linux/commit/aa8d3a716b59db6c1ad6c68fb8aa05e31980da60

link/xdp_test.go
```diff
@@ -0,0 +1,47 @@
...
+		License: "MIT",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	l, err := AttachXDP(XDPOptions{
+		Program:   prog,
+		Interface: IfIndexLO,
```
lmb, line 30: `net.InterfaceByName`?

link/xdp_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	testLink(t, l, testLinkOptions{
+		prog: prog,
+		loadPinned: func(s string, opts *ebpf.LoadPinOptions) (Link, error) {
+			return LoadPinnedRawLink(s, XDPType, opts)
+		},
+	})
+
+	err = l.Close()
```
lmb, line 43: Not necessary, testLink closes the link I think.

link/xdp.go
mehrdadrad: It explained more right after this line:
```
// Only one XDP mode should be set, without flag defaults
// to driver/generic mode (best effort).
```
but I just rephrased it line#29 please let me know if you want it something else.

link/xdp_test.go
mehrdadrad: Updated.

link/xdp_test.go
mehrdadrad: Removed.

link/xdp_test.go
mehrdadrad: We don't need: https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/prog_tests/xdp_link.c#L7
But if you want I can add this instead.

lmb (approved):

lmb:
Thanks!

PR #486, opened 2021-11-15
link: clean up tracing and raw_tp links

lmb:
The link package currently has `FreplaceLink` and `btfIDLink`, however those types don't map cleanly to what the kernel exposes.

By looking at their bpf_link_type we can discover the following:
* freplaceLink has type BPF_LINK_TRACING and is created via BPF_LINK_CREATE
* btfIDLink usually has type BPF_LINK_TRACING and is created via BPF_RAW_TRACEPOINT_OPEN
* BPF_RAW_TRACEPOINT_OPEN returns type BPF_LINK_RAW_TP for some program types

Reorganise the code to always use the same Go type for a given bpf_link_type. This means `btfIDLink` is gone, we instead use `rawTracepoint` and `tracing` as appropriate.


lmb:
cc @mehrdadrad please take a look.

link/tracing.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"fmt"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/btf"
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+type tracing struct {
+	RawLink
```
mehrdadrad, line 12: I think it's better to have embedded by pointer, (a bit less memory allocation)
```suggestion
	*RawLink
```

link/tracing_test.go
```diff
@@ -0,0 +1,153 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+				Type:       tt.programType,
+				AttachType: tt.attachType,
+				AttachTo:   tt.attachTo,
+				Instructions: asm.Instructions{
+					asm.LoadImm(asm.R0, 0, asm.DWord),
+					asm.Return(),
+				},
+				License: "GPL",
```
mehrdadrad, line 99: It can change to MIT.

link/tracing_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type:       ebpf.LSM,
+		AttachType: ebpf.AttachLSMMac,
+		AttachTo:   "file_mprotect",
+		Instructions: asm.Instructions{
+			asm.LoadImm(asm.R0, 0, asm.DWord),
+			asm.Return(),
+		},
+		License: "GPL",
```
mehrdadrad, line 135: It can change to MIT.

link/tracing.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	raw := RawLink{fd: fd}
+	info, err := raw.Info()
+	if err != nil {
+		raw.Close()
+		return nil, err
+	}
+
+	if info.Type == RawTracepointType {
```
mehrdadrad, line 123: I think we can not do anything for now but we can add section name to spec then program type. once we have secName then we can check it through prefix. what do you think? 
```go
if strings.HasPrefix(program.SecName, "tp_btf/") {
   return &rawTracepoint{raw}, nil
}
return &tracing{RawLink: RawLink{fd: fd}}, nil
```

link/raw_tracepoint.go
```diff
@@ -33,29 +34,57 @@ func AttachRawTracepoint(opts RawTracepointOptions) (Link, error) {
...
 	}
 
-	return &progAttachRawTracepoint{fd: fd}, nil
+	if err := haveBPFLink(); errors.Is(err, ErrNotSupported) {
+		// Prior to commit 70ed506c3bbc ("bpf: Introduce pinnable bpf_link abstraction")
+		// raw_tracepoints are just a plain fd.
+		return &simpleRawTracepoint{fd}, nil
+	} else if err != nil {
+		return nil, err
+	}
```
mehrdadrad, line 43:
```suggestion
	if err = haveBPFLink(); errors.Is(err, ErrNotSupported) {
		// Prior to commit 70ed506c3bbc ("bpf: Introduce pinnable bpf_link abstraction")
		// raw_tracepoints are just a plain fd.
		return &simpleRawTracepoint{fd}, nil
	}
	if err != nil {
		return nil, err
	}
```

link/tracing.go
lmb: All of the other link types currently embed non-pointer RawLink, so it's better to be consistent here. I think I had a reason for non-pointer at some point, but I can't remember it now. Probably doesn't make a big difference either way.

link/tracing.go
lmb: Adding this to `Program` is problematic, since there is no way to retrieve the section name from a pinned program (see LoadPinnedProgram). We could go via Program.Info() but that isn't better than RawLink.Info().

I have another commit that gets rid of all the different `LoadPinnedTracing`, etc. functions which uses this construct as well.

link/raw_tracepoint.go
lmb: It took me a bit to figure out what your suggestion is about, basically you don't want to shadow `err`?

mehrdadrad: I just wanted to remove extra `else` once we have return at first if statement. a bit cleaner.

link/tracing.go
mehrdadrad: Right, I got this idea from libbpf. https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L10163-L10187

link/raw_tracepoint.go
lmb: Got rid of the else.

mehrdadrad (approved):
LGTM. Thanks!

PR #487, opened 2021-11-16
Fix LINUX_VERSION_CODE handling of 4.14.252+

brycekahle:
Because the mainline kernel tree for 4.14 started forcing the sublevel to `255` early at version `252`, we need a special case to correctly match the `LINUX_VERSION_CODE` expected when loading kprobes.


ti-mo:
Ugh, dear lord. The gift that keeps on giving! Thanks for the patch!

ti-mo (approved):

lmb:
Closing in favor of #489

PR #488, opened 2021-11-16
use clang 13

lmb:
clang 13 has been out for a while and is now available on apt.llvm.org. Switch testdata, etc. to the new version. We still need `clang-9` on CI since `bpf2go` tests rely on it.


ti-mo:
Since we have the llvm (and Go) builder image now, could we consider running the bpf2go tests containerized on CI? I can't get a working clang9 build on Arch anymore since a few weeks now (my libffi is too new) and not sure if this will be fixed. As a result, I have to skip the bpf2go tests for now locally.

lmb:
Yeah we can consider doing that. The downside is that builds will probably take longer due to the 500MiB container image we need to pull. Maybe a good middle ground is to add another makefile target to use the container but not use that on CI yet? Either way, can we tackle that in a separate PR?

ti-mo:
My problem is mostly that `go test` is currently broken on my machine. Couldn't we use the `clang` binary in the Go test suite (should point to _something_ on CI, right?) and a fixed version in the Makefile, since that runs containerized and its output needs to be stable?

lmb:
That's why I added https://github.com/cilium/ebpf/pull/488/commits/76957b98bf03721823bcc82e4358fa2d97b7f2aa to explain why we require clang-9. It's not called by the Makefile (unfortunately, since that already runs in the container).

Some options: skip the tests if clang-9 is missing (kinda suboptimal); use an env var to override the version we test against (less meh, also not great); bump the minimum version and hope we don't introduce incompatibilities.

Do you mind if we merge this PR though? I need clang-13 for some bpf2go stuff.

ti-mo (approved):

PR #489, opened 2021-11-16
Introduce reading of LINUX_VERSION_CODE from vDSO ELF header

brycekahle:
This reads the LINUX_VERSION_CODE from a `note` section of the ELF header for the vDSO dynamically linked library. It reads the memory address of the vDSO library from `/proc/self/auxv`. From my initial research, this is available since 2.x kernels.

I kept the other version parsing code as a fallback, but I'm happy to change/remove it.


lmb (changes requested):
This seems like a big win, not needing to parse kernel version from uname would be great. I'm in favor of chucking the other code. cc @ti-mo

internal/vdso.go
```diff
@@ -0,0 +1,140 @@
...
+			if val == 0 {
+				continue
+			}
+			return elf.NewFile(newUnsafeReader(unsafe.Pointer(uintptr(val))))
+		}
+	}
+	return nil, fmt.Errorf("not found")
+}
+
+type note struct {
```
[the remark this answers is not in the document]
mmat11: https://refspecs.linuxbase.org/elf/elf.pdf search for "Note Section"

ti-mo:
Very neat! Feel free to yank out all of the uname code and replace `detectKernelVersion()` with your vDSO implementation so adding a `LinuxVersionCode()` is not necessary.

I would also like to see this piece of code pedantically commented for posterity and readability. :)

internal/vdso.go
```diff
@@ -0,0 +1,140 @@
...
+		n := note{}
+		if err := binary.Read(sr, hdr.ByteOrder, &n); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return 0, fmt.Errorf("read note: %w", err)
+		}
+		var name string
+		if n.NameSize > 0 {
+			nameData, err := readAligned4(sr, n.NameSize)
```
mmat11, line 108: is it possible to use `binary.Read` + `internal.Align`? (same for desc)

```
		nameData := make([]byte, internal.Align(n.NameSize, 4)))
		err = binary.Read(sr, hdr.ByteOrder, &nameData)
		if err != nil {
			return 0, err
		}
```

internal/vdso.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func vdsoELFHeader() (*elf.File, error) {
+	av, err := os.Open("/proc/self/auxv")
+	if err != nil {
+		return nil, fmt.Errorf("open auxv: %w", err)
+	}
+	defer av.Close()
+
+	abuf, err := ioutil.ReadAll(av)
```
mmat11, line 36:
```suggestion
	abuf, err := io.ReadAll(av)
```

brycekahle:
> Very neat! Feel free to yank out all of the uname code and replace `detectKernelVersion()` with your vDSO implementation so adding a `LinuxVersionCode()` is not necessary.

Can do. Please note that the `LINUX_VERSION_CODE` has the sub-level clamped, so we can back out the `major.minor.sublevel`, but it won't be accurate for sublevel 256+ (and 4.14.252+ 😡 ).

brycekahle:
Updated and added some tests.

internal/vdso.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// loop through all the type+value pairs until we find `AT_SYSINFO_EHDR`, described as:
+	// The address of a page containing the virtual Dynamic
+	// Shared Object (vDSO) that the kernel creates in order to
+	// provide fast implementations of certain system calls.
+	for i := 0; i < len(buf)-1 && buf[i] != _AT_NULL; i += 2 {
+		tag, val := buf[i], buf[i+1]
+		switch tag {
+		case _AT_SYSINFO_EHDR:
+			if val == 0 {
```
lmb, line 71: So there may be multiple _AT_SYSINFO_EHDR? I would expect `0` to be an error or something.

internal/version.go
```diff
@@ -98,66 +95,11 @@ func KernelVersion() (Version, error) {
...
-		// through any other method.
-		v, err := findKernelVersion(string(pvs))
-		if err != nil {
-			return Version{}, err
-		}
-		return v, nil
-	}
-
-	var uname unix.Utsname
-	if err := unix.Uname(&uname); err != nil {
```
lmb, line 124: You can probably remove `Uname` from `internal/unix` as well.

internal/version_test.go
```diff
@@ -46,47 +47,39 @@ func TestKernelVersion(t *testing.T) {
...
+
+func TestVersionFromCode(t *testing.T) {
 	var tests = []struct {
 		name string
-		s    string
+		code uint32
 		v    Version
-		err  bool
 	}{
-		{"ubuntu version_signature", "Ubuntu 4.15.0-91.92-generic 4.15.18", Version{4, 15, 18}, false},
```
lmb, line 56: So these "weird" kernels export which version in the vDSO? `4.15.0` or `4.15.18`?

lmb (approved):
Very nice, thank you.

> we can back out the major.minor.sublevel, but it won't be accurate for sublevel 256+ (and 4.14.252+ rage ).

If I understand correctly we use the version for two reasons: unit tests of feature tests and various skipping logic (only uses major.minor) and to load kprobes. As long as kprobes always accept the version we get from the vDSO I'm fine with this change.

To make this more explicit we could change the formatting of `.255` sublevel to `.255+`?

internal/version_test.go
brycekahle: `4.15.18` would be exported

internal/version.go
brycekahle: It appears to still be used by `KernelRelease`, which is still used by the BTF logic.

internal/vdso.go
brycekahle: I don't think there will be multiple. As-is, it will result in the `not found` error, but after looping through the rest of the pairs.

lmb: In that case I'd either skip the check and allow zero (it'll fail elsewhere i assume) or return a dedicated error here.

internal/version.go
lmb: Ah I see, thanks for checking.

internal/vdso.go
brycekahle: Returning an error now.

brycekahle:
> we could change the formatting of `.255` sublevel to `.255+`?

I don't think we need to do that. I was just making sure you were aware of the limitations.

lmb:
@ti-mo is going to take another look and get this merged then.

ti-mo:
Continued in https://github.com/cilium/ebpf/pull/500.

PR #491, opened 2021-11-17
link: add LoadPinnedLink

lmb:
Add a function that allows loading any pinned link. This allows us to
deprecate all of the separate LoadPinned* functions. In the future we
can use LoadPinnedLink to build APIs that work with any kind of link.

### DIFF
--- a/link/link_test.go
+++ b/link/link_test.go
@@ -46,12 +46,7 @@ func TestRawLink(t *testing.T) {
 		t.Error("Link program ID doesn't match program ID")
 	}
 
-	testLink(t, link, testLinkOptions{
-		prog: prog,
-		loadPinned: func(f string, opts *ebpf.LoadPinOptions) (Link, error) {
-			return LoadPinnedRawLink(f, UnspecifiedType, opts)
-		},
-	})
+	testLink(t, &linkCgroup{*link}, prog)
 }
 
 func TestRawLinkLoadPinnedWithOptions(t *testing.T) {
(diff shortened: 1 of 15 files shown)


mehrdadrad:
https://github.com/cilium/ebpf/blob/c301abe35264aadf77e260d0416f1a890f2e673e/link/link.go#L168
```go
 return nil, fmt.Errorf("get pinned link info: %w", err) 
```

link/link_test.go
```diff
@@ -137,17 +127,17 @@ func testLink(t *testing.T, link Link, opts testLinkOptions) {
...
 
-		link2, err := opts.loadPinned(path, nil)
+		link2, err := LoadPinnedLink(path, nil)
 		if err != nil {
 			t.Fatalf("Can't load pinned %T: %s", link, err)
 		}
 		link2.Close()
 
-		if reflect.TypeOf(link) != reflect.TypeOf(link2) {
+		if _, raw := link.(*RawLink); !raw && reflect.TypeOf(link) != reflect.TypeOf(link2) {
```
mehrdadrad, line 136: This type assertion checking skips all RawLink type like XDP. I think it's better to have this if statement without this logic like below:
```Go
if reflect.TypeOf(link) != reflect.TypeOf(link2) {
```
Then fix the TestRawLink test: as we know this is a cGroup type then we need casting it with appropriate type.
```Go
testLink(t, &linkCgroup{*link}, prog)
```

link/link_test.go
```diff
@@ -46,12 +46,7 @@ func TestRawLink(t *testing.T) {
 		t.Error("Link program ID doesn't match program ID")
 	}
 
-	testLink(t, link, testLinkOptions{
-		prog: prog,
-		loadPinned: func(f string, opts *ebpf.LoadPinOptions) (Link, error) {
-			return LoadPinnedRawLink(f, UnspecifiedType, opts)
-		},
-	})
+	testLink(t, link, prog)
```
mehrdadrad, line 49:
```suggestion
	testLink(t, &linkCgroup{*link}, prog)
```

lmb: Nice, thanks!

PR #492, opened 2021-11-17
asm: add Instructions.SizeBytes() for convenience and readability

ti-mo:
This fixes a subtle bug where buf would always be reallocated if the program contains dword loads.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -275,11 +275,25 @@ ref:
 	}
 }
 
+// Size returns the amount of bytes ins would occupy in binary form.
+func (ins Instruction) Size() uint64 {
+	return uint64(InstructionSize * ins.OpCode.rawInstructions())
+}
+
 // Instructions is an eBPF program.
 type Instructions []Instruction
 
 func (insns Instructions) String() string {
 	return fmt.Sprint(insns)
+}
+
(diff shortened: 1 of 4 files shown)


asm/instruction.go
```diff
@@ -275,13 +275,27 @@ ref:
 	}
 }
 
+// SizeBytes returns the amount of bytes ins would occupy in binary form.
+func (ins Instruction) SizeBytes() uint64 {
```
lmb, line 279: Maybe `Size` is enough?

lmb (approved):
Maybe you've seen this already, there is `asm.RawInstructionOffset` which serves a similar purpose. It feels like there is some overlap, but no clever ideas come to me. Maybe you have one.

lmb:
> This fixes a subtle bug where buf would always be reallocated if the program contains dword loads.

Super nit: not a bug imo, everything still works :P

ti-mo:
> > This fixes a subtle bug where buf would always be reallocated if the program contains dword loads.
> 
> Super nit: not a bug imo, everything still works :P

Yup, but if insns has a single dword, the size of the buffer will invariably be doubled, which isn't great.

asm/instruction.go
ti-mo: Yup, makes sense :+1:

PR #494, opened 2021-11-18
btf: buffered SectionReader, document BTF parsing, vmlinux benchmark

ti-mo:
This PR aims to improve the readability of the logic and obviates the need to call `r.Seek()` and `io.LimitReader()`. It is part of a larger effort to support multiple progs per ELF section, which also requires some changes to lineinfo/funcinfo/corerelo handling. Moving things around this way makes those patches a bit easier to digest. :)

Convenience methods were added to determine the absolute starting offsets of type info, string table, funcInfos, lineInfos and coreRelos within their respective ELF sections.

Documented all functions in ext_info.go.

---

Also added a benchmark over `ParseVmLinux` because I'd like to make this more efficient in the future. (:scream_cat: :scream_cat:)

```
λ  ebpf tb/buffered-section-reader ✓ go test -exec sudo ./internal/btf  -bench=. -benchmem -memprofile memprofile.out -cpuprofile profile.out
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf/internal/btf
BenchmarkParseVmlinux-16    	      10	 111009555 ns/op	75976636 B/op	  511036 allocs/op
PASS
ok  	github.com/cilium/ebpf/internal/btf	3.042s
```

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -53,6 +53,18 @@ type btfHeader struct {
 	StringLen uint32
 }
 
+// typeStart returns the offset from the beginning of the .BTF section
+// to the start of its type entries.
+func (h *btfHeader) typeStart() int64 {
+	return int64(h.HdrLen + h.TypeOff)
+}
+
+// stringStart returns the offset from the beginning of the .BTF section
+// to the start of its string table.
+func (h *btfHeader) stringStart() int64 {
+	return int64(h.HdrLen + h.StringOff)
+}
+
 // LoadSpecFromReader reads BTF sections from an ELF.
 //
 // Returns ErrNotFound if the reader contains no BTF.
@@ -129,7 +141,7 @@ func loadSpecFromELF(file *internal.SafeELFFile, variableOffsets map[variable]ui
 		return spec, nil
 	}
 
-	spec.funcInfos, spec.lineInfos, spec.coreRelos, err = parseExtInfos(btfExtSection.Open(), file.ByteOrder, spec.strings)
+	spec.funcInfos, spec.lineInfos, spec.coreRelos, err = parseExtInfos(btfExtSection, file.ByteOrder, spec.strings)
 	if err != nil {
 		return nil, fmt.Errorf("can't read ext info: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


internal/btf/ext_info.go
```diff
@@ -12,88 +12,126 @@ import (
...
+// lineInfoStart returns the offset from the beginning of the .BTF.ext section
+// to the start of its line_info entries.
+func (h *btfExtHeader) lineInfoStart() int64 {
+	return int64(h.HdrLen + h.LineInfoOff)
+}
+
+// coreReloStart returns the offset from the beginning of the .BTF.ext section
+// to the start of its CO-RE relocation entries.
+func (h *btfExtHeader) coreReloStart(ch *btfExtCoreHeader) int64 {
+	if ch == nil {
```
lmb, line 72: Does the code rely on passing nil to coreReloStart? Seems a bit subtle to me. I'd prefer to not check for nil and crash instead.

internal/btf/ext_info.go
```diff
@@ -12,88 +12,126 @@ import (
...
+// parseBTFExtCoreHeader parses the tail of the .BTF.ext header. If additional
+// header bytes are present, extHeader.HdrLen will be larger than the struct,
+// indicating the presence of a CO-RE extension header.
+func parseBTFExtCoreHeader(r io.Reader, bo binary.ByteOrder, extHeader *btfExtHeader) (*btfExtCoreHeader, error) {
+	var coreHeader btfExtCoreHeader
+	extHdrSize := int64(binary.Size(&extHeader))
 	coreHdrSize := int64(binary.Size(&coreHeader))
+	remainder := int64(extHeader.HdrLen) - extHdrSize
+
 	if remainder >= coreHdrSize {
```
lmb, line 94: Now I wonder if this should be `remainder >= 0`?

internal/btf/ext_info.go
```diff
@@ -12,88 +12,126 @@ import (
...
 		return nil, nil, nil, fmt.Errorf("line info: %w", err)
 	}
 
-	if coreHeader.CoreReloOff > 0 && coreHeader.CoreReloLen > 0 {
-		if _, err := r.Seek(int64(header.HdrLen+coreHeader.CoreReloOff), io.SeekStart); err != nil {
-			return nil, nil, nil, fmt.Errorf("can't seek to CO-RE relocation section: %v", err)
-		}
-
-		relos, err = parseExtInfoRelos(io.LimitReader(r, int64(coreHeader.CoreReloLen)), bo, strings)
+	if coreHeader != nil && coreHeader.CoreReloOff > 0 && coreHeader.CoreReloLen > 0 {
```
lmb, line 132: Looks to me like `coreHeader` can never be nil?

internal/btf/btf_test.go
```diff
@@ -110,6 +107,20 @@ func TestParseVmlinux(t *testing.T) {
...
 }
 
+func BenchmarkParseVmlinux(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		spec, err := parseVmLinux(b)
+		if err != nil {
+			b.Fatal(err)
+		}
+
+		// Access spec to prevent it from being optimized away.
```
lmb, line 117: Does that actually happen?

internal/btf/btf_test.go
```diff
@@ -110,6 +107,20 @@ func TestParseVmlinux(t *testing.T) {
 	}
 }
 
+func BenchmarkParseVmlinux(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		spec, err := parseVmLinux(b)
```
lmb, line 112: This benchmarks gzip decompression as well, which doesn't make sense. This should decompress first and only call loadRawSpec in the loop.

internal/btf/btf_test.go
ti-mo: Thanks, was a leftover from a troubleshooting session. (forgot `b.N` loop)

internal/btf/btf_test.go
ti-mo: Good point!
```
# with gzip
BenchmarkParseVmlinux-16    	      14	 123495986 ns/op
# without gzip
BenchmarkParseVmlinux-16    	      14	  88655530 ns/op
```

internal/btf/ext_info.go
ti-mo: I'm fine either way, I think the chance is slim for upstream to insert another kind of header here instead of a CO-RE header, so we should try to parse if there are _any_ more bytes to read. Then at least it blows up loudly if there aren't `coreHdrSize` amount of bytes to read.

internal/btf/btf.go
```diff
@@ -243,54 +255,58 @@ func loadKernelSpec() (*Spec, error) {
...
-	rawStrings, err := readStringTable(io.LimitReader(rd, int64(header.StringLen)))
+	header, err := parseBTFHeader(rd, bo)
 	if err != nil {
-		return nil, nil, fmt.Errorf("can't read type names: %w", err)
+		return nil, nil, fmt.Errorf("parsing .BTF header: %v", err)
 	}
 
-	if _, err := rd.Seek(int64(header.HdrLen+header.TypeOff), io.SeekStart); err != nil {
-		return nil, nil, fmt.Errorf("can't seek to start of type section: %v", err)
+	buf := internal.NewBufferedSectionReader(rd, header.stringStart(), int64(header.StringLen))
```
lmb, line 302: Something to keep in mind: rd is already in memory, so another layer of buffering isn't useful. Just use SectionReader directly?

internal/btf/btf_test.go
```diff
@@ -110,6 +109,35 @@ func TestParseVmlinux(t *testing.T) {
...
+		b.Fatal(err)
+	}
+
+	var buf bytes.Buffer
+	if _, err := buf.ReadFrom(gr); err != nil {
+		b.Fatal(err)
+	}
+	rd := bytes.NewReader(buf.Bytes())
+
+	for n := 0; n < b.N; n++ {
```
lmb, line 130: This is missing a `b.ResetTimer()` before entering the loop.

lmb (changes requested):

lmb (approved):

PR #496, opened 2021-11-23
prog: update program types from libbpf

mehrdadrad:


elf_reader.go
```diff
@@ -1005,27 +1011,31 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
...
-		"cgroup/sendmsg4":    {CGroupSockAddr, AttachCGroupUDP4Sendmsg, 0},
-		"cgroup/sendmsg6":    {CGroupSockAddr, AttachCGroupUDP6Sendmsg, 0},
-		"cgroup/recvmsg4":    {CGroupSockAddr, AttachCGroupUDP4Recvmsg, 0},
-		"cgroup/recvmsg6":    {CGroupSockAddr, AttachCGroupUDP6Recvmsg, 0},
-		"cgroup/sysctl":      {CGroupSysctl, AttachCGroupSysctl, 0},
-		"cgroup/getsockopt":  {CGroupSockopt, AttachCGroupGetsockopt, 0},
-		"cgroup/setsockopt":  {CGroupSockopt, AttachCGroupSetsockopt, 0},
-		"classifier":         {SchedCLS, AttachNone, 0},
-		"action":             {SchedACT, AttachNone, 0},
+		"struct_ops+":           {StructOps, AttachNone, 0},
```
lmb, line 1014: Madness. Could you add this at then end please?

elf_reader.go
```diff
@@ -1005,27 +1011,31 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
 		"freplace/":             {Extension, AttachNone, 0},
 		"lsm/":                  {LSM, AttachLSMMac, 0},
 		"lsm.s/":                {LSM, AttachLSMMac, unix.BPF_F_SLEEPABLE},
-
-		"cgroup_skb/ingress": {CGroupSKB, AttachCGroupInetIngress, 0},
```
lmb, line 1009: Urgh, what a garbage diff github.

elf_reader.go
```diff
@@ -991,10 +994,13 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
 		"sockops":               {SockOps, AttachCGroupSockOps, 0},
 		"sk_skb/stream_parser":  {SkSKB, AttachSkSKBStreamParser, 0},
 		"sk_skb/stream_verdict": {SkSKB, AttachSkSKBStreamVerdict, 0},
+		"sk_skb":                {SkSKB, AttachNone, 0},
 		"sk_msg":                {SkMsg, AttachSkMsgVerdict, 0},
 		"lirc_mode2":            {LircMode2, AttachLircMode2, 0},
 		"flow_dissector":        {FlowDissector, AttachFlowDissector, 0},
+		"syscall":               {Syscall, AttachNone, 0},
```
lmb, line 1001: Should be after `iter/`

elf_reader.go
```diff
@@ -981,8 +981,11 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
 		"tracepoint/":           {TracePoint, AttachNone, 0},
 		"raw_tracepoint/":       {RawTracepoint, AttachNone, 0},
 		"raw_tp/":               {RawTracepoint, AttachNone, 0},
-		"tp_btf/":               {Tracing, AttachTraceRawTp, 0},
+		"raw_tracepoint.w/":     {RawTracepointWritable, AttachNone, 0},
+		"raw_tp.w/":             {RawTracepointWritable, AttachNone, 0},
 		"xdp":                   {XDP, AttachNone, 0},
+		"xdp_devmap/":           {XDP, AttachXDPDevMap, 0},
```
lmb, line 987: Both of these should be after `syscall`

lmb (changes requested):

elf_reader.go
```diff
@@ -981,8 +981,11 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
 		"tracepoint/":           {TracePoint, AttachNone, 0},
 		"raw_tracepoint/":       {RawTracepoint, AttachNone, 0},
 		"raw_tp/":               {RawTracepoint, AttachNone, 0},
-		"tp_btf/":               {Tracing, AttachTraceRawTp, 0},
```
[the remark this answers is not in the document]
mehrdadrad: Follow the order of libbpf.c is a great idea! I just did and found one more update: "tp/" and one extra "seccomp" (it's not available at libbpf.c but I didn't remove it from the list).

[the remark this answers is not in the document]
lmb: Can you extend the comment at the top and mention that the list should follow libbpf.c?

elf_reader.go
```diff
@@ -973,64 +973,74 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
...
-		"cgroup/getsockopt":  {CGroupSockopt, AttachCGroupGetsockopt, 0},
-		"cgroup/setsockopt":  {CGroupSockopt, AttachCGroupSetsockopt, 0},
-		"classifier":         {SchedCLS, AttachNone, 0},
-		"action":             {SchedACT, AttachNone, 0},
-
-		"cgroup/getsockname4": {CGroupSockAddr, AttachCgroupInet4GetSockname, 0},
-		"cgroup/getsockname6": {CGroupSockAddr, AttachCgroupInet6GetSockname, 0},
-		"cgroup/getpeername4": {CGroupSockAddr, AttachCgroupInet4GetPeername, 0},
-		"cgroup/getpeername6": {CGroupSockAddr, AttachCgroupInet6GetPeername, 0},
+		"seccomp": {SocketFilter, AttachNone, 0},
```
lmb, line 1043: This is not in https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/tree/tools/lib/bpf/libbpf.c yet?

elf_reader.go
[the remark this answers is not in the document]
mehrdadrad: Comment added.

elf_reader.go
mehrdadrad: I couldn't find seccomp anywhere, even old versions. please double check.
It lives here at least since [OCT/2019](https://github.com/cilium/ebpf/blob/729999fbaf2da20b749155e360bcb90a6f72c5da/elf_reader.go) I think we can remove it as it's like an alias for "socket" type.
```go
"socket":              {SocketFilter, AttachNone, 0},
"seccomp":             {SocketFilter, AttachNone, 0},
```

lmb: Seems like this was introduced in https://github.com/newtools/ebpf/commit/ae95a77165a239f8dfab8ac8d6fa6b5d8144f35f which references https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/samples/bpf/bpf_load.c?id=fb40c9ddd66b9c9bb811bbee125b3cb3ba1faee7#n60 but the upstream source doesn't contain `seccomp` either. Let's drop it.

lmb: @nathanjsweet do you remember where `seccomp` came from?

lmb (approved):

elf_reader.go
nathanjsweet: seccomp is a hold-over from an earlier version of the library when we would do an alternate loading logic for seccomp programs. It can safely be removed.

PR #498, opened 2021-11-23
examples: add $BPF_STRIP to bpf2go invocation

lmb:
Update the version of cilium/ebpf used by the examples, and add $BPF_STRIP
to the invocation. Finally, strip the examples by rebuilding them.


ti-mo (approved):
Dependency bump should probably be split from the rest.

PR #499, opened 2021-11-24
Add support for function pointers

### LINKED ISSUE #467: program: support function pointers
It's now possible to pass function pointers to BPF helpers: https://github.com/torvalds/linux/blob/d25f27432f80a800a3592db128254c8140bd71bf/tools/testing/selftests/bpf/progs/for_each_array_map_elem.c#L40-L59

This needs fixing up when loading a program into the kernel: https://github.com/torvalds/linux/commit/53eddb5e04ac5c53a4ccef9f1f900562a5a75246

pippolo84:
This PR adds support for function pointer relocations, following the same logic in libbpf: https://github.com/torvalds/linux/commit/53eddb5e04ac5c53a4ccef9f1f900562a5a75246

For example, this kind of relocation is needed when using helpers such as:
`long bpf_for_each_map_elem(struct bpf_map *map, void *callback_fn, void *callback_ctx, u64 flags)`
which takes a callback as a parameter.

The kernel selftests `for_each_array_{array,map}_elem` have been enabled as they should be supported now.

Fixes #467


elf_reader.go
```diff
@@ -344,6 +344,19 @@ func (ec *elfCode) loadInstructions(section *elfSection) (asm.Instructions, uint
 	}
 }
 
+func offsetToSymbol(section *elfSection, offset int64) (elf.Symbol, error) {
+	if offset < 0 {
+		return elf.Symbol{}, fmt.Errorf("invalid offset %d", offset)
```
lmb, line 349: This isn't much more informative than `no symbol at offset -1` from the error below, so I'd probably skip it. Which means this function is necessary.

elf_reader.go
```diff
@@ -413,44 +426,69 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 		}
 
 	case programSection:
-		if ins.OpCode.JumpOp() != asm.Call {
-			return fmt.Errorf("not a call instruction: %v", ins)
+		if ins.OpCode.JumpOp() != asm.Call && !ins.OpCode.IsDWordLoad() {
```
lmb, line 429: Shouldn't this check that `ins.Src == asm.PseudoFunc`? Might be worth adding a helper like `Instruction.IsLoadOfFunctionPointer()` for this, see `IsFunctionCall`.

elf_reader.go
```diff
@@ -413,44 +426,69 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
+		case elf.STT_NOTYPE:
 			if bind != elf.STB_GLOBAL {
 				return fmt.Errorf("call: %s: unsupported binding: %s", name, bind)
 			}
+		case elf.STT_FUNC:
+			if bind != elf.STB_GLOBAL {
+				return fmt.Errorf("load/call: %s: unsupported binding: %s", name, bind)
+			}
+
+			if ins.OpCode.IsDWordLoad() {
```
lmb, line 447: So it's not valid for a call to reference STT_FUNC, but it's ok for a load? If that is true we should return an error for calls rather than doing nothing.

elf_reader.go
```diff
@@ -413,44 +426,69 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
+		if ins.OpCode.JumpOp() != asm.Call && !ins.OpCode.IsDWordLoad() {
+			return fmt.Errorf("neither a call nor a load instruction: %v", ins)
 		}
 
-		if ins.Src != asm.PseudoCall {
+		if ins.OpCode.JumpOp() == asm.Call && ins.Src != asm.PseudoCall {
 			return fmt.Errorf("call: %s: incorrect source register", name)
 		}
 
 		switch typ {
```
lmb, line 437: I think it might be better to switch on ins.OpCode and then deal with `typ` instead of the other way round. That makes introducing unwanted behaviour for a given opcode less likely.

linker.go
```diff
@@ -61,7 +61,11 @@ func needSection(insns, section asm.Instructions) (bool, error) {
 			continue
 		}
 
-		if ins.OpCode.JumpOp() != asm.Call || ins.Src != asm.PseudoCall {
```
lmb, line 64: This can be rewritten as `!ins.IsFunctionCall()`. Then you can use `!ins.IsFunctionCall() && !ins.IsLoadOfFunctionPointer()` instead of the two ifs.

linker.go
```diff
@@ -112,6 +116,8 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
 		}
 
 		switch {
+		case ins.Src == asm.PseudoFunc:
```
lmb, line 119: This should also be `ins.IsLoadOfFunctionPointer() && ins.Constant == -1`.

lmb (changes requested):
Thanks for the PR! I've left a couple of comments, it looks good overall :)

Can you add some tests outside of the libbpf integration please? It's not trivial to debug those if they fail, and we're pretty lax when getting errors from these tests. I think you should be able to adjust loader.c to include a load of a function pointer.

pippolo84:
> Thanks for the PR! I've left a couple of comments, it looks good overall :)
> 
> Can you add some tests outside of the libbpf integration please? It's not trivial to debug those if they fail, and we're pretty lax when getting errors from these tests. I think you should be able to adjust loader.c to include a load of a function pointer.

Hi @lmb , thank you for your comments. I'll polish the code and add some tests as you suggested.

elf_reader.go
pippolo84: If I didn't miss something, in libbpf the correspondant `asm.PseudoFunc` is used to mark a `ld_imm64` that points to a subprogram and thus need relocation: https://github.com/torvalds/linux/blob/c5c17547b778975b3d83a73c8d84e8fb5ecf3ba5/tools/lib/bpf/libbpf.c#L6091
So it is up to the elf reader to check for such `ld_imm64` and then mark the instruction as a relocatable `asm.PseudoFunc`.
But I added the helper `IsLoadOfFunctionPointer` anyway, to make the code more readable in the relocation phase in `linker.go`.

pippolo84:
I added a test in elf_reader_test.go for subprogram relocation. AFAIK this kind of relocation is supported from 5.13 onwards (running the test with my distro kernel, v5.10, makes it fail, instead it passes on a v5.15).

To avoid complicating things too much in loader.c, I wrote a separate BPF program and Go test, adding `testutils.SkipOnOldKernel` to skip it for older kernel.

pippolo84:
I realized that I misunderstood the reason why the test needs a kernel >= 5.13.
I used the bpf helper `for_each_map_elem`, that's the reason for the requirement. Is this acceptable or should I modify the test to run on earlier kernel version?
I thought about using that helper because it seemed the most common case to exercise the subprogram relocation with `ld_imm64` instruction.

Regarding the CI failure I find it surprising: maybe just a matter of flakyness? Unfortunately I am not able to see the output from the failing test on semaphore.

linker.go
```diff
@@ -112,6 +112,8 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
 		}
 
 		switch {
+		case ins.IsLoadOfFunctionPointer() && ins.Constant == -1:
+			fallthrough
```
lmb, line 116: Nit: this is going to return a confusing error `call at %d` when it's really a load. Maybe replace with `%s at %d` and substitute `ins.OpCode`? Then you can also make the function call case fall through to jump resolution.

testdata/common.h
```diff
@@ -29,10 +29,14 @@ struct bpf_map_def {
...
+
 static long (*trace_printk)(const char *fmt, uint32_t fmt_size, ...) = (void *)6;
 
 static long (*tail_call)(void *ctx, void *prog_array_map, uint32_t index) = (void *)12;
 
 static int (*perf_event_output)(const void *ctx, const void *map, uint64_t index, const void *data, uint64_t size) = (void *)25;
 
 static uint32_t (*get_smp_processor_id)(void) = (void *)8;
+
+static long (*for_each_map_elem)(const void *map, void *callback_fn, void *callback_ctx, uint64_t flags) = (void *)164;
```
lmb, line 42: Nit: newline.

elf_reader_test.go
```diff
@@ -456,6 +456,43 @@ func TestTailCall(t *testing.T) {
 	})
 }
 
+func TestSubprogRelocation(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.13", "subprog addr loaded with ld_imm64")
```
lmb, line 460: This should say that for_each_map_elem is > 5.13 only.

lmb (commented):
I think requiring 5.13 for the test is OK as long as it's documented why. What use is there for subprog loads before for_each_map_elem appeared?

elf_reader.go
```diff
@@ -413,44 +413,66 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
+				}
 
-			// The function we want to call is in the indicated section,
-			// at the offset encoded in the instruction itself. Reverse
-			// the calculation to find the real function we're looking for.
-			// A value of -1 references the first instruction in the section.
-			offset := int64(int32(ins.Constant)+1) * asm.InstructionSize
-			if offset < 0 {
-				return fmt.Errorf("call: %s: invalid offset %d", name, offset)
+			case elf.STT_SECTION:
```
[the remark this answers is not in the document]
pippolo84: No, they are relocated in a slightly different way: https://github.com/libbpf/libbpf/blob/472c0726e84d821186a315889c885b23895b155e/src/libbpf.c#L5963-L5966

linker.go
pippolo84: In case of a jump resolution we need to update the `Offset` field instead of `Constant`.
I tried to rewrite it anyway so to avoid repeating the same relocation logic, but I am not sure the result is that better.

pippolo84:
> I think requiring 5.13 for the test is OK as long as it's documented why. What use is there for subprog loads before for_each_map_elem appeared?

TBH I am not aware of any other case. Digging a little bit through the relevant kernel patches it seems that this kind of relocation has been introduced to support that helper in the first place: https://www.spinics.net/lists/bpf/msg35623.html

linker.go
lmb: Sorry i missed that bit, my bad.

lmb: I ended up amending you commit, I hope you don't mind.

lmb (approved):

linker.go
pippolo84: Np, that's totally fine :+1:

lmb:
Thanks!

pippolo84:
> Thanks!

Thank you for your review and comments!

PR #503, opened 2021-11-24
btf: rename FindType to TypeByName, add TypesByName and TypeByID methods

ti-mo:
This patch adds two querying primitives to btf.Spec:

- `TypeByID()` looks up a `btf.Type` by its ID in the btf.Spec
- `TypesByName()` returns a list of `btf.Type`s that satisfy the given essential name. This can be used to obtain all CO-RE flavors of a struct, to perform a quick lookup without retaining the result, or if the kind or signature of the type you're looking for is not known up front.

In order to simplify the internal graph bookkeeping, harmonize on a single `btf.Type` interface all BTF types must implement instead of the `Type` and `NamedType` interfaces. This facilitates implementing an intuitive API for querying the graph.

Further document the purpose of `essentialName()`.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -31,14 +31,20 @@ type ID uint32
 
 // Spec represents decoded BTF.
 type Spec struct {
-	rawTypes   []rawType
-	strings    stringTable
+	// Data from .BTF.
+	rawTypes []rawType
+	strings  stringTable
+
+	// Inflated Types.
 	types      []Type
-	namedTypes map[string][]NamedType
-	funcInfos  map[string]extInfo
-	lineInfos  map[string]extInfo
-	coreRelos  map[string]coreRelos
-	byteOrder  binary.ByteOrder
+	namedTypes map[string][]Type
+
+	// Data from .BTF.ext.
+	funcInfos map[string]extInfo
+	lineInfos map[string]extInfo
+	coreRelos map[string]coreRelos
+
+	byteOrder binary.ByteOrder
 }
 
 type btfHeader struct {
@@ -373,11 +379,12 @@ func fixupDatasec(rawTypes []rawType, rawStrings stringTable, sectionSizes map[s
 // Copy creates a copy of Spec.
 func (s *Spec) Copy() *Spec {
 	types, _ := copyTypes(s.types, nil)
-	namedTypes := make(map[string][]NamedType)
+
+	namedTypes := make(map[string][]Type)
 	for _, typ := range types {
-		if named, ok := typ.(NamedType); ok {
-			name := essentialName(named.TypeName())
-			namedTypes[name] = append(namedTypes[name], named)
+		if name := typ.TypeName(); name != "" {
+			name = essentialName(name)
+			namedTypes[name] = append(namedTypes[name], typ)
 		}
 	}
 
@@ -484,14 +491,47 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
 	return &Program{s, length, funcInfos, lineInfos, relos}, nil
 }
 
-// FindType searches for a type with a specific name.
+// TypeByID returns the BTF Type with the given type ID.
 //
-// Called T a type that satisfies Type, typ must be a non-nil **T.
-// On success, the address of the found type will be copied in typ.
+// Returns an error wrapping ErrNotFound if a Type with the given ID
+// does not exist in the Spec.
+func (s *Spec) TypeByID(id TypeID) (Type, error) {
+	if int(id) > len(s.types) {
+		return nil, fmt.Errorf("type ID %d: %w", id, ErrNotFound)
+	}
+	return s.types[id], nil
+}
+
+// AnyTypesByName returns a list of BTF Types with the given name.
+//
+// Multiple Types of the same name can exist in case multiple 'struct flavors'
+// are described in BTF info. These are used to deal with changes in kernel
+// data structures.
+//
+// If the BTF blob describes multiple compilation units like vmlinux, multiple
+// Types with the same name and kind can exist, but might not describe the same
+// data structure.
+//
+// Returns an error wrapping ErrNotFound if no matching Type exists in the Spec.
+func (s *Spec) AnyTypesByName(name string) ([]Type, error) {
+	types, ok := s.namedTypes[essentialName(name)]
+	if !ok {
+		return nil, fmt.Errorf("type name %s: %w", name, ErrNotFound)
+	}
+	return types, nil
+}
+
+// TypeByName searches for a Type with a specific name. Since multiple
+// Types with the same name can exist, the parameter typ is taken to
+// narrow down the search in case of a clash.
+//
+// typ must be a non-nil pointer to an implementation of a Type.
+// On success, the address of the found Type will be copied to typ.
 //
 // Returns an error wrapping ErrNotFound if no matching
-// type exists in spec.
-func (s *Spec) FindType(name string, typ interface{}) error {
+// Type exists in the Spec. If multiple candidates are found,
+// an error is returned.
+func (s *Spec) TypeByName(name string, typ interface{}) error {
 	typValue := reflect.ValueOf(typ)
 	if typValue.Kind() != reflect.Ptr {
 		return fmt.Errorf("%T is not a pointer", typ)
@@ -507,13 +547,19 @@ func (s *Spec) FindType(name string, typ interface{}) error {
 		return fmt.Errorf("%T does not satisfy Type interface", typ)
 	}
 
+	types, err := s.AnyTypesByName(name)
+	if err != nil {
+		return err
+	}
+
 	var candidate Type
-	for _, typ := range s.namedTypes[essentialName(name)] {
+	for _, typ := range types {
 		if reflect.TypeOf(typ) != wanted {
 			continue
 		}
 
-		// Match against the full name, not just the essential one.
+		// Match against the full name, not just the essential one
+		// in case the type being looked up is a struct flavor.
 		if typ.TypeName() != name {
 			continue
 		}
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -29,6 +29,10 @@ var (
 // ID represents the unique ID of a BTF object.
 type ID uint32
 
+// essentialName represents the name of a BTF type stripped of any flavor
+// suffixes after a ___ delimiter.
+type essentialname string
+
 // Spec represents decoded BTF.
 type Spec struct {
 	// Data from .BTF.
@@ -36,8 +40,11 @@ type Spec struct {
 	strings  stringTable
 
 	// Inflated Types.
-	types      []Type
-	namedTypes map[string][]Type
+	types []Type
+
+	// Types indexed by essential name.
+	// Includes all struct flavors and types with the same name.
+	namedTypes map[essentialname][]Type
 
 	// Data from .BTF.ext.
 	funcInfos map[string]extInfo
@@ -380,11 +387,11 @@ func fixupDatasec(rawTypes []rawType, rawStrings stringTable, sectionSizes map[s
 func (s *Spec) Copy() *Spec {
 	types, _ := copyTypes(s.types, nil)
 
-	namedTypes := make(map[string][]Type)
+	namedTypes := make(map[essentialname][]Type)
 	for _, typ := range types {
 		if name := typ.TypeName(); name != "" {
-			name = essentialName(name)
-			namedTypes[name] = append(namedTypes[name], typ)
+			en := essentialName(name)
+			namedTypes[en] = append(namedTypes[en], typ)
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


internal/btf/btf.go
```diff
@@ -484,14 +491,44 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
...
-// On success, the address of the found type will be copied in typ.
+// Multiple Types of the same name can exist in case multiple 'struct flavors'
+// are described in BTF info. These are used to deal with changes in kernel
+// data structures.
 //
 // Returns an error wrapping ErrNotFound if no matching
-// type exists in spec.
-func (s *Spec) FindType(name string, typ interface{}) error {
+// Type exists in the Spec.
+func (s *Spec) TypesByName(name string) ([]Type, error) {
```
lmb, line 513: It's possible for multiple types `foo` to exist even without flavours since BTF is merged from multiple compilation units without name spacing. Put another way, the `[]Type` returned from this function may contain types with identical names. It's also possible for multiple types with the same flavour name to exist, but lets not go there. TypeByName flags this problem by returning an error if there are ambiguous matches, and allows resolving it by passing a name including a flavour.

1. TypeByName("foo___bar") returns the exact flavour.
2. TypeByName("foo") returns the single type that has the name "foo".

Case 2 is somewhat problematic since it doesn't find `foo___bar` if that is the only existing type.

Your proposed TypesByName instead returns all types with the same essential name, so `TypesByName("foo___bar")` will return types named `foo`. How about the following instead:

* TypesByName("foo___bar") returns all types for the flavour
* TypesByName("foo") returns all types with essentialName == foo
* TypeByName("foo") returns a type with essentialName == foo as long as there is only a single one

That last point may make some lookups fail that currently work, but it's unlikely to cause issues.

Another discrepancy is that this function doesn't have a `typ` parameter like `TypeByName` does. I'm working on generating types from bpf2go, and for that I wrote [`FindAnyType`](https://github.com/lmb/ebpf/blob/820745cec7d5822f83ed15441c48a26bf54662c5/internal/btf/btf.go#L526):

```Go
func (s *Spec) FindAnyType(name string) (NamedType, error)
```

How about we call the methods `AnyType(s)ByName` and `Type(s)ByName`, with `TypesByName` being unimplemented since it doesn't seem super useful?

lmb (changes requested):
I don't understand the idea behind getting rid of `NamedType`. Is the problem that `[]Type` and `[]NamedType` are not assignable to each other? The code I've written for bpf2go is nicer with `NamedType` vs. having to check the function return: https://github.com/lmb/ebpf/blob/7f73e6889d42ae315d66e8027140850d71c49d0d/cmd/bpf2go/output.go#L258-L292

internal/btf/btf.go
```diff
@@ -484,14 +491,44 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
 	return &Program{s, length, funcInfos, lineInfos, relos}, nil
 }
 
-// FindType searches for a type with a specific name.
+// TypeByID returns the BTF Type with the given type ID.
+//
+// Returns an error wrapping ErrNotFound if a Type with the given ID
+// does not exist in the Spec.
+func (s *Spec) TypeByID(id TypeID) (Type, error) {
```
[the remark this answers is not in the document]
ti-mo: Ah, my bad. I'm using this in the multiprog patch set to resolve func_info type IDs to type (func) names: https://github.com/ti-mo/ebpf/commit/a999aeec82d35fcfeb5319beb1c7decff4e561ae#diff-12372158b51fb6bfad0ba49ec1a51dffe6ac1b8dc4e202d26488c93b53db8107R111.

I'll add a test for this.

ti-mo:
> I don't understand the idea behind getting rid of NamedType. Is the problem that []Type and []NamedType are not assignable to each other?

Yes, the fact that they're both discrete types and that they are/were stored in separate arrays makes it slightly awkward to access them. Also, I don't think it made sense from a data modelling perspective:

- whether a type can have a name or not is a property of a type, not a defining characteristic
- named and unnamed types are not semantically different; even named types can be anonymous, so any code that operates on named types must account for `TypeName() == ""`, which kind of defeats the purpose of having a separate namedType interface to begin with
- the querying interface (here e.g. `TypeByID`) would need a variant for both kinds of types, which will just cause confusion and might be more painful to implement/maintain than necessary

> The code I've written for bpf2go is nicer with NamedType vs. having to check the function return: https://github.com/lmb/ebpf/blob/7f73e6889d42ae315d66e8027140850d71c49d0d/cmd/bpf2go/output.go#L258-L292

Sorry, lacking context, not clear what this code is doing. (comments might help)

---

Now I've had time to mull it over, I think I've also changed my mind about `TypeName()` returning `string, bool`:
```
λ  ebpf tb/multiprog ✓ grep -R "\.TypeName()"
internal/cmd/gentypes/main.go:		if n, _ := cs.TypeName(); !ok || n != "" {
internal/btf/core.go:		localTypeName, _ := localType.TypeName()
internal/btf/ext_info.go:	name, _ := typ.TypeName()
internal/btf/btf.go:		if name, _ := typ.TypeName(); name != "" {
internal/btf/btf.go:		if n, _ := typ.TypeName(); n != name {
internal/btf/types.go:		typeName, _ := typ.TypeName()
internal/btf/format_test.go:				n, _ := t.TypeName()
```

Currently, there's not a call site that actually uses the bool, and I think this is just more expressive:
```
if typ.TypeName == "" && typ.IsNamed() {
  return errors.New("some_error")
}
```

WDYT?

internal/btf/btf.go
ti-mo: That all sounds good :+1: I'm just starting out learning about struct flavors, so not aware of many of the corner cases yet. Could we add tests for the gotchas you've outlined?

My main goal with this proposal was to give the caller the flexibility to resolve ambiguity on their own, by their own rules, to avoid having to overload the lib with specialized filtering logic. I think it makes sense to have this as a general building block.

lmb:
`IsNamed() bool` is confusing I think because it could mean both `name != ""` or `name can be != ""`. Let's go with `TypeName() string` for now and skip the dedicated function. We can figure out what to do once we have the need.

internal/btf/btf.go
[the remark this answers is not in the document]
lmb: This function exposes an awkwardness of the current data model: it gives you a type by the original ID from the BTF. Modifying some Type.ID won't update this. I wish there was a way around that, but I don't have any good ideas. Maybe worth documenting?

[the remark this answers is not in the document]
ti-mo: I wouldn't change the current data model, at least not right now. Currently, a `btf.Spec` is still primarily an abstraction over a raw BTF blob (despite the rawType/Type distinction), and the ID being carried in the Type makes it a leaky abstraction.

I reckon we could remove `Type.ID()` at some point, since the ID is only meaningful within the scope of a raw BTF spec (or a btf.Handle). Once the graph has been inflated from an ELF and the types linked together, a Type's initial ID is no longer meaningful. We should probably move towards something like `btf.Spec.ID(t btf.Type) btf.TypeID` and `btf.Handle.ID(t btf.Type) btf.TypeID`.

Let's revisit this discussion later, we'll likely want to change this anyway during the lazy type unmarshaling work.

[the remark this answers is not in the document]
lmb: I like `Handle.ID` and `Spec.ID`.

> during the lazy type unmarshaling work.

You haven't sold me on that idea yet :P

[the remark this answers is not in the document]
ti-mo: I've added a test that looks up id 0 and asserts it to be `Void` since it's the only type ID (iirc) that's fixed.

internal/btf/btf.go
```diff
@@ -484,14 +494,65 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
...
+//
+// If the BTF blob describes multiple compilation units like vmlinux, multiple
+// Types with the same name and kind can exist, but might not describe the same
+// data structure.
+//
+// Returns an error wrapping ErrNotFound if no matching Type exists in the Spec.
+func (s *Spec) AnyTypesByName(name string) ([]Type, error) {
+	en := essentialName(name)
+
+	// Named types are indexed by essential name.
```
lmb, line 525: An aside: maybe we should add `type essentialname string`, then we could have `map[essentialName][]Type`, etc.

internal/btf/btf.go
```diff
@@ -484,14 +494,65 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
...
+	types, ok := s.namedTypes[en]
+	if !ok {
+		return nil, fmt.Errorf("type name %s: %w", name, ErrNotFound)
+	}
+
+	// If name contains a flavor suffix, require an exact match.
+	if en != name {
+		for _, t := range types {
+			if t.TypeName() == name {
+				return []Type{t}, nil
```
lmb, line 535: I think this isn't correct, there might be multiple flavours after all. So this needs to iterate and filter.

internal/btf/btf.go
```diff
@@ -484,14 +494,65 @@ func (s *Spec) Program(name string, length uint64) (*Program, error) {
...
+	en := essentialName(name)
+
+	// Named types are indexed by essential name.
+	types, ok := s.namedTypes[en]
+	if !ok {
+		return nil, fmt.Errorf("type name %s: %w", name, ErrNotFound)
+	}
+
+	// If name contains a flavor suffix, require an exact match.
+	if en != name {
```
lmb, line 532: What about this instead?

```suggestion
	if en == name {
		return types, nil
	}
```

I think we should probably also copy `types` to avoid the user modifying `namedTypes`.

internal/btf/btf.go
```diff
@@ -507,13 +568,19 @@ func (s *Spec) FindType(name string, typ interface{}) error {
...
-	for _, typ := range s.namedTypes[essentialName(name)] {
+	for _, typ := range types {
 		if reflect.TypeOf(typ) != wanted {
 			continue
 		}
 
-		// Match against the full name, not just the essential one.
+		// Match against the full name, not just the essential one
+		// in case the type being looked up is a struct flavor.
 		if typ.TypeName() != name {
```
lmb, line 584: Doesn't this mean that `TypeByName("foo", ...)` wont return `foo___flavour`?

internal/btf/btf_test.go
```diff
@@ -34,13 +34,61 @@ func parseVmLinux(tb testing.TB) (*Spec, error) {
...
+		tb.Fatalf("ELF %s does not contain .BTF section", file)
+	}
+	rd := sec.Open()
+
+	return loadRawSpec(rd, elf.ByteOrder, nil, nil)
+}
+
+func TestAnyTypesByName(t *testing.T) {
+	testutils.Files(t, testutils.Glob(t, "testdata/relocs-*.elf"), func(t *testing.T, file string) {
+		spec, err := parseELFBTF(t, file)
```
lmb, line 60: Why not just plain `LoadSpecFromReader`?

internal/btf/btf_test.go
```diff
@@ -34,13 +34,61 @@ func parseVmLinux(tb testing.TB) (*Spec, error) {
...
+		types, err := spec.AnyTypesByName("ambiguous")
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if len(types) != 2 {
+			t.Fatalf("expected to receive exactly 2 types from querying ambiguous type, got: %v", types)
+		}
+
+		types, err = spec.AnyTypesByName("ambiguous___flavour")
```
lmb, line 74: Can you add a similar flavour test for `TypeByName`?

lmb (changes requested):
If dealing with struct flavours is too cumbersome we could also change the API, so that resolving struct flavours is the responsibility of the caller? Basically, export `essentialName()` and add an unexported wrapper for CO-RE which does two lookups. Just an idea.

internal/btf/btf.go
ti-mo: Great idea!

internal/btf/btf.go
ti-mo: When you say `there might be multiple flavours`, do you actually mean `there might be multiple flavors with the same name`? (due to multiple compilation units being linked together, etc.)

If `en != name`, `name` contains a flavor suffix, so the first-found match could be returned IMO. This might not be 100% correct if there are multiple instances of the same flavor, but is this something we should account for? If e.g. there are multiple instances of `sk_buff___v46`, shouldn't they all be considered identical? If there are multiple types present with identical names but different layouts, things will blow up at some point down the line anyway...

internal/btf/btf.go
ti-mo: `return types, nil` is already the fallthrough case at the end of the function. This clause here implements matching exact struct flavors.

internal/btf/btf.go
ti-mo: True, but if I understand it correctly, this has always been the case. What should be the behaviour of `TypeByName("foo", ...)` if e.g. `foo`, `foo___v46` and `foo___v51` are present?

internal/btf/btf_test.go
ti-mo: Nice catch, also plumbed the helper into other similar tests.

internal/btf/btf_test.go
ti-mo: Done. 64a9d86 This at least encodes the behaviour you've discussed in https://github.com/cilium/ebpf/pull/503#discussion_r764924572, but we can change this later to include a priority system of some sort.

lmb (approved):

PR #504, opened 2021-11-25
cmd/bpf2go: infer llvm-strip suffix from clang binary

lmb:
llvm is often available in multiple versions, with binaries having the
major version attached as a suffix, for example clang-13. Since adding
stripping it's now necessary to pass two flags when using such
versioned binaries:

    bpf2go -cc clang-13 -strip llvm-strip-13 ...

Add a heuristic which derives the version suffix from the clang binary
and uses that to look for llvm-strip. This allows skipping the -strip
flag in the example above.


ti-mo (approved):
Cool, wanted to suggest something similar :+1:

Issue #506, opened 2021-11-26
link: LoadPinnedLink doesn't expose link information


mehrdadrad:
[LoadPinnedLink](https://github.com/cilium/ebpf/blob/master/link/link.go#L43) doesn't expose [RawLinkInfo](https://github.com/cilium/ebpf/blob/master/link/link.go#L96) like LoadPinnedRawLink (deprecated). The RawLinkInfo already has type, id and programID fields and we can add more type-specific information later to this structure like ifindex for XDP, attach_type for tracing/cgroup/netns/iter etc.
Any thoughts?

lmb:
You're right, the same problem applies to `RawLink.UpdateArgs`. As a workaround you can assert that the Link implements an interface:

```
interface {
    Info() (*RawLinkInfo, error)
}
```

There are a couple of options to fix this:
1. Add `Info() (*RawLinkInfo, error)` to `Link`. This is a problem for BPF link types that have custom metadata like [`bpf_iter_link_info`](https://elixir.bootlin.com/linux/v5.15.5/source/include/uapi/linux/bpf.h#L90). For those we probably want something like `Info() (*IterInfo, error)` with IterInfo embedding RawLinkInfo. With `Info()` part of the interface this becomes impossible.
2. Add structs like `NetNsInfo` for all link types we currently have defined, and export the currently unexported link types. Then the user can simply assert `link.(*Iter).Info()` which embeds a RawLinkInfo. Downside: there is no way to get the RawLinkInfo for a generic `Link` without type asserting it.
3. Add a `struct Info` which contains all the possible metadata. Access to individual fields would have to happen via getters like [`ProgramInfo.MapIDs`](https://pkg.go.dev/github.com/cilium/ebpf#ProgramInfo.MapIDs). Then we can add `Info() (*Info, error)` to `Link`.

Needs some experimentation, but I think the third one seems simplest / the one with the least amount of drawbacks. We  can even make it somewhat backwards compatible by aliasing `NetNsInfo` to `Info`.

mehrdadrad:
@lmb Actually I tried to add type-specific information and expose all link_info last weekend and I just cleaned up the code and added it as draft pull request #509 please review and let me know what do you think.

lmb:
Most of the link types now have info thanks to @mehrdadrad's PR. The two missing ones are RawTracepointInfo and IterInfo since they contain strings.

prash2512:
Hi 👋 
I'm new here.
Would like to work on this issue.

[...]

PR #509, opened 2021-11-29
link: expose link info with type-specific information

mehrdadrad:
This PR is related to issue #506 
- add new method: Info() (*RawLinkInfo, error) to Link interface.
- add bpf_link_info union member types to types generator.
- update sys/types with generator.
- add info method to types that they don't support.
- enhance a few tests to cover info. (needs to add more) 
- fix NetNsInfo type.

### DIFF
--- a/internal/cmd/gentypes/main.go
+++ b/internal/cmd/gentypes/main.go
@@ -168,7 +168,8 @@ import (
 			[]patch{
 				replace(enumTypes["LinkType"], "type"),
 				replace(linkID, "id"),
-				truncateAfter("prog_id"),
+				name(3, "extra"),
+				replaceUnionWithBytes("extra"),
 			},
 		},
 		{"FuncInfo", "bpf_func_info", nil},
@@ -355,7 +356,31 @@ import (
 		return attrs[i].goType < attrs[j].goType
 	})
 
-	attrTypes, err := splitBpfAttr(spec)
+	var bpfAttr *btf.Union
+	if err := spec.TypeByName("bpf_attr", &bpfAttr); err != nil {
+		return nil, err
+	}
+	attrTypes, err := splitUnion(bpfAttr, types{
+		{"map_create", "map_type"},
+		{"map_elem", "map_fd"},
+		{"map_elem_batch", "batch"},
+		{"prog_load", "prog_type"},
+		{"obj_pin", "pathname"},
+		{"prog_attach", "target_fd"},
+		{"prog_run", "test"},
+		{"obj_next_id", ""},
+		{"info_by_fd", "info"},
+		{"prog_query", "query"},
+		{"raw_tracepoint_open", "raw_tracepoint"},
+		{"btf_load", "btf"},
+		{"task_fd_query", "task_fd_query"},
+		{"link_create", "link_create"},
+		{"link_update", "link_update"},
+		{"link_detach", "link_detach"},
+		{"enable_stats", "enable_stats"},
+		{"iter_create", "iter_create"},
+		{"prog_bind_map", "prog_bind_map"},
+	})
 	if err != nil {
 		return nil, fmt.Errorf("splitting bpf_attr: %w", err)
 	}
@@ -381,6 +406,55 @@ import (
 		}
 	}
 
+	// Link info type specific
+
+	linkInfoExtraTypes := []struct {
+		goType  string
+		cType   string
+		patches []patch
+	}{
+		{"CgroupLinkInfo", "cgroup", []patch{replace(enumTypes["AttachType"], "attach_type")}},
+		{"IterLinkInfo", "iter", []patch{replace(pointer, "target_name"), truncateAfter("target_name_len")}},
+		{"NetNsLinkInfo", "netns", []patch{replace(enumTypes["AttachType"], "attach_type")}},
+		{"RawTracepointLinkInfo", "raw_tracepoint", []patch{replace(pointer, "tp_name")}},
+		{"TracingLinkInfo", "tracing", []patch{replace(enumTypes["AttachType"], "attach_type")}},
+		{"XDPLinkInfo", "xdp", nil},
+	}
+
+	sort.Slice(linkInfoExtraTypes, func(i, j int) bool {
+		return linkInfoExtraTypes[i].goType < linkInfoExtraTypes[j].goType
+	})
+
+	var bpfLinkInfo *btf.Struct
+	if err := spec.TypeByName("bpf_link_info", &bpfLinkInfo); err != nil {
+		return nil, err
+	}
+
+	member := bpfLinkInfo.Members[len(bpfLinkInfo.Members)-1]
+	bpfLinkInfoUnion, ok := member.Type.(*btf.Union)
+	if !ok {
+		return nil, fmt.Errorf("there is not type-specific union")
+	}
+
+	linkInfoTypes, err := splitUnion(bpfLinkInfoUnion, types{
+		{"raw_tracepoint", "raw_tracepoint"},
+		{"tracing", "tracing"},
+		{"cgroup", "cgroup"},
+		{"iter", "iter"},
+		{"netns", "netns"},
+		{"xdp", "xdp"},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("splitting linkInfo: %w", err)
+	}
+
+	for _, s := range linkInfoExtraTypes {
+		t := linkInfoTypes[s.cType]
+		if err := outputPatchedStruct(gf, w, s.goType, t, s.patches); err != nil {
+			return nil, fmt.Errorf("output %q: %w", s.goType, err)
+		}
+	}
+
 	return w.Bytes(), nil
 }
 
@@ -403,42 +477,16 @@ func outputPatchedStruct(gf *btf.GoFormatter, w *bytes.Buffer, id string, s *btf
 	return nil
 }
 
-func splitBpfAttr(spec *btf.Spec) (map[string]*btf.Struct, error) {
-	var bpfAttr *btf.Union
-	if err := spec.TypeByName("bpf_attr", &bpfAttr); err != nil {
-		return nil, err
-	}
+type types []struct {
+	name                string
+	cFieldOrFirstMember string
+}
 
-	types := []struct {
-		name                string
-		cFieldOrFirstMember string
-	}{
-		{"map_create", "map_type"},
-		{"map_elem", "map_fd"},
-		{"map_elem_batch", "batch"},
-		{"prog_load", "prog_type"},
-		{"obj_pin", "pathname"},
-		{"prog_attach", "target_fd"},
-		{"prog_run", "test"},
-		{"obj_next_id", ""},
-		{"info_by_fd", "info"},
-		{"prog_query", "query"},
-		{"raw_tracepoint_open", "raw_tracepoint"},
-		{"btf_load", "btf"},
-		{"task_fd_query", "task_fd_query"},
-		{"link_create", "link_create"},
-		{"link_update", "link_update"},
-		{"link_detach", "link_detach"},
-		{"enable_stats", "enable_stats"},
-		{"iter_create", "iter_create"},
-		{"prog_bind_map", "prog_bind_map"},
-	}
-
+func splitUnion(union *btf.Union, types types) (map[string]*btf.Struct, error) {
 	structs := make(map[string]*btf.Struct)
 
 	for i, t := range types {
-		member := bpfAttr.Members[i]
-
+		member := union.Members[i]
 		s, ok := member.Type.(*btf.Struct)
 		if !ok {
 			return nil, fmt.Errorf("%q: %s is not a struct", t.name, member.Type)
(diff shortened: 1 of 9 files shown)


internal/sys/types.go
```diff
@@ -72,6 +74,7 @@ const (
 	BPF_PROG_ATTACH                 Cmd = 8
 	BPF_PROG_DETACH                 Cmd = 9
 	BPF_PROG_TEST_RUN               Cmd = 10
+	BPF_PROG_RUN                    Cmd = 10
```
lmb, line 77: Did you update `btf/testdata/vmlinux-btf.gz` to get the new definitions? If yes that file should be checked in as well.

link/link.go
```diff
@@ -97,6 +102,26 @@ type RawLinkInfo struct {
 	Type    Type
 	ID      ID
 	Program ebpf.ProgramID
+	Extra   interface{}
+}
+
+func (r RawLinkInfo) ExtraRawTP() *sys.RawTPExtraLinkInfo {
```
lmb, line 108: This should be `(*RawTracepointLinkInfo, bool)` so to that it's possible to type assert without panicking.

Types from `sys` are awkward for users, since they can't import `sys`. So we might have to have aliases for the extra link info.

link/link.go
```diff
@@ -260,9 +285,34 @@ func (l *RawLink) Info() (*RawLinkInfo, error) {
...
+	case XDPType:
+		extra = &sys.XDPExtraLinkInfo{}
+	default:
+		return nil, fmt.Errorf("unknown link info type: %d", info.Type)
+	}
+
+	buf := bytes.NewReader(info.Extra[:])
+	err := binary.Read(buf, internal.NativeEndian, extra)
+	if err != nil {
+		return nil, err
```
lmb, line 309: Nit: the error needs some context.

lmb (changes requested):
This is quite nice! Shoving the `extra interface{}` in there is a good idea, thanks for that.

internal/sys/types.go
mehrdadrad: I switched to testdata/vmlinux-btf.gz to generate types.go

internal/cmd/gentypes/main.go
```diff
@@ -615,3 +701,22 @@ func rename(from, to string) patch {
 		return fmt.Errorf("no member named %q", from)
 	}
 }
+
+func addUnion(name string) patch {
```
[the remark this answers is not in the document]
mehrdadrad: Renamed to replaceUnionWithBytes and it supports name and position for anonymous union.
I added a long comment for this func please let me know if it doesn't need.

internal/cmd/gentypes/main.go
```diff
@@ -412,6 +453,51 @@ func outputPatchedStruct(gf *btf.GoFormatter, w *bytes.Buffer, id string, s *btf
...
+		{"netns", "netns_ino"},
+		{"xdp", "ifindex"},
+	}
+
+	u, ok := member.Type.(*btf.Union)
+	if !ok {
+		return nil, fmt.Errorf("there is not type-specific union")
+	}
+
+	for i, t := range types {
```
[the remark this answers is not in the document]
mehrdadrad: Added splitUnion for both.

internal/cmd/gentypes/main.go
```diff
@@ -390,6 +390,47 @@ import (
...
+			"CGroupExtraLinkInfo", "cgroup",
+			[]patch{},
+		},
+		{
+			"IterExtraLinkInfo", "iter",
+			[]patch{},
+		},
+		{
+			"NetNsExtraLinkInfo", "netns",
+			[]patch{rename("netns_ino", "net_ns_ino")},
```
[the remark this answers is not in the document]
mehrdadrad: Removed all renames.

internal/cmd/gentypes/main.go
```diff
@@ -390,6 +390,47 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("splitting bpf_attr: %w", err)
+	}
+	linkInfoExtraTypes := []struct {
+		goType  string
+		cType   string
+		patches []patch
+	}{
+		{
+			"CGroupExtraLinkInfo", "cgroup",
```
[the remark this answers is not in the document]
mehrdadrad: Renamed to CgroupLinkInfo.

link/link.go
```diff
@@ -35,6 +37,9 @@ type Link interface {
 	// not called.
 	Close() error
 
+	// Info returns metadata on a link.
+	Info() (*RawLinkInfo, error)
```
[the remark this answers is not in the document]
mehrdadrad: Renamed RawLinkInfo to LinkInfo and type RawLinkInfo = LinkInfo added.
The "function may return ErrNotSupported" comment added.
Testing link info added into testLink.

link/link.go
```diff
@@ -97,6 +102,26 @@ type RawLinkInfo struct {
 	Type    Type
 	ID      ID
 	Program ebpf.ProgramID
+	Extra   interface{}
```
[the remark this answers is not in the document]
mehrdadrad: It became unexported.

link/link.go
mehrdadrad: Fixed.
I think we don't need to add aliases but let me know if we need to add.

link/link.go
mehrdadrad: Context added.

link/netns.go
```diff
@@ -58,5 +56,5 @@ func (nns *NetNsLink) Info() (*NetNsInfo, error) {
 	if err != nil {
```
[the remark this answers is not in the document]
mehrdadrad: Removed.

internal/cmd/gentypes/main.go
```diff
@@ -390,6 +390,47 @@ import (
...
+		{
+			"TracingExtraLinkInfo", "tracing",
+			[]patch{},
+		},
+		{
+			"XDPExtraLinkInfo", "xdp",
+			[]patch{rename("ifindex", "if_index")},
+		},
+	}
+	for _, s := range linkInfoExtraTypes {
```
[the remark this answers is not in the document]
mehrdadrad: It's already sorted alphabetically, should I add sort func like attr?

internal/sys/types.go
```diff
@@ -920,3 +922,34 @@ func RawTracepointOpen(attr *RawTracepointOpenAttr) (*FD, error) {
 	}
 	return NewFD(int(fd))
 }
+
+type CgroupLinkInfo struct {
+	CgroupId   uint64
+	AttachType uint32
```
lmb, line 928: Needs replace(enumTypes["AttachType"], "attach_type")

link/link.go
```diff
@@ -92,11 +99,40 @@ type RawLinkOptions struct {
...
-type RawLinkInfo struct {
+// LinkInfo contains metadata on a link.
+type LinkInfo struct {
 	Type    Type
 	ID      ID
 	Program ebpf.ProgramID
+	extra   interface{}
+}
+
+// RawLinkInfo deprecated: use LinkInfo instead.
```
lmb, line 110: I think the Deprecated has to be on it's own line for linters to do their work.

```suggestion
// RawLinkInfo contains information on a raw link.
//
// Deprecated: use LinkInfo instead.
```

link/link.go
```diff
@@ -92,11 +99,40 @@ type RawLinkOptions struct {
 	Flags uint32
 }
 
-// RawLinkInfo contains metadata on a link.
-type RawLinkInfo struct {
+// LinkInfo contains metadata on a link.
+type LinkInfo struct {
```
lmb, line 103: Nit: this reads `link.LinkInfo`, `link.Info` is better imo.

link/link.go
```diff
@@ -92,11 +99,40 @@ type RawLinkOptions struct {
...
 	Type    Type
 	ID      ID
 	Program ebpf.ProgramID
+	extra   interface{}
+}
+
+// RawLinkInfo deprecated: use LinkInfo instead.
+type RawLinkInfo = LinkInfo
+
+func (r LinkInfo) ExtraRawTracepoint() *sys.RawTracepointLinkInfo {
```
lmb, line 113: I was thinking of returning `(*RawInfo, bool)` but I guess this works too. Can you document that the functions can return nil?

internal/cmd/gentypes/main.go
```diff
@@ -615,3 +669,32 @@ func rename(from, to string) patch {
 		return fmt.Errorf("no member named %q", from)
 	}
 }
+
+// replaceUnionWithBytes replaces the union with bytes within a struct.
+// you can specify the union name or the union position (anonymous).
+// the position can be -1, it means the first union in struct will be replaced
+// otherwise with a given position or name, it replaces exactly the specified
+// union name or union position with bytes.
+func replaceUnionWithBytes(name string, position int, bytesName string) patch {
```
lmb, line 678: I think you can decompose this into two steps instead: `name(3, "extra"), replaceUnionWithBytes("extra")`. See the discussion re `.Extra` though, I think the approach won't work.

```
func name(member int, name string) patch {
	return modifyNth(func(m *btf.Member) error {
		if m.Name != "" {
			return fmt.Errorf("member already has name %q", m.Name)
		}

		m.Name = name
		return nil
	}, member)
}
```

lmb (changes requested):
I realised now that `LinkInfo.Extra` being a blob of bytes is a problem: we can't specify input arguments to BPF_OBJ_GET_INFO_BY_FD. This means we can never pass a pointer to the kernel to retrieve things, for example `IterLinkInfo.TargetName`. It also means that we have to know the link type to be able to retrieve info from the kernel. :facepalm: 

I think we'll have to do the following:
* `link.RawLink` needs a new `typ link.Type` field which is populated when creating a link and when LoadPinnedLink is called.
* For link info types that require input arguments we generate a flattened `struct bpf_link_info`. This means including `prog_id` etc. `sys.IterLinkInfo` is an example of this.
* Each link type has a dedicated `link.IterInfo`, `link.NetNsInfo`, etc. struct which basically contains the info specific to that link type (with pointer, length pairs replaced with strings, etc.)
* `RawLink.Info()` switches on `RawLink.typ` to figure out which flattened struct has to be used and passes pointers if necessary. If no input arguments are required we can use `sys.LinkInfo` and parse from .Extra. The result of the syscall is broken down into `link.Info` and `link.IterInfo`, etc. as they work now.

internal/sys/types.go
mehrdadrad: Updated.

internal/sys/types.go
```diff
@@ -920,3 +922,34 @@ func RawTracepointOpen(attr *RawTracepointOpenAttr) (*FD, error) {
...
+}
+
+type RawTracepointLinkInfo struct {
+	TpName    uint64
+	TpNameLen uint32
+	_         [4]byte
+}
+
+type TracingLinkInfo struct {
+	AttachType  uint32
```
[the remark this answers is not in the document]
mehrdadrad: Updated.

mehrdadrad:
> I realised now that `LinkInfo.Extra` being a blob of bytes is a problem: we can't specify input arguments to BPF_OBJ_GET_INFO_BY_FD. This means we can never pass a pointer to the kernel to retrieve things, for example `IterLinkInfo.TargetName`. It also means that we have to know the link type to be able to retrieve info from the kernel. 🤦
> 
@lmb I fixed the pass a pointer to the kernel through bytes please review it. I think this way we have less code and less complexity. I addressed other feedbacks as well.

internal/cmd/gentypes/main.go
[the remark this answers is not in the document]
mehrdadrad: I forgot to fix this. I'll do this later.

link/link.go
mehrdadrad: Renamed.

link/link.go
mehrdadrad: Updated.

internal/sys/types.go
mehrdadrad: Updated.

internal/sys/types.go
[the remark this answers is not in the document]
mehrdadrad: Updated.

internal/sys/types.go
```diff
@@ -920,3 +922,34 @@ func RawTracepointOpen(attr *RawTracepointOpenAttr) (*FD, error) {
...
+	Map           struct{ MapId uint32 }
+}
+
+type NetNsLinkInfo struct {
+	NetnsIno   uint32
+	AttachType uint32
+}
+
+type RawTracepointLinkInfo struct {
+	TpName    uint64
```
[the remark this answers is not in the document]
mehrdadrad: Updated.

internal/cmd/gentypes/main.go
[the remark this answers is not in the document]
mehrdadrad: sort.Slice added.

internal/sys/types.go
```diff
@@ -920,3 +922,34 @@ func RawTracepointOpen(attr *RawTracepointOpenAttr) (*FD, error) {
...
 }
+
+type CgroupLinkInfo struct {
+	CgroupId   uint64
+	AttachType uint32
+	_          [4]byte
+}
+
+type IterLinkInfo struct {
+	TargetName    uint64
```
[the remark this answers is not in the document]
mehrdadrad: IterInfo and RawTracepointInfo added.
```go
type RawTracepointInfo struct {
	TPName string
}
type IterInfo struct {
	TargetName string
}
```
Also I added the below types (not alias):
```go
type TracingInfo sys.TracingLinkInfo
type CgroupInfo sys.CgroupLinkInfo
type NetNsInfo sys.NetNsLinkInfo
type XDPInfo sys.XDPLinkInfo
```

internal/sys/types.go
```diff
@@ -920,3 +922,34 @@ func RawTracepointOpen(attr *RawTracepointOpenAttr) (*FD, error) {
...
+	TargetNameLen uint32
+	Map           struct{ MapId uint32 }
```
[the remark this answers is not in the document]
mehrdadrad: truncateAfter("target_name_len") added.

internal/sys/types.go
```diff
@@ -920,3 +922,34 @@ func RawTracepointOpen(attr *RawTracepointOpenAttr) (*FD, error) {
...
+}
+
+type NetNsLinkInfo struct {
+	NetnsIno   uint32
+	AttachType uint32
```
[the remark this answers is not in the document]
mehrdadrad: Updated.

link/link.go
mehrdadrad: Document added.

internal/cmd/gentypes/main.go
mehrdadrad: Decomposed.

mehrdadrad:
> Manually encoding pointers into `LinkInfo.Extra` is problematic, I'm not sure how to best handle this at the moment. Could you get along without RawTracepointInfo and IterInfo? If we can skip these two the PR is good to go, otherwise it needs a bunch more work.

I removed IterType and RawTracepointType.

lmb (approved):

lmb:
Thanks!

PR #510, opened 2021-11-29
Load BTF for maps and parse program rodata.

player-two:
The library allows users change the value of volatile consts, which clang writes
to the rodata section of the elf file, before loading the program and its map
into the kernel.  This change allows users to read the values back out of the
map by parsing its contents with the linked btf data.

The use case for this is knowing if a program should be reloaded due to desired changes in these constants.  It's easy to see if the program has changed by looking at `ProgramInfo.Tag`, but for anyone using `CollectionSpec.RewriteConstants`, that's only half of the story.

I'm very new to btf.  I got this working for the included example/test by referencing other parts of the parsing code, but I only have a half-baked understanding of how the btf data is parsed into these go structures, mostly by studying `bpftool btf dump` output for some example programs.


lmb:
> it would be a bit odd to combine that with CollectionSpec.RewriteConstants, since that method takes a map[string]interface{}.

True, the interface for using the generated type would look different, probably via CollectionOptions. You'd always use that type, not `map[string]interface{}`.

> The nice thing about the map is that it's unordered, but then when reading the rodata back as a []byte, I would have to assume the order.

Are you trying to compare .rodata across two different compilations (versions) of the same program? My gut feeling is that if you first compare program tag and then binary compare .rodata you're probably OK. I don't have proof though.

lmb:

PR #511, opened 2021-11-30
Support passing the input context to BPF_PROG_TEST_RUN

player-two:
This allows setting program-specific metadata during test runs.  For example,
SKB programs can also set certain fields on the SKB instead of just the packet
data.

The feature was first added to the kernel in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=b0b9395d865e3060d97658fbc9ba3f77fecc8da1

I want to test a skb program that references the firewall mark, similar to this example: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/bpf/prog_tests/skb_ctx.c

I opted to add a new method for this, but I could add the argument to `Program.Test()` if you're okay with a breaking change.  I could also go the route of using the option pattern so that it's easier to support other fields of `sys.ProgRunAttr` down the road.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -459,7 +459,26 @@ func (p *Program) Close() error {
 //
 // This function requires at least Linux 4.12.
 func (p *Program) Test(in []byte) (uint32, []byte, error) {
-	ret, out, _, err := p.testRun(in, 1, nil)
+	ret, out, _, err := p.testRun(in, nil, 1, nil)
+	if err != nil {
+		return ret, nil, fmt.Errorf("can't test program: %w", err)
+	}
+	return ret, out, nil
+}
+
+// Test runs the Program in the kernel with the given input and metadata and
+// returns the value returned by the eBPF program. outLen may be zero.
+//
+// Note: the kernel expects at least 14 bytes input for an ethernet header for
+// XDP and SKB programs.
+//
+// This function requires at least the following Linux versions by program type:
+// * flow dissector: 5.4
+// * SK lookup: 5.13
+// * SKB: 5.2
+// * XDP: 5.15
+func (p *Program) TestWithContext(dataIn []byte, ctxIn []byte) (uint32, []byte, error) {
+	ret, out, _, err := p.testRun(dataIn, ctxIn, 1, nil)
 	if err != nil {
 		return ret, nil, fmt.Errorf("can't test program: %w", err)
 	}
@@ -478,7 +497,7 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 //
 // This function requires at least Linux 4.12.
 func (p *Program) Benchmark(in []byte, repeat int, reset func()) (uint32, time.Duration, error) {
-	ret, _, total, err := p.testRun(in, repeat, reset)
+	ret, _, total, err := p.testRun(in, nil, repeat, reset)
 	if err != nil {
 		return ret, total, fmt.Errorf("can't benchmark program: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


prog.go
```diff
@@ -543,16 +562,20 @@ func (p *Program) testRun(in []byte, repeat int, reset func()) (uint32, []byte,
...
-		DataSizeOut: uint32(len(out)),
-		DataIn:      sys.NewSlicePointer(in),
-		DataOut:     sys.NewSlicePointer(out),
+		DataSizeIn:  uint32(len(dataIn)),
+		DataSizeOut: uint32(len(dataOut)),
+		DataIn:      sys.NewSlicePointer(dataIn),
+		DataOut:     sys.NewSlicePointer(dataOut),
 		Repeat:      uint32(repeat),
 	}
+	if len(ctxIn) != 0 {
```
lmb, line 575: What about CtxOut?

lmb (commented):
Some recent kernel version introduced `BPF_PROG_RUN`. It's just an alias for `BPF_PROG_TEST_RUN` but it shows that upstream considers RUN to be a basic primitive on programs. Another thing that has become apparent is that the input required by different types of programs differs wildly.

* For some prog types ctx must be null, for others it's optional
* For some prog types data must be null and ctx is required (SK_LOOKUP)

Instead of `TestWithContext` I would add a method `Program.Run(RunOptions) (RunResult, error)`. Names TBD. `RunOptions` contains all the possible fields like `Data`, `Context`, etc. RunResult contains data output, context output, etc. `Test` and `Benchmark` keep the same API but are wrappers around `Run`.

cc @ti-mo

ti-mo:
Agree with @lmb with the additional point that we should try to avoid the term `Context` wherever possible as to not cause any confusion with `context.Context`. Using it as a struct field should be fine, but what bpf calls a 'context' is really an `Input`, not an actual context. The user should still be able to make the connection with how this is exposed on the bpf side (`ctx`), but comments might be sufficient for this.

ti-mo:
I've also been looking at this recently, see my comment https://github.com/cilium/ebpf/pull/537#issuecomment-1015434776.

From `uapi/linux/bpf.h`:
```
 *		Some program types have particular requirements:
 *
 *		**BPF_PROG_TYPE_SK_LOOKUP**
 *			*data_in* and *data_out* must be NULL.
 *
 *		**BPF_PROG_TYPE_RAW_TRACEPOINT**,
 *		**BPF_PROG_TYPE_RAW_TRACEPOINT_WRITABLE**
 *
 *			*ctx_out*, *data_in* and *data_out* must be NULL.
 *			*repeat* must be zero.
```

Should we perhaps encode these constraints in API? e.g.:

- `Program.Run(dataIn []byte)` as an alias to `Program.Test()`
- `Program.RunSkLookup(ctxIn SkLookupContext, ctxOut *SkLookupContext)`
- `Program.RunRawTracepoint(ctxIn []byte)`

This feels easier to maintain and more user-friendly than a catch-all `RunOptions` where we'll have to account for all permutations of options over time.

lmb:
> This feels easier to maintain and more user-friendly than a catch-all RunOptions where we'll have to account for all permutations of options over time.

It's going to add a ton of methods to Program. And we probably want structs anyways to be able to amend the API in the future.

ti-mo:
> > This feels easier to maintain and more user-friendly than a catch-all RunOptions where we'll have to account for all permutations of options over time.
> 
> It's going to add a ton of methods to Program. And we probably want structs anyways to be able to amend the API in the future.
> 
> P.S.: I'm always torn about encoding kernel API constraints. It ends up being a lot of work, only for upstream to change the API and us being incompatible with the upstream change. Definitely nicer for users though.

Agreed, `Run(RunOptions)` sounds good. :+1:

lmb:
I'm going to close this for now. Feel free to reopen at a later point.

PR #513, opened 2021-12-01
map: allow empty ProgramArray to not be assigned

lmb:
xdpcap [1] used ProgramArrays to implement hooks in XDP programs.
So far, it's been possible to add an xdpcap hook without calling
into the library to patch the ProgramArray. The change to support
pre-populating tail call maps broke this, since we now don't allow
ProgramArrays which are loaded but not requested by the user.

However, this case it is OK for the ProgramArray to be closed since
there are no programs in it anyways. Make the error less strict by
allowing empty ProgramArrays to be unassigned.

1: https://github.com/cloudflare/xdpcap

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -244,9 +244,14 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 		switch m.typ {
 		case ProgramArray:
 			// Require all lazy-loaded ProgramArrays to be assigned to the given object.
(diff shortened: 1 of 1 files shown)


arthurfabre (approved):
Thanks for figuring this out and making a PR!

ti-mo (approved):
Thanks! I've expanded the comment since this is somewhat tricky stuff.

PR #514, opened 2021-12-02
asm: Add Jump32 support

arthurfabre:
The unused 0x6 class now represents jumps with 32bit comparisons:

https://lore.kernel.org/all/1548523574-18316-16-git-send-email-jiong.wang@netronome.com/T/
http://patchwork.ozlabs.org/project/netdev/list/?series=88386&state=*

This complements 32bit ALU operations by allowing comparisons without
having to sign extend registers beforehand.

Add a Jump32Class to represent this. As some JumpOps are not supported
by Jump32 (Call, Exit, Ja), centralize the logic for validating jump
OpCodes in OpCode.JumpOp().

### DIFF
--- a/asm/opcode.go
+++ b/asm/opcode.go
@@ -30,6 +30,9 @@ const (
 	ALUClass Class = 0x04
 	// JumpClass jump operators
 	JumpClass Class = 0x05
+	// Jump32Class jump operators with 32 bit comparaisons
+	// Requires kernel 5.1
+	Jump32Class Class = 0x06
 	// ALU64Class arithmetic in 64 bit mode
 	ALU64Class Class = 0x07
 )
@@ -53,9 +56,9 @@ func (cls Class) IsALU() bool {
 	return cls == ALUClass || cls == ALU64Class
 }
 
-// IsJump checks if this is JumpClass.
+// IsJump checks if this is either JumpClass or Jump32Class.
 func (cls Class) IsJump() bool {
-	return cls == JumpClass
+	return cls == JumpClass || cls == Jump32Class
 }
 
 func (cls Class) isJumpOrALU() bool {
@@ -134,11 +137,20 @@ func (op OpCode) Endianness() Endianness {
 }
 
 // JumpOp returns the JumpOp.
+// Returns InvalidJumpOp if it doesn't encode a jump.
 func (op OpCode) JumpOp() JumpOp {
 	if !op.Class().IsJump() {
 		return InvalidJumpOp
 	}
-	return JumpOp(op & jumpMask)
+
+	jumpOp := JumpOp(op & jumpMask)
+
+	// Some JumpOps are only supported by JumpClass, not Jump32Class.
+	if op.Class() == Jump32Class && (jumpOp == Exit || jumpOp == Call || jumpOp == Ja) {
+		return InvalidJumpOp
+	}
+
+	return jumpOp
 }
 
 // SetMode sets the mode on load and store operations.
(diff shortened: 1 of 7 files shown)


arthurfabre:
Fixed a typo in a test.

arthurfabre:
I force pushed to re-run the tests, but CI failed again with:

```
E: Failed to fetch https://storage.googleapis.com/bazel-apt/dists/stable/jdk1.8/binary-amd64/Packages.gz  File has unexpected size (6503 != 6709). Mirror sync in progress? [IP: 142.250.186.112 443]
```

I'll try again in a bit.

asm/opcode_test.go
```diff
@@ -0,0 +1,49 @@
...
+			qt.Assert(t, opcode, qt.Not(qt.Equals), InvalidOpCode)
+			qt.Assert(t, opcode.JumpOp(), qt.Equals, op)
+		} else {
+			qt.Assert(t, opcode, qt.Equals, InvalidOpCode)
+			qt.Assert(t, opcode.JumpOp(), qt.Equals, InvalidJumpOp)
+		}
+	}
+
+	// Exit, call and JA aren't allowed with Jump32
+	test(Jump32Class, Exit, false)
```
lmb, line 23: All of these tests are going to assert on the same line, can you turn this into a table driven test instead?

asm/opcode.go
```diff
@@ -125,11 +129,20 @@ func (op OpCode) Endianness() Endianness {
...
-	if op.Class().encoding() != jumpOrALU {
+	if op.Class() != JumpClass && op.Class() != Jump32Class {
 		return InvalidJumpOp
 	}
-	return JumpOp(op & jumpMask)
+
+	jumpOp := JumpOp(op & jumpMask)
+
+	// Some JumpOps are only supported by JumpClass, not Jump32Class.
+	if op.Class() == Jump32Class && (jumpOp == Exit || jumpOp == Call || jumpOp == Ja) {
```
lmb, line 141: I don't think we should do this here. There is an asymmetry in using `asm`:
* When using it to assemble programs we shouldn't allow constructing invalid programs, so `SetJumpOp` should return Invalid.
* When using it to introspect programs we should allow decoding instructions as much as possible. `JumpOp` should do the minimum checks necessary.

linker.go
```diff
@@ -121,7 +121,7 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
 
 			ins.Constant = int64(callOffset - offset - 1)
 
-		case ins.OpCode.Class() == asm.JumpClass && ins.Offset == -1:
+		case ins.OpCode.JumpOp() != asm.InvalidJumpOp && ins.Offset == -1:
```
lmb, line 124: I find this harder to read. How about assing `Class.IsJump()` and then doing the following:

```suggestion
		case ins.OpCode.Class().IsJump() && ins.Offset == -1:
```

We could apply `IsJump`, `IsALU` in other places as well.

lmb (changes requested):

lmb:
Ping @arthurfabre do you want to pick this up again?

asm/opcode.go
arthurfabre: The kernel docs specifically call them out as BPF_JUMP only: https://www.kernel.org/doc/Documentation/networking/filter.txt

```
  BPF_JA    0x00  /* BPF_JMP only */
  BPF_CALL  0x80  /* eBPF BPF_JMP only: function call */
  BPF_EXIT  0x90  /* eBPF BPF_JMP only: function return */
```

Right now the kernel rejects them. If it does accept them in the future, I'm half worried they'll be abused for some different purpose - it feels like labeling them as Invalid / Unknown would help a user notice something is up. But I'm happy to change it if you feel differently.

linker.go
arthurfabre: Good idea! Added a commit to introduce these, and `IsStore()`, `IsLoad()` as well.

asm/opcode.go
lmb: Ok, there are a bunch of places in the library which use JumpOp to to function rewrity things, so returning invalidJumpOp here is the cautious option.

lmb (approved):

PR #520, opened 2021-12-07
btf: support raw BTF in LoadSpecFromReader

lmb:
Allow passing BTF not wrapped in an ELF to LoadSpecFromReader. This
allows users to use external BTF for vmlinux, which is usually shipped
without the ELF.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -78,12 +78,18 @@ func (h *btfHeader) stringStart() int64 {
 	return int64(h.HdrLen + h.StringOff)
 }
 
-// LoadSpecFromReader reads BTF sections from an ELF.
+// LoadSpecFromReader reads from an ELF or a raw BTF blob.
 //
-// Returns ErrNotFound if the reader contains no BTF.
+// Returns ErrNotFound if reading from an ELF which contains no BTF.
 func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 	file, err := internal.NewSafeELFFile(rd)
 	if err != nil {
+		if bo := guessRawBTFByteOrder(rd); bo != nil {
+			// Try to parse a naked BTF blob. This will return an error if
+			// we encounter a Datasec, since we can't fix it up.
+			return loadRawSpec(io.NewSectionReader(rd, 0, math.MaxInt64), bo, nil, nil)
+		}
+
 		return nil, err
 	}
 	defer file.Close()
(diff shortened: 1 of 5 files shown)


internal/btf/btf_test.go
```diff
@@ -2,47 +2,44 @@ package btf
...
+	sync.Once
+	err error
+	raw []byte
+}
 
-	rd, err := gzip.NewReader(fh)
-	if err != nil {
-		tb.Fatal(err)
-	}
+func readVmLinux(tb testing.TB) *bytes.Reader {
```
lmb, line 23: This is a drive by refactoring, I can split it into a separate commit.

internal/btf/btf.go
```diff
@@ -65,10 +65,25 @@ func (h *btfHeader) stringStart() int64 {
 	return int64(h.HdrLen + h.StringOff)
 }
 
-// LoadSpecFromReader reads BTF sections from an ELF.
+// LoadSpecFromReader reads from an ELF or a raw BTF blob.
 //
-// Returns ErrNotFound if the reader contains no BTF.
+// Returns ErrNotFound if reading from an ELF which contains no BTF.
 func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
```
ti-mo, line 71: I think the control flow could be cleaned up here. I don't think we really need to check for ELF magic, `NewSafeELFFile` already does that.

Instead, `guessBTFByteOrder()` could be called on the happy path as a first step and do an early return of `loadRawSpec()` if BTF magic is found. Otherwise, fall through to parsing it as an ELF, and the existing code further down will complain if anything's wrong.

internal/btf/btf_test.go
ti-mo: Sure, would be nice if it's not too difficult.

ti-mo (changes requested):
Awesome, this would add some nice flexibility. Left a minor comment, I think we can get away with less code.

internal/btf/btf.go
lmb: I ended up reshuffling this so that we fall back to raw BTF if opening the elf file fails for some reason. This means that ELF stays the happy path.

lmb:
This is good to go if someone has time for a look.

internal/btf/btf.go
ti-mo: Sounds good :+1:

internal/btf/btf.go
```diff
@@ -299,6 +295,20 @@ func parseBTFHeader(r io.Reader, bo binary.ByteOrder) (*btfHeader, error) {
 	return &header, nil
 }
 
+func guessRawBTFByteOrder(r io.ReaderAt) binary.ByteOrder {
+	for _, bo := range []binary.ByteOrder{
+		binary.LittleEndian,
+		binary.BigEndian,
+	} {
+		_, err := parseBTFHeader(internal.NewBufferedSectionReader(r, 0, math.MaxInt64), bo)
```
ti-mo, line 303: I think the use of `NewBufferedSectionReader` is unnecessary, `io.NewSectionReader` should be sufficient.

ti-mo (approved):
One nit, looks great!

PR #521, opened 2021-12-07
tidy up examples

lmb:
Clean up a couple of small things which have been bugging me for a bit.


ti-mo (approved):
Thanks, all makes sense :+1: kprobepin .o's were not regenerated for some reason, pushed a fixup. Will merge when CI is green.

PR #523, opened 2021-12-09
Add ReadTimeout to perf/ring buffer

brycekahle:
Adds the function `ReadTimeout` to allow non-indefinite waits when reading from a perf buffer.

I'm not in love with the `ErrNoRecords` sentinel error, but I wanted to see what you folks thought. Maybe `io.EOF` or exporting `errEOR` instead?

Alternatively, I could make it more Go-ish, changing the function signature to `ReadContext(ctx context.Context)`, but it would need to handle manual cancellation too.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -17,9 +17,8 @@ import (
 )
 
 var (
-	ErrClosed    = os.ErrClosed
-	ErrNoRecords = errors.New("no records")
-	errEOR       = errors.New("end of ring")
+	ErrClosed = os.ErrClosed
+	errEOR    = errors.New("end of ring")
 )
 
 // perfEventHeader must match 'struct perf_event_header` in <linux/perf_event.h>.
@@ -271,17 +270,17 @@ func (pr *Reader) Read() (Record, error) {
 	return pr.read(-1)
 }
 
-// ReadTimeout is Read but will timeout and return ErrNoRecords
-// if the timeout expires. The minimum timeout is 1 millisecond.
+// ReadTimeout is Read but will time out and return os.ErrDeadlineExceeded
+// if the timeout expires.
 func (pr *Reader) ReadTimeout(timeout time.Duration) (Record, error) {
-	if timeout < time.Millisecond {
-		return Record{}, fmt.Errorf("timeout too small")
+	if timeout > 0 && timeout < time.Millisecond {
+		return Record{}, fmt.Errorf("non-zero timeout too small")
 	}
 
-	return pr.read(int(timeout.Nanoseconds() / int64(time.Millisecond)))
+	return pr.read(timeout)
 }
 
-func (pr *Reader) read(msec int) (Record, error) {
+func (pr *Reader) read(timeout time.Duration) (Record, error) {
 	pr.mu.Lock()
 	defer pr.mu.Unlock()
 
@@ -291,12 +290,12 @@ func (pr *Reader) read(msec int) (Record, error) {
 
 	for {
 		if len(pr.epollRings) == 0 {
-			nEvents, err := pr.poller.Wait(pr.epollEvents, msec)
+			nEvents, err := pr.poller.Wait(pr.epollEvents, int(timeout.Milliseconds()))
 			if err != nil {
 				return Record{}, err
 			}
 			if nEvents == 0 {
-				return Record{}, ErrNoRecords
+				return Record{}, os.ErrDeadlineExceeded
 			}
 
 			for _, event := range pr.epollEvents[:nEvents] {
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -290,7 +290,12 @@ func (pr *Reader) read(timeout time.Duration) (Record, error) {
 
 	for {
 		if len(pr.epollRings) == 0 {
-			nEvents, err := pr.poller.Wait(pr.epollEvents, int(timeout.Milliseconds()))
+			msec := int(timeout.Milliseconds())
+			if timeout == -1 {
+				msec = -1
+			}
+
+			nEvents, err := pr.poller.Wait(pr.epollEvents, msec)
 			if err != nil {
 				return Record{}, err
 			}
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -288,13 +288,13 @@ func (pr *Reader) read(timeout time.Duration) (Record, error) {
 		return Record{}, fmt.Errorf("perf ringbuffer: %w", ErrClosed)
 	}
 
+	msec := int(timeout.Milliseconds())
+	if timeout == -1 {
+		msec = -1
+	}
+
 	for {
 		if len(pr.epollRings) == 0 {
-			msec := int(timeout.Milliseconds())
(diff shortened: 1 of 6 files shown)


perf/reader.go
```diff
@@ -266,6 +268,20 @@ func (pr *Reader) Close() error {
 //
 // Calling Close interrupts the function.
 func (pr *Reader) Read() (Record, error) {
+	return pr.read(-1)
+}
+
+// ReadTimeout is Read but will timeout and return ErrNoRecords
+// if the timeout expires. The minimum timeout is 1 millisecond.
+func (pr *Reader) ReadTimeout(timeout time.Duration) (Record, error) {
+	if timeout < time.Millisecond {
```
ti-mo, line 277: I'm not sure if we should enforce this. A timeout of 0 is valid and can be used for polling if the caller wants to do that.

What we should clearly document, though, is the following:
```
	ms := time.Millisecond
	fmt.Println(ms, ms.Milliseconds())  // 1ms 1
	ms -= time.Microsecond
	fmt.Println(ms, ms.Milliseconds()) // 999µs 0
	ms /= 2
	fmt.Println(ms, ms.Milliseconds()) // 499.5µs 0
```

Looks like it always gets rounded down, so perhaps we could error on `timeout > 0 && timeout < time.Millisecond` to prevent accidental busypolling?

perf/reader.go
```diff
@@ -275,10 +291,13 @@ func (pr *Reader) Read() (Record, error) {
 
 	for {
 		if len(pr.epollRings) == 0 {
-			nEvents, err := pr.poller.Wait(pr.epollEvents)
+			nEvents, err := pr.poller.Wait(pr.epollEvents, msec)
```
ti-mo, line 294: Do the `timeout.Milliseconds()` conversion here instead.

ringbuf/reader.go
```diff
@@ -7,6 +7,7 @@ import (
 	"io"
```
ti-mo, line 7: See my comments on the perf part.

ti-mo (changes requested):
As a concept, I think taking `context.Context` is fine (this is Go after all), but for such a low-level API, it doesn't make much sense performance-wise. `context.WithTimeout` makes several allocations, takes out a lock and needs to be called for every call to `ReadTimeout()`. A simple value for setting a timeout would be just fine here. :+1:

Overall I think this is a good addition. Implementing cancellation is not needed IMO, either read with timeouts or close the reader if you're done with it.

brycekahle:
@ti-mo Addressed all your feedback if you want to give it another look.

perf/reader.go
```diff
@@ -266,6 +267,20 @@ func (pr *Reader) Close() error {
 //
 // Calling Close interrupts the function.
 func (pr *Reader) Read() (Record, error) {
+	return pr.read(-1)
+}
+
+// ReadTimeout is Read but will time out and return os.ErrDeadlineExceeded
+// if the timeout expires.
+func (pr *Reader) ReadTimeout(timeout time.Duration) (Record, error) {
```
lmb, line 275: Should this be a deadline vs. a time out?

brycekahle: Do you mean taking a `time.Time` and computing the millisecond timeout value from that? If so, that would make it hard/impossible to end up with a timeout value of `0` which is useful for polling.

lmb: Not sure I understand the use case of polling to be honest, so I'd be OK to forego that for now.

As a user I'd expect the timeout to specify the time until the next Record. However, what it really does is bound how long we block in epoll. Those are two distinct, since we might in the future generate a spurious wakeup (e.g. to implement suspension without map shenanigans). As a user I don't (and shouldn't care) that epoll was woken. So taking a deadline here is simpler to implement and easier to understand I think.

brycekahle: > So taking a deadline here is simpler to implement

Would you mind branching and showing me? A deadline seems to be more complicated to me, since the underlying epoll API supports the timeout, so we simply pass it on.

ti-mo: Why not stick with a timeout? For network i/o, expressing timeouts as deadlines makes sense e.g. to bound a send/recv to the same window, but I'm not sure if that would be very useful here.

> So taking a deadline here is simpler to implement and easier to understand I think.

Hmm, epoll doesn't have the equivalent of `(net.Conn).SetReadDeadline` iirc. With your suggestion, would the caller pass in a `time.Time` using `time.Now().Add(time.Millisecond * 100)`, which the library code then has to reverse using `t.Sub(time.Now())` to obtain a relative timeout value to pass to `epoll_pwait`? That's going to be less accurate than passing a `time.Duration` and won't allow the caller to busypoll.

lmb: Here you go: https://github.com/cilium/ebpf/commit/26fab34199f4ff02e816bb7d0f991306f22433f0

* Pushes the deadline handling into the poller, so that `ringbuf` can later also benefit from the same code
* Fixes an incosistent timeout if EpollWait returns a temporary error

lmb: That commit should `if msec < 0 { msec = 0 }` then you can get polling by doing `ReadDeadline(time.Now())`

ti-mo: @lmb Your proposal seems to be more or less what I described. Is this really useful if the user is mostly interested in waiting for a duration instead of a deadline? Alternatively, why not both `ReadTimeout` and `ReadDeadline`)?

lmb: We can implement both, I'm just trying to say that `ReadTimeout()` would be implemented in terms of `ReadDeadline(time.Now().Add(to))`.

ti-mo: Since msec timeout is the native format taken by epoll_wait, why not the other way around? :)

lmb: I'm trying to say: the deadline or timeout parameter should indicate "the time i'm willing to wait until Read returns a record", not "the time I'm willing to wait for the next epoll notification". The current implementation does the latter.

Why is this relevant? For example, `ringbuf` can be woken from epoll but not produce a Record, because items in the ring may be marked as "discarded". The same can happen here if we end up changing the implementation. To avoid this the natural implementation is for Poller to take a deadline until which it may block.

If we add both timeout and deadline methods, how about `SetTimeout` and `SetDeadline` like for net.Conn etc.?

mythi: (sorry for joining late here)

> Why is this relevant? For example, `ringbuf` can be woken from epoll but not produce a Record,

Wanted to add the other use case `ringbuf` has: read from raw ring without epoll. Should this be handled as part of the `Set*` as a special case or we want it to be completely separate `ReadRaw()` call or something?

lmb: @mythi I probably need a more fully fleshed out example, maybe open a discussion about it?

lmb:
@brycekahle do you want me to do another round of review?

brycekahle:
> @brycekahle do you want me to do another round of review?

I think we probably need to figure out the API design issue first.

lmb:

PR #524, opened 2021-12-09
Add WakeupEvents support to perf buffers

brycekahle:
- Add support for `wakeup_events` as an alternative to `wakeup_watermark`.
- Adds a `TriggerAll` function to `perf.Reader` to allow reading of queued events that haven't triggered the threshold of `wakeup_events` yet.

Note: This depends on `ReadTimeout` from #523

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -134,6 +134,11 @@ type Reader struct {
 // ReaderOptions control the behaviour of the user
 // space reader.
 type ReaderOptions struct {
+	// The number of events required in any per CPU buffer before
+	// Read will process data. This is mutually exclusive with Watermark.
+	// The default is zero, which means Watermark will take precedence.
+	WakeupEvents int
+
 	// The number of written bytes required in any per CPU buffer before
 	// Read will process data. Must be smaller than PerCPUBuffer.
 	// The default is to start processing as soon as data is available.
@@ -185,7 +190,7 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 	// but doesn't allow using a wildcard like -1 to specify "all CPUs".
 	// Hence we have to create a ring for each CPU.
 	for i := 0; i < nCPU; i++ {
-		ring, err := newPerfEventRing(i, perCPUBuffer, opts.Watermark)
+		ring, err := newPerfEventRing(i, perCPUBuffer, opts.Watermark, opts.WakeupEvents)
 		if errors.Is(err, unix.ENODEV) {
 			// The requested CPU is currently offline, skip it.
 			rings = append(rings, nil)
@@ -303,6 +308,18 @@ func (pr *Reader) Read() (Record, error) {
 		}
 
 		return record, err
+	}
+}
+
+// TriggerAll forces all rings to be queued up reading on the next call to Read.
+// This is useful to forcefully read pending events that haven't met the threshold of WakeupEvents yet.
+func (pr *Reader) TriggerAll() {
+	pr.mu.Lock()
+	defer pr.mu.Unlock()
+
+	for _, ring := range pr.rings {
+		pr.epollRings = append(pr.epollRings, ring)
+		ring.loadHead()
 	}
 }
 
--- a/perf/reader_test.go
+++ b/perf/reader_test.go
@@ -279,7 +279,7 @@ func TestPerfReaderClose(t *testing.T) {
 }
 
 func TestCreatePerfEvent(t *testing.T) {
-	fd, err := createPerfEvent(0, 1)
+	fd, err := createPerfEvent(0, 1, 0)
 	if err != nil {
 		t.Fatal("Can't create perf event:", err)
 	}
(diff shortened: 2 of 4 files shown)


perf/ring.go
```diff
@@ -86,17 +86,25 @@ func (ring *perfEventRing) Close() {
 	ring.mmap = nil
 }
 
-func createPerfEvent(cpu, watermark int) (int, error) {
-	if watermark == 0 {
-		watermark = 1
+func createPerfEvent(cpu, wakeupBytes int, wakeupEvents int) (int, error) {
```
lmb, line 89: This should return an error if both bytes and events are set.

perf/reader.go
```diff
@@ -306,6 +311,18 @@ func (pr *Reader) Read() (Record, error) {
 	}
 }
 
+// TriggerAll forces all rings to be queued up reading on the next call to Read.
+// This is useful to forcefully read pending events that haven't met the threshold of WakeupEvents yet.
+func (pr *Reader) TriggerAll() {
```
lmb, line 316: How do you want to use this? I've played with this idea in the past, but it was never clear to me how to make this square with Read() semantics. You can't avoid blocking in Read to get all the remaining samples without knowing how many samples there are.

brycekahle: That's a good point. We might need a new function that reads each ring until `errEOR` and returns all those records at once? That or a way to opt-out of the call to `poller.Wait` once `len(pr.epollRings) == 0`.

lmb: How do you interrupt that `errEOR` function though? It's turtles all the way down. Hence my question what your use case is, maybe there is some other way to solve it.

brycekahle: No need for interruption because it isn't blocking. It would essentially be this loop:
```
for {
        if len(pr.epollRings) == 0 {
	        return Record{}, io.EOF
        }
        // Start at the last available event. The order in which we
        // process them doesn't matter, and starting at the back allows
        // resizing epollRings to keep track of processed rings.
        record, err := readRecordFromRing(pr.epollRings[len(pr.epollRings)-1])
        if err == errEOR {
	        // We've emptied the current ring buffer, process
	        // the next one.
	        pr.epollRings = pr.epollRings[:len(pr.epollRings)-1]
	        continue
        }
        return record, err
}
```

Our use case is using `wakeup_events` to do a sort of batching. We observe a lot of CPU being used to wakeup for every event, so we want to wakeup every X. We have a synchronization point in time where we need to read all queued events, even if they haven't triggered the batch size X.

lmb: > We have a synchronization point in time where we need to read all queued events, even if they haven't triggered the batch size X.

But you still keep reading afterwards, right? That's easier to solve than flush and read remaining,

joamaki: I'd suggest adding e.g. `ReadAll(maxRecords int, timeoutMsec int, receive func(record Record)) error`, which would first poll all rings and then if there's something to read in any of them, it'll loop through all of the rings until they're empty.

Though... the tricky part is probably figuring out all the different ways this might need to be parameterized, and for some high-perf use-cases we might want to have thread-per-ring (LockOSThread'd and SetAffinity'd to the same cpu as the ring). So perhaps providing the low-level primitives to build this in any way you want plus few sane wrappers would be the way to go?

perf/reader.go
```diff
@@ -134,6 +134,11 @@ type Reader struct {
 // ReaderOptions control the behaviour of the user
 // space reader.
 type ReaderOptions struct {
+	// The number of events required in any per CPU buffer before
+	// Read will process data. This is mutually exclusive with Watermark.
+	// The default is zero, which means Watermark will take precedence.
+	WakeupEvents int
```
ti-mo, line 140: > ... This is mutually exclusive with Watermark.
> // The default is zero, which means Watermark will take precedence.

It would be much nicer if the caller was actually informed of this configuration conflict. There's no way to introspect how the reader is ultimately configured, so the caller won't be aware of one parameter being chosen over the other.

perf/reader_test.go
```diff
@@ -279,7 +279,7 @@ func TestPerfReaderClose(t *testing.T) {
 }
 
 func TestCreatePerfEvent(t *testing.T) {
-	fd, err := createPerfEvent(0, 1)
+	fd, err := createPerfEvent(0, 1, 0)
```
ti-mo, line 282: I would probably pass `ReaderOptions` here, this call site is quickly becoming incrutable. :)

perf/ring.go
ti-mo: This should probably take a `ReaderOptions` instead of more ints.

perf/ring.go
```diff
@@ -22,12 +22,12 @@ type perfEventRing struct {
 	*ringReader
 }
 
-func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error) {
-	if watermark >= perCPUBuffer {
+func newPerfEventRing(cpu, perCPUBuffer, wakeupBytes int, wakeupEvents int) (*perfEventRing, error) {
```
ti-mo, line 25: Idem, `ReaderOptions` would be nicer.

perf/reader.go
brycekahle: The CPU usage isn't necessarily the fault of this library, but rather what we do with those events, which is why we batch them to reduce the frequency. Doing this batching in eBPF ends up costing us stack size and larger maps to store the batched data. Being able to do this batching directly in the perf buffer would be a huge improvement for us.

Let me think about the flush use case, to see if a better design shakes out.

lmb: I'm happy with WakeupEvents as it stands if that is of any use to you on its own?

lmb:

PR #525, opened 2021-12-13
btf: only return exact matches from AnyTypesByName

lmb:
Change AnyTypesByName so that essential names stay an implementation
detail of the btf package: we need spec.namedTypes for fast
flavour lookup, but it's not clear that we should expose the flavour
name distinction to the users.

Fix this by only returning types which have an exact name match. This
means that `AnyTypesByName("ambiguous")` does not return ambiguous___flavour.

### DIFF
--- a/internal/btf/btf_test.go
+++ b/internal/btf/btf_test.go
@@ -58,8 +58,8 @@ func TestAnyTypesByName(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		if len(types) != 2 {
-			t.Fatalf("expected to receive exactly 2 types from querying ambiguous type, got: %v", types)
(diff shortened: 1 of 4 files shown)


internal/btf/btf_test.go
```diff
@@ -58,7 +58,7 @@ func TestAnyTypesByName(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		if len(types) != 2 {
+		if len(types) != 1 {
 			t.Fatalf("expected to receive exactly 2 types from querying ambiguous type, got: %v", types)
```
ti-mo, line 62: Error message no longer corresponds, will fix this up

ti-mo (approved):

PR #526, opened 2021-12-14
cmd/bpf2go: allow to specify native target

tklauser:
This allows to invoke `bpf2go` to build for the native architecture and
have `__TARGET_ARCH_xxx` defined based on `GOARCH` without having to
derive the  architecture e.g. as part of the build system.

For example this would be useful in github.com/cilium/pwru:

    //go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cc clang -target native KProbePWRU ./bpf/kprobe_pwru.c -- -DOUTPUT_SKB -I./bpf/headers

### DIFF
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"sort"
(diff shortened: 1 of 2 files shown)


lmb (approved):
Looks good, thanks!

PR #527, opened 2021-12-14
map, prog: fix typo in MEMLOCK error string

tklauser:
s/bay be to low/may be to low/


lmb (approved):

ti-mo (approved):
I've read this line so many times and never spotted this. Should I be worried?

lmb:
Thanks!

PR #529, opened 2021-12-16
elf_reader: emit `UnspecifiedProgram`s into `CollectionSpec`

ti-mo:
eBPF programs emitted to sections with unknown prefixes were previously withheld from CollectionSpec.Programs. Add a ProgramSpec.SectionName field to allow the caller to determine (Attach)Type manually, and include all programs that were explicitly assigned to an ELF section.


collection.go
```diff
@@ -286,7 +286,11 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (*Co
 		}
 	}
 
-	for progName := range spec.Programs {
+	for progName, prog := range spec.Programs {
+		if prog.Type == UnspecifiedProgram {
```
ti-mo, line 290: I feel like this would be the expected behaviour, and should keep things backwards-compatible. Instead of filtering `UnspecifiedProgram`s out of the `CollectionSpec`, filter them out on their way into the kernel.

Any other ideas to accomplish this?

lmb: Hmm. What happens if a prog array points at an UnspecifiedProgram?

Seems OK on the surface.

lmb (approved):

collection.go
lmb: Probably worth documenting this.

ti-mo: The lazy-loader doesn't take this code path afaik, but the kernel should complain about an invalid program. Good point, will investigate.

lmb: FWIW I think returning an error when a programarray references unspecifiedprogram is OK.

ti-mo: I've added an explicit `UnspecifiedProgram` check to `(collectionLoader).loadProgram()` since there's not much point in trying to let the kernel load these progs.

I've tested putting `btf_map_init.c`'s `tail_main()` into .text, and it turns out this check isn't tripped currently, since `loadProgram()` operates on `cs.Programs`, which .text is not allowed to emit programs into. Maybe in a future patch it will, or into a `cs.Functions` field, etc. but in the meantime I think it doesn't hurt to have the check there.

PR #532, opened 2021-12-20
Eliminate binary.Read in instruction (un)marshaler

ti-mo:
This still needs to be fuzzed, but the gains are nice so far:

```
name                    old time/op    new time/op    delta
Read64bitImmediate-16      479ns ±12%      74ns ± 1%  -84.62%  (p=0.008 n=5+5)
Write64BitImmediate-16     417ns ±10%      63ns ± 2%  -84.97%  (p=0.008 n=5+5)

name                    old alloc/op   new alloc/op   delta
Read64bitImmediate-16      32.0B ± 0%      8.0B ± 0%  -75.00%  (p=0.008 n=5+5)
Write64BitImmediate-16     24.0B ± 0%      8.0B ± 0%  -66.67%  (p=0.008 n=5+5)

name                    old allocs/op  new allocs/op  delta
Read64bitImmediate-16       4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.008 n=5+5)
Write64BitImmediate-16      3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.008 n=5+5)
```

The use of `internal.DiscardZeroes{}` was considered for checking for `64bit immediate has non-zero fields`, but ended up being a lot slower than just including those bits in the read. The fact that it aligns with a single u32 interpretation is quite convenient. :)

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -47,33 +47,45 @@ func (ins Instruction) Sym(name string) Instruction {
 
 // Unmarshal decodes a BPF instruction.
 func (ins *Instruction) Unmarshal(r io.Reader, bo binary.ByteOrder) (uint64, error) {
-	var bi bpfInstruction
-	err := binary.Read(r, bo, &bi)
-	if err != nil {
+	data := make([]byte, InstructionSize)
+	if _, err := io.ReadFull(r, data); err != nil {
 		return 0, err
 	}
 
-	ins.OpCode = bi.OpCode
-	ins.Offset = bi.Offset
-	ins.Constant = int64(bi.Constant)
-	ins.Dst, ins.Src, err = bi.Registers.Unmarshal(bo)
-	if err != nil {
-		return 0, fmt.Errorf("can't unmarshal registers: %s", err)
+	ins.OpCode = OpCode(data[0])
+
+	regs := data[1]
+	switch bo {
+	case binary.LittleEndian:
+		ins.Dst, ins.Src = Register(regs&0xF), Register(regs>>4)
+	case binary.BigEndian:
+		ins.Dst, ins.Src = Register(regs>>4), Register(regs&0xf)
 	}
 
-	if !bi.OpCode.IsDWordLoad() {
+	ins.Offset = int16(bo.Uint16(data[2:4]))
+	// Convert to int32 before widening to int64
+	// to ensure the signed bit is carried over.
+	ins.Constant = int64(int32(bo.Uint32(data[4:8])))
+
+	if !ins.OpCode.IsDWordLoad() {
 		return InstructionSize, nil
 	}
 
-	var bi2 bpfInstruction
-	if err := binary.Read(r, bo, &bi2); err != nil {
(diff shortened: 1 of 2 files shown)


asm/instruction.go
```diff
@@ -47,33 +47,46 @@ func (ins Instruction) Sym(name string) Instruction {
...
+
+	if !ins.OpCode.IsDWordLoad() {
 		return InstructionSize, nil
 	}
 
-	var bi2 bpfInstruction
-	if err := binary.Read(r, bo, &bi2); err != nil {
+	// Pull another instruction from the stream to retrieve the second
+	// half of the 64-bit immediate value.
+	data = make([]byte, InstructionSize)
```
brycekahle, line 76: Do you need the second `make` here? You are going to fill the entire slice since you are using `io.ReadFull`.

ti-mo: Thanks, not sure how I missed that.

asm/instruction.go
```diff
@@ -97,26 +109,22 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
...
 	if !isDWordLoad {
 		return InstructionSize, nil
 	}
 
-	bpfi = bpfInstruction{
-		Constant: int32(ins.Constant >> 32),
-	}
-
-	if err := binary.Write(w, bo, &bpfi); err != nil {
+	data = make([]byte, InstructionSize)
```
lmb, line 125: Idea: re-use the existing allocation or allocate 2*InstructionSize above?

asm/instruction.go
```diff
@@ -527,13 +535,6 @@ func (iter *InstructionIterator) Next() bool {
...
 }
 
-type bpfInstruction struct {
-	OpCode    OpCode
-	Registers bpfRegisters
-	Offset    int16
-	Constant  int32
-}
-
 type bpfRegisters uint8
```
lmb, line 538: Is this still used?

asm/instruction_test.go
```diff
@@ -48,6 +60,21 @@ func TestWrite64bitImmediate(t *testing.T) {
...
+func BenchmarkWrite64BitImmediate(b *testing.B) {
+	insns := Instructions{
+		LoadImm(R0, math.MinInt32-1, DWord),
+	}
+
+	var buf bytes.Buffer
+	for i := 0; i < b.N; i++ {
+		buf.Reset()
+
+		if err := insns.Marshal(&buf, binary.LittleEndian); err != nil {
```
lmb, line 72: Nit: why not benchmark Instruction.Marshal directly?

lmb (approved):
Nice!

asm/instruction.go
ti-mo: Good catch, I think zeroing out the first 4 bytes is the way to go.

asm/instruction_test.go
ti-mo: Not sure either, fixed.

PR #534, opened 2021-12-22
Remove BTF type IDs on map creation if not supported 

chenhengqi:
In libbpf commit 962f24137978 ([0]), we teach libbpf to recognize
BPF maps that do not support BTF-defined key/value and remove
BTF type IDs quietly on map creation. Make ebpf library support
this feature.

  [0]: https://github.com/libbpf/libbpf/commit/962f24137978bf3321bed97785bae59b5a3f3b06

### DIFF
--- a/map.go
+++ b/map.go
@@ -102,6 +102,12 @@ func (ms *MapSpec) Copy() *MapSpec {
 	return &cpy
 }
 
+// hasBTF returns true if the MapSpec has a valid BTF spec and if its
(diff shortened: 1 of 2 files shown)


ti-mo:
Hi @chenhengqi, thanks for the patch! I've pushed a small helper to your branch to simplify the logic at the end of `createMap()` and renamed the mt helper to `hasBTF()`.

PR #536, opened 2021-12-27
info: Use unix.ByteSliceToString instead of internal.CString

ti-mo:
Also document availability of tag and name fields somewhere it's not lost on the user.


lmb:
LGTM

PR #537, opened 2021-12-27
prog: default Program.Test() 'in' argument to 14-byte slice

ti-mo:
As a shorthand for providing an empty input, allow nil to be specified.
This seems to be the common case in the test suite.

---

I haven't been able to pinpoint where exactly the 14-byte requirement comes from within the kernel. Sure, it's the length of an ethernet header, but I'd like to be able to document why it's a requirement for tracing progs etc. that don't expect to operate on network frames/packets.

### DIFF
--- a/collection_test.go
+++ b/collection_test.go
@@ -145,7 +145,7 @@ func TestCollectionSpecRewriteMaps(t *testing.T) {
 	}
 	defer coll.Close()
 
-	ret, _, err := coll.Programs["test-prog"].Test(make([]byte, 14))
+	ret, _, err := coll.Programs["test-prog"].Test(nil)
 	testutils.SkipIfNotSupported(t, err)
 	if err != nil {
 		t.Fatal(err)
--- a/elf_reader_test.go
+++ b/elf_reader_test.go
@@ -211,7 +211,7 @@ func TestLoadCollectionSpec(t *testing.T) {
 		}
 		defer coll.Close()
 
-		ret, _, err := coll.Programs["xdp_prog"].Test(make([]byte, 14))
+		ret, _, err := coll.Programs["xdp_prog"].Test(nil)
 		if err != nil {
 			t.Fatal("Can't run program:", err)
 		}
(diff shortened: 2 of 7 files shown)


lmb:
The check is here: https://elixir.bootlin.com/linux/latest/source/net/bpf/test_run.c#L259 Don't know whether there is a good explanation.

A problem with this approach: sk_lookup programs don't allow passing in data, parameters are passed via a different argument. In this case passing the 14 byte slice will return an error. To fix this you'd have change behaviour based on the program type, which is cumbersome to maintain and not worth the effort imo.

ti-mo:
Cool, thanks for checking!

> In this case passing the 14 byte slice will return an error.

Then isn't this already broken today? https://github.com/cilium/ebpf/blob/36960939521f7570a134334bb4e69b79550a06c4/prog.go#L624-L626 Looks like passing a zero-length slice is not supported.

lmb:
Good point, that probably needs to go.

ti-mo:
There's a bit more work necessary to make testing SkLookups work, see `bpf_prog_test_run_sk_lookup()` in the kernel. (I think you even contributed that feature :slightly_smiling_face:)

`DataSizeIn/Out` and `DataIn/Out` must all be 0/NULL, but are currently unconditionally populated in `testRun()`. Also, `CtxIn/Out+Size` must be specified, which `Program.Test()` currently doesn't have a way of passing, so we'd need new API. Apparently, another such case is `RawTracepoint(Writable)`, where CtxOut, DataIn and DataOut must all be NULL.

We'd also need the Go equivalent of the kernel's `struct bpf_sk_lookup` in the `sys` package for this to be really useful.

I think merging this PR as-is makes sense, as we're looking at a different kind of API than `Program.Test()` currently provides for supporting SkLookup and RawTP. For these prog types, perhaps a `Program.TestContext()` that omits `data_in` would be more appropriate.

lmb:
> I think merging this PR as-is makes sense, as we're looking at a different kind of API than Program.Test() currently provides for supporting SkLookup and RawTP. For these prog types, perhaps a Program.TestContext() that omits data_in would be more appropriate.

Can we tackle this when we actually come up with a better BPF_PROG_RUN interface? I think we are going to tie our hands if we merge this now: any future API should (in the best case) then also allow to pass nil to mean "an empty slice of 14 bytes". I don't have enough information to say whether that is a good idea or not at the moment.

ti-mo:
Agreed, will leave it around as a draft to keep it on the radar.

lmb:
This should be handled as part of #576

Issue #444, opened 2021-10-03
retrieve instructions of loaded program


klueska:
As the title suggests -- is there an API to read the eBPF instructions from an already loaded program back out from the kernel?

I would like to be able to read the programs existing instructions, prepend some new instructions to it, and then reload these instructions over old program. Is there any way to achieve this?

klueska:
I dug a little and answered my own question -- there isn't a public API readily available to get the instructions through this package, but there is a way to get them  by making direct syscalls.

Anyone coming across this issue in the future, I wrote the following (after reading through the C code for [`bpftool`](https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/prog.c#L792) and borrowing some of the code in the `internal` package of this repo) and it is working well for my use case:

```
package internal

import (
	"fmt"
	"unsafe"

	"github.com/cilium/ebpf"
	"golang.org/x/sys/unix"
)

type ProgInfo struct {
	Prog_type                uint32
	Id                       uint32
	Tag                      [unix.BPF_TAG_SIZE]byte
	Jited_prog_len           uint32
	Xlated_prog_len          uint32
	Jited_prog_insns         Pointer
	Xlated_prog_insns        Pointer
	Load_time                uint64
	Created_by_uid           uint32
	Nr_map_ids               uint32
	Map_ids                  Pointer
	Name                     [unix.BPF_OBJ_NAME_LEN]byte
	Ifindex                  uint32
	Gpl_compatible           uint32
	Netns_dev                uint64
	Netns_ino                uint64
	Nr_jited_ksyms           uint32
	Nr_jited_func_lens       uint32
	Jited_ksyms              Pointer
	Jited_func_lens          Pointer
	Btf_id                   uint32
	Func_info_rec_size       uint32
	Func_info                Pointer
	Nr_func_info             uint32
	Nr_line_info             uint32
	Line_info                Pointer
	Jited_line_info          Pointer
	Nr_jited_line_info       uint32
	Line_info_rec_size       uint32
	Jited_line_info_rec_size uint32
	Nr_prog_tags             uint32
	Prog_tags                Pointer
	Run_time_ns              uint64
	Run_cnt                  uint64
}

type ProgInfoLinear struct {
	Prog_type                uint32
	Id                       uint32
	Tag                      [unix.BPF_TAG_SIZE]byte
	Jited_prog_len           uint32
	Xlated_prog_len          uint32
	Jited_prog_insns         []byte
[...]

PR #538, opened 2021-12-27
program: retrieve xlated Instructions from Program

ti-mo:
Allow introspecting instructions after they have been translated by and loaded into the kernel.

Just putting this up here for discussion, I remember there being some prior discussion around this a while back.

---

There's also the following cosmetic bug:
```
tail_main:
	0: LoadMapPtr dst: r2 fd: 226
```
226 is the map's BPF ID, not its fd. Not sure if this needs to be taken into account.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -309,6 +309,31 @@ func (ins Instruction) Size() uint64 {
 // Instructions is an eBPF program.
 type Instructions []Instruction
 
+// Unmarshal unmarshals an Instructions from a binary instruction stream.
+// All instructions in insns are replaced by instructions decoded from r.
+func (insns *Instructions) Unmarshal(r io.Reader, bo binary.ByteOrder) error {
+	if len(*insns) > 0 {
+		*insns = nil
+	}
+
+	var offset uint64
+	for {
+		var ins Instruction
+		n, err := ins.Unmarshal(r, bo)
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("offset %d: %w", offset, err)
+		}
+
+		*insns = append(*insns, ins)
+		offset += n
+	}
+
+	return nil
+}
+
 // Name returns the name of the function insns belongs to, if any.
 func (insns Instructions) Name() string {
 	if len(insns) == 0 {
--- a/info.go
+++ b/info.go
@@ -2,6 +2,7 @@ package ebpf
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -12,6 +13,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/btf"
 	"github.com/cilium/ebpf/internal/sys"
@@ -95,7 +97,8 @@ type ProgramInfo struct {
 	btf   btf.ID
 	stats *programStats
 
-	maps []MapID
+	maps  []MapID
+	insns []byte
 }
 
 func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
@@ -129,7 +132,13 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 		info2.MapIds = sys.NewPointer(unsafe.Pointer(&pi.maps[0]))
 	}
 
-	if info.NrMapIds > 0 {
+	if info.XlatedProgLen > 0 {
+		pi.insns = make([]byte, info.XlatedProgLen)
+		info2.XlatedProgLen = info.XlatedProgLen
+		info2.XlatedProgInsns = sys.NewSlicePointer(pi.insns)
+	}
+
+	if info.NrMapIds > 0 || info.XlatedProgLen > 0 {
 		if err := sys.ObjInfo(fd, &info2); err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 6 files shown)


prog.go
```diff
@@ -546,6 +548,52 @@ func (p *Program) Close() error {
...
 }
 
+// Instructions returns the 'translated' instruction stream of the program
+// after verification. For example, map accesses are made to reference their
+// kernel map IDs, not the FDs they had when the program was inserted.
+//
+// The first instruction is marked as a symbol using the Program's name.
+//
+// Available from 4.13. Requires CAP_BPF or equivalent.
+func (p *Program) Instructions() (asm.Instructions, error) {
```
lmb, line 558: Could the method be on `ProgramInfo` instead? We could delay unmarshaling until the function is called and only hold onto `[]byte` until then.

As you point out the returned instructions contain ids, which means it's not possible to load the program into the kernel again. The function signature kind of implies that however. Some ideas:

* Return a different type `TranslatedInstructions []asm.Instruction` to document this
* Somehow encode into the instruction that its referencing an id, not an fd (per insns metadata?). Return an error when trying to marshal that.

The latter would also allow better pretty printing which seems nice. Maybe we do both?

ti-mo: > Could the method be on `ProgramInfo` instead? We could delay unmarshaling until the function is called and only hold onto `[]byte` until then.

Yup, in retrospect, I should've put this onto `ProgramInfo` already, since two calls to `sys.ObjInfo` are necessary to fetch the actual instructions. The first call will populate `info.XlatedProgLen`, which will happen during `Program.Info()`, and the subsequent call to `ProgramInfo.Instructions()` will use the field to allocate an appropriately-sized buffer and fetch the insn stream.

> Return a different type `TranslatedInstructions []asm.Instruction` to document this

I would only go this route if aliasing the type is actually practically needed, e.g. if we want to implement extra methods on it. Not really the case here, I think.

> Somehow encode into the instruction that its referencing an id, not an fd (per insns metadata?). Return an error when trying to marshal that.

IMO this approach makes a type alias redundant, so eventually we could use this to do better pretty-prints. It won't practically solve the fact that this is surprising to the caller, whether an error is returned by the lib during marshaling or by the verifier. (although the lib will probably be friendlier :slightly_smiling_face:)

All in all, I think we can probably ignore this for now, as trying to re-insert these insns into the kernel will error out one way or another; reading the docstring will be necessary regardless.

lmb: > reading the docstring will be necessary regardless.

Ok, please make this clearer in the doc then. Right now the reader needs understand the subtleties of BPF passed to the kernel to make the connection that these insns can't be loaded.

ti-mo: Done, I've done some refactoring in `newProgramInfoFromFd` to coalesce fetching map IDs and binary insns when they're available. The decoding is deferred until `ProgramInfo.Instructions()`.

asm/instruction.go
```diff
@@ -309,6 +309,26 @@ func (ins Instruction) Size() uint64 {
 // Instructions is an eBPF program.
 type Instructions []Instruction
 
+// Unmarshal unmarshals an Instructions from a binary instruction stream.
+func (insns *Instructions) Unmarshal(r io.Reader, bo binary.ByteOrder) error {
```
lmb, line 313: Seems like calling this on non-empty `Instructions` will append to the instruction stream. Should we replace the instructions instead? (Maybe there is precedent for this behavior somewhere in the library?) If we append it should be mentioned in the doc I think.

info.go
```diff
@@ -109,33 +111,40 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
...
+	if info.NrMapIds > 0 {
+		pi.maps = make([]MapID, info.NrMapIds)
+		info2.NrMapIds = info.NrMapIds
+		info2.MapIds = sys.NewPointer(unsafe.Pointer(&pi.maps[0]))
+	}
+
+	if info.XlatedProgLen > 0 {
+		pi.insns = make([]byte, info.XlatedProgLen)
+		info2.XlatedProgLen = info.XlatedProgLen
+		info2.XlatedProgInsns = sys.NewPointer(unsafe.Pointer(&pi.insns[0]))
```
lmb, line 138: Nit: `NewSlicePointer`?

info.go
```diff
@@ -199,13 +208,42 @@ func (pi *ProgramInfo) Runtime() (time.Duration, bool) {
...
+// For example, map accesses are made to reference their kernel map IDs,
+// not the FDs they had when the program was inserted.
+//
+// The first instruction is marked as a symbol using the Program's name.
+//
+// Available from 4.13. Requires CAP_BPF or equivalent.
+func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
+	// If the calling process is not BPF-capable, the field will be empty.
+	if len(pi.insns) == 0 {
+		return nil, fmt.Errorf("getting xlated instructions: %w", os.ErrPermission)
```
lmb, line 225: This might be a bit misleading on pre 4.13? Maybe return `ErrNotSupported` instead?

prog_test.go
```diff
@@ -686,6 +686,61 @@ func TestProgramBindMap(t *testing.T) {
...
+	}
+
+	name := "test_prog"
+
+	spec := &ProgramSpec{
+		Type: SocketFilter,
+		Name: name,
+		Instructions: asm.Instructions{
+			asm.LoadImm(asm.R0, -1, asm.DWord).Sym(name),
+			asm.LoadMapPtr(asm.R1, arr.FD()),
```
lmb, line 712: If you get rid of the map pointer you don't need to do fix ups on the result.

lmb (approved):

info.go
ti-mo: True! Forgot that existed.

asm/instruction.go
ti-mo: That's not intended, the expected behaviour attached to the term 'unmarshaling' is definitely to replace.

info.go
ti-mo: Agreed, done.

klueska:
I see the comment that the instructions read out cannot be written back in. My original use case from #444 was to do exactly this (i.e. read out, prepend to, write back).

Is there someway to retrieve the original stream of instructions to accomplish what I need from what’s been added here?

ti-mo:
@klueska I think this implementation does what you asked for, at least it provides the API for you to build your use case on top of.

xlated instructions have BPF map IDs where map FDs used to be during load time. Walking the insns, calling `NewMapFromID` on all map IDs you encounter (maybe taking care you don't open the same map ID more than once) and rewriting those map IDs with the newly-opened FDs should work. There might be some other differences that I'm not aware of, but this should you most of the way there. :slightly_smiling_face:

At least now, you can also diff your ProgramSpec insns with the loaded Program insns and see what changed. Good luck!

klueska:
I have verified that moving from my custom code to this implementation works identically.

The diff is fairly small: https://gitlab.com/nvidia/container-toolkit/libnvidia-container/-/merge_requests/127/diffs?commit_id=d21ac33cb0459e9277f61622ef99fc8d8fa91ca5

PR #540, opened 2021-12-29
Support attaching SocketFilter programs to raw sockets

folbricht:
Adds support for attaching Socket Filter programs to raw sockets. It would look like this:

```go
link.SocketFilterAttachProgram(conn, objs.Filter)

link.SocketFilterDetachProgram(conn)
```

Any feedback is welcome. I'll also provide an example.

### DIFF
--- a/internal/unix/types_linux.go
+++ b/internal/unix/types_linux.go
@@ -66,6 +66,9 @@ const (
 	PERF_RECORD_SAMPLE       = linux.PERF_RECORD_SAMPLE
 	AT_FDCWD                 = linux.AT_FDCWD
 	RENAME_NOREPLACE         = linux.RENAME_NOREPLACE
+	SO_ATTACH_BPF            = linux.SO_ATTACH_BPF
+	SO_DETACH_BPF            = linux.SO_DETACH_BPF
+	SOL_SOCKET               = linux.SOL_SOCKET
 )
 
 // Statfs_t is a wrapper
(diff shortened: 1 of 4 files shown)


lmb (commented):
Hi Frank,

How are you opening the raw sockets currently? Go stdlib or via some third part package (e.g. mdlayher/raw)?

* Taking a plain `SockFD int` is problematic, since it doesn't work well with the stdlib. It invites calling `os.File.Fd()` to get the file, but that puts the socket into blocking mode. It's also problematic because it may lead to closed fd reuse. We should take https://golang.google.cn/pkg/syscall/#Conn
* We don't have to shoehorn filters into the `Link` interface, it's not possible to call most of the methods. Two free functions to attach and detach would be better i think, similar to RawAttachProgram.

folbricht:
I've been using the stdlib to open the raw socket (similar to the example above). If I understand your suggestion correctly, I should:
- Wrap the FD into a syscall.Conn which would prevent direct access to it
- Add a couple of functions to attach/detach. As for naming them, how does `RawAttachSocketFilter(syscall.Conn, *ebpf.Program)` sound?

lmb:
> Add a couple of functions to attach/detach. 

Yes, without introducing a new link type though. So "replacing linkSocketFilter with two free function" is what I'd like to do.

Can you add a test as well please?

folbricht:
Replaced the Link interface implementation with two functions, and added a simple test (which just checks that attaching/detaching) works. Using a full raw socket involves a bit more code to implement syscall.Conn but I could do that in an example.

lmb:
Renamed the free functions to be closer to other functions in the package.

lmb (approved):

lmb:
Thanks!

PR #543, opened 2022-01-18
delete useless code

ningmingxiao:
Signed-off-by: ningmingxiao <ning.mingxiao@zte.com.cn>
int64(^uint(0)>>1) is 9223372036854775807
int64(offset) max is 9223372036854775807
int64(offset) is not bigger than  int64(^uint(0)>>1)

### DIFF
--- a/internal/btf/strings.go
+++ b/internal/btf/strings.go
@@ -31,10 +31,6 @@ func readStringTable(r io.Reader) (stringTable, error) {
 }
 
 func (st stringTable) Lookup(offset uint32) (string, error) {
-	if int64(offset) > int64(^uint(0)>>1) {
-		return "", fmt.Errorf("offset %d overflows int", offset)
-	}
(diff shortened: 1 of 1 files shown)


lmb:
> int64(^uint(0)>>1) is 9223372036854775807

This is only true on 64 bit arches.

ningmingxiao:
thank you,I see
```
64 bit  os
because of  offset uint32  max is  4294967295,
int64(^uint(0)>>1) is 9223372036854775807
int64(offset) max is  4294967295
int64(offset) > int64(^uint(0)>>1)   value  always false

32  bit  os
offset uint32   max is  4294967295,
int64(^uint(0)>>1) is 2147483647
int64(offset) max is    4294967295
int64(offset) > int64(^uint(0)>>1)  value  always true
```
32bit  is difference from 64bit

PR #544, opened 2022-01-20
linker: Allow calling into unimplemented forward function declarations

### LINKED ISSUE #531: Function calls can't be linked after loading an ELF
**Describe the bug**
We have a program that makes a call to a function that is declared, but not defined in C:

```
// Declared, but not defined
int foo();

__section("socket") int link_call() {
    return foo();
}
```

The implementation of `foo()` is generated at runtime in assembly using the `asm` package.

Before https://github.com/cilium/ebpf/commit/9d739bdd0fe0fd0651ceb3e7bdcb0c9e9e4e92e1, we could append a suitable definition to `ProgramSpec.Instructions`, and `NewCollection()` would "link" it in by rewriting the call instruction offset.

**To Reproduce**
Branch `afabre/link_call` has a test to reproduce this: https://github.com/cilium/ebpf/commit/8607ff9f2c20e11a5e548e2da280bc11789bbd82

The test currently fails:

```
> sudo -E (which go) test -count=1 ./
--- FAIL: TestCollectionSpecLinkCall (0.00s)
    collection_test.go:162: Can't parse ELF: file testdata/link_call-el.elf: load programs: populating references: finding references: symbol reference foo not present in progs
```

But the test passes using the commit before https://github.com/cilium/ebpf/commit/9d739bdd0fe0fd0651ceb3e7bdcb0c9e9e4e92e1:

```
> git checkout 9d739bdd0fe0fd0651ceb3e7bdcb0c9e9e4e92e1~1
> git cherry-pick 8607ff9f2c20e11a5e548e2da280bc11789bbd82
> sudo -E (which go) test -count=1 ./
ok      github.com/cilium/ebpf  0.524s
```

We don't have to settle on this exact API, happy to work on a different way of doing things (both from the C / clang side and the go side).

ti-mo:
After #508, the linker became more restrictive regarding jumps to symbols that don't appear in the same section, as these usually mean a function in another section is called. One notable exception to this rule is forward function declarations that are left unimplemented.

This patch adds the possibility for the caller to provide an implementation for these symbols after the CollectionSpec has been emitted. Instead, throw an error when the ProgramSpec is instantiated into a Program.

Fixes #531 

@arthurfabre Please check if this works as expected for your production code.


linker.go
```diff
@@ -164,7 +166,7 @@ func fixupJumpsAndCalls(insns asm.Instructions) error {
 			continue
 		}
 
-		return fmt.Errorf("%s at %d: reference to missing symbol %q", ins.OpCode, i, ins.Reference)
+		return fmt.Errorf("%s at insn %d: symbol %q: %w", ins.OpCode, i, ins.Reference, errUnsatisfiedReference)
```
lmb, line 169: I think this could lead to a somewhat confusing error message: https://github.com/cilium/ebpf/blob/1946eefb4ab804a7d40b5d0d65ddd4158be73979/prog.go#L233-L235

Use a different error?

lmb (approved):
Seems ok to me, not sure how the changes you made make it work though :D

ti-mo:
> not sure how the changes you made make it work though

`references` is only really used in `flatten()`, where it's iterated over, not looked up in. This results in the fwd function not being concatenated to the full insn stream (since there is no ProgramSpec and thus no body), so `fixupJumpsAndCalls` will find a jump without a target at load time.

linker.go
ti-mo: Done.

PR #545, opened 2022-01-21
elf_reader_test: enable test_sk_assign.o, handle iproute2 map

### LINKED ISSUE #469: Enable `test_sk_assign.o` selftest
Non-standard non-BTF map definitions can be loaded since https://github.com/cilium/ebpf/pull/362, so this can technically be enabled.

Some extra glue might be needed for parsing the extra bytes, but worth it to get extra test coverage IMO.

ti-mo:
Fixes #469

### DIFF
--- a/elf_reader_test.go
+++ b/elf_reader_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


elf_reader_test.go
```diff
@@ -674,6 +673,18 @@ func TestLibBPFCompat(t *testing.T) {
...
 
+		switch file {
+		case "test_sk_assign.o":
+			// Test contains a legacy iproute2 bpf_elf_map definition.
+			for _, m := range spec.Maps {
+				if m.Extra.Len() == 0 {
+					t.Fatalf("Expected extra bytes in map %s", m.Name)
+				}
+				_, _ = io.Copy(io.Discard, &m.Extra)
+				m.Pinning = PinByName
```
lmb, line 684: Is pinning required for the test to function?

lmb (approved):

elf_reader_test.go
ti-mo: Nope, removed.

PR #546, opened 2022-01-21
map: make Map.Extra field a pointer to a bytes.Reader

ti-mo:
This avoids allocating a bytes.Reader for every MapSpec, as its presence is rather uncommon.

This is a breaking change that needs to be documented in the release.


lmb (approved):
Fixed up the documentation of MapSpec.Extra.

PR #549, opened 2022-01-25
elf_reader: Fix non-determinism in determining prog type

joamaki:
The section prefix to types mapping was stored in a map and the type
was found by iterating over the map. Since map iteration order is
non-deterministic and there's overlapping prefixes (e.g. cgroup/sock
and cgroup/sock_create), this sometimes caused the wrong type to
be picked.

Fix this by using a slice instead and thus respecting the order in
which the types were specified.


ti-mo (approved):
Thanks!

PR #552, opened 2022-01-28
bpf2go: generate Go types from BTF

lmb:
This PR has been sitting in my repo for too long. The basic idea is to use BTF to generate Go type equivalents of C types. By default we will do this for all types referenced as `key` or `value` from BTF map definitions.

Adapting examples would look like this: https://github.com/lmb/ebpf/commit/1ce17d8aaa820f1df1530b367b417672ff62be6f

cmd/bpf2go: generate types from BTF
    
    Use GoFormatter to generate type definitions for types used as map keys
    and values. Allow users to emit additional types via a command line flag.

collection: expose BTF spec

btf: export UnderlyingType
    
    Export an infallible version of skipQualifiersAndTypedef. Circular
    types / very long chains of qualifiers or typedefs are a corner
    case that shouldn't really happen, so let's not burden the callers
    with this detail.

btf: allow finding type by name alone
    
    Add a function which allows finding a type by name alone. This
    is useful for code that doesn't care about the concrete type
    returned.


danobi:
FWIW, I think it'd be really useful to also generate BTF_KIND_DATASEC types as well. Maybe not in this PR since this library doesn't support reading globals from collections yet.

I added similar stuff in libbpf-rs and it was quite useful there. I miss it a lot in go land :)

lmb:
Yes, definitely. The code can already emit Datasec, but I don't yet know how to expose a sensible API for that.

ti-mo:
Neat! Looks like you found a use for `AnyTypeByName` after all. :smile:

ti-mo (approved):

PR #555, opened 2022-02-01
btf: remove symbol value bounds check to support vmlinux ELFs

ti-mo:
In vmlinux ELFs, symbol values are (virtual) memory addresses and don't necessarily correspond to offsets within ELF sections or even files.

For example:

```
~ llvm-objdump -t /usr/lib/modules/*/build/vmlinux | head -5

/usr/lib/modules/5.15.17-1-lts/build/vmlinux:	file format elf64-x86-64

SYMBOL TABLE:
ffffffff81000000 l    d  .text	0000000000000000 .text
```

Explanation on why this is the case: https://0xax.gitbooks.io/linux-insides/content/Theory/linux-theory-2.html#vmlinux

Remove the check as the MaxU32 assumption no longer holds. This allows the caller to specify an unstripped vmlinux TargetBTF and would result in correct fallback behaviour to one of the files in /boot, /lib and /usr as specified in loadKernelSpec(), which is currently also broken as a result.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -90,6 +90,17 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 	}
 	defer file.Close()
 
+	return loadSpecFromELF(file)
+}
+
+// variableOffsets extracts all symbols offsets from an ELF and indexes them by
+// section and variable name.
+//
+// References to variables in BTF data sections carry unsigned 32-bit offsets.
+// Some ELF symbols (e.g. in vmlinux) may point to virtual memory that is well
+// beyond this range. Since these symbols cannot be described by BTF info,
+// ignore them here.
+func variableOffsets(file *internal.SafeELFFile) (map[variable]uint32, error) {
 	symbols, err := file.Symbols()
 	if err != nil {
 		return nil, fmt.Errorf("can't read symbols: %v", err)
@@ -115,10 +126,10 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 		variableOffsets[variable{secName, symbol.Name}] = uint32(symbol.Value)
 	}
 
-	return loadSpecFromELF(file, variableOffsets)
+	return variableOffsets, nil
 }
 
-func loadSpecFromELF(file *internal.SafeELFFile, variableOffsets map[variable]uint32) (*Spec, error) {
+func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
 	var (
 		btfSection    *elf.Section
 		btfExtSection *elf.Section
@@ -148,7 +159,12 @@ func loadSpecFromELF(file *internal.SafeELFFile, variableOffsets map[variable]ui
 		return nil, fmt.Errorf("btf: %w", ErrNotFound)
 	}
 
-	spec, err := loadRawSpec(btfSection.Open(), file.ByteOrder, sectionSizes, variableOffsets)
+	vars, err := variableOffsets(file)
+	if err != nil {
+		return nil, err
+	}
+
+	spec, err := loadRawSpec(btfSection.Open(), file.ByteOrder, sectionSizes, vars)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


internal/btf/btf.go
```diff
@@ -107,9 +107,6 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 		}
 
 		secName := file.Sections[symbol.Section].Name
-		if symbol.Value > math.MaxUint32 {
-			return nil, fmt.Errorf("section %s: symbol %s: size exceeds maximum", secName, symbol.Name)
-		}
 
 		variableOffsets[variable{secName, symbol.Name}] = uint32(symbol.Value)
```
lmb, line 111: In that case variable offsets should also be 64 bit? With your change we'd just silently truncate.

lmb:
Is there a way to add a test for this?

internal/btf/btf.go
joamaki: Yep was wondering about this. As a datapoint, at least on the vmlinux I have here the symbol addresses are either of the form `ffffffffXXXXXXXX`or `00000000XXXXXXXX`: 
```
155662: ffffffff81968bc0    23 FUNC    GLOBAL DEFAULT     1 dm_target_exit
155663: ffffffff831546b8     0 NOTYPE  GLOBAL DEFAULT    30 __stop___trace_bprintk_fmt
155664: 000000009aaeefce     0 NOTYPE  GLOBAL DEFAULT   ABS __crc_sysctl_nf_log_all_netns
155665: 00000000949b5176     0 NOTYPE  GLOBAL DEFAULT   ABS __crc_devlink_region_snapshot_create
155666: ffffffff810b5450   143 FUNC    GLOBAL DEFAULT     1 __x64_sys_ssetmask
155667: 000000004456717a     0 NOTYPE  GLOBAL DEFAULT   ABS __crc_devlink_trap_policers_register
155668: ffffffff831004e0   432 OBJECT  GLOBAL DEFAULT    30 ping_prot
155669: ffffffff81703420    33 FUNC    GLOBAL DEFAULT     1 clk_hw_unregister_gate
155670: ffffffff81317320   599 FUNC    GLOBAL DEFAULT     1 zbud_reclaim_page
```

Whether that means we can ignore the first 32 bits or not, I don't know :shrug:

ti-mo:
Loading any vmlinux ELF will do, will try to come up with something.

internal/btf/btf.go
ti-mo: Okay, taking another look at this, this is indeed not a good idea. Truncation would cause symbols `ffffffff81317320` and `0000000081317320` to have the same offset. Not an issue in practice, because:
- symbol values aren't used as an index, and lookups are always performed based on sec/symbol name.
- BTF datasec vars have 32-bit offsets, so symbols sitting in the `ffffffff00000000` range simply cannot be encoded

My vmlinux seems to confirm this. There is only one datasec:
```
~ bpftool btf dump file /usr/lib/modules/5.16.2-zen1-1-zen/build/vmlinux
...
[107260] DATASEC '.data..percpu' size=209156 vlen=380
        type_id=5390 offset=4096 size=4096 (VAR 'cpu_debug_store')
...
```

```
~ llvm-readelf -s /usr/lib/modules/5.16.2-zen1-1-zen/build/vmlinux | grep cpu_debug_store
...
142762: 0000000000001000  4096 OBJECT  GLOBAL DEFAULT    37 cpu_debug_store
```

Since vmlinux `.text` sits in `ffffffff81000000` (and is not a datasec), there cannot be any BTF references to it.

Because of all this, I think the straightforward solution would be to skip any symbols whose addresses overflow u32, since they cannot be referenced by BTF.

internal/btf/btf.go
```diff
@@ -102,22 +108,26 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 			continue
 		}
 
+		if symbol.Value > math.MaxUint32 {
+			// References to variables in BTF Datasecs carry 32-bit offsets
```
lmb, line 112: Put that in the function doc instead?

lmb (approved):

joamaki (approved):

PR #560, opened 2022-02-04
Look for Linux note in all elf note sections instead of just the first one

paulcacheux:
When looking for the Linux version, we currently look for the "Linux" note in the vdso.

For strange and unknown reasons, on some kernels (specially on a Container Optimized OS v85 with a 5.4.x kernel), the `.rodata` can also have an elf section type of `SHT_NOTE`.

This means that only looking at the first note section can result in just looking at the wrong section.

This PR fixes this issue by looking at all the note sections.


lmb:
Thanks for your PR! Could you dump the vdso that is causing the problem into `testdata` and extend `TestLinuxVersionCodeEmbedded`?

paulcacheux:
Good point, I will do that

lmb (approved):
Thanks!

ti-mo:
Thanks!

lmb:
We don't have a specific release cycle. Since there are only bugfixes on master I tagged 0.8.1 for you.

PR #562, opened 2022-02-09
CI: try to fix error 42

lmb:
We get somewhat regular "exit status 42" failures on CI. This happens
when a file indicating success of the test doesn't exist:

    if [[ ! -e "${output}/success" ]]; then
      exit 42
    fi

Weirdly, error 42 always happens without any output from the Go test
runner itself. I have a theory that maybe it's not the test that fails,
but that creating the success file doesn't complete before the VM
stops. Try to work around this by syncing the filesystem which the
success file resides on before exiting.


lmb:
We have a workaround in place, and upstream doesn't seem responsive.

Issue #522, opened 2021-12-09
Removal of `btf.Program`


ti-mo:
Currently, a program's BTF info is kept in a separate structure, pointed to by `ebpf.ProgramSpec`'s `BTF` field. Now the use case for BTF has grown as substantially as it did, it's time to merge this data structure into the core types for ease of maintenance.

Having this in a separate structure is both convenient (checking BTF presence is checking a single value) and inconvenient in some ways, but overall it's a burden to have the data spread out between two structs, especially when it comes to marshaling BTF info back into binary form.

My initial idea is to move things around as follows:

- FuncInfo: remove in favor of a `btf.TypeID` field in `ebpf.ProgramSpec`
- LineInfos: resolve file and lineinfo into strings, move to `ebpf.ProgramSpec`, exported (?)
- CoreRelos: not quite sure what to do with this yet, could be moved and unexported initially

https://github.com/cilium/ebpf/pull/508 already made some small steps to remove the need for `FuncInfo` and `LineInfo` to resemble their wire formats so closely. Ultimately, we should end up with something like `[]*ProgramSpec.MarshalFuncInfos()`.

More suggestions:
- https://github.com/cilium/ebpf/pull/508#discussion_r760258743: export `coreRelocate` and get rid of `Program`?
- https://github.com/cilium/ebpf/pull/508#discussion_r760271764: Why not have `FuncInfos.Marshal`?
- https://github.com/cilium/ebpf/pull/508#discussion_r760269474: Store `Type *Func` instead of bare TypeID

lmb:
We could solve many of these with flexible metadata in `asm.Instruction`, then concatenation of `asm.Instructions` does the right thing by default. There is a proof of concept for per insn metadata in 483ca4a65c08a117afcd25cc65e5407a27972762, but I think a real implementation needs to be more flexible, `instructionMeta` should probably be a `map[interface{}]interface{}` and work similarly to context.Value. We'd also need copy on write semantics. With that in place:

* `LineInfos` could become metadata on `asm.Instruction` pretty easily.
* `FuncInfo` could also become per insn metadata, but this is less straighforward.
* `CoreRelos` could also become per insn metadata, but we'd still have to collect all relos into a slice, see `coreRelocate`.
* We could solve #515 and #517 by adding metadata that references a `*Map`. This could also allow `NewProgram(CollectionSpec.Programs["foo"])` to resolve maps, which currently returns an error.

[...]

PR #567, opened 2022-02-13
Add add sparse per-instruction metadata 

### LINKED ISSUE #515: Map file descriptor is closed after `RewriteMaps`
**Describe the bug**

The file descriptor of an eBPF map uses in `RewriteMaps` is closed before loading the program, causing a failure when loading the program. 

**To Reproduce**

I prepared a repository with the reproducer of this problem -> https://github.com/mauriciovasquezbernal/ebpfissue. 

```
$ git clone https://github.com/mauriciovasquezbernal/ebpfissue
...
$ cd ebpfissue
$ make 
...
$ sudo bpftool map create /sys/fs/bpf/mount_ns_set type hash key 8 value 4 entries 1024 name mount_ns_set
...
$ sudo ./foo
2021/12/02 18:34:34 loading objects: field VfsReadEntry: program vfs_read_entry: load program: invalid argument: fd 3 is not pointing to valid bpf_map
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 
```

I see the problem is that the file descriptor of the map is being closed before loading the program:

```
$ sudo strace -f -e signal=SIGINT -e trace=close,bpf ./foo
close(3)                                = 0
strace: Process 466748 attached
strace: Process 466749 attached
strace: Process 466750 attached
strace: Process 466751 attached
[pid 466747] bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=4, max_entries=1, map_flags=0, inner_map_fd=0, map_name="", map_ifindex=0, btf_fd=0, btf_key_type_id=0, btf_value_type_id=0}, 60) = 3
[pid 466747] close(3)                   = 0
[pid 466747] bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=4, max_entries=1, map_flags=0, inner_map_fd=0, map_name="feature_test", map_ifindex=0, btf_fd=0, btf_key_type_id=0, btf_value_type_id=0}, 60) = 7
[pid 466747] close(7)                   = 0
[pid 466747] bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=4, max_entries=1, map_flags=0, inner_map_fd=0, map_name=".test", map_ifindex=0, btf_fd=0, btf_key_type_id=0, btf_value_type_id=0}, 60) = 7
[pid 466747] close(7)                   = 0
[pid 466747] close(3)                   = 0

---> fd 3 is returned here

[pid 466747] bpf(BPF_OBJ_GET, {pathname="/sys/fs/bpf/mount_ns_set", bpf_fd=0, file_flags=0}, 16) = 3
[pid 466747] bpf(BPF_OBJ_GET_INFO_BY_FD, {info={bpf_fd=3, info_len=80, info=0xc00001c320}}, 16) = 0
[pid 466747] close(7)                   = 0
strace: Process 466752 attached
strace: Process 466753 attached

---> but it's closed here before BPF_PROG_LOAD!!!

[pid 466752] close(3)                   = 0
strace: Process 466754 attached
strace: Process 466755 attached
strace: Process 466756 attached
[pid 466755] close(7)                   = 0
[pid 466755] bpf(BPF_BTF_LOAD, {btf="\237\353\1\0\30\0\0\0\0\0\0\0\34\0\0\0\34\0\0\0\3\0\0\0\0\0\0\0\0\0\0\2"..., btf_log_buf=NULL, btf_size=55, btf_log_size=0, btf_log_level=0}, 32) = 3
[pid 466755] close(3)                   = 0
[pid 466755] bpf(BPF_BTF_LOAD, {btf="\237\353\1\0\30\0\0\0\0\0\0\0\30\0\0\0\30\0\0\0\3\0\0\0\0\0\0\0\0\0\0\r"..., btf_log_buf=NULL, btf_size=51, btf_log_size=0, btf_log_level=0}, 32) = 3
[pid 466755] close(3)                   = 0
[pid 466755] bpf(BPF_BTF_LOAD, {btf="\237\353\1\0\30\0\0\0\0\0\0\0\244D\0\0\244D\0\0\373-\0\0\0\0\0\0\0\0\0\2"..., btf_log_buf=NULL, btf_size=29367, btf_log_size=0, btf_log_level=0}, 32) = 3
[pid 466755] bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_KPROBE, insn_cnt=6, insns=0xc001011590, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(0, 0, 0), prog_flags=0, prog_name="", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0}, 120) = 7
[pid 466755] close(7)                   = 0
[pid 466755] bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_KPROBE, insn_cnt=33, insns=0xc000c90240, license="Dual BSD/GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(5, 11, 22), prog_flags=0, prog_name="vfs_read_entry", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=3, func_info_rec_size=8, func_info=0xc0026e2a78, func_info_cnt=1, line_info_rec_size=16, line_info=0xc000ae8540, line_info_cnt=12, attach_btf_id=0, attach_prog_fd=0}, 120) = -1 EINVAL (Invalid argument)
[pid 466755] bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_KPROBE, insn_cnt=33, insns=0xc000c90240, license="Dual BSD/GPL", log_level=1, log_size=65536, log_buf="", kern_version=KERNEL_VERSION(5, 11, 22), prog_flags=0, prog_name="vfs_read_entry", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=3, func_info_rec_size=8, func_info=0xc0026e2a78, func_info_cnt=1, line_info_rec_size=16, line_info=0xc000ae8540, line_info_cnt=12, attach_btf_id=0, attach_prog_fd=0}, 120) = -1 EINVAL (Invalid argument)
[pid 466755] close(3)                   = 0
[pid 466755] close(3)                   = 0
2021/12/02 18:34:44 loading objects: field VfsReadEntry: program vfs_read_entry: load program: invalid argument: fd 3 is not pointing to valid bpf_map
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
[pid 466756] +++ exited with 1 +++
[pid 466755] +++ exited with 1 +++
[pid 466752] +++ exited with 1 +++
[pid 466754] +++ exited with 1 +++
[pid 466749] +++ exited with 1 +++
[pid 466753] +++ exited with 1 +++
[pid 466748] +++ exited with 1 +++
[pid 466751] +++ exited with 1 +++
[pid 466750] +++ exited with 1 +++
+++ exited with 1 +++
```

It's somehow related to the garbage collection, if I comment out the following line it works fine:

https://github.com/cilium/ebpf/blob/f40ce35c95205b0acb530b281495a9e2357f0a86/internal/sys/fd.go#L21

Two additional details:

1. It only happens (in this case) when there is a core relocation. (commenting out https://github.com/mauriciovasquezbernal/ebpfissue/blob/05a7074afb97b833836bc02e7c0e8df79d54a888/bpf/foo.bpf.c#L9 makes it work)
2. Doesn't happen when the map is created instead of loaded from a pinned one (https://github.com/mauriciovasquezbernal/ebpfissue/blob/05a7074afb97b833836bc02e7c0e8df79d54a888/foo.go#L33)

My environment is:

```sh
$ go version
go version go1.16.3 linux/amd64
$ uname -r
5.11.0-41-generic
uname -a
$ clang --version
Ubuntu clang version 12.0.0-3ubuntu1~20.04.4
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

I hope I gave enough information to track this down. 

/cc @alban

dylandreimerink:
This PR improves the existing per-instruction metadata, as discussed in #564 and fixes #515.

This PR moves the existing `Instruction.Reference` and `Instruction.Symbol` metadata to a separate struct called `metadata`. `asm.Instruction` now has a pointer to this `metadata` struct. The `Instruction` struct now has exported getters and setters to get and modify this metadata. The design is such that we only assign metadata to instructions if we have to, since most instructions don't have metadata.

Copying instructions will not automatically copy the metadata since `Instruction` has a pointer to metadata. Therefor the setters will perform a copy-on-write, so upon the first modification of metadata, the instruction will get a new metadata object which is copied from the existing one and then modified. 

These setters all have value receivers which return the modified instruction. This choice was made because it makes it easier to use the setters in the DSL, since you can't call functions with pointer receivers on literals. By returning a modified instruction we make method chaining possible, which fits well with the rest of the DSL.

We also added Filename, Line and LineCol to the `metadata` and getters and setters for them to `Instruction`. The ELF reader was modified to add looked-up `LineInfo` from the .BTF.ext section to the program instructions. This allows us to print the original source code info along with the program instructions when formatting `asm.Instructions`. 

By attaching this info the the instruction makes it relocation-proof. In future PRs we might even be able to marshal blobs to be loaded into the kernel so would guarantee correct BTF line info for relocated programs and BTF line info for programs created via the DSL.

Lastly, this PR adds map references to the per-instruction metadata. This allows us to add references to `ebpf.Map` to the program instructions themselves, specifically during rewriting, which prevents the maps from being garbage collected, thus fixing #515

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"strings"
 
+	"github.com/cilium/ebpf/internal/sys"
 	"github.com/cilium/ebpf/internal/unix"
 )
 
@@ -19,6 +20,7 @@ const InstructionSize = 8
 // RawInstructionOffset is an offset in units of raw BPF instructions.
 type RawInstructionOffset uint64
 
+var ErrUnreferencedSymbol = errors.New("unreferenced symbol")
 var ErrUnsatisfiedMapReference = errors.New("unsatisfied map reference")
 var ErrUnsatisfiedProgramReference = errors.New("unsatisfied program reference")
 
@@ -127,31 +129,67 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 	return 2 * InstructionSize, nil
 }
 
-// RewriteMapPtr changes an instruction to use a new map fd.
+// AssociateMap associates a Map with this Instruction.
 //
-// Returns an error if the instruction doesn't load a map.
-func (ins *Instruction) RewriteMapPtr(fd int) error {
-	if !ins.OpCode.IsDWordLoad() {
-		return fmt.Errorf("%s is not a 64 bit load", ins.OpCode)
-	}
-
-	if ins.Src != PseudoMapFD && ins.Src != PseudoMapValue {
+// Implicitly clears the Instruction's Reference field.
+//
+// Returns an error if the Instruction is not a map load.
+func (ins *Instruction) AssociateMap(m FDer) error {
+	if !ins.IsLoadFromMap() {
 		return errors.New("not a load from a map")
 	}
 
+	ins.setMap(m)
+
+	// Remove any Reference as it has now been resolved.
+	*ins = ins.WithReference("")
+
+	return nil
+}
+
+// RewriteMapPtr changes an instruction to use a new map fd.
+//
+// Returns an error if the instruction doesn't load a map.
+//
+// Deprecated: use AssociateMap instead. If you cannot provide a Map,
+// wrap an fd in a type implementing FDer.
+func (ins *Instruction) RewriteMapPtr(fd int) error {
+	if !ins.IsLoadFromMap() {
+		return errors.New("not a load from a map")
+	}
+
+	ins.encodeMapFD(fd)
+
+	return nil
+}
+
+func (ins *Instruction) encodeMapFD(fd int) {
 	// Preserve the offset value for direct map loads.
 	offset := uint64(ins.Constant) & (math.MaxUint32 << 32)
 	rawFd := uint64(uint32(fd))
 	ins.Constant = int64(offset | rawFd)
-	return nil
 }
 
 // MapPtr returns the map fd for this instruction.
 //
 // The result is undefined if the instruction is not a load from a map,
 // see IsLoadFromMap.
+//
+// Deprecated: use Map() instead.
 func (ins *Instruction) MapPtr() int {
-	return int(int32(uint64(ins.Constant) & math.MaxUint32))
+	// If there is a map associated with the instruction, return its FD.
+	if ins.metadata != nil && ins.metadata.bpfMap != nil {
+		return ins.metadata.bpfMap.FD()
+	}
+
+	// Fall back to the fd stored in the Constant field
+	return ins.mapFd()
+}
+
+// mapFd returns the map file descriptor stored in the 32 least significant
+// bits of ins' Constant field.
+func (ins *Instruction) mapFd() int {
+	return int(int32(ins.Constant))
 }
 
 // RewriteMapOffset changes the offset of a direct load from a map.
@@ -233,13 +271,22 @@ func (ins Instruction) Format(f fmt.State, c rune) {
 	}
 
 	if ins.IsLoadFromMap() {
-		fd := ins.MapPtr()
+		fd := ins.mapFd()
+		m := ins.Map()
 		switch ins.Src {
 		case PseudoMapFD:
-			fmt.Fprintf(f, "LoadMapPtr dst: %s fd: %d", ins.Dst, fd)
+			if m != nil {
+				fmt.Fprintf(f, "LoadMapPtr dst: %s map: %s", ins.Dst, m)
+			} else {
+				fmt.Fprintf(f, "LoadMapPtr dst: %s fd: %d", ins.Dst, fd)
+			}
 
 		case PseudoMapValue:
-			fmt.Fprintf(f, "LoadMapValue dst: %s, fd: %d off: %d", ins.Dst, fd, ins.mapOffset())
+			if m != nil {
+				fmt.Fprintf(f, "LoadMapValue dst: %s, map: %s off: %d", ins.Dst, m, ins.mapOffset())
+			} else {
+				fmt.Fprintf(f, "LoadMapValue dst: %s, fd: %d off: %d", ins.Dst, fd, ins.mapOffset())
+			}
 		}
 
 		goto ref
@@ -348,6 +395,13 @@ func (ins Instruction) Reference() string {
 	return ins.metadata.Reference()
 }
 
+// Map returns the Map referenced by ins, if any.
+// An Instruction will contain a Map if e.g. it references an existing,
+// pinned map that was opened during ELF loading.
+func (ins Instruction) Map() FDer {
+	return ins.metadata.Map()
+}
+
 // copyMetadata is a convenience method for copying ins.metadata, assigning
 // the new copy to its metadata field and returning a pointer to the copy
 // so one access can be chained.
@@ -356,12 +410,31 @@ func (ins *Instruction) copyMetadata() *metadata {
 	return ins.metadata
 }
 
+// setMap sets the given Map m in the metadata of this instruction.
+func (ins *Instruction) setMap(m FDer) {
+	if ins.metadata.Map() == m {
+		return
+	}
+
+	ins.copyMetadata().bpfMap = m
+}
+
+// FDer represents a resource tied to an underlying file descriptor.
+// Used as a stand-in for e.g. ebpf.Map since that type cannot be
+// imported here and FD() is the only method we rely on.
+type FDer interface {
+	FD() int
+}
+
 // metadata holds metadata about an Instruction.
 type metadata struct {
 	// reference denotes a reference (e.g. a jump) to another symbol.
 	reference string
 	// symbol denotes an instruction at the start of a function body.
 	symbol string
+
+	// bpfMap denotes the Map whose fd is used by this instruction.
+	bpfMap FDer
 }
 
 // Reference is a safe accessor to metadata's reference field.
@@ -380,6 +453,15 @@ func (m *metadata) Symbol() string {
 		return ""
 	}
 	return m.symbol
+}
+
+// Map is a safe accessor to metadata's bpfMap field.
+// It can be called on a nil m, in which case it will return the default value.
+func (m *metadata) Map() FDer {
+	if m == nil {
+		return nil
+	}
+	return m.bpfMap
 }
 
 // copy returns a copy of metadata.
@@ -441,22 +523,25 @@ func (insns Instructions) Size() uint64 {
 	return sum
 }
 
-// RewriteMapPtr rewrites all loads of a specific map pointer to a new fd.
+// AssociateMap updates all Instructions that Reference the given symbol
+// to point to an existing Map m instead.
 //
-// Returns an error if the symbol isn't used, see IsUnreferencedSymbol.
-func (insns Instructions) RewriteMapPtr(symbol string, fd int) error {
+// Returns ErrUnreferencedSymbol error if no references to symbol are found
+// in insns. If symbol is anything else than the symbol name of map (e.g.
+// a bpf2bpf subprogram), an error is returned.
+func (insns Instructions) AssociateMap(symbol string, m FDer) error {
 	if symbol == "" {
 		return errors.New("empty symbol")
 	}
 
-	found := false
+	var found bool
 	for i := range insns {
 		ins := &insns[i]
 		if ins.Reference() != symbol {
 			continue
 		}
 
-		if err := ins.RewriteMapPtr(fd); err != nil {
+		if err := ins.AssociateMap(m); err != nil {
 			return err
 		}
 
@@ -464,7 +549,40 @@ func (insns Instructions) RewriteMapPtr(symbol string, fd int) error {
 	}
 
 	if !found {
-		return &unreferencedSymbolError{symbol}
+		return fmt.Errorf("symbol %s: %w", symbol, ErrUnreferencedSymbol)
+	}
+
+	return nil
+}
+
+// RewriteMapPtr rewrites all loads of a specific map pointer to a new fd.
+//
+// Returns ErrUnreferencedSymbol if the symbol isn't used.
+//
+// Deprecated: use AssociateMap instead.
+func (insns Instructions) RewriteMapPtr(symbol string, fd int) error {
+	if symbol == "" {
+		return errors.New("empty symbol")
+	}
+
+	var found bool
+	for i := range insns {
+		ins := &insns[i]
+		if ins.Reference() != symbol {
+			continue
+		}
+
+		if !ins.IsLoadFromMap() {
+			return errors.New("not a load from a map")
+		}
+
+		ins.encodeMapFD(fd)
+
+		found = true
+	}
+
+	if !found {
+		return fmt.Errorf("symbol %s: %w", symbol, ErrUnreferencedSymbol)
 	}
 
 	return nil
@@ -588,7 +706,11 @@ func (insns Instructions) Format(f fmt.State, c rune) {
 // Returns ErrUnsatisfiedProgramReference if there is a Reference Instruction
 // without a matching Symbol Instruction within insns.
 func (insns Instructions) Marshal(w io.Writer, bo binary.ByteOrder) error {
-	if err := insns.resolveFunctionReferences(); err != nil {
+	if err := insns.encodeFunctionReferences(); err != nil {
+		return err
+	}
+
+	if err := insns.encodeMapPointers(); err != nil {
 		return err
 	}
 
@@ -619,7 +741,7 @@ func (insns Instructions) Tag(bo binary.ByteOrder) (string, error) {
 	return hex.EncodeToString(h.Sum(nil)[:unix.BPF_TAG_SIZE]), nil
 }
 
-// resolveFunctionReferences populates the Offset (or Constant, depending on
+// encodeFunctionReferences populates the Offset (or Constant, depending on
 // the instruction type) field of instructions with a Reference field to point
 // to the offset of the corresponding instruction with a matching Symbol field.
 //
@@ -628,7 +750,7 @@ func (insns Instructions) Tag(bo binary.ByteOrder) (string, error) {
 //
 // Returns ErrUnsatisfiedProgramReference if there is a Reference Instruction
 // without at least one corresponding Symbol Instruction within insns.
-func (insns Instructions) resolveFunctionReferences() error {
+func (insns Instructions) encodeFunctionReferences() error {
 	// Index the offsets of instructions tagged as a symbol.
 	symbolOffsets := make(map[string]RawInstructionOffset)
 	iter := insns.Iterate()
@@ -676,6 +798,33 @@ func (insns Instructions) resolveFunctionReferences() error {
 
 			ins.Offset = int16(symOffset - offset - 1)
 		}
+	}
+
+	return nil
+}
+
+// encodeMapPointers finds all Map Instructions and encodes their FDs
+// into their Constant fields.
+func (insns Instructions) encodeMapPointers() error {
+	iter := insns.Iterate()
+	for iter.Next() {
+		ins := iter.Ins
+
+		if !ins.IsLoadFromMap() {
+			continue
+		}
+
+		m := ins.Map()
+		if m == nil {
+			continue
+		}
+
+		fd := m.FD()
+		if fd < 0 {
+			return fmt.Errorf("map %s: %w", m, sys.ErrClosedFd)
+		}
+
+		ins.encodeMapFD(m.FD())
 	}
 
 	return nil
@@ -729,17 +878,10 @@ func newBPFRegisters(dst, src Register, bo binary.ByteOrder) (bpfRegisters, erro
 	}
 }
 
-type unreferencedSymbolError struct {
-	symbol string
-}
-
-func (use *unreferencedSymbolError) Error() string {
-	return fmt.Sprintf("unreferenced symbol %s", use.symbol)
-}
-
 // IsUnreferencedSymbol returns true if err was caused by
 // an unreferenced symbol.
+//
+// Deprecated: use errors.Is(err, asm.ErrUnreferencedSymbol).
 func IsUnreferencedSymbol(err error) bool {
-	_, ok := err.(*unreferencedSymbolError)
-	return ok
+	return errors.Is(err, ErrUnreferencedSymbol)
 }
--- a/asm/instruction_test.go
+++ b/asm/instruction_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -141,7 +142,7 @@ func TestInstructionLoadMapValue(t *testing.T) {
 	if !ins.IsLoadFromMap() {
 		t.Error("isLoadFromMap returns false")
 	}
-	if fd := ins.MapPtr(); fd != 1 {
+	if fd := ins.mapFd(); fd != 1 {
 		t.Error("Expected map fd to be 1, got", fd)
 	}
 	if off := ins.mapOffset(); off != 123 {
@@ -171,7 +172,7 @@ func TestInstructionsRewriteMapPtr(t *testing.T) {
 		t.Error("Constant should be 2, have", insns[0].Constant)
 	}
 
-	if err := insns.RewriteMapPtr("bad", 1); !IsUnreferencedSymbol(err) {
+	if err := insns.RewriteMapPtr("bad", 1); !errors.Is(err, ErrUnreferencedSymbol) {
 		t.Error("Rewriting unreferenced map doesn't return appropriate error")
 	}
 }
--- a/collection.go
+++ b/collection.go
@@ -5,13 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"reflect"
 	"strings"
 
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal/btf"
-	"github.com/cilium/ebpf/internal/sys"
 )
 
 // CollectionOptions control loading a collection into the kernel.
@@ -70,15 +68,14 @@ func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
 	for symbol, m := range maps {
 		// have we seen a program that uses this symbol / map
 		seen := false
-		fd := m.FD()
 		for progName, progSpec := range cs.Programs {
-			err := progSpec.Instructions.RewriteMapPtr(symbol, fd)
+			err := progSpec.Instructions.AssociateMap(symbol, m)
 
 			switch {
 			case err == nil:
 				seen = true
 
-			case asm.IsUnreferencedSymbol(err):
+			case errors.Is(err, asm.ErrUnreferencedSymbol):
 				// Not all programs need to use the map
 
 			default:
@@ -469,9 +466,10 @@ func (cl *collectionLoader) loadProgram(progName string) (*Program, error) {
 			continue
 		}
 
-		if uint32(ins.Constant) != math.MaxUint32 {
-			// Don't overwrite maps already rewritten, users can
-			// rewrite programs in the spec themselves
+		// Don't overwrite map loads containing non-zero map fd's,
+		// they can be manually included by the caller.
+		// Map FDs/IDs are placed in the lower 32 bits of Constant.
+		if int32(ins.Constant) > 0 {
 			continue
 		}
 
@@ -480,11 +478,7 @@ func (cl *collectionLoader) loadProgram(progName string) (*Program, error) {
 			return nil, fmt.Errorf("program %s: %w", progName, err)
 		}
 
-		fd := m.FD()
-		if fd < 0 {
-			return nil, fmt.Errorf("map %s: %w", ins.Reference(), sys.ErrClosedFd)
-		}
-		if err := ins.RewriteMapPtr(m.FD()); err != nil {
+		if err := ins.AssociateMap(m); err != nil {
 			return nil, fmt.Errorf("program %s: map %s: %w", progName, ins.Reference(), err)
 		}
 	}
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -450,12 +450,6 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 
 		ins.Src = asm.PseudoMapFD
 
-		// Mark the instruction as needing an update when creating the
-		// collection.
-		if err := ins.RewriteMapPtr(-1); err != nil {
-			return err
-		}
-
 	case dataSection:
 		var offset uint32
 		switch typ {
@@ -488,12 +482,6 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 		// The kernel expects the offset in the second basic BPF instruction.
 		ins.Constant = int64(uint64(offset) << 32)
 		ins.Src = asm.PseudoMapValue
-
-		// Mark the instruction as needing an update when creating the
-		// collection.
-		if err := ins.RewriteMapPtr(-1); err != nil {
-			return err
-		}
 
 	case programSection:
 		switch opCode := ins.OpCode; {
--- a/linker.go
+++ b/linker.go
@@ -120,7 +120,8 @@ func fixupAndValidate(insns asm.Instructions) error {
 	for iter.Next() {
 		ins := iter.Ins
 
-		if ins.IsLoadFromMap() && ins.MapPtr() == -1 {
+		// Map load was tagged with a Reference, but does not contain a Map pointer.
+		if ins.IsLoadFromMap() && ins.Reference() != "" && ins.Map() == nil {
 			return fmt.Errorf("instruction %d: map %s: %w", iter.Index, ins.Reference(), asm.ErrUnsatisfiedMapReference)
 		}
 


lmb (commented):
Hi Dylan,

Thanks for the PR! I started the review yesterday, but only managed to finish it today.

- Can you explain what the kernel uses the various bits of information for? I know that SetLine ends up in the verifier log, but what about File, Line, Column?
- How does the library use this information? Seems like we ignore File, Line, Column?
- Is it possible to only pass e.g. Line, but leave other bits empty?

Regarding the metadata, I'd prefer if the setters modified the instruction instead of making a copy, so pointer receivers. Behind the scenes metadata would still be copied. Intuitively I expect a setter to make changes to the object I'm calling it on. Calling the setters also becomes awkward judging from the changes in the PRs.

I also saw that you merged master into your feature branch, maybe that was an accident? Doing that makes merges much messier, please rebase instead.

asm/instruction.go
```diff
@@ -133,15 +145,39 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 	return 2 * InstructionSize, nil
 }
 
-// RewriteMapPtr changes an instruction to use a new map fd.
+// RewriteMap changes an instruction to use a new map's fd.
+// This function also adds a reference to `m` in the instruction so the passed map isn't garbage collected if no other
```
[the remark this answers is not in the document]
dylandreimerink: Agreed

asm/instruction.go
```diff
@@ -296,16 +332,202 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
+		return ins
+	}
+
+	ins.metadata = ins.metadata.copy()
+	ins.metadata.bpfMap = ebpfMap
+	return ins
+}
+
+// Map denotes the *ebpf.Map from which this instruction preforms a data.
+func (ins Instruction) Map() ebpfMap {
```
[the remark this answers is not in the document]
dylandreimerink: This was a remnant from when the SetMap was still exported, I made it un-exported but forgot to un-export the getter, will fix that.

asm/instruction.go
```diff
@@ -296,16 +332,202 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
+		return nil
+	}
+
+	return ins.metadata.bpfMap
+}
+
+// ebpfMap isn't actually used as a meaningful interface, rater it is used since we can't directly use ebpf.Map as type
+// since this would cause in import loop. The methods were chosen because they are unique to ebpf.Map and don't
+// reference other types in the ebpf package
+type ebpfMap interface {
```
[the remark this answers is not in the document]
dylandreimerink: I thought narrowing it as much as possible might be desirable to avoid mistakes, but it was a workaround to begin with. Will change it to an `FDer`

asm/instruction.go
```diff
@@ -490,8 +743,14 @@ func (insns Instructions) Format(f fmt.State, c rune) {
...
 	for iter.Next() {
-		if iter.Ins.Symbol != "" {
-			fmt.Fprintf(f, "%s%s:\n", symIndent, iter.Ins.Symbol)
+		if iter.Ins.Symbol() != "" {
+			fmt.Fprintf(f, "%s%s:\n", symIndent, iter.Ins.Symbol())
+		}
+		if line := iter.Ins.Line(); line != "" {
+			line := strings.TrimSpace(line)
+			if line != "" {
+				fmt.Fprintf(f, "%s%*s; %s\n", indent, offsetWidth, " ", line)
```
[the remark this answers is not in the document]
dylandreimerink: Ends up looking like this (From the TestDataSections test)
```
  ; __section("socket/3") int data_sections() {
 0: MovImm dst: r0 imm: 150
  ; if (uneg != (unsigned int)-1)
 1: LoadMapValue dst: r1, fd: -1 off: 8 <.rodata>
 3: LdXMemW dst: r1 src: r1 off: 0 imm: 0
 4: LdImmDW dst: r2 imm: 4294967295
  ; if (uneg != (unsigned int)-1)
 6: JNEReg dst: r1 off: 22 src: r2
 7: MovImm dst: r0 imm: 153
  ; if (neg != -2)
 8: LoadMapValue dst: r1, fd: -1 off: 12 <.rodata>
10: LdXMemW dst: r1 src: r1 off: 0 imm: 0
11: LdImmDW dst: r2 imm: 4294967294
  ; if (neg != -2)
13: JNEReg dst: r1 off: 15 src: r2
14: MovImm dst: r0 imm: 156
  ; if (static_uneg != (unsigned int)-3)
15: LoadMapValue dst: r1, fd: -1 off: 16 <.rodata>
17: LdXMemW dst: r1 src: r1 off: 0 imm: 0
18: LdImmDW dst: r2 imm: 4294967293
  ; if (static_uneg != (unsigned int)-3)
20: JNEReg dst: r1 off: 8 src: r2
  ; if (static_neg != -4)
21: LoadMapValue dst: r1, fd: -1 off: 20 <.rodata>
23: LdXMemW dst: r1 src: r1 off: 0 imm: 0
24: MovImm dst: r0 imm: 0
25: LdImmDW dst: r2 imm: 4294967292
27: JEqReg dst: r1 off: 1 src: r2
28: MovImm dst: r0 imm: 159
  ; }
29: Exit
```

elf_reader.go
```diff
@@ -393,19 +395,48 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
+				name = insns.Name()
+			}
+
+			lineInfos = ec.btf.GetFuncLineInfos(name)
+		}
+		for _, lineInfo := range lineInfos {
+			if uint64(lineInfo.InsnOff) == funcOffset {
+				fileName, err := lineInfo.FileName(ec.btf)
+				if err != nil {
+					continue
```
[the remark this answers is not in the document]
dylandreimerink: Should have been break, this point is moot anyway since we will be dropping this look in favor of the map lookup

elf_reader.go
```diff
@@ -393,19 +395,48 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
 			}
 		}
 
+		if ec.btf != nil {
+			name := ins.Symbol()
+			if name == "" {
+				name = insns.Name()
+			}
+
+			lineInfos = ec.btf.GetFuncLineInfos(name)
```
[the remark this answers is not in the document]
dylandreimerink: We need to re-get it if the current function changes, have found a better way to do this already

asm/instruction.go
```diff
@@ -296,16 +332,202 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
+		a = *m
+	}
+	if other != nil {
+		b = *other
+	}
+
+	return a == b
+}
+
+func (m *metadata) copy() *metadata {
```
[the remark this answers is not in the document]
dylandreimerink: agreed

dylandreimerink:
> Can you explain what the kernel uses the various bits of information for? I know that SetLine ends up in the verifier log, but what about File, Line, Column?

As far as I know, the kernel doesn't use any of the `.BTF.ext` data(LineInfos and FuncInfos) itself, as far as I can tell from the [initial commit](https://github.com/torvalds/linux/commit/c454a46b5efd8eff8880e88ece2976e60a26bf35) the main goal was to add source code references for in the verifier as well as when reading the xlated and jitted instructions back from the kernel.

> How does the library use this information? Seems like we ignore File, Line, Column?

I used the `Line` to add to `Instructions.Format` which now has a output similar to bpftool and llvm when outputting instructions with lineinfo.

The Filename and Column information isn't currently used by the library directly. My motivation for adding is was to use them outside of the library in [edb](https://github.com/dylandreimerink/edb/blob/58c0e340a3930172de2df972c28509f0a92d9917/cmd/debug/cmd.go#L145) where I use the information to display the relevant source file and place the cursor at the correct line within that file. 

> Is it possible to only pass e.g. Line, but leave other bits empty?

Yea it should be. If we were to create a `LineInfo` object from the `metadata` we can leave the `LineCol` and `FileNameOff` at 0. Which I believe are valid values, since the line and column can both be zero and a offset of 0 within the strings table will just result in an empty string. I think the most important part is to not have duplicate `InsnOff` values within the same function, which should not be an issue.

internal/btf/btf.go
```diff
@@ -172,6 +172,15 @@ func loadSpecFromELF(file *internal.SafeELFFile, variableOffsets map[variable]ui
 	return spec, nil
 }
 
+// GetFuncLineInfos returns the LineInfos for the given function name or nil if LineInfos can be found
+func (spec *Spec) GetFuncLineInfos(funcName string) LineInfos {
```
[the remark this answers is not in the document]
lmb: I'll see whether I can improve this.

asm/instruction.go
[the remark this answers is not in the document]
dylandreimerink: Yea, shouldn't be hard, I can modify `ExampleInstructions_Format`

elf_reader.go
```diff
@@ -393,19 +395,48 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
+		}
+		for _, lineInfo := range lineInfos {
```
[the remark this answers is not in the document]
dylandreimerink: agreed

asm/instruction.go
```diff
@@ -296,16 +332,202 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
 	}
 }
 
+func (ins Instruction) Equal(other Instruction) bool {
+	return ins.OpCode == other.OpCode &&
+		ins.Dst == other.Dst &&
+		ins.Src == other.Src &&
+		ins.Offset == other.Offset &&
+		ins.Constant == other.Constant &&
+		ins.metadata.equal(other.metadata)
```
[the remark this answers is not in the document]
dylandreimerink: Currently `Reference` and `Symbol` are checked when comparing two instructions with ==, so I attempted to preserve that. But I think checking the equality of instructions outside of the tests is rarely used, at least I can't think of a use case. So un-exporting it for now seems fine.

asm/jump.go
```diff
@@ -62,48 +62,44 @@ func (op JumpOp) Op(source Source) OpCode {
...
-		Offset:    -1,
-		Constant:  int64(value),
-		Reference: label,
-	}
+	return (Instruction{
+		OpCode:   op.opCode(JumpClass, ImmSource),
+		Dst:      dst,
+		Offset:   -1,
+		Constant: int64(value),
+	}).SetReference(label)
```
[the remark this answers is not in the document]
dylandreimerink: The point of the setter is to abstract the metadata struct away so we might change its implementation later. It is not strictly necessary since we are in the same package,  but this might save us refactoring work down the line. I don't feel strongly either way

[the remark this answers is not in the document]
lmb: Let's keep it as it is.

dylandreimerink:
Got almost everything. The only thing left is to make `btf.LineInfo` a `fmt.Stringer`, which we can do as soon as #572 has been merged. I can rebase and update `WIP: asm: Replaced seperate getters/setters with Context` once that is done.

asm/instruction_test.go
```diff
@@ -280,3 +288,46 @@ func TestInstructionIterator(t *testing.T) {
...
+	}
+
+	if insn[0].metadata == insn2[0].metadata {
+		t.Fatal("changed instruction should not be equal")
+	}
+
+	// Set the reference, then clear it. Causing us to have a nil in one instruction and an empty metadata in the
+	// other.
+	insn[1].SetReference("SomeValue")
+	insn[1].SetReference("")
```
ti-mo, line 320: I think this sets the reference on the wrong insn.

asm/instruction_test.go
```diff
@@ -280,3 +288,46 @@ func TestInstructionIterator(t *testing.T) {
...
+
+	// Metadata is value compared, not pointer compared, so so these should still be equal
+	if !insn[1].equal(insn2[1]) {
```
ti-mo, line 323: `equal()` no longer compares metadata.

asm/instruction_test.go
```diff
@@ -280,3 +288,46 @@ func TestInstructionIterator(t *testing.T) {
...
+		t.Fatal("instructions with nil and empty metadata should be equal")
+	}
+
+	insn[1].SetReference("abc")
```
ti-mo, line 327: Idem, this is not testing anything anymore.

collection.go
```diff
@@ -452,27 +450,21 @@ func (cl *collectionLoader) loadProgram(progName string) (*Program, error) {
 	for i := range progSpec.Instructions {
 		ins := &progSpec.Instructions[i]
 
-		if !ins.IsLoadFromMap() || ins.Reference == "" {
-			continue
-		}
-
-		if uint32(ins.Constant) != math.MaxUint32 {
```
ti-mo, line 459: I think this might incur an unnecessary breaking change since we've never enforced clearing the Reference field if the caller rewrites their own map fds.

@lmb any thoughts?

elf_reader.go
```diff
@@ -393,19 +395,51 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
+			if err != nil {
+				fileName = ""
+			}
+
+			line, err := lineInfo.Line(ec.btf)
+			if err != nil {
+				line = ""
+			}
+
+			// TODO we should use the *btf.LineInfo here directly, as soon as it implements fmt.Stringer
```
ti-mo, line 430: I don't think this is necessary. We should avoid adding any  dependencies to the `asm` package. Just a string is fine here (see my thoughts around `SimpleContext` above).

elf_reader.go
```diff
@@ -393,19 +395,51 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
 				}
 
-				ins.Reference = sym
+				ins.SetReference(sym)
 				ins.Constant = -1
 			}
 		}
 
+		// Only rebuild lineInfos at the start of a function
+		if funcOffset == 0 {
```
ti-mo, line 410: This block should probably be moved to around line 381 above where the function body is rolled over.

I would make `ec.btf != nil` the enclosing check to avoid reallocating `lineInfos` for each function body, even if there is no BTF.

elf_reader.go
```diff
@@ -393,19 +395,51 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
+			lineInfos = make(map[uint64]btf.LineInfo)
+			if ec.btf != nil {
+				for _, lineInfo := range ec.btf.GetFuncLineInfos(ins.Symbol()) {
```
ti-mo, line 413: Could `GetFuncLineInfos` already return lineinfos by insn offset? Let's try to keep `loadFunctions` lean.

elf_reader.go
```diff
@@ -393,19 +395,51 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
+					lineInfos[uint64(lineInfo.InsnOff)] = lineInfo
+				}
+			}
+		}
+
+		if lineInfo, found := lineInfos[funcOffset]; found {
```
ti-mo, line 419: Could a `String()` equivalent be implemented onto `btf.LineInfo` instead of having printing logic here? Might be easier to make rendering decisions, but maybe @lmb's PR (haven't reviewed yet) will make this a bit more straightforward.

elf_reader.go
```diff
@@ -393,19 +395,51 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
...
+					return nil, fmt.Errorf("offset %d: no jump target found at offset %d", secOffset, tgt)
 				}
 
-				ins.Reference = sym
+				ins.SetReference(sym)
 				ins.Constant = -1
 			}
 		}
 
+		// Only rebuild lineInfos at the start of a function
```
ti-mo, line 409: What about a subprog's FuncInfo? Could we also pull that in here?

internal/btf/ext_info.go
```diff
@@ -328,6 +328,26 @@ type LineInfo struct {
...
+}
+
+// LineNum returns the line number within the FileName
+func (li LineInfo) LineNum() uint32 {
+	return li.LineCol >> 10
+}
+
+// ColNumber returns the column number within the Line
+func (li LineInfo) ColNum() uint32 {
+	return li.LineCol & 0x3ff
```
ti-mo, line 348: The `0x3ff` magic constant should have a name.

internal/btf/ext_info.go
```diff
@@ -328,6 +328,26 @@ type LineInfo struct {
...
+}
+
+// Line returns the line of source code as a string
+func (li LineInfo) Line(btfSpec *Spec) (string, error) {
+	return btfSpec.strings.Lookup(li.LineOff)
+}
+
+// LineNum returns the line number within the FileName
+func (li LineInfo) LineNum() uint32 {
+	return li.LineCol >> 10
```
ti-mo, line 343: Why 10 bits to the right?

ti-mo (commented):
Went through it in detail, great work! This has been on our minds for a few months now and will open some exciting new possibilities.

asm/instruction.go
```diff
@@ -99,6 +111,13 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 		return 0, errors.New("invalid opcode")
 	}
 
+	if ins.IsLoadFromMap() && ins.metadata != nil && ins.metadata.bpfMap != nil {
+		err := ins.RewriteMapPtr(ins.metadata.bpfMap.FD())
+		if err != nil {
```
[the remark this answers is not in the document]
dylandreimerink: agreed

asm/instruction.go
```diff
@@ -99,6 +111,13 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 		return 0, errors.New("invalid opcode")
 	}
 
+	if ins.IsLoadFromMap() && ins.metadata != nil && ins.metadata.bpfMap != nil {
+		err := ins.RewriteMapPtr(ins.metadata.bpfMap.FD())
```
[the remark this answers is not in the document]
dylandreimerink: That is a valid point

asm/instruction.go
```diff
@@ -136,15 +155,30 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
...
+
+// RewriteMapPtr changes an instruction to use a new map fd.
+//
+// Returns an error if the instruction doesn't load a map.
+//
+// Deprecated: use RewriteMap instead.
+func (ins *Instruction) RewriteMapPtr(fd int) error {
+	if !ins.IsLoadFromMap() {
 		return errors.New("not a load from a map")
 	}
```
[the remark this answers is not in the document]
dylandreimerink: agreed

asm/instruction.go
```diff
@@ -299,16 +339,111 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
+func (ins Instruction) Reference() string {
+	if ins.metadata == nil {
+		return ""
+	}
+
+	return ins.metadata.reference
+}
+
+// WithContext adds information about the origin/source of the instruction.
+func (ins Instruction) WithContext(ctx fmt.Stringer) Instruction {
```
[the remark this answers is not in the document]
dylandreimerink: > What about `WithInfo`, from extInfos? This is still reminiscent of the original term.

I don't mind that, this came about because we would like to define a comment for example which is not related to extInfos as such.

> Also, I think this could just be a string; a `SimpleContext` feels a bit awkward.

We discussed having just a string with a custom format e.g. `foo.c:12:34: {line or comment}` with optional parts so a plain string `These instructions perform ### function.` or `firewall.go:123: asm.Ja.Label("some-label")`. Downside is that this is somewhat less flexible than defining a interface, and @lmb was concerned about the the code size/performance implication of a string parser.

> Thinking about marshaling this later, if an instruction has an Info and is a Symbol, we emit a funcInfo

We also need a BTF type ID to creating funcInfo, which we currently don't have metadata for. I wanted to implement that when we start work on marshalling.

> The kernel requires there be exactly one funcInfo pointing at the start of each subprog

I don't think that the kernel requires this. Perhaps in some cases, but I have loaded programs without BTF info before without issue.

> iirc an instruction cannot have both a lineInfo and a funcInfo

Not the case, I have seen a few programs which have both at the first instruction.

asm/instruction.go
```diff
@@ -299,16 +339,111 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
+func (m *metadata) copy() *metadata {
+	var copy metadata
+	if m != nil {
+		copy = *m
+	}
+	return &copy
+}
+
+// SimpleContext can be used to set a string as context of an Instruction.
+type SimpleContext string
```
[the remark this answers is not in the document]
dylandreimerink: Yea, also not my preference, makes the DSL messy, `...WithContext(SimpleContext("some comment"))`. But lets discuses that in the above comment/thread.

asm/instruction.go
```diff
@@ -493,8 +662,14 @@ func (insns Instructions) Format(f fmt.State, c rune) {
 
 	iter := insns.Iterate()
 	for iter.Next() {
-		if iter.Ins.Symbol != "" {
-			fmt.Fprintf(f, "%s%s:\n", symIndent, iter.Ins.Symbol)
+		if iter.Ins.Symbol() != "" {
+			fmt.Fprintf(f, "%s%s:\n", symIndent, iter.Ins.Symbol())
+		}
+		if ctx := iter.Ins.Context(); ctx != nil {
+			line := strings.TrimSpace(ctx.String())
```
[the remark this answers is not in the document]
dylandreimerink: Indentation in the source code itself, which doesn't make sense when displaying the between the non-indented instructions

asm/instruction.go
```diff
@@ -299,16 +339,111 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
+	symbol string
+	// context denotes information about the origin/source of the instruction.
+	context fmt.Stringer
+
+	// bpfMap denotes the *ebpf.Map from which this instruction preforms a data.
+	bpfMap FDer
+}
+
+// returns a copy of metadata, return value is never nil, even if called on a nil value
+func (m *metadata) copy() *metadata {
```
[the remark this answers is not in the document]
dylandreimerink: Hmm. That is indeed an issue arising out of the whole context object situation. I see the following solutions:
* Using reflection to perform a deep-copy(limited to only exported fields)
* Defining a `Context` interface and ask all implementations to implement a `Copy` function.
* Have only value types in the metadata(have context be a string which is parsed to LineInfo when during marshaling)
* Give up on the copy-on-write principle and accept that we can't just value copy Instructions

asm/instruction_test.go
dylandreimerink: This is intentional, since this will result in an allocated metadata object with an empty string. Testing equality in such an edge case. But this is no longer important since we don't compare metadata anymore in `equal()`, so we can ditch it.

asm/instruction_test.go
dylandreimerink: Right, bit of a theme, have to update the unit tests after the changed made during the review

asm/instruction_test.go
dylandreimerink: Indeed

collection.go
dylandreimerink: iirc, I removed the check at some point because it was causing issues with the tests, but adding it back now doesn't break the tests. So I guess keeping it doesn't hurt, and allows people to rewrite their own FDs if they so wish.

elf_reader.go
dylandreimerink: Right, didn't realize we also enter that block on the very first iteration as well

elf_reader.go
dylandreimerink: Yea, this would also be the location to do the same for FuncInfo. That is something we will have to add before we can marshall metadata back into wire formatted LineInfo and FuncInfo. 

FuncInfo is nothing more than a name and BTF type, however I skipped this because dealing with references to BTF might have some implications. Especially on how to marshall a reference back to type IDs in the final blob, and if we would need additional info. So I decided to not implement the whole funcInfo part yet until we have a better plan for the marshalling part. I have no use case for it as of now since we already have symbols which are basically the same just without the BTF type.

internal/btf/ext_info.go
dylandreimerink: Agreed

internal/btf/ext_info.go
dylandreimerink: That is the amount of `1`s in `0x3ff`. LineCol is has 22 bits of line number and 10 bits of column number packed into one value. Perhaps a link to the kernels BTF docs can help. But I think this is moot since #572 will change the LineInfo to just have separate Line and Col numbers exported which replaces this.

asm/instruction.go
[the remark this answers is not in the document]
ti-mo: > I don't think that the kernel requires this. Perhaps in some cases, but I have loaded programs without BTF info before without issue.

Loading programs without BTF has always been possible (BTF is relatively recent), but if a BTF blob is present, each subprog must be accompanied by a valid funcInfo with the right offset.

> Not the case, I have seen a few programs which have both at the first instruction.

Yep, then we need an extra field. I forgot that funcInfo is actual type info of `func -> func_proto` that also contains the args and their types, so we can't just synthesize it from the Symbol() value. :/ Let's take this into account before settling on the `WithContext` name, let me mull it over for a bit. LineInfo and FuncInfo are both 'contextual' information, I still find the term 'context' too generic for what it actually describes.

Maybe something like `WithLineInfo()` and `WithFuncInfo()` would be more clear? An important use case is checking whether the instruction contains a FuncInfo, so a single func/line Stringer field makes this rather difficult from the caller pov.

Any ideas?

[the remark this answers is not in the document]
dylandreimerink: > subprog must be accompanied by a valid funcInfo with the right offset.

Yea, discovered that the hard way recently, a good reason to add funcInfo before moving on the generating BTF. I already have something brewing for that. But adding lineInfo without the func info doesn't hurt until that point.

>  LineInfo and FuncInfo are both 'contextual' information, I still find the term 'context' too generic for what it actually describes.

Agreed, I think `WithLineInfo` is good, that is after all exactly what this is, information about a line of source code. And while we allow any stringer to serve as the actual `Line` part of it for the sake of comments, the name still fits, and doesn't clash with funcInfo.

asm/instruction_test.go
```diff
@@ -280,3 +288,46 @@ func TestInstructionIterator(t *testing.T) {
...
+	if insn[0].Reference() != "my_func" {
+		t.Fatal("metadata modification to copied instruction effected the old instruction")
+	}
+
+	if insn2[0].Reference() != "my_func2" {
+		t.Fatal("SetReference didn't update new instruction")
+	}
+
+	if insn[0].metadata == insn2[0].metadata {
+		t.Fatal("changed instruction should not be equal")
```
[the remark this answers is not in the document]
dylandreimerink: Since the metadata is copy on write, it should not.

internal/btf/btf.go
```diff
@@ -190,6 +190,22 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
 	return spec, nil
 }
 
+// GetFuncLineInfos returns a map of LineInfo's for the given function name. The map is indexes by the
+// instruction offset of the lineInfo.
+func (spec *Spec) GetFuncLineInfos(funcName string) map[uint32]LineInfo {
+	if spec.lineInfos == nil {
+		return make(map[uint32]LineInfo)
```
lmb, line 197: Just `nil` should also work?

internal/btf/btf.go
```diff
@@ -190,6 +190,22 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
 	return spec, nil
 }
 
+// GetFuncLineInfos returns a map of LineInfo's for the given function name. The map is indexes by the
+// instruction offset of the lineInfo.
+func (spec *Spec) GetFuncLineInfos(funcName string) map[uint32]LineInfo {
```
lmb, line 195: Can you use `Spec.Program` and `Program.LineInfos` instead? The implementation of this function can be private to the elf reader.

collection.go
```diff
@@ -484,7 +483,7 @@ func (cl *collectionLoader) loadProgram(progName string) (*Program, error) {
 		if fd < 0 {
```
lmb, line 483: `fd` is not necessary anymore?

internal/btf/btf.go
lmb, line 197: This can just be `nil`, it's allowed to do a lookup in a nil map.

asm/instruction.go
```diff
@@ -356,12 +398,34 @@ func (ins *Instruction) copyMetadata() *metadata {
...
 // metadata holds metadata about an Instruction.
 type metadata struct {
 	// reference denotes a reference (e.g. a jump) to another symbol.
 	reference string
 	// symbol denotes an instruction at the start of a function body.
 	symbol string
+	// source denotes information about the origin/source of the instruction.
+	source fmt.Stringer
+
+	// bpfMap denotes the *ebpf.Map from which this instruction preforms a data.
```
lmb, line 427: `preforms a data`?

asm/instruction.go
```diff
@@ -356,12 +398,34 @@ func (ins *Instruction) copyMetadata() *metadata {
 	return ins.metadata
 }
 
+// setMap sets the *ebpf.Map from which this instruction preforms a data.
+func (ins *Instruction) setMap(m FDer) {
+	if (ins.metadata != nil && ins.metadata.bpfMap == m) ||
```
lmb, line 403: Add `metadata.Map` as for other getters / setters?

asm/instruction.go
```diff
@@ -579,6 +688,12 @@ func (insns Instructions) Format(f fmt.State, c rune) {
 		if iter.Ins.Symbol() != "" {
 			fmt.Fprintf(f, "%s%s:\n", symIndent, iter.Ins.Symbol())
 		}
+		if ctx := iter.Ins.Source(); ctx != nil {
```
lmb, line 691: `ctx`?

collection.go
dylandreimerink: It is only used here to make sure that the passed map is valid. It returns a `sys.ErrClosedFd` which I guess holds meaning. 

Should we move this check to `Instruction.Marshal`? Or do you propose to not check map validity at all?

internal/btf/btf.go
dylandreimerink: :open_mouth: I thought that reading from an nil map also caused a nil pointer panic like writing, learned something today. Will update this

asm/instruction.go
```diff
@@ -356,12 +398,34 @@ func (ins *Instruction) copyMetadata() *metadata {
 	return ins.metadata
 }
 
+// setMap sets the *ebpf.Map from which this instruction preforms a data.
+func (ins *Instruction) setMap(m FDer) {
```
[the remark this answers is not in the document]
dylandreimerink: We could by embedding `fmt.Stringer` in the `FDer` interface, since `*ebpf.Map` does have a [String()](https://github.com/cilium/ebpf/blob/d4100a1f430f7792dd6ab3113a2dbbe6a1c10e82/map.go#L461) func.

We currently already have the reference as `<ref>` behind instructions, so after replacing a string reference with a map we could replace it with the string'ed version of the map(its type, name and fd). So that seems logical.

On the other hand, I would rather do that in a separate PR to avoid feature creep in this PR, otherwise it will never be merged :smiley:

asm/instruction.go
```diff
@@ -127,30 +131,53 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
...
 	rawFd := uint64(uint32(fd))
 	ins.Constant = int64(offset | rawFd)
-	return nil
 }
 
 // MapPtr returns the map fd for this instruction.
 //
 // The result is undefined if the instruction is not a load from a map,
 // see IsLoadFromMap.
 func (ins *Instruction) MapPtr() int {
```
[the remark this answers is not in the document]
dylandreimerink: Don't know if we want to just yet, I recall an earlier change to this PR suggested by @ti-mo to still support use cases where users can specify their own FD's instead of using our facilities. Though I suppose a user could always implement their own `FDer` to cover that scenario. 

Perhaps we should then also provide a `Map() FDer` getter? Or decide to not support custom FD setting from outside the lib.

internal/btf/btf.go
dylandreimerink: Is that the direction we want to be headed in?, especially regarding #522

asm/instruction.go
```diff
@@ -356,12 +398,34 @@ func (ins *Instruction) copyMetadata() *metadata {
...
+	if (ins.metadata != nil && ins.metadata.bpfMap == m) ||
+		(ins.metadata == nil && m == nil) {
+		return
+	}
+
+	ins.metadata = ins.metadata.copy()
+	ins.metadata.bpfMap = m
+}
+
+// FDer isn't actually used as a meaningful interface, rater it is used because we can't directly use types from the
```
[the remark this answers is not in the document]
ti-mo:
```suggestion
// FDer represents a resource tied to an underlying file descriptor.
// Used as a stand-in for e.g. ebpf.Map since that type cannot be
// imported here and FD() is the only method we rely on.
```

asm/instruction.go
```diff
@@ -127,30 +131,53 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
...
-	if !ins.OpCode.IsDWordLoad() {
-		return fmt.Errorf("%s is not a 64 bit load", ins.OpCode)
+func (ins *Instruction) AssociateMap(m FDer) error {
+	if !ins.IsLoadFromMap() {
+		return errors.New("not a load from a map")
 	}
 
-	if ins.Src != PseudoMapFD && ins.Src != PseudoMapValue {
+	// Set fd to -1 since it is technically not yet rewritten
+	ins.encodeMapFD(-1)
```
ti-mo, line 143: I'm not sure if we still need to do this. -1 flags to `fixupAndValidate` that a `Reference()` to a map hasn't been resolved into an fd yet, but as of this patch, that's no longer necessary.

Either:
- ins has a non-nil `Map()`, which can be resolved using `ins.Map().FD()` during marshaling
- the caller has set the FD manually (`> 0`)
- `ins.IsLoadFromMap()` is true and Constant is 0

I would suggest we drop this line and raise `ErrUnsatisfiedMapReference` when `Constant <= 0` in `fixupAndValidate()`.

asm/instruction.go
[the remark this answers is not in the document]
ti-mo: Maybe the comment was a bit terse, this method should be sunset with `// Deprecated:` to be removed in 0.9.

Caller-supplied FDer would indeed be an excellent workaround for users that are somehow unable to switch. This should go into the patch notes and in the deprecation notices on both `MapPtr()` and `RewriteMapPtr()`.

asm/instruction.go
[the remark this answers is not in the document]
ti-mo: Yep, let's do this in a follow-up. I agree it's a must-have, though.

asm/instruction_test.go
```diff
@@ -176,12 +176,19 @@ func TestInstructionsRewriteMapPtr(t *testing.T) {
 	}
 }
 
+type Comment string
```
ti-mo, line 179: @lmb What are your thoughts on putting this in the package proper? I think providing this as part of the DSL makes sense, I see little point in letting the caller declare this manually.

asm/instruction.go
```diff
@@ -441,9 +521,43 @@ func (insns Instructions) Size() uint64 {
...
+		}
+
+		// Remove the reference, since we have now resolved it, so we will not attempt to do so again later.
+		*ins = ins.WithReference("")
+
+		found = true
+	}
+
+	if !found {
+		return &unreferencedSymbolError{symbol}
```
ti-mo, line 550: This (barely) predates `%w`, let's make this an unexported sentinel.

collection.go
lmb: Yeah `Instruction.Marshal` sounds like the place to do this.

asm/instruction.go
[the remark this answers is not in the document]
lmb: Makes me wish I had used `Map.Fd` instead of `Map.FD`. That way callers could use `os.File` as an Fder.

asm/instruction.go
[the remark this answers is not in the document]
lmb: SGTM. We don't need fmt.Stringer in `FDer`, calling `fmt.Sprint` is enough and will figure out the correct way to stringify.

asm/instruction_test.go
lmb: I don't feel strongly about it, but would probably not include it in the package. It's four lines of code.

internal/btf/btf.go
lmb: It's better than adding new API which we need to take care of later, and it doesn't necessarily make #522 harder I think.

lmb:
@dylandreimerink can you add a `Fixes #515` to the commit with map metadata?

asm/instruction.go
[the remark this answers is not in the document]
dylandreimerink: I was in the neighborhood anyway to remove `MapPtr()`, added this anyway in the PR, wasn't that much work

asm/instruction.go
dylandreimerink: Good suggestion

asm/instruction_test.go
dylandreimerink: It makes sense to me to include it in the package, I know it is only 4 lines, but it saves the user having to figure this out. So I have to agree with @ti-mo on this one.

asm/instruction.go
dylandreimerink: No quite sure what you mean, do you want to replace this line with `fmt.Errorf("unreferenced symbol %s", symbol)` or do you wat to export `unreferencedSymbolError` and remove the `IsUnreferencedSymbol` so it can be used with the `errors.Is` function?

ti-mo (commented):
Have a few last comments. Looks like the commit title and message of `asm,elf_reader: added instruction context` is now outdated.

As discussed offline, let's only merge the Map support here, the .BTF.ext unmarshaler needs a few changes since we're processing LineInfos per section again.

ti-mo:
The LineInfo metadata was split off from this PR as this one already touched several subtle interactions regarding maps between the loader, linker and marshaler. This one is good to go for now.

PR #569, opened 2022-02-15
asm: correct opcode Class descriptions

ti-mo:
Removed use of the confusing term 'constant'. The term 'immediate' is more appropriate when describing memory/register values embedded in the instruction stream.

Also, Ld(X) and St(X) appear to have been reversed. The X variants operate on memory, the non-X ones operate on registers. Use both 'load' and 'store' to describe the respective classes.

### DIFF
--- a/asm/opcode.go
+++ b/asm/opcode.go
@@ -18,22 +18,23 @@ type Class uint8
 const classMask OpCode = 0x07
 
 const (
-	// LdClass load memory
+	// LdClass loads immediate values into registers.
+	// Also used for non-standard load operations from cBPF.
 	LdClass Class = 0x00
-	// LdXClass load memory from constant
+	// LdXClass loads memory into registers.
 	LdXClass Class = 0x01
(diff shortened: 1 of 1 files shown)


asm/opcode.go
```diff
@@ -18,22 +18,22 @@ type Class uint8
 const classMask OpCode = 0x07
 
 const (
-	// LdClass load memory
```
lmb, line 21: Maybe we just copy what https://www.kernel.org/doc/html/latest/bpf/instruction-set.html#instruction-classes has?

ti-mo: For LD, it states 'non-standard load operations', which is as vague as can be. Then IMO what's in this patch is clearer. :shrug:

lmb: But the current description doesn't capture the quirkiness of LoadIndOp which loads from an sk_buff for cBPF compat.

ti-mo: Done.

PR #570, opened 2022-02-16
features: Expose KernelVersion function

brycekahle:


lmb:
There is a lint failure: https://ebpf.semaphoreci.com/jobs/b3859009-9587-4599-8d94-c01addc57277#L543

brycekahle:
@lmb the CI seems to be stuck on the 5.10 build

ti-mo:
@brycekahle Just curious, but what is this going to be used for in practice? Wouldn't you prefer more feature probes being exposed over a version number? I think we (and you) have established it's essentially meaningless with the backporting going on in RHEL and friends, so this is adding a footgun. If we do decide to add it, it should come with warning bells. :slightly_smiling_face:

brycekahle:
@ti-mo So we need this for a very specific reason. We are working on a feature to automatically download kernel headers. In order to validate that we have the correct headers to match the running kernel, we need to know what the `LINUX_VERSION_CODE` of the running kernel is, so we can compare it to the one from the header file.

ti-mo:
Superseded by https://github.com/cilium/ebpf/pull/585.

PR #571, opened 2022-02-16
features: centralize error semantics, remove maxMiscType, etc.

ti-mo:
Best reviewed per commit.

These are some changes I'd liked to see in #541, but did not have the bandwidth to review at the time. apologies.

The function documentation was getting a bit repetitive to my liking, I think it fits better in the package doc, with any exceptions (currently none) documented on the function in question.

I don't feel very strongly about `ErrUnsatisfiedProgramReference`, but it's potentially confusing during troubleshooting, so IMO it's better to be specific.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -560,31 +560,26 @@ func (insns Instructions) FixupReferences() error {
 			continue
 		}
 
-		symOffset, ok := symbolOffsets[ins.Reference]
 		switch {
 		case ins.IsFunctionReference() && ins.Constant == -1:
+			symOffset, ok := symbolOffsets[ins.Reference]
 			if !ok {
-				break
+				return fmt.Errorf("%s at insn %d: symbol %q: %w", ins.OpCode, i, ins.Reference, ErrUnsatisfiedProgramReference)
 			}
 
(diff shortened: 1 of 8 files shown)


asm/instruction.go
```diff
@@ -584,7 +584,7 @@ func (insns Instructions) FixupReferences() error {
 			// no fixup needed
 			continue
 		}
-		return fmt.Errorf("%s at insn %d: symbol %q: %w", ins.OpCode, i, ins.Reference, ErrUnsatisfiedReference)
+		return fmt.Errorf("%s at insn %d: symbol %q: %w", ins.OpCode, i, ins.Reference, ErrUnsatisfiedProgramReference)
```
lmb, line 587: I think this came about because I keep misreading the switch & loop. I'll push something that makes it a bit easier to read.

lmb: See https://github.com/cilium/ebpf/pull/571/commits/3880052a663a028fedeb5d4f3a0697201d629d97

lmb (approved):

asm/instruction.go
ti-mo: Sure, thanks :+1:

PR #572, opened 2022-02-16
Hide BTF wire format of FuncInfo and LineInfo

lmb:
Prompted by https://github.com/cilium/ebpf/pull/567/files#r805999952

Make it so that FuncInfo and LineInfo don't expose the BTF wire format outside of the package, and decode FileName, Line, etc. I used getters to avoid users being able to modify `LineInfo.FileName` and expecting that to show up in BTF.

Somewhat WIP since I need to write better commit descriptions.

cc @dylandreimerink

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -247,11 +247,30 @@ func (spec *Spec) splitExtInfos(info *extInfo) error {
 					return fmt.Errorf("section %s: error looking up FuncInfo for LineInfo %v", secName, li)
 				}
 
-				// Offsets are ELF section-scoped, make them function-scoped by
-				// subtracting the function's start offset.
-				li.InsnOff -= fnOffset
+				line, err := spec.strings.Lookup(li.LineOff)
+				if err != nil {
+					return fmt.Errorf("lookup of line: %w", err)
+				}
 
-				oli[fn.Name] = append(oli[fn.Name], li)
+				fileName, err := spec.strings.Lookup(li.FileNameOff)
+				if err != nil {
+					return fmt.Errorf("lookup of filename: %w", err)
+				}
+
+				lineNumber := li.LineCol >> bpfLineShift
+				lineColumn := li.LineCol & bpfColumnMax
+
+				oli[fn.Name] = append(oli[fn.Name], LineInfo{
+					fileName,
+					line,
+					lineNumber,
+					lineColumn,
+					// Offsets are ELF section-scoped, make them function-scoped by
+					// subtracting the function's start offset.
+					li.InsnOff - fnOffset,
+					li.FileNameOff,
+					li.LineOff,
+				})
 			}
 		}
 
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"os"
 	"reflect"
+	"sort"
 	"sync"
 
 	"github.com/cilium/ebpf/internal"
@@ -42,7 +43,7 @@ type Spec struct {
 	// Includes all struct flavors and types with the same name.
 	namedTypes map[essentialName][]Type
 
-	// Data from .BTF.ext.
+	// Data from .BTF.ext. indexed by function name.
 	funcInfos map[string]FuncInfo
 	lineInfos map[string]LineInfos
 	coreRelos map[string]CoreRelos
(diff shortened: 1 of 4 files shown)


internal/btf/ext_info.go
```diff
@@ -310,44 +273,85 @@ func parseFuncInfoRecords(r io.Reader, bo binary.ByteOrder, recordSize uint32, r
...
 // Marshal writes the binary representation of the LineInfo to w.
 // The instruction offset is converted from bytes to instructions.
-func (li LineInfo) Marshal(w io.Writer, offset uint64) error {
-	li.InsnOff += uint32(offset)
-	// The kernel expects offsets in number of raw bpf instructions,
-	// while the ELF tracks it in bytes.
-	li.InsnOff /= asm.InstructionSize
-	return binary.Write(w, internal.NativeEndian, li)
+func (li *LineInfo) Marshal(w io.Writer, offset uint64) error {
+	// TODO: Error out if lineNumber, lineColumn are out of bounds.
```
lmb, line 336: Also TODO

dylandreimerink: I don't think you have to worry about out-of-bound `lineNumber`s and `lineColumn`s, they are never used as array/slice index and verifying them would require you to inspect the original source code which might very well not be available on the machine loaded the ELF file.

lmb: True, I was thinking more along the lines of robustness when we let people generate their own context.

internal/btf/btf.go
```diff
@@ -193,67 +194,98 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
...
+					fileName,
+					line,
+					lineNumber,
+					lineColumn,
+					// Offsets are ELF section-scoped, make them function-scoped by
+					// subtracting the function's start offset.
+					li.InsnOff - fnOffset,
+					li.FileNameOff,
+					li.LineOff,
+				})
```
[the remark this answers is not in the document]
lmb: True, the downside of using named keys is that you don't get a compile error when extending a struct definition. Since extension happens more often than reordering I tend to prefer this style for internal types.

internal/btf/btf.go
```diff
@@ -193,67 +194,79 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
...
+			}
+
+			ofi[fn.Name] = FuncInfo{fn}
+			funcs = append(funcs, fn)
+		}
+
+		// Consider an ELF section that contains 3 functions (a, b, c)
+		// at offsets 0, 10 and 15 respectively. Offset 5 will return function a,
+		// offset 12 will return b, offset >= 15 will return c, etc.
+		sort.Slice(secFuncs, func(i, j int) bool {
```
ti-mo, line 227: Could we get away without this? They appear in the list here in the order they were read from .BTF.ext, which should already be ordered by offset. I'm fine with leaving this in, we're at the mercy of the compiler otherwise.

internal/btf/btf.go
```diff
@@ -193,67 +194,79 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
...
+		})
+		funcForOffset := func(offset uint32) (fn *Func, fnOffset uint32) {
+			for i, fi := range secFuncs {
+				if fi.InsnOff > offset {
+					break
+				}
+				fn = funcs[i]
+				fnOffset = fi.InsnOff
+			}
+			return
```
ti-mo, line 238: The naked return makes this more difficult to read, we don't have much code using this style.

internal/btf/btf.go
```diff
@@ -193,67 +194,79 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
 // transforms them to be indexed by function. Retrieves function names from
 // the BTF spec.
 func (spec *Spec) splitExtInfos(info *extInfo) error {
-
 	ofi := make(map[string]FuncInfo)
 	oli := make(map[string]LineInfos)
 	ocr := make(map[string]CoreRelos)
 
 	for secName, secFuncs := range info.funcInfos {
```
ti-mo, line 201: We should probably rename `secFuncs` to `secFuncInfos` with the addition of `funcs` below. This function didn't become easier to grok.

internal/btf/btf.go
```diff
@@ -193,67 +194,79 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
...
+			funcs = append(funcs, fn)
+		}
+
+		// Consider an ELF section that contains 3 functions (a, b, c)
+		// at offsets 0, 10 and 15 respectively. Offset 5 will return function a,
+		// offset 12 will return b, offset >= 15 will return c, etc.
+		sort.Slice(secFuncs, func(i, j int) bool {
+			return secFuncs[i].InsnOff < secFuncs[j].InsnOff
+		})
+		funcForOffset := func(offset uint32) (fn *Func, fnOffset uint32) {
```
ti-mo, line 230: Perhaps move the a,b,c comment from above here? Some of the context around the old `funcForOffset` was lost. Looking back, I should've probably named this `funcForInstruction`, too.

internal/btf/ext_info.go
```diff
@@ -303,14 +317,42 @@ type bpfLineInfo struct {
...
+func (li *LineInfo) FileName() string {
+	return li.fileName
+}
+
+func (li *LineInfo) Line() string {
+	return li.line
+}
+
+func (li *LineInfo) LineNumber() uint32 {
+	return li.lineNumber
```
ti-mo, line 329: Why not `return li.lineCol << bpfLineShift` and remove `lineNumber`?

internal/btf/ext_info.go
```diff
@@ -303,14 +317,42 @@ type bpfLineInfo struct {
...
+}
+
+func (li *LineInfo) LineColumn() uint32 {
+	return li.lineColumn
```
ti-mo, line 333: Why not `return li.lineCol & bpfColumnMax` and remove `lineColumn`?

ti-mo (changes requested):
Thanks! Have a few nits, and I think we can avoid having to store `lineNumber` and `lineColumn` as they can easily be calculated on access. Only `fileName` and `line` really need to be resolved up front as those require the string table.

internal/btf/btf.go
lmb: I went back and forth on this as well. Sorting here has two purposes: it "documents" that funcForInstructions relies on this ordering to work correctly and at the same time enforces that invariant. That seems worth it to me.

ti-mo: Yep, makes sense. :+1:

internal/btf/ext_info.go
ti-mo: What's the difference? These fields are unexported and should always be accessed using the getters.

ti-mo (approved):

PR #574, opened 2022-02-17
asm,features: unexport jump fixup, fix misc probes and tests

ti-mo:
See commit messages.

The ISA probes didn't match the bpftool ones and the test skips were all off by one kernel release.

### DIFF
--- a/features/misc_test.go
+++ b/features/misc_test.go
@@ -1,32 +1,31 @@
 package features
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/cilium/ebpf/internal/testutils"
 )
 
 func TestHaveMisc(t *testing.T) {
-	tests := map[miscType]struct {
+	tests := map[string]struct {
+		typ       miscType
 		probe     func() error
 		minKernel string
 	}{
-		largeInsn:    {probe: HaveLargeInstructions, minKernel: "5.1"},
-		boundedLoops: {probe: HaveBoundedLoops, minKernel: "5.2"},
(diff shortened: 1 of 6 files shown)


features/misc_test.go
```diff
@@ -1,32 +1,31 @@
...
+	tests := map[string]struct {
+		typ       miscType
 		probe     func() error
 		minKernel string
 	}{
-		largeInsn:    {probe: HaveLargeInstructions, minKernel: "5.1"},
-		boundedLoops: {probe: HaveBoundedLoops, minKernel: "5.2"},
-		v2ISA:        {probe: HaveV2ISA, minKernel: "4.13"},
-		v3ISA:        {probe: HaveV3ISA, minKernel: "5.0"},
+		"large instructions": {typ: largeInsn, probe: HaveLargeInstructions, minKernel: "5.2"},
```
lmb, line 15: Makes me think the probes should just have been `internal.FeatureTest` in the first place.

linker.go
```diff
@@ -115,10 +115,6 @@ func marshalLineInfos(layout []reference) ([]byte, error) {
 }
 
 func fixupJumpsAndCalls(insns asm.Instructions) error {
```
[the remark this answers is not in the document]
ti-mo: Yup, was considering the same. This is now `fixupAndValidate()` since the nil fd map pointer check has been moved there.

features/misc_test.go
ti-mo: This can still be moved later on, but I think over time we want to achieve the opposite. We want to expose as many of these as possible to the caller to provide `bpftool feature probe` in Go.

PR #575, opened 2022-02-18
examples: generate types via bpf2go

lmb:
examples: generate types via bpf2go
    
    Use bpf2go to generate a type for struct event. This requires an ugly hack to
    convince clang to emit struct event into the ELF in the first place.

examples/headers: add bpf_endian.h and LICENSE
    
    Stick to libbpf 0.6.1 for now, since 0.7.0 deprecates non-BTF
    map definitions. This requires removing the duplicate definition of PT_REGS_RC.
    
    Also guard struct pt_regs for amd64, the definition isn't valid on
    other arches. Finally, get rid of the __VMLINUX_H__ define, it trips up some
    logic in bpf_tracing.h.
    
    See https://github.com/libbpf/libbpf/commit/d788cd57b54c179a657f823cbe56c868f6974924


ti-mo (approved):
Now if bpf2go could emit efficient unmarshalers for types, that would be something. :stuck_out_tongue:

examples/headers/LICENSE.BSD-2-Clause
```diff
@@ -0,0 +1,32 @@
+Valid-License-Identifier: BSD-2-Clause
+SPDX-URL: https://spdx.org/licenses/BSD-2-Clause.html
+Usage-Guide:
+  To use the BSD 2-clause "Simplified" License put the following SPDX
+  tag/value pair into a comment according to the placement guidelines in
+  the licensing rules documentation:
+    SPDX-License-Identifier: BSD-2-Clause
+License-Text:
+
+Copyright (c) <year> <owner> . All rights reserved.
```
lmb, line 10: See libbpf/libbpf#461

PR #580, opened 2022-02-21
elf_reader: Fix cgroup/sock_{create,release} program type

joamaki:
The program type for cgroup/sock_{create,release} was CGroupSKB,
when it should've been CGroupSock. The section to type array now
matches with libbpf.


joamaki:
libbpf definitions for the prog types: https://github.com/libbpf/libbpf/blob/6edaacad4f7c31f08f63205e2b3569b9f946c8cb/src/libbpf.c#L8649

ti-mo:
Thanks!

PR #584, opened 2022-02-23
collection: don't copy btf.Spec in CollectionSpec.Copy()

ti-mo:
Copying btf.Spec requires updating Map/Program.BTF.(S,s)pec, which gets complicated due to btf.Map/Program being pointers.

Instead of implementing a full deep-copy, avoid copying btf.Spec for now, since it's considered read-only. Since we're planning to get rid of btf.Program/Map anyway, this would not be worth the investment.

Add a test that loads both a copy and an original CollectionSpec.

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -28,6 +28,7 @@ type CollectionSpec struct {
 	Programs map[string]*ProgramSpec
 
 	// Types holds type information about Maps and Programs.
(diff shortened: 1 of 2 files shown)


lmb:
Ci errors are real, you're missing a `testutils.SkipIfNotSupported`.

PR #587, opened 2022-03-02
asm: remove explicit metadata nil checks

ti-mo:
See individual commits. With... methods were split making navigation more difficult than it should be, so they were moved. It's time to split up the file, too.

Having to check for a nil metadata is tedious, error-prone and hurts readability. All naked metadata reads were replaced by nil-safe accessors so nil checks are not necessary. Metadata only needs to be copied and set if the passed-in value differs from the default field value, which significantly simplifies the `With*` methods.

fyi @dylandreimerink

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -39,31 +39,6 @@ type Instruction struct {
 	metadata *metadata
 }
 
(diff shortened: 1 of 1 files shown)


dylandreimerink:
Seems like a good improvement to the readability

PR #590, opened 2022-03-04
Execute and test CO-RE relocations of bpf2bpf subprogs

### LINKED ISSUE #586: Relocation is not called for some `BPF_CORE_READ`
Hi.

**Describe the bug**
With changes added in 9d739bdd0fe0fd0651ceb3e7bdcb0c9e9e4e92e1, the relocations sometimes does not occur.

**To Reproduce**
I created a reproducer which loads two eBPF code, one which works smoothly the other which shows the problem.
The only difference between the two is that in the one which works, 3 lines are commented.
You can test it with:

```bash
# In a first terminal:
$ git clone https://github.com/eiffel-fl/cilium-ebpf-problem-reproducer.git
$ cd cilium-ebpf-problem-reproducer
$ docker build -t mntns-erange .
$ docker run --rm --privileged -ti mntns-erange
# In a second terminal:
$ sudo cat /sys/kernel/debug/tracing/trace_pipe
# In a third terminal:
$ nc -lvp 9090                                                           
nc: getnameinfo: Temporary failure in name resolution
# Go back to second terminal.
# problem is the output of the problematic eBPF code, ret -34 is -ERANGE returned BPF_CORE_READ_INTO.
# It occurs because this BPF_CORE_READ_INTO is not rellocated at runtime.
nc-107741  [007] d...  9949.590175: bpf_trace_printk: problem: ret: -34; mntns_id: 0

nc-107741  [007] d...  9949.590191: bpf_trace_printk: ok: ret: 0; mntns_id: -268435456
```

**Expected behavior**
`BPF_CORE_READ_INTO` should not return `-ERANGE` because it should had been relocated.

ti-mo:
btf: include CO-RE relocations of bpf2bpf subprograms

> ProgramSpecs read from an ELF contain the full instruction streams from the
bpf2bpf subprograms they call. When marshaling a program to be inserted into
the kernel, FuncInfos and LineInfos are collected from all called subprogs,
but CO-RE relocations were notably skipped after the linker rework in https://github.com/cilium/ebpf/commit/9d739bdd0fe0fd0651ceb3e7bdcb0c9e9e4e92e1.
>
> This commit gathers CO-RE relos for all subprogs included in the final
instruction stream and modifies their offsets to point to their new locations.

btf: add CO-RE read test that fails unless relocations were correctly applied

> The linker rework in https://github.com/cilium/ebpf/commit/9d739bdd0fe0fd0651ceb3e7bdcb0c9e9e4e92e1 caused a piece of code that appended CORERelos
to the entrypoint instruction stream to be forgotten. The existing CO-RE tests
were all written in a way where the BPF programs would return 0 (success) if
no CO-RE relocations were applied at all.
>
> This caused CO-RE relocations against functions (subprogs) that were only
included by a bpf2bpf call to be skipped.
>
> This patch adds a broken test program that returns non-zero if left unaltered,
and requires CO-RE relocation against another BTF blob to return 0.
Additionally, the checks are performed from a non-inlined subprogram to ensure
CO-RE relocations against subprogs work.

---

@eiffel-fl could you verify if this solves your problem?

Fixes #586


eiffel-fl:
Hi.

I tested it with the initial tool I was testing and specifying `__attribute((noinline))` and it worked fine!
I also take a quick look to your code and it is cool!
Thank you for this :D :D :D

If you need test or review or anything, feel free to ping!

Best regards.

linker.go
```diff
@@ -80,6 +81,24 @@ func findReferences(progs map[string]*ProgramSpec) error {
 	return nil
 }
 
+// collectCoreRelos returns a list of CO-RE relocations of the layout's progs
+// in order.
+func collectCoreRelos(layout []reference) btf.CoreRelos {
+	if len(layout) == 0 {
```
lmb, line 87: Not required?

linker.go
```diff
@@ -80,6 +81,24 @@ func findReferences(progs map[string]*ProgramSpec) error {
 	return nil
 }
 
+// collectCoreRelos returns a list of CO-RE relocations of the layout's progs
+// in order.
+func collectCoreRelos(layout []reference) btf.CoreRelos {
```
lmb, line 86: Nit: we should decide on CORE or Core. I think CORE is more appropriate since it's an acronym.

linker.go
```diff
@@ -80,6 +81,24 @@ func findReferences(progs map[string]*ProgramSpec) error {
...
+	if len(layout) == 0 {
+		return nil
+	}
+
+	var relos btf.CoreRelos
+	for _, sym := range layout {
+		if sym.spec.BTF == nil {
+			continue
```
lmb, line 94: Do we support mixing btf and non btf programs otherwise? Seems like this could be an error condition instead.

lmb (approved):

linker.go
ti-mo: Yup, but then it needs changing across the board. Better now than never!

PR #594, opened 2022-03-09
asm: add Instruction Source metadata

ti-mo:
This commit adds Source metadata to Instruction, which is data about the origin of a series of Instructions.

A Source can be any fmt.Stringer. The ELF reader will associate corresponding btf.LineInfo objects with the Instruction.

Instructions.Format will now print the source information in-line with the Instruction on which it is set. This results in instructions being annotated with original source code when loaded from an ELF file, or custom annotations if custom Stringers are assigned using the Instruction.WithSource.

---

Leaving this here as work in progress because the LineInfo unmarshaler should be reverted to operate per section instead of per function. lineInfos are encoded per section, so it's a bit unnecessary to split during unmarshaling and join for instruction tagging.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -402,6 +402,23 @@ func (ins Instruction) Map() FDer {
 	return ins.metadata.Map()
 }
 
+// WithSource adds source information about the Instruction.
+func (ins Instruction) WithSource(src fmt.Stringer) Instruction {
+	if ins.Source() == src {
+		return ins
+	}
+
+	ins.copyMetadata().source = src
+	return ins
+}
+
+// Source returns source information about the Instruction. The field is
+// present when the compiler emits BTF line info about the Instruction and
+// usually contains the line of source code responsible for it.
+func (ins Instruction) Source() fmt.Stringer {
+	return ins.metadata.Source()
+}
+
 // copyMetadata is a convenience method for copying ins.metadata, assigning
 // the new copy to its metadata field and returning a pointer to the copy
 // so one access can be chained.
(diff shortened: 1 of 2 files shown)


ti-mo:
The groundwork for adding this to `loadFunctions` has been done. I'm planning to pull LineInfos into `elfSection` indexed by raw insn count, so that we can pull them in as metadata in the iterator in `loadFunctions` and remove them from btf.Program.

@dylandreimerink One issue here is that it won't be possible to load programs with caller-provided `Stringer`s into the kernel yet, because we haven't yet implemented BTF marshaling. LineInfos point into the main BTF string table, but we can't yet insert our own entries.

Do you need this piece of DSL to land for edb? Worst case, we could document that `WithSource` doesn't technically cause its inputs to end up in the kernel yet, and remove the warning when BTF marshaling lands.

lmb:
If I remember correctly, we settled on collecting all the .Source() that are btf.LineInfo, so no custom line info in the kernel yet.

dylandreimerink:
> The groundwork for adding this to loadFunctions has been done. I'm planning to pull LineInfos into elfSection indexed by raw insn count, so that we can pull them in as metadata in the iterator in loadFunctions and remove them from btf.Program.

That seems fair. Once all of this is in, I am hoping to also add FuncInfo to the metadata, so it would be nice if that is also available in `elfSection`.

> One issue here is that it won't be possible to load programs with caller-provided Stringers into the kernel yet, because we haven't yet implemented BTF marshaling. LineInfos point into the main BTF string table, but we can't yet insert our own entries.

Yes, we still have some work ahead of us to get marshaling to fully work, likely in a number of small steps to keep diffs manageable. 

> Do you need this piece of DSL to land for edb? 

For `edb` the most important feature is `(Instruction) Source() fmt.Stringer`. I am not planning on using `WithSource` until marshalling is fully implemented.

> If I remember correctly, we settled on collecting all the .Source() that are btf.LineInfo, so no custom line info in the kernel yet.

Yes, that is how I recall it as well. The idea was to just use the `Stringer.String()` result as the `Line` when marshalling, unless it specifically is a `*btf.LineInfo`.

lmb:
Superseded by #606

PR #605, opened 2022-03-24
kprobe: fall back to tracefs if kernel doesn't support dots in PMU symbol

chenhengqi:
Centos 8 (kernel 4.18) has kprobe PMU, but don't allow `.` in symbol name,
add a new option to workaround this.

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strings"
 	"sync"
 	"unsafe"
 
@@ -263,6 +264,12 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 
 	rawFd, err := unix.PerfEventOpen(&attr, args.pid, 0, -1, unix.PERF_FLAG_FD_CLOEXEC)
 
+	// On some old kernels, kprobe PMU doesn't allow `.` in symbol names and
+	// return -EINVAL. Return ErrNotSupported to allow falling back to tracefs.
+	// https://github.com/torvalds/linux/blob/94710cac0ef4/kernel/trace/trace_kprobe.c#L340-L343
+	if errors.Is(err, unix.EINVAL) && strings.Contains(args.symbol, ".") {
+		return nil, fmt.Errorf("symbol '%s': older kernels don't accept dots: %w", args.symbol, ErrNotSupported)
+	}
 	// Since commit 97c753e62e6c, ENOENT is correctly returned instead of EINVAL
 	// when trying to create a kretprobe for a missing symbol. Make sure ENOENT
 	// is returned to the caller.
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -385,7 +385,7 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		// subsampling or rate limiting logic can be more accurately implemented in
 		// the eBPF program itself.
 		// See Documentation/kprobes.txt for more details.
-		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, args.symbol, args.symbol)
+		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, sanitizedSymbol(args.symbol), args.symbol)
 	case uprobeType:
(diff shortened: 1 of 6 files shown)


link/kprobe.go
```diff
@@ -49,6 +49,13 @@ type KprobeOptions struct {
 	//
 	// Needs kernel 5.15+.
 	Cookie uint64
+
+	// Fallback to legacy tracefs kprobe interface
+	// event if kernel has kprobe PMU
+	//
+	// This allows old kernels which has kprobe PMU
+	// but don't support `.` in symbol names to work properly
```
mmat11, line 57: are there kernels that supports PMU and don't support `.` in the symbol name and kernels that support PMU and `.` in the symbol name? (I am pretty sure on my kernel (5.16.x) I can attach uprobes PMU with `.` in the symbol)

If yes, what about implementing some feature detection to do this transparently? (kernel supports PMU && kernel doesn't support `.` -> go for tracefs)

link/kprobe.go
```diff
@@ -173,7 +180,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 	if err == nil {
 		return tp, nil
 	}
-	if err != nil && !errors.Is(err, ErrNotSupported) {
+	if err != nil && !errors.Is(err, ErrNotSupported) && !opts.Legacy {
```
mmat11, line 183: opts can be `nil` here

link/kprobe.go
chenhengqi: > are there kernels that supports PMU and don't support . in the symbol name and kernels that support PMU and . in the symbol name?

Centos 8 with kernel v4.18. Did not find the related commit.

chenhengqi: > If yes, what about implementing some feature detection to do this transparently? (kernel supports PMU && kernel doesn't support . -> go for tracefs)

Don't know how to reliably implement such detection. :(

lmb:
@chenhengqi how does the centos kernel fail when `.` is in the name? In the best case we could add an `if errors.Is(err, EFOO)` case that enables the already existing tracefs fallback.

chenhengqi:
> @chenhengqi how does the centos kernel fail when `.` is in the name? In the best case we could add an `if errors.Is(err, EFOO)` case that enables the already existing tracefs fallback.

The version of kernel shipped in CentOS 8 Linux is 4.18.

In 4.18, the kernel don't allow dot in symbol name in kprobe PMU case:

https://github.com/torvalds/linux/blob/94710cac0ef4ee177a63b5227664b38c95bbf703/kernel/trace/trace_kprobe.c#L340-L343

Can we always fallback to tracefs interface ? Or check the error against EINVAL ?

ti-mo:
No need for feature detection IMO, falling back under a specific error like Lorenz suggested sounds like the best option to me. Is this possible?

chenhengqi:
OK, just updated the code as you suggest.

link/kprobe.go
```diff
@@ -335,7 +343,7 @@ func tracefsProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 	}
 
 	// Get the newly-created trace event's id.
-	tid, err := getTraceEventID(group, args.symbol)
+	tid, err := getTraceEventID(group, sanitizedSymbol(args.symbol))
```
ti-mo, line 346: Is this still necessary with the fallback to tracefs? I'm a bit foggy on this, but doesn't this modify the symbol name in a way that it no longer matches the kernel's? Couldn't we retry a PMU probe with a sanitized symbol in that case?

chenhengqi: Yes, on line L396, we have to sanitize the `EVENT` part of kprobe_event.

ti-mo: Okay, so this means the 'name' part of the event. That makes sense.

ti-mo:
I've added a missing call to `sanitizedSymbol()` in `closeTraceFSProbeEvent()`. Closing a tracefs event created with a sanitized name would of course fail otherwise.

I've also converted `TestK(ret)probe` to be table-driven tests and found some kernel symbols with local visibility that were present in kernel versions ranging from 4.4 to 5.15, so this fallback behaviour can actually be exercised.

chenhengqi:
Thanks.

Issue #550, opened 2022-01-25
Support custom "kernel BTF" locations


joamaki:
For projects that ship with BTF for kernels without `CONFIG_DEBUG_INFO_BTF` it would useful
to be able to point the library to a custom location, so that a project does not need to invasively
symlink/copy/mount the BTF file into a known location.

Current alternative is to set `TargetBTF` with `NewCollectionWithOptions`, but this takes an `io.Reader`,
and requires parsing the BTF file for each collection that's loaded, which is rather memory and CPU intensive.

Would be great if this could be solved, either by allowing giving hints about the BTF location, or by having `TargetBTF`
take a parsed `btf.Spec`.

lmb:
My preference would be to make it a `btf.Spec`. We haven't done this so far since users don't have access to `btf`, but that is something we should change anyways.

I know that @ti-mo has ideas about moving some parts of `btf` into a public package.

ti-mo:
We'll most likely end up exporting the BTF parser and making the lib take a `btf.Spec` or similar where applicable.

One point to add: we need to add support for parsing kernel module BTF (e.g. `/sys/kernel/btf/nf_conntrack`), since those refer to types declared in vmlinux.

lmb:
I'm guessing the data in /sys only shows up after doing a modprobe, which begs the question how we could test this in CI. We don't build / ship modules at the moment to keep it simple, but we might have to change that.

ti-mo:
Indeed, they show up after modprobes. We could always ship a few BTF blobs (including vmlinux) from `/sys/kernel/btf` from a modules-enabled system in the repo. The downside is vmlinux is a few megabytes big. but this might not be that much of a dealbreaker.

This would also make benchmarking BTF unmarshaling in a reproducible way a bit more straightforward.

lmb:
We already have vmlinux btf, and use it for benchmarks via readVMLinux​(​b​). Is that what you mean?

ti-mo:
Ah, turns out we do! Then adding one or two kernel module BTFs shouldn't pose an issue at all.

PR #606, opened 2022-03-28
Use instruction metadata to simplify linking and CO-RE relocation

lmb:
btf: move CORERelocate to core.go

btf: add constructors for various *Info

    Pull constructors for FuncInfo, LineInfo and CORERelocation out of
    splitExtInfos. This makes the function smaller and makes removing it
    later easier.

    Instead of taking a *Spec in the constructors, take a []Type and
    stringTable directly. This makes splitting ExtInfos out of *Spec
    easier later on.

btf: track bpfCORERelo.InsnOff in instructions instead of bytes

    This makes bpfCORERelo consistent with bpfFuncInfo and bpfLineInfo. It seems like
    there is a latent bug here as well, since we used to subtract bpfFuncInfo.InsnOff
    from bpfCORERelo.InsnOff, but for some reason the tests pass.

btf: store ExtInfos in instruction metadata

    Use per-instruction metadata to store FuncInfo, LineInfo and CO-RE relocations.
    As a result, simply concatenating two Instruction slices preserves the ext_info
    without extra code to keep track of offsets. The simplest way to achieve is to
    assign ext_infos per section, before we split into individual functions. This
    also avoids having to split ext_infos in the first place.

    Storing ext_infos in metadata in turn allows / requires removing code that relies
    on stable offsets, the most notable being applying COREFixups. Instead of tracking
    which offset a fixup should be applied to we change coreRelocate to instead return
    results in the same order as CORERelocations are passed. In the caller we remember
    which instruction originated the relocation and apply the fixup directly instead of
    iterating all instructions once more.

    Instead of storing ExtInfos in Spec we split it off and keep a reference to the
    necessary types and string table around. This makes more sense since every BTF ELF
    has a Spec, but ExtInfos are optional. It turns out that most code except the ELF
    loader doesn't care about ExtInfos in the first place, so we might be able to
    add an API that doesn't parse ExtInfos in the future.

    Finally we can remove btf.Program since we don't need an intermediate type to hold
    metadata anymore.

btf: remove Map

    btf.Program is gone, so let's get rid of Map as well.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -38,7 +38,7 @@ type Spec struct {
 	strings  *stringTable
 
 	// Inflated Types.
-	types []Type
+	types types
 
 	// Types indexed by essential name.
 	// Includes all struct flavors and types with the same name.
@@ -205,24 +205,14 @@ func (spec *Spec) splitExtInfos(info *extInfo) error {
 	for secName, secFuncInfos := range info.funcInfos {
 		// Collect functions from each section and organize them by name.
 		var funcs []*Func
-		for _, fi := range secFuncInfos {
-			typ, err := spec.TypeByID(fi.TypeID)
+		for _, bfi := range secFuncInfos {
+			fi, err := newFuncInfo(bfi, spec.types)
 			if err != nil {
 				return err
 			}
 
-			fn, ok := typ.(*Func)
-			if !ok {
-				return fmt.Errorf("type ID %d is a %T, but expected a Func", fi.TypeID, typ)
-			}
-
-			// C doesn't have anonymous functions, but check just in case.
-			if fn.Name == "" {
-				return fmt.Errorf("func with type ID %d doesn't have a name", fi.TypeID)
-			}
-
-			ofi[fn.Name] = FuncInfo{fn}
-			funcs = append(funcs, fn)
+			ofi[fi.fn.Name] = *fi
+			funcs = append(funcs, fi.fn)
 		}
 
 		sort.Slice(secFuncInfos, func(i, j int) bool {
@@ -245,52 +235,43 @@ func (spec *Spec) splitExtInfos(info *extInfo) error {
 
 		// Attribute LineInfo records to their respective functions, if any.
 		if lines := info.lineInfos[secName]; lines != nil {
-			for _, li := range lines {
-				fn, fnOffset := funcForInstruction(li.InsnOff)
+			for _, bli := range lines {
+				fn, fnOffset := funcForInstruction(bli.InsnOff)
 				if fn == nil {
-					return fmt.Errorf("section %s: error looking up FuncInfo for LineInfo %v", secName, li)
+					return fmt.Errorf("section %s: error looking up FuncInfo for offset %v", secName, bli.InsnOff)
 				}
 
-				line, err := spec.strings.Lookup(li.LineOff)
+				li, err := newLineInfo(bli, spec.strings)
 				if err != nil {
-					return fmt.Errorf("lookup of line: %w", err)
+					return err
 				}
 
-				fileName, err := spec.strings.Lookup(li.FileNameOff)
-				if err != nil {
-					return fmt.Errorf("lookup of filename: %w", err)
-				}
+				// Offsets are ELF section-scoped, make them function-scoped by
+				// subtracting the function's start offset.
+				li.insnOff -= fnOffset
 
-				lineNumber := li.LineCol >> bpfLineShift
-				lineColumn := li.LineCol & bpfColumnMax
-
-				oli[fn.Name] = append(oli[fn.Name], LineInfo{
-					fileName,
-					line,
-					lineNumber,
-					lineColumn,
-					// Offsets are ELF section-scoped, make them function-scoped by
-					// subtracting the function's start offset.
-					li.InsnOff - fnOffset,
-					li.FileNameOff,
-					li.LineOff,
-				})
+				oli[fn.Name] = append(oli[fn.Name], *li)
 			}
 		}
 
 		// Attribute CO-RE relocations to their respective functions, if any.
 		if relos := info.relos[secName]; relos != nil {
 			for _, r := range relos {
-				fn, fnOffset := funcForInstruction(r.insnOff)
+				fn, fnOffset := funcForInstruction(r.InsnOff)
 				if fn == nil {
-					return fmt.Errorf("section %s: error looking up FuncInfo for CO-RE relocation %v", secName, r)
+					return fmt.Errorf("section %s: error looking up FuncInfo for offset %v", secName, r.InsnOff)
+				}
+
+				relo, err := newCORERelocation(r, spec.types, spec.strings)
+				if err != nil {
+					return err
 				}
 
 				// Offsets are ELF section-scoped, make them function-scoped by
 				// subtracting the function's start offset.
-				r.insnOff -= fnOffset
+				relo.insnOff -= fnOffset
 
-				ocr[fn.Name] = append(ocr[fn.Name], r)
+				ocr[fn.Name] = append(ocr[fn.Name], *relo)
 			}
 		}
 	}
@@ -641,10 +622,7 @@ func (s *Spec) Program(name string) (*Program, error) {
 // Returns an error wrapping ErrNotFound if a Type with the given ID
 // does not exist in the Spec.
 func (s *Spec) TypeByID(id TypeID) (Type, error) {
-	if int(id) > len(s.types) {
-		return nil, fmt.Errorf("type ID %d: %w", id, ErrNotFound)
-	}
-	return s.types[id], nil
+	return s.types.ByID(id)
 }
 
 // AnyTypesByName returns a list of BTF Types with the given name.
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -689,12 +689,6 @@ func (h *Handle) FD() int {
 	return h.fd.Int()
 }
 
-// Map is the BTF for a map.
-type Map struct {
-	Spec       *Spec
-	Key, Value Type
-}
-
 func marshalBTF(types interface{}, strings []byte, bo binary.ByteOrder) []byte {
 	const minHeaderLength = 24
 
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -848,25 +848,6 @@ func (p *Program) Spec() *Spec {
 	return p.spec
 }
 
-// CORERelocate returns the changes required to adjust the program to the target.
-//
-// Passing a nil target will relocate against the running kernel.
-func CORERelocate(local, target *Spec, relos CORERelos) (COREFixups, error) {
-	if len(relos) == 0 {
-		return nil, nil
-	}
-
-	if target == nil {
-		var err error
-		target, err = LoadKernelSpec()
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	return coreRelocate(local, target, relos)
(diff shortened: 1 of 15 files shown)


lmb:
Split off the `WithSource` + sparse metadata changes into #609

lmb:
I'll move `program: remove references from ProgramSpec` to a follow up PR once #609 is merged.

lmb:
This is now in a state where it can be reviewed! I've updated the PR description with the latest commit messages, please take a look.

I'm a little bit unhappy with `ExtInfos.Assign`, it seems like it belongs into the ELF loader. I'd prefer to convert bpfFuncInfo to FuncInfo, etc. at BTF load time and then expose something like the following:

```
type ExtInfos struct{
    []FuncInfo
    []LineInfo
    []CORERelocation
}

// Per-section ExtInfos
func LoadSpecFromReader(...) (*Spec ,map[string]ExtInfos, error)
```

The snag is that `FuncInfo`, etc. don't contain the InsnOff, since having an offset doesn't make sense when they are attached to an instruction. I can see two ways to fix this:

```
type ExtInfo struct {
    map[uint64]FuncInfo
    ...
}
```

By abusing the map we can avoid introducing a new type, but none of our code really wants to use a map. An ordered slice is much better. For that we'd need:

```
type FuncInfoWithOffset struct {
    FuncInfo
    Offset uint32
}

type ExtInfo struct {
    []FuncInfoWithOffset
    ...
}
```

This means a lot of additional types, which I don't particularly like. Maybe you have another idea?

internal/btf/core.go
```diff
@@ -197,35 +157,84 @@ func (k coreKind) String() string {
 	}
 }
 
-func coreRelocate(local, target *Spec, relos CORERelos) (COREFixups, error) {
+// ApplyRelocations calculates the difference in types between local
+// and target and adjusts insns to match.
+//
+// Passing a nil target will relocate against the running kernel. insns are
+// modified in place.
+func ApplyRelocations(local, target *Spec, insns asm.Instructions) error {
```
ti-mo, line 165: Put insns as the first argument to signal that it's being modified in-place?

internal/btf/core.go
```diff
@@ -197,35 +157,84 @@ func (k coreKind) String() string {
...
+	var relos []*CORERelocation
+	var reloInsns []*asm.Instruction
+	iter := insns.Iterate()
+	for iter.Next() {
```
ti-mo, line 169: Could coreRelocate somehow be refactored to operate on a single Instruction? That way, getting relo from metadata, calculating the fixup and applying it can be done in a single loop without copying much. Not speaking about performance, but it might shorten this function and make it easier to follow. I haven't fully understood `coreRelocate()` though, just a suggestion.

internal/btf/btf.go
```diff
@@ -78,21 +72,33 @@ func (h *btfHeader) stringStart() int64 {
 
 // LoadSpecFromReader reads from an ELF or a raw BTF blob.
 //
-// Returns ErrNotFound if reading from an ELF which contains no BTF.
-func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
+// Returns ErrNotFound if reading from an ELF which contains no BTF. ExtInfos
+// may be nil.
+func LoadSpecFromReader(rd io.ReaderAt) (*Spec, *ExtInfos, error) {
```
ti-mo, line 77: I've seen you refer to eliminating `*ExtInfos` here in the future in your commit message, but can we get away with doing that sooner rather than later? Could we put `*ExtInfos` (back?) into the `btf.Spec` instead of returning it as a separate value here? IMO `LoadSpec*` should return a single value.

ti-mo (commented):
Looks great, have a few first observations that might or might not apply, let me know what you think.

internal/btf/core.go
```diff
@@ -197,35 +157,84 @@ func (k coreKind) String() string {
...
+	for i, relo := range relos {
 		if relo.kind == reloTypeIDLocal {
 			// Filtering out reloTypeIDLocal here makes our lives a lot easier
 			// down the line, since it doesn't have a target at all.
 			if len(relo.accessor) > 1 || relo.accessor[0] != 0 {
 				return nil, fmt.Errorf("%s: unexpected accessor %v", relo.kind, relo.accessor)
 			}
 
-			result[uint64(relo.insnOff)] = COREFixup{
+			result[i] = COREFixup{
```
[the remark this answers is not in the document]
lmb: This is libbpf behaviour, since `relo.kind == reloTypeIDLocal` we need an ID.

internal/btf/core.go
lmb: Unfortunately that doesn't work. coreRelocate splits all relocations into per-type relocations, and then uses the per-type relocations as "constraints" to find the best target type. That isn't possible when looking at a single instruction (and therefore relocation) in isolation.

internal/btf/btf.go
lmb: I'd like to remove ExtInfos completely, but https://github.com/cilium/ebpf/pull/606#issuecomment-1104201245 is my best idea how to do it and I don't like it very much.

Embedding ExtInfos in Spec has it the wrong way round: when reading BTF we always have a Spec, but we don't always have ExtInfos. So modeling the relationship as ExtInfo{Spec} is more natural. For example, naked vmlinux BTF has no ext infos. Check the call sites of `loadRawSpec`, almost always we pass `nil, nil` and aren't interested in ExtInfos in the first place. With ExtInfos split out / removed a good amount of code can be refactored to be less deeply nested.

I was considering adding a separate constructor `LoadExtInfos(io.Reader)` and making `LoadSpec` skip ExtInfos. However I can't come up with a way to achieve that without (a) adding `*elf.File` to the `btf` API surface like `LoadExtInfos(*elf.File)` (and all the SafeELFFile baggage) or (b) parsing the ELF one more time. (We already parse the ELF header twice in the ELF loader due to this.)

lmb: Well, check this out: https://github.com/cilium/ebpf/commit/640c4b0385e39775010880e92bd7711797e39e73 Kinda terrible, but we could use it to have `LoadExtInfosFromReader(io.ReaderAt)` and `LoadSpecFromReader(io.ReaderAt)`.

internal/btf/core.go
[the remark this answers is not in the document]
lmb: This also creates problems for #643 of course.

internal/btf/core.go
```diff
@@ -197,35 +157,84 @@ func (k coreKind) String() string {
...
+func coreRelocate(local, target *Spec, relos []*CORERelocation) ([]COREFixup, error) {
 	if local.byteOrder != target.byteOrder {
 		return nil, fmt.Errorf("can't relocate %s against %s", local.byteOrder, target.byteOrder)
 	}
 
 	var ids []TypeID
-	relosByID := make(map[TypeID]CORERelos)
-	result := make(COREFixups, len(relos))
-	for _, relo := range relos {
+	relosByID := make(map[TypeID][]*CORERelocation)
```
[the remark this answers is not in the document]
lmb: This turned out nice, thanks!

lmb:
As a follow up to https://github.com/cilium/ebpf/pull/606#issuecomment-1104201245, what if ExtInfos becomes this:

```
type ExtInfos struct{
    Relocations []*CORERelocations
    RelocationOffsets []uint32
    ...
}

func Foo(...) (map[string]*ExtInfos, error)
```

internal/btf/btf.go
ti-mo: Regarding ExtInfos: could we somehow break out that review and discussion into a separate PR? Taking `*elf.File` is not the way to go for the reasons you mentioned, but for me primarily due to #550.

If we're reworking the btf/extinfo reader API, we should definitely take the opportunity to address this current limitation and come to a solution that works for module BTF as well. Actually parsing (and 'linking') module BTF might be a bigger chunk of work, but at least the API should allow for it to be added later if needed.

internal/btf/core.go
ti-mo: Okay, is this due to the way CO-RE is supposed to work, or just a decision that was made while doing the initial implementation?

lmb: My initial implementation operated on a single instruction / relocation, and then failed some of the libbpf tests. As a result I changed the approach to what we have right now. It's not the same as what libbpf does (I think) but I'm fairly certain that only looking at a single relocation doesn't work.

lmb:
Please take a look at https://github.com/cilium/ebpf/pull/606/commits/64dbe60367826214020d9600f599d6038aadf3c0:

- I've gotten rid of ApplyRelocations and ExtInfos.Assign and MarshalExtInfos and moved them into linker.go instead. This is much closer to what we had before. This is nice because instruction metadata is now private to the ebpf package.
- FuncInfo, LineInfo, CORERelocationInfo are now used to expose "metadata + offset" via `ExtInfos`. The real metadata is called `Func` (as you suggested), `Line` and `CORERelocation`.
- I've split the constuctor for Spec in two, with one returning ExtInfos along with the Spec. The nice side effect is that we don't decode useless ExtInfo when loading BTF via `handleCache`.

internal/btf/core.go
ti-mo: I'm curious to know why this is, but I might need to further my understanding of CO-RE first. We'll leave this as-is now.

internal/btf/btf.go
```diff
@@ -94,6 +90,29 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 	return loadSpecFromELF(file)
 }
 
+// LoadSpecAndExtInfosFromReader reads from an ELF.
+//
+// ExtInfos may be nil if the ELF doesn't contain section metadta.
+// Returns ErrNotFound if the ELF contains no BTF.
+func LoadSpecAndExtInfosFromReader(rd io.ReaderAt) (*Spec, *ExtInfos, error) {
```
ti-mo, line 97: Following the `ExtInfos{Spec}` relationship you suggested, following suit would make sense:

```suggestion
func LoadExtInfosFromReader(rd io.ReaderAt, spec *Spec) (*ExtInfos, error) {
```

internal/btf/core.go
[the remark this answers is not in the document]
ti-mo: Yes, this is probably something that needs solving in order to get rid of IDs.

linker.go
```diff
@@ -111,50 +111,100 @@ func findReferences(progs map[string]*ProgramSpec) error {
...
+	return nil
 }
 
-// marshalFuncInfos returns the BTF func infos of all progs in order.
-func marshalFuncInfos(layout []reference) ([]byte, error) {
-	if len(layout) == 0 {
-		return nil, nil
+// marshalExtInfos encodes function and line info embedded in insns into kernel
+// wire format.
+func marshalExtInfos(insns asm.Instructions) (funcInfos, lineInfos []byte, _ error) {
```
ti-mo, line 146: Could this live in package btf, or is there an issue with a dependency on `asm`?

My hope was to get rid of FuncInfo and LineInfo, since all the information they carry can now be represented by `asm.Instructions`. e.g. *Func and the instruction's position in the slice as you've shown below.

Ultimately, we should end up with only Line and Func, and their respective wire formats, without the `Func/LineInfo` intermediates.

ti-mo (approved):
Exciting! Left a few more comments. Should the latest commit be squashed?

internal/btf/btf.go
lmb: That was my initial plan, but this would allow using a Spec loaded from a different ELF.

PR #609, opened 2022-03-28
asm: add Instruction Source metadata

lmb:
asm: add Instruction Source metadata

    This commit adds Source metadata to Instruction, which is data about
    the origin of a series of Instructions. A Source can be any fmt.Stringer.

    Instructions.Format will now print the source information in-line with the
    Instruction on which it is set. This results in instructions being annotated
    with original source code when loaded from an ELF file, or custom annotations
    if fmt.Stringers are ssigned using Instruction.WithSource.

asm: use sparse metadata

    Replace the struct based metadata with a linked list. This reduces the amount of memory
    needed when the metadata is sparse, aka when there are more different kinds of metadata.
    This also allows packages to have their own private metadata, which we'll use to store
    BTF ext_infos.

    The most common operation on metadata is Get, since we need to check for the presence of
    multiple bits of metadata for every single instruction. This is captured by the get_miss
    benchmark. The second most operation will be repeated calls to Set. This is captured
    by the add_first and add_last benchmarks.

    My initial implementation used a map[interface{}]interface{} to store metadata, but it
    turns out that a linked list is more efficient. When using a map, each additional call
    to Set has to copy all previously set keys. We can avoid this with the linked list since
    we can simply prepend a new item.

    Timo proposed an implementation based on a slice, but this too suffers from having to copy
    previous KVs.

        name \ time/op           map.txt      slice.txt    ll.txt
        Metadata/add_first-4     97.6ns ± 2%  48.4ns ± 0%  25.1ns ± 2%
        Metadata/add_last-4       162ns ± 1%   104ns ± 1%    27ns ± 1%
        Metadata/add_existing-4  15.0ns ± 0%   9.0ns ± 0%   7.2ns ± 1%
        Metadata/get_miss-4      8.10ns ± 0%  3.73ns ± 0%  2.80ns ± 0%

        name \ alloc/op          map.txt      slice.txt    ll.txt
        Metadata/add_first-4       336B ± 0%     56B ± 0%     48B ± 0%
        Metadata/add_last-4        336B ± 0%    216B ± 0%     48B ± 0%
        Metadata/add_existing-4   0.00B        0.00B        0.00B
        Metadata/get_miss-4       0.00B        0.00B        0.00B

        name \ allocs/op         map.txt      slice.txt    ll.txt
        Metadata/add_first-4       2.00 ± 0%    2.00 ± 0%    1.00 ± 0%
        Metadata/add_last-4        2.00 ± 0%    3.00 ± 0%    1.00 ± 0%
        Metadata/add_existing-4    0.00         0.00         0.00
        Metadata/get_miss-4        0.00         0.00         0.00

elf: add simple benchmark

### DIFF
--- a/asm/metadata.go
+++ b/asm/metadata.go
@@ -1,0 +1,80 @@
+package asm
+
+// Metadata contains metadata about an instruction.
+type Metadata struct {
+	head *metaElement
+}
+
+type metaElement struct {
+	next       *metaElement
+	key, value interface{}
+}
+
+// Find the element containing key.
+//
+// Returns nil if there is no such element.
+func (m *Metadata) find(key interface{}) *metaElement {
+	for e := m.head; e != nil; e = e.next {
+		if e.key == key {
+			return e
+		}
+	}
+	return nil
+}
+
+// Remove an element from the linked list.
+//
+// Copies as many elements of the list as necessary to remove r, but doesn't
+// perform a full copy.
+func (m *Metadata) remove(r *metaElement) {
+	current := &m.head
+	for e := m.head; e != nil; e = e.next {
+		if e == r {
+			// We've found the element we want to remove.
+			*current = e.next
+
+			// No need to copy the tail.
+			return
+		}
+
+		// There is another element in front of the one we want to remove.
+		// We have to copy it to be able to change metaElement.next.
+		cpy := &metaElement{key: e.key, value: e.value}
+		*current = cpy
+		current = &cpy.next
+	}
+}
+
+// Set a key to a value.
+//
+// If value is nil, the key is removed. Avoids modifying old metadata by
+// copying if necessary.
+func (m *Metadata) Set(key, value interface{}) {
+	if e := m.find(key); e != nil {
+		if e.value == value {
+			// Key is present and the value is the same. Nothing to do.
+			return
+		}
+
+		// Key is present with a different value. Create a copy of the list
+		// which doesn't have the element in it.
+		m.remove(e)
+	}
+
+	// m.head is now a linked list that doesn't contain key.
+	if value == nil {
+		return
+	}
+
+	m.head = &metaElement{key: key, value: value, next: m.head}
+}
+
+// Get the value of a key.
+//
+// Returns nil if no value with the given key is present.
+func (m *Metadata) Get(key interface{}) interface{} {
+	if e := m.find(key); e != nil {
+		return e.value
+	}
+	return nil
+}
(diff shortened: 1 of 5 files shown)


asm/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+	for e := m.head; e != nil; e = e.next {
+		if e == r {
+			// We've found the element we want to remove. No need to copy the tail.
+			*prev = e.next
+			break
+		}
+
+		cpy := &metaElement{key: e.key, value: e.value}
+		*prev = cpy
+		prev = &cpy.next
```
ti-mo, line 40: This line only seems to be hit once during the test suite, and the tests pass if it's removed. Could we add another test to exercise this better? (e.g. removing an element from a list where more than 1 element is copied?)

This is my understanding of how it works: after a copy, `prev` is supposed to point at the slot the next item will occupy, iow; it's supposed to point at a nil pointer initially. This fact is not 100% clear, but then I'm not super familiar with the data structure.

Nit: I think the name 'prev' is a bit misleading here, it's more like the 'current' item being operated on. My confusion during the first read-through was more due to lack of context, so just adding some comments about what each variable's meaning is would help a lot. :+1:

asm/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+func (m *Metadata) find(key interface{}) *metaElement {
+	for e := m.head; e != nil; e = e.next {
+		if e.key == key {
+			return e
+		}
+	}
+	return nil
+}
+
+// Remove an element from the linked list.
```
ti-mo, line 25: I would briefly document that this function is where the copying takes place.

ti-mo (commented):
`.remove()` and `.Set()` aren't super easy to understand, but that's partially a layer 8 problem.

I'm wondering if we could fold the `value == nil` case into `m.remove()` so we can avoid the fallthrough behaviour and make it more concise.

I already like it in its current state, though. It's barely more complex than a list-based implementation, and the efficiency is a really nice bonus.

asm/metadata.go
lmb: Done.

asm/metadata.go
lmb: Renamed to current.

lmb: Also extended the test to have more bits of metadata.

lmb:
> I'm wondering if we could fold the value == nil case into m.remove() so we can avoid the fallthrough behaviour and make it more concise.

I looked into this, and it's not more straight forward. I think of `Set` in two stages: first, obtain a `head` that doesn't contain an element with `key` in it. Second, depending on `value`, prepend an element or use the new head directly.

lmb:
I've rejigged `Set` once more, I think it's now as clear as it'll get. Let me know what you think!

PR #612, opened 2022-03-29
Upgrade to LLVM 14, introduce clang-format

ti-mo:
This PR upgrades testdata and examples to build against LLVM14. Call clang-format during `make`.


examples/headers/bpf_helpers.h
```diff
@@ -8,6 +8,7 @@
  * in advance since bpf_helper_defs.h uses such types
  * as __u64.
  */
+#include "common.h"
```
lmb, line 11: This will break the next time we sync changes from libbpf upstream.

ti-mo: True, addressed in https://github.com/cilium/ebpf/pull/614.

PR #613, opened 2022-03-29
link/kprobe: specify symbol offset

nickcooper-zhangtonghao:
Signed-off-by: Tonghao Zhang <zhangtonghao@didiglobal.com>

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -49,6 +49,8 @@ type KprobeOptions struct {
 	//
 	// Needs kernel 5.15+.
 	Cookie uint64
+	// Kprobe symbol offset, or uprobe path offset.
+	Offset uint64
 }
 
 const (
@@ -162,6 +164,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 
 	if opts != nil {
 		args.cookie = opts.Cookie
+		args.offset = opts.Offset
 	}
 
 	// Use kprobe PMU if the kernel has it available.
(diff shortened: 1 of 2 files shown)


link/kprobe.go
```diff
@@ -49,6 +49,8 @@ type KprobeOptions struct {
 	//
 	// Needs kernel 5.15+.
 	Cookie uint64
+	// Kprobe symbol offset, or uprobe path offset.
```
mmat11, line 52: this is not the uprobe offset as this options are only used in the context of kprobes

PS. don't close and open PRs for small changes, just add a commit and optionally rebase/squash

nickcooper-zhangtonghao: ok, thanks! I am unfamiliar with github PR.

mmat11: 👍🏻

mmat11 (approved):

link/kprobe.go
```diff
@@ -49,6 +49,8 @@ type KprobeOptions struct {
 	//
 	// Needs kernel 5.15+.
 	Cookie uint64
+	// Kprobe symbol offset.
```
ti-mo, line 52: This needs more context. How is it used? Why is it necessary? How would the caller determine the correct offset to use?

ti-mo (changes requested):
Missing some comments/documentation. Looks fine otherwise.

link/kprobe.go
```diff
@@ -489,3 +496,14 @@ func kretprobeBit() (uint64, error) {
 	})
 	return kprobeRetprobeBit.value, kprobeRetprobeBit.err
 }
+
+// kprobeToken creates the SYM[+offs] token for the tracefs api.
+func kprobeToken(args probeArgs) string {
```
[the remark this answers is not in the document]
nickcooper-zhangtonghao: Hi, as suggested by @mmat11 , I think we should use the kprobeToken such as uprobeToken.

link/kprobe.go
nickcooper-zhangtonghao: Hi, This offset is used for kprobe, because some functions in kernel is static or inline, we can't insert one probe. but we can use probe the caller with offset. This "offset" is same as tracefs kprobe_events

https://www.kernel.org/doc/html/latest/trace/kprobetrace.html

nickcooper-zhangtonghao: Hi, This offset is used for kprobe, because some functions in kernel is static or inline, we can't insert one probe. but we can use probe the caller with offset. This "offset" is same as tracefs kprobe_events

https://www.kernel.org/doc/html/latest/trace/kprobetrace.html

ti-mo:
```suggestion

	// Offset of the kprobe relative to the traced symbol.
	// Can be used to insert kprobes at arbitrary offsets in kernel functions,
	// e.g. in places where functions have been inlined.
```

ti-mo:
Adding a test similar to `TestUprobeToken` would be nice, as well as one where the offset is set and meaningfully used.

nickcooper-zhangtonghao:
>

PR #614, opened 2022-03-30
examples: include bpf_helpers.h from common.h

ti-mo:
Since bpf_helpers.h includes bpf_helper_defs.h, which depends on certain types like __u64 being defined, work with a single common.h include in examples to make sure this is all set up properly.

This ensures any future vendor updates to bpf_*.h continue to function.


lmb:
LGTM

PR #615, opened 2022-03-30
examples: fentry - add comments to illustrate difference with tcprtt

ti-mo:
The tcprtt example was added recently, which relies on CO-RE information to work across different kernel versions. fentry, on the other hand, will break if structs change in the running kernel, so document this fact.

Replaced unused fields with padding and addressed some nits.


examples/fentry/fentry.c
```diff
@@ -8,8 +8,25 @@
...
+/**
+ * This example copies parts of struct sock_common and struct sock from
+ * the Linux kernel, but doesn't cause any CO-RE information to be emitted
+ * into the ELF object. This requires the struct layout (up until the fields
+ * that are being accessed) to match the kernel's, and the example will break
+ * or misbehave when this is no longer the case.
+ *
+ * Also note that BTF-enabled programs like fentry, fexit, fmod_ret, tp_btf,
+ * lsm, etc. declared using the BPF_PROG macro can read kernel memory without
+ * needing to call bpf_probe_read*().
```
mmat11, line 22: shouldn't this be `BPF_CORE_READ*()`?

ti-mo: BPF_CORE_READ and friends emit a CO-RE relocation and call `bpf_probe_read_*()` behind the scenes, but that's not what is meant here specifically. This comment is just about the memory accesses in the kernel, not the CO-RE part.

PR #617, opened 2022-03-31
Remove extra allocations from perf.Reader.Read()

brycekahle:
Before:
```
goos: linux
goarch: arm64
BenchmarkReader
BenchmarkReader-4   	  287356	      3732 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  305671	      3701 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  305148	      3669 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  306469	      3675 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  295496	      3691 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  297704	      3710 ns/op	     410 B/op	       6 allocs/op
BenchmarkReader-4   	  302049	      3667 ns/op	     410 B/op	       6 allocs/op
BenchmarkReader-4   	  299085	      3680 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  305419	      3673 ns/op	     409 B/op	       6 allocs/op
BenchmarkReader-4   	  305707	      3691 ns/op	     409 B/op	       6 allocs/op
PASS
ok  	github.com/cilium/ebpf/perf	11.532s
```

After:
```
goos: linux
goarch: arm64
BenchmarkReader
BenchmarkReader-4   	  318324	      3476 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  330291	      3492 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  325369	      3509 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  325884	      3508 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  327452	      3497 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  326408	      3599 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  320997	      3507 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  320274	      3512 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  318555	      3506 ns/op	     384 B/op	       2 allocs/op
BenchmarkReader-4   	  329978	      3499 ns/op	     384 B/op	       2 allocs/op
PASS
ok  	github.com/cilium/ebpf/perf	11.788s
```

The remaining 2 allocs are:
1. The creation of the raw sample slice
2. The output slice from `program.Test`

### DIFF
--- a/internal/endian_le.go
+++ b/internal/endian_le.go
@@ -1,0 +1,10 @@
+//go:build amd64 || arm64
+// +build amd64 arm64
+
+package internal
+
+import "encoding/binary"
+
+var NativeEndian binary.ByteOrder = binary.LittleEndian
+
+const ClangEndian = "el"
--- a/perf/ring.go
+++ b/perf/ring.go
@@ -10,8 +10,16 @@ import (
 	"sync/atomic"
 	"unsafe"
 
+	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/unix"
 )
+
+type ringReaderSeeker interface {
+	io.Reader
+	readUint32() (uint32, error)
+	readUint64() (uint64, error)
+	seek(int) error
+}
 
 // perfEventRing is a page of metadata followed by
 // a variable number of pages which form a ring buffer.
@@ -136,9 +144,20 @@ func (rr *ringReader) writeTail() {
 }
 
 func (rr *ringReader) Read(p []byte) (int, error) {
-	start := int(rr.tail & rr.mask)
+	start, end := rr.getBounds(len(p))
+	n := copy(p, rr.ring[start:end])
+	rr.tail += uint64(n)
 
-	n := len(p)
+	if rr.tail == rr.head {
+		return n, io.EOF
+	}
+
+	return n, nil
+}
+
+func (rr *ringReader) getBounds(len int) (start, end int) {
+	n := len
+	start = int(rr.tail & rr.mask)
 	// Truncate if the read wraps in the ring buffer
 	if remainder := cap(rr.ring) - start; n > remainder {
 		n = remainder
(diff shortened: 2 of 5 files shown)


internal/endian_le.go
```diff
@@ -0,0 +1,10 @@
+//go:build amd64 || arm64
+// +build amd64 arm64
+
+package internal
+
+import "encoding/binary"
+
+var NativeEndian binary.ByteOrder = binary.LittleEndian
```
lmb, line 8: Can you explain the purpose of hard coding the byte order for amd64?

perf/ring.go
```diff
@@ -148,13 +167,37 @@ func (rr *ringReader) Read(p []byte) (int, error) {
...
+	end = start + n
+	return
+}
 
-	copy(p, rr.ring[start:start+n])
+func (rr *ringReader) readUint32() (uint32, error) {
+	n := 4
+	start, end := rr.getBounds(n)
+	if end-start < n {
+		return 0, io.EOF
```
lmb, line 178: This doesn't handle the case where a number is split across the beginning / end of the ring buffer?

lmb (commented):
Hi Bryce,

From the commit I'm inferring that the problem is allocations from the perf event headers, is that correct?

I think in the current state the code doesn't handle headers split across the ring correctly. Fixing that without a temporary buffer is going to be cumbersome. What if instead of a custom reader we did the following instead:
* Implement `UnmarshalBinary([]byte) error` on the header types.
* Add a sync.Pool with slices of length `binary.Size(header)`.
* To read a header, retrieve a slice from the pool, call read and then UnmarshalBinary.

This needs an extra copy, but avoids doing the allocation all over again.

Let me know what you think!

internal/endian_le.go
brycekahle: My thought was that it might help the compiler do more optimizations since it changes the determination from runtime to build time.

perf/ring.go
brycekahle: Good call! Let me work on that.

internal/endian_le.go
lmb: Hmm I wonder. Can you do a little test to check whether that is the case? If it turns out to be helpful it'd be nice to have it in a separate commit with an explanation for what is going on.

perf/ring.go
lmb: Just to make sure, you saw https://github.com/cilium/ebpf/pull/617#pullrequestreview-928250434 ?

internal/endian_le.go
brycekahle: Analysis in https://github.com/cilium/ebpf/pull/649

brycekahle:
Superseded by #650

PR #618, opened 2022-04-02
link/kprobe: specify symbol offset

nickcooper-zhangtonghao:
This patch adds kprobe symbol offset and its unit test.
With symbol offset, we can insert kprobe for functions
inlined.

[PR closed https://github.com/cilium/ebpf/pull/613]

Signed-off-by: Tonghao Zhang <zhangtonghao@didiglobal.com>

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -50,6 +50,10 @@ type KprobeOptions struct {
 	//
 	// Needs kernel 5.15+.
 	Cookie uint64
+	// Offset of the kprobe relative to the traced symbol.
+	// Can be used to insert kprobes at arbitrary offsets in kernel functions,
+	// e.g. in places where functions have been inlined.
+	Offset uint64
 }
 
 const (
@@ -163,6 +167,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 
 	if opts != nil {
 		args.cookie = opts.Cookie
+		args.offset = opts.Offset
 	}
 
 	// Use kprobe PMU if the kernel has it available.
@@ -235,8 +240,12 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 		}
 
 		attr = unix.PerfEventAttr{
+			// The minimum size required for PMU kprobes is PERF_ATTR_SIZE_VER1,
+			// since it added the config2 (Ext2) field. Use Ext2 as probe_offset.
+			Size:   unix.PERF_ATTR_SIZE_VER1,
 			Type:   uint32(et),          // PMU event type read from sysfs
 			Ext1:   uint64(uintptr(sp)), // Kernel symbol to trace
+			Ext2:   args.offset,         // Kernel symbol offset
 			Config: config,              // Retprobe flag
 		}
 	case uprobeType:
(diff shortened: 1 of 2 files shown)


link/kprobe.go
```diff
@@ -235,8 +240,12 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 		}
 
 		attr = unix.PerfEventAttr{
+			// The minimum size required for PMU kprobes is PERF_ATTR_SIZE_VER1,
+			// since it added the config2 (Ext2) field. Use Ext2 as probe_offset.
+			Size:   unix.PERF_ATTR_SIZE_VER1,
```
lmb, line 245: According to https://classes.engineering.wustl.edu/cse522/man-pages/perf_event_open.2.pdf VER1 is available from 2.6 so this doesn't change the minimum required version.

lmb (commented):
I'm assuming that the kprobe only fires if the PC is exactly that of sym + offset. Is there a way to add a test where `Offset != 0` to make sure this actually works all the way?

link/kprobe.go
nickcooper-zhangtonghao: if we don't set .size , the kernel will use the PERF_ATTR_SIZE_VER0 64B as default. The config2 will be not included in perf_event_attr.

the kernel code:
static int perf_copy_attr(struct perf_event_attr __user *uattr,  struct perf_event_attr *attr)
        .....
        /* ABI compatibility quirk: */
        if (!size)
                size = PERF_ATTR_SIZE_VER0; // config2 is not in VER0

nickcooper-zhangtonghao: Offset == 0 is ok, it is an option for linux kernel.
https://www.kernel.org/doc/html/latest/trace/kprobetrace.html

lmb: I should've been clearer, using ver1 LGTM.

lmb:
I'm happy with this change, but would like a test. Maybe we can have a negative test where we get an error if we pass math.MaxUint32 or similar if using a valid offset is too complicated.

nickcooper-zhangtonghao:
@lmb Hi
This patch add the kprobeToken unit test only. Offset is different for different kernel function. Offset is set by user who should know what they are doing. For math.MaxUint32, the kprobe(symbol tcp_recvmsg) does't work, because the vmlinux .text section + offset < .text section + math.MaxUint32, then the kernel return err. I think we should allow the user to set Offset they want. If Offset is not suitable, the kernel will return err value.
https://elixir.bootlin.com/linux/v5.13.3/source/kernel/kprobes.c#L1557

mmat11:
@nickcooper-zhangtonghao as @lmb suggested you could test this either with a negative test that will fail when using a bad offset or with a normal test (if you already know the offset for `tcp_recvmsg` and it's more or less stable from kernel to kernel, you could use that)

nickcooper-zhangtonghao:
> @nickcooper-zhangtonghao as @lmb suggested you could test this either with a negative test that will fail when using a bad offset or with a normal test (if you already know the offset for `tcp_recvmsg` and it's more or less stable from kernel to kernel, you could use that)

HI @mmat11
I add kprobe offset test. I use the ksym(vprintk) as symbol, because it present on all tested kernels.
the offset is 0/math.MaxUint64, because use them as offset to insert kprobe, we can get return value we expected

lmb (approved):

PR #622, opened 2022-04-07
Update doc.go

ashish-billore:
Minor typo fix for the wording.


lmb:
I think in this case we're referring to the noun "application exit" so the current wording is correct.

PR #630, opened 2022-04-20
sys: update types

florianl:

Update the available types with `gentypes` on a `v5.17` vmlinux.


lmb:
Did you update this by manually invoking gentypes? The idea is to let go generate do the work and use vmlinux-btf.gz as the source of type data.

The "canonical" way to update types.go is to use `make VMLINUX=/path/to/vmlinux vmlinux-btf` followed by `go generate ./internal/sys/`

florianl:
yes - I manully invoked `gentypes` like: `$ ./gentypes /path/to/vmlinux`.
but I think I get your point, that `internal/btf/testdata/vmlinux-btf.gz` is missing here.

internal/sys/types.go
```diff
@@ -918,8 +924,11 @@ type ProgLoadAttr struct {
 	LineInfoCnt        uint32
 	AttachBtfId        uint32
 	AttachProgFd       uint32
-	_                  [4]byte
+	CoreReloCnt        uint32
```
lmb, line 927: Interesting :D

lmb (approved):

PR #633, opened 2022-04-21
btf: stop relying on TypeID when printing Types

lmb:
btf: stop relying on TypeID when printing Types

Using the TypeID when printing is convenient, but it precludes us from
removing TypeID from the Type interface. Rework the printing code to
use the pointer address to identify a specific Type.

At the same time, make printing a lot more powerful. We now support
printing cyclical types without having to worry about infinite recursion.
For example, formatting a typedef with %v will now ouput

    Typedef:"foo"[Pointer]

We can include details of Pointer by formatting with %1v:

    Typedef:"foo"[Pointer[target=Void]]

Updates #643

### DIFF
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -2,6 +2,7 @@ package btf
 
 import (
 	"fmt"
+	"io"
 	"math"
 	"reflect"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/btf/types.go
```diff
@@ -19,6 +20,16 @@ func (tid TypeID) ID() TypeID {
 
 // Type represents a type described by BTF.
 type Type interface {
+	// Type can be formatted using the %s and %v verbs. %s outputs only the
+	// identity of the type, without any detail. %v outputs additional detail.
+	//
+	// Use the '+' flag to include the address of the type.
+	//
+	// Use the width flag to specify how many levels of detail to output.
```
ti-mo, line 28: This is genius.

We should probably give a quick example (e.g. `%5s`) as a small improvement. The number before the verb is technically not a flag, I think those are `+`, `-`, `#`, `' '` and 0.

lmb: Thanks! I reworded the help text.

ti-mo (approved):

PR #637, opened 2022-04-22
btf: split string table ahead of time

lmb:
We currently read the string table into one big byte slice, and
allocate substrings in Lookup. This forces us to hang on to one
large allocation for the string table, and many small allocations
for substrings. Calling Lookup for the same offset also allocates
on each invocation.

Instead, split the string table into substrings at load time. In
Lookup we can then use a binary search to find the correct string.
As a result, both overall memory and allocations drop:

    ParseVmlinux-4    76.6ms ±12%    73.9ms ± 4%     ~     (p=0.343 n=4+4)

    ParseVmlinux-4    92.9MB ± 0%    84.3MB ± 0%   -9.22%  (p=0.029 n=4+4)

    ParseVmlinux-4      786k ± 0%      693k ± 0%  -11.87%  (p=0.029 n=4+4)

To keep the speed roughly the same I had to "vendor" sort.Search and
specialised it for uint32, since the function was prominent in profiles.

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -35,7 +35,7 @@ type ID uint32
 type Spec struct {
 	// Data from .BTF.
 	rawTypes []rawType
-	strings  stringTable
+	strings  *stringTable
 
 	// Inflated Types.
 	types []Type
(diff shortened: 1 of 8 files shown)


lmb:
cc @paulcacheux, this was the idea I mentioned, please take a look.

lmb:
I also tried replacing `offsets, strings` with a `map[uint32]string` but its slower:

```
ParseVmlinux-4    74.3ms ± 5%    82.4ms ± 3%  +10.87%  (p=0.029 n=4+4)

ParseVmlinux-4    84.3MB ± 0%    87.8MB ± 0%   +4.16%  (p=0.029 n=4+4)

ParseVmlinux-4      693k ± 0%      693k ± 0%     ~     (p=0.943 n=4+4)
```

paulcacheux:
This looks great, thanks for working on this

Issue #542, opened 2022-01-10
Reduce BTF parser's memory footprint


ti-mo:
In its current form, the BTF parser scans and allocates all types in a BTF blob. This can be cost-prohibitive on machines with smaller memory budgets when parsing vmlinux.

One potential example has been reported here: https://github.com/cilium/pwru/issues/45.

This issue is for discussing how we could make BTF parsing less memory-intensive.

---

One possible approach could be JIT-parsing as much as possible by providing an iterator over all types. The string table should be parsed in full, as we need to perform random lookups for inflating types' names. Strings aren't fixed-width, so we'd need to scan the whole string table from the start for performing random lookups.

Aside from a type iterator, we can also support type lookups by ID, which is required for inflating the type graph starting from any existing type with dependent types. (e.g., a struct)

lmb:
> The string table should be parsed in full, as we need to perform random lookups for inflating types' names.

The look up is usually offset -> name, that seems fine without parsing the table? Do you mean an `index` of `name -> id`?

lmb:
I've just merged a couple of improvements which significantly reduce the used memory:

- 2c93040e14ead1005b18d467e4553050133252c3
- 31c4e1cf522f532fafd3d3ae6cec8ceeed8bc059
- 0cb94e8393544cb4eef1d209dbff1965d9409704

lmb:
Memory usage is now at less than half for reading vmlinux, is this still an issue?

PR #638, opened 2022-04-23
btf: fix up inflated types immediately if possible

lmb:
Converting []rawType into []Type currently creates a lot of temporary
data in the form of fixups. They are necessary since a type may reference
itself, for example in a linked list:

    struct node {
        struct node* next;
        ...
    }

No matter which order we encode struct node and struct node* in, we
can't resolve one of the types until we've inflated both.

However, such self referential types are rare. Most of the time we
are referencing an already inflated type. We can use this to our
advantage and perform fixups opportunistically. At the same time
we can split out type assertions from fixups, since most fixups
don't care about types at all.

    ParseVmlinux-4    68.9ms ± 2%    58.4ms ± 1%  -15.22%  (p=0.029 n=4+4)

    ParseVmlinux-4    48.4MB ± 0%    45.2MB ± 0%   -6.69%  (p=0.029 n=4+4)

    ParseVmlinux-4      721k ± 0%      721k ± 0%     ~     (p=0.057 n=4+4)

Updates #542

### DIFF
--- a/internal/btf/types.go
+++ b/internal/btf/types.go
@@ -3,6 +3,7 @@ package btf
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


lmb:
@paulcacheux looking at your commit I realised that there is even more improvements to be had. Please take a look.

paulcacheux (approved):

Issue #616, opened 2022-03-31
Marshal BTF from `btf.Type`


ti-mo:
Over the past few months, while working on per-instruction metadata, we've identified the need for marshaling valid BTF blobs based on `btf.Type`s.

A few use cases:

- caller-provided `btf.Type`s for map keys/values, for pretty-printing dumps and creating valid StructOps maps
- caller-provided FuncInfo (e.g. `Instruction.WithFunc(f *btf.Func)`) for manually assembling (and modifying) programs with bpf-to-bpf calls
- caller-provided source information (e.g. `Instruction.WithSource(s fmt.Stringer)`), for adding comments or other info on instructions
- .. probably a few things I haven't thought of

Optimizations/cleanups:
- BTF blobs accompanying maps can consist of just 2 (best case) types, removing the need for the BTF handle cache.
- Loading the complete BTF blob from larger ELFs is no longer required. For programs, this only needs to contain types+strings of FuncInfos and strings from LineInfos.

This should also make fixing https://github.com/cilium/ebpf/issues/344 possible.

ti-mo:
This was shipped in #796

PR #641, opened 2022-04-25
btf: support marshaling Type

lmb:
This is a first cut at adding a `btf.Builder` which allows turning `Type` into BTF that the kernel understands. The code can round-trip vmlinux BTF and then load the result into the kernel, which is a good indication that we covered most of the bases.

As it stands, the builder is not exported or used in the package, since there is currently no way to add strings necessary for `LineInfo` to the BTF. This should become a lot easier once #606 lands.

Best reviewed per commit.

btf: add btfInt
    
    Replace intEncoding with a type. This allows adding setters.

btf: add Array.Index
    
    The kernel refuses to load an array type without an index type, so probably
    btf.rst is outdated.

btf: move typeDeque to separate file
    
    Split typeDeque into a separate file to reduce the amount of code in types.go.

btf: add stringTableBuilder
    
    A stringTableBuilder helps build a BTF-style string table, which delimits
    individual strings with a 0 byte. Each unique string is only added to the
    table once.

btf: replace Type.walk with walkType
    
    Type.walk takes a *typeDeque as argument. Go escape analysis doesn't see through
    interface boundaries and therefore always allocates such pointers on the heap.
    This forces typeDeque to always be heap allocated, which makes traversal more
    expensive than it has to be.
    
    Replace the interface function with walkType, which essentially contains a big
    type switch.

btf: add builder
    
    Allow building BTF wire format from scratch. We can compare a full vmlinux build to
    just re-encoding raw type data as we currently do in Spec.marshal:
    
        name                    time/op
        BuildVmlinux/builder-4  77.9ms ± 2%
        BuildVmlinux/native-4   23.3ms ± 2%
    
        name                    alloc/op
        BuildVmlinux/builder-4  33.7MB ± 0%
        BuildVmlinux/native-4   20.1MB ± 0%
    
        name                    allocs/op
        BuildVmlinux/builder-4    473k ± 0%
        BuildVmlinux/native-4     394k ± 0%
    
    A full rebuild is about 3x slower and uses 1.6x the memory. We can't use the builder
    to replace Spec.marshal yet, since we don't have a good way to include LineInfos
    in the generated BTF. This will become easier once LineInfos are stored in
    Instruction metadata.
    
    Updates #616


lmb:
As always, needs some more work!

lmb:
This works now!

PR #642, opened 2022-04-25
link: do not use regexp pkg

kolyshkin:
For ID validity checks, replace regular expressions with custom functions checking input strings
byte by byte.

For `sanitizeSymbol`, replace a regexp with a for loop. Add some more test cases while at it.
    
Regular expressions bring in a big library as a dependency, and they
incur some overhead during startup (to compile regexes).

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -429,7 +429,7 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		// the eBPF program itself.
 		// See Documentation/kprobes.txt for more details.
 		token = kprobeToken(args)
-		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, sanitizedSymbol(args.symbol), token)
+		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, sanitizeSymbol(args.symbol), token)
 	case uprobeType:
 		// The uprobe_events syntax is as follows:
 		// p[:[GRP/]EVENT] PATH:OFFSET [FETCHARGS] : Set a probe
@@ -481,7 +481,7 @@ func closeTraceFSProbeEvent(typ probeType, group, symbol string) error {
 
 	// See [k,u]probe_events syntax above. The probe type does not need to be specified
 	// for removals.
-	pe := fmt.Sprintf("-:%s/%s", group, sanitizedSymbol(symbol))
+	pe := fmt.Sprintf("-:%s/%s", group, sanitizeSymbol(symbol))
 	if _, err = f.WriteString(pe); err != nil {
 		return fmt.Errorf("writing '%s' to '%s': %w", pe, typ.EventsPath(), err)
 	}
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -269,7 +269,7 @@ func unsafeStringPtr(str string) (unsafe.Pointer, error) {
 // name automatically has its invalid symbols converted to underscores so the caller
 // can pass a raw symbol name, e.g. a kernel symbol containing dots.
 func getTraceEventID(group, name string) (uint64, error) {
-	name = sanitizedSymbol(name)
+	name = sanitizeSymbol(name)
 	tid, err := uint64FromFile(tracefsPath, "events", group, name, "id")
 	if errors.Is(err, os.ErrNotExist) {
 		return 0, fmt.Errorf("trace event %s/%s: %w", group, name, os.ErrNotExist)
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
+	"strings"
 	"sync"
 
 	"github.com/cilium/ebpf"
@@ -15,10 +15,6 @@ import (
 
 var (
 	uprobeEventsPath = filepath.Join(tracefsPath, "uprobe_events")
-
-	// rgxEventSymbol is used to strip invalid characters from the [k,u]probe symbol
-	// as they are not allowed to be used as the EVENT token in tracefs.
-	rgxEventSymbol = regexp.MustCompile("[^a-zA-Z0-9]+")
 
 	uprobeRetprobeBit = struct {
 		once  sync.Once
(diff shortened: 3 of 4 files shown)


link/kprobe.go
```diff
@@ -144,6 +138,34 @@ func Kretprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions) (Link, er
...
 
+func isValidSymbol(s string) bool {
+	// The code below produces the same results as
+	// regexp.MustCompile("^[a-zA-Z_][0-9a-zA-Z_.]*$").MatchString(s).
+
+	if len(s) < 1 {
+		return false
+	}
+
+	for i := 0; i < len(s); i++ {
```
lmb, line 149: Nit: `for i, c := range []byte{s} {`

Casting to []byte in a range loop doesn't allocate.

link/kprobe.go
```diff
@@ -144,6 +138,34 @@ func Kretprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions) (Link, er
 	return lnk, nil
 }
 
+func isValidSymbol(s string) bool {
```
lmb, line 141: `isValidKprobeSymbol`?

link/uprobe.go
```diff
@@ -316,8 +312,29 @@ func tracefsUprobe(args probeArgs) (*perfEvent, error) {
 }
 
 // sanitizedSymbol replaces every invalid character for the tracefs api with an underscore.
-func sanitizedSymbol(symbol string) string {
-	return rgxEventSymbol.ReplaceAllString(symbol, "_")
+// It is equivalent to calling regexp.MustCompile("[^a-zA-Z0-9]+").ReplaceAllString("_").
+func sanitizedSymbol(s string) string {
+	var b strings.Builder
```
lmb, line 317: Can you use https://pkg.go.dev/strings#Map instead?

link/uprobe.go
```diff
@@ -316,8 +312,29 @@ func tracefsUprobe(args probeArgs) (*perfEvent, error) {
 }
 
 // sanitizedSymbol replaces every invalid character for the tracefs api with an underscore.
-func sanitizedSymbol(symbol string) string {
-	return rgxEventSymbol.ReplaceAllString(symbol, "_")
+// It is equivalent to calling regexp.MustCompile("[^a-zA-Z0-9]+").ReplaceAllString("_").
+func sanitizedSymbol(s string) string {
```
lmb, line 316: Can you rename to `sanitizeSymbol`?

lmb (commented):
Thanks, looks good to me. Out of curiosity, do you have numbers for size / start up time impact?

link/kprobe.go
```diff
@@ -144,6 +138,34 @@ func Kretprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions) (Link, er
 	return lnk, nil
 }
```
ti-mo, line 140:
```suggestion
// isValidKprobeSymbol implements the equivalent of a regex match
// against "^[a-zA-Z_][0-9a-zA-Z_.]*$".
```

link/perf_event.go
```diff
@@ -373,3 +367,27 @@ var haveBPFLinkPerfEvent = internal.FeatureTest("bpf_link_perf_event", "5.15", f
 	}
 	return err
 })
+
```
ti-mo, line 370:
```suggestion

// isValidTraceID implements the equivalent of a regex match
// against "^[a-zA-Z_][0-9a-zA-Z_]*$".
//
```

ti-mo (changes requested):
Is this really worth it? I'd be curious to see some numbers.

kolyshkin:
> Is this really worth it? I'd be curious to see some numbers.

The initialization overhead I mentioned in the description is probably negligible, so not doing any measurements (yet it calls `regexp.MustCompile` from the `init` section for every global variable like the removed `rgxEventSymbol`).

Let's see the binary size overhead. Using Go 1.18.1 on Linux/amd64, `-orig` is with cilium/ebpf v0.8.1, the no-suffix version is compiled with the version from this PR:

```console
[kir@kir-rhat ebpf]$ ls -l ebpf-test-orig ebpf-test
-rwxrwxr-x. 1 kir kir 3187473 Apr 26 20:22 ebpf-test
-rwxrwxr-x. 1 kir kir 3387158 Apr 26 20:20 ebpf-test-orig
```

The difference is close to 200K. Let's see the `size` results:
```console
$ size ebpf-test ebpf-test-orig 
   text	   data	    bss	    dec	    hex	filename
1993489	 112896	 217376	2323761	 237531	ebpf-test
2121479	 121456	 217760	2460695	 258c17	ebpf-test-orig
```

So, about 130KB from text+data+bss, I guess most of it is regexp and regexp/syntax packages.

Note the measurements might not be entirely valid as I'm comparing the latest released version of github.com/cilium/ebpf (v0.8.1) against the code from this PR.

I have used this non-program calling uprobe (which is being patched to not use regexp):

```go
package main

import (
	"log"

	"github.com/cilium/ebpf"
	"github.com/cilium/ebpf/asm"
	"github.com/cilium/ebpf/link"
)

func main() {
	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
		Type:    ebpf.Kprobe,
		License: "MIT",
		Instructions: asm.Instructions{
			asm.Mov.Imm(asm.R0, 0),
			asm.Return(),
		},
	})
	if err != nil {
		log.Fatal(err)
	}

	bashEx, err := link.OpenExecutable("/bin/bash")
	if err != nil {
		log.Fatal(err)
	}

	up, err := bashEx.Uprobe("main", prog, nil)
	if err != nil {
		log.Fatal(err)
	}
	up.Close()
}
```

with this go.mod file:
```
module github.com/kolyshkin/ebpf-test

go 1.18

require (
	github.com/cilium/ebpf v0.8.1
	golang.org/x/sys v0.0.0-20210906170528-6f6e22806c34 // indirect
)

replace github.com/cilium/ebpf => github.com/kolyshkin/ebpf v0.8.2-0.20220425190318-df1de70d40a1
```

(Remove the last line to go back to stock v0.8.1)

link/kprobe.go
kolyshkin: You meant ` range []byte(s)` I guess. Yes, this looks much better!

link/kprobe.go
kolyshkin: done

link/uprobe.go
kolyshkin: This will slightly change the output -- currently, each sequence of symbols that are not allowed are replaced with a single `_`.

If this is an OK change to made, let me know and I will do it.

link/uprobe.go
kolyshkin: Done

kolyshkin:
Resolved all review comments except https://github.com/cilium/ebpf/pull/642#discussion_r858448166, rebased.

link/uprobe.go
lmb: I think that is acceptable.

lmb:
Thanks for the numbers, seems worth it.

PR #644, opened 2022-04-26
internal: make ErrorWithLog slightly more efficient

lmb:
Trim the log before converting it into a string will save an allocation.
Also document that BPF_LOAD_BTF never returns ENOSPC.


lmb:
If type checking fails the code doesn't check log buffer overflow, so ENOSPC is never generated.

Issue #517, opened 2021-12-06
`LoadAndAssign` fails after `RewriteMaps`


mauriciovasquezbernal:
**Describe the bug**

`LoadAndAssign` fails when `RewriteMaps` and bpf2go are used with a `missing map` error. 

**To Reproduce**
There is a reproducer available at https://github.com/mauriciovasquezbernal/ebpfissue/tree/1bfab92fea1c988c032d6b5715fdfee9ed80c1ce

```sh
$ go version
go version go1.16.3 linux/amd64
$ uname -r
5.11.0-41-generic
$ clang --version
Ubuntu clang version 12.0.0-3ubuntu1~20.04.4
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

**Expected behavior**
The object should load fine and IMO the rewritten map should be assigned to `fooMaps.MountNsSet`

Note: I added a call to `runtime.KeepAlive` to avoid hitting https://github.com/cilium/ebpf/issues/515.

lmb:
I can see why you'd like behaviour like this, but it's hard to make RewriteMaps work the way you want it to. We'd need a new API as pointed out by timo https://github.com/cilium/ebpf/issues/515#issuecomment-986761701

FWIW, you can just set `pinning` on `mount_ns_net` and have the library handle pinning for you ([example here](https://github.com/cilium/ebpf/issues/515#issuecomment-986761701)). Why is that not an option here?

mauriciovasquezbernal:
> FWIW, you can just set `pinning` on `mount_ns_net` and have the library handle pinning for you ([example here](https://github.com/cilium/ebpf/issues/515#issuecomment-986761701)). Why is that not an option here?

I ended up doing exactly that. My only reason to prefer a RewriteMaps like approach is that I don't have to pin the map if it's not already pinned. (If I'm creating the map or if I already have a reference to it). 

I'm fine with always using the pinning, but I think `RewriteMaps` should also work in this case.

lmb:
Hmm. Why does pinning / no pinning change dynamically for you? I don't fully understand the use case, sorry.

> I'm fine with always using the pinning

You shouldn't have to if you don't care about persisting the state. You're saying you want to persist without pinning?

mauriciovasquezbernal:
> Hmm. Why does pinning / no pinning change dynamically for you? I don't fully understand the use case, sorry.

[...]

PR #646, opened 2022-04-27
collection: Add ReplaceMaps to CollectionOptions

mauriciovasquezbernal:
```
    collection: Add MapReplacements to CollectionOptions
    
    Add a new way to replace references to specific maps when creating a
    new collection by introducing a MapReplacements field to the
    CollectionOptions struct.
    
    The goal of this functionality is the same provided by RewriteMaps(),
    but this new approach plays nicely with bpf2go as https://github.com/cilium/ebpf/issues/517
    is fixed.
```

Fixes https://github.com/cilium/ebpf/issues/517

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -18,6 +18,15 @@ import (
 type CollectionOptions struct {
 	Maps     MapOptions
 	Programs ProgramOptions
+
+	// MapReplacements defines a set of maps that will be used
+	// instead of creating new ones when loading the object. The
+	// maps' specs should be compatible and there must exist a map
+	// in CollectionSpec.Maps for each map in MapReplacements.
+	// MapReplacements are cloned before being used in the
+	// Collection, so the user can Close() them if not needed
+	// anymore.
+	MapReplacements map[string]*Map
 }
 
 // CollectionSpec describes a collection.
@@ -214,7 +223,10 @@ func (cs *CollectionSpec) Assign(to interface{}) error {
 // Returns an error if any of the fields can't be found, or
 // if the same Map or Program is assigned multiple times.
 func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions) error {
-	loader := newCollectionLoader(cs, opts)
+	loader, err := newCollectionLoader(cs, opts)
+	if err != nil {
+		return err
+	}
 	defer loader.cleanup()
 
 	// Support assigning Programs and Maps, lazy-loading the required objects.
@@ -289,7 +301,10 @@ func NewCollection(spec *CollectionSpec) (*Collection, error) {
 // Omitting Collection.Close() during application shutdown is an error.
 // See the package documentation for details around Map and Program lifecycle.
 func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (*Collection, error) {
-	loader := newCollectionLoader(spec, &opts)
+	loader, err := newCollectionLoader(spec, &opts)
+	if err != nil {
+		return nil, err
+	}
 	defer loader.cleanup()
 
 	// Create maps first, as their fds need to be linked into programs.
@@ -379,9 +394,16 @@ type collectionLoader struct {
 	handles  *handleCache
 }
 
-func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) *collectionLoader {
+func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) (*collectionLoader, error) {
 	if opts == nil {
 		opts = &CollectionOptions{}
+	}
+
+	// Check for existing MapSpecs in the CollectionSpec for all provided replacement maps.
+	for name := range opts.MapReplacements {
+		if _, ok := coll.Maps[name]; !ok {
+			return nil, fmt.Errorf("replacement map %s not found in CollectionSpec", name)
+		}
 	}
 
 	return &collectionLoader{
@@ -390,7 +412,7 @@ func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) *collect
 		make(map[string]*Map),
 		make(map[string]*Program),
 		newHandleCache(),
-	}
+	}, nil
 }
 
 // finalize should be called when all the collectionLoader's resources
@@ -424,6 +446,20 @@ func (cl *collectionLoader) loadMap(mapName string) (*Map, error) {
 
 	if mapSpec.BTF != nil && cl.coll.Types != mapSpec.BTF.Spec {
 		return nil, fmt.Errorf("map %s: BTF doesn't match collection", mapName)
+	}
+
+	if replaceMap, ok := cl.opts.MapReplacements[mapName]; ok {
+		if err := mapSpec.checkCompatibility(replaceMap); err != nil {
+			return nil, fmt.Errorf("use replacement map %s: %w", mapSpec.Name, err)
+		}
+		// Clone the map to avoid closing user's map later on.
+		m, err := replaceMap.Clone()
+		if err != nil {
+			return nil, err
+		}
+
+		cl.maps[mapName] = m
+		return m, nil
 	}
 
 	m, err := newMapWithOptions(mapSpec, cl.opts.Maps, cl.handles)
--- a/collection_test.go
+++ b/collection_test.go
@@ -1,6 +1,7 @@
 package ebpf
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"testing"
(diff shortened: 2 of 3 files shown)


collection.go
```diff
@@ -426,9 +445,19 @@ func (cl *collectionLoader) loadMap(mapName string) (*Map, error) {
...
 
-	m, err := newMapWithOptions(mapSpec, cl.opts.Maps, cl.handles)
-	if err != nil {
-		return nil, fmt.Errorf("map %s: %w", mapName, err)
+	var m *Map
+	var err error
+	var ok bool
+
+	if m, ok = cl.opts.OverrideMaps[mapName]; ok {
+		if err := mapSpec.checkCompatibility(m); err != nil {
```
lmb, line 453: You can move this check into newCollectionLoader and then initialise cl.maps with opts.OverrideMaps.

collection.go
```diff
@@ -18,6 +18,12 @@ import (
 type CollectionOptions struct {
 	Maps     MapOptions
 	Programs ProgramOptions
+
+	// OverrideMaps defines a set of maps that will be used instead
+	// of creating new ones when loading the object. The maps' specs
+	// should be compatible and there must exist a map in
+	// CollectionSpec.Maps for each map in OverrideMaps.
+	OverrideMaps map[string]*Map
```
lmb, line 26: Bikeshed: `MapOverrides`, `MapReplacements`?

lmb (commented):
Seems fine, wdyt @ti-mo?

collection.go
mauriciovasquezbernal: What should happen to the maps in opts.OverrideMaps when a call to `NewCollectionWithOptions()` / `LoadAndAssign()` fails? Should they be closed? I'm asking because if we initialize `cl.maps` with `opts.OverrideMaps` those will be closed in case of error in `(cl *collectionLoader) cleanup()`.

collection.go
mauriciovasquezbernal: No idea. I had named it before `ReplaceMaps` but then realized it was not a very good one. I'll let up to you folks.

ti-mo: `MapReplacements` gmv :+1:

map.go
```diff
@@ -24,7 +24,7 @@ var (
 	ErrKeyNotExist      = errors.New("key does not exist")
 	ErrKeyExist         = errors.New("key already exists")
 	ErrIterationAborted = errors.New("iteration aborted")
-	ErrMapIncompatible  = errors.New("map's spec is incompatible with pinned map")
+	ErrMapIncompatible  = errors.New("map's spec is incompatible")
```
ti-mo, line 27: I see the initial "map's spec" is/was a bit ambiguous, since a MapSpec is always caller-provided, and doesn't necessarily relate to an existing Map. We could improve this as follows:

```suggestion
	ErrMapIncompatible  = errors.New("map spec is incompatible with existing map")
```

This is clear both in the pinned Map case and in the manually-created Map case (which wasn't necessarily pinned). Good catch. :+1:

collection.go
```diff
@@ -379,18 +391,25 @@ type collectionLoader struct {
 	handles  *handleCache
 }
 
-func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) *collectionLoader {
+func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) (*collectionLoader, error) {
 	if opts == nil {
 		opts = &CollectionOptions{}
 	}
 
+	// check that there exist maps in coll.Maps for all maps in opts.OverrideMaps
```
ti-mo, line 399: Ultra-nit:
```suggestion
	// Check for existing MapSpecs in the CollectionSpec for all provided replacement Maps.
```

collection.go
```diff
@@ -379,18 +391,25 @@ type collectionLoader struct {
...
+	for name := range opts.OverrideMaps {
+		if _, ok := coll.Maps[name]; !ok {
+			return nil, fmt.Errorf("override map %s doesn't exist in the spec", name)
```
ti-mo, line 402: Going with the 'replacement' term:
```suggestion
			return nil, fmt.Errorf("replacement map %s not found CollectionSpec", name)
```

collection.go
ti-mo: > What should happen to the maps in `opts.OverrideMaps` when a call to `NewCollectionWithOptions()` / `LoadAndAssign()` fails? Should they be closed?

Probably not. We don't own those Maps, and it will cause bugs in user code (and cause data loss if not pinned).

> those will be closed in case of error in `(cl *collectionLoader) cleanup()`.

Still the case in the current proposal, right? You're pulling a Map from OverrideMaps and putting it into `cl.maps` below.

I agree with having the compatibility check here, but just make `loadMap()` early-return the Map here if it exists in the overrides, without assigning it to `cl.maps`. The overrides then effectively serve as a second Map cache.

ti-mo (changes requested):
Thanks for this, good to go with a few changes. :+1: Let's use the term 'replacement' instead of 'override' to stick to the existing terminology.

collection.go
lmb: Good catch re: what happens on error. This leads to the follow up question: what should happen if I use `NewCollectionWithOptions` directly (so not via `LoadAndAssign`)? In that case `coll.Maps` will contain references to `Replacements`. Who owns those? Maybe we need to `Clone()` the replacement and stick it in cl.maps after all?

lmb:
TBD, could happen in another PR: should we deprecate `RewriteMaps` in favour of this?

collection.go
mauriciovasquezbernal: > Still the case in the current proposal, right? You're pulling a Map from OverrideMaps and putting it into cl.maps below.

Yes, almost the same. The small difference is that I'm not pulling all maps at once. 

> Who owns those?

It's a good question, not only in the error case but in general: 

```
m := ebpf.NewMap(...)
col := ebpf.NewCollectionWithOptions(..., ebpf.CollectionOptions{MapReplacements: ... "mymap": m})

....

col.Close()

// should m be valid here? 
```

> Maybe we need to Clone() the replacement and stick it in cl.maps after all?

It seems it'd simplify things.

mauriciovasquezbernal:
> TBD, could happen in another PR: should we deprecate RewriteMaps in favour of this?

I think yes. I'm also wondering if `RewriteConstants` should also follow this approach (to have a `Constants` field in `CollectionOptions`)?

collection.go
```diff
@@ -379,18 +394,25 @@ type collectionLoader struct {
...
-func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) *collectionLoader {
+func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) (*collectionLoader, error) {
 	if opts == nil {
 		opts = &CollectionOptions{}
 	}
 
+	// Check for existing MapSpecs in the CollectionSpec for all provided replacement maps.
+	for name := range opts.MapReplacements {
+		if _, ok := coll.Maps[name]; !ok {
+			return nil, fmt.Errorf("replacement map %s not found CollectionSpec", name)
```
ti-mo, line 405:
```suggestion
			return nil, fmt.Errorf("replacement map %s not found in CollectionSpec", name)
```

lmb:
> I'm also wondering if RewriteConstants should also follow this approach (to have a Constants field in CollectionOptions)?

Caveat: this needs some thinking to make it work nice with `bpf2go`. I'd like to generate a `type constants struct` and be able to pass that to `NewCollection` / `LoadAndAssign` somehow. For that to work `CollectionOptions.Constants` would have to be `interface{}`. If we want to replace `RewriteConstants` at the same time we'd need it to be `map[string]interface{}` however.

collection_test.go
```diff
@@ -184,6 +187,112 @@ func TestCollectionSpecRewriteMaps(t *testing.T) {
...
+		t.Fatal("new / override map not used")
+	}
+
+	// Check that newMap isn't closed when the collection is closed
+	coll1.Close()
+	if _, err := unix.FcntlInt(uintptr(newMap.FD()), syscall.F_GETFD, 0); err != nil {
+		t.Fatal("replaced map was closed")
+	}
+
+	// Override non-existing map
```
ti-mo, line 261: Everything past this point would be better off as a separate test. We can get away with a collspec without a prog for this.

collection_test.go
```diff
@@ -184,6 +187,112 @@ func TestCollectionSpecRewriteMaps(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if ret != 2 {
+		t.Fatal("new / override map not used")
+	}
+
+	// Check that newMap isn't closed when the collection is closed
+	coll1.Close()
+	if _, err := unix.FcntlInt(uintptr(newMap.FD()), syscall.F_GETFD, 0); err != nil {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Right. I thought using newMap.Put() was too simple, but you're right, it's clearer.

collection_test.go
mauriciovasquezbernal: I split it in 3 different tests.

collection_test.go
```diff
@@ -184,6 +187,112 @@ func TestCollectionSpecRewriteMaps(t *testing.T) {
...
+	newMap2, err := NewMap(&MapSpec{
+		Type:       Array,
+		KeySize:    4,
+		ValueSize:  8, // this is different
+		MaxEntries: 1,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	coll3, err := NewCollectionWithOptions(cs, CollectionOptions{
```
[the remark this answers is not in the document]
mauriciovasquezbernal: It's not anymore an issue after separating the tests.

collection.go
mauriciovasquezbernal: I've trying to reproduce this problem but I can't.  

> since loadMaps isn't called for all maps when using LoadAndAssign.

If loadMaps is not called it means the map is not used, hence not replace and there's nothing to worry about its compatibility. What I am missing?

ti-mo: > If loadMaps is not called it means the map is not used

Correct.

> What I am missing?

Lorenz wants to make sure all given MapReplacements are always compatibility-checked, regardless of whether they are loaded or not. This avoids any surprises.

I'll pick this up in a follow-up together with the ReplaceMaps() deprecation.

Issue #576, opened 2022-02-18
add Program.Run to allow passing all parameters to BPF_PROG_RUN


lmb:
Some recent kernel version introduced `BPF_PROG_RUN`. It's just an alias for `BPF_PROG_TEST_RUN` but it shows that upstream considers RUN to be a basic primitive on programs. Another thing that has become apparent is that the input required by different types of programs differs wildly.

* For some prog types ctx must be null, for others it's optional
* For some prog types data must be null and ctx is required (SK_LOOKUP)

Instead of `TestWithContext` I would add a method `Program.Run(RunOptions) (RunResult, error)`. Names TBD. `RunOptions` contains all the possible fields like `Data`, `Context`, etc. RunResult contains data output, context output, etc. `Test` and `Benchmark` keep the same API but are wrappers around `Run`.

cc @ti-mo

_Originally posted by @lmb in https://github.com/cilium/ebpf/pull/511#pullrequestreview-823817507_

lmb:
See #537 for some discussion on what the default argument for input should be.

lmb:
RunOptions should also include `Reset`, so that program interruption can be handled by the caller.

dxuuu:
`Run(RunOptions)` sounds reasonable to me. Adding a reset callback also sounds good. Just lmk on #647 and I'll send an update

lmb:
This is now in master!

PR #647, opened 2022-04-28
program: Add TestWithOptions() API

danobi:
This PR adds support for a new `TestWithOptions()` API
for `Program`s. The extra options are useful for more advanced
testing.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -488,12 +488,24 @@ func (p *Program) Close() error {
 
 // Various options for Run'ing a Program
 type RunOptions struct {
-	Data    []byte
+	// Program's data input. Required field.
+	Data []byte
+	// Program's data after Program has run. Caller must allocate. Optional field.
 	DataOut []byte
-	Repeat  uint32
-	Flags   uint32
-	Cpu     uint32
-	Reset   func()
+	// Program's context input. Optional field.
+	Context interface{}
+	// Program's context after Program has run. Must be a pointer or slice. Optional field.
+	ContextOut interface{}
+	// Number of times to run Program. Optional field. Defaults to 1.
+	Repeat uint32
+	// Optional flags.
+	Flags uint32
+	// CPU to run Program on. Optional field.
+	// Note not all program types support this field.
+	CPU uint32
+	// Called whenever the syscall is interrupted, and should be set to testing.B.ResetTimer
+	// or similar. Typically used during benchmarking. Optional field.
+	Reset func()
 }
 
 // Test runs the Program in the kernel with the given input and returns the
@@ -522,6 +534,17 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 		return ret, nil, fmt.Errorf("can't test program: %w", err)
 	}
 	return ret, opts.DataOut, nil
+}
+
+// Run runs the Program in kernel with given RunOptions.
+//
+// Note: the same restrictions from Test apply.
+func (p *Program) Run(opts *RunOptions) (uint32, error) {
+	ret, _, err := p.testRun(opts)
+	if err != nil {
+		return ret, fmt.Errorf("can't test program: %w", err)
+	}
+	return ret, nil
 }
 
 // Benchmark runs the Program with the given input for a number of times
@@ -611,6 +634,18 @@ func (p *Program) testRun(opts *RunOptions) (uint32, time.Duration, error) {
 		return 0, 0, err
 	}
 
+	ctx := new(bytes.Buffer)
+	if opts.Context != nil {
+		if err := binary.Write(ctx, internal.NativeEndian, opts.Context); err != nil {
+			return 0, 0, fmt.Errorf("cannot serialize context: %v", err)
+		}
+	}
+
+	var ctxOut []byte
+	if opts.ContextOut != nil {
+		ctxOut = make([]byte, binary.Size(opts.ContextOut))
+	}
+
 	attr := sys.ProgRunAttr{
 		ProgFd:      p.fd.Uint(),
 		DataSizeIn:  uint32(len(opts.Data)),
@@ -618,6 +653,12 @@ func (p *Program) testRun(opts *RunOptions) (uint32, time.Duration, error) {
 		DataIn:      sys.NewSlicePointer(opts.Data),
 		DataOut:     sys.NewSlicePointer(opts.DataOut),
 		Repeat:      uint32(opts.Repeat),
+		CtxSizeIn:   uint32(len(ctx.Bytes())),
+		CtxSizeOut:  uint32(len(ctxOut)),
+		CtxIn:       sys.NewSlicePointer(ctx.Bytes()),
+		CtxOut:      sys.NewSlicePointer(ctxOut),
+		Flags:       opts.Flags,
+		Cpu:         opts.CPU,
 	}
 
 	for {
@@ -647,6 +688,13 @@ func (p *Program) testRun(opts *RunOptions) (uint32, time.Duration, error) {
 			panic("kernel wrote past end of output buffer")
 		}
 		opts.DataOut = opts.DataOut[:int(attr.DataSizeOut)]
+	}
+
+	if len(ctxOut) != 0 {
+		b := bytes.NewReader(ctxOut)
+		if err := binary.Read(b, internal.NativeEndian, opts.ContextOut); err != nil {
+			return 0, 0, fmt.Errorf("failed to decode ContextOut: %v", err)
+		}
 	}
 
 	total := time.Duration(attr.Duration) * time.Nanosecond
--- a/prog.go
+++ b/prog.go
@@ -486,6 +486,16 @@ func (p *Program) Close() error {
 	return p.fd.Close()
 }
 
+// Various options for Run'ing a Program
+type RunOptions struct {
+	Data    []byte
+	DataOut []byte
+	Repeat  uint32
+	Flags   uint32
+	Cpu     uint32
+	Reset   func()
+}
+
 // Test runs the Program in the kernel with the given input and returns the
 // value returned by the eBPF program. outLen may be zero.
 //
@@ -494,11 +504,24 @@ func (p *Program) Close() error {
 //
 // This function requires at least Linux 4.12.
 func (p *Program) Test(in []byte) (uint32, []byte, error) {
-	ret, out, _, err := p.testRun(in, 1, nil)
+	// Older kernels ignore the dataSizeOut argument when copying to user space.
+	// Combined with things like bpf_xdp_adjust_head() we don't really know what the final
+	// size will be. Hence we allocate an output buffer which we hope will always be large
+	// enough, and panic if the kernel wrote past the end of the allocation.
+	// See https://patchwork.ozlabs.org/cover/1006822/
+	out := make([]byte, len(in)+outputPad)
+
+	opts := RunOptions{
+		Data:    in,
+		DataOut: out,
+		Repeat:  1,
+	}
+
+	ret, _, err := p.testRun(&opts)
 	if err != nil {
 		return ret, nil, fmt.Errorf("can't test program: %w", err)
 	}
-	return ret, out, nil
+	return ret, opts.DataOut, nil
 }
 
 // Benchmark runs the Program with the given input for a number of times
@@ -513,7 +536,17 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 //
 // This function requires at least Linux 4.12.
 func (p *Program) Benchmark(in []byte, repeat int, reset func()) (uint32, time.Duration, error) {
-	ret, _, total, err := p.testRun(in, repeat, reset)
+	if uint(repeat) > math.MaxUint32 {
+		return 0, 0, fmt.Errorf("repeat is too high")
+	}
+
+	opts := RunOptions{
+		Data:   in,
+		Repeat: uint32(repeat),
+		Reset:  reset,
+	}
+
+	ret, total, err := p.testRun(&opts)
 	if err != nil {
 		return ret, total, fmt.Errorf("can't benchmark program: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


lmb:
Thank you for your PR, I'll try to do a full review tomorrow. In the meantime, please take a look at #576 which discusses adding this feature.

prog_test.go
```diff
@@ -90,6 +91,56 @@ func TestProgramRun(t *testing.T) {
...
+	}
+	defer prog.Close()
+
+	buf := make([]byte, 14)
+	xdp := sys.XdpMd{
+		Data:    0,
+		DataEnd: 14,
+	}
+	xdpOut := sys.XdpMd{}
+	ctx := (*(*[unsafe.Sizeof(xdp)]byte)(unsafe.Pointer(&xdp)))[:]
```
lmb, line 122: See my comment on `binary.Write` above, usage of `unsafe` should not be necessary / encouraged.

prog.go
```diff
@@ -549,6 +549,16 @@ func (p *Program) Close() error {
...
 }
 
+// Extented Program test options and attributes
+type TestOptions struct {
+	In     []byte
+	Ctx    []byte
+	CtxOut []byte
+	Repeat uint32
+	Flags  uint32
+	Cpu    uint32
```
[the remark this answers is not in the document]
dxuuu: Seems like some prog types (raw tp) setting CPU tells the handler to run the prog on that specific CPU by sending an IPI.

danobi:
Pushed a commit that addresses all comments except for RunOptions/RunResult split

prog.go
```diff
@@ -549,6 +549,16 @@ func (p *Program) Close() error {
 	return p.fd.Close()
 }
 
+// Extented Program test options and attributes
+type TestOptions struct {
+	In     []byte
+	Ctx    []byte
+	CtxOut []byte
```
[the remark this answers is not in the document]
danobi: I'm not an expert on go, but how would you convert from `[]byte` -> `interface{}` for returned context?

danobi:
Not sure what's up with CI. Seems like one of the jobs got stuck (can't get the page to load)

prog.go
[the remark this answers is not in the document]
ti-mo: When working with converting Go types to/from binary, my mind immediately goes to `encoding.Binary(Un)Marshaler`. We could take a `BinaryMarshaler` for `Context`.

The following needs to be taken into account:
```
 *		**BPF_PROG_TYPE_SK_LOOKUP**
 *			*data_in* and *data_out* must be NULL.
 *
 *		**BPF_PROG_TYPE_RAW_TRACEPOINT**,
 *		**BPF_PROG_TYPE_RAW_TRACEPOINT_WRITABLE**
 *
 *			*ctx_out*, *data_in* and *data_out* must be NULL.
 *			*repeat* must be zero.
```

With the current patch, I don't think these prog types will work, since all fields (data_in/out, ctx_in/out) unconditionally receive slice pointers, even if their lengths are zero. This will need to be changed to conditionally set the respective fields when their input slices are > len 0.

On the bright side, this constraint might guide us in designing the API:

```go
type RunContext struct {
	// Program's data input. Required field.
	DataIn []byte

	// Program's context input. Optional field.
	ContextIn encoding.BinaryMarshaler

	// Program's context after Program has run. Optional field.
	ContextOut encoding.BinaryUnmarshaler
	// .. or
	ContextOut []byte
// ...
}
```

- DataOut could be the return value of the function or caller-provided. Both could work given the input struct is called `RunContext`. The forced allocation of the output buffer w/ padding could be moved to `Test()` and we don't need to break API. `Benchmark()` could then run with SkipOutput implicitly enabled.
- If ContextIn is provided, a buffer for ContextOut can be allocated with the same length as the marshaling result. If ContextOut is not provided, this buffer would not be allocated and the CtxOut arg not provided to the syscall to accomodate for `RAW_TRACEPOINT*`.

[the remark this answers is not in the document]
danobi: With your proposed `RunContext`, we still need to give a way for the user to provide the context value, right? `ContextIn` and `ContextOut` are interfaces. We still need the data somewhere. For that reason, I think it might be cleaner to have `Context` and `ContextOut` to be `interface{}`.

> With the current patch, I don't think these prog types will work, since all fields (data_in/out, ctx_in/out) unconditionally receive slice pointers, even if their lengths are zero.

I think even without this patch, BPF_PROG_TYPE_RAW_TRACEPOINT* was broken: https://github.com/cilium/ebpf/blob/e0825791eaf28e70c8c6403054c85b5d37450290/prog.go#L585-L587 . `data_in` was never permitted to be null.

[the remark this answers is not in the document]
danobi: I'd like to also better understand the concern about allocating for `Benchmark()`. The prog iterations are run inside the kernel, so allocation time is not measured.

danobi:
I've cleaned up the commits and changed `ContextOut` to `interface{}`. IMO this approach balances ergonomics with reflection. Please let me know what you think.

prog.go
```diff
@@ -506,7 +523,23 @@ func (p *Program) Close() error {
...
+	if err != nil {
+		return ret, nil, fmt.Errorf("can't test program: %w", err)
+	}
+	return ret, out, nil
+}
+
+// Run runs the Program in kernel with given RunOptions.
+//
+// Note: the same restrictions from Test apply.
+func (p *Program) Run(opts RunOptions) (uint32, []byte, error) {
```
lmb, line 541: See the discussion on `DataOut`, this should not return `[]byte`.

prog.go
```diff
@@ -498,6 +498,23 @@ func (p *Program) Close() error {
...
+	// Program's context after Program has run. Must be a pointer or slice. Optional field.
+	ContextOut interface{}
+	// Number of times to run Program. Optional field. Defaults to 1.
+	Repeat uint32
+	// Optional flags.
+	Flags uint32
+	// CPU to run Program on. Optional field.
+	// Note not all program types support this field.
+	CPU uint32
+}
```
lmb, line 516: This should include `Reset func()` with appropriate documentation.

prog_test.go
```diff
@@ -90,6 +90,54 @@ func TestProgramRun(t *testing.T) {
...
+	buf := make([]byte, 14)
+	xdp := sys.XdpMd{
+		Data:    0,
+		DataEnd: 14,
+	}
+	xdpOut := sys.XdpMd{}
+	opts := RunOptions{
+		Data:       buf,
+		Context:    xdp,
+		ContextOut: &xdpOut,
```
lmb, line 124: Could this just be `&xdp` as well?

prog_test.go
```diff
@@ -90,6 +90,54 @@ func TestProgramRun(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if ret != 0 {
+		t.Error("Expected return value to be 0, got", ret)
+	}
+
+	if xdp != xdpOut {
+		t.Errorf("Expect xdp (%+v) != xdpOut (%+v)", xdp, xdpOut)
```
lmb, line 137: Incorrect error message?

prog_test.go
```diff
@@ -90,6 +90,54 @@ func TestProgramRun(t *testing.T) {
...
+	testutils.SkipIfNotSupported(t, err)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if ret != 0 {
+		t.Error("Expected return value to be 0, got", ret)
+	}
+
+	if xdp != xdpOut {
```
lmb, line 136: It might be better to assert that we actually decoded the kernel output ctx. Right now this doesn't really test anything.

prog.go
```diff
@@ -498,6 +498,23 @@ func (p *Program) Close() error {
 	return p.fd.Close()
 }
 
+// Various options for Run'ing a Program
+type RunOptions struct {
```
lmb, line 502: Name bikeshed. Timo suggested `RunContext`, but then we'd have `RunContext.Context`. `RunOptions` is consistent with other option structs we have in the library, but it's a little bit weird in that we actually write into the `Out` fields. Probably not a big deal, so I'm fine with `RunOptions`.

prog.go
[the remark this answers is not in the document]
lmb: > When working with converting Go types to/from binary, my mind immediately goes to encoding.Binary(Un)Marshaler.

I'm not a fan of either interface. Because they are usually implemented manually they invite bugs and bloat the code base. BinaryMarshaler also doesn't allow controlling allocation, which really should be possible if the reason to use BinaryMarshaler is performance in the first place.

Hence, I think that interface{} is actually better than the marshalers, since it allows us to cut out all that junk in the common case.

> I think even without this patch, BPF_PROG_TYPE_RAW_TRACEPOINT* was broken.

That's correct. I think Timo is saying that our new API should allow passing NULL in some fashion. We can do that as a follow up PR though, I think.

> I'd like to also better understand the concern about allocating for Benchmark().

The issue is that we allocate an output slice even if the caller discards it: https://github.com/cilium/ebpf/blob/e0825791eaf28e70c8c6403054c85b5d37450290/prog.go#L602

prog_test.go
danobi: It could, but I had it this way to be extra sure that serialization/deserialization path was being hit.

prog_test.go
danobi: Hmm could you clarify? The current code tests that `DataEnd` was decoded as 14. Cuz when it went into the `Run`, `xdpOut.DataEnd == 0`.

prog.go
```diff
@@ -577,37 +634,43 @@ var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() e
...
-	// enough, and panic if the kernel wrote past the end of the allocation.
-	// See https://patchwork.ozlabs.org/cover/1006822/
-	out := make([]byte, len(in)+outputPad)
+	ctx := new(bytes.Buffer)
+	var ctxOut []byte
+	if opts.Context != nil {
+		if err := binary.Write(ctx, internal.NativeEndian, opts.Context); err != nil {
+			return 0, 0, fmt.Errorf("cannot serialize context: %v", err)
+		}
+		if opts.ContextOut != nil {
```
lmb, line 656: Nit: why make this dependent on `.Context`? This can be a stand alone if branch i think.

prog.go
```diff
@@ -617,28 +680,37 @@ func (p *Program) testRun(in []byte, repeat int, reset func()) (uint32, []byte,
...
-		// Houston, we have a problem. The program created more data than we allocated,
-		// and the kernel wrote past the end of our buffer.
-		panic("kernel wrote past end of output buffer")
+	if opts.DataOut != nil {
+		if int(attr.DataSizeOut) > cap(opts.DataOut) {
+			// Houston, we have a problem. The program created more data than we allocated,
+			// and the kernel wrote past the end of our buffer.
+			panic("kernel wrote past end of output buffer")
+		}
+		opts.DataOut = opts.DataOut[:int(attr.DataSizeOut)]
```
lmb, line 702: This has no effect since `opts` is passed by value. `Run`, etc. should take `*RunOptions` I think.

prog.go
```diff
@@ -498,6 +498,29 @@ func (p *Program) Close() error {
...
+	// Program's context after Program has run. Must be a pointer or slice. Optional field.
+	ContextOut interface{}
+	// Number of times to run Program. Optional field. Defaults to 1.
+	Repeat uint32
+	// Optional flags.
+	Flags uint32
+	// CPU to run Program on. Optional field.
+	// Note not all program types support this field.
+	CPU uint32
+	// Called whenever the syscall is interrupted, and should be set to testing.B.ResetTime
```
lmb, line 519:
```suggestion
	// Called whenever the syscall is interrupted, and should be set to testing.B.ResetTimer
```

prog.go
```diff
@@ -498,6 +498,29 @@ func (p *Program) Close() error {
 	return p.fd.Close()
 }
 
+// Various options for Run'ing a Program
+type RunOptions struct {
+	// Program's data input. Required field.
+	Data []byte
+	// Program's data after Program has run. Caller must allocate. Optional field.
+	// If the data output from the kernel is larger than the provided buffer, Run will panic.
```
lmb, line 506: I think we should keep this as an implementation detail and not document it. On newer kernels we will get ENOSPC, and on older kernels only programs that extend the buffer may trigger this.

lmb (changes requested):
Looks great, thanks!

prog.go
danobi: Oh right, cuz I was relying on `ctx.Bytes()` to figure out ctxOut's size. Updated PR with reflection to get size.

lmb:
Hi @danobi, I'll only be able to take another look at this towards Wednesday afternoon, European timezone. Sorry!

prog.go
```diff
@@ -565,37 +622,44 @@ var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() e
...
 
-	// Older kernels ignore the dataSizeOut argument when copying to user space.
-	// Combined with things like bpf_xdp_adjust_head() we don't really know what the final
-	// size will be. Hence we allocate an output buffer which we hope will always be large
-	// enough, and panic if the kernel wrote past the end of the allocation.
-	// See https://patchwork.ozlabs.org/cover/1006822/
-	out := make([]byte, len(in)+outputPad)
+	var ctxOut []byte
+	if opts.ContextOut != nil {
+		ctxOut = make([]byte, reflect.Indirect(reflect.ValueOf(opts.ContextOut)).Type().Size())
```
lmb, line 647: Using reflection here is incorrect, since the size of a type in memory != the size of its encoded form. For example, go type may contain padding due to alignment which binary.Size will ignore.

```suggestion
		ctxOut = make([]byte, binary.Size(opts.ContextOut))
```

prog.go
```diff
@@ -565,37 +622,44 @@ var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() e
...
 
-	if uint(len(in)) > math.MaxUint32 {
-		return 0, nil, 0, fmt.Errorf("input is too long")
+	if err := haveProgTestRun(); err != nil {
+		return 0, 0, err
 	}
 
-	if err := haveProgTestRun(); err != nil {
-		return 0, nil, 0, err
+	ctx := new(bytes.Buffer)
```
lmb, line 638: Nit: this forces an allocation even if we don't have opts.Context. We can fix that in a follow up though.

lmb (changes requested):
This looks really nice, thanks for putting in so much work. I've found one problem, once that is gone this is good to go.

prog.go
```diff
@@ -565,37 +622,44 @@ var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() e
 	return err
 })
 
-func (p *Program) testRun(in []byte, repeat int, reset func()) (uint32, []byte, time.Duration, error) {
-	if uint(repeat) > math.MaxUint32 {
-		return 0, nil, 0, fmt.Errorf("repeat is too high")
+func (p *Program) testRun(opts *RunOptions) (uint32, time.Duration, error) {
+	if len(opts.Data) == 0 {
```
ti-mo, line 626: data_in must be zero for `SK_LOOKUP`, `RAW_TRACEPOINT` and `RAW_TRACEPOINT_WRITABLE`, so let's not make the assumption that `Data` is/needs to be populated.

ti-mo (commented):
A few remaining comments, probably ok to merge after these.

prog.go
dxuuu: Yeah I agree. I think there was some discussion above. This was a carry-over from existing behavior, so probably best to change in a separate PR. I can file a followup.

prog.go
dxuuu: Ok, will send followup

lmb (approved):

lmb:
Thanks for your contribution, sorry it took so long to get merged!

PR #649, opened 2022-05-01
Hard code endianness

brycekahle:
This results in a reduction of the instructions necessary for endian-related operations. The wall clock time difference for a single call is miniscule (picoseconds).

An example, with the `ByteOrder.Uint32` function:
Runtime instructions: https://godbolt.org/z/7vW1Pfnso
Buildtime instructions: https://godbolt.org/z/baxW1Wb7W

The function call on `amd64` changes from:
```
MOVQ    "".NativeEndian(SB), CX
MOVQ    "".NativeEndian+8(SB), DX
MOVQ    64(CX), CX
MOVQ    AX, BX
MOVL    $4, DI
MOVQ    DX, AX
MOVQ    CX, DX
MOVQ    DI, CX
CALL    DX
```
to
```
XCHGL   AX, AX
```

The difference on `arm64` is equivalent, but uses different specific instructions.

### DIFF
--- a/internal/endian.go
+++ b/internal/endian.go
@@ -1,3 +1,6 @@
+//go:build !386 && !amd64 && !amd64p32 && !arm && !arm64 && !mipsle && !mips64le && !mips64p32le && !ppc64le && !riscv64 && !armbe && !arm64be && !mips && !mips64 && !mips64p32 && !ppc64 && !s390 && !s390x && !sparc && !sparc64
+// +build !386,!amd64,!amd64p32,!arm,!arm64,!mipsle,!mips64le,!mips64p32le,!ppc64le,!riscv64,!armbe,!arm64be,!mips,!mips64,!mips64p32,!ppc64,!s390,!s390x,!sparc,!sparc64
+
 package internal
 
 import (
@@ -9,7 +12,7 @@ import (
 // depending on the host's endianness.
 var NativeEndian binary.ByteOrder
 
-// Clang is set to either "el" or "eb" depending on the host's endianness.
+// ClangEndian is set to either "el" or "eb" depending on the host's endianness.
 var ClangEndian string
 
 func init() {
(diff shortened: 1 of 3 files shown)


internal/endian_le.go
```diff
@@ -0,0 +1,10 @@
+//go:build amd64 || arm64
+// +build amd64 arm64
+
+package internal
+
+import "encoding/binary"
+
+var NativeEndian binary.ByteOrder = binary.LittleEndian
```
lmb, line 8: Please duplicate the comments so that generated documentation is consistent.

internal/endian.go
```diff
@@ -1,3 +1,6 @@
+//go:build !arm64 && !amd64
```
lmb, line 1: Can you use https://github.com/cilium/ebpf/blob/5fd10f532dbc2e31c5a3a1f715c4cef223998136/cmd/bpf2go/main.go#L44-L65 as a reference to make both little and big endian arches compile time?

lmb (changes requested):

lmb:
Interesting, thanks!

internal/endian_le.go
brycekahle: Done!

internal/endian.go
brycekahle: Done!

lmb:
I'm getting the following when trying to pull your branch:
```
$ git pull git@github.com:DataDog/ebpf.git bryce.kahle/endian-build-time
From github.com:DataDog/ebpf
 * branch            bryce.kahle/endian-build-time -> FETCH_HEAD
fatal: refusing to merge unrelated histories
```
Similar result when checking out via github CLI, it only shows me a single commit with all source code smushed into one blob. Did you do any unorthodox git mangling on your repo that could cause this?

brycekahle:
> Did you do any unorthodox git mangling on your repo that could cause this?

I don't believe so. Did you try `gh pr checkout 649`?

brycekahle:
> like you have maintainer edits disabled?

Apparently this isn't [supported](https://github.com/isaacs/github/issues/1681) for org-owned forks at the moment.

> Let me know if you want to force push that into your branch or if I should make a new PR.

Whatever is easier for you, is fine with me.

lmb:
See #662. Can you use a personal fork in the future?

brycekahle:
> Can you use a personal fork in the future?

Yeah. That shouldn't be a problem.

PR #651, opened 2022-05-02
Deprecate CollectionSpec.RewriteMaps()

ti-mo:
Some follow-ups to https://github.com/cilium/ebpf/pull/646.

Adds a deprecation notice to RewriteMaps().
Moves the Map(Spec) compatibility check out of the lazy-loading path.

cc @mauriciovasquezbernal

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -19,13 +19,15 @@ type CollectionOptions struct {
 	Maps     MapOptions
 	Programs ProgramOptions
 
-	// MapReplacements defines a set of maps that will be used
-	// instead of creating new ones when loading the object. The
-	// maps' specs should be compatible and there must exist a map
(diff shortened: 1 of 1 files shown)


collection.go
```diff
@@ -73,6 +75,9 @@ func (cs *CollectionSpec) Copy() *CollectionSpec {
 // when calling NewCollection. Any named maps are removed from CollectionSpec.Maps.
 //
 // Returns an error if a named map isn't used in at least one program.
+//
+// Deprecated: Pass CollectionOptions.MapReplacements when loading the Collection
+// instead. This function will be removed in version 1.0.
```
lmb, line 80:
```suggestion
// instead.
```

lmb (approved):

PR #652, opened 2022-05-02
syscalls: omit AttachType field in haveProgAttach

ti-mo:
Discovered while running the test suite on kernel 4.14.

With the addition BPF_PROG_ATTACH tests in package link, the haveProgAttach feature test was added, gated on 4.10. This helper uses the AttachType field, which sets the bpf_attr expected_attach_type field, available as of 4.17.

There are more issues to address for 4.14, but this patch fixes these tests on kernels between 4.10 and 4.17.

cc @rgo3

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -58,7 +58,12 @@ type ProgramSpec struct {
 	Name string
 
 	// Type determines at which hook in the kernel a program will run.
-	Type       ProgramType
+	Type ProgramType
+
+	// AttachType of the program, needed to differentiate allowed context
+	// accesses in some newer program types like CGroupSockAddr.
+	//
+	// Available on kernels 4.17 and later.
 	AttachType AttachType
 
 	// Name of a kernel data structure or function to attach to. Its
--- a/link/syscalls.go
+++ b/link/syscalls.go
@@ -27,9 +27,8 @@ const (
 
 var haveProgAttach = internal.FeatureTest("BPF_PROG_ATTACH", "4.10", func() error {
 	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
-		Type:       ebpf.CGroupSKB,
-		AttachType: ebpf.AttachCGroupInetIngress,
-		License:    "MIT",
+		Type:    ebpf.CGroupSKB,
+		License: "MIT",
 		Instructions: asm.Instructions{
 			asm.Mov.Imm(asm.R0, 0),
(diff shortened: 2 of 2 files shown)


lmb:
CI seems currently buggy, see the thread on slack.

prog.go
```diff
@@ -58,7 +58,13 @@ type ProgramSpec struct {
 	Name string
 
 	// Type determines at which hook in the kernel a program will run.
-	Type       ProgramType
+	Type ProgramType
+
+	// AttachType of the program, needed to differentiate allowed context
+	// accesses in some newer program types like CGroupSockAddr. Should be
+	// set to AttachNone if not required.
```
lmb, line 65: I think this is not necessary, since AttachNone is the default.

syscalls.go
```diff
@@ -244,3 +244,32 @@ var haveProbeReadKernel = internal.FeatureTest("bpf_probe_read_kernel", "5.5", f
 	_ = fd.Close()
 	return nil
 })
+
+var haveProgAttachType = internal.FeatureTest("expected_attach_type", "4.17", func() error {
```
lmb, line 248: This needs a corresponding call to `testutils.CheckFeatureTest` or whatever it's called.

lmb (commented):
Do we really need the fallback behaviour? What if we change the feature test to omit AttachType instead? I'm a bit fuzzy what the difference is when AttachType is present vs. not present to be honest, so I might be missing the point.

prog.go
ti-mo: True, it's a bit redundant. I copied this from the docstring on `type AttachType`, will remove it here.

syscalls.go
ti-mo: Good catch, added `TestHaveProgAttachType`.

ti-mo:
> What if we change the feature test to omit AttachType instead?

Turns out I had a .gitignored UAPI header in my linux repo that contained `expected_attach_type`, so it popped up in grep even after checking out `v4.10`.. :disappointed: This field wasn't added until 4.17, so this is just a bug in the test suite that we didn't notice because of the 4.9-4.19 gap.

And indeed, probing CGroupSKB is a representative proxy for `BPF_PROG_ATTACH`, since it was introduced in the same patch series as you commented in the code. No need for the AttachType field.

lmb (approved):

PR #653, opened 2022-05-02
prog: recognize ENOTSUPP in testRun()

ti-mo:
Not all program types support PROG_TEST_RUN on all kernel versions. Previously, when testing a program type that didn't have a test runner implemented, the following error would be returned:

> Error when running: can't test program: can't run test: errno 524

Since this isn't overly descriptive, make it clear to the caller that the kernel supports test runs, just not for the particular program type.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -590,6 +590,7 @@ func (p *Program) Benchmark(in []byte, repeat int, reset func()) (uint32, time.D
 
 var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() error {
 	prog, err := NewProgram(&ProgramSpec{
+		// SocketFilter does not require privileges on newer kernels.
 		Type: SocketFilter,
 		Instructions: asm.Instructions{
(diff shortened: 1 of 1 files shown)


prog.go
```diff
@@ -585,7 +585,9 @@ func (p *Program) Benchmark(in []byte, repeat int, reset func()) (uint32, time.D
 
 var haveProgTestRun = internal.FeatureTest("BPF_PROG_TEST_RUN", "4.12", func() error {
 	prog, err := NewProgram(&ProgramSpec{
-		Type: SocketFilter,
+		// The first PROG_TEST_RUN patches shipped in 4.12 only supported testing
+		// SKB and XDP programs.
+		Type: SchedACT,
```
lmb, line 590: I changed this to SocketFilter since that can be run without privileges on newer kernels. What if we keep SocketFilter add a special case for ENOTSUPP here as well?

lmb: ENOTSUPP in this case indicating that we have prog run support, since we otherwise would get EINVAL.

ti-mo: Yep, this works too!

PR #657, opened 2022-05-04
Probe kernel bpf2bpf support

ti-mo:
```
The adjustments performed by fixupAndValidate() don't rely on CO-RE being
applied first. Execute it earlier in the chain to potentially avoid loading
BTF and executing CO-RE when we could bail out instead.

Also avoids having to touch test cases due to changes in a subsequent commit.
```

```
When trying to load a program containing bpf2bpf calls on a kernel that
doesn't support it, we bubble up the verifier error to the caller:

  program xdp_prog: load program: invalid argument: unreachable insn 28

This is clear, but doesn't allow skipping tests when loading e.g. loader.c
on older kernels due to it containing bpf2bpf calls.

Implement a feature probe for bpf2bpf, return an unsupported error when
trying to load a program with multiple subprogs on an older kernel.

Update the test suite to skip unsupported features instead of hardcoding
a skip on 4.16 and earlier.
```

- Provide clear(er) and earlier error feedback about the kernel's bpf2bpf support
- Allow test skips when loading a bpf2bpf prog on a kernel that doesn't support it
- Remove hardcoded feature gates for bpf2bpf on kernel 4.16

### DIFF
--- a/linker.go
+++ b/linker.go
@@ -110,6 +110,17 @@ func findReferences(progs map[string]*ProgramSpec) error {
 	return nil
 }
 
+// hasReferences returns true if insns contains one or more bpf2bpf
+// function references.
+func hasReferences(insns asm.Instructions) bool {
+	for _, i := range insns {
+		if i.IsFunctionReference() {
+			return true
+		}
+	}
+	return false
+}
+
 // applyRelocations collects and applies any CO-RE relocations in insns.
 //
 // Passing a nil target will relocate against the running kernel. insns are
--- a/prog.go
+++ b/prog.go
@@ -361,6 +361,12 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		}
 	}
 
+	if (errors.Is(err, unix.EINVAL) || errors.Is(err, unix.EPERM)) && hasReferences(spec.Instructions) {
+		if err := haveBPFToBPFCalls(); err != nil {
+			return nil, fmt.Errorf("load program: %w", internal.ErrorWithLog(err, logBuf, logErr))
+		}
+	}
+
 	if errors.Is(logErr, unix.EPERM) && len(logBuf) > 0 && logBuf[0] == 0 {
 		// EPERM due to RLIMIT_MEMLOCK happens before the verifier, so we can
 		// check that the log is empty to reduce false positives.
--- a/syscalls.go
+++ b/syscalls.go
@@ -38,6 +38,21 @@ func invalidBPFObjNameChar(char rune) bool {
 	}
 }
 
+func progLoad(insns asm.Instructions, typ ProgramType, license string) (*sys.FD, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, insns.Size()))
+	if err := insns.Marshal(buf, internal.NativeEndian); err != nil {
+		return nil, err
+	}
+	bytecode := buf.Bytes()
+
+	return sys.ProgLoad(&sys.ProgLoadAttr{
+		ProgType: sys.ProgType(typ),
+		License:  sys.NewStringPointer(license),
+		Insns:    sys.NewSlicePointer(bytecode),
+		InsnCnt:  uint32(len(bytecode) / asm.InstructionSize),
+	})
+}
+
 var haveNestedMaps = internal.FeatureTest("nested maps", "4.12", func() error {
 	_, err := sys.MapCreate(&sys.MapCreateAttr{
 		MapType:    sys.MapType(ArrayOfMaps),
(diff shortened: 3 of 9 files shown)


prog.go
```diff
@@ -256,6 +256,10 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		return nil, fmt.Errorf("can't load %s program on %s", spec.ByteOrder, internal.NativeEndian)
 	}
 
+	if err := fixupAndValidate(spec.Instructions); err != nil {
```
lmb, line 259: This modifies spec.Instruction instead of operating on a copy. It's a bit opaque, currently the copy happens in core.Apply. This changes with the metadata PR.

syscalls.go
```diff
@@ -226,21 +241,30 @@ var haveProbeReadKernel = internal.FeatureTest("bpf_probe_read_kernel", "5.5", f
...
+
+var haveBPFToBPFCalls = internal.FeatureTest("bpf2bpf calls", "4.16", func() error {
+	insns := asm.Instructions{
+		asm.Call.Label("prog2").WithSymbol("prog1"),
+		asm.Return(),
+		asm.Mov.Imm(asm.R0, 0).WithSymbol("prog2"),
+		asm.Return(),
+	}
+
+	fd, err := progLoad(insns, SocketFilter, "MIT")
```
lmb, line 261: There is some trickiness with bpf2bpf calls, they require capabilities: https://elixir.bootlin.com/linux/v5.17.5/source/kernel/bpf/verifier.c#L2004 I think we could add a case for `EPERM` and treat that as supported (but not permitted)?

See my comment about relying on this in the first place above.

linker.go
```diff
@@ -169,6 +169,10 @@ func fixupAndValidate(insns asm.Instructions) error {
 			return fmt.Errorf("instruction %d: map %s: %w", iter.Index, ins.Reference(), asm.ErrUnsatisfiedMapReference)
 		}
 
+		if ins.IsFunctionReference() && haveBPFToBPFCalls() != nil {
+			return fmt.Errorf("kernel does not support bpf2bpf function calls: %w", haveBPFToBPFCalls())
```
lmb, line 173: The error message is going to be a bit repetitive here. `instruction %d: %w`?

linker.go
```diff
@@ -169,6 +169,10 @@ func fixupAndValidate(insns asm.Instructions) error {
 			return fmt.Errorf("instruction %d: map %s: %w", iter.Index, ins.Reference(), asm.ErrUnsatisfiedMapReference)
 		}
 
+		if ins.IsFunctionReference() && haveBPFToBPFCalls() != nil {
```
lmb, line 172: I'm a bit torn about this to be honest, refusing to load is a pretty big hammer. Feature tests are kind of brittle, so the less we rely on them the better.

It'd probably also be better to only call haveBPFToBPFCalls once per program instead of once per bpf2bpf call? It's guarded by a RWMutex, but its a mutex after all.

lmb (changes requested):

linker.go
ti-mo: Yes, but how to make it clear that it's lack of support in the kernel, not in the library?

linker.go
ti-mo: > It'd probably also be better to only call haveBPFToBPFCalls once per program instead of once per bpf2bpf call? It's guarded by a RWMutex, but its a mutex after all.

True, but I didn't want to call it if the prog didn't contain any func references.

> refusing to load is a pretty big hammer

Agreed, maybe it's better to use the probe as a hint when a prog load fails, like we do for EPERM. I'll mull this over for a bit.

prog.go
ti-mo: I've respun this for the check to take place if program creation fails, this should no longer be an issue.

prog.go
```diff
@@ -361,6 +361,13 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		}
 	}
 
+	if (errors.Is(logErr, unix.EINVAL) || errors.Is(logErr, unix.EPERM)) &&
```
lmb, line 364: I think `EINVAL` happens a lot, so we'll do another traversal of all instructions for most errors. Plus we'll allocate a slice just to get it's length. What if you return `hasB2BCalls bool` from `fixupAndValidate` or similar and gate on that here instead? Alternatively, a function that returns on the first found call would also be an improvement.

lmb:
Thanks, checking after loading is nicer!

ti-mo:
@lmb I've found a bug introduced in 9b5a039e00c6df7f25e76eac13fede067b0a48a1 and made some changes to `ErrorWithLog`. PTAL :pray: 

I've also added `hasReferences()` and looked at `err` instead of `logErr`.

lmb (approved):
Looks good, thanks for looking at ErrorWithLog as well

PR #658, opened 2022-05-04
Assorted fixes for kernel 4.14

ti-mo:
- Run `TestMapLoadPinnedWithOptions` as of 4.15
- Don't specify AttachType in `mustCgroupFixtures`
- Remove map pointer from `TestProgramInstructions`

Together with #657, this makes tests pass on our 4.14 ci-kernels image. To be enabled later when we move to self-hosted CI runners.

### DIFF
--- a/prog_test.go
+++ b/prog_test.go
@@ -687,17 +687,12 @@ func TestProgramBindMap(t *testing.T) {
 }
 
 func TestProgramInstructions(t *testing.T) {
-	arr := createArray(t)
-	defer arr.Close()
-
 	name := "test_prog"
 	spec := &ProgramSpec{
 		Type: SocketFilter,
 		Name: name,
 		Instructions: asm.Instructions{
 			asm.LoadImm(asm.R0, -1, asm.DWord).WithSymbol(name),
-			asm.LoadMapPtr(asm.R1, arr.FD()),
-			asm.Mov.Imm32(asm.R0, 0),
 			asm.Return(),
 		},
 		License: "MIT",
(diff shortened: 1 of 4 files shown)


prog_test.go
```diff
@@ -687,17 +687,12 @@ func TestProgramBindMap(t *testing.T) {
 }
 
 func TestProgramInstructions(t *testing.T) {
```
lmb, line 689: Does comparing the Tag in our tests make sense in the first place if the output is so unpredictable?

lmb: Might be worth an issue as well.

lmb (approved):

prog_test.go
ti-mo: We compare Tag instead of comparing Instructions. I don't think it does any harm, and when a check like this starts to fail, it usually leads to an interesting discovery, like in this case.

> Might be worth an issue as well.

What do you mean?

lmb: Can we sanitise the Instructions ourselves? Should we store map ID in metadata?

ti-mo: For the test case as it was before this patch; not really. Given the following prog:

```
	0: LdImmDW dst: r0 imm: -1
	2: LoadMapPtr dst: r1 fd: 6
	4: Mov32Imm dst: r0 imm: 0
	5: Exit
```

This is what the xlated insns look like:
```
	0: LdImmDW dst: r0 imm: -1
	2: LdImmDW dst: r1 imm: -103942818276608
	4: Mov32Imm dst: r0 imm: 0
	5: Exit
```

Src of instruction 2 is unset. Sanitizing the pointer is no problem, but no way to figure out this is a map load if I understand correctly.

PR #659, opened 2022-05-04
btf: move type IDs into Spec

### LINKED ISSUE #643: btf: remove Type.ID()
The `Type` interface in `btf` currently has an `ID()` method. This is a problem since a type doesn't really have an ID, it's only when you have a list of types that we can assign IDs.

As such, we should remove `Type.ID()` and instead provide a function on `Spec` that allows looking up the ID of a Type.

lmb:
The ID of a type is really its index in the list of types passed to the kernel
when loading BTF. As such, it's incorrect for a Type to have an ID. The ID only
exists for []Type. Remove ID() from the Type interface, and instead provide a
function to look up an ID from a Spec. This function requires a map[Type]TypeID
to avoid a linear scan of the potentially large Spec.types. It seems like removing
the type ID from types themselves does make up for the increase however.

    ParseVmlinux-4    58.6ms ± 1%    61.7ms ± 1%  +5.40%  (p=0.029 n=4+4)

    ParseVmlinux-4    45.2MB ± 0%    44.2MB ± 0%  -2.20%  (p=0.029 n=4+4)

    ParseVmlinux-4      721k ± 0%      718k ± 0%  -0.32%  (p=0.029 n=4+4)

Fixes #643

### DIFF
--- a/internal/btf/btf.go
+++ b/internal/btf/btf.go
@@ -43,6 +43,9 @@ type Spec struct {
 	// Includes all struct flavors and types with the same name.
 	namedTypes map[essentialName][]Type
 
+	// Type IDs indexed by type.
+	typeIDs map[Type]TypeID
+
 	byteOrder binary.ByteOrder
 }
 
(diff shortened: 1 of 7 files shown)


dylandreimerink (approved):
LGTM

ti-mo (approved):
Very clean, one nit :ok_hand:

PR #660, opened 2022-05-04
btf: introduce Bits type and remove Int.Bits and Int.Offset

lmb:
btf: introduce Bits type

    BTF expresses some lengths in bits, others in bytes. So far we don't have a
    good convention how to distinguish the two units. Add a Bits type to make
    the distinction clearer.

btf: remove Int.Bits and Int.Offset

    Unfortunately BTF has two ways to encode bitfields. You can
    create a Member and set Offset / BitfieldSize or create an Int that
    has the offset and bitfield size baked in. The background for this
    can be found in [1].

    Remove Int.Bits and Int.Offset in favour of Member.BitfieldSize and
    Member.Offset. Also get rid of a bunch of incorrect checks for bitfields,
    which we didn't catch due to the weird encoding. This cleans up the API,
    but may make our lives worse if we end up having to support writing out
    legacy bitfields. Let's hope it doesn't come to that.

    The performance of parsing vmlinux hasn't changed dramatically:

        ParseVmlinux-4    62.1ms ± 1%    61.8ms ± 2%    ~     (p=0.686 n=4+4)

        ParseVmlinux-4    44.2MB ± 0%    44.7MB ± 0%  +1.14%  (p=0.029 n=4+4)

        ParseVmlinux-4      718k ± 0%      718k ± 0%  +0.00%  (p=0.029 n=4+4)

    1: https://github.com/torvalds/linux/commit/9d5f9f701b1891466fb3dbb1806ad97716f95cc3


lmb:
Added `TestInflateLegacyBitfield` after discussing with Timo.

ti-mo (approved):

PR #667, opened 2022-05-10
ringbuf: reduce allocations and syscalls

lmb:
ringbuf: reduce allocations on the discard and busy paths

    There is no need to wrap internal sentinel errors, since they are
    never returned to the user. Not wrapping saves us some allocations.

ringbuf: reduce the number of epoll_wait calls

    We don't have to call back into epoll_wait until we've read the
    full contents of the ring buffer. This reduces the number of syscalls
    made during busy periods.

### DIFF
--- a/ringbuf/reader.go
+++ b/ringbuf/reader.go
@@ -66,7 +66,7 @@ func readRecord(rd *ringbufEventRing, rec *Record, buf []byte) error {
 		// the next sample in the ring is not committed yet so we
 		// exit without storing the reader/consumer position
 		// and start again from the same position.
-		return fmt.Errorf("%w", errBusy)
+		return errBusy
 	}
 
 	/* read up to 8 byte alignment */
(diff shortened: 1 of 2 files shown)


ringbuf/reader.go
```diff
@@ -167,16 +169,25 @@ func (r *Reader) Read() (Record, error) {
...
 
-		record, err := readRecord(r.ring)
-		if errors.Is(err, errBusy) || errors.Is(err, errDiscard) {
-			continue
+		for {
+			record, err := readRecord(r.ring)
+			if err == errBusy || err == errDiscard {
+				continue
+			}
+			if err == errEOR {
```
florianl, line 185: Is this a left over from previous experiments? `readRecord()` does not return `errEOR` afaict.

Other than that, the changes work fine here.

florianl, line 185: Is this a left over from previous experiments? `readRecord()` does not return `errEOR` afaict.

Other than that, the changes work fine here.

lmb: Thanks for catching that! Modified one of the tests to exercise this behavior.

mythi:
@lmb thanks! I was just about to submit my github approval. It looks nice to me. AFAICS with this users can also start controlling samples submission better with `linux.BPF_RB_NO_WAKEUP` and `linux.BPF_RB_FORCE_WAKEUP`.

PR #671, opened 2022-05-11
Export `LoadKernelSpecUncached` function to load kernel spec

paulcacheux:
With https://github.com/cilium/ebpf/pull/665 being merged, we can expect users that would use `cilium/ebpf` only for the BTF parsing/exploration facilities. For some of them the cache used by `LoadKernelSpec` might not be desirable, for example if they only want to load the BTF at one point of time and be done with it.

This PR exports the inner function that load the kernel spec.

As a side note: I'm not a huge fan of the name `LoadKernelSpecUncached` but I can't find a better one. Happy to update this PR with another name if desired (maybe `LoadKernelSpecRaw` or `LoadKernelSpecWithoutCache` ?)

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -273,14 +273,14 @@ func LoadKernelSpec() (*Spec, error) {
 	}
 
 	var err error
-	kernelBTF.Spec, err = loadKernelSpec()
+	kernelBTF.Spec, err = LoadKernelSpecUncached()
 	return kernelBTF.Spec, err
 }
 
-// loadKernelSpec attempts to load the raw vmlinux BTF blob at
+// LoadKernelSpecUncached attempts to load the raw vmlinux BTF blob at
 // /sys/kernel/btf/vmlinux and falls back to scanning the file system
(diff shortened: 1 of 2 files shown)


lmb:
Do you have a situation where this shows up in profiles or similar? If contention is an issue we can change `kernelBTF.Mutex` to an `RWMutex`, I'm hesitant to add a new API for this.

paulcacheux:
Putting a bit more thought into it, I think you are right. I will close this

ti-mo:
@lmb Alternatively, I thought we could also make `LoadKernelSpec` always parse vmlinux BTF, and provide an unexported wrapper that does the caching for our uses in the root package in `applyRelocations()` and `findTargetInKernel()`.

That way, if the caller never loads an ELF with CO-RE or prog.AttachTo, they would never incur this memory overhead, and they could decide for themselves to hang on to the spec or discard it.

I would expect most users to already cache the spec anyway, since it currently requires reading the source code to figure out the library already does this under the covers, it's not documented.

PR #673, opened 2022-05-13
features: add HaveMapFlag API

paulcacheux:
`HaveMapFlag(flag uint32) error` would allow a user to probe running kernel for support for map flags.
The main target of this would be to probe for availability of mmapable array maps.

This PR could also be replaced by just exporting https://github.com/cilium/ebpf/blob/314c7732ecc76e7587edf487d1048fbc165983e3/syscalls.go#L91

### DIFF
--- a/features/map.go
+++ b/features/map.go
@@ -13,6 +13,7 @@ import (
 
 func init() {
 	mc.mapTypes = make(map[ebpf.MapType]error)
+	mc.mapFlags = make(map[uint32]error)
 }
 
 var (
@@ -22,6 +23,7 @@ var (
 type mapCache struct {
 	sync.Mutex
 	mapTypes map[ebpf.MapType]error
+	mapFlags map[uint32]error
 }
 
 func createMapTypeAttr(mt ebpf.MapType) *sys.MapCreateAttr {
@@ -120,29 +122,8 @@ func haveMapType(mt ebpf.MapType) error {
 		fd.Close()
 	}
 
-	switch {
-	// For nested and storage map types we accept EBADF as indicator that these maps are supported
-	case errors.Is(err, unix.EBADF):
-		if isMapOfMaps(mt) || isStorageMap(mt) {
-			err = nil
-		}
-
-	// ENOTSUPP means the map type is at least known to the kernel.
-	case errors.Is(err, sys.ENOTSUPP):
-		if mt == ebpf.StructOpsMap {
-			err = nil
-		}
-
-	// EINVAL occurs when attempting to create a map with an unknown type.
-	// E2BIG occurs when MapCreateAttr contains non-zero bytes past the end
-	// of the struct known by the running kernel, meaning the kernel is too old
-	// to support the given map type.
-	case errors.Is(err, unix.EINVAL), errors.Is(err, unix.E2BIG):
-		err = ebpf.ErrNotSupported
-	}
-
+	err = convertProbeError(mt, err)
 	mc.mapTypes[mt] = err
-
 	return err
 }
 
@@ -159,6 +140,84 @@ func isStorageMap(mt ebpf.MapType) bool {
 	case ebpf.SkStorage, ebpf.InodeStorage, ebpf.TaskStorage:
 		return true
 	}
-
 	return false
+}
+
+// HaveMapFlag probes the running kernel for the availability of the specified map flag.
+//
+// See the package documentation for the meaning of the error return value.
+func HaveMapFlag(flag uint32) (err error) {
+	defer func() {
+		// This closure modifies a named return variable.
+		err = wrapProbeErrors(err)
+	}()
+
+	return haveMapFlag(flag)
+}
+
+func haveMapFlag(flag uint32) error {
+	mc.Lock()
+	defer mc.Unlock()
+	err, ok := mc.mapFlags[flag]
+	if ok {
+		return err
+	}
+
+	attr, err := createMapFlagTypeAttr(flag)
+	if err != nil {
+		return err
+	}
+
+	fd, err := sys.MapCreate(attr)
+	if err == nil {
+		fd.Close()
+	}
+
+	err = convertProbeError(ebpf.MapType(attr.MapType), err)
+	mc.mapFlags[flag] = err
+	return err
+}
+
+func createMapFlagTypeAttr(flag uint32) (*sys.MapCreateAttr, error) {
+	a := &sys.MapCreateAttr{
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 1,
+		MapFlags:   flag,
+	}
+
+	switch flag {
+	case unix.BPF_F_MMAPABLE, unix.BPF_F_INNER_MAP, unix.BPF_F_RDONLY_PROG, unix.BPF_F_WRONLY_PROG:
+		a.MapType = sys.MapType(ebpf.Array)
+		return a, nil
+	case unix.BPF_F_NO_PREALLOC:
+		a.MapType = sys.MapType(ebpf.Hash)
+		return a, nil
+	}
+
+	return nil, ebpf.ErrNotSupported
+}
+
+func convertProbeError(mt ebpf.MapType, err error) error {
+	switch {
+	// For nested and storage map types we accept EBADF as indicator that these maps are supported
+	case errors.Is(err, unix.EBADF):
+		if isMapOfMaps(mt) || isStorageMap(mt) {
+			return nil
+		}
+
+	// ENOTSUPP means the map type is at least known to the kernel.
+	case errors.Is(err, sys.ENOTSUPP):
+		if mt == ebpf.StructOpsMap {
+			return nil
+		}
+
+	// EINVAL occurs when attempting to create a map with an unknown type.
+	// E2BIG occurs when MapCreateAttr contains non-zero bytes past the end
+	// of the struct known by the running kernel, meaning the kernel is too old
+	// to support the given map type.
+	case errors.Is(err, unix.EINVAL), errors.Is(err, unix.E2BIG):
+		return ebpf.ErrNotSupported
+	}
+	return err
 }
--- a/features/map.go
+++ b/features/map.go
@@ -122,8 +122,29 @@ func haveMapType(mt ebpf.MapType) error {
 		fd.Close()
 	}
 
-	err = convertProbeError(mt, err)
+	switch {
+	// For nested and storage map types we accept EBADF as indicator that these maps are supported
+	case errors.Is(err, unix.EBADF):
+		if isMapOfMaps(mt) || isStorageMap(mt) {
+			err = nil
+		}
+
+	// ENOTSUPP means the map type is at least known to the kernel.
+	case errors.Is(err, sys.ENOTSUPP):
+		if mt == ebpf.StructOpsMap {
+			err = nil
+		}
+
+	// EINVAL occurs when attempting to create a map with an unknown type.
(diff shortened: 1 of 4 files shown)


rgo3:
Thanks for the PR @paulcacheux, I'll give it a closer look tomorrow. Back when we added the `HaveMapType` API there was some discussion around adding a `HaveMapFlags` API, roughly starting at this [message](https://github.com/cilium/ebpf/issues/235#issuecomment-871386930) on #235. Maybe you can give it a look in the meantime. But the gist of it was that having `HaveMapFlag(MapType, Flag)` opens up a lot of possible combinations that we would need to test for.

features/map.go
```diff
@@ -129,37 +147,61 @@ func haveMapType(mt ebpf.MapType) error {
...
 	// Wrap unexpected errors.
 	case err != nil:
-		err = fmt.Errorf("unexpected error during feature probe: %w", err)
+		return fmt.Errorf("unexpected error during feature probe: %w", err)
 
-	default:
-		fd.Close()
 	}
 
-	mc.mapTypes[mt] = err
```
florianl, line 160: Removing this makes the caching obsolete and performs the check every time. Is there a particular reason to do so?

paulcacheux: This is done [here](https://github.com/cilium/ebpf/pull/673/files#diff-df55ec79721c085fabb3b83f2498ad96eff807311889d298b09693c4e1cda8a4R152) now. Because this part is extracted into a separate function

features/map.go
```diff
@@ -129,37 +147,61 @@ func haveMapType(mt ebpf.MapType) error {
...
+	attr.MapFlags |= flags
+	err = checkMapCreation(mt, attr)
+
+	mc.mapFlags[cacheEntry] = err
+
+	return err
+}
+
+func checkMapCreation(mt ebpf.MapType, attr *sys.MapCreateAttr) error {
 	fd, err := sys.MapCreate(createMapTypeAttr(mt))
```
rgo3, line 177:
```suggestion
	fd, err := sys.MapCreate(attr)
```
Flags won't get set otherwise.

features/map_test.go
```diff
@@ -64,9 +65,25 @@ func TestHaveMapType(t *testing.T) {
 				t.Fatalf("Map type %s isn't supported even though kernel is at least %s: %v", mt.String(), minVersion, err)
 			}
 		})
-
 	}
+}
+
+var mapFlagMinVersion = map[mapFlagCacheEntry]string{
+	{ebpf.Array, unix.BPF_F_MMAPABLE}: "5.5",
```
rgo3, line 72: We should cover more combinations with tests and test failing outcomes as well.

features/map_test.go
```diff
@@ -64,9 +65,25 @@ func TestHaveMapType(t *testing.T) {
...
+}
+
+func TestHaveMapFlags(t *testing.T) {
+	for entry, minVersion := range mapFlagMinVersion {
+		feature := fmt.Sprintf("%s:%d", entry.mt.String(), entry.flags)
```
rgo3, line 77: `entry.flags` will print as a decimal number, we should somehow print the flag as its string value in the test output.

rgo3 (changes requested):
Apart from my comments I think there still needs to be more investigation to see whether or not `HaveMapFlags(MapType, Flag)` is exhaustive enough to be able to check all Map/Flag combinations correctly. Does any flag need additional things set up in `sys.MapCreateAttr` so the probe runs successfully? It's probably fine to not be able to test every flag, as long as we document what works, what doesn't and have respective tests showcasing that.

As you and @lmb mentioned something like `HaveMmapableMaps()` could be a simpler alternative (only one MapType that supports it) if a more generalized approach has too much complexity to get it right.

features/map.go
```diff
@@ -129,37 +147,61 @@ func haveMapType(mt ebpf.MapType) error {
...
 	// of the struct known by the running kernel, meaning the kernel is too old
 	// to support the given map type.
 	case errors.Is(err, unix.EINVAL), errors.Is(err, unix.E2BIG):
-		err = ebpf.ErrNotSupported
+		return ebpf.ErrNotSupported
 
 	// EPERM is kept as-is and is not converted or wrapped.
 	case errors.Is(err, unix.EPERM):
-		break
+		return nil
```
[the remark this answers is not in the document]
paulcacheux: Fixed in last commit

features/map.go
paulcacheux: Fixed in last commit

features/map_test.go
paulcacheux: Added more tests, covering the internal feature tests functions

ti-mo:
@paulcacheux Feel free to swing by `#ebpf-go-dev` on our community Slack if you want to discuss this more interactively. We have a few questions around this proposal:

1. What is the practical use case? Please include this in any future contributions so we get a better idea of what you're trying to achieve. (also, commit message :pray:)
2. Would an API that only takes a `flag` (not a map type) be sufficient? We could probe using a map type that is known to support the flag.
3. Building on the previous point, are there known cases in the kernel's development history where a flag was introduced on one map type, and later implemented on more map types? This is important to know if we can get away with (2).

Something else to take into account is the caller can probe for map type A with flag B, but the kernel doesn't necessarily support map type A. Of course, this means that the (A, B) combo is definitely not supported, but this overlaps a bit too much with `HaveMapType`. Knowing more about the use case would really help here.

Thanks!

features/map.go
```diff
@@ -120,29 +122,8 @@ func haveMapType(mt ebpf.MapType) error {
...
-
-	// EINVAL occurs when attempting to create a map with an unknown type.
-	// E2BIG occurs when MapCreateAttr contains non-zero bytes past the end
-	// of the struct known by the running kernel, meaning the kernel is too old
-	// to support the given map type.
-	case errors.Is(err, unix.EINVAL), errors.Is(err, unix.E2BIG):
-		err = ebpf.ErrNotSupported
-	}
-
+	err = convertProbeError(mt, err)
```
lmb, line 125: I think reusing this logic for flag detection is incorrect: we craft some very specific map attributes to elicit specific Errno. The same error isn't necessarily caused when flags are present. I'd suggest you leave this part untouched.

features/map.go
```diff
@@ -159,6 +140,84 @@ func isStorageMap(mt ebpf.MapType) bool {
...
+	if ok {
+		return err
+	}
+
+	attr, err := createMapFlagTypeAttr(flag)
+	if err != nil {
+		return err
+	}
+
+	fd, err := sys.MapCreate(attr)
```
lmb, line 171: In general `HaveMapFlag` needs to do two feature probes:
1. Is the map type we use to probe supported?
2. Is the flag on that map type supported?

We know that all supported kernels have Array and Hash, so it's OK to skip the first check here and only check for (2). Instead of calling `convertProbeError` I'd therefore just check that `err == nil` to determine whether the flag is supported.

features/map.go
```diff
@@ -159,6 +140,84 @@ func isStorageMap(mt ebpf.MapType) bool {
...
 		return true
 	}
-
 	return false
 }
+
+// HaveMapFlag probes the running kernel for the availability of the specified map flag.
+//
+// See the package documentation for the meaning of the error return value.
+func HaveMapFlag(flag uint32) (err error) {
```
lmb, line 149: There is no way to discover which flags the library supports probing for.

I think we should have a `type MapFlag uint32` here with aliases to `unix.BPF_F_*` consts defined for the flags that we support probing for.

`MapFlag` could also be in `ebpf` proper, but that raises the question of circular imports again.

lmb (changes requested):

features/map.go
paulcacheux: Noted, this reverted in the latest commit

features/map.go
paulcacheux: Agreed, maybe you would like be to add a comment explaining why we skip (1) for now, but it may be necessary in the future ? I changed the convertProbeError to just check for `EINVAL` and convert it to `ErrNotSupported` in the latest commit

features/map.go
paulcacheux: I agree with you but I just matched what the rest of the API is using, like https://pkg.go.dev/github.com/cilium/ebpf#Map.Flags. Maybe we should change this as well (but this would warrants another PR)

internal/unix/types_other.go
```diff
@@ -27,14 +27,14 @@ const (
 	EACCES = syscall.Errno(0)
 	EILSEQ = syscall.Errno(0)
 
-	BPF_F_NO_PREALLOC        = 0
+	BPF_F_NO_PREALLOC        = 1 << 0
```
rgo3, line 30: Why do we need to touch those? I always thought they were set to 0 on purpose.

rgo3 (changes requested):

internal/unix/types_other.go
lmb: The Go compiler rejects the switch statement due to duplicate values I think.

features/map.go
lmb: I think we can add `sys.MapFlags` and re-export that via type aliases.

internal/unix/types_other.go
paulcacheux: Yes this is the issue, that it's trying to fix

features/map.go
```diff
@@ -159,6 +140,84 @@ func isStorageMap(mt ebpf.MapType) bool {
...
+	switch flag {
+	case unix.BPF_F_MMAPABLE, unix.BPF_F_INNER_MAP, unix.BPF_F_RDONLY_PROG, unix.BPF_F_WRONLY_PROG:
+		a.MapType = sys.MapType(ebpf.Array)
+		return a, nil
+	case unix.BPF_F_NO_PREALLOC:
+		a.MapType = sys.MapType(ebpf.Hash)
+		return a, nil
+	}
+
+	return nil, ebpf.ErrNotSupported
```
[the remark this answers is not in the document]
paulcacheux: very good point, I will change that

lmb (approved):
I'm okay with this if you're okay with making a follow up PR for https://github.com/cilium/ebpf/pull/673/files#r942294110

@rgo3 any concerns from your side?

features/map.go
lmb: Adding that comment would be good, thanks!

rgo3 (approved):
No, apart from adding that missing comment and maybe creating an issue to add the MapFlag type as a follow up it looks good.

paulcacheux:
Just added the comment, absolutely agreed for the followup PR. I will launch a discussion on slack on the best way to implement this

lmb:
@paulcacheux thanks for reviving this PR! Can you tell me when you'll be able to do the follow up work? I'd like the new map flags type before we tag any more releases.

paulcacheux:
I have been working a bit on the follow up work, the issue I'm facing is that the map flags are defined in an anonymous enum by the kernel (https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L1192), making it not easy to generate the go code like it's done for maps. Here is a PR to start the work on https://github.com/cilium/ebpf/pull/766

lmb:
Cool! I think for now it's ok to:

1. Add `sys.MapFlags`
2. Make the various `sys.MapCreate`, etc. struct use `MapFlags`
3. Alias `ebpf.MapFlags`, `features.MapFlags` to `sys.MapFlags`.
4. Add `const (...)` to `features`

Tackling the anonymous enum can happen at some other point.

PR #675, opened 2022-05-16
Make BTF for data sections optional

ti-mo:
This is a workaround for a compiler bug/limitation when working with arrays declared as preprocessor macros.

This bit us on two occasions:

- https://github.com/cilium/cilium/pull/18602 (@borkmann)
- https://github.com/cilium/cilium/issues/18616(@vincentmli)

Discussion on the list:
https://lore.kernel.org/bpf/CAK3+h2wcDceeGyFVDU3n7kPm=zgp7r1q4WK0=abxBsj9pyFN-g@mail.gmail.com

Best reviewed per commit.


lmb:
I'll try to review this Wednesday afternoon after #647.

btf/types.go
```diff
@@ -514,6 +514,13 @@ type VarSecinfo struct {
 	Size   uint32
 }
 
+func (vsi *VarSecinfo) TypeName() string {
```
lmb, line 517: Seems a bit confusing, since `TypeName` is part of `Type` but `VarSecInfo` doesn't implement the latter. Also missing context, commit message?

testdata/loader.c
```diff
@@ -166,3 +166,26 @@ __section("socket/3") int data_sections() {
...
+ * Up until LLVM 14, this program results in an .rodata.cst32 section
+ * that is accessed by 'return values[i]'. For this section, no BTF is
+ * emitted. 'values' cannot be rewritten, since there is no BTF info
+ * describing the data section.
+ */
+__section("socket/4") int anon_const() {
+	volatile int ctx = 0;
+
+// 32 bytes wide results in a .rodata.cst32 section.
+#define values \
```
lmb, line 183: Could we just fake a `.rodata.foo` via explicit section annotations? Might be more robust on newer versions of LLVM as well.

elf_reader_test.go
```diff
@@ -131,6 +131,12 @@ func TestLoadCollectionSpec(t *testing.T) {
 				SectionName: "static",
 				License:     "MIT",
 			},
+			"anon_const": {
```
lmb, line 134: Seems like there is no fix for the problem here, this means that the tests will fail on this commit? If yes, that's problematic when trying to bisect problems.

elf_reader.go
```diff
@@ -1041,9 +1032,16 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
 			MaxEntries: 1,
 			Contents:   []MapKV{{uint32(0), data}},
-			Key:        &btf.Void{},
-			Value:      datasec,
-			BTF:        ec.btf,
+		}
+
+		// It is possible for a data section to exist without a corresponding BTF Datasec
+		// if it only contains anonymous values like macro-defined arrays.
+		value := &btf.Datasec{}
```
lmb, line 1048: Why allocate value here, not in the if branch below?

collection.go
```diff
@@ -122,34 +122,63 @@ func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
...
+			}
+
+			b, err := marshalBytes(replacement, int(v.Size))
+			if err != nil {
+				return fmt.Errorf("marshaling constant replacement %s: %w", vname, err)
+			}
+
+			copy(cpy[v.Offset:v.Offset+v.Size], b)
+
+			replaced[vname] = true
```
lmb, line 165: Should we allow replacing the same variable in multiple rodata sections? Seems wonky.

lmb (commented):
Looks fine to me, please make sure that running tests in the individual commits doesn't fail though.

btf/types.go
lmb: Reading through the rest of the PR I still don't know where this is used.

testdata/loader.c
ti-mo: I've tried all possible ways I could think of to force this to be emitted to a particular section, but haven't managed. Since it's not a real symbol, clang doesn't support setting the section attribute on it.

elf_reader_test.go
ti-mo: Okay, didn't know we really cared about that, will respin the set to take this into account.

collection.go
ti-mo: What's the reasoning behind not allowing it? Detecting compiler bugs? Technically, the same var being present in multiple rodata sections would also require having 2 equally-named ELF symbols with global vis, which is wonky indeed. So, we error here instead?

ti-mo: Done, added a check that makes sure we don't replace the same var twice.

elf_reader_test.go
ti-mo: Done.

collection.go
ti-mo: I guess this could also happen when the caller copies .rodata from a CollectionSpec and adds it under a different key in `Maps`.

PR #677, opened 2022-05-17
btf: Add getter for Spec.namedTypes

brb:
This commit adds the `(*Spec).GetNamedTypes()` getter.

One consumer of such an API is "pwru" \[1\] which needs to iterate over
all types of a given spec.

\[1\]: https://github.com/cilium/pwru

---

I didn't bother making the `essentialName` type public, as my consumer doesn't need to access the returned map by a key (it iterates over it with the `range` operator).

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -655,6 +655,11 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
 	return nil
 }
 
+// GetNamedTypes returns a reference to named types.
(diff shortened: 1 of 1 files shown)


brb:
Discussed offline with @ti-mo offline. He suggested that it's not safe to return a ref to `namedTypes`, as it needs to stay in sync with other `Spec` fields, and he suggested to add an iterator instead.

PR #678, opened 2022-05-18
btf: Add spec types iterator

brb:
This commit introduces the `TypesIterator` which is used to iterate over all types of a given spec.

One notable user is "pwru" \[1\] which needs to iterate over all vmlinux BTF types to find functions which accept SKB as a param.

The iterator is very similar to the existing `InstructionIterator`, as both export the `Next()` method and set the values and indices in the public fields.

\[1\]: https://github.com/cilium/pwru

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -655,6 +655,31 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
 	return nil
 }
 
+// TypesIterator iterates over types of a given spec.
+type TypesIterator struct {
+	spec *Spec
+	// The index of the last visited type in the spec.
+	Index int
+	// The last visited type in the spec.
+	Type Type
+}
+
+// NewTypesIterator returns the types iterator.
+func (s *Spec) NewTypesIterator() *TypesIterator {
+	return &TypesIterator{spec: s, Index: 0}
+}
+
+// Next returns true as longs as there are any remaining types.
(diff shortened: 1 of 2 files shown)


btf/btf.go
```diff
@@ -655,6 +655,31 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
...
+type TypesIterator struct {
+	spec *Spec
+	// The index of the last visited type in the spec.
+	Index int
+	// The last visited type in the spec.
+	Type Type
+}
+
+// NewTypesIterator returns the types iterator.
+func (s *Spec) NewTypesIterator() *TypesIterator {
```
lmb, line 668: Nit: Just `Iterate` would be ok I think, in line with Map.Iterate.

btf/btf.go
```diff
@@ -655,6 +655,31 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
 	return nil
 }
 
+// TypesIterator iterates over types of a given spec.
+type TypesIterator struct {
+	spec *Spec
+	// The index of the last visited type in the spec.
+	Index int
```
lmb, line 662: Do you need to expose the index? I'd prefer to keep it an implementation detail for now.

lmb (changes requested):

tklauser (approved):

lmb (approved):
Thanks!

PR #679, opened 2022-05-19
program: remove references from ProgramSpec

lmb:
Now that we don't need ProgramSpec.layout anymore to reconstruct
CO-RE relocations we can remove ProgramSpec.references. At the
same time we can simplify the linking logic.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"sort"
 	"strings"
 
 	"github.com/cilium/ebpf/internal/sys"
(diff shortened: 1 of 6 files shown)


ti-mo (commented):
You've revived the old linking code in the form of `flattenInstructions` and I still find it hard to wrap my head around coming back to it a few months later.

Could we maybe go at this another way and introduce a step that resolves stringly prog references to `*ProgramSpec`s in insn metadata? We could support either a string or `*ProgramSpec` in the same meta field. This would mean we no longer need to pass `refs` around, which will make the linking code a lot easier to understand.

lmb:
> I still find it hard to wrap my head around coming back to it a few months later.

Can you be more specific, what part of it is hard to understand? From the rest of you comment it seems `refs` is kind of ambiguous?

> Could we maybe go at this another way and introduce a step that resolves stringly prog references to *ProgramSpecs in insn metadata?

I think it only makes sense to use metadata if we need to pass info from the ELF loader to NewProgram or similar. So if the only use case is in the ELF loader we should come up with a different solution.

PR #682, opened 2022-05-24
Add `RelativeOffset` field to `UprobeOptions` struct

mertyildiran:
Alternative to https://github.com/cilium/ebpf/pull/683

This PR, for a [bpftrace](https://github.com/iovisor/bpftrace) program like below:

```
#!/usr/bin/env bpftrace

uprobe:./main:"main.probeMe"+42
{
    printf("rax: %d\n", reg("ax"));
}
```

enables probing without the need of recalculating the offset of `main.probeMe` (which is already calculated and stored inside [`offsets` field](https://github.com/cilium/ebpf/blob/951bb28908d23e50fca063a2d51098ca028352bf/link/uprobe.go#L46)), adding `0x2a` and setting `Offset` field to the resulting value:

```go
up, err := ex.Uprobe("main.probeMe", objs.ExampleUprobe, &link.UprobeOptions{RelativeOffset: 0x2a})
if err != nil {
	log.Fatalf("creating uprobe: %s", err)
}
defer up.Close()
```

The PR also improves some of the comments because it's hard to understand at first glance that with the block below, the `Offset` field nullifies the offset that's calculated from the given symbol and the `Offset` is actually an **absolute offset** instead of a relative one:

https://github.com/cilium/ebpf/blob/951bb28908d23e50fca063a2d51098ca028352bf/link/uprobe.go#L255-L262

### DIFF
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -49,7 +49,7 @@ type Executable struct {
 // UprobeOptions defines additional parameters that will be used
 // when loading Uprobes.
 type UprobeOptions struct {
-	// Symbol offset (absolute). Must be provided in case of external symbols (shared libs).
+	// Symbol offset. Must be provided in case of external symbols (shared libs).
 	// If set, overrides the offset eventually parsed from the executable.
 	Offset uint64
 	// The offset relative to given symbol. Useful when tracing an arbitrary point
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -49,9 +49,13 @@ type Executable struct {
 // UprobeOptions defines additional parameters that will be used
 // when loading Uprobes.
 type UprobeOptions struct {
-	// Symbol offset. Must be provided in case of external symbols (shared libs).
+	// Symbol offset (absolute). Must be provided in case of external symbols (shared libs).
 	// If set, overrides the offset eventually parsed from the executable.
 	Offset uint64
+	// The offset relative to given symbol. Useful when tracing an arbitrary point
+	// inside the frame of given symbol and eliminates the need of recalculating
+	// the absolute offset.
+	RelativeOffset uint64
 	// Only set the uprobe on the given process ID. Useful when tracing
 	// shared library calls or programs that have many running instances.
 	PID int
@@ -186,6 +190,8 @@ func (ex *Executable) offset(symbol string) (uint64, error) {
 //
 //  up, err := ex.Uprobe("main", prog, &UprobeOptions{Offset: 0x123})
 //
+// Note: Setting the Offset field in the options supersedes the symbol's offset.
+//
 // Losing the reference to the resulting Link (up) will close the Uprobe
 // and prevent further execution of prog. The Link must be Closed during
 // program shutdown to avoid leaking system resources.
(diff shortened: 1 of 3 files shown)


link/uprobe.go
```diff
@@ -49,9 +49,13 @@ type Executable struct {
 // UprobeOptions defines additional parameters that will be used
 // when loading Uprobes.
 type UprobeOptions struct {
-	// Symbol offset. Must be provided in case of external symbols (shared libs).
+	// Symbol offset (absolute). Must be provided in case of external symbols (shared libs).
```
mmat11, line 52: this shouldn't be the offset from the start (so maybe `absolute` is not the best term). Maybe leave this comment as is and specify in the comment below that `RelativeOffset` is an additional offset relative to the calculated or provided offset?

link/uprobe.go
```diff
@@ -49,9 +49,13 @@ type Executable struct {
...
 	// If set, overrides the offset eventually parsed from the executable.
 	Offset uint64
+	// The offset relative to given symbol. Useful when tracing an arbitrary point
+	// inside the frame of given symbol and eliminates the need of recalculating
```
mmat11, line 56: > eliminates the need of recalculating the absolute offset.

could you clarify this?

link/uprobe.go
```diff
@@ -254,11 +262,12 @@ func (ex *Executable) uprobe(symbol string, prog *ebpf.Program, opts *UprobeOpti
 
 	offset := opts.Offset
 	if offset == 0 {
+		offset = opts.RelativeOffset
```
mmat11, line 265: nit: what about doing `offset: offset+opts.RelativeOffset` in `probeArgs` on line 287? this way the "relative offset" is added both for a specified one and for an internally calculated one

link/uprobe.go
mertyildiran: @mmat11 reverted with https://github.com/mertyildiran/ebpf/commit/ddbe83e447b992cedb652554f987d0a8c83e0f50

link/uprobe.go
mertyildiran: @mmat11 the *"need of recalculating the absolute offset"* is that to achieve `[symbol]+[offset]` without the `RelativeOffset` field, the users need to recalculate the offset of the `[symbol]` by themselves then add the desired relative offset (relative to the symbol). Which means that they need to duplicate the code in [here](https://github.com/cilium/ebpf/blob/951bb28908d23e50fca063a2d51098ca028352bf/link/uprobe.go#L76-L158) and run it before the `Uprobe` call while all of offsets for all of the symbol in the binary is already calculated before by this library with `OpenExecutable` call.

Please suggest the correct comment if the current one is unsuitable.

link/uprobe.go
mertyildiran: @mmat11 done with https://github.com/mertyildiran/ebpf/commit/8a7955ac654c14d0744dc3e48185802c84d94944

link/uprobe.go
mmat11: thanks! the comment looks good to me

mmat11 (approved):

lmb (changes requested):
Hi, can you add a test that covers the feature?

mertyildiran:
@lmb added a test case with https://github.com/mertyildiran/ebpf/commit/89cb6f25c563af9a4a656c2d750b0d1aee1f3dd1 But IMHO, exporting the `Offset` method was a simpler change. Which I have provided as an alternative in the PR description: https://github.com/cilium/ebpf/pull/683 I'm not sure why the `offset` method is preferred as unexported:

https://github.com/cilium/ebpf/blob/0f694846dcf82318fdaf28ce6db73c110caaadc7/link/uprobe.go#L160

PR #683, opened 2022-05-24
Export the `Offset` method of `Executable`

mertyildiran:
Alternative to https://github.com/cilium/ebpf/pull/682

This PR, for a [bpftrace](https://github.com/iovisor/bpftrace) program like below:

```
#!/usr/bin/env bpftrace

uprobe:./main:"main.probeMe"+42
{
    printf("rax: %d\n", reg("ax"));
}
```

enables probing without the need of recalculating the offset of `main.probeMe` (which is already calculated and stored inside [`offsets` field](https://github.com/cilium/ebpf/blob/951bb28908d23e50fca063a2d51098ca028352bf/link/uprobe.go#L46)):

```go
symbol := "main.probeMe"
offset, err := ex.Offset(readSymbol)
if err != nil {
	log.Fatalf("error getting offset: %s", err)
}
offset += 0x2a
up, err := ex.Uprobe(symbol, objs.ExampleUprobe, &link.UprobeOptions{Offset: offset})
if err != nil {
	log.Fatalf("creating uprobe: %s", err)
}
defer up.Close()
```

### DIFF
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -157,7 +157,7 @@ func (ex *Executable) load(f *internal.SafeELFFile) error {
 	return nil
 }
 
-func (ex *Executable) offset(symbol string) (uint64, error) {
+func (ex *Executable) Offset(symbol string) (uint64, error) {
 	if off, ok := ex.offsets[symbol]; ok {
 		// Symbols with location 0 from section undef are shared library calls and
 		// are relocated before the binary is executed. Dynamic linking is not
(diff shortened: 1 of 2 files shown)


mmat11:
Imho https://github.com/cilium/ebpf/pull/682 looks better from the api point of view

mertyildiran:
@mmat11 closing in favor of https://github.com/cilium/ebpf/pull/682

PR #685, opened 2022-05-26
prog: Allow empty data for prog.Test and prog.Run

danobi:
Certain prog types (SK_LOOKUP, RAW_TRACEPOINT and
RAW_TRACEPOINT_WRITABLE) require that data_in is NULL. The previous
check prevented 0 length data_in.

Fix by removing check.

Followup from #647.

### DIFF
--- a/internal/cmd/gentypes/main.go
+++ b/internal/cmd/gentypes/main.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"sort"
@@ -173,12 +174,19 @@ import (
 				replace(enumTypes["LinkType"], "type"),
 				replace(linkID, "id"),
 				name(3, "extra"),
-				replaceUnionWithBytes("extra"),
+				replaceWithBytes("extra"),
 			},
 		},
 		{"FuncInfo", "bpf_func_info", nil},
 		{"LineInfo", "bpf_line_info", nil},
 		{"XdpMd", "xdp_md", nil},
+		{
+			"SkLookup", "bpf_sk_lookup",
+			[]patch{
+				choose(0, "cookie"),
+				replaceWithBytes("remote_ip4", "remote_ip6", "local_ip4", "local_ip6"),
+			},
+		},
 	}
 
 	sort.Slice(structs, func(i, j int) bool {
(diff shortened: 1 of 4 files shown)


lmb (changes requested):
Could you add a test for this? AFAIR SK_LOOKUP programs will refuse to run if given DataIn, so you can use that to make sure this works.

dxuuu:
Ping

prog_test.go
```diff
@@ -136,6 +136,45 @@ func TestProgramRunWithOptions(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	skLookup := sys.BpfSkLookup{
+		Family: 2, /* AF_INET */
+	}
+	opts := RunOptions{
+		Data:    []byte{},
```
ti-mo, line 164: I think this could be removed, the call to `sys.NewSlicePointer(opts.Data)` in testRun() will return a `Pointer{}` whether Data is nil or length 0.

prog_test.go
```diff
@@ -136,6 +136,45 @@ func TestProgramRunWithOptions(t *testing.T) {
...
+		Instructions: ins,
+		License:      "MIT",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer prog.Close()
+
+	skLookup := sys.BpfSkLookup{
+		Family: 2, /* AF_INET */
```
ti-mo, line 161: Feel free to add an entry to `internal/unix/types_*.go`.

ti-mo (commented):
Sorry for the long wait, there has been some PTO in between. Thanks for your patience! Left some minor comments.

internal/cmd/gentypes/main.go
```diff
@@ -179,6 +179,13 @@ import (
 		{"FuncInfo", "bpf_func_info", nil},
 		{"LineInfo", "bpf_line_info", nil},
 		{"XdpMd", "xdp_md", nil},
+		{
+			"BpfSkLookup", "bpf_sk_lookup",
```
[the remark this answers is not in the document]
lmb: It's easier to automate generating names like `BtfInfo` vs capitalising some things.

lmb (approved):

lmb:
Sorry for the long delay, I wasn't aware that you had made changes and then I was busy with life when you pinged me. Thanks for following up :)

PR #686, opened 2022-05-26
prog: Avoid allocating for ctx if ctx not provided

dxuuu:
Avoid allocating an extraneous `bytes.Buffer` if we don't need to.

Followup from #647.


dxuuu:
Whoops wrong account

PR #691, opened 2022-05-29
btf: Add support for loading split BTF

brb:
This PR adds a support for split BTF parsing. The idea behind the split BTF is explained in https://github.com/libbpf/libbpf/commit/2811d54f8b7168875ee87bfabab9f4438f66ecfb. The usage example - https://github.com/cilium/pwru/pull/64.

Updates #705.


btf/types.go
```diff
@@ -790,9 +806,18 @@ func inflateRawTypes(rawTypes []rawType, rawStrings *stringTable) ([]Type, error
...
 	fixup := func(id TypeID, typ *Type) {
-		if id < TypeID(len(types)) {
+		if id < TypeID(len(baseTypes)) {
+			*typ = baseTypes[id]
+			return
+		}
+
+		idx := id
+		if baseTypes != nil {
+			idx = id - typeIDOffset
```
lmb, line 816: Please use `len(baseTypes)` instead of `typeIDOffset`.

btf/types.go
```diff
@@ -1033,11 +1056,11 @@ func inflateRawTypes(rawTypes []rawType, rawStrings *stringTable) ([]Type, error
 
 	for _, fixup := range fixups {
 		i := int(fixup.id)
-		if i >= len(types) {
+		if i >= len(types)+len(baseTypes) {
```
lmb, line 1059: Can you bail out `if i < len(baseTypes)` here? Otherwise the interaction between `fixup()` and the `types[i-len(baseTypes)]` is pretty subtle.

btf/btf.go
```diff
@@ -199,26 +199,39 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
 		return nil, fmt.Errorf("compressed BTF is not supported")
 	}
 
-	return loadRawSpec(btfSection.ReaderAt, file.ByteOrder, sectionSizes, vars)
+	return loadRawSpec(btfSection.ReaderAt, file.ByteOrder, nil, nil, sectionSizes, vars)
 }
 
-func loadRawSpec(btf io.ReaderAt, bo binary.ByteOrder, sectionSizes map[string]uint32, variableOffsets map[variable]uint32) (*Spec, error) {
```
lmb, line 205: `loadRawSpec` is already too messy, I think this breaks the camel's back. I'll try to split it out some to make your life easier.

btf/btf_test.go
```diff
@@ -377,3 +377,34 @@ func TestTypesIterator(t *testing.T) {
...
+	// net/netfilter/xt_nat.c:static int xt_nat_checkentry(const struct xt_tgchk_param *par)
+	typ, err := splitSpec.AnyTypeByName("xt_nat_checkentry")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fnType := typ.(*Func)
+	fnProto := fnType.Type.(*FuncProto)
+
+	// 'int' is defined in the base BTF
+	intType, err := spec.AnyTypeByName("int")
```
lmb, line 401: Needs an assertion that `"int"` is not  in `splitSpec`.

btf/types.go
```diff
@@ -1033,11 +1056,11 @@ func inflateRawTypes(rawTypes []rawType, rawStrings *stringTable) ([]Type, error
...
 			return nil, fmt.Errorf("reference to invalid type id: %d", fixup.id)
 		}
 
-		*fixup.typ = types[i]
+		*fixup.typ = types[i-len(baseTypes)]
 	}
 
 	for _, bitfieldFixup := range bitfieldFixups {
```
lmb, line 1066: This PR doesn't address legacy bitfield fixups at all, which needs another look. Basically, what if a struct member in a kmod points at a type in base which used a legacy bitfield encoding? At this point we've lost the bitfield information from `base`, so there is probably nothing we can do. Except return an error.

btf/btf.go
lmb: See #708

btf/types.go
brb: > At this point we've lost the bitfield information from base, so there is probably nothing we can do

Also, the new bitfield format was introduced in v4.20 (https://github.com/torvalds/linux/commit/9d5f9f701b1891466fb3dbb1806ad97716f95cc3), while the kmod BTF - in v5.9 (https://github.com/torvalds/linux/commit/36e68442d1afd4f720704ee1ea8486331507e834).

> Except return an error.

Added the check.

brb:
Once I get the `/sys/kernel/btf/xt_nat` from @florianl, I am going to adjust my last commit accordingly.

brb:
> Once I get the /sys/kernel/btf/xt_nat from @florianl, I am going to adjust my last commit accordingly.

Done.

brb:
The 5.4 CI build time out seems unrelated to my changes.

btf/btf_test.go
```diff
@@ -377,3 +377,58 @@ func TestTypesIterator(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// net/netfilter/xt_nat.c:static int xt_nat_checkentry(const struct xt_tgchk_param *par)
+	typ, err := splitSpec.AnyTypeByName("xt_nat_checkentry")
+	if err != nil {
+		t.Fatal(err)
+	}
+	typeID := splitSpec.typeIDs[typ]
```
lmb, line 403: Nit: Please use `Spec.TypeID`.

btf/strings.go
```diff
@@ -63,6 +71,13 @@ func splitNull(data []byte, atEOF bool) (advance int, token []byte, err error) {
 	return i + 1, data[:i], nil
 }
 
+func lookupString(offset uint32, rawStringTable, baseStringTable *stringTable) (string, error) {
```
lmb, line 74: This should be folded into `stringTable.Lookup`, see https://github.com/cilium/ebpf/pull/691#discussion_r894622433.

btf/strings.go
```diff
@@ -19,15 +19,23 @@ type sizedReader interface {
 	Size() int64
 }
 
-func readStringTable(r sizedReader) (*stringTable, error) {
+func readStringTable(r sizedReader, baseStrings *stringTable) (*stringTable, error) {
+	// When parsing split BTF's string table, the first entry offset is derived
+	// from the last entry offset of the base BTF.
+	nextStringOffset := uint32(0)
```
lmb, line 25: Nit: `firstStringOffset`.

btf/types.go
```diff
@@ -776,12 +776,23 @@ func (dq *typeDeque) all() []*Type {
...
-	types := make([]Type, 0, len(rawTypes)+1)
-	types = append(types, (*Void)(nil))
+// If baseTypes and baseRawStrings are provided, then the raw types are
+// considered to be of a split BTF (e.g., a kernel module).
+//
+// Returns  a slice of types indexed by TypeID. Since BTF ignores compilation
+// units, multiple types may share the same name. A Type may form a cyclic graph
+// by pointing at itself.
+func inflateRawTypes(rawTypes []rawType, baseTypes types, rawStrings, baseRawStrings *stringTable) ([]Type, error) {
+	types := make([]Type, 0, len(rawTypes))
```
lmb, line 786: cap needs `+1` otherwise we reallocate for vmlinux. Overallocating is OK, underallocating is expensive.

btf/types.go
lmb: The loop here relies on the fact that `fixup` will unconditionally take care of base types, so we can never see an id < len(baseTypes). That isn't easily apparent from the code, adding `if i < len(baseTypes)` could serve as an indication for this. The alternative is a comment, but that is less robust to changes in `fixup`.

btf/btf_test.go
```diff
@@ -377,3 +377,58 @@ func TestTypesIterator(t *testing.T) {
...
 	}
 }
+
+func TestLoadSplitSpecFromReader(t *testing.T) {
+	spec, err := LoadSpecFromReader(readVMLinux(t))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	f, err := os.Open("testdata/xt_nat")
```
lmb, line 387: This is fine for now, but we should replace this with split BTF for `bpf_testmod` from one of the kernels we use on CI. See https://github.com/cilium/ebpf/commit/544f053c951ed3b02a881a4fe349ec7203a99225

lmb (changes requested):

btf/btf_test.go
brb: Are you suggesting instead of having the copy of kmod BTF (e.g., `xt_nat`) to load the `bpf_testmod`, and then to read the split BTF from `/sys/kernel/btf/bpf_testmod`? If yes, then my worry is that it will not be so easy to test the pkg `btf` with just `go test ./btf`.

lmb (approved):
I've amended your commit to add `stringTable.base` which cuts down on the diff.

PR #694, opened 2022-06-01
link: unify KprobeOptions.Offset and UprobeOptions.Offset

lmb:
Both options currently have subtly different definitions. Kprobe.Offset is counted
from the start of the symbol, while Uprobe.Offset counts from the start of the
address space.

Rename UprobeOptions.Offset to Address and RelativeOffset to Offset. Existing
users of UprobeOptions.Offset will get an error when they first upgrade the
library, so make the error message even more explicit.

### DIFF
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -42,20 +42,22 @@ var (
 type Executable struct {
 	// Path of the executable on the filesystem.
 	path string
-	// Parsed ELF symbols and dynamic symbols offsets.
-	offsets map[string]uint64
+	// Parsed ELF and dynamic symbols' addresses.
+	addresses map[string]uint64
 }
 
 // UprobeOptions defines additional parameters that will be used
 // when loading Uprobes.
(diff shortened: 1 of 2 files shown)


link/uprobe.go
```diff
@@ -176,19 +178,24 @@ func (ex *Executable) offset(symbol string, opts *UprobeOptions) (uint64, error)
...
 	} else {
 		return 0, fmt.Errorf("symbol %s: %w", symbol, ErrNoSymbol)
 	}
 
-	// Always add the relative offset regardless where we got the absolute
-	// offset from.
-	offset += opts.RelativeOffset
-	return offset, nil
+	// Always add the relative offset regardless where we got the address from.
+	address += opts.Offset
```
mmat11, line 197: shouldn't we do this only in case `opts.Address > 0` to retain old behavior?

mmat11 (approved):
lgtm

PR #697, opened 2022-06-03
map: suggest incorrect MaxEntries on EINVAL

markpash:
If a map creation fails and the max entries provided by the spec is
zero, it's likely that it's because of that zero value.
Help the user by suggesting that MaxEntries may be incorrectly set to
zero.

Signed-off-by: Mark Pashmfouroush <mark@isovalent.com>


lmb (approved):
Nice, thanks!

PR #699, opened 2022-06-04
link: close tracefs event if perf event cannot be opened

nashuiliang:
When getTraceEventID() or openTracepointPerfEvent() returns an error,
tracefs event (e.g. /sys/kernel/debug/tracing/events/<group>), which has
been created by createTraceFSProbeEvent(), is not cleaned up.

Signed-off-by: Chuang Wang <nashuiliang@gmail.com>

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -350,7 +350,7 @@ func tracefsKprobe(args probeArgs) (*perfEvent, error) {
 // Path and offset are only set in the case of uprobe(s) and are used to set
 // the executable/library path on the filesystem and the offset where the probe is inserted.
 // A perf event is then opened on the newly-created trace event and returned to the caller.
-func tracefsProbe(typ probeType, args probeArgs) (*perfEvent, error) {
+func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
 	// Generate a random string for each trace event we attempt to create.
 	// This value is used as the 'group' token in tracefs to allow creating
 	// multiple kprobe trace events with the same name.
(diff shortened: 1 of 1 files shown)


mmat11 (approved):
I remember pointing this out in the initial Kprobe implementation, here's the discussion: https://github.com/cilium/ebpf/pull/265#discussion_r602535385. I have no strong opinion but prefer cleaning it up /cc @ti-mo

nashuiliang:
Thanks, I read the discussion: https://github.com/cilium/ebpf/pull/265#discussion_r602535385.

I should briefly describe my scene：
> When **livepatch** and **kprobe** are using the same kernel function X, tracefs event (e.g. /sys/kernel/debug/tracing/events/), which has been created by createTraceFSProbeEvent(), cannot be enabled by syscall `perf_event_open`

Therefore, I should add `closeTraceFSProbeEvent()` on failed `openTracepointPerfEvent()`.
Hmm, I saw `getTraceEventID()` as well, and added it by the way, although I thought it might not return an error.

nashuiliang:
@ti-mo Wow, cool!

The test method is more complicated. It can be triggered manually, and it is difficult to write test code at the same time.

lmb:
> I should briefly describe my scene:
> > When **livepatch** and **kprobe** are using the same kernel function X, tracefs event (e.g. /sys/kernel/debug/tracing/events/), which has been created by createTraceFSProbeEvent(), cannot be enabled by syscall `perf_event_open`.

Interesting! Out of curiosity, which errno do you get from perf_event_open?

lmb (approved):

nashuiliang:
Error: `opening tracepoint perf event: device or resource busy`.
Errno: `-EBUSY`

PR #702, opened 2022-06-10
run-tests: fix shebang to find bash more reliably

markpash:
Not all systems have bash located at /bin/bash.
A better way to run bash with shebang is with:
`#!/usr/bin/env bash`

Signed-off-by: Mark Pashmfouroush <mark@isovalent.com>

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Test the current package under a different kernel.
 # Requires virtme and qemu to be installed.
 # Examples:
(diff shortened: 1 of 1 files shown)


rgo3:
CI seems to fail because the ca-certs we download in semaphore don't exist in the [index](http://security.ubuntu.com/ubuntu/pool/main/c/ca-certificates/) anymore.

lmb:
Let's see whether #703 fixes it.

lmb (approved):

lmb:
Thanks!

Issue #645, opened 2022-04-27
Add some XDP examples


diegobes:
There are several examples of different ebpf program types but none for XDP.
Can you please add an example of how to use link/program to attach, deatach, pin, unpin xdp programs and maps ?

Thanks

Cyc1ing1313:
agree

ti-mo:
We'll add an example for attaching XDP progs using the `link` package. We could add a fallback using netlink later on if the kernel doesn't support bpf_link, but let's start with a simple example.

@diegobes Pinning is already demonstrated in other examples, is there something specific that needs to be highlighted?

lmb:
We have an XDP link example thanks to @wedaly now.

PR #706, opened 2022-06-10
examples: add example for XDP

wedaly:
Add an example for attaching an eBPF program to a network
interface with XDP. The example program parses the IPv4
source address from packets (if available) and records
packet counts by IP address in an LRU hash map.

Issue: https://github.com/cilium/ebpf/issues/645


examples/xdp/main.go
```diff
@@ -0,0 +1,83 @@
...
+// +build linux
+
+// This program demonstrates attaching an eBPF program to a network interface
+// with XDP (eXpress Data Path). The program parses the IPv4 source address
+// from packets and writes the packet count by IP to an LRU hash map.
+// The userspace program (Go code in this file) prints the contents
+// of the map to stdout every second.
+// It is possible to modify the XDP program to drop or redirect packets
+// as well -- give it a try!
+package main
```
florianl, line 11: This example makes use of techniques only available in Kernel 5.7 and newer. I think it would be nice to make the reader/user of this example aware of this in the comment.

wedaly: good call, updated!

wedaly:
Working on fixing this error in the build:
```
In file included from /ebpf/examples/xdp/xdp.c:5:
In file included from /usr/include/linux/if_ether.h:25:
/usr/include/linux/types.h:5:10: fatal error: 'asm/types.h' file not found
#include <asm/types.h>
         ^~~~~~~~~~~~~
```

wedaly:
Copied the types I needed into examples/common.h. Maybe I should pull in the header changes from https://github.com/cilium/ebpf/pull/711 (or wait until after that PR merges?) to avoid conflicts?

examples/xdp/main.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+func formatMapContents(m *ebpf.Map) (string, error) {
+	var sb strings.Builder
+	var key, val uint32
+	iter := m.Iterate()
+	for iter.Next(&key, &val) {
+		sourceIP := net.IPv4(byte(key>>24), byte(key>>16), byte(key>>8), byte(key))
+		packetCount := val
```
markpash, line 80:
```suggestion
	var (
		sb   strings.Builder
		key  []byte
		val  uint32
		iter = m.Iterate()
	)

	for iter.Next(&key, &val) {
		sourceIP := net.IP(key)
		packetCount := val
```

markpash: Since you do the byte swap of the IP in BPF, we can just get `key` in the form of `[]byte` and coerce it to a Go `net.IP`. (This should be even easier with go1.18 and `netip.Addr` since it implements `encoding.BinaryUnmarshaler`!)

It would also make it super easy if someone wanted to adapt this to support IPv6!

examples/xdp/xdp.c
```diff
@@ -0,0 +1,74 @@
...
+		// The protocol is not IPv4, so we can't parse an IPv4 source address.
+		return 0;
+	}
+
+	// Then parse the IP header.
+	struct iphdr *ip = pos;
+	hdrsize          = sizeof(*ip);
+	if (pos + hdrsize > data_end) {
+		return 0;
+	}
```
markpash, line 45:
```suggestion
	void *data_end = (void *)(long)ctx->data_end;
	void *data     = (void *)(long)ctx->data;

	// First, parse the ethernet header.
	struct ethhdr *eth = data;
	if ((void *)(eth + 1) > data_end) {
		return 0;
	}

	if (eth->h_proto != bpf_htons(ETH_P_IP)) {
		// The protocol is not IPv4, so we can't parse an IPv4 source address.
		return 0;
	}

	// Then parse the IP header.
	struct iphdr *ip = (void *)(eth + 1);
	if ((void *)(ip + 1) > data_end) {
		return 0;
	}
```

examples/xdp/xdp.c
```diff
@@ -0,0 +1,74 @@
...
+
+	// Return the source IP address in host byte order.
+	*ip_src_addr = (__u32)bpf_ntohl(ip->saddr);
+	return 1;
+}
+
+SEC("xdp_prog")
```
markpash, line 52:
```suggestion
SEC("xdp")
```
Not sure how big of an issue it is, but libbpf [has decided](https://github.com/libbpf/libbpf/wiki/Libbpf:-the-road-to-v1.0#stricter-and-more-uniform-bpf-program-section-name-sec-handling) this should just be `xdp`.

markpash (changes requested):
Nice example! Just a few non-critical suggestions.

examples/xdp/xdp.c
```diff
@@ -0,0 +1,74 @@
...
+
+const int MAX_MAP_ENTRIES = 16;
+
+/* Define an LRU hash map for storing packet count by source IPv4 address */
+struct bpf_map_def SEC("maps") xdp_stats_map = {
+	.type        = BPF_MAP_TYPE_LRU_HASH,
+	.key_size    = sizeof(__u32), // source IPv4 address
+	.value_size  = sizeof(__u32), // packet count
+	.max_entries = MAX_MAP_ENTRIES,
+};
```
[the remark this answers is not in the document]
wedaly: thanks! updated to use BTF-defined maps

examples/xdp/xdp.c
wedaly: updated!

markpash (approved):

lmb (approved):

lmb:
Thanks!

examples/xdp/main.go
dadav: I'm having a hard time getting the ipv6 network byte order thingy working, would be awesome if someone could share an example someday :)

ultperf: anyone care to share the ipv6 example?

PR #708, opened 2022-06-11
btf: simplify loadRawSpec

lmb:
We currently call fixupDatasec from within loadRawSpec, but only
the call site in loadSpecFromELF actually requires the call to be
present. Move the call into loadSpecFromELF, which allows dropping
two arguments from loadRawSpec.


brb (approved):
LGTM, thanks.

Issue #705, opened 2022-06-10
Support for kernel module BTF


lmb:
Newer kernels ship with separate BTF for kernel modules (kmod):

```
$ ls /sys/kernel/btf/ | head -n5
aes_ce_blk
aes_ce_cipher
aes_neon_blk
aes_neon_bs
async_memcpy
```

The library is currently oblivious to their existence. This means that kmod BTF is not taken into account for CO-RE relocations and it's not possible to attach fentry / fexit (others?) programs to kernel modules. Both should be supported without the user needing to take additional steps in the happy case.

- [x] LoadSplitSpecFromReader
- [x] CO-RE / Merge BTF
- [x] Load kmod BTF
- [x] fentry / fexit

# Proposed API

```
// LoadSplitSpecFromReader loads split BTF from a reader.
//
// Types from base are used to resolve references in the split BTF.
// The returned Spec only contains types from the split BTF, not from the base.
func LoadSplitSpecFromReader(r io.ReaderAt, base *Spec) (*Spec, error)

// Merge the types from other into the current Spec.
//
// Merged types don't have a canonical type ID, and therefore TypeID() will
// return an error for them.
func (*Spec) Merge(other *Spec) error

// Spec retrieves the types associated with the Handle from the kernel.
//
// base is used to decode split BTF and may be nil.
func (h *Handle) Spec(base *Spec) (*Spec, error) // BREAKING CHANGE
```

# CO-RE

CO-RE is currently done entirely in the library, by matching types used by the BPF program to types found in the kernel.

To extend CO-RE to kmods, we should merge the set of `vmlinux` and all `kmod` types and search this superset. Even though a kprobe is attached to a particular kmod's function, it
might still want to pull types from other modules, for example to cast void* to
appropriate types.
I believe this is what [libbpf does](https://github.com/libbpf/libbpf/blob/e93b1010f30bce09ca68db3f6db61fc100c1a41c/src/libbpf.c#L5649-L5657).

Creating the superset of types will be done by `Spec.Merge()`. Merged types won't
have a canonical type ID, so `Spec.TypeID()` will return a well known error.

**Problem**: `bpf_core_type_id_kernel` is supposed to return the ID of a type in the kernel.
According to the documentation we could probably return `0` and be "compatible":

```
/*
 * Convenience macro to get BTF type ID of a target kernel's type that matches
 * specified local type.
 * Returns:
 *    - valid 32-bit unsigned type ID in kernel BTF;
[...]

PR #709, opened 2022-06-12
CI: load bpf_testmod in VM

lmb:
Load the bpf_testmod into the kernel when booting the test VM. This
allows getting rid of the workaround to load vmlinux at BTF ID 1,
since loading a module also loads vmlinux BTF.

Later on we can use this to test split BTF / kmod support.

Updates #705

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -83,6 +83,10 @@ elif [[ "${1:-}" = "--exec-test" ]]; then
     export KERNEL_SELFTESTS="/run/input/bpf"
   fi
 
+  if [[ -f "/run/input/bpf/bpf_testmod/bpf_testmod.ko" ]]; then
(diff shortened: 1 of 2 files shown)


run-tests.sh
```diff
@@ -83,6 +83,10 @@ elif [[ "${1:-}" = "--exec-test" ]]; then
     export KERNEL_SELFTESTS="/run/input/bpf"
   fi
 
+  if [[ -f "/run/input/bpf/bpf_testmod/bpf_testmod.ko" ]]; then
+    insmod "/run/input/bpf/bpf_testmod/bpf_testmod.ko"
```
lmb, line 87: Shipped in selftests.tgz as of [7744b00ee32e31f6c43d79911d27d704b45de83a](https://github.com/cilium/ci-kernels/commit/7744b00ee32e31f6c43d79911d27d704b45de83a)

PR #710, opened 2022-06-12
Remove 4.4 from CI

lmb:


lmb:
Thanks, added names back to the testcases.

PR #711, opened 2022-06-13
examples: add a XDP firewall example using LPM trie

lobshunter:
Signed-off-by: lob <pengyu@pingcap.com>

This PR adds _another_ XDP example. This program is a XDP version of firewall.
It uses `BPF_MAP_TYPE_LPM_TRIE` to match IP of incoming packets against the block CIDR. IPv4 packets with source address belonging to block CIDR will be dropped.

Related issue: #645.


lmb:
Thanks for writing up this example, however I think that its scope is too big for this repository. Adding the LPM map makes for a nice demo but it doesn't make XDP usage clearer than the existing example.

What would be nice is to add a netlink fallback to the example from #706.

PR #714, opened 2022-06-16
map: Do not chec maxEntries for PerfEventArray map

olsajiri:
When I use perf event map in MapReplacements for ebpf.CollectionOptions,
I'm getting error for maxEntries check.

The perf event map is created with maxEntries initialized to number
of cpus, so we can't check maxEntries before the map is created.

I'm not sure we can postpone the check, but we can surely skip it
for perf event map.

### DIFF
--- a/map.go
+++ b/map.go
@@ -170,7 +170,8 @@ func (ms *MapSpec) checkCompatibility(m *Map) error {
 	case m.valueSize != ms.ValueSize:
 		return fmt.Errorf("expected value size %v, got %v: %w", ms.ValueSize, m.valueSize, ErrMapIncompatible)
 
-	case m.maxEntries != ms.MaxEntries:
+	case !(ms.Type == PerfEventArray && ms.MaxEntries == 0) &&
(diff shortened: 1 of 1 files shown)


lmb:
Ah, good point! We already have a bit of special casing for PerfEventArray:

https://github.com/cilium/ebpf/blob/be4d5714f9da20d394df98a957f7c6f6380f8ab3/map.go#L381-L387

What if you only skip the check if `ms.Type == PerfEventArray && ms.MaxEntries == 0`? That way it's in line with the other special case we have.

olsajiri:
> What if you only skip the check if `ms.Type == PerfEventArray && ms.MaxEntries == 0`? That way it's in line with the other special case we have.

sounds good, will change, thanks

olsajiri:
@lmb any idea what's wrong with the CI? looks like I can't check results, thanks

lmb:
Sometimes it gets stuck, I think due to the VM crashing. I've triggered a rerun.

Issue #701, opened 2022-06-06
LoadAndAssign() should explicitly close all resources it doesn't assign


ti-mo:
I got nerdsniped on Slack and replaced the `sys.FD` finalizer with:
```go
func (fd *FD) finalize() {
	if fd.raw != -1 {
		panic(fmt.Sprintf("forgot to close fd %d", fd.raw))
	}
}
```

This broke a few tests, some where Close() was omitted, mostly around the use of `LoadAndAssign`.

`collectionLoader.finalize()` should only remove the maps and progs that were successfully assigned at least once. It shouldn't zero out `cl.maps` and `cl.programs`, because that relies on finalizers running, which might just never happen.

lmb:
Interesting! Maybe we could add some code that surfaces these bugs when running tests?

Something like:
- allow changing the behaviour to panic at runtime. This could be `var internal.FinalizerShouldPanic bool`
- set the flag from tests and invoke runtime.GC at the end of TestMain.

ti-mo:
I've been experimenting with a few approaches, but it does feel like we need some sort of global to toggle this behaviour. I was also thinking I'd probably want to run actual applications with this enabled in the future, since there will be a lot of upcoming refactoring work in Cilium where this would be incredibly handy. (it currently stuffs maps (created by an old internal bpf/ package, not the lib) containing live fds into package-global vars, and their lifecycle is kind of undefined)

I have a branch with all 'internal' leaks ironed out: https://github.com/ti-mo/ebpf/tree/tb/leaking-fds (grab some eye bleach)

For tests, panicking is okay, but also very annoying to debug, since the panic occurs within the runtime and there's no stack info available. In my hack branch, I've added two string fields, one `metadata` for annotating the fd with a name from within `newProgramWithOptions` etc., and another string to stuff the value of `runtime.Stack()` into from `NewFD()` so when the FD is finalized, there's info available about where it was created.

For anything other than tests, `panic()` is not acceptable, so I was thinking of some kind of notification mechanism that callers can subscribe to in order to receive updates when a GC closes an FD.

There are a few shortcomings to this approach:

- Stack/metadata collection would only start when there is at least one subscriber, but that might result in messages without any metadata if fds were created before the first subscriber. Probably unavoidable and acceptable.
[...]

PR #724, opened 2022-07-01
collection: close all unassigned resources before returning from LoadAndAssign

ti-mo:
```
    collection: close all unassigned resources before returning from LoadAndAssign

    Before, loader.finalize() would nil out cl.maps and cl.progs before returning
    from LoadAndAssign, leaving the closing of fds up to the garbage collector.
    This is non-deterministic and potentially causes rlimit pressure when loading
    many objects in quick succession. In case the GC is disabled, this leaks fds.

    .rodata is one such common case that often gets lazy-loaded, but rarely
    gets assigned due to the availability of RewriteConstants. This Map's fd would
    then stick around until the Map gets collected.

    This commit replaces cl.finalize() with inline logic in LoadAndAssign and
    NewCollectionWithOptions and renames loader.cleanup() to .close().
```

Fixes: https://github.com/cilium/ebpf/issues/701

cc @alban @mauriciovasquezbernal

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -264,14 +264,17 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 	if err != nil {
 		return err
 	}
-	defer loader.cleanup()
+	defer loader.close()
 
 	// Support assigning Programs and Maps, lazy-loading the required objects.
 	assignedMaps := make(map[string]bool)
+	assignedProgs := make(map[string]bool)
+
 	getValue := func(typ reflect.Type, name string) (interface{}, error) {
 		switch typ {
 
 		case reflect.TypeOf((*Program)(nil)):
+			assignedProgs[name] = true
 			return loader.loadProgram(name)
 
 		case reflect.TypeOf((*Map)(nil)):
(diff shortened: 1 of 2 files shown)


alban:
I confirm that the fd leak is fixed when using this branch.

Tested with go.mod:
```
// https://github.com/cilium/ebpf/pull/724
replace github.com/cilium/ebpf v0.9.0 => github.com/ti-mo/ebpf v0.5.1-0.20220701135211-fe77f4b1476b
```

alban (approved):
I am not a maintainer, but the patch looks good to me.

collection.go
```diff
@@ -259,19 +259,22 @@ func (cs *CollectionSpec) Assign(to interface{}) error {
...
 	loader, err := newCollectionLoader(cs, opts)
 	if err != nil {
 		return err
 	}
-	defer loader.cleanup()
 
-	// Support assigning Programs and Maps, lazy-loading the required objects.
 	assignedMaps := make(map[string]bool)
+	assignedProgs := make(map[string]bool)
+	defer loader.close(assignedMaps, assignedProgs, err)
```
brycekahle, line 270: I don't know if this intentional, but this will fail to capture any changes to `err` after this point. See https://go.dev/play/p/BzoyZysNcyT

ti-mo: Good catch, we need to close over `err` here to make sure it captures the return value. Will have to add a test for this.

PR #728, opened 2022-07-05
Tidy up btf.Handle related APIs

lmb:
btf: lazily decode types in Handle.Spec

    We currently decode BTF when calling NewHandleFromID. This can be very
    expensive for large BTF blobs like vmlinux, even though not all users
    will want to inspect decoded BTF in the first place.

    Move the decoding into Handle.Spec, which is a breaking change.

    Updates #705

program: add getter for btf.Handle

    Given a loaded program, it takes four calls to retrieve parsed BTF:

    1. Program.Info()
    2. ProgramInfo.BTFID()
    3. NewHandleFromID()
    4. btf.Handle.Spec()

    This chain of calls has a subtle requirement: the loaded program
    must be kept alive for steps 1-3, since otherwise the BTF ID might
    be deallocated or reused by the kernel. Program.Handle encapsulates
    these steps while keeping program alive.

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -734,6 +734,8 @@ func NewHandle(spec *Spec) (*Handle, error) {
 
 // NewHandleFromID returns the BTF handle for a given id.
 //
+// Prefer calling [ebpf.Program.Handle] or [ebpf.Map.Handle] if possible.
+//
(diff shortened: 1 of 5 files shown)


lmb:
Sorry, this isn't ready yet!

PR #729, opened 2022-07-05
btf: fix NewHandleFromID for kernel modules

lmb:
btf: add split kmod BTF test

    Use HandleIterator to find vmlinux and the first loaded module, and
    attempt to parse the split BTF.

    This requires exporting HandleInfo since we use external tests.

btf: add HandleIterator

    Add an iterator for BTF objects / handles. The API is the same as MapIterator.

    Used to test loading kmod split BTF from a BTF ID in a subsequent commit.

btf: rename info.go to handle.go

btf: fix NewHandleFromID for kernel modules

    Kernel modules only expose split BTF when queried via BPF_OBJ_GET_INFO_BY_FD. This
    causes NewHandleFromID to fail for kmod BTF, since the function attempts to inflate
    types.

    Avoid inflating types in NewHandleFromID and change Handle.Spec to accept a base
    Spec to allow decoding types from modules.

    This is a breaking change.

    Updates #705


btf/handle.go
```diff
@@ -0,0 +1,60 @@
...
+	return &HandleIterator{0, nil}
+}
+
+// Next retrieves a handle for the next BTF blob.
+//
+// [Handle.Close] is called if *handle is non-nil to avoid leaking fds.
+//
+// Returns true if another BTF blob was found. Call [HandleIterator.Err] after
+// the function returns false.
+func (it *HandleIterator) Next(handle **Handle) bool {
```
lmb, line 27: Maybe this should just be `Next()`, with handle exposed via `HandleIterator.Handle` or similar? See https://pkg.go.dev/github.com/cilium/ebpf/btf#TypesIterator

lmb: Turns out that this makes it harder to properly clean up the Handle.

PR #731, opened 2022-07-07
btf: use ci-kernels to source kernel BTF

lmb:
Add a make target generate-btf which recreates checked-in BTF from one
of the canonical CI kernels we test against. Turns out that the currently
checked in BTF is from 5.19, which is still in release candidate stage.

At the same time replace the xt_nat BTF with btf_testmod BTF, since we
already ship that as part of selftests.tgz.


lmb:
Seems like our checked in vmlinux BTF is > 5.15: https://ebpf.semaphoreci.com/jobs/b80a9fe9-1746-41e3-a435-79f3dcbed26f#L490 I'll take a look later.

PR #732, opened 2022-07-07
sys: add fd leak tracing instrumentation

ti-mo:
In an effort to improve the library's fd hygiene, this patch adds optional
metadata to sys.FD to hold a name and a stack trace of where the fd was
created, as well as a simple subscription mechanism for packages to
get notified when a live fd gets garbage collected due to losing its reference.

Metadata is only collected when there are active tracers, so they must be
registered before creating the first sys.FD.

This has shaken out a few minor bugs so far. Most of them from the test suite,
but also a few where we forgot to close resources that were no longer needed
before returning from LoadAndAssign and some other internal APIs

All test packages now also include a TestMain that sets up leak tracing and
starts a GC round before finishing the test, so we can catch these kinds of
errors more easily in the future.

The idea is to export the sys package over time, which would make this
instrumentation available to all, after it has been given some time to mature.

### DIFF
--- a/internal/sys/fd.go
+++ b/internal/sys/fd.go
@@ -17,9 +17,31 @@ type FD struct {
 }
 
 func newFD(value int) *FD {
+	if onLeakFD != nil {
+		fds.Store(value, callersFrames())
+	}
+
 	fd := &FD{value}
-	runtime.SetFinalizer(fd, (*FD).Close)
+	runtime.SetFinalizer(fd, (*FD).finalize)
 	return fd
+}
+
+// finalize is set as the FD's runtime finalizer and
+// sends a leak trace before calling FD.Close().
+func (fd *FD) finalize() {
+	if fd.raw < 0 {
+		return
+	}
+
+	// Invoke the fd leak callback. Calls LoadAndDelete to guarantee the callback
+	// is invoked at most once for one sys.FD allocation, runtime.Frames can only
+	// be unwound once.
+	f, ok := fds.LoadAndDelete(fd.Int())
+	if ok && onLeakFD != nil {
+		onLeakFD(f.(*runtime.Frames))
+	}
+
+	_ = fd.Close()
 }
 
 // NewFD wraps a raw fd with a finalizer.
@@ -66,6 +88,8 @@ func (fd *FD) Close() error {
 
 	value := int(fd.raw)
 	fd.raw = -1
+
+	fds.Delete(value)
 
 	fd.Forget()
 	return unix.Close(value)
--- a/internal/sys/fd_trace.go
+++ b/internal/sys/fd_trace.go
@@ -1,0 +1,78 @@
+package sys
+
+import (
+	"runtime"
+	"sync"
+)
+
+// OnLeakFD controls tracing [FD] lifetime to detect resources that are not
+// closed by Close().
+//
+// If fn is not nil, tracing is enabled for all FDs created going forward. fn is
+// invoked for all FDs that are closed by the garbage collector instead of an
+// explicit Close() by a caller. Calling OnLeakFD twice with a non-nil fn
+// (without disabling tracing in the meantime) will cause a panic.
+//
+// If fn is nil, tracing will be disabled. Any FDs that have not been closed are
+// considered to be leaked, fn will be invoked for them, and the process will be
+// terminated.
+//
+// fn will be invoked at most once for every unique sys.FD allocation since a
+// runtime.Frames can only be unwound once.
+func OnLeakFD(fn func(*runtime.Frames)) {
+	// Enable leak tracing if new fn is provided.
+	if fn != nil {
+		if onLeakFD != nil {
+			panic("OnLeakFD called twice with non-nil fn")
+		}
+
+		onLeakFD = fn
+		return
+	}
+
+	// fn is nil past this point.
+
+	if onLeakFD == nil {
+		return
+	}
+
+	// Call onLeakFD for all open fds.
+	if fs := flushFrames(); len(fs) != 0 {
+		for _, f := range fs {
+			onLeakFD(f)
+		}
+	}
+
+	onLeakFD = nil
+}
+
+var onLeakFD func(*runtime.Frames)
+
+// fds is a registry of all file descriptors wrapped into sys.fds that were
+// created while an fd tracer was active.
+var fds sync.Map // map[int]*runtime.Frames
+
+// flushFrames removes all elements from fds and returns them as a slice. This
+// deals with the fact that a runtime.Frames can only be unwound once using
+// Next().
+func flushFrames() []*runtime.Frames {
+	var frames []*runtime.Frames
+	fds.Range(func(key, value any) bool {
+		frames = append(frames, value.(*runtime.Frames))
+		fds.Delete(key)
+		return true
+	})
+	return frames
+}
+
+func callersFrames() *runtime.Frames {
+	c := make([]uintptr, 32)
+
+	// Skip runtime.Callers and this function.
+	i := runtime.Callers(2, c)
+	if i == 0 {
+		return nil
+	}
+
+	return runtime.CallersFrames(c)
+}
--- a/internal/testutils/fdtrace/fd.go
+++ b/internal/testutils/fdtrace/fd.go
@@ -1,0 +1,47 @@
+package fdtrace
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"runtime"
+	"testing"
+
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+// TestMain runs m with sys.FD leak tracing enabled.
+func TestMain(m *testing.M) {
+	// fn can either be invoked asynchronously by the gc or during disabling of
+	// the leak tracer below. Don't terminate the program immediately, instead
+	// capture a boolean that will be used to set the exit code. This avoids races
+	// and gives all events the chance to be written to stderr.
+	var leak bool
+	sys.OnLeakFD(func(fs *runtime.Frames) {
+		fmt.Fprintln(os.Stderr, "leaked fd created at:")
+		fmt.Fprintln(os.Stderr, formatFrames(fs))
+		leak = true
+	})
+
+	ret := m.Run()
+
+	sys.OnLeakFD(nil)
+
+	if leak {
+		ret = 99
+	}
+
+	os.Exit(ret)
+}
+
+func formatFrames(fs *runtime.Frames) string {
+	var b bytes.Buffer
+	for {
+		f, more := fs.Next()
+		b.WriteString(fmt.Sprintf("\t%s+%#x\n\t\t%s:%d\n", f.Function, f.PC-f.Entry, f.File, f.Line))
+		if !more {
+			break
+		}
+	}
+	return b.String()
+}
--- a/link/cgroup.go
+++ b/link/cgroup.go
@@ -41,6 +41,13 @@ func AttachCgroup(opts CgroupOptions) (Link, error) {
 
 	var cg Link
 	cg, err = newLinkCgroup(cgroup, opts.Attach, clone)
+	if err == nil {
+		cgroup.Close()
+		clone.Close()
+		return cg, nil
+	}
+
+	// cgroup and clone are retained by progAttachCgroup.
 	if errors.Is(err, ErrNotSupported) {
 		cg, err = newProgAttachCgroup(cgroup, opts.Attach, clone, flagAllowMulti)
 	}
@@ -67,6 +74,8 @@ var _ Link = (*progAttachCgroup)(nil)
 
 func (cg *progAttachCgroup) isLink() {}
 
+// newProgAttachCgroup attaches prog to cgroup using BPF_PROG_ATTACH.
+// cgroup and prog are retained by [progAttachCgroup].
 func newProgAttachCgroup(cgroup *os.File, attach ebpf.AttachType, prog *ebpf.Program, flags cgroupAttachFlags) (*progAttachCgroup, error) {
 	if flags&flagAllowMulti > 0 {
 		if err := haveProgAttachReplace(); err != nil {
@@ -151,6 +160,7 @@ type linkCgroup struct {
 
 var _ Link = (*linkCgroup)(nil)
 
+// newLinkCgroup attaches prog to cgroup using BPF_LINK_CREATE.
 func newLinkCgroup(cgroup *os.File, attach ebpf.AttachType, prog *ebpf.Program) (*linkCgroup, error) {
 	link, err := AttachRawLink(RawLinkOptions{
 		Target:  int(cgroup.Fd()),
--- a/map_test.go
+++ b/map_test.go
@@ -31,6 +31,21 @@ var (
 	}
 )
 
+// newHash returns a new Map of type Hash. Cleanup is handled automatically.
+func newHash(t *testing.T) *Map {
+	hash, err := NewMap(&MapSpec{
+		Type:       Hash,
+		KeySize:    5,
+		ValueSize:  4,
+		MaxEntries: 10,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { hash.Close() })
+	return hash
+}
+
 func TestMap(t *testing.T) {
 	m := createArray(t)
 	defer m.Close()
@@ -1182,8 +1197,7 @@ func TestMapGuessNonExistentKey(t *testing.T) {
 }
 
 func TestNotExist(t *testing.T) {
-	hash := createHash()
-	defer hash.Close()
+	hash := newHash(t)
 
 	var tmp uint32
 	err := hash.Lookup("hello", &tmp)
@@ -1214,8 +1228,7 @@ func TestNotExist(t *testing.T) {
 }
 
 func TestExist(t *testing.T) {
-	hash := createHash()
-	defer hash.Close()
+	hash := newHash(t)
 
 	if err := hash.Put("hello", uint32(21)); err != nil {
 		t.Errorf("Failed to put key/value pair into hash: %v", err)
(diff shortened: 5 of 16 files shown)


lmb (commented):
Assumptions:
- The overhead of leak tracking must be minimal when not enabled, since sys.FD is a foundational building block.
- Some users will use leak tracking in their testsuite, see `TestMainWithTracing`.
- Some users will use leak tracking in a "debug build" of their software.
- Some users may not just be interested in leaking fds, but also having a trace of where FDs are closed.

Observations:
- Storing `*metadata` in `FD` increases the size even if leak tracking is disabled.
- `SetName(string)` allocates even when leak tracking is not enabled due to calling `fmt.Sprintf`.
- `LeakTracers` assumes that there may be multiple tracers, which is overkill for the "leak tracing in the testsuite" use case.

My suggestion would be to split your proposal into two parts.

First, a set of callbacks exposed by `sys`.

```Go
type Tracer interface {
    // Create is called whenever a new FD is allocated.
    Create(*FD)

    // Remove is called whenever FD.Close is called.
    Remove(*FD)

    // Finalize is called whenever a finalizer is run on an unclosed FD.
    // It is not invoked on closed FDs.
    Finalize(*FD)
}

// ActiveTracer is called during the lifecycle of an FD if it is non-nil.
//
// Changing ActiveTracer must happen when no FDs are being created or destroyed to
// avoid race conditions.
var ActiveTracer Tracer
```

There is one problem here that I'm not 100% sure how to solve: in this design, per-FD metadata would have to be stored in a map of some sort. The natural way to do this would be `map[*FD]metadata`, but that of course keeps a reference to FD alive and therefore prevents finalizers from running. I think there are two solutions:

1. Use `map[uintptr]metadata` and import https://pkg.go.dev/go4.org/unsafe/assume-no-moving-gc. If go ever gains a moving gc we'd have to break compatibility, or hope that weak references are a thing at that point. Since this is internal to the tracer implementation we could accept this.
2. Use the approach outlined in https://groups.google.com/g/golang-nuts/c/1ItNOOj8yW8/m/9xBlzy6DNKYJ I think it boils down to introducing another type, and changing the tracer to refer to the new type:

```Go
// No finalizer.
type TraceFD struct {
    raw int
}

// Has a finalizer.
type FD struct {
    *TraceFD
}

type Tracer interface {
    Create(*TraceFD)
    ...
}
```

An open question here is what the layout of FD looks like in this case. Is there really an extra pointer indirection, or is the go compiler smart enough to remove it?

I'd probably go for the `uintptr` route at this point.

Second, an implementation of `Tracer` for the in-library use case of checking that we don't have leaks. This can be simpler than the current implementation which supports multiple listeners, changing listeners dynamically, etc.

One problem I don't know how to solve is that unconditionally enabling leak tracing in TestMain will affect benchmarks. Maybe a function in testutils that sets `ActiveTracer` to nil before the benchmark and restores the old value afterwards? Kind of annoying that we'd have to add that to every benchmark though. Maybe we just live with it?

internal/sys/fd.go
```diff
@@ -13,15 +14,33 @@ import (
...
 }
 
 func newFD(value int) *FD {
-	fd := &FD{value}
-	runtime.SetFinalizer(fd, (*FD).Close)
+	// Annotate the fd if there are active fd leak tracers.
+	var meta *metadata
+	if HaveLeakTracers() {
+		meta = &metadata{
+			stack: debug.Stack(),
```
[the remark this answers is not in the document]
ti-mo: Thanks, I found it a bit strange not being able to take a more compact representation of a stack and only rendering it out when needed.

Can't immediately find how to go from a `runtime.Callers` to a full stack trace, but I'll look around a bit more.

ti-mo:
Thanks for the feedback. :pray: 

Something else I considered is making `sys.FD` an interface, since it's already mostly consumed from outside the package through a clean API, but I don't think that's an improvement performance-wise if weighed against adding a single `*metadata` pointer in the struct. I would suggest we carry the field initially, maybe we can store metadata in a dedicated lookaside map later on.

lmb:
> Something else I considered is making sys.FD an interface, since it's already mostly consumed from outside the package through a clean API, but I don't think that's an improvement performance-wise if weighed against adding a single *metadata pointer in the struct.

Agreed. sys.FD as interface means additional allocation and it also prevents function inlining, which it'd be good to avoid.

> I would suggest we carry the field initially, maybe we can store metadata in a dedicated lookaside map later on.

Metadata would have to become `any` for that to work I guess?

internal/sys/fd.go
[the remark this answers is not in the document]
lmb: https://pkg.go.dev/runtime#CallersFrames is what you want.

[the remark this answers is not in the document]
ti-mo: No longer using `debug.Stack()`. `runtime.CallersFrames` output is stored in the metadata and I've added our own minimal formatter for a `*runtime.Frames`.

ti-mo:
@lmb Needs a bit of polish, but swapped this out for a much simpler design instead of the pub-sub thing. sys.FD no longer has a metadata pointer, all metadata is memoized in a package global.

We keep merging leaks, so we should at least consider running this internally. For working on Cilium itself, I can keep a local branch with some setters for `sys.{Create,Close,Finalize}`.

lmb (commented):
I like your idea to just use the fd as a key, much more straightforward than
my overengineered idea. I've read through my earlier longer comment, and I now
think I was wrong on a couple of things :D

* Tracing a regular (non-GC) fd.Close doesn't seem like a realistic use case to
  me. One can grep the code base or strace.
* You're probably right that making fd a bit bigger isn't the end of
  the world.

The design I came up with tried to minimize the code we need in the library, but
I think that came at the expense of ergonomics. I wasn't a big fan of the idea
and it shows ;)

# What callbacks do we need?

Currently we have this:

```go
var Create, Close, Leak func()
```

Create and Close are only needed to maintain metadata. This approach forces
a global lock (or similar) to make concurrent callbacks safe. This isn't a great
API if you want to do leak detection, especially if you want to run it in production.

I think we should sacrifice some memory, add a metadata pointer to sys.FD and slim down
the exported API in exchange.

```go
// TraceLeakedFd controls tracing of file descriptor leaks.
//
// fn is called whenever a file descriptor is freed from a GC finalizer instead
// of being explicitly closed. Passing nil for fn disables tracing.
//
// Only fds created after the function call returns are traced.
func TraceLeakedFd(fn func(...))

// Later on we could allow stacking tracers via this extension:

// Returns the previously active tracer or nil. It is the responsibility of the
// caller to invoke the previous tracer.
func TraceLeakedFd(fn ...) func(...)
```

We should have a proper setter, so that behind the scenes we can make things non-racy.
For example using https://pkg.go.dev/sync/atomic#example-Value-ReadMostly

# What metadata do we need?

You currently track a name and a backtrace of the creator. Name seems like the
minimum viable thing, but it creates code churn (thinking of SetName here) and
is brittle.

The backtrace is very useful, since it always works without having to do anything
on our part. I'm worried it's expensive to generate a traceback, but that is
probably premature.

I think we should add `*runtime.Frames` to sys.FD, but skip name.

# Are fds the right level of abstraction to trace?

Probably not feasible, but hear me out. We're discussing adding metadata to
sys.FD, but in reality we already have a lot of it in Map, Program, Link, etc.

Would it be possible to have an API like this instead?

```go
func TraceLeakedMap(func(*Map))
func TraceLeakedProgram(func(*Program))
...
```

Maybe we could get away with not adding runtime.Frames after all? The downside
is that we'd have to categorize all kinds of fds, especially the link package
is problematic here.

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,81 @@
...
+
+// TraceTestMain registers an fd leak tracer, runs the test suite, triggers
+// garbage collection and exits the process with the return code obtained
+// from the test suite.
+func TraceTestMain(m *testing.M) {
+	Finalize = Exiter
+
+	ret := m.Run()
+
+	if len(fdMeta) > 0 {
```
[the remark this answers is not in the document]
ti-mo: A single `runtime.GC` does/can not collect objects with finalizers. During mark phase, when an unreachable object is hit, its finalizer is stripped, scheduled onto a separate goroutine for execution, and collection of the object itself gets punted to the next cycle. Even a `runtime.GC` + `runtime.Gosched` is not a guarantee for finalizers running. Initially there was a `GC()` + `time.Sleep(100 * time.Millisecond)` here, which is bad for obvious reasons, and _still_ not a guarantee. :)

Checking fdMeta is very cheap and mostly waterproof.

link/cgroup.go
```diff
@@ -32,12 +32,13 @@ func AttachCgroup(opts CgroupOptions) (Link, error) {
 	if err != nil {
 		return nil, fmt.Errorf("can't open cgroup: %s", err)
 	}
+	defer cgroup.Close()
```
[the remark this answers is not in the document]
ti-mo: Damn, that's subtle. I'll add some comments.

internal/sys/fd_metadata.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (m metadata) String() string {
+	return fmt.Sprintf("name '%s': created at:\n%s", m.name, formatFrames(m.stack))
+}
+
+func callersFrames() *runtime.Frames {
+	c := make([]uintptr, 16)
+	for {
+		// Skip runtime.Callers and this function.
+		i := runtime.Callers(2, c)
```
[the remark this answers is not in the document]
ti-mo: There are multiple call sites within the test suite (think subtests etc.) that are already more than 16 deep. Without the bottom frame, a stack trace can be pretty useless to the caller, especially in larger projects, so I'd definitely keep the loop. Maybe we should set the default size to 32, though.

I'll account for a zero `i`.

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,81 @@
+package sys
+
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"testing"
```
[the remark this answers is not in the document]
ti-mo: Looks like I lost sight of this again. Putting this in testutils seems to work for all packages except for `internal`. There's already a  `testutils -> internal` dependency. Do we need to create a package just for this?

internal/sys/fd.go
```diff
@@ -87,10 +110,29 @@ func (fd *FD) Dup() (*FD, error) {
...
+	return newFD(dup, fd.Name()), nil
 }
 
 func (fd *FD) File(name string) *os.File {
 	fd.Forget()
 	return os.NewFile(uintptr(fd.raw), name)
 }
+
+// SetName annotates fd with a name if leak tracing is enabled.
+func (fd *FD) SetName(name string) {
```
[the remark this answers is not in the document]
ti-mo: Dropped name from metadata altogether, it was of marginal value if only to be used for internal purposes.

ti-mo:
Re-spun this a bit. We've concluded in offline discussion that this kind of API is best kept internal for now. Cilium can do its own resource tracking for higher-level objects.

In the last version, I've:
- merged metadata and FDTrace
- protected accesses to fdmap and made it accessible through methods on `sys.FDs`
- changed the default frame count to 32
- removed all hooks except for `sys.Finalize`

Looking for more ideas for simplification. We need an fd registry to be able to query open FDs at the end of a test run, since any reliance on GC for this purpose will flake/race. Adding a metadata pointer to sys.FD is not enough and incurs memory overhead on all users, however limited it may be.

Marking this as ready to review.

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,137 @@
...
+// Finalize is called when the Go runtime frees an [FD].
+var Finalize FDTracer
+
+// traceFDs returns true if there are active fd tracers registered to the
+// package.
+func traceFDs() bool {
+	return Finalize != nil
+}
+
+type fdmap struct {
```
lmb, line 24: This should be replaced with a https://pkg.go.dev/sync#Map instead, to save on code we have to maintain.

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	f(t)
+}
+
+func callersFrames() *runtime.Frames {
+	c := make([]uintptr, 32)
+	for {
+		// Skip runtime.Callers and this function.
+		i := runtime.Callers(2, c)
```
lmb, line 111: Like I said before, retrying doesn't seem like the right choice here. "32 frames should be enough for everyone" applies I think. We just need a good enough stack to be able to figure out where the object was created. All of this CallersFrames stuff is a bit black magic, and I'd prefer to stick as close to the examples as possible.

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,137 @@
...
+// created while an fd tracer was active.
+var FDs = newFDMap()
+
+// Finalize is called when the Go runtime frees an [FD].
+var Finalize FDTracer
+
+// traceFDs returns true if there are active fd tracers registered to the
+// package.
+func traceFDs() bool {
+	return Finalize != nil
```
lmb, line 21: Doesn't seem worth it to have a function for a single boolean?

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,137 @@
...
+	"runtime"
+	"strings"
+	"sync"
+)
+
+// FDs is a registry of all file descriptors wrapped into sys.FDs that were
+// created while an fd tracer was active.
+var FDs = newFDMap()
+
+// Finalize is called when the Go runtime frees an [FD].
```
lmb, line 15: Document that it only fires for FDs that are created after Finalize is set?

link/cgroup.go
```diff
@@ -35,12 +35,18 @@ func AttachCgroup(opts CgroupOptions) (Link, error) {
...
 
 	var cg Link
 	cg, err = newLinkCgroup(cgroup, opts.Attach, clone)
+	if err == nil {
+		cgroup.Close()
+		clone.Close()
+		return cg, nil
+	}
+
+	// cgroup and clone are retained by progAttachCgroup.
```
lmb, line 49: Aaah, subtle and not great. My fault I guess :D

map_test.go
```diff
@@ -31,6 +31,19 @@ var (
 	}
 )
 
+func createHash(t *testing.T) *Map {
```
lmb, line 34: Also add `t.Cleanup(func() { hash.Close() })`?

lmb (changes requested):
So, the new API surface is something like this:

```go
type sys.FDTracer ...
type sys.FDTrace ...
sys.Finalize = foo
sys.FDs.Len() // etc.
```

I think we can trim this down further. Especially sys.FDs leaks an internal consideration ("how are leaks tracked?") into the API.

Here's my proposal:

```go
// TraceLeakedFD controls tracing FD lifetime to detect FDs that are not closed via Close.
//
// If fn is not nil, tracing is enabled for all FDs going forward. The function is then invoked
// for all FDs which have their finalizer invoked by the GC.
//
// If fn is nil, tracing is disabled. All FDs that have not been closed yet are considered to be leaked, and fn will be invoked for them.
//
// TODO: What happens when fn is not nil, but tracing is already active? Maybe ok to panic for now.
func TraceLeakedFD(fn func(*runtime.Frames))
```

`FDTracer` is gone, it's not necessary. `FDTrace` is gone since we can easily update internal callers, no need to be extensible. `TraceLeakedFDs` works because we usually want to disable tracing and check for unclosed FDs at the same time.

`formatFrames` can go into the testutil.

internal/sys/fd_trace.go
ti-mo: Ack on hardcoding 32 frames, that's fine for internal API. I found the dynamic resizing somewhere on Sourcegraph, but not sure if it was from the stdlib. This should be fine to do either way, but I'll remove it.

internal/sys/fd_trace.go
ti-mo: sync.Map doesn't have the equivalent of len() and iterating is quite clunky. I'll give it a try after addressing the other feedback.

map_test.go
ti-mo: Good idea!

internal/sys/fd_trace.go
ti-mo: There used to be multiple hooks, so this made sense. `if traceFDs() { ... }` is still a bit more expressive than `if TraceLeakedFD != nil { ... }`, but ok, I'll remove it.

lmb: Keep it if you prefer it.

internal/sys/fd_trace.go
ti-mo: Went with sync.Map after all.

internal/sys/fd_trace.go
ti-mo: Ended up removing it since `FD.finalize()` needed a nil check on `onLeakFD`.

internal/testutils/fdtrace/fd.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"runtime"
+	"testing"
+
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+// TestMain runs m with sys.FD leak tracing enabled.
+func TestMain(m *testing.M) {
+	sys.OnLeakFD(func(fs *runtime.Frames) {
```
lmb, line 15: Second part of my suggestion:

```suggestion
	var ret int
	sys.OnLeakFD(func(fs *runtime.Frames) {
		ret = 42
		// ...
	})
	
	ret = m.Run()
	sys.OnLeakFD(nil)
	os.Exit(ret)
```

internal/sys/fd_trace.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// fn is nil past this point.
+
+	if onLeakFD == nil {
+		return
+	}
+
+	// Call onLeakFD for all open fds.
+	if fs := flushFrames(); len(fs) != 0 {
+		fmt.Fprintln(os.Stderr, "found open fds while disabling leak tracing:")
```
lmb, line 43: I think output and os.Exit should be up to onLeakFD. I can imagine later on adding a helper that does `t.Fail` instead of os.Exit for example. See my comment on TestMain.

internal/testutils/fdtrace/fd.go
ti-mo: Brilliant!

PR #736, opened 2022-07-14
perf add timeout argument

xpu22:


perf/reader.go
```diff
@@ -162,6 +163,7 @@ type ReaderOptions struct {
 	// Read will process data. Must be smaller than PerCPUBuffer.
 	// The default is to start processing as soon as data is available.
 	Watermark int
+	TimeOut   int
```
florianl, line 166: I think this option should have at least document what it is used for so users can reason about its impact. Also if there is a default value, it should be mentioned here as well I think.

lmb:
Hi @xpu22, thanks for your PR!

There was an earlier proposal for this in https://github.com/cilium/ebpf/pull/523 The gist of the discussion is:

- Instead of a timeout option, there should be a `Reader.SetDeadline` function.
- You can take the implementation on `Poller` from https://github.com/cilium/ebpf/commit/26fab34199f4ff02e816bb7d0f991306f22433f0#diff-1f29559e5dfacc2fda670f413123949365e356c977391255950d29a89a6e983bR124-R141

Please also add some unit tests and take a look at the failing CI: https://ebpf.semaphoreci.com/jobs/b80f08ab-6ffd-4473-8887-27ad0c4c0508#L572

perf/reader.go
xpu22: Ok, thanks

lmb:
https://ebpf.semaphoreci.com/jobs/690fd5ce-e7d0-4400-9f58-3da6f915e3a0#L573 is still failing. Please ping me explicitly when you're ready for another round of review.

xpu22:
> https://ebpf.semaphoreci.com/jobs/690fd5ce-e7d0-4400-9f58-3da6f915e3a0#L573 is still failing. Please ping me explicitly when you're ready for another round of review.

Hi @lmb please review

internal/epoll/poller.go
```diff
@@ -118,8 +119,8 @@ func (p *Poller) Add(fd int, id int) error {
 // Wait for events.
 //
 // Returns the number of pending events or an error wrapping os.ErrClosed if
-// Close is called.
-func (p *Poller) Wait(events []unix.EpollEvent) (int, error) {
+// Close is called, or os.ErrDeadlineExceeded if EpollWait timeout.
+func (p *Poller) Wait(events []unix.EpollEvent, deadline time.Time) (int, error) {
```
florianl, line 123: From a personal point of view, I would prefer `time.Duration` over `time.Time` as argument type of `deadline`.
Is there a use case to specify a date like `2022-08-01 00:00:00 +0000 UTC` as deadline?

lmb: That would be SetTimeout. I mentioned this in the other PR: we can add SetTimeout as a convenience wrapper around SetDeadline, but the underlying implementation must use time.Time.

lmb: Or do you mean Duration here and keep time.Time for the public API?

internal/epoll/poller.go
```diff
@@ -128,7 +129,12 @@ func (p *Poller) Wait(events []unix.EpollEvent) (int, error) {
 	}
 
 	for {
-		n, err := unix.EpollWait(p.epollFd, events, -1)
+		msec := int(-1)
+		if !deadline.IsZero() {
+			msec = int(time.Until(deadline).Milliseconds())
```
lmb, line 134: Bug in my own code: this doesn't handle deadlines in the past or so far in the future that msec is overflowed.

I'll fix that up in a follow up.

perf/reader.go
```diff
@@ -16,8 +17,9 @@ import (
 )
 
 var (
-	ErrClosed = os.ErrClosed
-	errEOR    = errors.New("end of ring")
+	ErrClosed  = os.ErrClosed
+	ErrTimeOut = os.ErrDeadlineExceeded
```
lmb, line 21: This alias isn't necessary. Just document that `os.ErrDeadlineExceeded` is returned.

perf/reader.go
lmb: With `SetDeadline` this Option isn't used anymore.

perf/reader.go
```diff
@@ -280,6 +285,18 @@ func (pr *Reader) Close() error {
 	return nil
 }
 
+// SetDeadline the Read timeout
```
lmb, line 288:
```suggestion
// SetDeadline controls how long Read and ReadInto will block waiting for samples.
//
// Passing a zero time.Time will remove the deadline.
```

internal/epoll/poller.go
```diff
@@ -138,6 +144,10 @@ func (p *Poller) Wait(events []unix.EpollEvent) (int, error) {
 			return 0, err
 		}
 
+		if n == 0 && msec != -1 {
+			return 0, os.ErrDeadlineExceeded
```
lmb, line 148: I think we should wrap the error here, otherwise we return a naked ErrDeadlineExceeded.

```suggestion
			return 0, fmt.Errorf("poll: %w", os.ErrDeadlineExceeded)
```

lmb (changes requested):
Looks good, thanks for following up so quickly!

internal/epoll/poller.go
florianl: Setting a fixed date (month/day/year ...) in the future (or past) as a deadline to read from a socket feels strange to me. Usually something like `s`, `ms` or `ns` is used. But maybe that is just my usage of sockets.

even for the public API I would not expect so see a `time.Time` as input for such a functionality.

florianl: yeah make sense. `SetDeadline(time.Time)` and `SetTimeout(Duration)`

nashuiliang:
Cool, adding the timeout of ebpf-perf is very necessary!
Hi, @lmb @florianl @xpu22, have you considered using [context.Context WithCancel/WithDeadline](https://pkg.go.dev/context) for this?

The new [Read()](https://github.com/cilium/ebpf/blob/master/perf/reader.go#L293) API is as follows:

```golang
func (pr *Reader) ReadWithContext(ctx context.Context) (Record, error)
```

ti-mo:
> have you considered using [context.Context WithCancel/WithDeadline](https://pkg.go.dev/context) for this?

We have, but IMO the performance characteristics of `context.Context` aren't very well-suited to this kind of API. Granted, maybe that's not our call to make and we could just offer the functionality. If the ring isn't that busy, the overhead of Context could be acceptable if it makes it much more ergonomic to use.

Could you give a practical use case for it? In any case, I would suggest we take this discussion to a dedicated issue.

lmb (approved):
Thanks for finishing this work, I was waiting for you to ping me for another review. Sorry it took a while.

Issue #741, opened 2022-07-20
Load string sections as global data maps


dylandreimerink:
I ran into the limitation that we don't support globally allocated strings when attempting to load [tracee](https://github.com/aquasecurity/tracee) with [edb](https://github.com/dylandreimerink/edb) ([related issue](https://github.com/dylandreimerink/edb/issues/25)).

When attempting to load these programs we get the `string is not stack allocated: not supported` error. This is because clang/LLVM create a `.rodata.str1.1` section for string literals. Tracee [uses string literals](https://github.com/aquasecurity/tracee/blob/0b29052c9b89e445e731e9a80c3e621b6df4b3a6/pkg/ebpf/c/tracee.bpf.c#L3145) in a few locations which seem to work well for them. Libbpf seems to handle this by creating a separate global data map(frozen arraymap with 1 key and the section contents as value). 

For example:
```c
#include <stddef.h>
#include <linux/bpf.h>
#include "bpf_helpers.h"
#include "bpf_trace_common.h"
#include <linux/ptrace.h>

typedef char custkey[16];

struct
{
    __uint(type, BPF_MAP_TYPE_HASH);
    __uint(max_entries, 2);
    __type(key, custkey);
    __type(value, __u32);
} my_map SEC(".maps");

SEC("xdp")
int bpf_prog1(struct xdp_md *ctx)
{
    __u32 *value = bpf_map_lookup_elem(&my_map, "Hello World!!!!");
    if (value)
        (*value)++;

    return XDP_PASS;
}
```
When loaded via `bpftool`: `sudo bpftool prog load example_prog /sys/fs/bpf/example_prog`
Will load without errors:
```
$ sudo bpftool prog show name bpf_prog1                            
55561: xdp  name bpf_prog1  tag 829c1a82180cf970
        loaded_at 2022-07-20T16:05:10+0200  uid 0
        xlated 104B  jited 67B  memlock 4096B  map_ids 28210,28212
        btf_id 6001
        
$ sudo bpftool map show id 28212                                   
28212: array  name .rodata.str1.1  flags 0x480
        key 4B  value 16B  max_entries 1  memlock 4096B
        frozen
        
$ sudo bpftool map dump id 28212
key: 00 00 00 00  value: 48 65 6c 6c 6f 20 57 6f  72 6c 64 21 21 21 21 00
Found 1 element
```

So I propose we also allow the loading of string sections as global data.

PR #742, opened 2022-07-20
elf_reader: Allow strings read-only global data sections

dylandreimerink:
In the past we had the limitation that string sections like
`.rodata.str1.1` could not not be loaded since clang/LLVM doesn't
provide BTF information for such sections. In #675 we lifted the
requirement to have BTF information for global data sections. This means
that we can now also allow string sections to be loaded as global data.

All facilities to do so are already in place, this commit just removes
the check for references to sections with the `SHF_STRINGS` flag set.

Fixes: #741
Signed-off-by: Dylan Reimerink <dylan.reimerink@isovalent.com>


ti-mo (dismissed):
Thanks! Could we also try to load these on kernels that support them?

elf_reader_test.go
```diff
@@ -422,13 +422,64 @@ func TestLoadInvalidInitializedBTFMap(t *testing.T) {
...
+	if prog == nil {
+		t.Fatal("program not found")
+	}
+
+	testMap := coll.Maps["my_map"]
+	if testMap == nil {
+		t.Fatal("test map not found")
+	}
+
+	const runCount = 3
```
lmb, line 464: Why run this multiple times, isn't once enough?

lmb (commented):
Nice!

elf_reader_test.go
dylandreimerink: Just habit, doesn't matter in this case since actually testing the program is not the point. It is to prove the section loaded and can be used as map key. I changed it to exec just once now.

lmb (approved):

Issue #344, opened 2021-07-14
Assemble a dummy BTF blob for probing StructOps maps


ti-mo:
With https://github.com/cilium/ebpf/pull/321, an API for probing available map types in the kernel was added. However, a StructOps map requires a valid BTF blob to be specified in order to make creation work.

@qmonnet was able to get this (partially?) working in bpftool: https://github.com/cilium/ebpf/pull/321#discussion_r662944737.

This issue is for implementing the equivalent using a pre-baked (or assembled at runtime) BTF blob to be able to probe this map type successfully.

cc @rgo3

qmonnet:
Works completely for bpftool, but I haven't submitted a patch upstream yet.

<details><summary>Full working patch:</summary>

```diff
diff --git a/tools/lib/bpf/libbpf_probes.c b/tools/lib/bpf/libbpf_probes.c
index ecaae2927ab8..629d39c98f10 100644
--- a/tools/lib/bpf/libbpf_probes.c
+++ b/tools/lib/bpf/libbpf_probes.c
@@ -203,6 +203,22 @@ bool bpf_probe_map_type(enum bpf_map_type map_type, __u32 ifindex)
 	__u32 btf_key_type_id = 0, btf_value_type_id = 0;
 	struct bpf_create_map_attr attr = {};
 	int fd = -1, btf_fd = -1, fd_inner;
+	int btf_vmlinux_value_type_id = 0;
+	struct btf *btf_vmlinux;
+
+	/* [1] INT 'int' size=4 bits_offset=0 nr_bits=32 encoding=SIGNED */
+	__u8 const btf_data[] = {
+		0x9f, 0xeb, 0x01, 0x00, 0x18, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x30, 0x00, 0x00, 0x00,
+		0x30, 0x00, 0x00, 0x00, 0x09, 0x00, 0x00, 0x00,
+		0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
+		0x04, 0x00, 0x00, 0x00, 0x20, 0x00, 0x00, 0x01,
+		0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x0d,
+		0x00, 0x00, 0x00, 0x00, 0x07, 0x00, 0x00, 0x00,
+		0x01, 0x00, 0x00, 0x00, 0x05, 0x00, 0x00, 0x00,
[...]

PR #746, opened 2022-07-22
features: implement probes for StructOps maps and programs

ti-mo:
Turns out this was never really blocking: https://github.com/cilium/ebpf/issues/344. Specifying any invalid vmlinux type id when creating a map is enough to get some information other than EINVAL.

### DIFF
--- a/features/map.go
+++ b/features/map.go
@@ -75,6 +75,10 @@ func createMapTypeAttr(mt ebpf.MapType) *sys.MapCreateAttr {
 		a.BtfKeyTypeId = 1   // BTF_KIND_INT
 		a.BtfValueTypeId = 3 // BTF_KIND_ARRAY
 		a.BtfFd = ^uint32(0)
+	case ebpf.StructOpsMap:
+		// StructOps requires setting a vmlinux type id, but id 1 will always
+		// resolve to some type of integer. This will cause ENOTSUPP.
+		a.BtfVmlinuxValueTypeId = 1
 	}
 
 	return a
@@ -100,14 +104,6 @@ func validateMaptype(mt ebpf.MapType) error {
 	if mt > mt.Max() {
 		return os.ErrInvalid
 	}
-
-	if mt == ebpf.StructOpsMap {
-		// A probe for StructOpsMap has vmlinux BTF requirements we currently
-		// cannot meet. Once we figure out how to add a working probe in this
-		// package, we can remove this check.
-		return errors.New("a probe for MapType StructOpsMap isn't implemented")
-	}
-
 	return nil
 }
 
@@ -128,6 +124,12 @@ func haveMapType(mt ebpf.MapType) error {
 	// For nested and storage map types we accept EBADF as indicator that these maps are supported
 	case errors.Is(err, unix.EBADF):
 		if isMapOfMaps(mt) || isStorageMap(mt) {
+			err = nil
+		}
+
+	// ENOTSUPP means the map type is at least known to the kernel.
+	case errors.Is(err, sys.ENOTSUPP):
+		if mt == ebpf.StructOpsMap {
 			err = nil
 		}
 
--- a/features/map.go
+++ b/features/map.go
@@ -25,73 +25,59 @@ type mapCache struct {
 }
 
 func createMapTypeAttr(mt ebpf.MapType) *sys.MapCreateAttr {
-	var (
-		keySize        uint32 = 4
-		valueSize      uint32 = 4
-		maxEntries     uint32 = 1
-		innerMapFd     uint32
-		flags          uint32
-		btfKeyTypeID   uint32
-		btfValueTypeID uint32
-		btfFd          uint32
-	)
+	a := &sys.MapCreateAttr{
+		MapType:    sys.MapType(mt),
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 1,
+	}
 
 	// switch on map types to generate correct MapCreateAttr
 	switch mt {
 	case ebpf.StackTrace:
 		// valueSize needs to be sizeof(uint64)
-		valueSize = 8
+		a.ValueSize = 8
 	case ebpf.LPMTrie:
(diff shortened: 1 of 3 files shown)


features/prog.go
```diff
@@ -248,7 +252,7 @@ func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
 
 func progLoadProbeNotImplemented(pt ebpf.ProgramType) bool {
 	switch pt {
-	case ebpf.Tracing, ebpf.StructOps, ebpf.Extension, ebpf.LSM:
```
rgo3, line 251: We might want to consider the StructOps program probe separately. This works fine for only probing for the program type but will allow using it in the HaveProgramHelper API as well which probably won't work: 
assuming e.g. `HaveProgramHelper(ebpf.StructOps, asm.FnExampleHelper)` will return ENOTSUPP as with the regular `HaveProgramType` feature test, but won't actually tell us anything about the StructOps/Helper combination.

features/map.go
```diff
@@ -138,6 +136,10 @@ func haveMapType(mt ebpf.MapType) error {
 			err = nil
 		}
 
+	// ENOTSUPP means the map type is at least known to the kernel.
+	case errors.Is(err, unix.ENOTSUPP):
```
rgo3, line 140:
```suggestion
	case errors.Is(err, unix.ENOTSUPP):
	    if mt == ebpf.StructOpsMap {
	        err = nil
	    }
	    fallthrough // only works if we put it above the err != nil case
```
Should we only accept it for StructOps(Map), but return inconclusive for other (future) map types where it might mean something else? Same for the program probe.

features/prog.go
rgo3: We could allow it for `HaveProgramType` but not for `HaveProgramHelper`.

ti-mo: Yep, this approach won't work for probing helper types. libbpf hasn't implemented this yet, either: https://elixir.bootlin.com/linux/v5.18.13/source/tools/lib/bpf/libbpf_probes.c#L403.

Crafting a correct program could be a follow-up goal, and would require finding a correct type ID in vmlinux. (I assume @qmonnet's patch suggestion in #344 successfully created a map?) From the looks of it, BTF marshaling might be necessary for that.

features/map.go
ti-mo: Yes, being defensive here might be better.

features/map.go
```diff
@@ -27,14 +27,15 @@ type mapCache struct {
 
 func createMapTypeAttr(mt ebpf.MapType) *sys.MapCreateAttr {
 	var (
-		keySize        uint32 = 4
```
lmb, line 30: Somewhat OT: just filling in `MapCreateAttr` would avoid duplicating all field names.

features/map.go
```diff
@@ -80,18 +81,23 @@ func createMapTypeAttr(mt ebpf.MapType) *sys.MapCreateAttr {
 		btfKeyTypeID = 1   // BTF_KIND_INT
 		btfValueTypeID = 3 // BTF_KIND_ARRAY
 		btfFd = ^uint32(0)
+	case ebpf.StructOpsMap:
+		// StructOps requires setting a vmlinux type id, but id 1 will always
+		// be some type of integer. This will cause ENOTSUPP.
+		btfVmlinuxValueTypeID = 1
```
lmb, line 87: How does setting this not cause EBADF / EINVAL from btf_get_by_fd? https://elixir.bootlin.com/linux/v5.18.14/source/kernel/bpf/syscall.c#L890

features/map.go
lmb: A general observation: `createMapTypeAttr` invites subtle bugs, because we split crafting the syscall payload from checking the specific error we expect. See the calls to `isMapOfMaps`, etc. in `haveMapType`.

features/map.go
rgo3: [`find_map_and_alloc()`](https://elixir.bootlin.com/linux/v5.18.14/source/kernel/bpf/syscall.c#L864) will return the `ENOTSUPP` when calling [`map_alloc()`](https://elixir.bootlin.com/linux/v5.18.14/source/kernel/bpf/syscall.c#L129) I think.

features/map.go
ti-mo: Sorry, could you explain why this is a problem?

features/map.go
ti-mo: Sure, will pick this up.

features/prog.go
```diff
@@ -151,6 +151,12 @@ func haveProgramType(pt ebpf.ProgramType) error {
 	case errors.Is(err, unix.EPERM):
 		break
 
+	// ENOTSUPP means the program type is at least known to the kernel.
+	case errors.Is(err, unix.ENOTSUPP):
+		if pt == ebpf.StructOps {
+			err = nil
+		}
+
```
rgo3, line 159: Should we add a `fallthrough` here in case `ENOTSUPP` happens unexpectedly? Same question for the map type probe. If it happens somewhere unexpectedly we would return it unwrapped.

rgo3 (changes requested):

features/prog.go
ti-mo: `fallthrough` hurts code readability somewhat, maybe it's probably time to drop the EPERM case and rewrite all outgoing errors in a `defer`, but I expect that approach to be met with resistance as well. :stuck_out_tongue:

ti-mo: I've made a separate proposal for this at https://github.com/cilium/ebpf/pull/749. Drafted this one for the time being.

features/prog.go
ti-mo: @rgo3 I think I see what you mean. Removing `ebpf.StructOps` from this switch case here will make `HaveProgramHelper(ebpf.StructOps,  ...)` calls now return `ENOTSUPP` instead of the more user-friendly `"a probe for ProgType ebpf.StructOps isn't implemented"`. I think taking this small hit is acceptable if it means we don't need to refactor this.

lmb (approved):

Issue #713, opened 2022-06-15
Support for BTF_KIND_DECL_TAG and BTF_KIND_TYPE_TAG


lmb:
Linux 5.18 added a two new BTF kinds [DECL_TAG and TYPE_TAG](https://github.com/torvalds/linux/blob/169e77764adc041b1dacba84ea90516a895d43b2/Documentation/bpf/btf.rst#2217-btf_kind_decl_tag). TYPE_TAG seems fine, the encoding for DECL_TAG looks like a headache.

Most relevant Linux commit: https://github.com/torvalds/linux/commit/7472d5a642c94a0ee1882ff3038de72ffe803a01

`pahole` only generates these kinds from v1.23 onwards:
* DECL_TAG https://git.kernel.org/pub/scm/devel/pahole/pahole.git/commit/?h=next&id=ec62499774c4bcac720706018385ced0c1f060ab
* TYPE_TAG https://git.kernel.org/pub/scm/devel/pahole/pahole.git/commit/?h=next&id=d99d551930cf4b4bc41eb5d68b7da69d718a6553

`clang-14` is also required: https://reviews.llvm.org/D111199

- [x] Parse / skip the new types
- [ ] Omit encoding new types in Spec.marshal if the kernel doesn't understand them
- [ ] Export types from `btf`

ti-mo:
With BTF-enabled kernels becoming generally available in distro's, we might need to think about tolerating/ignoring unknown types so older tools written with CO-RE are forwards-compatible with newer kernels.

Time for a `btf.ReaderOptions` of some sort? :)

lmb:
My current reading of this is that vmlinux BTF will only contain these new types if:
* clang >= 14
* pahole >= 1.23
* the kernel is **built with clang instead of gcc**

Not sure how many production kernels use clang instead of gcc.

lmb:
An idea I've been kicking around: DECL_TAG and TYPE_TAG are a bit of a bodge to add stringly typed metadata to the BTF wire format. What if we didn't represent them as `Type`, but instead added the metadata to `Pointer` or whatever?

```go
type Pointer struct {
	...
	Tags []string
}
```

When decoding BTF we'd add TYPE_TAGs to `Pointer.Tags` and during marshaling we'd create TYPE_TAG on the fly. Similar for DECL_TAG which applies to `Struct` and `Union` (and others?) This is the best solution from an API perspective, it's very easy to add and remove tags from a Type.

Downsides:
[...]

PR #747, opened 2022-07-25
btf: minimal support for BTF_KIND_DECL_TAG and BTF_KIND_TYPE_TAG

lmb:
btf: add limited support for BTF_KIND_DECL_TAG and BTF_KIND_TYPE_TAG

    Parse decl and type tags from BTF so that we can read BTF on new kernels.
    The new types are not exported since the encoding of declTag.Index especially
    is cumbersome to use. We can export the types later if necessary.

    Treat typeTag as a qualifier, since accoriding to the BTF documentation it
    is usually used in a chain of types:

        ptr -> [type_tag]*
            -> [const | volatile | restrict | typedef]*
            -> base_type

    Updates #713


ti-mo:
I think we should look at something like [bingo](https://www.bwplotka.dev/2020/bingo/) instead of `internal/tooldeps`. Could we drop that change from the PR and address it separately?

lmb:
Can you tell me why tooldeps doesn't work? What problem does bingo solve 
that is important to us?

PR #749, opened 2022-07-27
features: remove EPERM wrapping exception, automatically wrap errors

ti-mo:
```
In an attempt to simplify the switch blocks in all types of feature
probes, lift the arbitrary EPERM wrapping exception.

Also, move the call to fd.Close() out of the switch statement to live
closer to the call that returns the fd. The fd can always be immediately
closed.
```

```
Instead of relying on repeated calls to fmt.Errorf("...: %w", err) to ensure
we never return naked sentinels, add error wrappers at a few key locations.

This abstracts and centralizes the wrapping logic to make the probe logic
itself a bit leaner.
```

### DIFF
--- a/features/errors.go
+++ b/features/errors.go
@@ -1,0 +1,38 @@
+package features
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/cilium/ebpf"
+)
+
+// wrapProbeErrors wraps err to prevent callers from directly comparing
+// it to exported sentinels. Error rewriting in Go can be implemented by
+// deferring a closure over a named error return variable. This gives the
+// closure access to the stack space where the return value will be written,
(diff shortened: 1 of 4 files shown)


lmb (approved):
I think the closure is ok here, since it's the first thing that happens in the function.

There is a small downside to this approach: wrapping happens every time `Have*` is called, not just once. Probably not a big issue.

Ultimately I still wonder why we're not using `internal.FeatureTest` here. It has this figured out + it would allow feature tests to execute concurrently.

ti-mo:
> There is a small downside to this approach: wrapping happens every time Have* is called, not just once.

Yep, but errors are also no longer stored in wrapped form, so this trades (little) cpu for (little) memory!

I'll check if we can switch to `internal.FeatureTest`.

PR #751, opened 2022-07-28
link: (u|k)probe: don't treat EINVAL as os.ErrNotExist

ti-mo:
This is a follow-up to https://github.com/cilium/ebpf/pull/745.

No longer treat EINVAL from creating u/k(ret)probes as `os.ErrNotExist`, since EINVAL has many potential causes. This is a potentially breaking change, since callers scanning valid instruction offsets in symbols must now expect both `os.ErrNotExist` as well as `os.EINVAL`, depending on the kernel version and architecture.

Make errors shorter and more specific, e.g. 'bad insn boundary' no longer includes a 'not found' string. Token is now included in all errors, even unrecognized ones.

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -123,6 +123,9 @@ func Kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions) (Link, error
 // Losing the reference to the resulting Link (kp) will close the Kretprobe
 // and prevent further execution of prog. The Link must be Closed during
 // program shutdown to avoid leaking system resources.
+//
+// On kernels 5.10 and earlier, setting a kretprobe on a nonexistent symbol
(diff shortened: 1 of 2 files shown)


mmat11:
am I wrong or we are not catching EINVAL at all?

ti-mo:
This may be the sleep deprivation talking, but EINVAL would be handled by the `err != nil` catch-all.

mmat11:
> This may be the sleep deprivation talking, but EINVAL would be handled by the `err != nil` catch-all.

🙈 true, didn't notice it, LGTM!

looks like a 4.14 test needs to be adjusted

mmat11 (approved):

lmb:
Thanks Timo!

lmb (approved):

PR #754, opened 2022-08-01
prog: unconditionally enable verifier retries, add consts, document LogLevel

ti-mo:
With the recent addition of verifier error summaries, there are fewer reasons for disabling verifier output altogether. The existing behaviour is somewhat underspecified and underdocumented. Logging behaviour hinges on both LogLevel and LogSize, with an undocumented -1 value for LogSize to disable the verifier log. This gets rather complex with the retry mechanism involved.

This patch removes the local logSize variable, sets opts.LogSize directly to the default value, and log behaviour is now controlled by LogLevel only. Disabling the verifier log altogether can be done by setting the LogDisabled ProgramOption.

A type LogLevel was added to reflect the kernel's own constants for requesting logs on either branch or instruction level, as well as the stats summary of the current verifier pass on supported kernels. (v5.2 and up)

Documented all behaviour on the LogLevel and LogSize fields and constants.

This patch introduces one behavioural change: when ProgramOptions.LogLevel is left to its default value, a failed prog load will always result in a follow-up load with the verifier log level enabled (set to 1). The existing behaviour is to only request a verifier log if at least one of LogLevel or LogSize are set.

### DIFF
--- a/internal/cmd/gentypes/main.go
+++ b/internal/cmd/gentypes/main.go
@@ -65,21 +65,24 @@ func generateTypes(spec *btf.Spec) ([]byte, error) {
 	linkID := &btf.Int{Size: 4}
 	btfID := &btf.Int{Size: 4}
 	pointer := &btf.Int{Size: 8}
+	logLevel := &btf.Int{Size: 4}
 
-	// Pre-declare handwritten types sys.ObjName and sys.Pointer so that
-	// generated types can refer to them.
+	// Pre-declare handwritten types so that generated types can refer to them.
 	var (
-		_ sys.Pointer
 		_ sys.ObjName
 		_ sys.LinkID
+		_ sys.BTFID
+		_ sys.Pointer
+		_ sys.LogLevel
 	)
 
 	gf := &btf.GoFormatter{
 		Names: map[btf.Type]string{
-			objName: "ObjName",
-			linkID:  "LinkID",
-			btfID:   "BTFID",
-			pointer: "Pointer",
+			objName:  "ObjName",
+			linkID:   "LinkID",
+			btfID:    "BTFID",
+			pointer:  "Pointer",
+			logLevel: "LogLevel",
 		},
 		Identifier: internal.Identifier,
 		EnumIdentifier: func(name, element string) string {
@@ -276,6 +279,7 @@ import (
 				replace(objName, "prog_name"),
 				replace(enumTypes["ProgType"], "prog_type"),
 				replace(enumTypes["AttachType"], "expected_attach_type"),
+				replace(logLevel, "log_level"),
 				replace(pointer,
 					"insns",
 					"license",
--- a/prog.go
+++ b/prog.go
@@ -35,14 +35,43 @@ const (
 // verifier log.
 const DefaultVerifierLogSize = 64 * 1024
 
+// maxVerifierLogSize is the maximum size of verifier log buffer the kernel
+// will accept before returning EINVAL.
+const maxVerifierLogSize = math.MaxUint32 >> 2
+
 // ProgramOptions control loading a program into the kernel.
 type ProgramOptions struct {
-	// Controls the detail emitted by the kernel verifier. Set to non-zero
-	// to enable logging.
-	LogLevel uint32
-	// Controls the output buffer size for the verifier. Defaults to
-	// DefaultVerifierLogSize.
+	// Bitmap controlling the detail emitted by the kernel's eBPF verifier log.
+	// LogLevel-type values can be ORed together to request specific kinds of
+	// verifier output. See the documentation on [ebpf.LogLevel] for details.
+	//
+	//  opts.LogLevel = (ebpf.LogLevelBranch | ebpf.LogLevelStats)
+	//
+	// If left to its default value, the program will first be loaded without
+	// verifier output enabled. Upon error, the program load will be repeated
+	// with LogLevelBranch and the given (or default) LogSize value.
+	//
+	// Setting this to a non-zero value will unconditionally enable the verifier
+	// log, populating the [ebpf.Program.VerifierLog] field on successful loads
+	// and including detailed verifier errors if the program is rejected. This
+	// will always allocate an output buffer, but will result in only a single
+	// attempt at loading the program.
+	LogLevel LogLevel
+
+	// Controls the output buffer size for the verifier log, in bytes. See the
+	// documentation on ProgramOptions.LogLevel for details about how this value
+	// is used.
+	//
+	// If this value is set too low to fit the verifier log, the resulting
+	// [ebpf.VerifierError]'s Truncated flag will be true, and the error string
+	// will also contain a hint to that effect.
+	//
+	// Defaults to DefaultVerifierLogSize.
 	LogSize int
+
+	// Disables the verifier log completely, regardless of other options.
+	LogDisabled bool
+
 	// Type information used for CO-RE relocations.
 	//
 	// This is useful in environments where the kernel BTF is not available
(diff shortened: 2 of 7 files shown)


prog.go
```diff
@@ -35,14 +35,54 @@ const (
 // verifier log.
 const DefaultVerifierLogSize = 64 * 1024
 
+// LogLevel controls the verbosity of the kernel's eBPF program verifier.
+// These constants can be used for the ProgramOptions.LogLevel field.
+type LogLevel uint32
```
lmb, line 40: Nit: how about defining this in sys and using it from the attr?

prog.go
```diff
@@ -35,14 +35,54 @@ const (
...
+	// LogLevel-type values can be ORed together to request specific kinds of
+	// verifier output. See the documentation on [ebpf.LogLevel] for details.
+	//
+	//  opts.LogLevel = (ebpf.LogLevelBranch | ebpf.LogLevelStats)
+	//
+	// If left to its default value, the program will first be loaded without
+	// verifier output enabled. Upon error, the program load will be repeated
+	// with LogLevelBranch and the given (or default) LogSize value.
+	//
+	// Setting this to a value other than LogLevelDisabled will unconditionally
```
lmb, line 70: Not true for 0?

prog.go
```diff
@@ -294,10 +336,11 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		return &Program{unix.ByteSliceToString(logBuf), fd, spec.Name, "", spec.Type}, nil
 	}
 
-	if opts.LogLevel == 0 && opts.LogSize >= 0 {
-		// Re-run with the verifier enabled to get better error messages.
-		logBuf = make([]byte, logSize)
-		attr.LogLevel = 1
+	// If the caller did not specify a log level, re-run with a branch-level
+	// verifier logs and the default log buffer size.
```
lmb, line 340: Nit: not necessarily the default. The caller can specify their own size.

prog.go
lmb: What about adding stringer for LogLevel?

prog.go
```diff
@@ -35,14 +35,54 @@ const (
...
+	// Controls the output buffer size for the verifier log, in bytes. See the
+	// documentation on ProgramOptions.LogLevel for details about how this value
+	// is used.
+	//
+	// If this value is set too low to fit the verifier log, the resulting
+	// [ebpf.VerifierError]'s Truncated flag will be true, and the error string
+	// will also contain a hint to that effect.
+	//
+	// Defaults to DefaultVerifierLogSize.
+	LogSize uint64
```
lmb, line 84: Is uint64 really appropriate here? The go allocator is limited to int anyways (?) so there is no need to break compat.

prog.go
```diff
@@ -35,14 +35,54 @@ const (
...
+	// Print verifier state for every instruction.
+	// Available since Linux v5.2.
+	LogLevelInstruction
+
+	// Print verifier errors and stats at the end of the verification process.
+	// Available since Linux v5.2.
+	LogLevelStats
+
+	// Disable the verifier log.
+	LogLevelDisabled = math.MaxUint32
```
lmb, line 55: This is not kernel API right? What if we add `DisableDetailedErrors bool` to ProgramOptions instead?

lmb (changes requested):

prog.go
ti-mo: If we allow negative values, we need to sanitize, or the call to `make([]byte, opts.LogSize)` will panic. Previously, there was a check on `LogSize >= 0`, but why allow negative values to begin with?

prog.go
ti-mo: Sounds good!

prog.go
ti-mo: 'Setting' implying something other than 0, since that's the default value. Clear IMO, but this might change anyway if we go for a bool to disable the log.

prog.go
ti-mo: I get the rationale behind making the verifier log an implementation detail for future extensibility, but I'm not sure if that's really useful here. If we introduce a 'detailed errors' concept, that will need to be defined and explained to the caller. If the user is reaching for the option, it's mainly going to be for performance reasons, and they're going to be looking for a setting related to the verifier log.

Wouldn't `DisableVerifierLog bool` be more clear?

prog.go
lmb: I'm not saying that negative size makes sense, I'm saying that `int` is the natural choice for allocations. I'm fine panicking with negative size since that's what the runtime does.

Another thought that shows why we should stick to what the runtime does. why use uint64 not uint? We can't allocate that much memory on 32 bit anyways.

(One caveat of course: we used hacky negative size to disable the log, but never documented that. We'd have to hope that nobody outside of the lib uses that.)

ti-mo: > ... we used hacky negative size to disable the log, but never documented that. We'd have to hope that nobody outside of the lib uses that.

I'll change this back to an int and throw an error when it's negative, including a pointer to the new bool. Might save a user some time debugging after upgrading. As long as there's a way to disable the log and it's a quick fix, there should not be any complaints.

This should also be marked as a breaking change in the relnotes.

prog.go
ti-mo: Going with `LogDisabled` to stick to the `LogXYZ` pattern of field names.

lmb (approved):
LGTM, thanks!

types.go
```diff
@@ -282,3 +283,22 @@ type BatchOptions struct {
 	ElemFlags uint64
 	Flags     uint64
 }
+
+//go:generate stringer -type LogLevel -trimprefix Log
+
+// LogLevel controls the verbosity of the kernel's eBPF program verifier.
+// These constants can be used for the ProgramOptions.LogLevel field.
+type LogLevel sys.LogLevel
```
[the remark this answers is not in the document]
ti-mo: Stringer declares methods on `LogLevel`, which doesn't compile if LogLevel is an alias. Maybe I'm holding it wrong. I would also prefer an alias, because then we won't need the `LogLevel()` conversions anymore.

prog.go
```diff
@@ -180,6 +209,14 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 		return nil, fmt.Errorf("can't load %s program on %s", spec.ByteOrder, internal.NativeEndian)
 	}
 
+	if opts.LogSize < 0 {
+		return nil, errors.New("ProgramOptions.LogSize must be a positive value; disable verifier logs using ProgramOptions.LogDisabled")
+	}
+
+	if opts.LogSize > MaxVerifierLogSize {
```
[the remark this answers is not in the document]
ti-mo: Sounds good!

internal/cmd/gentypes/main.go
```diff
@@ -65,21 +65,24 @@ func generateTypes(spec *btf.Spec) ([]byte, error) {
 	linkID := &btf.Int{Size: 4}
 	btfID := &btf.Int{Size: 4}
 	pointer := &btf.Int{Size: 8}
+	logLevel := &btf.Int{Size: 4}
 
-	// Pre-declare handwritten types sys.ObjName and sys.Pointer so that
-	// generated types can refer to them.
+	// Pre-declare handwritten types so that generated types can refer to them.
 	var (
-		_ sys.Pointer
```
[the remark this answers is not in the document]
ti-mo: I reordered the list to match the block of btf types and the GoFormatter one below.

types.go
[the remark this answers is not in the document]
lmb: Duh!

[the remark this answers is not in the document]
ti-mo: After offline discussion, we'll skip the stringer and turn this into an alias for less tedium.

PR #755, opened 2022-08-03
link: add maxactive for kretprobe

alahaiyo:
In some case we want to missed as little as possible and
it is helpful to have this configuration option.

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -412,14 +412,9 @@ func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
 	if errors.Is(err, os.ErrNotExist) {
 		if typ == kprobeType && args.ret && args.retprobeMaxActive != 0 {
 			// In kernels earlier than 4.12, if maxactive is used to create kretprobe events,
-			// the event names created are not expected. We need to delete that event and
-			// create again without maxactive.
+			// the event names created are not expected.
 			_ = removeTraceFSProbeEvent(typ, fmt.Sprintf("-:kprobes/r_%s_0", sanitizeSymbol(args.symbol)))
-			args.retprobeMaxActive = 0
-			if err = createTraceFSProbeEvent(typ, args); err != nil {
-				return nil, fmt.Errorf("creating probe entry on tracefs: %w", err)
-			}
-			tid, err = getTraceEventID(group, args.symbol)
+			return nil, fmt.Errorf("create trace event with non-default maxactive: %w", ErrNotSupported)
 		}
 	}
 	if err != nil {
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -424,7 +424,7 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) (uint64, error) {
 	// entry, so we need to rely on reads for detecting uniqueness.
 	_, err := getTraceEventID(args.group, args.symbol)
 	if err == nil {
-		return 0, fmt.Errorf("trace event already exists: %s/%s", args.group, args.symbol)
+		return 0, fmt.Errorf("trace event %s/%s: %w", args.group, args.symbol, os.ErrExist)
 	}
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
 		return 0, fmt.Errorf("checking trace event %s/%s: %w", args.group, args.symbol, err)
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -403,7 +403,7 @@ func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
 			// If a livepatch handler is already active on the symbol, the write to
 			// tracefs will succeed, a trace event will show up, but creating the
 			// perf event will fail with EBUSY.
-			_ = closeTraceFSProbeEvent(typ, args.group, args.symbol, false)
+			_ = closeTraceFSProbeEvent(typ, args.group, args.symbol)
 		}
 	}()
 
@@ -414,7 +414,7 @@ func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
 			// In kernels earlier than 4.12, if maxactive is used to create kretprobe events,
 			// the event names created are not expected. We need to delete that event and
 			// create again without maxactive.
-			_ = closeTraceFSProbeEvent(typ, args.group, args.symbol, true)
+			_ = removeTraceFSProbeEvent(typ, fmt.Sprintf("-:kprobes/r_%s_0", sanitizeSymbol(args.symbol)))
 			args.retprobeMaxActive = 0
 			if err = createTraceFSProbeEvent(typ, args); err != nil {
 				return nil, fmt.Errorf("creating probe entry on tracefs: %w", err)
@@ -521,9 +521,13 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 }
 
 // closeTraceFSProbeEvent removes the [k,u]probe with the given type, group and symbol
-// from <tracefs>/[k,u]probe_events. The lowkernel flag is used to delete kretprobe
-// events that use maxactive in low version kernels (kernel < 4.12)
-func closeTraceFSProbeEvent(typ probeType, group, symbol string, lowKernel bool) error {
+// from <tracefs>/[k,u]probe_events.
+func closeTraceFSProbeEvent(typ probeType, group, symbol string) error {
+	pe := fmt.Sprintf("%s/%s", group, sanitizeSymbol(symbol))
+	return removeTraceFSProbeEvent(typ, pe)
+}
+
+func removeTraceFSProbeEvent(typ probeType, pe string) error {
 	f, err := os.OpenFile(typ.EventsPath(), os.O_APPEND|os.O_WRONLY, 0666)
 	if err != nil {
 		return fmt.Errorf("error opening %s: %w", typ.EventsPath(), err)
@@ -532,14 +536,8 @@ func closeTraceFSProbeEvent(typ probeType, group, symbol string, lowKernel bool)
 
 	// See [k,u]probe_events syntax above. The probe type does not need to be specified
 	// for removals.
-	var pe string
-	if lowKernel {
-		pe = fmt.Sprintf("-:kprobes/r_%s_0", symbol)
-	} else {
-		pe = fmt.Sprintf("-:%s/%s", group, sanitizeSymbol(symbol))
-	}
-	if _, err = f.WriteString(pe); err != nil {
-		return fmt.Errorf("writing '%s' to '%s': %w", pe, typ.EventsPath(), err)
+	if _, err = f.WriteString("-:" + pe); err != nil {
+		return fmt.Errorf("remove event %q from %s: %w", pe, typ.EventsPath(), err)
 	}
 
 	return nil
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -33,7 +33,7 @@ type probeType uint8
 type probeArgs struct {
 	symbol, group, path          string
 	offset, refCtrOffset, cookie uint64
-	pid                          int
+	pid, retprobeMaxActive       int
 	ret                          bool
 }
 
@@ -49,6 +49,12 @@ type KprobeOptions struct {
 	// Can be used to insert kprobes at arbitrary offsets in kernel functions,
 	// e.g. in places where functions have been inlined.
 	Offset uint64
+	// Increase the maximum number of concurrent invocations of a kretprobe.
+	// Required when tracing some long running functions in the kernel.
+	//
+	// Deprecated: this setting forces the use of an outdated kernel API and is not portable
+	// across kernel versions.
+	RetprobeMaxActive int
 }
 
 const (
@@ -191,6 +197,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 	}
 
 	if opts != nil {
+		args.retprobeMaxActive = opts.RetprobeMaxActive
 		args.cookie = opts.Cookie
 		args.offset = opts.Offset
 	}
@@ -241,6 +248,11 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 	et, err := getPMUEventType(typ)
 	if err != nil {
 		return nil, err
+	}
+
+	// Use tracefs if we want to set kretprobe's retprobeMaxActive.
+	if args.retprobeMaxActive != 0 {
+		return nil, fmt.Errorf("pmu probe: non-zero retprobeMaxActive: %w", ErrNotSupported)
 	}
 
 	var config uint64
@@ -391,12 +403,25 @@ func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
 			// If a livepatch handler is already active on the symbol, the write to
 			// tracefs will succeed, a trace event will show up, but creating the
 			// perf event will fail with EBUSY.
-			_ = closeTraceFSProbeEvent(typ, args.group, args.symbol)
+			_ = closeTraceFSProbeEvent(typ, args.group, args.symbol, false)
 		}
 	}()
 
 	// Get the newly-created trace event's id.
 	tid, err := getTraceEventID(group, args.symbol)
+	if errors.Is(err, os.ErrNotExist) {
+		if typ == kprobeType && args.ret && args.retprobeMaxActive != 0 {
+			// In kernels earlier than 4.12, if maxactive is used to create kretprobe events,
+			// the event names created are not expected. We need to delete that event and
+			// create again without maxactive.
+			_ = closeTraceFSProbeEvent(typ, args.group, args.symbol, true)
+			args.retprobeMaxActive = 0
+			if err = createTraceFSProbeEvent(typ, args); err != nil {
+				return nil, fmt.Errorf("creating probe entry on tracefs: %w", err)
+			}
+			tid, err = getTraceEventID(group, args.symbol)
+		}
+	}
 	if err != nil {
 		return nil, fmt.Errorf("getting trace event id: %w", err)
 	}
@@ -420,8 +445,15 @@ func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
 // createTraceFSProbeEvent creates a new ephemeral trace event by writing to
 // <tracefs>/[k,u]probe_events. Returns os.ErrNotExist if symbol is not a valid
 // kernel symbol, or if it is not traceable with kprobes. Returns os.ErrExist
-// if a probe with the same group and symbol already exists.
+// if a probe with the same group and symbol already exists. Returns os.ErrInvalid
+// if maxactive is != 0 for anything but a kretprobe.
 func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
+	if args.retprobeMaxActive != 0 {
+		if !args.ret || typ == uprobeType {
+			return fmt.Errorf("maxactive must be 0 except kretprobe: %w", os.ErrInvalid)
+		}
+	}
+
 	// Open the kprobe_events file in tracefs.
 	f, err := os.OpenFile(typ.EventsPath(), os.O_APPEND|os.O_WRONLY, 0666)
 	if err != nil {
@@ -447,7 +479,7 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		// the eBPF program itself.
 		// See Documentation/kprobes.txt for more details.
 		token = kprobeToken(args)
-		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, sanitizeSymbol(args.symbol), token)
+		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret, args.retprobeMaxActive), args.group, sanitizeSymbol(args.symbol), token)
 	case uprobeType:
 		// The uprobe_events syntax is as follows:
 		// p[:[GRP/]EVENT] PATH:OFFSET [FETCHARGS] : Set a probe
@@ -460,7 +492,7 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		//
 		// See Documentation/trace/uprobetracer.txt for more details.
 		token = uprobeToken(args)
-		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, args.symbol, token)
+		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret, 0), args.group, args.symbol, token)
 	}
 	_, err = f.WriteString(pe)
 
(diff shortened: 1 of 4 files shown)


lmb (commented):
Thank you for your PR!

Can you explain your use case for maxactive and why implementing your own sampling in BPF isn't sufficient?

As it stands, the PR is too special cased for me:
- doesn't work with pmu probes
- only supported for retprobes (?)

Kprobes have a large API, and supporting all of it is not possible. So we need a very compelling reason to add something like your proposed change.

alahaiyo:
> Thank you for your PR!
> 
> Can you explain your use case for maxactive and why implementing your own sampling in BPF isn't sufficient?
> 
> As it stands, the PR is too special cased for me:
> 
> * doesn't work with pmu probes
> * only supported for retprobes (?)
> 
> Kprobes have a large API, and supporting all of it is not possible. So we need a very compelling reason to add something like your proposed change.

alahaiyo:
> Thank you for your PR!
> 
> Can you explain your use case for maxactive and why implementing your own sampling in BPF isn't sufficient?
> 
> As it stands, the PR is too special cased for me:
> 
> * doesn't work with pmu probes
> * only supported for retprobes (?)
> 
> Kprobes have a large API, and supporting all of it is not possible. So we need a very compelling reason to add something like your proposed change.

If the traced function can sleep, then when the number of processes exceeds the number of CPUs, the kretprobe will not be triggered. Because the kernel defaults maxactive to NR_CPU. When we trace some kernel network functions (such as inet_accept), when the network connection pressure is very high, there will be a lot of kretprobe not triggered. 

When writing kprobe_event through tracefs, you can set maxactive to a maximum of 2048, which is enough. The implementation of pmu probe in the kernel will call create_local_trace_kprobe, in which maxactive is hardcoded to 0.

lmb:
Thanks for the background, that's interesting.

So what is the behaviour of a pmu retprobe in the same scenario? It drops retprobe calls, without being able to work around the limitation? Asked another way, what does it mean that maxactive is 0 in the pmu case?

What problem does maxactive guard against in the first place? Maybe we could just hardcode it to 2048 in the tracefs case?

Cc @mmat11

mmat11:
> It drops retprobe calls, without being able to work around the limitation? Asked another way, what does it mean that maxactive is 0 in the pmu case?

I believe this is the case

> What problem does maxactive guard against in the first place?

I found an explanation here https://github.com/iovisor/bcc/issues/1072#issue-217189257

> Maybe we could just hardcode it to 2048 in the tracefs case?

From this commit https://github.com/torvalds/linux/commit/696ced4fb1d76802f864d8848aa4716633f83c17 it seems the maximum is 4096. I guess there's some overhead (see https://elixir.bootlin.com/linux/v5.19/source/kernel/kprobes.c#L2202) in having a big number, else it could have been hardcoded in the kernel itself (?).

I wonder why this parameter isn't configurable via `perf_event_open`.
IMO we can support the feature but we should raise an error when it's not possible to do so (kernel < 4.12 or maxactive > 0 on non-return probes) and clearly document that setting this option will force the probe to be installed via tracefs

nashuiliang:
hi, @mmat11 

> I wonder why this parameter isn't configurable via `perf_event_open`. 

Tow related commits(not merged):

* https://lore.kernel.org/all/20191007223111.1142454-1-songliubraving@fb.com/

```text
Enable specifying maxactive for fd based kretprobe. This will be useful
for tracing tools like bcc and bpftrace. [1] discussed the need of this
in bpftrace. Use highest highest 12 bit (bit 52-63) to allow maximal
maxactive of 4095.

[1] https://github.com/iovisor/bpftrace/issues/835
```

* https://lore.kernel.org/bpf/20220714193403.13211-1-9erthalion6@gmail.com/

```text
From: Song Liu <songliubraving@fb.com>

Enable specifying maxactive for fd based kretprobe. This will be useful
for tracing tools like bcc and bpftrace (see for example discussion [1]).
Use highest 4 bit (bit 59-63) to allow specifying maxactive by log2.

The original patch [2] seems to be fallen through the cracks and wasn't
applied. I've merely rebased the work done by Song Liu, verififed it
still works, and modified to allow specifying maxactive by log2 per
suggestion from the discussion thread.

Note that changes in rethook implementation may render maxactive
obsolete.

[1]: https://github.com/iovisor/bpftrace/issues/835
[2]: https://lore.kernel.org/all/20191007223111.1142454-1-songliubraving@fb.com/
```

It seems to be discussing a more elegant way to implement.

> IMO we can support the feature but we should raise an error when it's not possible to do so (kernel < 4.12 or maxactive > 0 on non-return probes) and clearly document that setting this option will force the probe to be installed via tracefs

Agree, I also expect this feature.

alahaiyo:
Thank you for your reply！
> > It drops retprobe calls, without being able to work around the limitation? Asked another way, what does it mean that maxactive is 0 in the pmu case?
> 
> I believe this is the case

Agree
 
> > What problem does maxactive guard against in the first place?
> 
> I found an explanation here [iovisor/bcc#1072 (comment)](https://github.com/iovisor/bcc/issues/1072#issue-217189257)
> 
> > Maybe we could just hardcode it to 2048 in the tracefs case?
> 
> From this commit [torvalds/linux@696ced4](https://github.com/torvalds/linux/commit/696ced4fb1d76802f864d8848aa4716633f83c17) it seems the maximum is 4096. 

Oh my mistake, the maximum is 4096.

>I guess there's some overhead (see https://elixir.bootlin.com/linux/v5.19/source/kernel/kprobes.c#L2202) in having a big number, else it could have been hardcoded in the kernel itself (?).
> 
> I wonder why this parameter isn't configurable via `perf_event_open`. 

As @nashuiliang said, the feature is under discussion now.

> IMO we can support the feature but we should raise an error when it's not possible to do so (kernel < 4.12 or maxactive > 0 on non-return probes) and clearly document that setting this option will force the probe to be installed via tracefs

Yes, I agree. I will add this and commit again.

lmb:
@alahaiyo I'm not sure yet what the best approach is here, so there is a risk that your work might be wasted. I'll read through all the info here and try to reach an opinion today or tomorrow.

alahaiyo:
> @alahaiyo I'm not sure yet what the best approach is here, so there is a risk that your work might be wasted. I'll read through all the info here and try to reach an opinion today or tomorrow.

I'm really looking forward to your opinion.

lmb:
From the kernel side we have two separate APIs to create a kretprobe:

1. via `tracefs`: supports setting `maxactive` and missed events are reported via
[`/sys/kernel/debug/tracing/kprobe_profile`](https://docs.kernel.org/trace/kprobetrace.html#event-profiling).

2. via `perf_event_open`: supports neither `maxactive` nor missed events. Peter Zijlstra wants to [avoid adding maxactive to p_e_o if possible](https://lore.kernel.org/bpf/YtgMKDgNLnMIkHLI@worktop.programming.kicks-ass.net/). A patch to [expose missed events has gotten stuck, but the maintainer seems open to it](https://lore.kernel.org/bpf/20191028091229.GJ4131@hirez.programming.kicks-ass.net/).

On the user space side, maxactive is exposed in [`bcc`](https://github.com/iovisor/bcc/pull/2224) and [`gobpf`](https://github.com/iovisor/gobpf/pull/39). `bpftrace` [hardcodes maxactive to `0`](https://github.com/iovisor/bpftrace/blob/dcc105da3ff024f981143a8408b6d3956bfd73b6/src/attached_probe.cpp#L901) and would like to [print the number of missed events](https://github.com/iovisor/bpftrace/issues/835#issuecomment-515166844) instead.

Let's consider the implications of the `maxactive` API as proposed in the PR:

- On pre 4.12 kernels the user will receive an error from the kernel, since the feature is not supported. So users have to avoid setting maxactive on old kernels. This requires checking kernel version or similar, or feature probing from the library to avoid setting maxactive.
- On >= 4.12 kernels, we will force use of `tracefs` if maxactive!=0 even if `pmu` is available. `tracefs` behaves poorly when the tracer crashes.
- If `maxactive` becomes available on `pmu` after all, we have to do feature detection to
  fall back to `tracefs` for kernels that have `pmu` but don't have the maxactive API.
- If `maxactive` never becomes available on `pmu` we are forever locked into `tracefs` for kretprobes, even if regular pmu kprobes don't require maxactive anymore.
- It's completely unclear what users should set `maxactive` to, since it needs to account for all callers of a function on the whole system. So being able to set maxactive still doesn't guarantee that all retprobes fire.

All in all, `maxactive` seems like a bad API. I think `bpftrace` has it right with exposing the number of missed events instead. If we manage to resurrect the fd-based missed events patch we could provide a somewhat unified API across tracefs and pmu.

@alahaiyo in your PR description you say that you want to "miss as little as possible". Would it be acceptable to you if you instead had a way to see how many misses you had, and scaling your result accordingly?

If not, what value would you set `maxactive` to? How do you figure out whether the kretprobe needs `maxactive` in the first place? How do you plan on dealing with kernels < 4.12?

If just misses is not enough, we could add a higher level option like `FunctionMaySleep bool` or similar instead of `Maxactive int`. Behind the scenes the lib would try to figure out the best strategy, which for now would be setting `maxactive=4096` and forcing `tracefs`. This is better than `maxactive` since it avoids exposing the low level implementation detail and gives us flexibility in the future. The downside is that it requires a lot more feature probing and magic in the library itself.

lmb:
@alahaiyo do you notice the missed events on specific machines, or is your goal to reduce misses outright? If there is a problem with a specific machine, is CONFIG_PREEMPT enabled? What is the value of `/sys/devices/system/cpu/possible`?

lmb:
From https://github.com/iovisor/bcc/pull/2224#issue-412958330:

> For kernels without maxactive supports in debugfs, the error resolution is not straightforward. The kernel will still install a probe, but under a different name.

The issue is that < 4.12 kernels will treat the `r50:...` written to tracefs as the name of the probe. This will need to be handled if we end up adding maxactive after all.

alahaiyo:
> @alahaiyo do you notice the missed events on specific machines, or is your goal to reduce misses outright? If there is a problem with a specific machine, is CONFIG_PREEMPT enabled? What is the value of `/sys/devices/system/cpu/possible`?

It's not on specific machines. CONFIG_PREEMPT is not set and the value of `/sys/devices/system/cpu/possible` is `0-95`.

alahaiyo:
> From [iovisor/bcc#2224 (comment)](https://github.com/iovisor/bcc/pull/2224#issue-412958330):
> 
> > For kernels without maxactive supports in debugfs, the error resolution is not straightforward. The kernel will still install a probe, but under a different name.
> 
> The issue is that < 4.12 kernels will treat the `r50:...` written to tracefs as the name of the probe. This will need to be handled if we end up adding maxactive after all.

Yes, that is a good reference.

alahaiyo:
>On pre 4.12 kernels the user will receive an error from the kernel, since the feature is not supported. So users have to avoid setting maxactive on old kernels. This requires checking kernel version or similar, or feature probing from the library to avoid setting maxactive.

We can handle with that.

>On >= 4.12 kernels, we will force use of tracefs if maxactive!=0 even if pmu is available. tracefs behaves poorly when the tracer crashes.

In the interface of `link.Kprobe`, users can ignore maxactive, so the default is 0 and pmu will be used. If the user does not want to missed events, then they can set maxactive by themselves, especially when the misses is unbearable.

> If maxactive becomes available on pmu after all, we have to do feature detection to
fall back to tracefs for kernels that have pmu but don't have the maxactive API.
> If maxactive never becomes available on pmu we are forever locked into tracefs for kretprobes, even if regular pmu kprobes don't require maxactive anymore.
> It's completely unclear what users should set maxactive to, since it needs to account for all callers of a function on the whole system. So being able to set maxactive still doesn't guarantee that all retprobes fire.

Even setting `maxactive` to 4096 does not guarantee that events will not be missed, and if so we have no other way. But I think the user needs to be given the ability to adjust the possibility of event missed.

>If not, what value would you set maxactive to? How do you figure out whether the kretprobe needs maxactive in the first place? How do you plan on dealing with kernels < 4.12?

In fact, I don't know how to set maxactive. In our case it was only found that there were kretprobe events missing and this was not acceptable. The solution to this problem is to increase `maxactive` to a large enough value.

>we could add a higher level option like FunctionMaySleep bool or similar instead of Maxactive int

But I still think exposing `maxactive` is better than making the user think about whether the function might sleep. Maxactive is considered as an attribute of kretprobe, which can make it easier for users to find out whether the function of trace will have the problem of loss of kretprobe events.

link/kprobe.go
```diff
@@ -245,6 +249,10 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 
 	var config uint64
 	if args.ret {
+		// Use tracefs if we want to set kretprobe's maxactive.
+		if args.maxactive > 0 {
```
lmb, line 253: This should be outside of the `args.ret` block, so that we refuse maxactive for regular kprobes as well.

link/kprobe.go
```diff
@@ -245,6 +249,10 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
...
+			return nil, ErrNotSupported
```
lmb, line 254:
```suggestion
			return nil, fmt.Errorf("pmu probe: non-zero maxactive: %w", ErrNotSupported)
```

link/kprobe.go
```diff
@@ -460,7 +464,7 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		//
 		// See Documentation/trace/uprobetracer.txt for more details.
 		token = uprobeToken(args)
-		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret), args.group, args.symbol, token)
+		pe = fmt.Sprintf("%s:%s/%s %s", probePrefix(args.ret, args.maxactive), args.group, args.symbol, token)
```
lmb, line 467: I don't think uprobes support maxactive at all? Needs a test if they do after all.

link/kprobe.go
```diff
@@ -441,13 +449,9 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		// r:ebpf_1234/r_my_kretprobe nf_conntrack_destroy
 		// p:ebpf_5678/p_my_kprobe __x64_sys_execve
 		//
-		// Leaving the kretprobe's MAXACTIVE set to 0 (or absent) will make the
-		// kernel default to NR_CPUS. This is desired in most eBPF cases since
-		// subsampling or rate limiting logic can be more accurately implemented in
-		// the eBPF program itself.
 		// See Documentation/kprobes.txt for more details.
 		token = kprobeToken(args)
```
lmb, line 453: createTraceFSProbeEvent should return an error if maxactive is != 0 for anything but a kretprobe.

link/kprobe.go
```diff
@@ -49,6 +49,9 @@ type KprobeOptions struct {
 	// Can be used to insert kprobes at arbitrary offsets in kernel functions,
 	// e.g. in places where functions have been inlined.
 	Offset uint64
+	//Set kretprobe's MAXACTIVE if you need.
+	//Set to 0 (or absent) will make the kernel default to NR_CPUS.
+	Maxactive int
```
lmb, line 54: Let's add some more documentation and also deprecate this API to dissuade people from using it unless really necessary.

```suggestion
	// Increase the maximum number of concurrent invocations of a kretprobe.
	// Required when tracing some long running functions in the kernel.
	//
	// Deprecated: this setting forces the use of an outdated kernel API and is not portable
	// across kernel versions.
	RetprobeMaxActive int
```

link/kprobe.go
```diff
@@ -441,13 +449,9 @@ func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
 		// r:ebpf_1234/r_my_kretprobe nf_conntrack_destroy
 		// p:ebpf_5678/p_my_kprobe __x64_sys_execve
 		//
-		// Leaving the kretprobe's MAXACTIVE set to 0 (or absent) will make the
```
lmb, line 444: I think this bit of documentation can stay, it's useful knowledge.

link/kprobe.go
```diff
@@ -33,7 +33,7 @@ type probeType uint8
 type probeArgs struct {
 	symbol, group, path          string
 	offset, refCtrOffset, cookie uint64
-	pid                          int
+	pid, maxactive               int
```
lmb, line 36:
```suggestion
	pid, retprobeMaxActive               int
```

link/kprobe.go
```diff
@@ -193,6 +196,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 	if opts != nil {
 		args.cookie = opts.Cookie
 		args.offset = opts.Offset
+		args.maxactive = opts.Maxactive
```
lmb, line 199: Please return an error if `opts.Maxactive < 0` here.

lmb (changes requested):
Sorry for the long delay, wrapping my head around this issue has been tricky. I was hoping to find a better solution that exposing `maxactive`, but I'm drawing a blank.

I've left some comments for what I'd like you to change, mostly documentation and making sure we're refusing maxactive in as many cases as possible. Please also add tests that cover the maxactive API.

alahaiyo:
I'm glad that this PR can continue to push forward. I have made changes to the code based on the comments. @lmb

link/kprobe.go
```diff
@@ -191,6 +198,20 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 	}
 
 	if opts != nil {
+		if opts.RetprobeMaxActive < 0 {
+			return nil, fmt.Errorf("requires opts.RetprobeMaxActive >= 0: %w", errInvalidInput)
+		} else if opts.RetprobeMaxActive > 0 {
```
lmb, line 203: Nit: drop the unnecessary else.

```suggestion
		}
		
		if opts.RetprobeMaxActive > 0 {
```

link/kprobe.go
```diff
@@ -191,6 +198,20 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
...
+			// The kretprobe's maxactive can be set on kernels 4.12 and up.
+			v, err := internal.KernelVersion()
+			if err == nil {
+				kv, _ := internal.NewVersion("4.12")
+				if v.Less(kv) {
+					args.retprobeMaxActive = 0
```
lmb, line 209: Please remove the fallback based on kernel version or replace it with a full feature probe. Comparing kernel versions is not reliable.

link/kprobe.go
```diff
@@ -243,6 +264,11 @@ func pmuProbe(typ probeType, args probeArgs) (*perfEvent, error) {
 		return nil, err
 	}
 
+	// Use tracefs if we want to set kretprobe's retprobeMaxActive.
+	if args.retprobeMaxActive > 0 {
```
lmb, line 268: `!= 0`

link/kprobe.go
```diff
@@ -49,6 +50,12 @@ type KprobeOptions struct {
 	// Can be used to insert kprobes at arbitrary offsets in kernel functions,
 	// e.g. in places where functions have been inlined.
 	Offset uint64
+	// Increase the maximum number of concurrent invocations of a kretprobe.
+	// Required when tracing some long running functions in the kernel.
+	//
+	// Deprecated: this setting forces the use of an outdated kernel API and is not portable
+	// across kernel versions.
+	RetprobeMaxActive int
```
lmb, line 58: Should this be `uint` instead? Can you check whether the kernel parses this as signed or unsigned?

link/kprobe.go
```diff
@@ -191,6 +198,20 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 	}
 
 	if opts != nil {
+		if opts.RetprobeMaxActive < 0 {
```
lmb, line 201: I think we shouldn't check this here but let the kernel fail if it is invalid. Alternative is `uint` as above.

lmb (changes requested):

link/kprobe.go
alahaiyo: It's `int` in kernel.  If maxactive <= 0, it is set to a default value.

alahaiyo: > Should this be `uint` instead? Can you check whether the kernel parses this as signed or unsigned?

It's `int` in kernel.  If maxactive <= 0, it is set to a default value.

link/kprobe.go
alahaiyo: Maybe we should let it pass to the kernel and let kernel set to a default value.

link/kprobe.go
lmb: Tell me more, what do you have in mind? I'm not very familiar with tracefs myself.

link/kprobe.go
lmb: Agreed, let it be int and the kernel can deal with it.

lmb:
@alahaiyo I think the issue I pointed out in https://github.com/cilium/ebpf/pull/755#issuecomment-1211758219 is not addressed yet?

alahaiyo:
I found that it already has a query for the event ID after the event is created. So I just need to determine if I need to recreate it. My new commit is pushed. @lmb

lmb (commented):
I like how you integrated the check for unsupported maxactive! In the interests of getting this merged soon I'll push a couple of commits that contain changes I would've asked you to make. There is only one functional change: I'm dropping the fallback to disable maxactive.

Please take a look at my commits and let me know if you are OK with the modifications. Please also update the tests as mentioned.

link/kprobe.go
```diff
@@ -499,7 +532,12 @@ func closeTraceFSProbeEvent(typ probeType, group, symbol string) error {
 
 	// See [k,u]probe_events syntax above. The probe type does not need to be specified
 	// for removals.
-	pe := fmt.Sprintf("-:%s/%s", group, sanitizeSymbol(symbol))
+	var pe string
+	if lowKernel {
+		pe = fmt.Sprintf("-:kprobes/r_%s_0", symbol)
```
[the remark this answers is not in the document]
alahaiyo: In the old kernel, if the second character is not a`:`(https://elixir.bootlin.com/linux/v4.10/source/kernel/trace/trace_kprobe.c#L638), the event will not be created temporarily. Then here(https://elixir.bootlin.com/linux/v4.10/source/kernel/trace/trace_kprobe.c#L713), the kernel will create an event named `r_symbol_offset`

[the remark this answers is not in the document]
alahaiyo: > Why is this not sanitizing the symbol?

So I think we can't sanitizing the symbol when kernel would use full symbol name to create event.
There is an example：
`root@hecs-352404:~# echo "r1024:ebpf_abcdefg/_ib_umem_release __ib_umem_release" >> /sys/kernel/debug/tracing/kprobe_events`
`root@hecs-352404:~# cat /sys/kernel/debug/tracing/kprobe_events`
`r:kprobes/r___ib_umem_release_0 __ib_umem_release`

alahaiyo:
> I like how you integrated the check for unsupported maxactive! In the interests of getting this merged soon I'll push a couple of commits that contain changes I would've asked you to make. There is only one functional change: I'm dropping the fallback to disable maxactive.
> 
> Please take a look at my commits and let me know if you are OK with the modifications. Please also update the tests as mentioned.

I total OK with that. And I push my commits which deal with tests and a little bugfix.

link/kprobe.go
```diff
@@ -417,15 +408,32 @@ func tracefsProbe(typ probeType, args probeArgs) (_ *perfEvent, err error) {
...
+//
+// Returns os.ErrNotExist if symbol is not a valid
 // kernel symbol, or if it is not traceable with kprobes. Returns os.ErrExist
-// if a probe with the same group and symbol already exists.
-func createTraceFSProbeEvent(typ probeType, args probeArgs) error {
+// if a probe with the same group and symbol already exists. Returns an error if
+// args.retprobeMaxActive is used on non kprobe types. Returns ErrNotSupported if
+// the kernel is too old to support kretprobe maxactive.
+func createTraceFSProbeEvent(typ probeType, args probeArgs) (uint64, error) {
+	// Before attempting to create a trace event through tracefs,
```
lmb, line 421: @mmat11 does my refactor here make sense? I think it's nicer if createTraceFSProbeEvent encapsulates all the ugly bits of the tracefs interface.

link/kprobe.go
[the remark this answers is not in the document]
lmb: I updated the comment to reflect this.

lmb:
> deal with tests and a little bugfix.

A good catch. I've opted to return `os.ErrExist` from `create...` instead, that way we don't have to change the tests. I also included the symbol offset in the even name when removing, since that may be != 0.

link/kprobe.go
mmat11: lgtm!

PR #756, opened 2022-08-03
errors: remove verifier truncation heuristic, rely on ENOSPC

ti-mo:
In order to capture full verifier errors of large programs, the caller can
choose to repeatedly retry loading the program with progressively larger log
buffers, since the important info is always at the end of the log.

In doing so, it's incredibly likely for the last character of the buffer to
be a newline, resulting in the truncated flag being set to false when it should
instead be set to true, regardless of the kernel's ENOSPC return code.

This commit removes the newline heuristic that currently aims to detect log
truncation. Instead, we now require ENOSPC to be sent by the kernel in order
for the Truncated flag to be set in the VerifierError.

The heuristic was initially introduced for parsing BTF verifier errors,
since there's a latent kernel bug that causes ENOSPC only being returned
on a successful BTF load. Unfortunately, it turns out that not all BTF
verifier errors end in a newline either (see calls to btf_verifier_log()),
making the newline heuristic ineffective for BTF logs as well.

ErrorWithLog now takes an ancillary error from a second load operation that
is used to determine truncation status. This second error will not be wrapped
into the VerifierError.


lmb:
> There is no indication that kernels 5.18 and up don't return ENOSPC for BTF
loads.

Have you actually managed to get ENOSPC from loading BTF? In my experience an EINVAL or similar will override ENOSPC when loading invalid BTF. Maybe you're talking about ENOSPC from valid BTF with a small log?

Agreed re bringing the check for ENOSPC back, but I think we might still need the heuristic for BTF. We could fall back if the logErr is nil, or we could factor it out into a separate function and call that from the BTF loading code.

ti-mo:
> Have you actually managed to get ENOSPC from loading BTF? In my experience an EINVAL or similar will override ENOSPC when loading invalid BTF. Maybe you're talking about ENOSPC from valid BTF with a small log?

I see, I misinterpreted your `syscall will never return ENOSPC as of 5.18-rc4` comment to be 'as of 5.18-rc4...', implying something changed in 5.18. This should probably be `ENOSPC until at least 6.0`, as this is still the case on master. In `btf.c`, at the end of `btf_parse()`, ENOSPC is only returned if the log space check is hit, which requires a successful BTF load.

I'll think of something else.

ti-mo:
> We could fall back if the logErr is nil, or we could factor it out into a separate function and call that from the BTF loading code.

This feels clearer to me. BPF and BTF verifier logs have different styles and their APIs have different semantics.

btf/btf_test.go
```diff
@@ -285,6 +285,19 @@ func TestLoadSpecFromElf(t *testing.T) {
 	})
 }
 
+func TestVerifierError(t *testing.T) {
+	_, err := NewHandle(&Spec{})
```
lmb, line 289: What does an empty Spec represent though? Is it just type Void? Or something else?

internal/errors.go
```diff
@@ -2,39 +2,42 @@ package internal
...
 		log = log[:i]
-	} else if len(log) > 0 {
-		// No null byte? Dodgy!
-		truncated = true
 	}
 
 	log = bytes.Trim(log, whitespace)
 	if len(log) == 0 {
-		return &VerifierError{err, nil, false}
+		return &VerifierError{err, nil, truncated}
```
lmb, line 40: How can an empty log be truncated?

internal/errors.go
```diff
@@ -2,39 +2,42 @@ package internal
...
+// A secondary error (e.g. from a retried load with verifier log enabled)
+// can be provided that is checked for ENOSPC to determine if the kernel
+// truncated the log.
+func ErrorWithLog(err, ve error, log []byte) *VerifierError {
 	const whitespace = "\t\r\v\n "
 
+	// Either error can be ENOSPC depending on whether or now the verifier
+	// was explicitly enabled by the caller.
+	truncated := false
+	if errors.Is(err, unix.ENOSPC) || errors.Is(ve, unix.ENOSPC) {
```
lmb, line 28: What is the rationale to also check err? We've not done this before, and if the log is disabled ENOSPC can't mean that truncation occurred.

Put another way, we should only set truncated if `logEnabled && err == ENOSPC` to be conservative, no?

Nit: could also be `truncate := errors.Is ...`.

internal/errors.go
```diff
@@ -2,39 +2,42 @@ package internal
...
 //
-// A set of heuristics is used to determine whether the log has been truncated.
-func ErrorWithLog(err error, log []byte) *VerifierError {
+// A secondary error (e.g. from a retried load with verifier log enabled)
+// can be provided that is checked for ENOSPC to determine if the kernel
+// truncated the log.
+func ErrorWithLog(err, ve error, log []byte) *VerifierError {
 	const whitespace = "\t\r\v\n "
 
+	// Either error can be ENOSPC depending on whether or now the verifier
```
lmb, line 25: You mean the verifier log, not the verifier?

lmb (commented):
> In order to capture full verifier errors of large programs, the caller can
choose to repeatedly retry loading the program with progressively larger log
buffers, since the important info is always at the end of the log.

I think this is the first time I understand what you are trying to achieve, and why you need .Truncated to be accurate. I didn't consider that use case when introducing .Truncated, so I didn't mind that the heuristic isn't bomb proof.

You're trying to make .Truncated "reliable". Have you considered implementing resize-and-retry in the lib instead? Seems like a useful feature to have.

btf/btf_test.go
ti-mo: Not sure what to reply except the obvious: an empty type collection. On the wire, this results in a btf header, no types, and an empty string table, and this will result in a BTF verifier error saying 'No data'.

internal/errors.go
ti-mo: If the caller sets `LogLevel` > 0, `ve` is never assigned to. (the second load will be skipped) If the buffer is too small, ENOSPC will be in the first `err`.

The only other ENOSPC case I can find in PROG_LOAD is when a btf ID fails to be allocated for the prog (id's 1..INT_MAX exhausted). If a machine gets into this state, a false-positive truncated flag sounds minor in comparison.

That's not to say I don't agree. Thinking about it, setting the truncated flag doesn't belong in `ErrorWithLog` to begin with. Only setting Truncated when we've requested a log does sound safer.

internal/errors.go
ti-mo: I changed this to simplify the behaviour and testing around the input params. If any of err or ve is ENOSPC, truncated would be true.

prog.go
```diff
@@ -329,16 +329,6 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
...
-		logBuf = make([]byte, opts.LogSize)
-		attr.LogLevel = LogLevelBranch
-		attr.LogSize = uint32(len(logBuf))
-		attr.LogBuf = sys.NewSlicePointer(logBuf)
-		_, _ = sys.ProgLoad(attr)
-	}
-
 	switch {
 	case errors.Is(err, unix.EPERM):
 		if len(logBuf) > 0 && logBuf[0] == 0 {
```
lmb, line 334: Moving `!opts.LogDisabled && opts.LogLevel == 0` below the switch will make it so that this branch is almost never executed since logBuf is nil in most cases.

It's probably safer to keep the retry where it is.

ti-mo: Thanks, missed that.

PR #757, opened 2022-08-09
Expose map section name

EItanya:
Currently `ProgramSpec` exposes the `SectionName` of each prog, but the `MapSpec` does not do the same. We are using  the map section name and would like to expose it in the same way.

The test ignores the field as it is optional.


lmb:
As far as I can tell this is only useful to encode metadata for legacy-style map definitions, since BTF definitions have to come from `.maps`?

https://github.com/cilium/ebpf/blob/7fb0b5681c1a17a3fb20413d9ab048f95b700b7e/elf_reader.go#L74-L77

I don't mind adding this, wdzt @ti-mo? The tests should check that the elf reader correctly populates the section names though.

EItanya:
I tried adding it to the tests, but it failed on certain clang versions. Is there a way to ignore on those specific versions. Specifically is was < 7.

As far as usefulness, you can also add a suffix to the map section name. We were reading the entirety of the map section name, and using the added suffix for stats, etc.

lmb:
Can you provide me with the failing test output?

EItanya:
I added the section name to the test so it should show up there now

lmb:
The tests fail because you need to also set `SectionName` in `elfCode.loadMaps`. You probably also need to change `.maps` to `maps` since the latter is used for legacy style map definitions.

ti-mo:
@EItanya Before we commit to this approach, could you provide a real-world example to demonstrate how this is applied? Could you get away with embedding this information in the map name instead?

ti-mo:
Closing this due to inactivity.

PR #760, opened 2022-08-14
examples: add example for eBPF sockops

bigmikes:
I added an example code that uses eBPF sockops to:
1. Track passive and active established TCP connections.
2. Extract sRTT value every time TCP socket recalculates it.
3. Log 4-tuple and RTT in user space.


examples/tcprtt_sockops/main.go
```diff
@@ -0,0 +1,149 @@
...
+	go readLoop(rd)
+
+	// Wait
+	<-stopper
+}
+
+func findCgroupPath() (string, error) {
+	cgroupPath := "/sys/fs/cgroup"
+
+	enabled, err := cgroupv2.Enabled()
```
lmb, line 104: Do you think you could inline / copy the cgroup check into the example? It's very little code that is unlikely to need change and inlining would allow us to avoid a dependency. I'm asking because we're in the process of getting rid of the `examples/` module, it causes a lot of friction.

examples/tcprtt_sockops/main.go
```diff
@@ -0,0 +1,149 @@
...
+		"Src addr",
+		"Port",
+		"Dest addr",
+		"Port",
+		"RTT (ms)",
+	)
+	go readLoop(rd)
+
+	// Wait
+	<-stopper
```
lmb, line 98: I think you can simplify the example by getting rid of stopper and inlining readLoop into main. That also means you can get rid of handling `ringbuf.ErrClosed`.

examples/tcprtt_sockops/tcprtt_sockops.c
```diff
@@ -0,0 +1,120 @@
...
+
+	// Initialize the 4-tuple key
+	init_sk_key(skops, &sk_info.sk_key);
+	sk_info.sk_type = sock_type;
+
+	// Store the socket info in map using the 4-tuple as key
+	// We keep track of TCP connections in 'established' state
+	err = bpf_map_update_elem(&map_estab_sk, &sk_info.sk_key, &sk_info, BPF_NOEXIST);
+	if (err != 0) {
+		bpf_printk("Error in storing sk_key in map");
```
lmb, line 34: I'd prefer not encouraging printk, since it logs to the global trace pipe. Maybe just return here and add a comment why this could happen?

examples/tcprtt_sockops/tcprtt_sockops.h
```diff
@@ -0,0 +1,45 @@
+#ifndef TCPRTT_SOCKOPS_H
```
lmb, line 1: Is there a need to split the C across two files? I would find it easier to follow if we just had the .c.

examples/tcprtt_sockops/tcprtt_sockops.c
```diff
@@ -0,0 +1,120 @@
...
+	bpf_ringbuf_submit(rtt_event, 0);
+}
+
+static inline void bpf_sock_ops_state_cb(struct bpf_sock_ops *skops) {
+	struct sk_key sk_key = {};
+
+	// Socket changed state. args[0] stores the previous state.
+	// Perform cleanup of map entry if socket is exiting
+	// the 'established' state,
+	if (skops->args[0] == TCP_ESTABLISHED) {
```
lmb, line 93: Out of curiosity: is there a difference between BPF_SOCK_OPS_ACTIVE_ESTABLISHED_CB and grabbing the state transition into ESTABLISHED here? Might be clearer to follow if updating of map_estab_sk happens in a single function.

lmb (commented):
Very cool, thanks for contributing this example! I left some suggestions, please take a look.

examples/tcprtt_sockops/main.go
bigmikes: Other examples use the `stopper` signal. 

Don't we need to exit gracefully from the program so that the various `defer` statements cleanup the attachment to cgroup and the ring buffer?

examples/tcprtt_sockops/tcprtt_sockops.c
bigmikes: Sockops does not invoke your callback for `BPF_SOCK_OPS_STATE_CB` operation, unless you have registered such callback with `bpf_sock_ops_cb_flags_set`. Whereas, `BPF_SOCK_OPS_ACTIVE_ESTABLISHED_CB` and `BPF_SOCK_OPS_PASSIVE_ESTABLISHED_CB` are guaranteed to always be invoked. In fact, those CBs are where you register the other operations you are interested in with `bpf_sock_ops_cb_flags_set`.

examples/tcprtt_sockops/main.go
lmb: "Historical" context: `stopper` comes from the perf reader example, which uses it to show how to check for ErrClosed. So not all examples have to use the stopper. That said:

> Don't we need to exit gracefully from the program so that the various defer statements cleanup the attachment to cgroup and the ring buffer?

I had a look, and you're right that on older kernels we need the deferred `Link.Close` to run. So you can ignore me :)

lmb:
LGTM, could you rebase on master?

bigmikes:
I did something wrong during the rebase, let me close and create a new one...

PR #766, opened 2022-08-23
feature: define `sys.MapFlags` and re-export it for map flags feature detection

paulcacheux:
As requested in https://github.com/cilium/ebpf/pull/673 this PR defines a new internal type `sys.MapFlags` describing the different map flags constant, and use a re-exported version of it as the type argument for `HaveMapFlag` feature detection API

### DIFF
--- a/features/map.go
+++ b/features/map.go
@@ -164,9 +164,10 @@ func isStorageMap(mt ebpf.MapType) bool {
 	return false
 }
 
-// MapFlags indicates the flags passed to the kernel during map creation
+// MapFlags document which flags may be feature probed.
 type MapFlags = sys.MapFlags
 
+// Flags which may be feature probed.
 const (
 	BPF_F_NO_PREALLOC MapFlags = unix.BPF_F_NO_PREALLOC
 	BPF_F_RDONLY_PROG MapFlags = unix.BPF_F_RDONLY_PROG
@@ -177,6 +178,7 @@ const (
 
 // HaveMapFlag probes the running kernel for the availability of the specified map flag.
 //
+// Returns an error if flag is not one of the flags declared in this package.
 // See the package documentation for the meaning of the error return value.
 func HaveMapFlag(flag MapFlags) (err error) {
 	defer func() {
--- a/features/map.go
+++ b/features/map.go
@@ -13,7 +13,7 @@ import (
 
 func init() {
 	mc.mapTypes = make(map[ebpf.MapType]error)
-	mc.mapFlags = make(map[uint32]error)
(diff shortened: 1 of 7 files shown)


internal/sys/map_flags.go
```diff
@@ -0,0 +1,13 @@
+package sys
+
+import "github.com/cilium/ebpf/internal/unix"
+
+type MapFlags uint32
+
+const (
```
lmb, line 7: This should be in `features` for now, and document that only these flags are supported as arguments to `HaveMapFlag`

internal/sys/map_flags.go
```diff
@@ -0,0 +1,13 @@
+package sys
+
+import "github.com/cilium/ebpf/internal/unix"
+
+type MapFlags uint32
```
lmb, line 5: Please update `gentypes` so that https://github.com/cilium/ebpf/blob/c662d71024c8d86e1b365ada06d66fcfd8ef3905/internal/sys/types.go#L683 uses MapFlags as well. The setup for that happens here: https://github.com/cilium/ebpf/blob/c662d71024c8d86e1b365ada06d66fcfd8ef3905/internal/cmd/gentypes/main.go#L64-L86
There might be other structs that need the same change, please take a look.

lmb (changes requested):

internal/sys/map_flags.go
paulcacheux: As yes, I made the change for `sys.MapInfo` and `sys.MapCreateAttr`

internal/sys/map_flags.go
paulcacheux: What I did is having it in `sys` (allowing usage in `sys` so no circular imports) and re-export it in `features`. Should I do it differently ?

lmb: `type MapFlags` is fine, but the `const` block for the constants should be moved.

lmb: (I think you could keep `MapFlags` in types.go, but it doesn't matter much.)

paulcacheux: Oh right, I moved the constants in the last commit and made the change so that `sys.MapFlags` is generated by `gentypes`. I'm not super happy if you have an improvement in mind that can help with this please don't hesitate

lmb (approved):
Thanks! I sent you down the wrong path with my comment about keeping `MapFlags` in types.go, I meant syscall.go. Sorry!

paulcacheux:
Oh no worries ! Please let me know if something else is required

PR #774, opened 2022-08-31
link: memoize PMU event types read from sysfs

### LINKED ISSUE #744: link: memoize pmu IDs for kprobe / uprobes
As @brycekahle points out we currently do a lot of reading files in /sys/ when creating kprobes, etc: https://github.com/cilium/ebpf/commit/cb5b8b6084b4042fad214d237f2cd223732bb3ed#r49230542

It'd be nice to fix this by only reading the necessary values from /sys once and memoizing the result.

lmb:
Instead of parsing the same files over and over, read them once and store
the result. This cuts down on unnecessary file I/O.

Fixes #744

### DIFF
--- a/link/kprobe_test.go
+++ b/link/kprobe_test.go
@@ -307,19 +307,6 @@ func TestKprobeTraceFSGroup(t *testing.T) {
 	c.Assert(err, qt.Not(qt.IsNil))
 }
 
-func TestDetermineRetprobeBit(t *testing.T) {
-	testutils.SkipOnOldKernel(t, "4.17", "perf_kprobe PMU")
-	c := qt.New(t)
-
-	rpk, err := kretprobeBit()
-	c.Assert(err, qt.IsNil)
-	c.Assert(rpk, qt.Equals, uint64(0))
-
-	rpu, err := uretprobeBit()
-	c.Assert(err, qt.IsNil)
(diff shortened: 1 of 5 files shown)


link/kprobe_test.go
```diff
@@ -307,19 +307,6 @@ func TestKprobeTraceFSGroup(t *testing.T) {
 	c.Assert(err, qt.Not(qt.IsNil))
 }
 
-func TestDetermineRetprobeBit(t *testing.T) {
```
lmb, line 310: I removed this because it tests something that might change in the future (if I understand correctly).

mmat11 (approved):

link/kprobe_test.go
ti-mo: Hey, could you elaborate?

lmb: My thinking was:  Asserting that the value is 0 hardcodes an assumption the kernel wants to be able to change. Checking that err == nil just tests uint64FromFile, which is covered indirectly via kprobe, etc. tests.

We can add an explicit test of uint64FromFile if you want.

ti-mo: > Checking that err == nil just tests uint64FromFile, which is covered indirectly via kprobe, etc. tests.

Realized that too late, thanks!

PR #776, opened 2022-08-31
features: reuse internal.FeatureTest instead of open coding probes

lmb:
This PR changes `features` to use `internal.FeatureTest` instead of manually implementing a "cache struct" for the various kinds of probes. I've mentioned this in previous discussions of the package and finally got round do implementing it.

The advantages are:

- All feature probes use the same wrapping / caching behaviour. In the past we returned bare errors from `features`, and there are subtle differences in what errors are cached between `features` and `internal.FeatureTest`.
- Probes using `FeatureMatrix` can run concurrently with each other. So `HaveProgramType(SocketFilter)` doesn't block `HaveProgramType(Kprobe)` and vice versa. This is important to avoid deadlocks if feature tests call other feature tests.
- ErrNotSupported returned from `features` now include minimum kernel versions and feature names, like the rest of the library.
- It allows to eliminate the kludgy `Max()` methods on a bunch of constants.

I've also cut out copy pasted code that tried to avoid an import cycle with `ebpf`. That ship has now firmly sailed, since we use `ebpf` types in the public API of `features`.


features/prog.go
```diff
@@ -186,69 +170,57 @@ func HaveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) (err error)
...
+
+	key := helperKey{pt, helper}
+	if err, ok := helperCache.results[key]; ok {
 		return err
 	}
 
-	attr, err := createProgLoadAttr(pt, helper)
-	if err != nil {
-		return fmt.Errorf("couldn't create the program load attribute: %w", err)
+	if err := HaveProgramType(pt); err != nil {
```
lmb, line 189: This is a change in behaviour: when probing for a helper we always probe the program type. Opinions?

rgo3: What is the benefit of doing it this way considering we call `probeProgram(spec)` both times?

features/prog.go
```diff
@@ -154,17 +45,110 @@ func haveProgramType(pt ebpf.ProgramType) error {
...
+		if ft.Fn == nil {
+			key := key // avoid the dreaded loop variable problem
+			ft.Fn = func() error { return probeProgram(&ebpf.ProgramSpec{Type: key}) }
 		}
 	}
+}
 
-	pc.types[pt] = err
+type helperKey struct {
+	typ    ebpf.ProgramType
```
rgo3, line 143: nit/typo:
```suggestion
	type    ebpf.ProgramType
```

features/prog.go
lmb: Good question, I can't remember myself xD As best as I can tell it has the following advantages:

- It returns a "program type not supported" instead of "helper for program type not supported" on old kernels. Tiny change, but maybe less confusing for users.
- It avoids having to handle E2BIG in this function, since HaveProgramType takes care of that.

internal/feature.go
```diff
@@ -47,54 +57,74 @@ type featureTest struct {
...
+	defer ft.mu.Unlock()
+	// check one more time on the off
+	// chance that two go routines
+	// were able to call into the write
+	// lock
+	if ft.successful {
+		return ft.result
+	}
+	err := ft.Fn()
+	switch {
```
ti-mo, line 89: This is still too contrived (I know it's old code). A more common pattern:

```go
if err := ft.Fn(); !errors.Is(err, ErrNotSupported) {
  return fmt.Errorf()
}

ft.done = true
ft.result = &UFE{}

return ft.result
```

types.go
```diff
@@ -11,11 +11,6 @@ import (
 // that will be initialized in the kernel.
 type MapType uint32
 
-// Max returns the latest supported MapType.
-func (MapType) Max() MapType {
-	return maxMapType - 1
```
ti-mo, line 16: maxMapType/maxProgramType can also be removed. They're still referred to by Stringer-generated files, so they're not flagged by unused.

ti-mo (changes requested):

features/map.go
```diff
@@ -181,63 +180,78 @@ const (
...
-		return a, nil
-	case unix.BPF_F_NO_PREALLOC:
-		a.MapType = sys.MapType(ebpf.Hash)
-		return a, nil
-	}
-
-	return nil, errors.New("probe not implemented")
+var haveMapFlagsMatrix = internal.FeatureMatrix[MapFlags]{
+	BPF_F_NO_PREALLOC: {
+		Name:    "BPF_F_NO_PREALLOC",
```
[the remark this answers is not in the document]
lmb: I've added `MapFlags.String` and added an `init()` to set the fields.

internal/feature.go
```diff
@@ -47,54 +57,74 @@ type featureTest struct {
...
+	case errors.Is(err, ErrNotSupported):
+		v, err := NewVersion(ft.Version)
```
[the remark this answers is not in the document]
lmb: This works well for NewFeatureTest, but we're also constructing `FeatureTest` as part of `FeatureTestMatrix`. The best I can come up with is to change `FeatureTest.Version` to `internal.Version`. Then initialisation of a matrix becomes something like this:

```Go
var haveMapTypeMatrix = internal.FeatureMatrix[ebpf.MapType]{
	ebpf.Hash:           {Version: internal.Version{3,19}},
	ebpf.Array:          {Version: internal.Version{3,19}},
	ebpf.ProgramArray:   {Version: internal.Version{4,2}},
	ebpf.PerfEventArray: {Version: internal.Version{4,3}},
	ebpf.PerCPUHash:     {Version: internal.Version{4,6}},
	ebpf.PerCPUArray:    {Version: internal.Version{4,6}},
	ebpf.StackTrace: {
		Version: internal.Version{4,6},
		Fn: func() error {
...
```

Not great, but maybe acceptable?

internal/feature.go
lmb: It's  not so straight forward because you also have to deal with the `err == nil` case which should create UFE. I changed it into an if, else, not sure it's much more readable tbh.

types.go
lmb: Thanks!

internal/feature.go
ti-mo:
```go
	err := ft.Fn()
	if err == nil {
		ft.done = true
		return nil
	}
	if errors.Is(err, ErrNotSupported) {
		ft.done = true
		ft.result = &UnsupportedFeatureError{
			// MinimumVersion: v,
			Name: ft.Name,
		}

		return ft.result
	}

	// We couldn't execute the feature test to a point
	// where it could make a determination.
	// Don't cache the result, just return it.
	return fmt.Errorf("detect support for %s: %w", ft.Name, err)
}
```

internal/feature.go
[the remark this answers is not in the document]
ti-mo: Let's stick with strings.

ti-mo:
Rebased this, giving this another round.

features/misc.go
```diff
@@ -1,165 +1,79 @@
...
-
-// HaveV2ISA probes the running kernel if instructions of the v2 ISA are supported.
-//
-// See the package documentation for the meaning of the error return value.
-func HaveV2ISA() error {
-	return probeMisc(v2ISA)
-}
-
-// HaveV3ISA probes the running kernel if instructions of the v3 ISA are supported.
+// Upstream commit c04c0d2b968ac45d6ef020316808ef6c82325a82.
```
ti-mo, line 12: c04c0d2b968a ("bpf: increase complexity limit and maximum program size")

features/misc.go
```diff
@@ -1,165 +1,79 @@
...
-	case largeInsn:
-		for i := 0; i < maxInsns; i++ {
-			insns = append(insns, asm.Mov.Imm(asm.R0, 1))
-		}
-		insns = append(insns, asm.Return())
-	case boundedLoops:
-		insns = asm.Instructions{
+// HaveBoundedLoops probes the running kernel if bounded loops are supported.
+//
+// Upstream commit 2589726d12a1b12eaaa93c7f1ea64287e383c7a5.
```
ti-mo, line 32: 2589726d12a1 ("bpf: introduce bounded loops")

features/misc.go
```diff
@@ -1,165 +1,79 @@
...
-		}
-	case v2ISA:
-		insns = asm.Instructions{
+		},
+	})
+})
+
+// HaveV2ISA probes the running kernel if instructions of the v2 ISA are supported.
+//
+// Upstream commit 92b31a9af73b3a3fc801899335d6c47966351830.
```
ti-mo, line 49: 92b31a9af73b ("bpf: add BPF_J{LT,LE,SLT,SLE} instructions")

features/misc.go
```diff
@@ -1,165 +1,79 @@
...
-		}
-	case v3ISA:
-		insns = asm.Instructions{
+		},
+	})
+})
+
+// HaveV3ISA probes the running kernel if instructions of the v3 ISA are supported.
+//
+// Upstream commit 092ed0968bb648cd18e8a0430cd0a8a71727315c.
```
ti-mo, line 66: 092ed0968bb6 ("bpf: verifier support JMP32")

ti-mo (approved):
Went through it in-depth, great stuff. Left a few nits.

features/map.go
```diff
@@ -13,155 +12,217 @@ import (
...
+	// btf* fields need to be set
+	// see alloc_check for local_storage map types
+	err := createMap(&sys.MapCreateAttr{
+		MapType:        mt,
+		KeySize:        4,
+		ValueSize:      4,
+		MaxEntries:     0,
+		MapFlags:       unix.BPF_F_NO_PREALLOC,
+		BtfKeyTypeId:   1,
+		BtfValueTypeId: 1,
```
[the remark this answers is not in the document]
lmb: Yes, the ID is abitrary but has to be non-zero. 1 seems better than 3 in that case.

features/prog_test.go
```diff
@@ -97,26 +63,12 @@ func TestHaveProgramHelper(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("Could not read kernel version required for helper: %v", err)
-		}
-
-		if progVersion.Less(helperVersion) {
-			minVersion = tc.version
-		}
-
 		t.Run(fmt.Sprintf("%s/%s", tc.prog.String(), tc.helper.String()), func(t *testing.T) {
+			testutils.SkipIfNotSupported(t, HaveProgramType(tc.prog))
```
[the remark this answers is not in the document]
lmb: Dropped

features/prog.go
```diff
@@ -231,6 +139,18 @@ func init() {
 	}
 }
 
+type helperKey struct {
+	typ    ebpf.ProgramType
+	helper asm.BuiltinFunc
+}
+
+var helperCache = struct {
```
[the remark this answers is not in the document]
lmb: I've added a dedicated type `FeatureCache`.

PR #777, opened 2022-08-31
iter: Add support to load sleepable iterator program type

olsajiri:
Adding support to load sleepable iterator program type,
defined by 'iter.s/...' section.


olsajiri:
cc @ti-mo

Issue #780, opened 2022-09-01
btf: support for BTF_KIND_ENUM64


lmb:
There is a new kind BTF_KIND_ENUM64 to allow encoding enums with values > 2^32-1. The encoding is specified in the [kernel.org documentation](https://docs.kernel.org/bpf/btf.html#btf-kind-enum64).

My initial thinking is that we shouldn't expose a separate Enum64 type, but rather transform KIND_ENUM64 into an Enum. That would require changing the type of `EnumValue.Value` from `int32` to something larger, breaking API compat. However, I think we have to break compat anyways, due to this nugget from [patch 1](https://patchwork.kernel.org/project/netdevbpf/patch/20220607062600.3716578-1-yhs@fb.com/):

```
The kflag is also introduced for BTF_KIND_ENUM and BTF_KIND_ENUM64
to indicate whether the value is signed or unsigned. [...]
The kflag value 0 means unsigned values, which is consistent to the default by libbpf
and should also cover most cases as well.
```

We currently treat enum values as signed, while libbpf seems to do the opposite. When I initially implemented enum support I was under the impression that enums are signed. In reality their signedness is implementation defined: https://stackoverflow.com/a/2579250/19544965 The right thing to do is treat them as unsigned by default.

Full details are in the [upstream patchset from Yonghong](https://patchwork.kernel.org/project/netdevbpf/cover/20220607062554.3716237-1-yhs@fb.com/).

- [x] Extend EnumValue to 64 bits and default to unsigned enums in bpf2go
- [x] Somehow expose the signedness of an enum (`Enum.Signed`?)
- [ ] Read KIND_ENUM64 into `Enum`

PR #781, opened 2022-09-01
btf: handle signed enums and fix default signedness

lmb:
btf: handle signed enums and fix default signedness

    Enums now carry signedness encoded in the KindFlag, with the default being
    an unsigned enum. This breaks our assumption that enums are signed by default.

    Add Enum.Signed and switch EnumValue.Value to uint64. I used uint64 instead of
    uint32 since we'll likely need 64 bits to support BTF_KIND_ENUM64 in the future.

    Users of bpf2go that generated enums will see their generated types changing from
    signed to unsigned types.

    Updates #780

btf: reject invalid structs when formatting

    GoFormatter currently outputs nonsensical padding given a struct definition
    like the following:

        &Struct{
            Size: 1,
            Members: []Member{
                 {Name: "foo", Type: &Int{Size: 2}, Offset: 0},
            },
        }

    The reason is that the member is large than the enclosing struct. Return
    an error in such cases instead of formatting a bogus type.

btf: fix the output of GoFormatter for small enums

    When adding support for sized enums I only adjusted the code that
    writes out a type declaration, but forgot to change the output
    of enum type literals. This would lead to the wrong type being used
    for an enum that is used as a struct field.

    Fix this by moving the code from writeTypeDecl to writeTypeLit.

### DIFF
--- a/btf/btf_types.go
+++ b/btf/btf_types.go
@@ -237,7 +237,7 @@ type btfVariable struct {
 
 type btfEnum struct {
 	NameOff uint32
-	Val     int32
+	Val     uint32
 }
 
 type btfParam struct {
--- a/btf/core_test.go
+++ b/btf/core_test.go
@@ -209,7 +209,7 @@ func TestCOREFindEnumValue(t *testing.T) {
 		name                    string
 		local, target           Type
 		acc                     coreAccessor
-		localValue, targetValue int32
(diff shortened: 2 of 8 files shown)


lmb:
Source: https://patchwork.kernel.org/project/netdevbpf/patch/20220607062600.3716578-1-yhs@fb.com/

PR #791, opened 2022-09-09
prep work for BTF marshaling

lmb:
This is the first of three PRs to add BTF marshaling to the library, and does some groundwork. The second PR will contain the actual marshaling code, while the third PR has some more memory optimizations.

For the curious, the whole shebang is on [my btf-encoder branch](https://github.com/lmb/ebpf/tree/btf-encoder).

internal: move Deque
    
    Move Deque from btf into internal and export methods. No other code changes.

btf: make typeDeque generic
    
    Make typeDeque a generic container type so that we can re-use it
    for postorderTraversal.

btf: avoid heap allocations when walking types
    
    Type.walk() forces the typeDeque argument to always escape, since the escape analyzer has
    to be conservative when dealing with interfaces. We can apply two tricks that reduce
    allocations.
    
    First, Type.walk() is replaced with
    
        children() []*Type
    
    This removes the typeDeque parameter, and for types that don't have any children (like Void) we
    never incur an allocation at all. For types that do have children we've unfortunately swapped
    heap allocating typeDeque with allocating the []*Type slice instead.
    
    Which brings us to the second trick: by using a type switch we allow the compiler to inline the
    children() function for the most important Types, which in turn enables allocating most []*Type on
    the stack instead of on the heap.
    
    One notable exception is FuncProto, which allocates the returned slice as such:
    
        types := make([]*Type, len(fp.Params)+1)
    
    It seems like this trips up escape analysis for some reason as the slice is heap allocated.
    
        name                                      old time/op    new time/op    delta
        Walk/Void-4                                 22.6ns ± 2%     2.8ns ± 2%   -87.40%  (p=0.029 n=4+4)
        Walk/Int[unsigned_size=0]-4                 22.7ns ± 1%     3.2ns ± 1%   -86.09%  (p=0.029 n=4+4)
        Walk/Pointer[target=<nil>]-4                61.3ns ± 1%    38.5ns ± 1%   -37.26%  (p=0.029 n=4+4)
        Walk/Array[index=<nil>_type=<nil>_n=0]-4    63.8ns ± 1%    42.2ns ± 0%   -33.91%  (p=0.029 n=4+4)
        Walk/Struct[fields=2]-4                     64.9ns ± 1%    59.5ns ± 0%    -8.29%  (p=0.029 n=4+4)
        Walk/Union[fields=2]-4                      66.6ns ± 3%    59.3ns ± 0%   -10.84%  (p=0.029 n=4+4)
        Walk/Enum[size=0_values=0]-4                22.5ns ± 4%     2.9ns ± 2%   -87.22%  (p=0.029 n=4+4)
        Walk/Fwd[struct]-4                          22.2ns ± 1%     2.8ns ± 1%   -87.23%  (p=0.029 n=4+4)
        Walk/Typedef[<nil>]-4                       60.8ns ± 1%    38.7ns ± 1%   -36.42%  (p=0.029 n=4+4)
        Walk/Volatile[<nil>]-4                      62.9ns ± 3%    38.4ns ± 2%   -38.89%  (p=0.029 n=4+4)
        Walk/Const[<nil>]-4                         61.9ns ± 1%    39.8ns ± 3%   -35.78%  (p=0.029 n=4+4)
        Walk/Restrict[<nil>]-4                      64.2ns ± 1%    39.2ns ± 1%   -38.91%  (p=0.029 n=4+4)
        Walk/Func[static_proto=<nil>]-4             63.3ns ± 4%    39.5ns ± 2%   -37.57%  (p=0.029 n=4+4)
        Walk/FuncProto[args=2_return=<nil>]-4       67.8ns ± 5%    77.3ns ± 2%   +14.03%  (p=0.029 n=4+4)
        Walk/Var[static]-4                          63.2ns ± 1%    38.5ns ± 0%   -39.03%  (p=0.029 n=4+4)
        Walk/Datasec-4                              67.5ns ± 2%    58.5ns ± 1%   -13.31%  (p=0.029 n=4+4)
    
        name                                      old alloc/op   new alloc/op   delta
        Walk/Void-4                                  48.0B ± 0%      0.0B       -100.00%  (p=0.029 n=4+4)
        Walk/Int[unsigned_size=0]-4                  48.0B ± 0%      0.0B       -100.00%  (p=0.029 n=4+4)
        Walk/Pointer[target=<nil>]-4                  112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Array[index=<nil>_type=<nil>_n=0]-4      112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Struct[fields=2]-4                       112B ± 0%       80B ± 0%   -28.57%  (p=0.029 n=4+4)
        Walk/Union[fields=2]-4                        112B ± 0%       80B ± 0%   -28.57%  (p=0.029 n=4+4)
        Walk/Enum[size=0_values=0]-4                 48.0B ± 0%      0.0B       -100.00%  (p=0.029 n=4+4)
        Walk/Fwd[struct]-4                           48.0B ± 0%      0.0B       -100.00%  (p=0.029 n=4+4)
        Walk/Typedef[<nil>]-4                         112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Volatile[<nil>]-4                        112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Const[<nil>]-4                           112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Restrict[<nil>]-4                        112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Func[static_proto=<nil>]-4               112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/FuncProto[args=2_return=<nil>]-4         112B ± 0%       88B ± 0%   -21.43%  (p=0.029 n=4+4)
        Walk/Var[static]-4                            112B ± 0%       64B ± 0%   -42.86%  (p=0.029 n=4+4)
        Walk/Datasec-4                                112B ± 0%       80B ± 0%   -28.57%  (p=0.029 n=4+4)
    
        name                                      old allocs/op  new allocs/op  delta
        Walk/Void-4                                   1.00 ± 0%      0.00       -100.00%  (p=0.029 n=4+4)
        Walk/Int[unsigned_size=0]-4                   1.00 ± 0%      0.00       -100.00%  (p=0.029 n=4+4)
        Walk/Pointer[target=<nil>]-4                  2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Array[index=<nil>_type=<nil>_n=0]-4      2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Struct[fields=2]-4                       2.00 ± 0%      2.00 ± 0%      ~     (all equal)
        Walk/Union[fields=2]-4                        2.00 ± 0%      2.00 ± 0%      ~     (all equal)
        Walk/Enum[size=0_values=0]-4                  1.00 ± 0%      0.00       -100.00%  (p=0.029 n=4+4)
        Walk/Fwd[struct]-4                            1.00 ± 0%      0.00       -100.00%  (p=0.029 n=4+4)
        Walk/Typedef[<nil>]-4                         2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Volatile[<nil>]-4                        2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Const[<nil>]-4                           2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Restrict[<nil>]-4                        2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Func[static_proto=<nil>]-4               2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/FuncProto[args=2_return=<nil>]-4         2.00 ± 0%      2.00 ± 0%      ~     (all equal)
        Walk/Var[static]-4                            2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.029 n=4+4)
        Walk/Datasec-4                                2.00 ± 0%      2.00 ± 0%      ~     (all equal)

btf: add benchmark for Type.Walk
    
    Benchmark the cost of adding a type and its children to a typeDeque.
    
        name                                      time/op
        Walk/Void-4                               22.8ns ± 3%
        Walk/Int[unsigned_size=0]-4               22.3ns ± 1%
        Walk/Pointer[target=<nil>]-4              61.3ns ± 0%
        Walk/Array[index=<nil>_type=<nil>_n=0]-4  62.4ns ± 0%
        Walk/Struct[fields=2]-4                   64.2ns ± 0%
        Walk/Union[fields=2]-4                    64.6ns ± 0%
        Walk/Enum[size=0_values=0]-4              22.1ns ± 0%
        Walk/Fwd[struct]-4                        22.1ns ± 0%
        Walk/Typedef[<nil>]-4                     61.2ns ± 0%
        Walk/Volatile[<nil>]-4                    61.2ns ± 0%
        Walk/Const[<nil>]-4                       61.2ns ± 0%
        Walk/Restrict[<nil>]-4                    61.2ns ± 0%
        Walk/Func[static_proto=<nil>]-4           61.6ns ± 0%
        Walk/FuncProto[args=2_return=<nil>]-4     66.1ns ± 0%
        Walk/Var[static]-4                        61.5ns ± 1%
        Walk/Datasec-4                            64.4ns ± 0%
    
        name                                      alloc/op
        Walk/Void-4                                48.0B ± 0%
        Walk/Int[unsigned_size=0]-4                48.0B ± 0%
        Walk/Pointer[target=<nil>]-4                112B ± 0%
        Walk/Array[index=<nil>_type=<nil>_n=0]-4    112B ± 0%
        Walk/Struct[fields=2]-4                     112B ± 0%
        Walk/Union[fields=2]-4                      112B ± 0%
        Walk/Enum[size=0_values=0]-4               48.0B ± 0%
        Walk/Fwd[struct]-4                         48.0B ± 0%
        Walk/Typedef[<nil>]-4                       112B ± 0%
        Walk/Volatile[<nil>]-4                      112B ± 0%
        Walk/Const[<nil>]-4                         112B ± 0%
        Walk/Restrict[<nil>]-4                      112B ± 0%
        Walk/Func[static_proto=<nil>]-4             112B ± 0%
        Walk/FuncProto[args=2_return=<nil>]-4       112B ± 0%
        Walk/Var[static]-4                          112B ± 0%
        Walk/Datasec-4                              112B ± 0%
    
        name                                      allocs/op
        Walk/Void-4                                 1.00 ± 0%
        Walk/Int[unsigned_size=0]-4                 1.00 ± 0%
        Walk/Pointer[target=<nil>]-4                2.00 ± 0%
        Walk/Array[index=<nil>_type=<nil>_n=0]-4    2.00 ± 0%
        Walk/Struct[fields=2]-4                     2.00 ± 0%
        Walk/Union[fields=2]-4                      2.00 ± 0%
        Walk/Enum[size=0_values=0]-4                1.00 ± 0%
        Walk/Fwd[struct]-4                          1.00 ± 0%
        Walk/Typedef[<nil>]-4                       2.00 ± 0%
        Walk/Volatile[<nil>]-4                      2.00 ± 0%
        Walk/Const[<nil>]-4                         2.00 ± 0%
        Walk/Restrict[<nil>]-4                      2.00 ± 0%
        Walk/Func[static_proto=<nil>]-4             2.00 ± 0%
        Walk/FuncProto[args=2_return=<nil>]-4       2.00 ± 0%
        Walk/Var[static]-4                          2.00 ± 0%
        Walk/Datasec-4                              2.00 ± 0%

btf: drop local btf.Spec argument to CORERelocate
    
    CORERelocate currently takes ProgramSpec.BTF as an argument, since a CO-RE relocation
    may ask for the local type ID. This is problematic, since we want to get rid of ProgramSpec.BTF.
    
    Instead, cache the type ID when constructing the CORERelocation itself.

### DIFF
--- a/btf/traversal.go
+++ b/btf/traversal.go
@@ -1,0 +1,56 @@
+package btf
+
+import "fmt"
+
+// walkType calls fn on each child of typ.
+func walkType(typ Type, fn func(*Type)) {
+	// Explicitly type switch on the most common types to allow the inliner to
+	// do its work. This avoids allocating intermediate slices from walk() on
+	// the heap.
+	switch v := typ.(type) {
+	case *Void, *Int, *Enum, *Fwd, *Float:
+		// No children to traverse.
+	case *Pointer:
+		fn(&v.Target)
+	case *Array:
+		fn(&v.Index)
+		fn(&v.Type)
+	case *Struct:
+		for i := range v.Members {
+			fn(&v.Members[i].Type)
+		}
+	case *Union:
+		for i := range v.Members {
+			fn(&v.Members[i].Type)
+		}
+	case *Typedef:
+		fn(&v.Type)
+	case *Volatile:
+		fn(&v.Type)
+	case *Const:
+		fn(&v.Type)
+	case *Restrict:
+		fn(&v.Type)
+	case *Func:
+		fn(&v.Type)
+	case *FuncProto:
+		fn(&v.Return)
+		for i := range v.Params {
+			fn(&v.Params[i].Type)
+		}
+	case *Var:
+		fn(&v.Type)
+	case *Datasec:
+		for i := range v.Vars {
+			fn(&v.Vars[i].Type)
+		}
+	case *declTag:
+		fn(&v.Type)
+	case *typeTag:
+		fn(&v.Type)
+	case *cycle:
+		// cycle has children, but we ignore them deliberately.
+	default:
+		panic(fmt.Sprintf("don't know how to walk Type %T", v))
+	}
+}
--- a/btf/core.go
+++ b/btf/core.go
@@ -873,8 +873,8 @@ func coreAreTypesCompatible(localType Type, targetType Type) error {
 
 		case *Pointer, *Array:
 			depth++
-			localType.walk(&localTs)
-			targetType.walk(&targetTs)
+			walkType(localType, localTs.push)
+			walkType(targetType, targetTs.push)
 
 		case *FuncProto:
 			tv := targetType.(*FuncProto)
@@ -883,8 +883,8 @@ func coreAreTypesCompatible(localType Type, targetType Type) error {
 			}
 
 			depth++
-			localType.walk(&localTs)
-			targetType.walk(&targetTs)
+			walkType(localType, localTs.push)
+			walkType(targetType, targetTs.push)
 
 		default:
 			return fmt.Errorf("unsupported type %T", localType)
--- a/btf/core.go
+++ b/btf/core.go
@@ -855,7 +855,7 @@ func coreAreTypesCompatible(localType Type, targetType Type) error {
 		depth             = 0
 	)
 
-	for ; l != nil && t != nil; l, t = localTs.shift(), targetTs.shift() {
+	for ; l != nil && t != nil; l, t = localTs.Shift(), targetTs.Shift() {
 		if depth >= maxTypeDepth {
 			return errors.New("types are nested too deep")
 		}
@@ -873,8 +873,8 @@ func coreAreTypesCompatible(localType Type, targetType Type) error {
 
 		case *Pointer, *Array:
 			depth++
-			walkType(localType, localTs.push)
-			walkType(targetType, targetTs.push)
+			walkType(localType, localTs.Push)
+			walkType(targetType, targetTs.Push)
 
 		case *FuncProto:
 			tv := targetType.(*FuncProto)
@@ -883,8 +883,8 @@ func coreAreTypesCompatible(localType Type, targetType Type) error {
 			}
 
 			depth++
-			walkType(localType, localTs.push)
-			walkType(targetType, targetTs.push)
+			walkType(localType, localTs.Push)
+			walkType(targetType, targetTs.Push)
 
 		default:
 			return fmt.Errorf("unsupported type %T", localType)
--- a/btf/core.go
+++ b/btf/core.go
@@ -156,16 +156,17 @@ func (k coreKind) String() string {
 	}
 }
 
-// CORERelocate calculates the difference in types between local and target.
+// CORERelocate calculates changes needed to adjust eBPF instructions for differences
+// in types.
 //
 // Returns a list of fixups which can be applied to instructions to make them
 // match the target type(s).
 //
 // Fixups are returned in the order of relos, e.g. fixup[i] is the solution
 // for relos[i].
-func CORERelocate(local, target *Spec, relos []*CORERelocation) ([]COREFixup, error) {
-	if local.byteOrder != target.byteOrder {
-		return nil, fmt.Errorf("can't relocate %s against %s", local.byteOrder, target.byteOrder)
+func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder) ([]COREFixup, error) {
+	if bo != target.byteOrder {
+		return nil, fmt.Errorf("can't relocate %s against %s", bo, target.byteOrder)
 	}
 
 	type reloGroup struct {
@@ -185,15 +186,14 @@ func CORERelocate(local, target *Spec, relos []*CORERelocation) ([]COREFixup, er
 				return nil, fmt.Errorf("%s: unexpected accessor %v", relo.kind, relo.accessor)
 			}
 
-			id, err := local.TypeID(relo.typ)
-			if err != nil {
-				return nil, fmt.Errorf("%s: %w", relo.kind, err)
-			}
-
 			result[i] = COREFixup{
-				kind:   relo.kind,
-				local:  uint32(id),
-				target: uint32(id),
+				kind:  relo.kind,
+				local: uint32(relo.id),
+				// NB: Using relo.id as the target here is incorrect, since
+				// it doesn't match the BTF we generate on the fly. This isn't
+				// too bad for now since there are no uses of the local type ID
+				// in the kernel, yet.
+				target: uint32(relo.id),
 			}
 			continue
 		}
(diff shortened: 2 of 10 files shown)


btf/traversal.go
```diff
@@ -0,0 +1,52 @@
+package btf
+
+// walkType calls fn on each child of typ.
+//
+// It's faster than directly invoking typ.children for some common types.
+func walkType(typ Type, fn func(*Type)) {
```
ti-mo, line 6: Since this now contains a type switch, wouldn't this technically allow us to remove `children()` from interface `Type` and `walk(nil)` in the default case?

That would also allow us to tailor method names to the receiver, e.g. `Pointer.children()` could become `Pointer.target()` and `FuncProto.params()`, etc. Just a naive thought from looking at it from a distance.

From a domain definition perspective, a `children()` method doesn't make much sense as the method of traversal is so highly specific to the type and is hard to define in a generic way.

lmb: Yes, that would be possible, that's kind of what this function started out as. The downside is that it's quite easy to forget to extend walkType when adding a new implementation of Type, since there is no compiler feedback. It just blows up at some point.

Maybe not such a big deal, since we rarely add new types and also never want to allow users to implement Type. Should I change it?

ti-mo: > It just blows up at some point.

Well, my guess would be the opposite. If `walkType` encounters an unknown type, the default clause will cause traversal to silently stop, which is always more difficult to debug than a hard error.

An interface method reminds us to define traversal behaviour when adding a new Type, though in the current iteration with the default clause, it's still possible to implement a Type and forget to add it to the type switch, which will incur a heap alloc when traversing that particular Type.

ti-mo: I've looked at linters for this, nothing existing fits the bill. Some combination of https://github.com/nishanths/exhaustive/blob/master/enum.go and https://github.com/dominikh/implements/blob/master/main.go could work.

Some redeeming factors:
- If a new type shows up in BTF in the wild, I assume things will fail early on in the pipeline? (e.g. during inflation/parsing?)
- Reminder to extend `walkType` should be documented on the `children()` interface method.

Then I think we should be good.

lmb: I pushed a change. `children()` is now gone since there are no other callers left, and it's just duplicate code. Please take a look!

ti-mo (approved):

PR #792, opened 2022-09-14
Provide 15-byte context inputs to `Program.{Run,Benchmark,Test}`

ti-mo:
Centralize allocation of the buffer to avoid some duplication and ease future changes to the minimum buffer size.

The patch that introduced this regression doesn't officially land until 6.0, but it has already been backported to kernels as early as 5.15.65 (potentially even earlier?).

cc @dylandreimerink

### DIFF
--- a/internal/prog.go
+++ b/internal/prog.go
@@ -1,0 +1,11 @@
+package internal
+
+// EmptyBPFContext is the smallest-possible BPF input context to be used for
+// invoking `Program.{Run,Benchmark,Test}`.
+//
+// Programs require a context input buffer of at least 15 bytes. Looking in
+// net/bpf/test_run.c, bpf_test_init() requires that the input is at least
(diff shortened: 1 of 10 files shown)


internal/prog.go
```diff
@@ -0,0 +1,14 @@
+package internal
+
+// Programs require a context input buffer of at least 15 bytes. Looking in
+// net/bpf/test_run.c, bpf_test_init() requires that the input is at least
+// ETH_HLEN (14) bytes. As of Linux commit fd18942 ("bpf: Don't redirect packets
+// with invalid pkt_len"), it also requires the skb to be non-empty after
+// removing the Layer 2 header.
+var emptyBPFContext = make([]byte, 15)
```
lmb, line 8: Since it's internal, just export the slice var directyl?

lmb (approved):

PR #796, opened 2022-09-20
btf: add support for marshaling Type and use it for Program and Map

lmb:
btf: manually marshal lineInfo and funcInfo
    
    Reduce the allocations needed to marshal funcInfo and lineInfo by avoiding
    binary.Write.
    
    The collection loading benchmark is now in an acceptable ballpark again:
    
        name             old time/op    new time/op    delta
        NewCollection-4     948µs ± 9%    1097µs ±14%     ~     (p=0.057 n=4+4)
    
        name             old alloc/op   new alloc/op   delta
        NewCollection-4    31.2kB ± 0%    35.7kB ± 0%  +14.50%  (p=0.029 n=4+4)
    
        name             old allocs/op  new allocs/op  delta
        NewCollection-4       680 ± 0%       747 ± 0%   +9.85%  (p=0.029 n=4+4)

map: avoid MapSpec.Copy
    
    MapSpec has grown over time, so copying it adds up. We only need to
    copy for a small number of map types, so lazily copy the spec only
    when we need to.

map: build BTF on the fly
    
    Use btf.Encoder to write out MapSpec.Key and MapSpec.Value if necessary.
    MapSpec.BTF is now unused, and therefore deprecated like ProgramSpec.BTF.

program: build BTF on the fly
    
    Use btf.Encoder to marshal ext infos embedded in the instruction stream
    on the fly. Care is taken to avoid allocating if no ext infos are present.
    
    As a side effect, ProgramSpec.BTF is now unused by the library. Document
    that is deprecated and point users towards CollectionSpec.Types.

btf: use Encoder to marshal Spec

collection: benchmark loading a collection
    
    Add a benchmark that exercises loading a somewhat complex ELF into the kernel.
    
        name             time/op
        NewCollection-4   908µs ± 2%
    
        name             alloc/op
        NewCollection-4  31.2kB ± 0%
    
        name             allocs/op
        NewCollection-4     680 ± 0%

btf: add Encoder
    
    Allow building BTF wire format from scratch. The most interesting problem is which order
    to output types in. To understand the issue it's best to start with a non-cyclical type:
    
        const long unsigned int
    
    libbpf encodes the type like this:
    
        [1] INT 'long unsigned int' size=8 bits_offset=0 nr_bits=64 encoding=(none)
        [2] CONST '(anon)' type_id=1
    
    Since CONST needs to refer to the long unsigned int somehow, INT is written out
    first and therefore gets ID 1. Afterwards the CONST is written with type_id=1.
    This is exactly the result we get from doing a postorder traversal on the CONST.
    
    Things get more interesting with cyclical types:
    
        struct list_head {
            struct list_head *next, *prev;
        }
    
    libbpf encodes the type like this:
    
        [86] STRUCT 'list_head' size=16 vlen=2
                'next' type_id=88 bits_offset=0
                'prev' type_id=88 bits_offset=64
        ...
        [88] PTR '(anon)' type_id=86
    
    Here STRUCT actually refers to type_id 88, even though that type has not been written
    out yet. Even if we turn things around and encode PTR before STRUCT we'd still have
    the same problem.
    
    As a consequence, we need a way to allocate IDs for types that haven't been marshalled
    yet and we need to ensure that we encode types according to their allocated ID.
    The solution is the pending queue: when allocating a new ID for a type we also push
    that type to the queue. We then always drain the queue before continuing our traversal.
    
    Some benchmark data that compares marshaling Types to just re-encoding
    raw type data as we currently do in Spec.marshal:
    
        name                  time/op
        BuildVmlinux-4        75.7ms ± 8%
        BuildVmlinuxLegacy-4  26.5ms ± 8%
    
        name                  alloc/op
        BuildVmlinux-4        28.6MB ± 0%
        BuildVmlinuxLegacy-4  20.8MB ± 0%
    
        name                  allocs/op
        BuildVmlinux-4          463k ± 0%
        BuildVmlinuxLegacy-4    340k ± 0%
    
    A full rebuild is about 3x slower and uses 1.4x the memory.
    
    Updates #616

btf: add postorderTraversal
    
    Marshaling a Type means writing out the Type and all of its children, the
    children's children, etc. Put another way, a Type may form a cyclic graph
    and we need to visit that graph in some order. For efficiency reasons it's
    better to encode a Type's children before the type itself. In graph lingo
    this is called a preorder traversal.
    
    Add an iterator which implements preorder traversal for Type graphs. I've
    written three to four different implementations before settling on this
    approach.
    
    - Allocations make this a very expensive operation
    - Some common Types like Int or Void don't have any children
    - On the other hand, some types in vmlinux form very large graphs
    - Some types are traversed a lot more often than others
    - Go maps are slow when the key is an interface like Type
    
    As a result, the iterator allows skipping the traversal of a Type and
    avoids allocations in the common case of a skipped root Type or Types
    that have no children. It also avoids map lookups as much as possible.
    
    State is split into three parts:
    
    1. types contains types that need to have their children explored or
       which need to be yielded to the user.
    2. walked contains booleans that indicate whether to explore children
       or to yield.
    3. pushed contains all Type that have been added to types. It prevents
       traversing multiple times.
    
    types and walked could be stored in a single Deque, but this would waste
    memory on padding between Type and bool.


lmb:
@ti-mo based on our review I did the following:
- Unexported Encoder and added MarshalMapKV instead. Since encoder is now not exported I didn't bother renaming encoder.Encode, we can do that later / never ;P
- I reworked encoder.Add to be less confusing. I think I found a middle ground between your idea and my initial implementation. The pending queue is now created directly in `Add()` which is a lot clearer, thanks for prodding me to do this.
- The coverage for `encoder` is OK, only declTag and typeTag are not covered.

collection.go
```diff
@@ -575,6 +536,7 @@ func (cl *collectionLoader) populateMaps() error {
 					if err != nil {
 						return fmt.Errorf("loading program %s, for map %s: %w", objName, mapName, err)
 					}
+					mapSpec = mapSpec.Copy()
```
ti-mo, line 539: This will copy the outer map for every item in `mapSpec.Contents`, which is not great.

An `if len(mapSpec.Contents) != 0 { }` around the existing `mapSpec = mapSpec.Copy()` above should do the trick.

Nice find, otherwise!

ti-mo (changes requested):
Found a small perf regression in `populateMaps()`, but everything else looks amazing! :+1: 

Didn't run into any particularly tricky code, I feel like everything is nicely explained and easy to follow.

Is there more related follow-up work that needs to go in after this? Perhaps we should change the title of the PR to stand out more in the release notes, it looks rather innocuous right now. :wink:

collection.go
lmb: The problem with that is that it copies `.rodata` maps as well, which it would be nice to avoid!

lmb: Solved it a little bit differently, should  be ok now hopefully.

prog.go
```diff
@@ -369,7 +359,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, handles *hand
 
 	err = internal.ErrorWithLog(err, logBuf, truncated)
 	if btfDisabled {
-		return nil, fmt.Errorf("load program: %w (kernel without BTF support)", err)
+		return nil, fmt.Errorf("load program: %w (BTF disabled)", err)
```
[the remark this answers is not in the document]
lmb: I vaguely remember that it confused someone, but I'm drawing a blank on the details. Searching issues and discussions hasn't yielded anything. I'll change it back.

lmb:
> Is there more related follow-up work that needs to go in after this?

I have some optimizations for `map[Type]T` in a branch, but the code in the PR can stand on it's own.

collection.go
ti-mo: I pushed a fixup that uses `canStoreMap()` and `canStoreProgram()` for everything here and removes a level of indentation.

ti-mo (approved):

PR #799, opened 2022-09-21
reopen may uint128 to [16]byte

Benjamin-Yim:
The latest code seems to have changed quite a bit so it was changed to:

[format.go#L185](https://github.com/Benjamin-Yim/ebpf/blob/fix_uint128/btf/format.go#L185)

```golang
...
	case Unsigned, Signed:
		stem := "uint"
		if i.Encoding == Signed {
			stem = "int"
		}
		if i.Size > 8 {
			fmt.Fprintf(&gf.w, "[%d]byte /* %s%d */", i.Size, stem, i.Size*8)
		} else {
			fmt.Fprintf(&gf.w, "%s%d", stem, bits)
		}
```

### DIFF
--- a/btf/format.go
+++ b/btf/format.go
@@ -174,8 +174,6 @@ func (gf *GoFormatter) writeIntLit(i *Int) error {
 			return fmt.Errorf("bool with size %d", i.Size)
 		}
 		gf.w.WriteString("bool")
-	case Signed:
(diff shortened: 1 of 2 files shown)


lmb:
@Benjamin-Yim thank you for your contribution, sorry it took so long to merge.

PR #803, opened 2022-09-26
Semaphore: remove manual Go installation, expedite 'cache restore'

ti-mo:
Some recommendations from Semaphore support.

'sem-version' apparently installs Go, so rely on it instead of isntalling Go manually.

Execute 'cache restore' before 'go install' to avoid tar errors unpacking the cache archive:

```
Unpacking command failed: tar: /home/semaphore/go/pkg/mod/github.com/jonboulle/clockwork@v0.2.2/.github/workflows/ci.yml: Cannot open: File exists
tar: /home/semaphore/go/pkg/mod/github.com/jonboulle/clockwork@v0.2.2/.github/workflows/.editorconfig: Cannot open: File exists
tar: /home/semaphore/go/pkg/mod/github.com/jonboulle/clockwork@v0.2.2/ticker.go: Cannot open: File exists
...
```

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -23,14 +23,13 @@ blocks:
       prologue:
         commands:
           - sudo sh -c 'swapoff -a && fallocate -l 2G /swapfile && chmod 0600 /swapfile && mkswap /swapfile && swapon /swapfile'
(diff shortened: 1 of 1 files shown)


lmb:
Semaphore is really slow in updating go versions unfortunately which is why that hack exists in the first place!

PR #804, opened 2022-09-26
README: update to reflect the project's situation in H2 2022

ti-mo:
- 'Officially' mention 'ebpf-go' as being a thing.
- Dropped the CF/Cilium notice as the project has received frequent and significant contributions from the wider community and was only partially relevant anymore. We no longer need it to convey the legitimacy of the project.
- Explicitly call out GH Discussions, which is has seen significant use so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,12 @@
 
 ![HoneyGopher](.github/images/cilium-ebpf.png)
 
-eBPF is a pure Go library that provides utilities for loading, compiling, and
+ebpf-go is a pure Go library that provides utilities for loading, compiling, and
 debugging eBPF programs. It has minimal external dependencies and is intended to
 be used in long running processes.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,14 +4,12 @@
 
 ![HoneyGopher](.github/images/cilium-ebpf.png)
 
-eBPF is a pure Go library that provides utilities for loading, compiling, and
+ebpf-go is a pure Go library that provides utilities for loading, compiling, and
```
lmb, line 7: Clear as mud? ;D https://pkg.go.dev/github.com/dropbox/goebpf Referring to it as `ebpf-go` when speaking about it would work for me, but I don't want to rename the package `ebpf-go` since that forces import renaming.

ti-mo: Of course not, no import renaming. :)

PR #805, opened 2022-09-27
Prevent pprof from causing BPF verifier livelocks

ti-mo:
Read https://dxuuu.xyz/bpf-go-pprof.html for more context.

If a thread receives a signal while blocked in BPF_PROG_LOAD, the verifier can cooperatively interrupt itself by checking pending signals for the thread and return -EAGAIN from the syscall to request userspace to retry.

When a Go program is built and run with pprof enabled, threads are routinely sent a SIGPROF to make them dump profiling information, which can lead to a runaway reaction if the program takes longer to verify than the interrupt frequency. To prevent this, mask SIGPROF before calling BPF_PROG_LOAD and remove it when done.

@danobi


internal/sys/signals.go
```diff
@@ -0,0 +1,41 @@
...
+		panic(fmt.Errorf("creating signal set: %w", err))
+	}
+}
+
+// MaskProfilerSignal locks the calling goroutine to its underlying OS thread
+// and adds SIGPROF to the thread's signal mask. This prevents pprof from
+// interrupting expensive syscalls like e.g. BPF_PROG_LOAD.
+//
+// Call defer sys.UnmaskProfilerSignal() to reverse the operation.
+func MaskProfilerSignal() {
```
lmb, line 23: Unexport? I hope there won't be users outside of this package.

internal/sys/signals.go
```diff
@@ -0,0 +1,41 @@
...
+	if err := unix.SigsetAdd(&profSet, unix.SIGPROF); err != nil {
+		panic(fmt.Errorf("creating signal set: %w", err))
+	}
+}
+
+// MaskProfilerSignal locks the calling goroutine to its underlying OS thread
+// and adds SIGPROF to the thread's signal mask. This prevents pprof from
+// interrupting expensive syscalls like e.g. BPF_PROG_LOAD.
+//
+// Call defer sys.UnmaskProfilerSignal() to reverse the operation.
```
lmb, line 22:
```suggestion
// You must defer sys.UnmaskProfilerSignal() to reverse the operation.
```

internal/sys/signals.go
```diff
@@ -0,0 +1,41 @@
...
+func MaskProfilerSignal() {
+	runtime.LockOSThread()
+
+	if err := unix.PthreadSigmask(unix.SIG_BLOCK, &profSet, nil); err != nil {
+		panic(fmt.Errorf("masking profiler signal: %w", err))
```
lmb, line 27: Maybe it doesn't matter, but with your proposed API panicking here will never execute `UnlockOSThread()`, because `UnmaskProfilerSignal` hasn't been deferred yet.

Easiest fix is a manual `UnlockOSThread` here I guess?

internal/unix/signals_mips.go
```diff
@@ -0,0 +1,8 @@
+//go:build linux && (mips || mipsle || mips64 || mips64le)
+
+package unix
+
+const (
+	SIG_BLOCK   = 1
```
lmb, line 6: DAMN IT MIPS! :D

internal/cmd/gentypes/main.go
```diff
@@ -24,6 +24,10 @@ const (
 	retFd
 )
 
+var (
+	maskProfiler = "MaskProfilerSignal(); defer UnmaskProfilerSignal();"
```
lmb, line 28: Please see my comment re `BPF()`.

lmb (changes requested):
Very nice, one of those PRs where the change is tiny but the impact is big!

internal/sys/signals.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+// MaskProfilerSignal locks the calling goroutine to its underlying OS thread
+// and adds SIGPROF to the thread's signal mask. This prevents pprof from
+// interrupting expensive syscalls like e.g. BPF_PROG_LOAD.
+//
+// Call defer sys.UnmaskProfilerSignal() to reverse the operation.
+func MaskProfilerSignal() {
+	runtime.LockOSThread()
```
brb, line 24: Just to be on the safe side - we don't create any goroutines in between `runtime.LockOSThread()` and `runtime.UnlockOSThread()`? Asking, as such a goroutine might end up running on a thread with a non-modified sigmask.

ti-mo: No worries, we are aware of this!

internal/sys/signals.go
ti-mo: True, that's the least we could do! An alternative could be `runtime.Goexit()`, as terminating a locked goroutine destroys its underlying thread. That wouldn't allow us to return error or panic, we'd have to log to stdout/err, so no go.

internal/unix/signals.go
```diff
@@ -0,0 +1,41 @@
+package unix
```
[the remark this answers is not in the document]
ti-mo: > once I figure out how to get gopls to pass linux build tags when editing.

In VSCode, I use:
```
    "gopls": {
        "build.buildFlags": [
            "-tags=linux,amd64,integration"
        ],
    },
```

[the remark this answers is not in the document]
ti-mo: > Avoids `signals_other.go`

How, though? Package `sys` also needs to build on macOS, so we'd still need a `signals_other.go` there.

internal/sys/types.go
```diff
@@ -996,6 +996,8 @@ type ProgLoadAttr struct {
 }
 
 func ProgLoad(attr *ProgLoadAttr) (*FD, error) {
+	MaskProfilerSignal()
```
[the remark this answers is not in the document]
ti-mo: Sounds good, hated the change to gentypes anyway.

internal/unix/signals.go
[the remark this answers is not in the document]
lmb: I had forgotten about SIG_BLOCK and friends, sorry. I'll try to submit another CL that add them to x/sys, but let's keep SIG_BLOCK in internal/unix for now.

internal/sys/signals_test.go
```diff
@@ -0,0 +1,61 @@
...
+	// same OS thread as the functions under test. UnlockOSThread needs to be
+	// called as many times as LockOSThread to unlock the goroutine.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	maskProfilerSignal()
+	unmaskProfilerSignal()
+
+	var old unix.Sigset_t
+	if err := unix.PthreadSigmask(0, nil, &old); err != nil {
```
lmb, line 54: Neat!

lmb (approved):
🚢 it!

PR #808, opened 2022-09-28
internal/unix: remove package

lmb:
The package exists so that editing the source on a non-Linux OS doesn't break due to types and functions missing from golang.org/x/sys/unix for that OS. The better solution is to pass gopls the linux tag.


ti-mo:
This might be more of a breaking change than you realize. It'll prevent mac users from running non-ebpf unit tests of all projects that import the lib.

lmb:
Huh! Do you have an example?

lmb:
Seems like a bad idea! Maybe I can come up with something else. The package really bugs me cause it always gets autocompleted wrong, and maintaining is so much boilerplate.

lmb:
@florianl follow up: that's what the change to the README is all about. If you pass a `linux` tag to your `gopls` things kind of work (I tried on macOS). Do you have a setup where that isn't true?

florianl:
passing the `linux` tag to `gopls` just fixes the issue for this repository. But the side effect on other packages might be bigger.

E.g. random pick from https://pkg.go.dev/github.com/cilium/ebpf?tab=importedby - If this branch is used by https://github.com/infobloxopen/ebpf compilation should fail if compiled on `GOOS` like `darwin`. Can't test it bymyself as I don't have access to OS X.

PR #812, opened 2022-10-06
link: Allow kprobe multi to be disabled in kernel

arthurfabre:
Kprobe multi support is gated on CONFIG_FPROBE, without it creating the link fails with ENOTSUP: https://elixir.bootlin.com/linux/latest/source/include/linux/fprobe.h#L56

Unfortunately the standard Debian sid kernel does not have it enabled.

Update the kprobe multi test to handle ENOTSUP, and drop the minimum kernel version check, tests otherwise fail with:

Feature 'bpf_link_kprobe_multi' isn't supported even though kernel v5.19.11 is newer than v5.18

### DIFF
--- a/link/kprobe_multi.go
+++ b/link/kprobe_multi.go
@@ -164,12 +164,16 @@ var haveBPFLinkKprobeMulti = internal.FeatureTest("bpf_link_kprobe_multi", "5.18
 		Count:      1,
 		Syms:       sys.NewStringSlicePointer([]string{"vprintk"}),
 	})
-	if errors.Is(err, unix.EINVAL) {
+	switch {
+	case errors.Is(err, unix.EINVAL):
 		return internal.ErrNotSupported
-	}
-	if err != nil {
+	// If CONFIG_FPROBE isn't set.
+	case errors.Is(err, unix.EOPNOTSUPP):
+		return internal.ErrNotSupported
+	case err != nil:
 		return err
 	}
+
 	fd.Close()
 
 	return nil
--- a/internal/unix/types_linux.go
+++ b/internal/unix/types_linux.go
@@ -9,21 +9,22 @@ import (
 )
 
 const (
-	ENOENT  = linux.ENOENT
-	EEXIST  = linux.EEXIST
-	EAGAIN  = linux.EAGAIN
-	ENOSPC  = linux.ENOSPC
-	EINVAL  = linux.EINVAL
-	EPOLLIN = linux.EPOLLIN
-	EINTR   = linux.EINTR
-	EPERM   = linux.EPERM
-	ESRCH   = linux.ESRCH
-	ENODEV  = linux.ENODEV
-	EBADF   = linux.EBADF
-	E2BIG   = linux.E2BIG
-	EFAULT  = linux.EFAULT
-	EACCES  = linux.EACCES
-	EILSEQ  = linux.EILSEQ
+	ENOENT     = linux.ENOENT
+	EEXIST     = linux.EEXIST
+	EAGAIN     = linux.EAGAIN
+	ENOSPC     = linux.ENOSPC
+	EINVAL     = linux.EINVAL
+	EPOLLIN    = linux.EPOLLIN
+	EINTR      = linux.EINTR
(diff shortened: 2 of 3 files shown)


link/kprobe_multi.go
```diff
@@ -167,9 +167,11 @@ var haveBPFLinkKprobeMulti = internal.FeatureTest("bpf_link_kprobe_multi", "5.18
 	if errors.Is(err, unix.EINVAL) {
 		return internal.ErrNotSupported
 	}
-	if err != nil {
-		return err
+	// If CONFIG_FPROBE isn't set.
+	if errors.Is(err, unix.ENOTSUP) {
+		return internal.ErrNotSupported
 	}
+
```
mmat11, line 174: shouldn't the `if err != nil` check be left here?

arthurfabre: :facepalm: Yes, my bad. Good catch. Fixed.

link/kprobe_multi.go
```diff
@@ -138,7 +138,7 @@ func (kml *kprobeMultiLink) Unpin() error {
 	return fmt.Errorf("unpin kprobe_multi: %w", ErrNotSupported)
 }
 
-var haveBPFLinkKprobeMulti = internal.FeatureTest("bpf_link_kprobe_multi", "5.18", func() error {
+var haveBPFLinkKprobeMulti = internal.FeatureTest("bpf_link_kprobe_multi", "0.0", func() error {
```
lmb, line 141: I think this version number should stay, otherwise our CI will miss this test failing.

lmb (commented):
Thanks! You're hitting this when trying to run the tests on your laptop?

link/kprobe_multi.go
arthurfabre: If we don't remove the version number, tests fail for me with:

```
Feature 'bpf_link_kprobe_multi' isn't supported even though kernel v5.19.11 is newer than v5.18
```

Maybe we can add some magic (`-X` define or a build tag) so the version number checks happen in CI tests, but not locally?

arthurfabre:
Yup I'm hitting it locally on my laptop with the stock debian sid kernel (5.19.11 right now).

link/kprobe_multi.go
lmb: Hmm, I'd like to keep the default (= when running locally) to throw an error when this happens. Otherwise developing a new feature test becomes cumbersome.

How about an environment variable that contains separated feature test names for which to ignore an error result? Basically, in https://github.com/cilium/ebpf/blob/8fceee5ca41b47a59739a14e2f6877886dc89023/internal/testutils/feature.go#L58-L62 check `tb.Name()` against an env var and call `tb.Skip()`. Brownie points if we only split the env var once or so.

This means you have to opt in to ignore the result, but you can stick that variable in your `.bashrc` or something. If your kernel changes and the test automagically isn't skipped anymore.

ti-mo: I also think the version number should stay, Build tags are a hard nack from me, and ldflags are quite cumbersome to specify.

We've also recently adopted envs in `cilium/cilium` to gate test behaviours. All in all, I think it's the lesser evil, because they can be set ambiently and are easy to slap onto any command, as there might be arbitrary tools/scripts in between your shell and the actual test binaries reading the env.

I'd probably go with a single env here that basically trusts the verdict of our `haveBPFXYZ()` functions and doesn't execute the kernel version check. We could add this kind of skip in `testutils.CheckFeatureTest()` or `testutils.checkKernelVersion()`, although the latter might cause more unintended side effects.

As long as the default doesn't change and we don't relax CI.

lmb:
I've dropped the minimum version change, and used `EOPNOTSUPP` instead of `ENOTSUP`. Those are the same on my system but using EOPNOTSUPP is what the kernel uses, sticking to that should be less confusing.

```
lorenz@ubuntares:~$ errno EOPNOTSUPP
EOPNOTSUPP 95 Operation not supported
lorenz@ubuntares:~$ errno ENOTSUP
ENOTSUP 95 Operation not supported
```

lmb (approved):

PR #814, opened 2022-10-07
perf: Add options to create over writable and backward written perf ring buffer

eiffel-fl:
Hi.

First, I hope you are fine and the same for your relatives.

For a specific tool, I need to create overwritable and backward written perf ring buffers.
So, this PR adds two new options to `ReaderOptions`:

1. `WriteBackward` which indicates to the kernel the perf ring buffer should be written backward as it was implemented in this upstream [commit](https://github.com/torvalds/linux/commit/9ecda41acb971ebd07c8fb35faf24005c0baea12).
2. `OverWritable` which means the older events will be over written by newer ones. To do so, we only `mmap` perf ring buffer as read only.

I already wrote a function to read this kind of buffer by invoking a callback for each read event but I cannot ensure this function is totally ready as I did not get any feedback on it for the moment and I still need to modify it as I use `reflect` a lot.
Nonetheless, I think this contribution can be easily split in a first part dealing with the buffer creation and a second one to actually read the buffer.

If you see any way to improve this PR, do not hesitate to share.

### DIFF
--- a/perf/ring.go
+++ b/perf/ring.go
@@ -30,7 +30,7 @@ type perfEventRing struct {
 // From C, we can deduce write_backward is the is the 27th bit.
 const perfBitWriteBackward = gounix.CBitFieldMaskBit27
 
-func newPerfEventRing(cpu, perCPUBuffer, watermark int, writeBackward bool) (*perfEventRing, error) {
+func newPerfEventRing(cpu, perCPUBuffer, watermark int, writeBackward, overWritable bool) (*perfEventRing, error) {
 	if watermark >= perCPUBuffer {
 		return nil, errors.New("watermark must be smaller than perCPUBuffer")
 	}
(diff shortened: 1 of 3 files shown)


lmb (changes requested):
Hi @eiffel-fl, thanks for your PR.
* Why do you introduce two bools to control the behaviour, instead of one?
* Where are the modifications that deal with PERF_EVENT_IOC_PAUSE_OUTPUT?
* Seems like we have to avoid touching the tail pointer, and the interpretation of head also changes. Where are those changes?
* You code doesn't have any tests.

perf/ring.go
```diff
@@ -22,12 +22,19 @@ type perfEventRing struct {
 	*ringReader
 }
 
-func newPerfEventRing(cpu, perCPUBuffer, watermark int) (*perfEventRing, error) {
+
+// In C, struct perf_event_attr has a write_backward field which is a bit in a
+// 64-length bitfield.
+// In Golang, there is a Bits field which 64 bits long.
+// From C, we can deduce write_backward is the is the 27th bit.
+const writeBackwardBit = 1 << 27
```
alban, line 30: Could use:
```
const perfBitWriteBackward = unix.CBitFieldMaskBit27
```
to keep the naming style from the unix package.

See:
https://cs.opensource.google/go/x/sys/+/refs/tags/v0.1.0:unix/ztypes_linux.go;l=1009
https://cs.opensource.google/go/x/sys/+/refs/tags/v0.1.0:unix/ztypes_linux_amd64.go;drc=95e765b1cc43ac521bd4fd501e00774e34401449;l=388

That would automatically benefit from supporting other architectures with other endianness:
```
unix/ztypes_linux_ppc64.go:326: CBitFieldMaskBit26 = 0x2000000000
unix/ztypes_linux_amd64.go:337: CBitFieldMaskBit26 = 0x4000000
```

lmb:
Sounds good. Some guidance based on what you've told me so far:

- Have you considered re-using `Read()` / `ReadInto()` instead of adding a new function with a callback?
- You mentioned reflection for decoding. It'd be better if we could separate reading from the buffer and decoding the buffer from each other.

The less new exported API we need to introduce for these backwards rings the better.

PR #817, opened 2022-10-11
Add IsPinned() to RawLink

boratanrikulu:
Map and Prog have IsPinned() func but RawLink doesn't have it. I added IsPinned() to RawLink. Also seems Unpin method of Rawlink doesn't have a test, I added it too.

### DIFF
--- a/link/link_test.go
+++ b/link/link_test.go
@@ -13,6 +13,8 @@ import (
 	"github.com/cilium/ebpf/internal/sys"
 	"github.com/cilium/ebpf/internal/testutils"
 	"github.com/cilium/ebpf/internal/unix"
+
+	qt "github.com/frankban/quicktest"
 )
 
 func TestRawLink(t *testing.T) {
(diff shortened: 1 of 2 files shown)


link/link_test.go
```diff
@@ -52,6 +54,49 @@ func TestRawLink(t *testing.T) {
 
 func TestRawLinkLoadPinnedWithOptions(t *testing.T) {
 	cgroup, prog := mustCgroupFixtures(t)
+	link, path := newRawLink(t, cgroup, prog)
+	c := qt.New(t)
```
lmb, line 58: Can you change tests to use `qt.Assert` directly? `qt.New` might go away in the future.

link/link_test.go
```diff
@@ -52,6 +54,49 @@ func TestRawLink(t *testing.T) {
...
+	err := link.Unpin()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pinned = link.IsPinned()
+	c.Assert(pinned, qt.IsFalse)
+}
+
+func mustCgroupFixtures(t *testing.T) (*os.File, *ebpf.Program) {
```
lmb, line 90: Can you leave the function where it was? It makes the diff larger than it has to be.

link/link_test.go
```diff
@@ -52,6 +54,49 @@ func TestRawLink(t *testing.T) {
...
+	t.Helper()
+
+	testutils.SkipIfNotSupported(t, haveProgAttach())
+
+	return testutils.CreateCgroup(t), mustLoadProgram(t, ebpf.CGroupSKB, 0, "")
+}
+
+func newRawLink(t *testing.T, cgroup *os.File, prog *ebpf.Program) (*RawLink, string) {
```
lmb, line 98: Nit: `newPinnedRawLink`

link/link_test.go
```diff
@@ -52,6 +54,49 @@ func TestRawLink(t *testing.T) {
 
 func TestRawLinkLoadPinnedWithOptions(t *testing.T) {
 	cgroup, prog := mustCgroupFixtures(t)
+	link, path := newRawLink(t, cgroup, prog)
+	c := qt.New(t)
+
+	pinned := link.IsPinned()
```
lmb, line 60: What's the benefit of introducing a variable for `pinned`?

lmb (changes requested):
Thanks for the PR! Some nits, but otherwise good to go.

boratanrikulu:
@lmb, thank you for the review. I made updates

ti-mo (approved):
Thanks!

PR #818, opened 2022-10-13
ebpf: Allow to pass custom attr to perf ring reader

olsajiri:
Adding support to pass custom perf event attribute object into
ReaderOptions that will be used to create event for perf reader.

Once there's custom attr defined the reader will not parse sample
events but will return the full event array in Record.RawEvent array.

This change allow user to create specific configuration for perf
event. At the same time it's now responsibility of the user to parse
received events.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -45,6 +45,10 @@ type Record struct {
 	// garbage from the ring depending on the input sample's length.
 	RawSample []byte
 
+	// The full perf event sample (including the perfEventHeader),
+	// it is defined when using custom attr via NewReaderFromAttr
+	RawEvent []byte
+
 	// The number of samples which could not be output, since
 	// the ring buffer was full.
 	LostSamples uint64
@@ -53,7 +57,7 @@ type Record struct {
 // Read a record from a reader and tag it as being from the given CPU.
 //
 // buf must be at least perfEventHeaderSize bytes long.
-func readRecord(rd io.Reader, rec *Record, buf []byte) error {
+func readRecord(rd io.Reader, rec *Record, buf []byte, rawEvents bool) error {
 	// Assert that the buffer is large enough.
 	buf = buf[:perfEventHeaderSize]
 	_, err := io.ReadFull(rd, buf)
@@ -69,14 +73,24 @@ func readRecord(rd io.Reader, rec *Record, buf []byte) error {
 		internal.NativeEndian.Uint16(buf[6:8]),
 	}
 
+	if rawEvents {
+		rec.LostSamples = 0
+		rec.RawSample = rec.RawSample[:0]
+
+		rec.RawEvent, err = readRawEvent(rd, int(header.Size), buf, rec.RawEvent)
+		return err
+	}
+
 	switch header.Type {
 	case unix.PERF_RECORD_LOST:
 		rec.RawSample = rec.RawSample[:0]
+		rec.RawEvent = rec.RawEvent[:0]
 		rec.LostSamples, err = readLostRecords(rd)
 		return err
 
 	case unix.PERF_RECORD_SAMPLE:
 		rec.LostSamples = 0
+		rec.RawEvent = rec.RawEvent[:0]
 		// We can reuse buf here because perfEventHeaderSize > perfEventSampleSize.
 		rec.RawSample, err = readRawSample(rd, buf, rec.RawSample)
 		return err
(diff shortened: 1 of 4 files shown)


perf/reader.go
```diff
@@ -45,6 +45,8 @@ type Record struct {
 	// garbage from the ring depending on the input sample's length.
 	RawSample []byte
 
+	RawEvent []byte
```
lmb, line 48: Could we just reuse RawSample instead of introducing a new slice?

If this is required it should be documented.

perf/reader.go
```diff
@@ -131,6 +146,16 @@ func readRawSample(rd io.Reader, buf, sampleBuf []byte) ([]byte, error) {
 	return data, nil
 }
 
+func readRawEvent(rd io.Reader, header perfEventHeader) ([]byte, error) {
```
lmb, line 149: Similar question to `RawSample`, why can't this reuse `readRawSample`?

This function should take a pre-allocated buffer like `readRawSample` so that reusing a record is zero allocation.

perf/reader.go
```diff
@@ -164,6 +191,8 @@ type ReaderOptions struct {
 	// Read will process data. Must be smaller than PerCPUBuffer.
 	// The default is to start processing as soon as data is available.
 	Watermark int
+
+	Attr *unix.PerfEventAttr
```
lmb, line 195: Using `unix.PerfEventAttr` in the API creates an interesting problem. The type is `internal/unix.PerfEventAttr` which on linux is an alias to the real type in `x/sys/unix`. On other platforms (macOS, sigh) this is a stub in `internal/unix`. Trying to fill `ReaderOptions.Attr` will cause a compile time on macOS error, since `x/sys/unix` doesn't contain the type and the user can't access `internal/unix`. Maybe we don't care about this though? It's certainly confusing when looking at generated documentation.

perf/ring.go
```diff
@@ -93,16 +93,22 @@ func createPerfEvent(cpu int, opts ReaderOptions) (int, error) {
...
-	attr := unix.PerfEventAttr{
-		Type:        unix.PERF_TYPE_SOFTWARE,
-		Config:      unix.PERF_COUNT_SW_BPF_OUTPUT,
-		Bits:        unix.PerfBitWatermark,
-		Sample_type: unix.PERF_SAMPLE_RAW,
-		Wakeup:      uint32(watermark),
+	var attr *unix.PerfEventAttr
+
+	if opts.Attr != nil {
+		attr = opts.Attr
```
lmb, line 99: This doesn't enforce that opts.Watermark, etc. are zero if opts.Attr is not nil. PerfEventAttr is really a superset of ReaderOptions, so it doesn't seem right to add it to the latter type. What if w add a new constructor `NewReaderFromAttr(*PerfEventAttr)` instead?

lmb (changes requested):
This is much nicer than having to support a lot of knobs in the library, thanks for reworking your previous approach.

Please investigate getting rid of `Record.RawEvent` and `readRawEvent`, it seems like we should be able to cover both use cases with the same code path?

@ti-mo can you also take a look please? Not sure about exposing `unix.PerfEventAttr` in the API but I also don't have a better idea.

perf/reader.go
olsajiri: yep, good idea

perf/ring.go
olsajiri: right, sounds less confusing

perf/reader.go
olsajiri: actually the readRawSample is reading the bpf sample which is different from what we want to read for raw events - the whole perf event including the header.. I pushed new version with that and added comments for that

perf/reader.go
```diff
@@ -69,14 +73,24 @@ func readRecord(rd io.Reader, rec *Record, buf []byte) error {
 		internal.NativeEndian.Uint16(buf[6:8]),
 	}
 
+	if rawEvents {
```
lmb, line 76: Wouldn't it be nice to check for `header.Type == PERF_RECORD_LOST` and emit `LostSamples` even with raw events?

lmb (changes requested):

olsajiri:
> Okay, my idea to expose `NewReaderFromAttr` this way was bad, sorry. It assumes that a user will never want to pass both Options and an Attr at the same time. Neither is true, I expect that power users (that want Attr) will also want to use something like #814 if it's ever added.
> 
> Instead we need something like `NewReaderFromAttr(..., Options, PerfEventAttr)`, or we resurrect the original idea of putting attr in Options. The key will be that **Options take precedence over Attr**. This means updating to a new version of the reader may break users if they relied on the reader not touching certain parts of attr, but otherwise we can't ever add new features. I think that is acceptable though since this API is already pretty rough. In the future we can try to detect if we are overriding user setting in Attr and return an error instead, which would turn a silent breakage into an error.

right, I'll add Options

perf/reader.go
olsajiri: yea, the separate header should make it easier.. not sure yet if I prefer raw data or the PerfEventHeader,
will check with the rest of the tetragon code first ;-)

same for lost events.. with this new API we actually parse LOST events in upper layer.. I will see how it will look like
to have also LOST events provided, but I'd think it's more straightforward (in already rough API) not to have any exceptions and just return raw events

perf/reader.go
olsajiri: so RawSample actually contains specific data for sample event PERF_COUNT_SW_BPF_OUTPUT:

```
perfEventHeader {
        Type uint32
        Misc uint16
        Size uint16
}
sample (perfEventHeader.Type == 9) {
        Size uint32
        RawSample []byte
}
```

for events other than sample above (like MMAP/COMM/FORK or SAMPLES for events other than bpf-output)
we need to return just the full byte array:

```
perfEventHeader {
        Type uint32
        Misc uint16
        Size uint16
}
event (perfEventHeader.Type == *) {
        RawEvent []byte
}
```

I can return it in the rec.RawSample, but I'd still prefer to add rec.RawEvent
also we need separate read function to do that

perf/reader.go
lmb: > not to have any exceptions and just return raw events

Yeah I'm coming round to this idea.

lmb:
I'm on a plane later today, so I'll give implementing this a shot.

lmb:
As per the discussion in #873 I now believe that a more low level perf api should live in a separate package, which this library could then consume.

PR #820, opened 2022-10-19
btf: support enum64

### LINKED ISSUE #780: btf: support for BTF_KIND_ENUM64

willfindlay:
Implement support for the BTF_KIND_ENUM64 added in Linux 6.0 kernels. As discussed in https://github.com/cilium/ebpf/issues/780, the idea is to reuse the existing Enum type for the new enum64 since it has been extended to fit 64 bit values.

The majority of this implementation has been compied over from the original enum implementation. btfEnum64 layout comes from https://docs.kernel.org/bpf/btf.html?highlight=btf#btf-kind-enum64.

Please review carefully, as I'm not super familiar with the requirements here. Mostly just working off of what's already been done for the original enum type.

Fixes #780

### DIFF
--- a/btf/marshal.go
+++ b/btf/marshal.go
@@ -200,11 +200,17 @@ func (e *encoder) deflateType(typ Type) (err error) {
 		raw.data, err = e.convertMembers(&raw.btfType, v.Members)
 
 	case *Enum:
-		raw.SetKind(kindEnum)
 		raw.SetSize(v.size())
 		raw.SetVlen(len(v.Values))
 		raw.SetSigned(v.Signed)
-		raw.data, err = e.deflateEnumValues(v.Values)
+
+		if v.has64BitValues() {
+			raw.SetKind(kindEnum64)
+			raw.data, err = e.deflateEnum64Values(v.Values)
+		} else {
+			raw.SetKind(kindEnum)
+			raw.data, err = e.deflateEnumValues(v.Values)
+		}
 
 	case *Fwd:
 		raw.SetKind(kindForward)
@@ -310,7 +316,7 @@ func (e *encoder) deflateEnumValues(values []EnumValue) ([]btfEnum, error) {
 		}
 
 		if value.Value > math.MaxUint32 {
-			return nil, fmt.Errorf("value of enum %q exceeds 32 bit", value.Name)
+			return nil, fmt.Errorf("value of enum %q exceeds 32 bits", value.Name)
 		}
 
 		bes = append(bes, btfEnum{
@@ -318,6 +324,25 @@ func (e *encoder) deflateEnumValues(values []EnumValue) ([]btfEnum, error) {
 			uint32(value.Value),
 		})
 	}
+
+	return bes, nil
+}
+
+func (e *encoder) deflateEnum64Values(values []EnumValue) ([]btfEnum64, error) {
+	bes := make([]btfEnum64, 0, len(values))
+	for _, value := range values {
+		nameOff, err := e.strings.Add(value.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		bes = append(bes, btfEnum64{
+			nameOff,
+			uint32(value.Value),
+			uint32(value.Value >> 32),
+		})
+	}
+
 	return bes, nil
 }
 
--- a/btf/types.go
+++ b/btf/types.go
@@ -276,6 +276,21 @@ func (e *Enum) copy() Type {
 	return &cpy
 }
 
+// has64BitValues returns true if the Enum contains a value larger than 32 bits.
+// Kernels before 6.0 have enum values that overrun u32 replaced with zeroes.
+//
+// 64-bit enums have their Enum.Size attributes correctly set to 8, but if we
+// use the size attribute as a heuristic during BTF marshaling, we'll emit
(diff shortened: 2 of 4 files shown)


btf/types.go
```diff
@@ -991,6 +991,20 @@ func inflateRawTypes(rawTypes []rawType, baseTypes types, rawStrings *stringTabl
...
 			typ = tt
 
+		case kindEnum64:
+			rawvals := raw.data.([]btfEnum64)
+			vals := make([]EnumValue, 0, len(rawvals))
+			signed := raw.KindFlag()
+			for i, btfVal := range rawvals {
+				name, err := rawStrings.Lookup(btfVal.NameOff)
+				if err != nil {
+					return nil, fmt.Errorf("get name for enum value %d: %s", i, err)
```
ti-mo, line 1001: Micro-nit to disambiguate errors from these code paths:

```suggestion
					return nil, fmt.Errorf("get name for enum64 value %d: %s", i, err)
```

btf/types.go
```diff
@@ -991,6 +991,20 @@ func inflateRawTypes(rawTypes []rawType, baseTypes types, rawStrings *stringTabl
...
+				}
+				value := (uint64(btfVal.ValHi32) << 32) + uint64(btfVal.ValLo32)
```
ti-mo, line 1003: I'm not sure if the compiler is smart enough to turn this into `hi | lo`, perhaps it's clearer to make this explicit?

ti-mo (commented):
Looks good to me! Could you add some tests to `btf/types_test.go` wherever the existing `Enum` is exercised?

willfindlay:
@ti-mo Thanks, I've fixed up your nits.

> Could you add some tests to btf/types_test.go wherever the existing Enum is exercised?

Since we're reusing the existing enum type I'm not sure that this makes sense. The only difference is the underlying btfEnum64 type and the way it's converted into an Enum type, but I don't see anything related to the raw btf* types in that file... Am I missing something?

lmb:
Thanks for the PR, would you mind if we landed this after #796 is in? It'd mean writing code to serialise out enum64 as well, but it shouldn't be too much work.

willfindlay:
@lmb sure thing, feel free to just ping me here when we can get this moving again.

Unfortunately this is currently preventing Tetragon from running on 6.0.x kernels though so it would be nice to get the fix in asap. However I think I'm one of the only users who has hit this so far (perks of a rolling release distro 😃).

olsajiri:
> @lmb sure thing, feel free to just ping me here when we can get this moving again.
> 
> Unfortunately this is currently preventing Tetragon from running on 6.0.x kernels though so it would be nice to get the fix in asap. However I think I'm one of the only users who has hit this so far (perks of a rolling release distro smiley).

+1 .. tested on tetragon on 6.0 kernel, works nicely ;-) thanks

ti-mo:
We just merged #796. I'll rebase and backfill the necessary marshaling code to this PR. (https://github.com/cilium/ebpf/pull/820#issuecomment-1286975612)

btf/marshal.go
```diff
@@ -200,11 +200,17 @@ func (e *encoder) deflateType(typ Type) (err error) {
 		raw.data, err = e.convertMembers(&raw.btfType, v.Members)
 
 	case *Enum:
-		raw.SetKind(kindEnum)
 		raw.SetSize(v.size())
 		raw.SetVlen(len(v.Values))
 		raw.SetSigned(v.Signed)
-		raw.data, err = e.deflateEnumValues(v.Values)
```
lmb, line 207: Another option would be to return `(any, error)` from deflateEnumValues and handle the enum / 64 details there. This is fine too.

lmb (approved):
Looks great, thanks!

PR #825, opened 2022-10-25
internal: detect if /proc/self/auxv is not readable due to file caps

lmb:
As reported by Daniel Xu, setting capabilities on an executable file will prevent reading /proc/self/auxv since the file is owned by root. The work arounds to this have various trade offs, so the best we can do is tell the user why we failed and hope they check out our discussion or create an issue.

See https://github.com/cilium/ebpf/pull/823


lmb:
cc @danobi

danobi:
LGTM, thanks!

PR #826, opened 2022-10-26
btf: add specialised typeMap

lmb:
This PR improves marshaling vmlinux BTF. I tried to benchmark `NewCollection` as well, but that causes oomkills due to some weird kernel interaction. I've not replaced all occurrences of `map[Type]` with `typeMap`, since it only makes sense in a few places I think.

I think it's also OK to decide that the speedup is not worth it.

btf: add typeMap
    
    Add a specialised map that has a Type as its key. Using an interface as
    map key has two downsides:
    
    1. An interface is two pointers, one for type information the other
       for actual data.
    2. Comparing an interface key is slower than comparing a uint64 key
    
    Replace the interface with typeKey, which is just a wrapper around
    an unsafe pointer. That pointer is usually the data pointer of the
    interface value.
    
    Special care needs to be taken for Void, since it is a zero sized
    type. Instead of using the interface's data pointer we use the type
    information pointer as the key.
    
    This makes lookups roughly 10% faster and saves some memory since
    we only have to store one instead of two pointers. It also
    gracefully deals with Void, which removes some corner cases.
    
    The downside is that keys become meaningless: it's not possible
    to iterate a typeMap and retrieve the Type from the key.
    
        BenchmarkTypeMapGet/typeMap-4       158417788  7.582 ns/op
        BenchmarkTypeMapGet/map[Type]-4     141518359  8.253 ns/op

btf: use typeMap for postorder traversal
    
    name                                      old time/op    new time/op    delta
    PostorderTraversal/single_type-4            18.5ns ± 0%    18.5ns ± 0%     ~     (p=0.514 n=4+4)
    PostorderTraversal/cycle(1)-4                197ns ± 1%     184ns ± 1%   -7.05%  (p=0.029 n=4+4)
    PostorderTraversal/cycle(10)-4              1.03µs ± 0%    0.94µs ± 0%   -8.54%  (p=0.029 n=4+4)
    PostorderTraversal/gov_update_cpu_data-4     923µs ± 0%     829µs ± 0%  -10.21%  (p=0.029 n=4+4)
    
    name                                      old alloc/op   new alloc/op   delta
    PostorderTraversal/single_type-4             0.00B          0.00B          ~     (all equal)
    PostorderTraversal/cycle(1)-4                 264B ± 0%      200B ± 0%  -24.24%  (p=0.029 n=4+4)
    PostorderTraversal/cycle(10)-4                716B ± 0%      515B ± 0%  -28.07%  (p=0.029 n=4+4)
    PostorderTraversal/gov_update_cpu_data-4     364kB ± 0%     211kB ± 0%  -42.15%  (p=0.029 n=4+4)
    
    name                                      old allocs/op  new allocs/op  delta
    PostorderTraversal/single_type-4              0.00           0.00          ~     (all equal)
    PostorderTraversal/cycle(1)-4                 4.00 ± 0%      4.00 ± 0%     ~     (all equal)
    PostorderTraversal/cycle(10)-4                7.00 ± 0%      7.00 ± 0%     ~     (all equal)
    PostorderTraversal/gov_update_cpu_data-4       152 ± 0%       133 ± 0%  -12.79%  (p=0.029 n=4+4)

btf: use typeMap when marshaling BTF
    
    Use typeMap to store allocated IDs. I had to run benchmarks with swap
    and the GC disabled to get reliable timings. I therefore expect the
    real-life impact of this change to be larger than what the benchmark
    indicates.
    
        BuildVmlinux-4    69.9ms ± 0%    66.3ms ± 1%   -5.16%  (p=0.029 n=4+4)
    
        BuildVmlinux-4    37.8MB ± 0%    33.1MB ± 0%  -12.58%  (p=0.029 n=4+4)
    
        BuildVmlinux-4      436k ± 0%      436k ± 0%   -0.01%  (p=0.029 n=4+4)

btf: use typeMap to store TypeID in Spec
    
    Replace map[Type]TypeID with typeMap[TypeID]. The primary benefit is that we
    can drop the explicit check for Void in TypeID() since typeMap correctly
    handles that type.

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -38,7 +38,7 @@ type Spec struct {
 	types []Type
 
 	// Type IDs indexed by type.
-	typeIDs map[Type]TypeID
+	typeIDs typeMap[TypeID]
 
 	// The last allocated type ID.
 	lastTypeID TypeID
@@ -82,14 +82,16 @@ func (h *btfHeader) stringStart() int64 {
 
 // NewSpec creates a Spec containing only Void.
 func NewSpec() *Spec {
-	return &Spec{
+	s := &Spec{
 		[]Type{(*Void)(nil)},
-		map[Type]TypeID{(*Void)(nil): 0},
+		make(typeMap[TypeID]),
 		0,
 		make(map[essentialName][]Type),
 		nil,
 		nil,
 	}
+	s.typeIDs.Set((*Void)(nil), 0)
+	return s
 }
 
 // LoadSpec opens file and calls LoadSpecFromReader on it.
@@ -252,7 +254,7 @@ func loadRawSpec(btf io.ReaderAt, bo binary.ByteOrder,
 	}, nil
 }
 
-func indexTypes(types []Type, typeIDOffset TypeID) (map[Type]TypeID, map[essentialName][]Type, TypeID) {
+func indexTypes(types []Type, typeIDOffset TypeID) (typeMap[TypeID], map[essentialName][]Type, TypeID) {
 	namedTypes := 0
 	for _, typ := range types {
 		if typ.TypeName() != "" {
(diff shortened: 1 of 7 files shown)


lmb:
Closing this for now. The gain really isn't that big.

PR #830, opened 2022-10-28
elf_reader: check if ELF is for BPF data

florianl:

Perfom a basic check on the given ELF. This change adopts the check that is happening also in other loader:

https://github.com/libbpf/libbpf/blob/c97b16d96c8b1771973dba1596477f7d7e9ba201/src/libbpf.c#L1380-L1385

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -51,6 +51,12 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 		return nil, err
 	}
 
+	// Checks if the ELF file is for BPF data.
+	// Old LLVM versions set e_machine to EM_NONE.
+	if f.File.Machine != unix.EM_NONE && f.File.Machine != elf.EM_BPF {
+		return nil, fmt.Errorf("unexpected machine type for BPF ELF: %s", f.File.Machine)
+	}
+
 	var (
 		licenseSection *elf.Section
 		versionSection *elf.Section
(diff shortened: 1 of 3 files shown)


elf_reader.go
```diff
@@ -29,6 +29,12 @@ type elfCode struct {
 	extInfo  *btf.ExtInfos
 }
 
+// isBpfElf checks if the ELF file is for BPF data.
+// Old LLVM versions set e_machine to EM_NONE.
+func isBpfElf(f *elf.File) bool {
```
lmb, line 34: Maybe just move this if into `NewSafeELFFile`? Then `btf.LoadSpecFromReader` also benefits.

Nit: is the extra function necessary?

lmb (changes requested):

elf_reader.go
florianl: I had the very same thought about putting it into `NewSafeELFFile()`. But this function is also used by functions like `vdsoLinuxVersionCode()` which opens an ELF that does not contain BPF data. So it didn't work out nicely.

lmb: Ah good point. New constructor like `NewBPFELFFile()`? Or a helper in `internal`

florianl: Sorry for the delay of my response :pray: 

I have moved the check into `LoadCollectionSpecFromReader()` as it is a one line check and it is the only place where it is used. So maybe `NewBPFELFFile()` might be too much? It think `NewBPFELFFile()` might be worth if the check is used at least twice. please let us know what you think

ti-mo: This is fine, thanks!

ti-mo:
Changed the error string to `unexpected machine type for BPF ELF: %s`. Thanks for the patch!

ti-mo (approved):

PR #832, opened 2022-10-31
asm: add .WithMetadata() for conveniently replacing individual Instructions

ti-mo:
When replacing an Instruction in an Instructions parsed from an ELF, preserving Metadata is important, so this patch adds a convenient method for in-line assignment of Metadata from other instructions.

With func_info and line_info marshaling being driven by Metadata, accidentally stripping a func_info from the first Instruction will result in the following verifier error:

`invalid argument: missing bpf_line_info for func#0`

The verifier rejects line_infos for subprogs that don't start with a func_info.

cc @rgo3

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -354,6 +354,13 @@ func (ins Instruction) Size() uint64 {
 	return uint64(InstructionSize * ins.OpCode.rawInstructions())
 }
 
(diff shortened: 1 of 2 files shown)


rgo3 (approved):

florianl (approved):
great idea :+1: works fine here :)

Issue #284, opened 2021-04-13
Enable cleanup of tracefs entries (kprobe_events)


ti-mo:
As raised by @etherandrius: https://cilium.slack.com/archives/CKC55JL8L/p1618310807065500.

Currently, the group prefix for entries in `kprobe_events` is [hardcoded to `ebpf_`](https://github.com/cilium/ebpf/blob/dccbd5364e5b55091d649c1d2a1de0e1b2ff5522/link/kprobe.go#L166), resulting in identifiers like `ebpf_aabbccdd`. This might prove inflexible in some environments where programs are frequently SIGKILLed and stale entries are left behind in `<tracefs>/kprobe_events`.

The main use case would be enabling cleanup: allowing the caller to evict entries from `kprobe_events` as they see fit, without having to resort to removing all entries starting with `ebpf_`, possibly disrupting other kprobe tools written using this library.

@lmb suggested using `filepath.Base(os.Args[0]) + os.Getpid()` for the group name, which is sufficiently identifiable, and the PID could eventually be correlated using logs etc.

A few caveats:
- Group names allow up to 63 characters, so the name of the binary would need to be truncated for safety.
- PIDs (integers) have a variable-length string representation, so they would need to be padded to a static length.
- The library currently makes the assumption that creating multiple identical probes will succeed because the group identifier is random on each attempt. When using the (static) PID instead, repeated kprobe creations will use the same group and name, preventing multiple attachments to the same symbol from the same program. We're no longer tracking Trace Events as separate entities that can have multiple perf events created from them, so this means we'll always need a random component in the group name, and PID won't do.

While there are drawbacks to all approaches (maybe someone should just backport kprobe PMU to 4.x kernels instead... :sweat_smile:), which one should we pick?

- Kprobe options: `link.Kprobe("printk", prog, &LinkOptions{TraceFSPrefix: "foo"})`
- Package-level prefix override: `link.TraceFSPrefix(prefix string)` (this will mess with tests)
- No override, but always use the truncated, sanitized name of the current binary instead of `ebpf`. This will bite in various other ways, e.g. some users naming their binaries `app` when dropping them in containers.
- Any other ideas?

---

Additionally, should a cleanup method be provided? Something like `link.DeleteTraceFSPrefix("foo")` that removes all probes containing the given prefix? This breaks our abstraction, but we need to provide this if we ever plan on changing the group/prefix format.

Gui774ume:
Hey there 👋🏻 

[...]

PR #842, opened 2022-11-05
Allow specifying the tracefs prefix in options

brycekahle:
In support of #284 this seemed like the most flexible option.

~~Note: If you supply a custom group name, you cannot attach to the same symbol twice.~~ I'm wondering if we should be using the program name instead of a sanitized symbol name, for the portion of the tracefs event name after the group.

Currently:
`<type>:<group>/<sanitized_symbol> <symbol>`
Proposed:
`<type>:<group>/<sanitized_program_name> <symbol>`

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -47,6 +47,10 @@ type KprobeOptions struct {
 	// Deprecated: this setting forces the use of an outdated kernel API and is not portable
 	// across kernel versions.
 	RetprobeMaxActive int
+	// Prefix used for the event name if the kprobe must be attached using tracefs.
+	// The group name will be formatted as `<prefix>_<randomstr>`.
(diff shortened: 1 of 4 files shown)


brycekahle:
@ti-mo @lmb Let me know if you think this would be better as the prefix and still fed into `randomGroup`.

brycekahle:
@lmb ACK. I'll try and update this within the next week or two.

brycekahle:
> My gut feeling would be to make this `TracePrefix` as you mentioned

Can you clarify exactly what you want? I think I'm confused about which specific variant you prefer. I think the only requirement on my end is a predictable prefix, so that we can do cleanup of stale entries.

brycekahle:
@lmb updated to just provide the prefix

lmb (approved):

ti-mo:
Thanks!

Issue #839, opened 2022-11-04
Revisit map BTF and `haveBTF()` behaviour


ti-mo:
Context: as of #796, loading a BTF-defined prog array with type annotations now fails, where it didn't before.

```diff
--- a/testdata/btf_map_init.c
+++ b/testdata/btf_map_init.c
@@ -11,7 +11,8 @@ int __section("socket/tail") tail_1() {
 // Tail call map (program array) initialized with program pointers.
 struct {
        __uint(type, BPF_MAP_TYPE_PROG_ARRAY);
-       __uint(key_size, sizeof(uint32_t));
+       __type(key, uint32_t);
+       __type(value, uint32_t);
        __uint(max_entries, 2);
        __array(values, int());
 } prog_array_init __section(".maps") = {

```

```
λ  ./run-tests.sh 4.19 ./ -run=TestTailCall
...
--- FAIL: TestTailCall (0.01s)
    --- FAIL: TestTailCall/btf_map_init-el.elf (0.00s)
        elf_reader_test.go:557: field TailMain: program tail_main: map prog_array_init: map create: operation not supported
```

This is due to a few factors:
- `haveBTF()` is no longer checked on the map creation happy path. It now asks for forgiveness-not-permission by calling `MarshalMapKV()`,  which is fine, but;
- BPF_LOAD_BTF actually _does_ exist since 4.18 for basic types up until BTF_KIND_RESTRICT. Since the BTF blobs we marshal for maps don't contain any KIND_VARs, the kernel now accepts them. Previously, declaring a BTF map in an ELF would always result in a VAR being emitted, and the whole blob needed to be loaded, which would fail on kernels before 5.1.
- PROG_ARRAY doesn't support map BTF in 4.18/4.19, support was only added later. However, `ProgramArray.supportsBTF()` will return true.

Stepping through `MapSpec.createMap()`:
- `supportsBTF()` returns true and `.Key` and `.Value` are set
- `btf.MarshalMapKV()` succeeds because the kernel supports `KIND_INT` etc. just fine
- `sys.MapCreateAttr.Btf*` fields get populated
- Map creation fails with ENOTSUPP (prog array `map_check_btf` map op is `map_check_no_btf` in 4.19)

I think there are a few things we should change:
- Either remove `haveBTF()` (only really used in `newHandleFromRawBTF`), or:
- Make it load just an int instead of a var. Map BTF is actually supported as of 4.18.
[...]

PR #846, opened 2022-11-07
map: Allow to create ProgramArray map on 4.19 kernel

olsajiri:
The ProgramArray maps are not created properly with latest cilium/ebpf on kernel 4.19. It's probably due to [1] and related changes.

The ProgramArray map on 4.19 kernel do not support BTF and even though BTF could be still loaded properly, the sys.MapCreate will fail with ENOTSUPP with that.

Adding check for this case and try to load the map without BTF.

[1] f7f055bf7f06 ("map: build BTF on the fly")

### DIFF
--- a/map.go
+++ b/map.go
@@ -8,6 +8,7 @@ import (
 	"math/rand"
 	"path/filepath"
 	"reflect"
+	"syscall"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


ti-mo:
Hey, thanks for the patch! I've also hit this and created https://github.com/cilium/ebpf/issues/839 for it.

I'm still weighing our options. We should only clear attr.Btf* and retry map load if the kernel  doesn't support kind_var and kind_datasec. Special-casing ProgArray isn't a good solution because this also fails when loading .bss and .rodata.

Can this wait until next week?

olsajiri:
> Hey, thanks for the patch! I've also hit this and created #839 for it.
> 
> I'm still weighing our options. We should only clear attr.Btf* and retry map load if the kernel doesn't support kind_var and kind_datasec. Special-casing ProgArray isn't a good solution because this also fails when loading .bss and .rodata.

perfect, I was hoping for better fix ;-)

> 
> Can this wait until next week?

sure, @willfindlay workaround-ed it on tetragon side for now

thanks

ti-mo:
Closing in favor of https://github.com/cilium/ebpf/pull/855.

PR #849, opened 2022-11-09
Set Program.name when constructing from file descriptor

aibor:
When a Program is constructed from a file descriptor, only the file descriptor and the program type are set. This results in the name not being present for `.String()`. Because of this the resulting string is different depending on how the Program was constructed.

As the program info is used in `newProgramFromFD()` already anyway, it can be used for setting the name as well.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -400,7 +400,7 @@ func newProgramFromFD(fd *sys.FD) (*Program, error) {
 		return nil, fmt.Errorf("discover program type: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


lmb:
Thanks for your PR! Can you add a small test please?

aibor:
I've added tests for the name and for the type right in the test function for programs by FD. Let me know if it would be better suited somewhere else.

ti-mo (approved):
Thanks!

PR #851, opened 2022-11-10
program: return unwrapped VerifierError

lmb:
Multiple users have been confused as to how to get at the full verifier log. Due to how error wrapping with %w works, callers have to use errors.As to retrieve a VerifierError before being able to format with %+v to get the full log. This confusing exists even though we've tried to document how to deal with it.

Try to make the problem less pressing by returning an unwrapped VerifierError, which the user can then directly format using %+v or similar.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -173,7 +173,7 @@ type Program struct {
 
 // NewProgram creates a new Program.
 //
-// See NewProgramWithOptions for details.
+// See [NewProgramWithOptions] for details.
 func NewProgram(spec *ProgramSpec) (*Program, error) {
 	return NewProgramWithOptions(spec, ProgramOptions{})
 }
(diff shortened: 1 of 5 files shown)


prog.go
```diff
@@ -358,11 +357,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 	}
 
 	truncated := errors.Is(err, unix.ENOSPC) || errors.Is(err2, unix.ENOSPC)
-	err = internal.ErrorWithLog(err, logBuf, truncated)
-	if btfDisabled {
-		return nil, fmt.Errorf("load program: %w (kernel without BTF support)", err)
```
ti-mo, line 363: Okay, as long as we have verifier output, having the `kernel without BTF support` isn't super useful. The verifier output should contain func/lineinfo.

ti-mo (approved):
Cool, this is less involved than I anticipated.

Issue #850, opened 2022-11-09
Error relocating bpf_sk_lookup on Linux 5.19 (Ubuntu 22.10)


SikoJobs:
**Version of cilium/ebpf:**
> v0.9.3 and 5fa0dc4

**Full error message:**
> field SkLookup: program sk_lookup: apply CO-RE relocations: relocate Struct:"bpf_sk_lookup": target Func:"bpf_sk_lookup": target Func:"bpf_sk_lookup": unsized type *btf.Func

**Value of `targets`:**
> [Struct:"bpf_sk_lookup"[fields=10] Func:"bpf_sk_lookup"[static proto=FuncProto]]
https://github.com/cilium/ebpf/blob/5fa0dc4ad0ce18242dc8bcf24ceebf10431f68c8/btf/core.go#L216-L220

**Type of `target`, value of `relo.kind`:**
> `*btf.Func`, `reloFieldByteOffset`
https://github.com/cilium/ebpf/blob/5fa0dc4ad0ce18242dc8bcf24ceebf10431f68c8/btf/core.go#L247-L255

SikoJobs:
I always switch to newer Linux distributions to acquire newer ebpf features, and I haven't seen such error on Linux 5.15 (Ubuntu 22.04) or 5.10 (Debian 11) using previous versions of cilium/ebpf. But on Ubuntu 22.10, I've tried several recent versions of cilium/ebpf and this error always exists (on bpf_sk_lookup only).

Making `targets` above keep the first member only will temporarily solve the problem.

lmb:
Interesting! It looks like CO-RE is trying to relocate a struct against a function, both named `bpf_sk_lookup`. Relocating against a function doesn't really make sense, which is why that unsized error pops up. Seems like [both types exist](https://elixir.bootlin.com/linux/v5.15.77/A/ident/bpf_sk_lookup) in the kernel, going back to 5.15 or earlier.

Are you using`struct bpf_sk_lookup` from bpf.h in your code? Or are you using `vmlinux.h`? It'd be great if you could provide me with a reproducer somewhere.

SikoJobs:
> Interesting! It looks like CO-RE is trying to relocate a struct against a function, both named `bpf_sk_lookup`. Relocating against a function doesn't really make sense, which is why that unsized error pops up. Seems like [both types exist](https://elixir.bootlin.com/linux/v5.15.77/A/ident/bpf_sk_lookup) in the kernel, going back to 5.15 or earlier.

[...]

PR #852, opened 2022-11-11
btf: check for compatibility first when searching for a CO-RE field

lmb:
btf: check for compatibility first when searching for a CO-RE field

    Some kernels have both a struct bpf_sk_lookup and a function
    bpf_sk_lookup() in their BTF. When trying to CO-RE relocate struct
    bpf_sk_lookup we attempt to calculate against the function. This should
    generate errImpossibleRelocation but in fact returns this:

        relocate Struct:"bpf_sk_lookup": target Func:"bpf_sk_lookup": unsized type *btf.Func

    This is because in coreFindField we attempt to adjust the offset
    before checking that local and target are actually compatible.

    Call coreAreMembersCompatible before trying to adjust the offset.

    Fixes https://github.com/cilium/ebpf/issues/850

btf: always include coreKind when CO-RE fails

    Include relo.kind in the error string in coreCalculateFixups to
    ensure we always output this important information.

### DIFF
--- a/btf/core.go
+++ b/btf/core.go
@@ -395,7 +395,7 @@ func coreCalculateFixup(relo *CORERelocation, target Type, targetID TypeID, bo b
 			return poison()
 		}
 		if err != nil {
-			return zero, fmt.Errorf("target %s: %w", target, err)
+			return zero, fmt.Errorf("relocation %s: %w", relo.kind, err)
 		}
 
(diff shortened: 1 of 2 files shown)


lmb:
@SikoJobs can you give this a try and let me know whether it solves your issue please?

SikoJobs:
> @SikoJobs can you give this a try and let me know whether it solves your issue please?

It works. Thank you!

PR #854, opened 2022-11-16
map: Add PinPathFinal bool to MapOptions

olsajiri:
Currently the final pin path is composed of the PinPath
option joined with map name.
    
We have a use case where we want to specify specific
name for the map pin file.
    
Adding PinPathFinal bool that assuments the provided
PinPath option is the final pin path.
    
### DIFF
--- a/map.go
+++ b/map.go
@@ -32,6 +32,7 @@ type MapOptions struct {
 	// Existing maps will be re-used if they are compatible, otherwise an
 	// error is returned.
 	PinPath        string
+	PinPathFinal   bool
 	LoadPinOptions LoadPinOptions
 }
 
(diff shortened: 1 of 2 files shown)


lmb (commented):
This breaks when passing MapOptions.PinPathFinal to NewCollection with multiple maps. Can you describe what your use case is? Maybe there is a different way to solve this.

olsajiri:
we have 2 programs that can be load multiple times and they need to share specific
map in between them, but not with other instances  of such 2 programs loaded,
so we add specific pin file prefix for such couple

one option with the current code is to use specific directory, but would mean bigger
changes and it seemed to me ebpf library should be able to pin under specific name
anyway

current code uses Map::Pin which works just fine, but there's checkCompatibility,
in the path of NewMapWithOptions->newMapWithOptions->PinByName that
I wanted to take advantage of when starting on top of previous pinned maps to
make sure the spec still matches, but it's easy enough to add that to our code if
we can't go with this change

lmb:
> there's checkCompatibility,
> in the path of NewMapWithOptions->newMapWithOptions->PinByName that
> I wanted to take advantage of when starting on top of previous pinned maps to
> make sure the spec still matches

We could expose that function I think. Maybe `Map.IsCompatibleWith(*MapSpec) error` would make more sense for a public API though?

olsajiri:
> ok, that would work.. will make another PR for that, thanks

https://github.com/cilium/ebpf/pull/858

PR #855, opened 2022-11-16
btf: distinguish 'map' and 'program' BTF

ti-mo:
```
bpf(BPF_BTF_LOAD) was merged in 4.18 with support for most basic types
like int, ptr, array, etc. The introduction of BTF marshaling in the lib
came with the removal of haveBTF() from NewHandle(), which used to make
NewHandle() return ErrNotSupported on all upstream kernels before 5.2.

The change from ask-for-permission to ask-for-forgiveness exposed a few
corner cases:

- Some map types like prog arrays don't support BTF k/v until later kernels,
  resulting in an error 'map create: operation not supported' when creating
  maps with .Key and .Value.
- Loading small BTF blobs with just the map k/v types now succeeds on more
  (older) kernels than before. Previously, loading an ELF's full BTF blob
  containing func(proto) and datasecs/vars from other maps would result in
  BTF being disabled for the whole Collection, even if the map in question
  only had simple integers. This obscured the fact that some map types don't
  support BTF k/v on older kernels.
- Support for FUNC(_PROTO) and func_info was added in 5.0, resulting in
  errors loading ELFs built with 'clang -g'.

This commit defines support for 'map BTF' as supporting BTF_KIND_VAR.
haveMapBTF() is called from MarshalMapKV if the map BTF fails to load.
If the kernel does not support BTF_KIND_VAR, ErrNotSupported is returned
and map creation will take place without BTF metadata from the start.

An additional retry mechanism was added in case the BTF blob is accepted,
but map creation fails with ENOTSUPP. Map creation will be attempted without
BTF to address the case of some maps types not yet supporting BTF k/v.

supportsBTF() was removed as it didn't solve the problem correctly.
```

```
As described in the previous commit, this patch defines support for
'program BTF' as successfully loading a BTF blob containing a Func(Proto).
Assume this means func_info is also supported.

In upstream Linux, this was part of the same patchset as func_info. Since
there is little value in supporting only the BTF kind, distro kernels are
most likely to backport the whole patchset.

haveProgBTF() is called before marshaling ext_infos in MarshalExtInfos().
This avoids the following error on kernels between 4.18 and 5.0:

`load ext_infos: invalid argument: [2] Invalid kind:13`
```

Fixes https://github.com/cilium/ebpf/issues/839.


btf/btf_test.go
```diff
@@ -7,12 +7,38 @@ import (
...
+
+func vmlinuxSpec(tb testing.TB) *Spec {
+	tb.Helper()
+
+	vmlinux.Do(func() {
+		vmlinux.spec, vmlinux.err = LoadKernelSpec()
+	})
+	err := vmlinux.err
+
+	if errors.Is(err, ErrNotSupported) {
```
lmb, line 32: Why not `SkipIfNotSupported`?

I'm a bit torn about skipping on ErrNotSupported here, I think it's better to hardcode `SkipOnOldKernel` to coincide with the availability of `/sys/fs/...`. ErrNotSupported is convenient but makes it too easy to skip tests in subtle ways.

btf/btf.go
```diff
@@ -824,7 +820,7 @@ var haveBTF = internal.FeatureTest("BTF", "5.1", func() error {
 })
 
 var haveFuncLinkage = internal.FeatureTest("BTF func linkage", "5.6", func() error {
-	if err := haveBTF(); err != nil {
+	if err := haveMapBTF(); err != nil {
```
lmb: Seems like a weird callchain.

map.go
```diff
@@ -420,34 +420,41 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
...
 		}
 		if errors.Is(err, unix.EINVAL) && attr.MaxEntries == 0 {
 			return nil, fmt.Errorf("map create: %w (MaxEntries may be incorrectly set to zero)", err)
 		}
-		if btfDisabled {
-			return nil, fmt.Errorf("map create: %w (kernel without BTF support)", err)
+		if attr.BtfFd == 0 {
+			return nil, fmt.Errorf("map create without BTF k/v: %w", err)
 		}
-		return nil, fmt.Errorf("map create: %w", err)
```
lmb, line 450: Nit: I'd keep this error message the same, the whole with / without BTF thingie is confusing to users. The happy case should have BTF (?) so if it'd be nice to keep that simple.

Maybe we should get rid of the "with / without BTF" altogether.

btf/btf.go
```diff
@@ -680,10 +680,6 @@ func newHandleFromRawBTF(btf []byte) (*Handle, error) {
 		return &Handle{fd, attr.BtfSize}, nil
 	}
 
-	if err := haveBTF(); err != nil {
```
lmb, line 683: Hmm, not sure we should really yank this. It's nice to get ErrNotSupported as a user, as can be seem from the tests that manually call feature probes to skip the test.

Why not redefine `haveBTF` to mean "when BPF_LOAD_BTF" was added by dropping the KIND_VAR? You can still introduce `haveMapBTF` and use it from MarshalKV.

Optionally: call `haveBTF` from `NewHandle` instead of `newHandleFromRawBTF`. That punts the problem to when we get rid of `Marshal*` at some point.

btf/btf.go
```diff
@@ -819,8 +819,42 @@ var haveMapBTF = internal.FeatureTest("Map BTF (Var/Datasec)", "5.2", func() err
 	return nil
 })
 
+// haveProgBTF attempts to load a BTF blob containing a Func and FuncProto. It
+// is used as a proxy for ext_info (func_info) support, which depends on
+// Func(Proto) by definition.
+var haveProgBTF = internal.FeatureTest("Program BTF (func/line_info)", "5.0", func() error {
```
lmb, line 825: If we keep `haveBTF` this function could / should short circuit that?

btf/ext_info.go
```diff
@@ -139,7 +139,15 @@ var nativeEncoderPool = sync.Pool{
 
 // MarshalExtInfos encodes function and line info embedded in insns into kernel
 // wire format.
+//
+// Returns ErrNotSupported if the kernel doesn't support BTF-associated programs.
 func MarshalExtInfos(insns asm.Instructions) (_ *Handle, funcInfos, lineInfos []byte, _ error) {
+	// Bail out early if the kernel doesn't support Func(Proto). If this is the
+	// case, func_info will also be unsupported.
+	if err := haveProgBTF(); err != nil {
```
lmb, line 147: Here we ask for permission, but for maps we ask for forgiveness. Why not ask for forgiveness in both cases?

Benefit: errors from marshaling ext infos are surfaced on old kernels as well. Downside: we pay the cost on old kernels that don't have support.

btf/marshal.go
```diff
@@ -379,19 +384,28 @@ func (e *encoder) deflateVarSecinfos(vars []VarSecinfo) []btfVarSecinfo {
 // The function is intended for the use of the ebpf package and may be removed
 // at any point in time.
 func MarshalMapKV(key, value Type) (_ *Handle, keyID, valueID TypeID, _ error) {
+	if key == nil && value == nil {
```
lmb, line 387: Don't we have `spec.Key != nil || spec.Value != nil` in `createMap`? We shouldn't treat this as a valid input.

btf/marshal.go
```diff
@@ -379,19 +384,28 @@ func (e *encoder) deflateVarSecinfos(vars []VarSecinfo) []btfVarSecinfo {
...
 	enc := nativeEncoderPool.Get().(*encoder)
 	defer nativeEncoderPool.Put(enc)
 
 	enc.Reset()
 
-	keyID, err := enc.Add(key)
-	if err != nil {
-		return nil, 0, 0, err
+	var err error
+	if key != nil {
```
lmb, line 397: When is it legit to only have a key but not a value, or vice versa? Datasec? Something else?

lmb (changes requested):
Thanks for tackling this, it's really quite tricky.

I think we'll hit this forgiveness / permission problem in the future. So far we've relied on asking for permission, which tends to be simpler but has the risk of false negatives. It also has marginally higher overhead on the happy path / modern kernels. Forgiveness is a bit faster and has no false negatives, but false positives. It also seems more complicated, for example when we have a chain of feature tests calling each other.

We need this PR regardless, but what should we do going forward? Right now we're in this in between spot which might have the worst of all worlds. I can imagine that the cleanest thing would be to call haveBTF uncontionally from `NewHandle` and move the map checks to permission as well.

btf/btf.go
ti-mo: I think you posted this comment on the particular commit where `haveBTF()` was renamed to `haveMapBTF()`. A subsequent commit changes this to `haveProgBTF()`.

map.go
ti-mo: > the whole with / without BTF thingie is confusing to users

Not sure if I agree. If the error message is confusing, we should improve it, but IMO having the context is helpful. I've seen this pop up in bug reports on Slack/GH and it makes it much easier to pinpoint which map create actually failed. If we'd have `map create: %w` in multiple places, we need to start adding printf's to figure out which one actually failed. This is especially true in `newProgramWithOptions()`, which has significant complexity around this part.

btf/btf.go
ti-mo: > Why not redefine haveBTF to mean "when BPF_LOAD_BTF" was added by dropping the KIND_VAR? You can still introduce haveMapBTF and use it from MarshalKV.

That's what I had on my first attempt at this patch (load a simple blob with an int), but it didn't address the problem of a map with int/int kv succeeding to load the BTF blob, but the map type not supporting BTF metadata. I only added the map load retry later.

I just realized removing `haveBTF()` here breaks the API contract of `NewHandle()`, which promises to return ErrNotSupported. I'll reintroduce a `haveBTF()` that loads an int.

> Optionally: call haveBTF from NewHandle

I think `newHandleFromRawBTF` is exactly where it belongs. These probes are actually quite cheap, and the more often they're called, the better they amortize. It's good to have accurate feedback from building blocks like these, that makes the logic on top more solid and less complicated.

btf/btf.go
ti-mo: Ack, makes sense, though doesn't really matter in the grand scheme of things since the results are all cached.

btf/ext_info.go
ti-mo: > Here we ask for permission, but for maps we ask for forgiveness. Why not ask for forgiveness in both cases?

This is intentional, and this question is part of what makes this issue so tricky.

- Traversing insns and marshaling a btf blob and extinfos is significantly more expensive than calling a (potentially cached) `haveProgBTF()`.
- Loading a func(proto) is a perfect oracle for both the kernel supporting the BTF kind, as well as func_infos. Unlike `haveMapBTF()`, which is still no guarantee that the map types supports BTF metadata.

> Benefit: errors from marshaling ext infos are surfaced on old kernels as well.

Not really. `haveProgBTF()` != nil means the kernel doesn't support func_infos (and, by proxy, line_infos), so there's no point in trying to marshal and load any of it. Short-circuiting here looks like the best approach to me.

btf/marshal.go
ti-mo: Sure, this may be overkill, but it's still an exported function. Will remove if you feel strongly about it.

btf/marshal.go
ti-mo: This actually bubbled up while running the selftests. Map `m_queue` in `map_ptr_kern.c` is a `BPF_MAP_TYPE_QUEUE` and only has type info on the value. And `ringbuf_arr` in `test_ringbuf_multi.c` only has the key defined.

btf/btf.go
ti-mo: Done.

map.go
ti-mo: > The happy case should have BTF (?) so if it'd be nice to keep that simple.

Ah, missed that point. Agreed, will do.

btf/btf_test.go
ti-mo: > Why not SkipIfNotSupported?
Good catch!

>  I think it's better to hardcode SkipOnOldKernel to coincide with the availability of /sys/fs/...

Agreed, this shouldn't cause any problems running on dev machines, and in CI we only run upstream kernels so I guess we can make a hard assumption here.

ti-mo: Done, also made `TestParseCurrentKernelBTF` use the cached `vmlinux`.

btf/ext_info.go
lmb: > extinfos is significantly more expensive than calling a (potentially cached) haveProgBTF()

True, however the happy path aka new kernels won't have that overhead.

btf/btf_test.go
```diff
@@ -324,8 +324,8 @@ func TestLoadSpecFromElf(t *testing.T) {
 		}
 
 		t.Run("Handle", func(t *testing.T) {
+			testutils.SkipIfNotSupported(t, haveMapBTF())
```
lmb, line 327: Nit: This change is not required anymore?

lmb (approved):

btf/btf_test.go
ti-mo: It is. `loader-el.elf` contains a VAR, the BTF blob is loaded in full, and `NewHandle()` only returns `ErrNotSupported` on kernels <4.18. On 4.19:

            btf_test.go:330: Can't load BTF: invalid argument: [12] Invalid kind:14 (25 line(s) omitted)

I think we could eventually call `haveBTF`, `haveMapBTF` and `haveProgBTF` (in that order) in `NewHandle()`.. Then we could accurately wrap an error with the corresponding missing BTF support level. Maybe good for a follow-up?

ti-mo: Yep, maybe an alternative could be adding a probe method on each BTF kind. If loading a BTF blob fails, we look for the highest kind in the spec and call its probe method to check if the kernel supports it. (the goal being accurate ErrNotSupported from NewHandle)

PR #861, opened 2022-11-20
btf: fix some split BTF shortcomings

lmb:
Fix some outright bugs and some annoyances with split BTF support. It's still not entirely great, but this is better than nothing.

Requirement for #859.


lmb:
API changes:
```
--- old.txt	2022-11-20 16:21:35.946436516 +0000
+++ new.txt	2022-11-20 16:21:29.014452875 +0000
@@ -21,6 +21,9 @@

 FUNCTIONS

+func ClearCachedKernelSpec()
+    ClearCachedKernelSpec flushes any cached kernel type information.
+
 func LoadSpecAndExtInfosFromReader(rd io.ReaderAt) (*Spec, *ExtInfos, error)
     LoadSpecAndExtInfosFromReader reads from an ELF.

@@ -270,11 +273,9 @@
 func (h *Handle) Info() (*HandleInfo, error)
     Info returns metadata about the handle.

-func (h *Handle) Spec(base *Spec) (*Spec, error)
+func (h *Handle) Spec() (*Spec, error)
     Spec parses the kernel BTF into Go types.

-    base is used to decode split BTF and may be nil.
-
 type HandleInfo struct {
 	// ID of this handle in the kernel. The ID is only valid as long as the
 	// associated handle is kept alive.
```

btf/btf.go
```diff
@@ -263,24 +264,71 @@ func indexTypes(types []Type, typeIDOffset TypeID) (map[Type]TypeID, map[essenti
...
+
+// ClearCachedKernelSpec flushes any cached kernel type information.
+func ClearCachedKernelSpec() {
+	kernelBTF.Lock()
+	defer kernelBTF.Unlock()
+
+	kernelBTF.spec, kernelBTF.fallback = nil, false
+}
+
+func loadKernelSpec() (*Spec, bool, error) {
```
ti-mo, line 286:
```suggestion
func kernelSpec() (*Spec, bool, error) {
```

ti-mo (changes requested):
Ack overall, but this needs to take the changes of https://github.com/cilium/ebpf/pull/855 into account, I think there's some overlap with `btf_spec.go:vmlinux{,Spec}`.

btf/btf.go
```diff
@@ -37,7 +37,7 @@ type Spec struct {
 
 	// All types contained by the spec. For the base type, the position of
 	// a type in the slice is its ID.
-	types types
+	types []Type
```
ti-mo, line 40: How come this was changed?

btf/btf.go
```diff
@@ -500,7 +496,14 @@ func (sw sliceWriter) Write(p []byte) (int, error) {
 // Returns an error wrapping ErrNotFound if a Type with the given ID
 // does not exist in the Spec.
 func (s *Spec) TypeByID(id TypeID) (Type, error) {
```
ti-mo, line 498: From the commit message:

> TypeByID currently assumes that the first ID in s.types is 0, which is not correct when dealing with split BTF.

Could you add some context (in comments, ideally) about why this is the case? I'm not so familiar with split BTF yet and it's been a while since we touched it.

Split BTF IDs start at `len(vmlinux) + 1`, but vmlinux always acts as a base. Can we have a Spec containing only split BTF without a base?

ti-mo (approved):
Thanks! Would like some clarification around the changes to TypeByID.

btf/btf.go
lmb: Because the only function on `types` is ByID, which doesn't take split BTF into account. `types` should be removed but that can happen in a follow up.

btf/btf.go
ti-mo: Right, makes sense.

> Otherwise it's not possible to query a spec to find out whether a type is in in the split or the base part.

Why is this important again? Any real-world application?

btf/btf.go
```diff
@@ -452,7 +496,14 @@ func (sw sliceWriter) Write(p []byte) (int, error) {
...
-	return s.types.ByID(id)
+	firstID := s.firstTypeID()
+	i := int(id - firstID)
+
+	if id < firstID || i >= len(s.types) {
```
ti-mo, line 502: I feel like these should maybe be separate errors? Or, since we know the first and last typeID, we could improve the one here? `expected type ID $id between $firstID and $(firstID + len(s.types))`

btf/btf.go
lmb: Yes, it's required for kmod freplace.

btf/btf.go
lmb: Changed the error message.

PR #863, opened 2022-11-21
perf: Add support for overwritable buffer.

eiffel-fl:
Hi.

This PR supersedes #814.
Compared to conventionnal perf buffer, an overwritable perf buffer never stops writing.
Instead, the oldest data are overwritten by the newest.

As a consequence, the `ReaderOptions` was modified to add a new boolean to indicate if perf buffer is overwritable.
To read these buffers, a new function `ReadCallback()` was introduced. This function reads the buffer from the current head until it wraps the buffer (or there are no more data to read).
For each read record, the callback is called.

This PR also adds test which validates the new function is working.
You can also find an example of this feature usage in [Inspektor Gadget code](https://github.com/inspektor-gadget/inspektor-gadget/blob/8c925785c17200557985136b80d2953955b3e8bd/pkg/gadgets/traceloop/tracer/tracer.go#L283) (not upstreamed at the moment).

If you see any way to improve this contribution, feel free to share.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -53,7 +53,7 @@ type Record struct {
 // Read a record from a reader and tag it as being from the given CPU.
 //
 // buf must be at least perfEventHeaderSize bytes long.
-func readRecord(rd io.Reader, rec *Record, buf []byte) error {
+func readRecord(rd io.Reader, rec *Record, buf []byte, overWritable bool) error {
 	// Assert that the buffer is large enough.
 	buf = buf[:perfEventHeaderSize]
 	_, err := io.ReadFull(rd, buf)
@@ -82,6 +82,9 @@ func readRecord(rd io.Reader, rec *Record, buf []byte) error {
 		return err
 
 	default:
+		if overWritable && header.Type == 0 && header.Misc == 0 && header.Size == 0 {
+			return errEOR
+		}
 		return &unknownEventError{header.Type}
 	}
 }
@@ -111,7 +114,7 @@ type perfEventSample struct {
 func readRawSample(rd io.Reader, buf, sampleBuf []byte) ([]byte, error) {
 	buf = buf[:perfEventSampleSize]
 	if _, err := io.ReadFull(rd, buf); err != nil {
-		return nil, fmt.Errorf("read sample size: %v", err)
+		return nil, fmt.Errorf("read sample size: %w", err)
 	}
 
 	sample := perfEventSample{
@@ -126,7 +129,7 @@ func readRawSample(rd io.Reader, buf, sampleBuf []byte) ([]byte, error) {
 	}
 
 	if _, err := io.ReadFull(rd, data); err != nil {
-		return nil, fmt.Errorf("read sample: %v", err)
+		return nil, fmt.Errorf("read sample: %w", err)
 	}
 	return data, nil
 }
@@ -155,6 +158,9 @@ type Reader struct {
 	// Read calls, which would otherwise need to be interrupted.
 	pauseMu  sync.Mutex
 	pauseFds []int
+
+	paused       bool
+	overWritable bool
 }
 
 // ReaderOptions control the behaviour of the user
@@ -163,7 +169,9 @@ type ReaderOptions struct {
 	// The number of written bytes required in any per CPU buffer before
 	// Read will process data. Must be smaller than PerCPUBuffer.
 	// The default is to start processing as soon as data is available.
-	Watermark int
+	Watermark     int
+	// This perf ring buffer will be written from backward and over writable.
+	OverWritable bool
 }
 
 // NewReader creates a new reader with default options.
@@ -211,7 +219,7 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 	// but doesn't allow using a wildcard like -1 to specify "all CPUs".
 	// Hence we have to create a ring for each CPU.
 	for i := 0; i < nCPU; i++ {
-		ring, err := newPerfEventRing(i, perCPUBuffer, opts.Watermark)
+		ring, err := newPerfEventRing(i, perCPUBuffer, opts.Watermark, opts.OverWritable)
 		if errors.Is(err, unix.ENODEV) {
 			// The requested CPU is currently offline, skip it.
 			rings = append(rings, nil)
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -12,5 +12,6 @@ require (
 	github.com/kr/pretty v0.3.0 // indirect
 	github.com/kr/text v0.2.0 // indirect
 	github.com/rogpeppe/go-internal v1.6.1 // indirect
+	github.com/sirupsen/logrus v1.9.0 // indirect
```
ti-mo, line 15: Drive-by: This is a no go for the library, along with outputting to stdio in general.

ti-mo:
Converted to draft as it looks like some discussion will be needed. I haven't followed the previous discussion in detail, but wouldn't it be better to agree on an approach up front over Slack or a call?

cc @brycekahle

perf/ring.go
```diff
@@ -158,3 +195,78 @@ func (rr *ringReader) Read(p []byte) (int, error) {
...
+	//
+	// read_pos                  previous_read_pos
+	//     |                             |
+	//     V                             V
+	// +---+------+----------+-------+---+--+
+	// |..E|D....D|C........C|B.....B|A..|E.|
+	// +---+------+----------+-------+---+--+
+	// This code snippet was highly inspired by gobpf:
+	// https://github.com/iovisor/gobpf/blob/16120a1bf4d4abc1f9cf37fecfb86009a1631b9f/elf/perf.go#L148
+	if (read & rr.mask) < previousReadMasked {
```
lmb, line 252: This is not necessary, since it's OK for read to return `n < len(p)` (see https://pkg.go.dev/io#Reader). It's the callers responsibility to call Read until the buffer is filled, which we do in `readRawSample`, etc via `io.ReadFull`: https://github.com/cilium/ebpf/blob/8c2039b96a96db0a60f9a897858c5f0fb70eb18f/perf/reader.go#L113-L115

Instead, copy what the current implementation does and truncate the read: https://github.com/cilium/ebpf/blob/8c2039b96a96db0a60f9a897858c5f0fb70eb18f/perf/ring.go#L142-L150

perf/reader_test.go
```diff
@@ -258,6 +260,305 @@ func TestPerfReaderLostSample(t *testing.T) {
...
+	}
+	tb.Cleanup(func() {
+		prog.Close()
+		events.Close()
+	})
+
+	return prog, events
+}
+
+func TestPerfReaderOverWritable(t *testing.T) {
```
lmb, line 344: TestPerfReaderOverWritable and TestPerfReaderOverWritableOverWritten seem to be copies of TestPerfReaderLostSample, which is a really involved test because it needs to trigger a LostEvent record. As a result it's really hard to figure out what you are trying to test here.

The good news is that you don't need all that complexity: an overwritable perf ring will never generate a lost record by definition. Please rewrite your tests based on TestPerfReader instead. It might even be possible to test the new feature without outputSamplesProgOverWritable.

perf/reader_test.go
```diff
@@ -258,6 +260,305 @@ func TestPerfReaderLostSample(t *testing.T) {
 	}
 }
 
+func outputSamplesProgOverWritable(sampleSizes ...int) (*ebpf.Program, *ebpf.Map, error) {
```
lmb, line 263: Why can't you use mustOutputSamplesProg?

perf/ring_test.go
```diff
@@ -61,12 +61,12 @@ func makeRing(size, offset int) *ringReader {
 		Data_size: uint64(len(ring)),
 	}
 
-	return newRingReader(&meta, ring)
+	return newRingReader(&meta, ring, false)
 }
 
 func TestPerfEventRing(t *testing.T) {
```
lmb, line 67: Please cover `reverseReader` here as well.

perf/reader.go
```diff
@@ -321,19 +335,33 @@ func (pr *Reader) ReadInto(rec *Record) error {
 
 	for {
 		if len(pr.epollRings) == 0 {
-			nEvents, err := pr.poller.Wait(pr.epollEvents, pr.deadline)
-			if err != nil {
-				return err
-			}
+			if pr.overWritable {
```
lmb, line 338: In the current implementation, `SetDeadline` has no effect on an overWritable reader.

```go
r.SetDeadline(time.Now().Add(time.Second))
r.Read() // returns io.EOF without blocking
```

I think `SetDeadline` should always work consistently. You would use the reader like so:

```go
// this can happen once after creation of r, it just has to
// be Now() or earlier. See Poller.Wait.
r.SetDeadline(time.Now())
// call repeatedly
r.Read() // returns ErrDeadlineExceeded now
```

This means that we don't need this change and the one in the `err == errEOR` branch.

P.S.: I realize now that the kernel might do funky things wrt epoll and backwards written rings. Still worth a try.

lmb (changes requested):
Thanks for your patience @eiffel-fl! I did some digging in the sources and learned some interesting things:

1. "Overwritable" perf rings have existed for a long time (more than 10 years), they just aren't very useful since you can't figure out where records start.
2. Writing a ring "backwards" fixes this, since the head pointer now always points at a record header.
3. In userspace, we have to deal with partially overwritten records by inspecting the last position we read from and where the kernel currently is

There is an interesting case where the kernel "runs laps" around userspace, which I need to think more about.

perf/reader.go
```diff
@@ -163,7 +165,9 @@ type ReaderOptions struct {
 	// The number of written bytes required in any per CPU buffer before
 	// Read will process data. Must be smaller than PerCPUBuffer.
 	// The default is to start processing as soon as data is available.
-	Watermark int
+	Watermark     int
+	// This perf ring buffer will be written from backward and over writable.
+	OverWritable bool
```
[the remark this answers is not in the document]
lmb: Kernel lingo makes sense for us since we've read the context, less so for other users! I'm fine with keeping the current name, but let's capitalize as `Overwritable` then since it's a single word. See https://en.wiktionary.org/wiki/overwritable

perf/reader.go
```diff
@@ -307,11 +316,16 @@ func (pr *Reader) SetDeadline(t time.Time) {
 // Returns os.ErrDeadlineExceeded if a deadline was set.
 func (pr *Reader) Read() (Record, error) {
 	var r Record
+
 	return r, pr.ReadInto(&r)
 }
 
 // ReadInto is like Read except that it allows reusing Record and associated buffers.
 func (pr *Reader) ReadInto(rec *Record) error {
+	if pr.overWritable && !pr.paused {
```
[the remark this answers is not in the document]
lmb: > But it seems indeed better to have the whole function running under lock.

I think that's not possible since then calling .Pause would also block. Not sure how to fix that yet.

perf/reader.go
lmb: The poller will wait forever only if there is *no deadline set*. If you set a deadline and there is nothing to read, you will get ErrDeadlineExceeded.

perf/ring.go
```diff
@@ -158,3 +195,78 @@ func (rr *ringReader) Read(p []byte) (int, error) {
 
 	return n, nil
 }
+
+func (rr *ringReader) readOverwritable(p []byte) (int, error) {
+	read := rr.tail
+
+	// If size is 0, it means we read all the data avaiable in the buffer and jump
```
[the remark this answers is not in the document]
lmb: > I nonetheless agree this is not perfect to have this here as we can call this function from other path than readRawSample().

That's what I'm trying to say, yes.

> I do not have this issue in Inspektor Gadget as I read the header and the sample in the same function:

You're assuming that the ring is zero filled, which is only true before anything is written to the ring. Later on you'll just have random garbage in there, since the kernel doesn't do any zeroing (afaik).

> I am not sure what would be a good solution here.

I think this case isn't needed since we can handle it like a partially overwritten record (which won't be zero), see below.

perf/reader.go
lmb: Maybe this is helpful: https://man7.org/linux/man-pages/man2/epoll_wait.2.html

> Specifying a
       timeout of -1 causes epoll_wait() to block indefinitely, while
       specifying a timeout equal to zero cause epoll_wait() to return
       immediately, even if no events are available.

PR #864, opened 2022-11-23
btf: Remove deprecated {Map,Program}Spec.BTF field

ti-mo:
Setting ProgramSpec.BTF to nil was a convenient way of making prog loads execute without func/lineinfos with the introduction of Instruction.Metadata.

For example, in case instructions are added at the start of Instructions, the caller would need to make sure to move the Func to the 'new' first instruction to set up a valid function boundary. Clearing ProgramSpec.BTF was a convenient hack to load without BTF if your program type didn't need it.

Unfortunately, since adding full BTF marshaling, the {Map,Program}Spec.BTF field was no longer checked while loading BTF. Callers nilling the field had their code happily compile and fall over loading a program in production.

This commit removes these .BTF fields to make this a compile-time error.

cc @willfindlay

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -442,10 +442,6 @@ func (cl *collectionLoader) loadMap(mapName string) (*Map, error) {
 		return nil, fmt.Errorf("missing map %s", mapName)
 	}
 
-	if mapSpec.BTF != nil && cl.coll.Types != mapSpec.BTF {
-		return nil, fmt.Errorf("map %s: BTF doesn't match collection", mapName)
-	}
-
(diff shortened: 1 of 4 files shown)


lmb (approved):

ti-mo:
> Do we have the APIs in place to support stripping BTF so that coverbee etc. can continue to work?

Not really, but I'm not sure what a good solution would look like. I've added `Instruction.WithMetadata` recently to make moving metadata between instructions much easier; the idea being that it's better to fix up a program than to disable BTF. After all, some program types need funcinfo.

PR #865, opened 2022-11-23
bpf2go: write dependencies to temporary file to support Windows

junjiexing:
I want to develop Android eBPF programs on Windows, but found that bpf2go uses `/dev/fd/3` to accept dependencies output, which is not supported on Windows. This commit modified it to use temporary files to accept dependencies.

### DIFF
--- a/cmd/bpf2go/compile.go
+++ b/cmd/bpf2go/compile.go
@@ -72,7 +72,7 @@ func compile(args compileArgs) error {
 
 	var depFile *os.File
 	if args.dep != nil {
-		depFile, err = os.CreateTemp(os.TempDir(), "bpf2go")
+		depFile, err = os.CreateTemp("", "bpf2go")
 		if err != nil {
 			return err
 		}
--- a/cmd/bpf2go/compile.go
+++ b/cmd/bpf2go/compile.go
@@ -76,6 +76,7 @@ func compile(args compileArgs) error {
 		if err != nil {
 			return err
 		}
+		defer depFile.Close()
 		defer os.Remove(depFile.Name())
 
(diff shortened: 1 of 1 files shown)


cmd/bpf2go/compile.go
```diff
@@ -70,45 +70,35 @@ func compile(args compileArgs) error {
 	)
 	cmd.Dir = args.dir
 
-	var depRd, depWr *os.File
+	var depFile *os.File
 	if args.dep != nil {
-		depRd, depWr, err = os.Pipe()
+		depFile, err = os.CreateTemp(os.TempDir(), "bpf2go")
```
lmb, line 75:
```suggestion
		depFile, err = os.CreateTemp("", "bpf2go")
```

lmb (approved):
One small nit. Please be aware that this doesn't mean we won't break Windows support going forward, since none of us use Windows to develop and we don't use it in CI.

cmd/bpf2go/compile.go
```diff
@@ -70,45 +70,35 @@ func compile(args compileArgs) error {
...
+	var depFile *os.File
 	if args.dep != nil {
-		depRd, depWr, err = os.Pipe()
+		depFile, err = os.CreateTemp("", "bpf2go")
 		if err != nil {
 			return err
 		}
-		defer depRd.Close()
-		defer depWr.Close()
+		defer os.Remove(depFile.Name())
```
ti-mo, line 79: This is missing a `defer depFile.Close()`.

ti-mo (dismissed):

lmb:
Thanks!

PR #867, opened 2022-11-25
link: add QueryPrograms API

rgo3:
This PR adds a wrapper around `BPF_PROG_QUERY` and with this API it's possible to query cgroups, network namespaces or LIRC2 devices for attached programs.

Signed-off-by: Robin Gögge <r.goegge@isovalent.com>

### DIFF
--- a/link/query.go
+++ b/link/query.go
@@ -1,0 +1,63 @@
+package link
+
+import (
+	"fmt"
+	"os"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+// QueryOptions defines additional parameters when querying for programs.
+type QueryOptions struct {
+	// Path can be a path to a cgroup, netns or LIRC2 device
+	Path string
+	// Attach specifies the AttachType of the programs queried for
+	Attach ebpf.AttachType
+	// QueryFlags are flags for BPF_PROG_QUERY, e.g. BPF_F_QUERY_EFFECTIVE
+	QueryFlags uint32
+}
+
+// QueryPrograms retrieves ProgramIDs associated with the AttachType.
+//
+// It only returns IDs of programs that were attached using PROG_ATTACH and not bpf_link.
+// Returns (nil, nil) if there are no programs attached to the queried kernel resource.
+// Calling QueryPrograms on a kernel missing PROG_QUERY will result in ErrNotSupported.
+func QueryPrograms(opts QueryOptions) ([]ebpf.ProgramID, error) {
+	if haveProgQuery() != nil {
+		return nil, fmt.Errorf("can't query program IDs: %w", ErrNotSupported)
+	}
+
+	f, err := os.Open(opts.Path)
+	if err != nil {
+		return nil, fmt.Errorf("can't open file: %s", err)
+	}
+	defer f.Close()
+
+	// query the number of programs to allocate correct slice size
+	attr := sys.ProgQueryAttr{
+		TargetFd:   uint32(f.Fd()),
+		AttachType: sys.AttachType(opts.Attach),
+		QueryFlags: opts.QueryFlags,
+	}
+	if err := sys.ProgQuery(&attr); err != nil {
+		return nil, fmt.Errorf("can't query program count: %w", err)
+	}
+
+	// return nil if no progs are attached
+	if attr.ProgCount == 0 {
+		return nil, nil
+	}
+
+	// we have at least one prog, so we query again
+	progIds := make([]ebpf.ProgramID, attr.ProgCount)
+	attr.ProgIds = sys.NewPointer(unsafe.Pointer(&progIds[0]))
+	attr.ProgCount = uint32(len(progIds))
+	if err := sys.ProgQuery(&attr); err != nil {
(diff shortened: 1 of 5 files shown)


link/query.go
```diff
@@ -0,0 +1,49 @@
...
+// QueryPrograms retrieves ProgramIDs associated with the AttachType
+// from a given kernel resource, e.g. a cgroup, netns or LIRC2 device.
+func QueryPrograms(opts QueryOptions) ([]ebpf.ProgramID, error) {
+	if haveProgQuery() != nil {
+		return nil, fmt.Errorf("can't query program IDs: %w", ErrNotSupported)
+	}
+
+	f, err := os.Open(opts.Path)
+	if err != nil {
+		return nil, fmt.Errorf("can't open cgroup: %s", err)
```
ti-mo, line 28: This is a leftover from cgroup-specific prog_query.

link/query.go
```diff
@@ -0,0 +1,63 @@
...
+// QueryOptions defines additional parametes when querying for programs
+// QueryOptions.Path can be a path to a cgroup, netns or LIRC2 device
+type QueryOptions struct {
+	Path        string
+	AttachType  ebpf.AttachType
+	AttachFlags ebpf.AttachFlags
+	QueryFlags  uint32
+}
+
+// QueryPrograms retrieves ProgramIDs associated with the AttachType
```
lmb, line 21: This only works for things that go via RawAttachProgram / BPF_PROG_ATTACH, right? If yes that would be good to document.

link/query_test.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+
+			progInfo, err := prog.Info()
+			if err != nil {
+				t.Fatal("Can't get program info:", err)
+			}
+
+			progId, ok := progInfo.ID()
+			if !ok {
+				t.Skip("Program ID not supported")
```
lmb, line 30: Nit: doesn't PROG_QUERY imply IDs being present? Probably fine to bail out here.

lmb (changes requested):
Looks good to me, just a few nits. Thanks!

link/query.go
```diff
@@ -0,0 +1,63 @@
...
+	f, err := os.Open(opts.Path)
+	if err != nil {
+		return nil, fmt.Errorf("can't open file: %s", err)
+	}
+	defer f.Close()
+
+	// query the number of programs to allocate correct slice size
+	attr := sys.ProgQueryAttr{
+		TargetFd:    uint32(f.Fd()),
+		AttachType:  sys.AttachType(opts.AttachType),
```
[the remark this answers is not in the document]
rgo3: I tried this but ran into an error with the `String()` function on `ebpf.AttachType`: `cannot define new methods on non-local type sys.AttachType`. So not quite sure what I need to do to make an alias work here.

link/query_test.go
rgo3: It does indeed. Should I just not check the `ok` var here? `ID` is available from 4.13 and we skip this test until 4.15 with the feature test anyway.

link/query.go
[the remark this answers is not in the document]
ti-mo: Did you try `type AttachType sys.AttachType` or ``type AttachType = sys.AttachType`? My guess is on the former.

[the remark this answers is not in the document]
rgo3: I tried `type AttachType = sys.AttachType`. I think a regular type declaration as `type AttachType  sys.AttachType` wouldn't get rid of all the `sys.AttachType(...)` code.

link/query_test.go
lmb: yes, drop the check

lmb (approved):
Looks nice, thanks!

link/query.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	defer f.Close()
+
+	// query the number of programs to allocate correct slice size
+	attr := sys.ProgQueryAttr{
+		TargetFd:   uint32(f.Fd()),
+		AttachType: sys.AttachType(opts.Attach),
+		QueryFlags: opts.QueryFlags,
+	}
+	err = sys.ProgQuery(&attr)
```
ti-mo, line 44: `if err := sys.ProgQuery(); err != nil { ... }`

link/query.go
```diff
@@ -0,0 +1,65 @@
...
+	if attr.ProgCount == 0 {
+		return nil, nil
+	}
+
+	// we have at least one prog, so we query again
+	progIds := make([]ebpf.ProgramID, attr.ProgCount)
+	attr.ProgIds = sys.NewPointer(unsafe.Pointer(&progIds[0]))
+	attr.ProgCount = uint32(len(progIds))
+	err = sys.ProgQuery(&attr)
+	if err != nil {
```
ti-mo, line 59: Same as above.

link/query_test.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			progId, _ := progInfo.ID()
+
+			for _, id := range ids {
+				if id == progId {
+					return
+				}
+			}
+			t.Fatal("Can't find program ID in query")
```
ti-mo, line 35: Nit: printing the relevant results in test output is quite convenient, e.g.:

```suggestion
			t.Fatalf("Can't find program ID %d in query result: %v", progId, ids)
```

Issue #698, opened 2022-06-03
Unable to use `__kconfig`


eiffel-fl:
**Describe the bug**
With `cilium/ebpf`, it is not possible reference kconfig variables like `LINUX_KERNEL_VERSION`.

**To Reproduce**
Steps to reproduce the behavior. Please include:

```sh
# Try to bpf2go a file which contains the following line:
# extern int LINUX_KERNEL_VERSION __kconfig;
foo_bpfel.go: can't load BPF from ELF: load BTF: reference to .kconfig: not supported
```

**Expected behavior**
We should be able to reference kconfig variables.

lmb:
At a very high level:
1. You need a parser for kconfig style files.
2. You need to extract the Datasec patching logic from `CollectionSpec.RewriteConstants` into a separate function.
3. You need to modify `elfCode.loadDataSections` to add a special case for `.kconfig` sections. In that special case you will have to iterate the variables in the section and patch the correct values from (1) into the raw `data`.
4. You need to remove the block on `.kconfig` from `fixupDatasec`

ti-mo:
@eiffel-fl This is `internal.KernelVersion()`.

lmb:
Yes, that's the idea. You can probably resurrect [patchValues](https://github.com/cilium/ebpf/commit/1ff15d21014336f653b09fe3842977f01c5945d2#diff-5ed7da7003773d8e0a4dca00d4fc08fc5582695eefe22a400db15edfd5aab840L1285-L1338) for step 2.

> I think I will give a try to implement the whole __kconfig parsing.

I think Timo has a point, can we land this in two steps? First LINUX_VERSION_CODE, followed by full kconfig support later?

lmb:
From https://facebookmicrosites.github.io/bpf/blog/2020/02/19/bpf-portability-and-co-re.html#dealing-with-kernel-version-and-configuration-differences

```C
extern u32 LINUX_KERNEL_VERSION __kconfig;
extern u32 CONFIG_HZ __kconfig;
```

`__kconfig` is from https://elixir.bootlin.com/linux/v5.19.1/source/tools/lib/bpf/bpf_helpers.h#L174:

```C
#define __kconfig __attribute__((section(".kconfig")))
```

So there should be a `.kconfig` section with a single VAR that has extern linkage. It seems like that isn't the case for you?
 Probably the first thing that needs to be fixed.

> Then another problem that I met is that the BPF loader complains about extern linkage:

[...]

PR #869, opened 2022-11-29
Add initial __kconfig support with LINUX_KERNEL_VERSION.

eiffel-fl:
Hi.

I tested it with some basic eBPF program I wrote:

```bash
root@81d71157c164:/mnt/cilium-ebpf-test-kconfig# more bpf/test.bpf.c 
// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
#include "vmlinux.h"
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_core_read.h>

extern int LINUX_KERNEL_VERSION __kconfig;

SEC("tp_btf/block_rq_insert")
int insert(u64 *ctx)
{
        bpf_printk("LINUX_KERNEL_VERSION: %d\n", LINUX_KERNEL_VERSION);

        return 0;
}

char LICENSE[] SEC("license") = "GPL";
root@81d71157c164:/mnt/ebpf# go run ./...
# Switch to other terminal
$ sudo cat /sys/kernel/debug/tracing/trace_pipe                                                               
 .../2-579     [006] d.... 11213.059870: bpf_trace_printk: LINUX_KERNEL_VERSION: 331584
```

Sadly, I was not able to make the added `TestLinuxKernelVersion()` works because the ELF misses BTF and I need `btfSpec` to not be `nil` to check data section.
I think I miss some steps in the ELF creation.


elf_reader.go
```diff
@@ -132,6 +133,60 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 		return nil, fmt.Errorf("load maps: %w", err)
 	}
 
+	// It is possible for btfSpec to be nil if it was not found, so better to
+	// check this here to avoid panic.
+	if ec.btf != nil {
+		t, err := ec.btf.AnyTypeByName(".kconfig")
+		if err == nil {
```
ti-mo, line 140: This is more or less how I would do this:

```go
var ds *btf.Datasec
if err := ec.btf.TypeByName(".kconfig", &ds); err != nil && !errors.Is(err, btf.ErrNotFound) {
	return nil, err
}

data := make([]byte, 0, ds.Size)
for _, vsi := range ds.Vars {
	v := vsi.Type

	s, err := btf.Alignof(v)
	if err != nil {
		return nil, fmt.Errorf("getting size of kconfig variable %s: %w", v.TypeName(), err)
	}
	d := make([]byte, s)

	switch n := v.TypeName(); n {
	case "LINUX_KERNEL_VERSION":
		if s < 4 {
			return nil, fmt.Errorf("kconfig variable %s must be at least u32, got %d", n, s)
		}
		kv, err := internal.KernelVersion()
		if err != nil {
			return nil, fmt.Errorf("getting kernel version: %w", err)
		}
		internal.NativeEndian.PutUint32(d, kv.Kernel())

	default:
		return nil, fmt.Errorf("unsupported kconfig variable: %s", n)
	}

	data = append(data, d...)
}
```

ti-mo (changes requested):
Gave this a first round. Running the tests doesn't work for me.

btf/btf.go
```diff
@@ -539,6 +542,12 @@ func (s *Spec) AnyTypeByName(name string) (Type, error) {
 	return types[0], nil
 }
 
+// SetTypeByName set a type by its name in the given spec.
+func (s *Spec) SetTypeByName(name string, typ Type) {
```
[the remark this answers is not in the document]
ti-mo: No need, this method is not necessary.

testdata/linux_kernel_version.c
```diff
@@ -0,0 +1,9 @@
+#include "common.h"
```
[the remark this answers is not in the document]
ti-mo: `strip` doesn't throw away the .BTF section and friends, and clearly `loader.c` ELFs carry BTF info:
```
λ  ebpf francis/ebpf-kconfig ✓ bpftool btf dump file testdata/loader-clang-14-el.elf
[1] PTR '(anon)' type_id=3
[2] INT 'int' size=4 bits_offset=0 nr_bits=32 encoding=SIGNED
...
```

elf_reader.go
```diff
@@ -27,6 +27,7 @@ type elfCode struct {
 	version  uint32
 	btf      *btf.Spec
 	extInfo  *btf.ExtInfos
+	maps     *map[string]*MapSpec
```
ti-mo, line 30:
```suggestion
	maps     map[string]*MapSpec
```
Go maps are already pointer types and nil by default, no need to take out pointers to them.

elf_reader.go
```diff
@@ -140,6 +141,12 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 		return nil, fmt.Errorf("load data sections: %w", err)
 	}
 
+	ec.maps = &maps
```
ti-mo, line 144: Please factor `maps` out from being a parameter on the methods above (in a separate commit) and initialize it when allocating the `elfCode` above. This is a no go, sorry.

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		return fmt.Errorf("elfCode map of MapSpec should not be nil")
+	}
+
+	return ec.populateKconfigSection()
+}
+
+func (ec *elfCode) populateKconfigSection() error {
+	mapSpec, ok := (*ec.maps)[".kconfig"]
+	if !ok {
+		// If .kconfig map does not exist it just means the code does not have any
```
ti-mo, line 1133: These kinds of comments are useful for holding your train of thought, but are a net negative on the clarity of the code, please avoid them.

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		// variable inside .kconfig data section.
+		return nil
+	}
+
+	ds, ok := mapSpec.Value.(*btf.Datasec)
+	if !ok {
```
ti-mo, line 1139: The code in `loadDataSections` already guarantees this is a ds, please drop the check to lower the verbosity.

elf_reader.go
```diff
@@ -575,6 +582,26 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 			return fmt.Errorf("asm relocation: %s: unsupported type %s", name, typ)
 		}
 
+		// If we arrive here, it can possibly be because we are dealing with
+		// .kconfig.
+		if mapSpec, ok := (*ec.maps)[".kconfig"]; ok {
+			if ds, ok := mapSpec.Value.(*btf.Datasec); ok {
+				for _, varSecInfo := range ds.Vars {
+					variable, ok := varSecInfo.Type.(*btf.Var)
```
ti-mo, line 590: You already perform this check in `populateDataSections()`. `v := vsi.Type.(*btf.Var)` is fine here.

elf_reader.go
```diff
@@ -575,6 +582,26 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 			return fmt.Errorf("asm relocation: %s: unsupported type %s", name, typ)
 		}
 
+		// If we arrive here, it can possibly be because we are dealing with
+		// .kconfig.
+		if mapSpec, ok := (*ec.maps)[".kconfig"]; ok {
+			if ds, ok := mapSpec.Value.(*btf.Datasec); ok {
+				for _, varSecInfo := range ds.Vars {
```
ti-mo, line 589: For consistency with the population counterpart:
```suggestion
				for _, vsi := range m.Value.(*btf.Datasec).Vars {
```
Also, m.Value is guaranteed to be a ds by preceding code.

elf_reader.go
```diff
@@ -575,6 +582,26 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 			return fmt.Errorf("asm relocation: %s: unsupported type %s", name, typ)
 		}
 
+		// If we arrive here, it can possibly be because we are dealing with
+		// .kconfig.
+		if mapSpec, ok := (*ec.maps)[".kconfig"]; ok {
```
ti-mo, line 587:
```suggestion
		if m := ec.maps[".kconfig"]; m != nil {
```

elf_reader.go
```diff
@@ -575,6 +582,26 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 			return fmt.Errorf("asm relocation: %s: unsupported type %s", name, typ)
 		}
 
+		// If we arrive here, it can possibly be because we are dealing with
```
ti-mo, line 585: This comment is not so useful.

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+			Value:      ds,
+			ValueSize:  ds.Size,
+		}
+	}
+
+	return nil
+}
+
+func (ec *elfCode) populateDataSections() error {
+	if ec.maps == nil {
```
ti-mo, line 1123: Not a useful check.

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		return fmt.Errorf("elfCode map of MapSpec should not be nil")
+	}
+
+	return ec.populateKconfigSection()
```
ti-mo, line 1127:
```suggestion
	if err := ec.populateKconfigSection(); err != nil {
		return fmt.Errorf(".kconfig: %w", err)
	}
	
	return nil
```

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+	ds, ok := mapSpec.Value.(*btf.Datasec)
+	if !ok {
+		return fmt.Errorf(".kconfig value must be of type btf.Datasec")
+	}
+
+	data := make([]byte, 0, ds.Size)
+	for _, vsi := range ds.Vars {
+		v, ok := vsi.Type.(*btf.Var)
+		if !ok {
+			return fmt.Errorf("info should point to a variable")
```
ti-mo, line 1147: This will make debugging easier on a production build in the wild:
```suggestion
			return fmt.Errorf("variable %s: unexpected type %T", vsi.Type.TypeName(), vsi.Type)
```

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		}
+
+		s, err := btf.Alignof(v.Type)
+		if err != nil {
+			return fmt.Errorf("getting size of kconfig variable %s: %w", v.TypeName(), err)
```
ti-mo, line 1152: For consistency with the above:
```suggestion
			return fmt.Errorf("variable %s: getting size: %w", v.Name, err)
```

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		}
+		d := make([]byte, s)
+
+		switch n := v.TypeName(); n {
+		case "LINUX_KERNEL_VERSION":
+			if s < 4 {
+				return fmt.Errorf("kconfig variable %s must be at least u32, got %d", n, s)
```
ti-mo, line 1159: Error wrapping will already annotate this with `.kconfig`.
```suggestion
				return fmt.Errorf("variable %s must be at least u32, got %d", n, s)
```

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+			}
+			kv, err := internal.KernelVersion()
+			if err != nil {
+				return fmt.Errorf("getting kernel version: %w", err)
+			}
+			internal.NativeEndian.PutUint32(d, kv.Kernel())
+
+		default:
+			return fmt.Errorf("unsupported kconfig variable: %s", n)
```
ti-mo, line 1168:
```suggestion
			return fmt.Errorf("unsupported variable: %s", n)
```

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		// We need to change Linkage for extern to global to avoid problems while
+		// marshaling the map value.
+		v.Linkage = btf.GlobalVar
+
+		data = append(data, d...)
+	}
+
+	length := uint32(len(data))
+
+	// We need to update DataSec size to avoid problems while marshaling the map
```
ti-mo, line 1180: Which problems?

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+		}
+
+		// We need to change Linkage for extern to global to avoid problems while
+		// marshaling the map value.
+		v.Linkage = btf.GlobalVar
+
+		data = append(data, d...)
+	}
+
+	length := uint32(len(data))
```
ti-mo, line 1178: This variable is only used once.

elf_reader.go
```diff
@@ -1067,6 +1094,96 @@ func (ec *elfCode) loadDataSections(maps map[string]*MapSpec) error {
...
+
+	// We need to update DataSec size to avoid problems while marshaling the map
+	// value.
+	ds.Size = uint32(len(data))
+
+	mapSpec.ValueSize = length
```
ti-mo, line 1184: Let's not initialize ValueSize in `loadDataSections` if we need to fix it up here anyway.

ti-mo (changes requested):

brycekahle:
@lmb @ti-mo We are interested in getting this over the finish line too, so if I can help, please let me know.

ti-mo:
> I did not abandon this work, I was just waiting for some feedback.

Sorry Francis, that's some bad communication on my part. You've addressed all my feedback nicely and things are in a good shape! I hadn't taken the time to look at this a bit closer up, so I discovered a few unknown unknowns this week. I was hoping I could collaborate by adding a few patches underneath or on top before merging this, that's all. Sorry for the misunderstanding, I wasn't implying you abandoned the work at all. :blush: 

> What is the purpose of asm_relocation()? Is this only proper to cilium/ebpf because when implementing this I got a lot of inspiration from what is done in libbpf.

Correct, this is an old hack that was relied on by some users to implement static data before the kernel had support for pseudo/frozen maps like .(ro)data. There are still kernels out there that don't support them, so we're not going to drop support for this right now.

We have a similar issue for adding kfunc support, since we've always permitted forward declarations in C (see `fwd_decl.c`). This allows the caller to append to `Instructions` with `.WithSymbol("fwd")` after loading the ELF, but before loading the prog into the kernel, to inject behaviour at runtime without recompiling.

Had a discussion with Lorenz about this, we worked out some of the compatibility concerns I had around `asm_relocation`:
- If .kconfig is present, all relos against NOTYPE/GLOBAL/UND symbols must be executed against .kconfig. Any such symbols that don't appear in .kconfig are an error.
- We won't support __weak in this PR, that can be done later (since only LINUX_KERNEL_VERSION atm)

We're still debating whether or not we should expose .kconfig's MapSpec.Contents in the CollectionSpec, or if we'll populate it during map load.

ti-mo:
> How do you plan to add the patches?

By default, maintainers of a repo are allowed to push to PR branches in forks, unless explicitly disabled. The sidebar of this PR says `Maintainers are allowed to edit this pull request.` for me, but I don't seem to manage to push to it. Could you check if this isn't disabled on your end? No need to create other PRs, let's keep the discussion centered here.

> So, you mean to put in .kconfig even symbols which are not marked as __kconfig?

No, I mean we wouldn't support both __kconfig and inline asm relocations (like `MY_CONST` in `asm_relocation()`) in the same ELF. If a .kconfig map is present, all NOTYPE/GLOBAL/UND symbols must exist in it. If no .kconfig map is present and we need to relocate a MY_CONST, we just call `ins.WithReference()` and call it a day. The linker will complain later if the caller didn't fix up the instruction and didn't remove the reference.

ti-mo:
@eiffel-fl Sorry, on PTO this week. Just gave this another try (this is after `gh pr checkout eiffel-fl:francis/ebpf-kconfig` and reworking existing commits and prepending one commit to the existing two:

```
λ  ebpf francis/ebpf-kconfig ✓ git push -f
ERROR: Permission to eiffel-fl/ebpf.git denied to ti-mo.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Do you have any kind of branch protection set up on your fork? There's a branch protection setting that controls force-pushes, but not sure if it's on or off by default.

ti-mo:
@eiffel-fl Could you try re-forking from `cilium/ebpf` instead, and recreating this PR using my branch at https://github.com/ti-mo/ebpf/tree/tb/kconfig? I've added some code that fixes the alignment of the .kconfig Datasec and moved populating .kconfig to happen at map load time instead of ELF load time. Otherwise, missing kconfig values would prevent the ELF from loading, leaving the user with no way to fix things up.

Please close this PR when you've recreated it.

PR #870, opened 2022-12-01
map: include generated pin path in newMapWithOptions error

ti-mo:
When attempting to create a new map with the Pinned field set, a pin path is automatically generated based on MapOptions.PinPath and the map's name.

This makes it cumbersome for the caller to figure out what the final path was that was used in the pin operation, with vague errors like:

'pin map: no such file or directory'

Wrap the error message with the final pin path.


rgo3 (approved):

florianl (approved):
great addition :+1:

PR #878, opened 2022-12-08
[WIP] examples: add fexit example for tracing tcp listen

CSWYF3634076:
I found that there are few examples in the project. This pr is used to add fexit examples for tracing tcp listen
image

I have a question:
Maybe it's because I don't have a thorough understanding of CO-RE. 
According to [BPF CO-RE reference guide](https://nakryiko.com/posts/bpf-core-reference-guide/), fexit BPF program type code can do direct memory dereference and even follow the pointers. 
But in this pr, [fexit.c](https://github.com/cilium/ebpf/compare/master...CSWYF3634076:ebpf:wyf/fexit_examples#diff-14c7edade42b6ff333f82456b02dba416470d5305669c11e4a21f7d20259f233) 143 line inet->inet_sport pointer access. There is an error: "access beyond struct sock at off 808 size 2"


CSWYF3634076:
@ti-mo Thanks for your reply. The reason for the error should be the same as you said
image

I think it may be because I have made a type conversion `inet = (struct inet_sock *)sock->sk;` , then access the memory `inet->inet_sport`.

If I modify the code as follows, no error will occur. Only “ `inet->inet_sport` ”access requires `BPF_CORE_READ`. And `->` can be used normally for others
```c
        sk = sock->sk;
	inet = (struct inet_sock *)sock->sk;
	family = sk->__sk_common.skc_family;
	type = sock->type;

	event->proto = ((__u32)family << 16) | type;
	// TODO There is a question with inet->inet_sport pointer access. error: "access beyond struct sock at off 808 size 2"
	event->port = bpf_ntohs(BPF_CORE_READ(inet,inet_sport));  
	if (family == AF_INET)
		event->addr[0] = sk->__sk_common.skc_rcv_saddr;
	else if (family == AF_INET6)
		bpf_probe_read_kernel(event->addr, sizeof(event->addr),
                           sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
```
I don't know why `BPF_CORE_READ` can, and `->` can't

This example mainly uses fexit to obtain parameters and return values to track the listen system call. The process of obtaining the address and port may be a little cumbersome after dividing ipv6 and ipv4

Issue #853, opened 2022-11-15
how to deal with syscall interruption due to (runtime) signals


lmb:
[BPF_PROG_LOAD] and [BPF_PROG_RUN] are interrupted when a signal is sent to the thread executing the syscall. This leads to an unfortunate interaction with the Go runtime, which uses signals to implement the following features:

* SIGURG for async preemption
* SIGPROF for runtime.CPUProfile
* SIGSETXID for syscall.AllThreadSyscall

In my opinion, the library must deal with all of these gracefully.

## Async preemption

Async preemption is not used when a thread is in a syscall, since that is a ["blocked safe point"] already. No need to worry about SIGURG therefore.

## CPU profiling

The runtime periodically sends SIGPROF to all threads if CPU profiling is enabled. This includes threads in a syscall.

We can block SIGPROF at the expense of less accurate profiles. This solves the most common cause of problems. We currently do this for BPF_PROG_LOAD but not for BPF_PROG_RUN.

## All threads syscall

syscall.AllThreadSyscall sends SIGSETXID to invoke runPerThreadSyscall from the signal handler.

Blocking SIGSETXID would prevent invoking runPerThreadSyscall and cause the runtime to hang. We have to leave the signal unblocked and retry the syscall instead.

Since we can't know when a syscall interruption was due to AllThreadsSyscall we have to always retry.

## Conclusion

I think we should extend the BPF_PROG_LOAD behaviour to BPF_PROG_RUN:
* Mask SIGPROF to reduce noise from the profiler
* Retry on EINTR from PROG_RUN due to AllThreadsSyscall

I expect that AllThreadsSyscall interrupting PROG_RUN is actually very rare. So I propose to also remove RunOptions.Reset and instead document that AllThreadsSyscall or signals may skew the results of a Run. Most importantly we have to document that `RunOptions.Repeat` is basically a lower bound of how often a program is run due to retries.

["blocked safe point"]: https://go.dev/src/runtime/preempt.go
[BPF_PROG_LOAD]: https://github.com/torvalds/linux/commit/c3494801cd1785e2c25f1a5735fa19ddcf9665da
[BPF_PROG_RUN]: https://github.com/torvalds/linux/commit/df1a2cb7c74b3d3abc8d8c2d690f82c8ebc3490a

PR #887, opened 2022-12-12
program: block SIGPROF during BPF_PROG_RUN

lmb:
Extend the profiler blocking we introduced for BPF_PROG_LOAD to BPF_PROG_RUN, with the difference that EINTR is not handled in the syscall wrapper itself. This allows us to have PROG_RUN specific retry logic.

Deprecate RunOptions.Reset since its main motivation was to account for profiler skew. Interruptions from the runtime are only possible via runtime.AllThreadSyscall, which is very specialised anyways. This doesn't protect from interruptions due to signals that are being sent from "outside" the process, which will still skew results and cause retries.

Document that RunOptions.Repeat is a lower bound for how often a program is run due to the retry behaviour. Also add a workaround that makes the common case of repeat == 1 not suffer from interruptions.

Fixes https://github.com/cilium/ebpf/issues/853

### DIFF
--- a/internal/sys/syscall.go
+++ b/internal/sys/syscall.go
@@ -19,7 +19,7 @@ var ENOTSUPP = syscall.Errno(524)
 func BPF(cmd Cmd, attr unsafe.Pointer, size uintptr) (uintptr, error) {
 	// Prevent the Go profiler from repeatedly interrupting the verifier,
 	// which could otherwise lead to a livelock due to receiving EAGAIN.
-	if cmd == BPF_PROG_LOAD {
+	if cmd == BPF_PROG_LOAD || cmd == BPF_PROG_RUN {
 		maskProfilerSignal()
 		defer unmaskProfilerSignal()
 	}
--- a/prog.go
+++ b/prog.go
@@ -554,7 +554,7 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 		Repeat:  1,
 	}
 
-	ret, _, err := p.testRun(&opts)
+	ret, _, err := p.run(&opts)
 	if err != nil {
 		return ret, nil, fmt.Errorf("test program: %w", err)
 	}
@@ -565,7 +565,7 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 //
 // Note: the same restrictions from Test apply.
 func (p *Program) Run(opts *RunOptions) (uint32, error) {
-	ret, _, err := p.testRun(opts)
+	ret, _, err := p.run(opts)
 	if err != nil {
 		return ret, fmt.Errorf("run program: %w", err)
 	}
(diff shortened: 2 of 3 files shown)


internal/sys/syscall.go
```diff
@@ -19,7 +19,7 @@ var ENOTSUPP = syscall.Errno(524)
 func BPF(cmd Cmd, attr unsafe.Pointer, size uintptr) (uintptr, error) {
 	// Prevent the Go profiler from repeatedly interrupting the verifier,
 	// which could otherwise lead to a livelock due to receiving EAGAIN.
-	if cmd == BPF_PROG_LOAD {
+	if cmd == BPF_PROG_LOAD || cmd == BPF_PROG_RUN {
```
lmb, line 22: @ti-mo should we move PROG_LOAD retry into the caller as well, to make PROG_LOAD and PROG_RUN consistent?

ti-mo: I think that was my initial proposal, but having this behaviour here is better from a clarity/doc perspective, since there's no way to misuse it this way.

prog.go
```diff
@@ -675,24 +680,37 @@ func (p *Program) testRun(opts *RunOptions) (uint32, time.Duration, error) {
...
 	for {
 		err := sys.ProgRun(&attr)
 		if err == nil {
-			break
+			break retry
 		}
 
 		if errors.Is(err, unix.EINTR) {
+			if attr.Repeat == 1 {
+				// Signal interruption is checked after each repetition on older
```
ti-mo, line 696: What about kernels post-patch? The difference should be mentioned here. From my reading of the linked commit, signals are checked in `bpf_test_timer_continue()` as the `do..while` condition, so interleaved in between executions, never before starting execution. Doesn't that still qualify as 'after each repetition'?

prog.go
```diff
@@ -675,24 +680,37 @@ func (p *Program) testRun(opts *RunOptions) (uint32, time.Duration, error) {
 		Cpu:         opts.CPU,
 	}
 
+	if attr.Repeat == 0 {
```
ti-mo, line 683: The initial comment on `RunOptions.Repeat` reads `Defaults to 1` already. Does this serve a purpose other than simplifying the break logic below?

prog.go
lmb: Rephrased.

prog.go
```diff
@@ -519,15 +525,14 @@ type RunOptions struct {
...
 	// or similar. Typically used during benchmarking. Optional field.
+	//
+	// Deprecated: use [testing.B.ReportMetric] with unit "ns/op" instead.
 	Reset func()
 }
 
 // Test runs the Program in the kernel with the given input and returns the
 // value returned by the eBPF program. outLen may be zero.
 //
-// Note: the kernel expects at least 14 bytes input for an ethernet header for
```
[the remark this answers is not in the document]
ti-mo: Yep, we can have it in both places IMO.

ti-mo (approved):

PR #888, opened 2022-12-13
Remove deprecated `+build` Go build tags

ti-mo:
Does what it says on the tin, remove `+build` in favour of `//go:build`. To keep this straightforward, it uses a `[][]string` internally, and the template engine ORs all inner slices and ANDs all outer slices. This causes a clean diff with everything currently in examples/.

There's one problem: specifying multiple exclusions is currently a bit broken. `-tags "!foo !bar"` will result in `//go:build (!foo || !bar)`, which cancels each other out. We could pick out all tags starting with `!` and render them out to a separate clause that's ANDed?

@lmb let me know what you think.


lmb (commented):
I've changed your commit so that we don't emit new build lines, but stick to the old command line interface. PTAL.

PR #890, opened 2022-12-15
Add missing program type feature probes

rgo3:
With this PR we support feature probes for all program types the library currently supports.

### DIFF
--- a/btf/types.go
+++ b/btf/types.go
@@ -408,6 +408,12 @@ func FuncMetadata(ins *asm.Instruction) *Func {
 	return fn
 }
 
+// WithFuncMetadata adds a btf.Func to the Metadata of asm.Instruction.
+func WithFuncMetadata(ins asm.Instruction, fn *Func) asm.Instruction {
+	ins.Metadata.Set(funcInfoMeta{}, fn)
+	return ins
+}
+
 func (f *Func) Format(fs fmt.State, verb rune) {
 	formatType(fs, verb, f, f.Linkage, "proto=", f.Type)
 }
--- a/features/prog.go
+++ b/features/prog.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/btf"
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/sys"
 	"github.com/cilium/ebpf/internal/unix"
@@ -92,7 +93,16 @@ var haveProgramTypeMatrix = internal.FeatureMatrix[ebpf.ProgramType]{
 			})
 		},
 	},
-	// ebpf.Tracing:               {Version: "5.5"},
+	ebpf.Tracing: {
+		Version: "5.5",
+		Fn: func() error {
+			return probeProgram(&ebpf.ProgramSpec{
+				Type:       ebpf.Tracing,
+				AttachType: ebpf.AttachTraceFEntry,
+				AttachTo:   "bpf_init",
+			})
+		},
+	},
 	ebpf.StructOps: {
 		Version: "5.6",
 		Fn: func() error {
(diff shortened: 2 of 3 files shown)


features/prog.go
```diff
@@ -92,7 +93,16 @@ var haveProgramTypeMatrix = internal.FeatureMatrix[ebpf.ProgramType]{
...
 		},
 	},
-	// ebpf.Tracing:               {Version: "5.5"},
+	ebpf.Tracing: {
+		Version: "5.5",
+		Fn: func() error {
+			return probeProgram(&ebpf.ProgramSpec{
+				Type:       ebpf.Tracing,
+				AttachType: ebpf.AttachTraceFEntry,
+				AttachTo:   "inet_dgram_connect",
```
ti-mo, line 102: I'm hesitant to hardcode ksyms outside of the test suite, even the oldest of subsystems are sometimes subject to change if there's a need to modernize/refactor. Since fentry relies on `findTargetInKernel()`, we can do the following without any real cost:
```go
func() error {
	s, err := btf.LoadKernelSpec()
	if err != nil {
		return err
	}

	i := s.Iterate()
	for i.Next() {
		// Find the first named kernel function.
		f, ok := i.Type.(*btf.Func)
		if !ok || f.Name == "" {
			continue
		}
		return probeProgram(&ebpf.ProgramSpec{
			Type:       ebpf.Tracing,
			AttachType: ebpf.AttachTraceFEntry,
			AttachTo:   f.Name,
		})
	}

	return errors.New("no suitable kernel function for fentry")
},
```

features/prog.go
```diff
@@ -107,16 +117,54 @@ var haveProgramTypeMatrix = internal.FeatureMatrix[ebpf.ProgramType]{
 			return err
 		},
 	},
-	// ebpf.Extension:             {Version: "5.6"},
+	ebpf.Extension: {
+		Version: "5.6",
+		Fn: func() error {
+			// create btf.Func to add to first ins of target and extension so both progs are btf powered
+			btfFn := btf.Func{Name: "global_func", Type: &btf.FuncProto{Return: &btf.Int{}}, Linkage: btf.GlobalFunc}
```
ti-mo, line 124: Über-nit: we usually go for `"a"` for probes that need named BTF types.

features/prog.go
```diff
@@ -107,16 +117,54 @@ var haveProgramTypeMatrix = internal.FeatureMatrix[ebpf.ProgramType]{
...
+			// probe for Extension prog with target
+			return probeProgram(&ebpf.ProgramSpec{
+				Type: ebpf.Extension,
+				Instructions: asm.Instructions{
+					ins,
+					asm.Return(),
+				},
+				AttachType:   ebpf.AttachNone,
+				AttachTarget: prog,
+				AttachTo:     "global_func",
```
ti-mo, line 155: Make this `btfFn.Name` for good measure.

ti-mo (commented):
Exciting!

features/prog.go
lmb: Some functions may not be traceable, I think this may lead to more harm than good.

btf/types.go
```diff
@@ -408,6 +408,11 @@ func FuncMetadata(ins *asm.Instruction) *Func {
 	return fn
 }
 
+// SetFuncMetadata adds a btf.Func to the Metadata of asm.Instruction.
+func SetFuncMetadata(ins *asm.Instruction, fn *Func) {
```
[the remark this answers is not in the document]
rgo3: Renamed and made it return a Instruction now instead working on the pointer directly.

features/prog.go
```diff
@@ -107,16 +117,54 @@ var haveProgramTypeMatrix = internal.FeatureMatrix[ebpf.ProgramType]{
...
+					LogDisabled: false,
+				},
+			)
+			if err != nil {
+				return err
+			}
+			defer prog.Close()
+
+			// probe for Extension prog with target
+			return probeProgram(&ebpf.ProgramSpec{
```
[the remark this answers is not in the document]
rgo3: So the call only would be `return probeProgram(&spec, &spec)` and I can drop the target program creation with `NewProgram..`? How would I pass the target program to the Extension program in `probeProgram()`?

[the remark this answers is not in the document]
ti-mo: As discussed, my suggestion won't work.

features/prog.go
ti-mo: True, I decided to gather some data. Aside from the obvious issue described in #894, there are currently two clear limitations to Tracing programs. Fortunately, the BTF verifier is being really helpful:

1. `invalid argument: The function kthread_create_worker_on_cpu with variable args is unsupported.`
```c
struct kthread_worker *
kthread_create_worker_on_cpu(int cpu, unsigned int flags,
			         const char namefmt[], ...)
```
So, anything with varargs currently isn't supported.

2. `invalid argument: The function __copy_xstate_to_uabi_buf arg0 type STRUCT is unsupported.`
```c
void __copy_xstate_to_uabi_buf(struct membuf to, struct fpstate *fpstate,
			           u32 pkru_val, enum xstate_copy_mode copy_mode)
```
Any function taking by-value struct args larger than 16 bytes is not supported.

This seems to be governed by [btf_distill_func_proto()](https://elixir.bootlin.com/linux/v6.1.1/source/kernel/bpf/btf.c#L5917).

With that off the table, there's still the option of retrying with up to N funcs if we're not hardcoding a symbol.

ti-mo: I ranked all Funcs containing `bpf` in my kernel's BTF blob by length and found a strong candidate for hardcoding as an fentry probe target: [`bpf_init`](https://elixir.bootlin.com/linux/v6.1.1/source/kernel/bpf/inode.c#L806). It's been part of the kernel since 4.4 (https://github.com/torvalds/linux/commit/b2197755b263), has been completely untouched, and has a simple signature. For probing a BPF program type, it makes sense to target a symbol from the BPF subsystem. @borkmann mentioned that this is unlikely to change.

I'll push a fixup to quickly put this through CI and go for merge.

lmb (approved):

PR #866, opened 2022-11-25
Update dependencies for current Go versions

thaJeztah:
### go.mod: github.com/google/go-cmp v0.5.9

Updating to drop the obsolete golang.org/x/xerrors package

### go.mod: github.com/frankban/quicktest v1.14.4

this updates the indirect dependency of github.com/kr/pretty v0.3.1,
which updates github.com/rogpeppe/go-internal v1.9.0, which contains
some fixes for go1.18 :)

### go.mod: golang.org/x/sys v0.2.0

The golang.org/x project started tagging releases; while the go.mod
should specify a minimum required version, updating to a tagged
helps a bit with code-churn in consumers of this module, as there'll
be less "variations" in the version of this package used.

PR #893, opened 2022-12-18
chore(deps): setup dependabot

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/frankban/quicktest v1.14.4
 	github.com/google/go-cmp v0.5.9
-	golang.org/x/sys v0.2.0
+	golang.org/x/sys v0.3.0
 )
 
 require (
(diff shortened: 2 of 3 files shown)


lmb:
Hey! Can you tell me a bit about why you think dependabot is necessary?

mmorel-35:
Hi @Imb , 

Dependabot is never a necessity, it's just a choice to help you to keep the dependencies of your project up-to-date.
Doing it manually is possible but can be time consuming.

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
```
ti-mo, line 11: Let's make this monthly, please. Following https://github.com/golang/go/issues/21324#issuecomment-1293689721, the expectation is for x/sys to tag ~monthly, which is our only real dependency.

I don't want to follow every quicktest point release, that's unnecessary. Monthly sounds like a good-enough interval to make PRs like https://github.com/cilium/ebpf/pull/866 unnecessary, while staying reasonably up-to-date regarding subtle bugs introduced by newer Go versions. Obviously this is only ever an issue around Go release windows, but having updated dependencies may reduce papercuts for downstream projects anyway.

lmb (changes requested):
Initially I thought this change might make sense since we'd prevent our dependents from using newer versions of our own dependencies. After reading up on MVS, here: https://go.dev/ref/mod#minimal-version-selection I don't think this is actually the case, and introducing dependabot just creates additional churn.

Unless I misunderstood something again I don't see the use of dependabot.

mmorel-35:
That's up to you. You can close the PR if you are not interested.

ti-mo:
@lmb Correct, we're not forcing anything downstream. Individual modules always build against the exact version as specified in `go.sum`, even when imported into larger projects. As far as I understand, this would mean a binary can contain multiple versions of a given symbol, but I haven't verified that.

FYI I'm also on the fence, but once a month is a rather low frequency, and it's a free PR that kicks off CI automatically, and will be green and ready to merge in 99.9% of cases.

lmb:
> As far as I understand, this would mean a binary can contain multiple versions of a given symbol, but I haven't verified that.

I read the MVS page to mean the opposite: only a single version will ever be chosen.

> At the end of the graph traversal, MVS returns a build list containing the bolded versions: A 1.2, B 1.2, C 1.4, and D 1.2.

lmb:
I think the net effect of this change would be that dependents of this package would have their indirect dependencies be bumped once a month. Due to how MVS works this new version would then be chosen for their build. All code has bugs of course, but new code has more, and dependabot automation would force using new code without "a need".

ti-mo:
@lmb This package doesn't release once a month. Dependabot will only ping our users when we create a new tag.

lmb:
Yes, you're right, it would happen less often. I still think that unless there is a good reason (new API, bugfix) we should not update dependencies. That gives consumers of the lib more choice as to when or if they upgrade their dependencies.

ti-mo:
@mmorel-35 Thanks for the proposal, but we won't be going forward with this.

Issue #894, opened 2022-12-22
Tracing programs cannot be attached to non-unique kernel symbols


ti-mo:
**Describe the bug**

While hacking on https://github.com/cilium/ebpf/pull/890, I decided to try and attach a fentry prog to all of my machine's symbols. Many of them fail with: `attach Tracing/TraceFEntry: find target for fentry update_persistent_clock64 in vmlinux: type update_persistent_clock64: multiple candidates for *btf.Func`

```
[1096] STRUCT 'timespec64' size=16 vlen=2
        'tv_sec' type_id=1095 bits_offset=0
        'tv_nsec' type_id=90 bits_offset=64
..
[28302] FUNC_PROTO '(anon)' ret_type_id=69 vlen=1
        'now64' type_id=1096
[28303] FUNC 'update_persistent_clock64' type_id=28302 linkage=static
..
[62944] FUNC_PROTO '(anon)' ret_type_id=69 vlen=1
        'now' type_id=1096
[62945] FUNC 'update_persistent_clock64' type_id=62944 linkage=static
```

`update_persistent_clock64` is a weak vmlinux symbol that has an arch-specific implementation:
https://elixir.bootlin.com/linux/v6.1.1/source/kernel/time/ntp.c#L568
https://elixir.bootlin.com/linux/v6.1.1/source/arch/x86/kernel/rtc.c#L103

The argument in the weak symbol is named `now64`, the arch-specific one is called `now`.

**Expected behavior**

The library should either go with the first candidate, or additionally verify binary compatibility between all candidates' function signatures.

Or, we might be stuck between a rock and a hard place if the kernel expects the function's specific BTF id to be given. Perhaps we may need to try loading a prog pointing at each candidate until one is accepted.

Looks similar to https://github.com/cilium/ebpf/issues/723 and https://github.com/cilium/ebpf/issues/466.

lmb:
Can both 28303 and 62945 be attached to? Why is `linkage=static` on both funcs? Shouldn't one be weak (maybe a clang / pahole version issue)?

ti-mo:
As it turns out, trying to attach Tracing programs to kernel symbols with overlapping names is currently subtly broken in Linux up until at least 6.1. Here's what I found:

```
λ  ~  sudo bpftool btf dump id 1 | grep "'type_show'"
[24381] FUNC 'type_show' type_id=3718 linkage=static
[28217] FUNC 'type_show' type_id=3798 linkage=static
[...]

PR #896, opened 2023-01-13
btf: implement Type equality using btf.Equal()

ti-mo:
Opening this as draft as this likely needs some work to fit into the bigger picture. This was created to gather data for #894, has several other minor use cases (aside from convenience) and might aid in C/Go code generation. Also, perhaps it could inspire a better/alternative implementation. All in all, I think this is relatively little code for what it accomplishes.

### DIFF
--- a/btf/types.go
+++ b/btf/types.go
@@ -1,6 +1,7 @@
 package btf
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -36,6 +37,9 @@ type Type interface {
 	// Make a copy of the type, without copying Type members.
 	copy() Type
 
+	// Compare the Type against another. Ignores Name and any descendant Types.
+	equal(Type) bool
+
 	// New implementations must update walkType.
 }
 
@@ -67,6 +71,62 @@ func (ts types) ByID(id TypeID) (Type, error) {
 	return ts[id], nil
 }
 
+type Skipper func(Type) bool
+
+type Comparer func(Type, Type) bool
+
+// SkipQualifierAndTypedef skips qualifiers and Typedefs.
+var SkipQualifierAndTypedef = func(t Type) bool {
+	switch t.(type) {
+	case qualifier, *Typedef:
+		return true
+	}
+	return false
+}
+
+// CompareTypeName returns true if a and b's names correspond.
+var CompareTypeName = func(a, b Type) bool {
+	return a.TypeName() == b.TypeName()
+}
+
+// Equal compares Type a to Type b. By default, only binary equivalency is
+// checked, ignoring type names.
+//
+// cmp allows providing extra comparison rules. It will be evaluated after
+// checking a and b's binary equivalency at each step of the type hierarchy.
+// skip allows individual Types to be omitted from comparisons, and is called
+// before checking a and b's binary equivalency.
+func Equal(a, b Type, cmp Comparer, skip Skipper) error {
+	if a == nil && b == nil {
+		return nil
+	}
+	if a == nil || b == nil {
+		return errors.New("one of a or b are nil")
+	}
+
+	at, bt := postorderTraversal(a, nil), postorderTraversal(b, nil)
+	for at.Next() && bt.Next() {
+		if skip != nil {
+			for skip(at.Type) {
+				at.Next()
+			}
+			for skip(bt.Type) {
+				bt.Next()
+			}
+		}
+
+		if !at.Type.equal(bt.Type) {
+			return fmt.Errorf("a: %v and b: %v differ", at.Type, bt.Type)
+		}
+
+		if cmp != nil && !cmp(at.Type, bt.Type) {
+			return fmt.Errorf("a: %v and b: %v did not satisfy cmp", at.Type, bt.Type)
+		}
+	}
+
+	return nil
+}
+
 // Void is the unit type of BTF.
 type Void struct{}
 
@@ -74,6 +134,10 @@ func (v *Void) Format(fs fmt.State, verb rune) { formatType(fs, verb, v) }
 func (v *Void) TypeName() string               { return "" }
 func (v *Void) size() uint32                   { return 0 }
 func (v *Void) copy() Type                     { return (*Void)(nil) }
+func (v *Void) equal(other Type) bool {
+	_, ok := other.(*Void)
+	return ok
+}
 
 type IntEncoding byte
 
@@ -125,6 +189,16 @@ func (i *Int) copy() Type {
 	return &cpy
 }
 
+func (i *Int) equal(other Type) bool {
+	oi, ok := other.(*Int)
+	if !ok {
+		return false
+	}
+
+	return i.Size == oi.Size &&
+		i.Encoding == oi.Encoding
+}
+
 // Pointer is a pointer to another type.
 type Pointer struct {
 	Target Type
@@ -139,6 +213,11 @@ func (p *Pointer) size() uint32     { return 8 }
 func (p *Pointer) copy() Type {
 	cpy := *p
 	return &cpy
+}
+
+func (p *Pointer) equal(other Type) bool {
+	_, ok := other.(*Pointer)
+	return ok
 }
 
 // Array is an array with a fixed number of elements.
@@ -157,6 +236,14 @@ func (arr *Array) TypeName() string { return "" }
 func (arr *Array) copy() Type {
 	cpy := *arr
 	return &cpy
+}
+
+func (arr *Array) equal(other Type) bool {
+	oa, ok := other.(*Array)
+	if !ok {
+		return false
+	}
+	return arr.Nelems == oa.Nelems
 }
 
 // Struct is a compound type of consecutive members.
@@ -185,6 +272,25 @@ func (s *Struct) members() []Member {
 	return s.Members
 }
 
+func (s *Struct) equal(other Type) bool {
+	os, ok := other.(*Struct)
+	if !ok {
+		return false
+	}
+	if s.Size != os.Size ||
+		len(s.Members) != len(os.Members) {
+		return false
+	}
+
+	for i, m := range s.Members {
+		if !m.equal(os.Members[i]) {
+			return false
+		}
+	}
+
+	return true
+}
+
 // Union is a compound type where members occupy the same memory.
 type Union struct {
 	Name string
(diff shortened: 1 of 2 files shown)


btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
 	return ts[id], nil
 }
 
+type Skipper func(Type) bool
+
+type Comparer func(Type, Type) bool
+
+// SkipQualifierAndTypedef skips qualifiers and Typedefs.
+var SkipQualifierAndTypedef = func(t Type) bool {
```
lmb, line 79: Why isn't this just a func?

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
 	return ts[id], nil
 }
 
+type Skipper func(Type) bool
```
lmb, line 74: Nice, could use this for postorderTraversal as well. Should have documentation if we decide to export it though, otherwise there is no point?

Nit: should this be `SkipFn` instead to distinguish from an interface?

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+
+// CompareTypeName returns true if a and b's names correspond.
+var CompareTypeName = func(a, b Type) bool {
+	return a.TypeName() == b.TypeName()
+}
+
+// Equal compares Type a to Type b. By default, only binary equivalency is
+// checked, ignoring type names.
+//
+// cmp allows providing extra comparison rules. It will be evaluated after
```
lmb, line 95: I think the semantics should be: if cmp == nil we default to an exported function that does what you call binary equivalency.  Otherwise we defer to cmp entirely.

That way we could use this function to replace `coreAreTypesCompatible` which has a completely different notion of equality.

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+		if !at.Type.equal(bt.Type) {
+			return fmt.Errorf("a: %v and b: %v differ", at.Type, bt.Type)
+		}
+
+		if cmp != nil && !cmp(at.Type, bt.Type) {
+			return fmt.Errorf("a: %v and b: %v did not satisfy cmp", at.Type, bt.Type)
+		}
+	}
+
+	return nil
```
lmb, line 127: This is missing a check for `!at.Empty() || !bt.Empty()`, see `coreAreTypesCompatible`.

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+// before checking a and b's binary equivalency.
+func Equal(a, b Type, cmp Comparer, skip Skipper) error {
+	if a == nil && b == nil {
+		return nil
+	}
+	if a == nil || b == nil {
+		return errors.New("one of a or b are nil")
+	}
+
+	at, bt := postorderTraversal(a, nil), postorderTraversal(b, nil)
```
lmb, line 107: I think pre-order traversal is better here: https://en.wikipedia.org/wiki/Tree_traversal#Data_structures_for_tree_traversal It would mean we fail faster when there is a difference. The implementation of preorder can also be faster since it doesn't need as much temporary storage.

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+}
+
+// Equal compares Type a to Type b. By default, only binary equivalency is
+// checked, ignoring type names.
+//
+// cmp allows providing extra comparison rules. It will be evaluated after
+// checking a and b's binary equivalency at each step of the type hierarchy.
+// skip allows individual Types to be omitted from comparisons, and is called
+// before checking a and b's binary equivalency.
+func Equal(a, b Type, cmp Comparer, skip Skipper) error {
```
lmb, line 99: Do we need `skip`? The user can either strip qualifiers, etc. with `Copy` or have a custom `cmp` which calls `UnderlyingType` or similar. See below why I think skip is actually dangerous.

What if this was `Compare(a, b Type, fn Comparer)` instead? `Equal` makes me think this should return a bool instead.

btf/types.go
```diff
@@ -36,6 +37,9 @@ type Type interface {
 	// Make a copy of the type, without copying Type members.
 	copy() Type
 
+	// Compare the Type against another. Ignores Name and any descendant Types.
+	equal(Type) bool
```
lmb, line 41: Why is ignoring names the right thing to do?

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+	}
+	return false
+}
+
+// CompareTypeName returns true if a and b's names correspond.
+var CompareTypeName = func(a, b Type) bool {
+	return a.TypeName() == b.TypeName()
+}
+
+// Equal compares Type a to Type b. By default, only binary equivalency is
```
lmb, line 92: "binary equivalency" is not really self explanatory.

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+	if a == nil && b == nil {
+		return nil
+	}
+	if a == nil || b == nil {
+		return errors.New("one of a or b are nil")
+	}
+
+	at, bt := postorderTraversal(a, nil), postorderTraversal(b, nil)
+	for at.Next() && bt.Next() {
+		if skip != nil {
```
lmb, line 109: Allowing skipping like this is dangerous, I think. postorderTraversal flattens a graph into a list, so we lose the shape of the graph. This is okay (I think, need to convince myself of that again) as long as we always proceed in lockstep, consuming one type from each list. This skipping breaks that invariant so I think you can end up comparing types from different branches of a type.

btf/types.go
```diff
@@ -67,13 +71,73 @@ func (ts types) ByID(id TypeID) (Type, error) {
...
+//
+// cmp allows providing extra comparison rules. It will be evaluated after
+// checking a and b's binary equivalency at each step of the type hierarchy.
+// skip allows individual Types to be omitted from comparisons, and is called
+// before checking a and b's binary equivalency.
+func Equal(a, b Type, cmp Comparer, skip Skipper) error {
+	if a == nil && b == nil {
+		return nil
+	}
+	if a == nil || b == nil {
```
lmb, line 103: IMO this should be deferred to `cmp`. Maybe even the `a == b == nil` case.

btf/types.go
```diff
@@ -244,6 +369,17 @@ type Member struct {
 	BitfieldSize Bits
 }
 
+func (m Member) equal(other Member) bool {
+	// Compare Member properties, but don't recurse into their Types. Post-order
```
lmb, line 373: Nit: skip the comment, it applies to Pointer, Const, etc. as well.

btf/types.go
```diff
@@ -568,6 +843,11 @@ func (c *cycle) copy() Type {
 	return &cpy
 }
 
+func (c *cycle) equal(other Type) bool {
+	_, ok := other.(*cycle)
+	return ok
```
lmb, line 848: I think this should always be `false`, because `walkType` never recurses into a cycle.

lmb (commented):
Left a bunch of comments, I think this could become a powerful API. Nice work.

It'd be great if we could somehow add a test that individual Type's `equal` actually does what we want. Maybe we can fiddle with the field via reflection?

btf/types.go
ti-mo: Oh absolutely, postorder is much too expensive for this. I initially tried comparing all vmlinux' types against themselves, and that didn't finish within 5 minutes or so. Preorder doesn't require any allocs except for the iterator that has some scratch space, and as you mentioned, would allow us to bail out much sooner.

btf/types.go
ti-mo: I wrote Equal() with comparing structs and unions in mind, which `UnderlyingType()` doesn't touch. Structs and unions can almost be considered type graphs of their own, and (especially in vmlinux) can take absolutely monstrous proportions, so copies are not ideal. Also, calling `UnderlyingType()` in a comparer would copy every type at every level in the hierarchy, which gets expensive quickly.

You say `The user can either strip qualifiers`, but that's not so trivial for those unfamiliar with the ins and outs of BTF, and is just annoying from the caller's perspective with `qualifier` currently being unexported. Also fragile long-term as new qualifiers get added/encoded into BTF. I'd expect the library to provide the definition of what is and what isn't a qualifier, and the caller just being able to opt out of them if needed.

> `Equal` makes me think this should return a bool instead.

That's how it started out, but I started adding printfs all over the place to get more info. I think the final implementation (using pre-order traversal) should record the path taken on both sides and include those in the error message. (e.g. `a.bar.baz` (int64) differs from `b.quux.foo` (int32) -- you get the point)

btf/types.go
ti-mo: Agreed, I think 'layout' should be the term used.

btf/types.go
ti-mo: For my purposes, struct/type/func layouts were the lowest-common denominator comparing kernel types. e.g. names are not considered to determine if a func satisfies a C prototype or Go interface.

btf/types.go
ti-mo: All valid concerns indeed, and of course skips need to be used with care, that's somewhat self-explanatory.

The hierarchies are indeed flattened, but a and b are flattened the exact same way, and the same skip() is executed on both of them. At some point the comparison will converge on a complex (struct/union/func) parent node. If either side skipped a whole branch, one side will still be evaluating the branch while the other hits the parent, and a mismatch will occur.

I don't think this is a huge concern as it will flag a mismatch at some point, but it does call into question the usefulness of skipping anything other than qualifiers and typedefs. At worst the output will be inaccurate.

lmb: > The hierarchies are indeed flattened, but a and b are flattened the exact same way [...]. If either side skipped a whole branch, one side will still be evaluating the branch while the other hits the parent, and a mismatch will occur.

I had something like these types in mind. Would this mechanism still work for them? I'm kinda sceptical (and now think that the CO-RE logic is buggy because of this).

```go
struct {
  a struct{}
  b int
  c int
}

struct {
  a struct {
    b int
    c int
  }
}
```

> and the same skip() is executed on both of them.

That assumes that skip is a pure function. I've already been bitten by that not being true when using `postorderTraversal` in the Spec.Add PR, and I wrote that code!

> but it does call into question the usefulness of skipping anything other than qualifiers and typedefs.

Yes, indeed. However, I seem to remember a case where I wanted to skip only qualifiers and not typedefs though (GoFormatter maybe?) That implies that just a `skip bool` doesn't cut it.

ti-mo: > That implies that just a `skip bool` doesn't cut it.

Totally, just being devil's advocate. Equal() should always take some form of function for maximum flexibility and composability.

ti-mo: > Would this mechanism still work for them? [..example..]

With the current PR, this should flatten both hierarchies into [c, b, a, <parent struct>], and both legs will visit them in the same order.

c and b will indeed match, even though they're not at the same level in the hierarchy. a will not match because the left side has 0 members and the right side has two members. Likewise, `<parent struct>` has 3 members on the left and 1 member on the right, so won't match either. See what I mean?

lmb:
Closing this for now. I think the way to implement this is something like https://github.com/cilium/ebpf/pull/1383 to do a postorder traversal on a type.

PR #902, opened 2023-01-16
cs.RewriteConstants: define error MissingConstantsError

alban:
RewriteConstants can return the error "spec is missing one or more
constants" if the constant given as parameter does not exist. This patch
defines the error as a specific type so it can be tested with:
```
    var mErr *MissingConstantsError
    if !errors.As(err, &mErr) {
```
New unit tests are added, for both the success and failure cases.

Tested with:
```
    go test -exec sudo -run TestCollectionRewriteCon  ./...
```
Signed-off-by: Alban Crequy <albancrequy@linux.microsoft.com>

---

This PR was suggested by @mauriciovasquezbernal in https://github.com/inspektor-gadget/inspektor-gadget/pull/1201#discussion_r1053515224 to avoid testing with
```
strings.Contains(err.Error(), "spec is missing one or more constants")
```

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -107,6 +107,12 @@ func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
 	return nil
 }
 
+type MissingConstantsError struct{ Constants []string }
+
+func (m *MissingConstantsError) Error() string {
+	return fmt.Sprintf("spec is missing one or more constants: %s", strings.Join(m.Constants, ", "))
+}
+
 // RewriteConstants replaces the value of multiple constants.
 //
 // The constant must be defined like so in the C program:
@@ -120,7 +126,7 @@ func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
 //
 // From Linux 5.5 the verifier will use constants to eliminate dead code.
 //
-// Returns an error if a constant doesn't exist.
+// Returns an error of type MissingConstantsError if a constant doesn't exist.
 func (cs *CollectionSpec) RewriteConstants(consts map[string]interface{}) error {
 	replaced := make(map[string]bool)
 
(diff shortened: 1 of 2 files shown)


collection.go
```diff
@@ -107,6 +107,8 @@ func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
 	return nil
 }
 
+var ErrMissingConstants = errors.New("spec is missing one or more constants")
```
lmb, line 110: Judging from the test this should really be a dedicated error:

```go
type MissingConstantsError struct{ Constants []string }
```

Then the test can `errors.As` and assert on `err.Constants` instead of string comparison. WDYT?

Please update the documentation on RewriteConstants to indicate that a specific error is returned.

collection_test.go
```diff
@@ -328,6 +329,57 @@ func TestCollectionSpecMapReplacements_SpecMismatch(t *testing.T) {
...
 
+func TestCollectionRewriteConstants(t *testing.T) {
+	cs := &CollectionSpec{
+		Maps: map[string]*MapSpec{
+			".rodata": {
+				Type:       Array,
+				KeySize:    4,
+				ValueSize:  4,
+				MaxEntries: 1,
+				Value: &btf.Datasec{
```
lmb, line 340: Ha, very neat that this works!

collection.go
```diff
@@ -184,7 +186,7 @@ func (cs *CollectionSpec) RewriteConstants(consts map[string]interface{}) error
 	}
 
 	if len(missing) != 0 {
-		return fmt.Errorf("spec is missing one or more constants: %s", strings.Join(missing, ","))
+		return fmt.Errorf("%w: %s", ErrMissingConstants, strings.Join(missing, ","))
```
lmb, line 189: Should probably join with `", "`

lmb (changes requested):

alban:
@lmb Thanks for the review. I updated the patch.

lmb:
@alban I made some fixes, but I can't push to the kinvolk fork of the repo. Can you recreate from a personal fork please?

alban:
Superseded by https://github.com/cilium/ebpf/pull/904

PR #904, opened 2023-01-17
cs.RewriteConstants: define error MissingConstantsError

alban:
RewriteConstants can return the error "spec is missing one or more constants" if the constant given as parameter does not exist. This patch defines the error as a specific type so it can be tested with:
```
    var mErr *MissingConstantsError
    if !errors.As(err, &mErr) {
```
New unit tests are added, for both the success and failure cases.

Tested with:
```
    go test -exec sudo -run TestCollectionRewriteCon  ./...
```
Signed-off-by: Alban Crequy <albancrequy@linux.microsoft.com>

-----

This PR was suggested by @mauriciovasquezbernal in https://github.com/inspektor-gadget/inspektor-gadget/pull/1201#discussion_r1053515224 to avoid testing with
```
strings.Contains(err.Error(), "spec is missing one or more constants")
```

-----

Supersedes https://github.com/cilium/ebpf/pull/902

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -107,6 +107,16 @@ func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
 	return nil
 }
 
+// MissingConstantsError is returned by [CollectionSpec.RewriteConstants].
+type MissingConstantsError struct {
+	// The constants missing from .rodata.
+	Constants []string
+}
+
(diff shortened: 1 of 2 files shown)


lmb (approved):

lmb:
Thanks!

PR #905, opened 2023-01-17
Fix ProgramInfo.MapIDs for programs without maps on old kernels

### LINKED ISSUE #535: `ProgramInfo`'s `MapIDs()` bool returns false negatives
**Describe the bug**
Currently, the `(*ProgramInfo).MapIDs()`'s bool return value is simply determined by `pi.ids != nil`, `pi.ids` being allocated when `NrMapIds > 0`. The `NrMapIds` field will always be zero on older kernels that don't yet support it, so it's currently impossible to tell if the prog actually references 0 maps, or whether the kernel is too old to populate the field.

**Expected behavior**
On kernels that do support the NrMapIds field (4.15 and up), always return `true` for the second argument so the caller can determine for sure that the program has no map references.

This would require implementing a probe that inserts a prog that references a map and checking if NrMapIds is populated.

lmb:
Fixes #535.

### DIFF
--- a/info.go
+++ b/info.go
@@ -130,6 +130,12 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 		pi.maps = make([]MapID, info.NrMapIds)
 		info2.NrMapIds = info.NrMapIds
 		info2.MapIds = sys.NewPointer(unsafe.Pointer(&pi.maps[0]))
+	} else if haveProgramInfoMapIDs() == nil {
+		// This program really has no associated maps.
+		pi.maps = make([]MapID, 0)
+	} else {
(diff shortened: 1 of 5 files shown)


info.go
```diff
@@ -321,3 +327,30 @@ func EnableStats(which uint32) (io.Closer, error) {
...
+	}, SocketFilter, "MIT")
+	if err != nil {
+		return err
+	}
+	defer prog.Close()
+
+	err = sys.ObjInfo(prog, &sys.ProgInfo{
+		// NB: Don't need to allocate MapIds since the program isn't using
+		// any maps.
+		NrMapIds: 1,
```
ti-mo, line 344: Clever!

ti-mo (approved):

PR #906, opened 2023-01-18
link: use /sys/kernel/tracing if available

spikat:
The tracefs should be mounted at `/sys/kernel/tracing`, and/or `/sys/kernel/debug/tracing`, or not mounted at all.
Today cilium/ebpf rely only on `/sys/kernel/debug/tracing`, this PR adds a fallback to also consider `/sys/kernel/tracing` (which should be the default one since kernel 4.1, based on what I understand from [the documentation](https://www.kernel.org/doc/Documentation/trace/ftrace.txt)).

### DIFF
--- a/examples/tracepoint_in_c/main.go
+++ b/examples/tracepoint_in_c/main.go
@@ -1,7 +1,7 @@
 // This program demonstrates attaching an eBPF program to a kernel tracepoint.
 // The eBPF program will be attached to the page allocation tracepoint and
 // prints out the number of times it has been reached. The tracepoint fields
-// are printed into /sys/kernel/debug/tracing/trace_pipe.
+// are printed into /sys/kernel/tracing/trace_pipe.
 package main
 
 import (
@@ -35,7 +35,7 @@ func main() {
 	// counter by 1. The read loop below polls this map value once per
 	// second.
 	// The first two arguments are taken from the following pathname:
-	// /sys/kernel/debug/tracing/events/kmem/mm_page_alloc
+	// /sys/kernel/tracing/events/kmem/mm_page_alloc
 	kp, err := link.Tracepoint("kmem", "mm_page_alloc", objs.MmPageAlloc, nil)
 	if err != nil {
 		log.Fatalf("opening tracepoint: %s", err)
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -17,7 +17,7 @@ import (
 )
 
 var (
-	kprobeEventsPath = filepath.Join(tracefsPath, "kprobe_events")
+	kprobeEventsPath = filepath.Join(getTracefsPath(), "kprobe_events")
 )
 
 type probeType uint8
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
@@ -14,10 +13,6 @@ import (
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal/sys"
 	"github.com/cilium/ebpf/internal/unix"
-)
-
-var (
-	kprobeEventsPath = filepath.Join(getTracefsPath(), "kprobe_events")
 )
 
 type probeType uint8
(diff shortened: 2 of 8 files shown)


examples/headers/bpf_helper_defs.h
```diff
@@ -113,17 +113,17 @@ static __u64 (*bpf_ktime_get_ns)(void) = (void *) 5;
  *
  * 	This helper is a "printk()-like" facility for debugging. It
  * 	prints a message defined by format *fmt* (of size *fmt_size*)
- * 	to file *\/sys/kernel/debug/tracing/trace* from DebugFS, if
```
lmb, line 116: Please don't update the headers, they are "vendored" from libbpf.

examples/tracepoint_in_c/main.go
```diff
@@ -1,7 +1,7 @@
 // This program demonstrates attaching an eBPF program to a kernel tracepoint.
 // The eBPF program will be attached to the page allocation tracepoint and
 // prints out the number of times it has been reached. The tracepoint fields
-// are printed into /sys/kernel/debug/tracing/trace_pipe.
+// are printed into /sys/kernel/[debug/]tracing/trace_pipe.
```
lmb, line 4: Let's just change the path to the non-deprecated /sys/kernel/tracing.

link/perf_event.go
```diff
@@ -432,3 +432,17 @@ func isValidTraceID(s string) bool {
 
 	return true
 }
+
+// getTracefsPath will return a correct path to the tracefs mount point.
+// Since kernel 4.1 tracefs should be mounted by default at /sys/kernel/tracing,
+// but may be also be available at /sys/kernel/debug/tracing if debugfs is mounted.
+// The available tracefs paths will depends on distribution choices.
+func getTracefsPath() string {
+	if tracefsPath == "" {
```
lmb, line 441: This creates a race condition, please remove tracefsPath and use sync.Once. Example from the codebase: https://github.com/cilium/ebpf/blob/63a3caefca29ce4a580a033044c71b57eb034092/internal/cpu.go#L10-L24

link/perf_event.go
```diff
@@ -432,3 +432,17 @@ func isValidTraceID(s string) bool {
...
+		if _, err := os.Stat("/sys/kernel/tracing/kprobe_events"); err == nil {
```
lmb, line 442: Why not stat just `/sys/kernel/tracing`?

lmb (changes requested):
Hey, thanks for your PR! Left some feedback, please take a look.

examples/headers/bpf_helper_defs.h
spikat: Change reverted on last commit

link/perf_event.go
spikat: Oh right .. Thanks for the catch and suggestion, this should be ok now.

link/perf_event.go
spikat: Sadly on some distro (like Debian 10 and Ubuntu 18 at least) tracefs is only mounted once in /sys/kernel/debug/tracing, but the mount point /sys/kernel/tracing still exist (but empty..). It really depends on each distributions/version init (it's well mounted twice in Ubuntu 20/22).
That's why I stat the kprobe_events entry instead of just the mount point

spikat:
Hi @lmb and thank you for your quick review. I've applied all of your suggestions. :)

spikat:
Gentle bump. Have you by chance the time to have a look on my last changes @lmb ? :)

lmb:
Sorry, this one slipped through the cracks. I had a review pending but didn't submit it somehow. I pushed another commit that uses `internal.Memoize` which I added recently.

link/kprobe.go
```diff
@@ -61,11 +56,13 @@ func (pt probeType) String() string {
 	return "uprobe"
 }
 
-func (pt probeType) EventsPath() string {
-	if pt == kprobeType {
-		return kprobeEventsPath
+func (pt probeType) EventsFile() (*os.File, error) {
+	path, err := sanitizeTracefsPath(fmt.Sprintf("%s_events", pt.String()))
```
lmb, line 60: @mmat11 FYI I'm rejigging things a little bit here.

lmb (approved):

spikat:
> Sorry, this one slipped through the cracks. I had a review pending but didn't submit it somehow. I pushed another commit that uses `internal.Memoize` which I added recently.

No worries. Thank you for the merge!

PR #908, opened 2023-01-18
sys: overhaul Pointer constructors

lmb:
Use generics to make the Pointer constructors more useful. Also remove the "New" prefix from constructors that do not allocate.

Introduce SliceLen and SliceElems which return 0 if the length of a slice exceeds uint32. This will make it much more obvious when a buffer is too large to be passed to a syscall, since passing a pointer with zero length will cause EFAULT or similar.

Signed-off-by: Lorenz Bauer <oss@lmb.io>

### DIFF
--- a/internal/sys/ptr.go
+++ b/internal/sys/ptr.go
@@ -1,43 +1,57 @@
 package sys
 
 import (
+	"math"
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal/unix"
 )
 
-// NewPointer creates a 64-bit pointer from an unsafe Pointer.
-func NewPointer(ptr unsafe.Pointer) Pointer {
+// UnsafePointer creates a 64-bit pointer from an unsafe Pointer.
+func UnsafePointer(ptr unsafe.Pointer) Pointer {
 	return Pointer{ptr: ptr}
 }
 
-// NewSlicePointer creates a 64-bit pointer from a byte slice.
-func NewSlicePointer(buf []byte) Pointer {
-	if len(buf) == 0 {
+// SlicePointer creates a 64-bit pointer from a slice.
+func SlicePointer[E any](slice []E) Pointer {
+	if len(slice) == 0 {
 		return Pointer{}
 	}
 
-	return Pointer{ptr: unsafe.Pointer(&buf[0])}
+	return Pointer{ptr: unsafe.Pointer(&slice[0])}
 }
 
-// NewSlicePointer creates a 64-bit pointer from a byte slice.
+// SliceLen returns the length of a slice as a uint32.
 //
-// Useful to assign both the pointer and the length in one go.
-func NewSlicePointerLen(buf []byte) (Pointer, uint32) {
-	return NewSlicePointer(buf), uint32(len(buf))
+// Returns zero if the length of the slice exceeds uint32.
+func SliceLen[E any](slice []E) uint32 {
+	return SliceElems(slice, 1)
 }
 
-// NewStringPointer creates a 64-bit pointer from a string.
+// SliceElems returns the number of equal sized elements in a slice.
+//
+// Returns zero if the number of elements exceeds uint32.
+func SliceElems[E any](slice []E, elemSize int) uint32 {
+	n := len(slice) / elemSize
+	if int64(n) > math.MaxUint32 {
+		return 0
+	}
+
+	return uint32(n)
+}
+
+// NewStringPointer allocates a null-terminated backing slice for str and returns
+// a pointer to it.
 func NewStringPointer(str string) Pointer {
-	p, err := unix.BytePtrFromString(str)
+	s, err := unix.ByteSliceFromString(str)
 	if err != nil {
 		return Pointer{}
 	}
 
-	return Pointer{ptr: unsafe.Pointer(p)}
+	return SlicePointer(s)
 }
 
-// NewStringSlicePointer allocates an array of Pointers to each string in the
+// NewStringSlicePointer allocates an array of Pointers for each string in the
 // given slice of strings and returns a 64-bit pointer to the start of the
 // resulting array.
 //
@@ -48,5 +62,5 @@ func NewStringSlicePointer(strings []string) Pointer {
 		sp = append(sp, NewStringPointer(s))
 	}
 
-	return Pointer{ptr: unsafe.Pointer(&sp[0])}
+	return SlicePointer(sp)
 }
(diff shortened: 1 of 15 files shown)


Diego3237:
CU

internal/sys/ptr.go
```diff
@@ -1,43 +1,57 @@
...
 import (
+	"math"
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal/unix"
 )
 
-// NewPointer creates a 64-bit pointer from an unsafe Pointer.
-func NewPointer(ptr unsafe.Pointer) Pointer {
+// UnsafePointer creates a 64-bit pointer from an unsafe Pointer.
```
ti-mo, line 10: This makes some cases stutter (`sys.UnsafePointer(unsafe.Pointer(&info))`) and makes for longer lines overall. It still allocates, just not necessarily on the heap. :wink: I think NewPointer was just fine, it still returns 'a new Pointer'.

internal/sys/ptr.go
```diff
@@ -1,43 +1,57 @@
...
 }
 
-// NewSlicePointer creates a 64-bit pointer from a byte slice.
+// SliceLen returns the length of a slice as a uint32.
 //
-// Useful to assign both the pointer and the length in one go.
-func NewSlicePointerLen(buf []byte) (Pointer, uint32) {
-	return NewSlicePointer(buf), uint32(len(buf))
+// Returns zero if the length of the slice exceeds uint32.
+func SliceLen[E any](slice []E) uint32 {
```
ti-mo, line 27: This is only really useful if E is []byte.

ti-mo (changes requested):
Nice overall, but I have my doubts about SliceElems and SliceLen.

internal/sys/ptr.go
```diff
@@ -1,43 +1,57 @@
...
+	return SliceElems(slice, 1)
 }
 
-// NewStringPointer creates a 64-bit pointer from a string.
+// SliceElems returns the number of equal sized elements in a slice.
+//
+// Returns zero if the number of elements exceeds uint32.
+func SliceElems[E any](slice []E, elemSize int) uint32 {
```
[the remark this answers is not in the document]
lmb: > elemSize > 1 only really makes sense if E is []byte, so doesn't really belong in a generic function.

> If we want to avoid the many uint32(len(..., maybe declaring a method Elems(size int) uint32 over a []byte alias would make more sense?

Fair point. What about `SliceElems([]byte, int)` instead then? I'm not sure what the alias type would do.

In a way, using `[]byte` in a couple of places is kind of a kludge. The syscall really expects `[]bpfInstruction` and `[]bpfFuncInfo`, etc. We've just decided to represent those as `[]byte` since we use binary.Write to marshal into the kernel representation. This is nice and safe, but in reality all the BPF UAPI types are manually laid out anyways, and we frequently just do `unsafe.Pointer(&sys.ProgLoadAttr)`.

I'm not at all convinced by my idea, but maybe we should just stick to "handwritten" structs instead? MarshalExtInfos would return `[]bpfLineInfo, []bpfFuncInfo` (naming to be bikeshedded, you get the idea) that we pass via `SlicePointer`. Those types need not have exported fields, which would keep them opaque from the users perspective. (Could also start generating func info and line info from BTF and stick them in `sys`?)

The upside is that we'd remove a bunch of allocations and reflection from commonly called code. The downside is that it seems like a big change just to get rid of `SliceElems` :D

> The overflow cutoff at maxu32 also feels somewhat arbitrarily tacked on without context around Func/LineInfoCnt and friends. If overflows are really a concern, newProgramWithOptions() should error instead of silently falling back to 0. EFAULT instead of EINVAL will still be a pain to debug.

There are three options:
1. Silently truncate as we do right now. The failure mode is that some calls will probably go through, other calls will fail in mysterious ways (thinking of func info / line info as you mentioned).
2. Silently use 0, as proposed by the PR. Calls fail reliably, but with a weird error.
3. Return an error, as proposed by you. Calls fail with a nice error.

I think the likelihood of someone running into this error is low, so I'm reluctant to go for (3) since it clutters call sites. Out of the remaining options 2 seems better than 1 to me.

internal/sys/ptr.go
lmb: Maybe you meant `SliceElems`? This is useful with other types, check `QueryPrograms` in query.go that calls it with `[]ProgramID`.

internal/sys/ptr.go
lmb: True about the stuttering! Pointer is a value type, so in my mind that means it doesn't allocate on it's own. Contrast that with NewSlicePointer, etc.

internal/sys/ptr.go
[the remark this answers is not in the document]
ti-mo: > We've just decided to represent those as []byte since we use binary.Write to marshal into the kernel representation.

This decision probably stems from reading files through io.Readers. Aren't C vs. Go struct alignment rules a potential show stopper here?

> The upside is that we'd remove a bunch of allocations and reflection from commonly called code.

(fyi, there's no such thing as `bpfInstruction`; `asm.Instruction` marshals directly to and from a binary stream)

For asm.Instruction specifically (our largest object pool by far), we could interpret a chunk of bytes as a hypothetical  `[]bpfInstruction` for a speed gain, but a copy is still needed into `asm.Instruction`. Whether we copy from `[]byte` or `[]bpfInstruction` makes no difference in that scenario. Also, instructions are read from files (thus io.Readers) in the majority of cases, not from existing memory.

> The downside is that it seems like a big change just to get rid of SliceElems

Could be worth it for clarity and overall maintainability. If we can generate types from uapi, that might be preferred over hand-written marshalers. Whether the output is []byte or []bpfFoo, allocs need to be made regardless, so performance would be equal to []byte at best.

> Out of the remaining options 2 seems better than 1 to me.

Sure, guess it doesn't matter that much.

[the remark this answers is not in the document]
lmb: > Aren't C vs. Go struct alignment rules a potential show stopper here?

They kind of are, but we already have that problem with structs in `sys`. I think if we use `gentypes` and add some additional tests we should be OK.

> Whether we copy from []byte or []bpfInstruction makes no difference in that scenario.

I'm specifically talking about `Instructions.Marshal`: instead of building []byte we'd assemble `[]sys.Instruction` (or whatever you want to call it). That avoids reflection and / or [manually writing the marshaler](https://github.com/cilium/ebpf/commit/36960939521f7570a134334bb4e69b79550a06c4).

lmb:
See #914 instead.

PR #910, opened 2023-01-21
bpf2go: spring cleaning

lmb:
Some light refactoring to clean up bpf2go a little.


cmd/bpf2go/main.go
```diff
@@ -192,13 +194,22 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
 
+	var g errgroup.Group
+	g.SetLimit(runtime.GOMAXPROCS(0))
+
 	for target, arches := range targets {
-		if err := b2g.convert(target, arches); err != nil {
-			return err
-		}
+		target, arches := target, arches // capture loop variables
+		g.Go(func() error {
```
lmb, line 202: Not sure this is really worth it.

cmd/bpf2go/main.go
```diff
@@ -192,13 +194,22 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
+			// Do some light stdout buffering to avoid concurrent runs from
+			// trampling on each other in the happy case.
+			stdout := bufio.NewWriter(stdout)
```
rgo3, line 205: How is using a buffer avoiding mixed output on stdout? Is it because we know that output from `convert` won't exceed buffer size, thus we can be sure all output is written in one go with `Flush`?

lmb: The idea is that in the happy (aka no errors case) convert will fit into the buffer, yes. If there is an error it will probably exceed the buffer and then the output of multiple targets will be mixed. It'll be mixed in chunks of buffer size though.

rgo3 (approved):
:broom: :sparkles:

PR #915, opened 2023-01-24
map: replace hacky unsafe.Pointer conversion with PROT_NONE page

lmb:
One of our legacy code paths currently does a dodgy pointer conversion:

    unsafe.Pointer(^uintptr(0))

This is (correctly) flagged by linters as being a bit bogus, since unsafe.Pointer conversions do have an effect on the GC (I believe). It also assumes that the highest address is never used.

Map an inaccessible page and use that instead of the bogus pointer. This has the same effect but is much safer.

Signed-off-by: Lorenz Bauer <oss@lmb.io>


lmb:
cc @nplanel

ti-mo (approved):
Nice! This has been bothering me since forever.

PR #916, opened 2023-01-25
Update CI dependencies

lmb:


.semaphore/semaphore.yml
```diff
@@ -23,13 +23,13 @@ blocks:
...
         commands:
           - sudo sh -c 'swapoff -a && fallocate -l 2G /swapfile && chmod 0600 /swapfile && mkswap /swapfile && swapon /swapfile'
-          - sem-version go 1.19.1
+          - sem-version go 1.19.5
           - export PATH="$PATH:$(go env GOPATH)/bin"
           - checkout
           # Disabled, see https://github.com/cilium/ebpf/issues/898
           # - cache restore
           - go mod tidy
-          - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.48.0
```
lmb, line 32: @ti-mo do you prefer sticking to `install.sh`? Seems superfluous if we have a go toolchain, but I'm probably missing something.

ti-mo: I vaguely remember some issues with a too-recent version of Go. If a new Go version drops and we're eager to bump it in CI, sometimes the golangci-lint release tags don't build from source with the newest toolchain, since they rely on some of the more obscure features of the stdlib. I only recall this being an issue when 1.18.0 dropped, but that was an extreme case of course.

Don't really care in general, but speed-wise install.sh should be faster, and there's no strong reason to build from source.

mmat11: this step seems to be failing in my pipeline -> https://ebpf.semaphoreci.com/workflows/3f7b2959-fa2c-47f3-a66f-bac69fbee7e4?pipeline_id=8f72e4bf-7f8c-4ba2-b18f-b8e92ec24a45
could this be related to the problem you are mentioning @ti-mo ?

PR #918, opened 2023-01-26
link: close perfEventLink.fd before perf event in Close()

ti-mo:
If a Kretprobe is created with RetprobeMaxActive > 0 on a kernel supporting bpf_link, a bpf_link is created based on a tracefs-backed perf event.

perfEventLink.Close() would try to remove the tracefs entry first, resulting in EBUSY because of the open link. This error, in turn, leaked the bpf_link fd. This also caused perf events to never be removed from tracefs/kprobe_events.

This commit closes the link and the perf event in the correct order.

Found by the fd tracer in https://github.com/cilium/ebpf/pull/732.

---

Unfortunately, this also means pinning a retprobe link with maxactive set means tracefs entries will leak by design. The only real solution I see is https://github.com/cilium/ebpf/pull/842.

### DIFF
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -136,10 +136,14 @@ func (pl *perfEventLink) Unpin() error {
 }
 
 func (pl *perfEventLink) Close() error {
-	if err := pl.pe.Close(); err != nil {
-		return fmt.Errorf("perf event link close: %w", err)
+	if err := pl.fd.Close(); err != nil {
+		return fmt.Errorf("perf link close: %w", err)
 	}
-	return pl.fd.Close()
+
+	if err := pl.pe.Close(); err != nil {
+		return fmt.Errorf("perf event close: %w", err)
+	}
+	return nil
 }
 
 func (pl *perfEventLink) Update(prog *ebpf.Program) error {
--- a/link/kprobe_test.go
+++ b/link/kprobe_test.go
@@ -72,25 +72,22 @@ func TestKprobeOffset(t *testing.T) {
 
 func TestKretprobeMaxActive(t *testing.T) {
 	prog := mustLoadProgram(t, ebpf.Kprobe, 0, "")
+	defer prog.Close()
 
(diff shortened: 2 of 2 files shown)


lmb:
One legitimate CI failure: https://ebpf.semaphoreci.com/jobs/97ac8e96-14ff-470f-8f67-6e581e07bf07#L756

link/perf_event.go
```diff
@@ -136,10 +136,15 @@ func (pl *perfEventLink) Unpin() error {
 }
 
 func (pl *perfEventLink) Close() error {
+	if err := pl.fd.Close(); err != nil {
+		return fmt.Errorf("perf link close: %w", err)
+	}
+
+	// This will fail with EBUSY if the link above was a pinned bpf_link.
```
lmb, line 143: From the code it seems like we don't support pinned perf events?

lmb:
> a bpf_link is created based on a tracefs-backed perf event.

And so far we only create bpf_link based on pmu perf event? Is the behaviour OK for link based on pmu?

> This also caused perf events to never be removed from tracefs/kprobe_events.

In general, or just for MaxActive > 0? If the latter it would be good to reword, the former interpretation would be pretty bad.

link/perf_event.go
mmat11: based on the commit message, the comment looks good if `s/pinned/tracefs-backed`

ti-mo: @lmb Perf events cannot be pinned, they are not bpf resources.
@mmat11 I'll make the comment a bit more elaborate, I really do mean 'pinned'. Closing a bpf_link's fd doesn't detach its event.

ti-mo:
> And so far we only create bpf_link based on pmu perf event?

Yes, that happened to be the case because kernels supporting bpf_link generally also support kprobe PMUs. 

>  Is the behaviour OK for link based on pmu?

Yes, closing a PMU that has an active bpf_link attached works just fine.

> In general, or just for MaxActive > 0?

Yes, only when maxactive > 0. Reworded.

PR #923, opened 2023-01-30
Fix pin rodata map error

zf1575192187:
1: fix pin .rodata map failed
2: fix populateMaps rodata map loaded from pinned file fail


chenhengqi:
What is your use-case ?

zf1575192187:
Pin out the map corresponding to some segments such as .rodata .bss and use it to share with other processes.

PR #930, opened 2023-02-08
map: ignore BPF_F_RDONLY_PROG for DevMap compatibility check

### LINKED ISSUE #925: Automatic pin reusing fails for devmaps
**Describe the bug**
Given a map of type `BPF_MAP_TYPE_DEVMAP_HASH` with pinning set to 1 loading it with `MapOptions.PinPath` set: It
works the first time when when there's no pinned file yet but fails with `use pinned map devmap: expected flags 0, got 128: map spec is incompatible with existing map`. 

**To Reproduce**
```go
// go version go1.19.5 linux/amd64
// github.com/cilium/ebpf@v0.10.0
//
// cat main_test.go
package main

import (
	"testing"

	"github.com/cilium/ebpf"
)

func TestMapReuse(t *testing.T) {
	spec := ebpf.MapSpec{
		Name:       "test",
		Type:       ebpf.DevMapHash,
		KeySize:    4,
		ValueSize:  4,
		MaxEntries: 1,
		Pinning:    ebpf.PinByName,
	}
	opts := ebpf.MapOptions{
		PinPath: "/sys/fs/bpf",
	}
	load := func() error {
		_, err := ebpf.NewMapWithOptions(&spec, opts)
		return err
	}

	if err := load(); err != nil {
		t.Fatalf("first load: %v", err)
	}

	if err := load(); err != nil {
		t.Fatalf("second load: %v", err)
	}
}
```

output:
```
=== RUN   TestMapReuse
    main_test.go:31: second load: creating map: use pinned map test: expected flags 0, got 128: map spec is incompatible with existing map
--- FAIL: TestMapReuse (0.05s)
FAIL

FAIL	ebpf-devmap-issue	10.114s
FAIL
```

**Expected behavior**
As with other types there should be no incompatibility error when reloading the exact same spec.

**Reasons for the flag change**
For devmaps the flag `BPF_F_RDONLY_PROG` (128) is set unconditionally for this type of map (see: https://github.com/torvalds/linux/blob/v6.1/kernel/bpf/devmap.c#L128). A workaround to manually set that flag does not work  (`field DevmapFlag: map devmap_flag: map create: invalid argument (without BTF k/v)`) .

**Possible solution**
For such unconditionally set flags, would it be an accepted change to handle that explicitly in `Map.Compatible`? At a first glance it appears only the types devmap and devmap_hash are affected. If it would be an accepted solution, I'd provide a PR.

aibor:
Allow DevMap and DevMapHash flags to differ in BPF_F_RDONLY_PROG.

Fixes #925.

### DIFF
--- a/map.go
+++ b/map.go
@@ -169,7 +169,10 @@ func (ms *MapSpec) Compatible(m *Map) error {
 		m.maxEntries != ms.MaxEntries:
 		return fmt.Errorf("expected max entries %v, got %v: %w", ms.MaxEntries, m.maxEntries, ErrMapIncompatible)
 
-	case m.flags != ms.Flags:
+	// BPF_F_RDONLY_PROG is set unconditionally for devmaps. Explicitly allow
+	// this mismatch.
+	case !((ms.Type == DevMap || ms.Type == DevMapHash) && m.flags^ms.Flags == unix.BPF_F_RDONLY_PROG) &&
+		m.flags != ms.Flags:
 		return fmt.Errorf("expected flags %v, got %v: %w", ms.Flags, m.flags, ErrMapIncompatible)
 	}
 	return nil
--- a/map_test.go
+++ b/map_test.go
@@ -621,6 +621,38 @@ func TestMapLoadPinned(t *testing.T) {
 	c.Assert(pinned, qt.IsTrue)
 }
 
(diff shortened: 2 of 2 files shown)


lmb (approved):

lmb:
Thanks for the detailed bug report + bugfix!

PR #932, opened 2023-02-09
program: set name from object info for pinned file

aibor:
When loading a program from a pinned file, the program name is set from the file name instead from the object info.

This commit adds a test that expects the program name to be set from the object info and changes the procedure to set the name from the object info, which is present already and used to set the program type.

Signed-off-by: Tobias Böhm <code@aibor.de>

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -783,7 +783,14 @@ func LoadPinnedProgram(fileName string, opts *LoadPinOptions) (*Program, error) 
 		return nil, fmt.Errorf("info for %s: %w", fileName, err)
 	}
 
-	return &Program{"", fd, filepath.Base(fileName), fileName, info.Type}, nil
+	var progName string
+	if haveObjName() == nil {
+		progName = info.Name
+	} else {
+		progName = filepath.Base(fileName)
+	}
+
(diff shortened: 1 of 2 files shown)


prog.go
```diff
@@ -783,7 +782,7 @@ func LoadPinnedProgram(fileName string, opts *LoadPinOptions) (*Program, error)
 		return nil, fmt.Errorf("info for %s: %w", fileName, err)
 	}
 
-	return &Program{"", fd, filepath.Base(fileName), fileName, info.Type}, nil
+	return &Program{"", fd, info.Name, fileName, info.Type}, nil
```
lmb, line 785: I think this would leave > 4.11 && < 4.15 without a name. Maybe keep filepath.Base if `!haveProgName()`?

lmb (changes requested):

prog.go
aibor: Good point. I changed my patch to check for the necessary feature and fall back to the file base name, just as before.

lmb (approved):

lmb:
Thanks :)

PR #936, opened 2023-02-16
docs: update various bits of documentation

lmb:
Adds a diagram to ARCHITECTURE.md and gets rid of the BTF section, which is orthogonal to how the library is architected. It existed mainly to explaint why the package wasn't exported, but that is now moot.

Make CONTRIBUTING.md shorter (do people read these things?) and give a list of steps how to contribute a feature.

Update README.md to mention the btf package.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,7 +1,18 @@
 Architecture of the library
 ===
 
-    ELF -> Specifications -> Objects -> Links
+```mermaid
+graph RL
+    Program --> ProgramSpec --> ELF
+    Map --> MapSpec --> ELF
+    Links --> Map & Program
+    subgraph Collection
+        Program & Map
+    end
+    subgraph CollectionSpec
+        ProgramSpec & MapSpec
+    end
+```
 
 ELF
 ---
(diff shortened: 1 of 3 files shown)


ARCHITECTURE.md
```diff
@@ -1,7 +1,18 @@
 Architecture of the library
 ===
 
-    ELF -> Specifications -> Objects -> Links
+```mermaid
```
lmb, line 4: Check out the rich diff!

ti-mo: Nice! Shouldn't the arrows be inverted? e.g. ELF -> ProgramSpec. That conveys the flow more naturally and reflects better what was written here before.

ARCHITECTURE.md
```diff
@@ -20,41 +31,32 @@ if possible.
...
+in Go. The returned `CollectionSpec` should be deterministic: reading the same ELF
+file on different systems must produce the same output.
+As a corollary, any changes that depend on the runtime environment like the
+current kernel version must happen when creating [Objects](#Objects).
 
 Specifications
 ---
 
-`CollectionSpec`, `ProgramSpec` and `MapSpec` are blueprints for in-kernel
+`CollectionSpec` is a very simple container for `ProgramSpec` and `MapSpec`. Avoid
```
ti-mo, line 42: I think this is missing `btf.Spec`. A paragraph that describes `CollectionSpec.Types` at a very high level would be nice to replace the BTF section that was removed above.

Something along the lines of CollectionSpec.Types allowing the spec's type info to be queried would be sufficient.

ti-mo (changes requested):
Thanks, left a few nits.

ARCHITECTURE.md
lmb: To me dependency order makes more sense, since that's also how the code ends up being structured. I kept ELF at the left since that makes it easier to explain however!

ti-mo (approved):

PR #944, opened 2023-02-23
Use FSType for tracefs mount detection

### LINKED ISSUE #934: link: use unix.Statfs to find debugfs, tracefs mount
Since #906 we try multiple locations to find a tracefs / debugfs mount, by os.Stat ing a file / directory. What we should do instead is use `unix.Statfs` and check `Statfs_t.Type`. We should extract the code we have for pinning into a helper and reuse that from `getTracefsPath`:
https://github.com/cilium/ebpf/blob/12702f3b089c3ffdfb80e8702956b33bd99145b0/internal/pinning.go#L25-L40

The magic numbers we need are documented here https://man7.org/linux/man-pages/man2/statfs.2.html and should be available from `unix`.

brycekahle:
Fixes #934

### DIFF
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -445,14 +445,13 @@ func isValidTraceID(s string) bool {
 // The available tracefs paths will depends on distribution choices.
 var getTracefsPath = internal.Memoize(func() (string, error) {
 	for _, p := range []struct {
-		file, path string
+		path   string
+		fsType int64
 	}{
-		// It's not enough to stat /sys/kernel/tracing, since some distros have
-		// a directory without a filesystem there.
-		{"/sys/kernel/tracing/kprobe_events", "/sys/kernel/tracing"},
-		{"/sys/kernel/debug/tracing", "/sys/kernel/debug/tracing"},
+		{"/sys/kernel/tracing", internal.TraceFSType},
+		{"/sys/kernel/debug/tracing", internal.DebugFSType},
 	} {
-		if _, err := os.Stat(p.file); err == nil {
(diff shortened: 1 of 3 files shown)


zhangzhiqiangcs:
Hi, @brycekahle @lmb I found there exists `/sys/kernel/debug/tracing` directory on my machine, but its filetype is tracefs, but not the expected debugfs type. what should I do to deal with this problem.
Thanks for any help.

link/perf_event.go
```diff
@@ -445,14 +445,13 @@ func isValidTraceID(s string) bool {
 // The available tracefs paths will depends on distribution choices.
 var getTracefsPath = internal.Memoize(func() (string, error) {
 	for _, p := range []struct {
-		file, path string
+		path   string
+		fsType int64
 	}{
-		// It's not enough to stat /sys/kernel/tracing, since some distros have
-		// a directory without a filesystem there.
-		{"/sys/kernel/tracing/kprobe_events", "/sys/kernel/tracing"},
```
ti-mo, line 452: Why was this line removed?

brycekahle:
@zhangzhiqiangcs I just looked at my mounts, and it is actually the same thing:
```
debugfs on /sys/kernel/debug type debugfs (rw,nosuid,nodev,noexec,relatime)
tracefs on /sys/kernel/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
tracefs on /sys/kernel/debug/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
```

I'll submit a PR to fix this issue

brycekahle:
https://github.com/cilium/ebpf/pull/958

PR #947, opened 2023-02-27
Support __kconfig macro for LINUX_KERNEL_VERSION

eiffel-fl:
Hi.

This supersedes #869.


ti-mo:
@eiffel-fl This is a PR from a fork in the `inspektor-gadget` org. It doesn't seem like 'allow updates from maintainers' is enabled, could you check? Need to push a small fixup to my first commit.

PR #949, opened 2023-02-27
Add an example for handling egress traffic via TC

fedepaol:
Adding an example where we use vishvananda/netlink to load an ebpf program as tc filter.
This particular program leverages the bpf_redirect_neigh to redirect egress traffic coming from 192.168.1.5 to an interface different from the default gateway.

I struggled a bit in finding a way to handle egress traffic via tc, hope this will help others. Also, this is my first interaction with ebpf so I might have done naive mistakes.

Thanks for the great library!


fedepaol:
I just realized there's one global go.mod file, and we can't add a dependency to netlink for the sake of the example, so we either provide a separate go.mod / go.sum file for the examples or this pr can't land as it is.

ti-mo:
Hi @fedepaol! Thanks for the proposal. `examples/` used to be a separate Go module, but that was causing pain when making breaking changes to bpf2go, so we decided to drop it and pull it into the main module. We've actively avoided adding a tc example due to the necessary library dependency.

Unless you're dealing with a legacy app that already depends heavily on `vishvananda/netlink`, I'd highly recommend you look at https://github.com/florianl/go-tc instead. It already depends directly on ebpf-go and comes with a clear example of how to do what you're proposing here: https://pkg.go.dev/github.com/florianl/go-tc#example-package-EBPF.

go-tc is built on https://github.com/mdlayher/netlink, see the diagram at the bottom for packages that interact with other netfilter families. Give those packages a try, I promise you'll enjoy the experience much better. :slightly_smiling_face:

If you hit another wall like this, especially since you're starting out, please don't hesitate to reach out in #ebpf-go on the Cilium Slack or skim our GH Discussions. It already contains many useful Q&A entries. I also covered this topic there in https://github.com/cilium/ebpf/discussions/769#discussioncomment-3481971.

Closing this for now.

cc @florianl

fedepaol:
I got this far because I skimmed both the discussions here and go-tc's one, and I have a working example with both the libraries, I resorted pushing an example using netlink because it's what cilium is currently using.
My two cents: I understand the dependency issues, but since this is THE place where to look for examples, a couple of lines in the README.md redirecting to @florianl 's repo would have been super helpful (and I'll see if I can push my example there, I still thing having a working egress example would help people). 

And again, I am writing because I struggled and I think this would help other people, on my side I (hope I) managed to find the right enchantment :-)

lmb:
> a couple of lines in the README.md redirecting to florianl's repo

SGTM!

ti-mo:
@florianl Since go-tc depends on ebpf-go, would there be a spot for a full BPF example including C in your repo? bpf2go can be used to keep things straightforward.

@fedepaol We can definitely add a link in examples/, that's not a problem. :slightly_smiling_face:

florianl:
[`cilium/ebpf`](https://github.com/cilium/ebpf/) only shows up as dependency of [`florianl/go-tc`](https://github.com/florianl/go-tc) as the [example for tc with eBPF](https://pkg.go.dev/github.com/florianl/go-tc#example-package-EBPF) is in a `_test.go` file and is executed with CI every time. To be fair, I would prefer to not add complexity and additional C code in [`florianl/go-tc`](https://github.com/florianl/go-tc). Adding the described example to [`florianl/go-tc`](https://github.com/florianl/go-tc) might solve it for this particular use case. But I think this will not scale for other similar requests as everything will be distributed and hard to find and detect for users.
E.g. with [`florianl/perf-ebpf`](https://github.com/florianl/perf-ebpf) I have started to show how the perf subsystem can be used with [`cilium/ebpf`](https://github.com/cilium/ebpf/). [`florianl/go-tc`](https://github.com/florianl/go-tc) and [`florianl/perf-ebpf`](https://github.com/florianl/perf-ebpf) are just two use cases.

Maybe we could add some kind of `showcases` sections where people can name such implementations and link to it? Even if I think having an example using `bpf2go` with a C file within [`florianl/go-tc`](https://github.com/florianl/go-tc) might not be the best case, I'm happy to create a repo to provide a proper show case and document it.

ti-mo:
I'm putting a TC example in the ebpf-go documentation project.

PR #951, opened 2023-02-28
Support __kconfig macro for LINUX_KERNEL_VERSION

eiffel-fl:
Hi.

This supersedes #869.


btf/types.go
```diff
@@ -655,18 +655,32 @@ func Sizeof(typ Type) (int, error) {
 	return 0, fmt.Errorf("type %s: exceeded type depth", typ)
 }
 
-// alignof returns the alignment of a type.
+// alignof returns the alignment of a type. Returns an error if the Type can't
```
lmb, line 658: Nit: new line.

btf/btf.go
```diff
@@ -487,6 +531,37 @@ func fixupDatasec(types []Type, sectionSizes map[string]uint32, offsets map[symb
 	return nil
 }
 
+// fixupDatasecLayout populates ds.Vars[].Offset according to var sizes and
+// alignment. Calculate and set ds.Size.
+func fixupDatasecLayout(ds *Datasec) error {
+	var off uint32
+
+	for i, vsi := range ds.Vars {
+		v := vsi.Type.(*Var)
```
lmb, line 540: This is not always a `Var`, I think it can be `Func` as well.

btf/btf.go
```diff
@@ -487,6 +531,37 @@ func fixupDatasec(types []Type, sectionSizes map[string]uint32, offsets map[symb
...
+		}
+		align, err := alignof(v.Type)
+		if err != nil {
+			return fmt.Errorf("variable %s: getting alignment: %w", v.Name, err)
+		}
+
+		// Calculate the padding (if any) between the end of the previous member
+		// and the start of the current member. Advance the offset, store it in vsi
+		// and increment by the current member's size.
+		off += padding(off, uint32(align))
```
lmb, line 553: I think you can replace this with `off = internal.Align(off, align)`. I've pushed a commit which makes that function generic.

btf/btf.go
```diff
@@ -487,6 +531,37 @@ func fixupDatasec(types []Type, sectionSizes map[string]uint32, offsets map[symb
...
+
+		ds.Vars[i].Offset = off
+
+		off += uint32(size)
+	}
+
+	ds.Size = uint32(off)
```
lmb, line 560: Idle curiosity: does the kernel require datasec size also have to be a multiple of 4 or something like that?

btf/btf.go
```diff
@@ -462,8 +488,26 @@ func fixupDatasec(types []Type, sectionSizes map[string]uint32, offsets map[symb
...
 		name := ds.Name
-		if name == ".kconfig" || name == ".ksyms" {
-			return fmt.Errorf("reference to %s: %w", name, ErrNotSupported)
+
+		// Some Datasecs are virtual and don't have corresponding ELF sections.
+		switch name {
+		case ".ksyms":
+			// .ksyms describes forward declarations of kfunc signatures.
+			// Nothing to fix up, all sizes and offsets are 0.
+			continue
```
lmb, line 497: Shouldn't we keep returning `ErrNotSupported` here?

lmb (changes requested):
Leaving some feedback for @ti-mo's part.

btf/btf.go
ti-mo: Yes, that's a change I actually wanted to push. :)

internal/align.go
```diff
@@ -1,6 +1,8 @@
 package internal
 
+import "golang.org/x/exp/constraints"
```
ti-mo, line 3: A little copying is better than a little dependency; can't we declare our own

```
type Integer interface {
	~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
}
```

for now until the dust settles and package constraints is promoted out of exp? Seems overkill to pull in the whole thing for a single constraint.

btf/types.go
```diff
@@ -655,18 +655,32 @@ func Sizeof(typ Type) (int, error) {
...
-		return int(t.size()), nil
+		n = int(t.size())
 	case *Int:
-		return int(t.Size), nil
+		n = int(t.Size)
 	default:
 		return 0, fmt.Errorf("can't calculate alignment of %T", t)
 	}
+
+	if !isPow(n) {
```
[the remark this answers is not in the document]
ti-mo: BTF allows encoding ints of any size, including odd-sized and zero-sized ints. Calculating the alignment of such types is not really possible as far as I understand. (https://en.wikipedia.org/wiki/Data_structure_alignment#Definitions)

Silently accepting this sounds like a bug; maybe not important in this patch set, but might go unnoticed in the future. `alignof()` was already fallible, so this made sense.

btf/btf.go
ti-mo: Cool, didn't know about that function. This works:

```
		off = internal.Align(off, uint32(align))
		ds.Vars[i].Offset = off
		off += uint32(size)
```

internal/align.go
lmb: The key advantage of x/exp is this:

> If future releases of Go add new predeclared integer types, this constraint will be modified to include them.

I don't mind much either way, I do think at some point we'll probably pull it in.

btf/btf.go
ti-mo: Indeed, looks like it's datasec -> vsi -> func for kfuncs. Will just `continue` here when the vsi target is not a var, since funcs don't have a size.

ti-mo: I ended up returning an error here, PTAL.

internal/align.go
ti-mo: True, it's likely this will get promoted out of exp in its current shape. I've amended your commit to make exp a direct dependency in go.mod, looks like you didn't `go tidy`.

btf/btf.go
ti-mo: Likely not, Datasec size can be 5 if it contains a u32 followed by a u8, for example.

lmb (commented):
LGTM!

PR #954, opened 2023-03-01
btf: work around kernel Datasec bug

### LINKED ISSUE #921: btf: vmlinux .data..percu Datasec is refused by the kernel
Since #859 TestRoundtripVMlinux sometimes fails. I've boiled the failure down to the kernel refusing the `.data..percpu` Datasec for an unknown reason.

```
          	[1] DATASEC .data..percpu size=205060 vlen=383
          		type_id=7811 offset=102416 size=2 Invalid type
```

lmb:
btf: work around kernel Datasec bug
    
    The kernel erroneously rejects Datasec where a Typedef, Volatile, Const,
    Restrict or typeTag follows a Pointer, Struct, Union or Array. Work
    around this by explicitly adding such types to the Spec first.
    
    The workaround only applies to BTF generated via Map.Value.
    Manually crafting a Datasec is not covered.
    
    Fixes #921
    
btf: fix marshaling of typeTag and declTag
    
    Marshaling of typeTag and declTag is currently broken. This is because
    we rely on marshaling vmlinux BTF for test coverage, which doesn't
    contain tags since we don't compile with clang.
    
cc @alban @mauriciovasquezbernal

### DIFF
--- a/btf/marshal.go
+++ b/btf/marshal.go
@@ -319,10 +319,13 @@ func (e *encoder) deflateType(typ Type) (err error) {
 
 	case *declTag:
 		raw.SetKind(kindDeclTag)
+		raw.SetType(e.id(v.Type))
 		raw.data = &btfDeclTag{uint32(v.Index)}
+		raw.NameOff, err = e.strings.Add(v.Value)
 
 	case *typeTag:
 		raw.SetKind(kindTypeTag)
+		raw.SetType(e.id(v.Type))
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal:
Great. It fixes the issues we have loading some of the programs of Inspektor Gadget. Thanks a lot for the fix!

ti-mo (approved):
Thanks!

PR #955, opened 2023-03-01
btf: preserve member order during marshaling

lmb:
Currently, we will marshal members of a composite type like Datasec, Struct, etc. in reverse order. The BTF ends up looking like this:

  [1] DATASEC a size=2 vlen=2
    type_id=7 offset=0 size=1
    type_id=4 offset=1 size=1

The types necessary for Datasec.Vars[1] are marshaled before the ones for Datasec.Vars[0]. Change the implementation so that the order is reversed. The resulting BTF is now:

  [1] DATASEC a size=2 vlen=2
    type_id=4 offset=0 size=1
    type_id=7 offset=1 size=1

### DIFF
--- a/btf/traversal.go
+++ b/btf/traversal.go
@@ -37,7 +37,7 @@ func postorderTraversal(root Type, skip func(Type) (skip bool)) postorderIterato
 	}
 
 	po := postorderIterator{root: root, skip: skip}
-	walkType(root, po.push)
+	walkType(root, po.push, true)
 
 	return po
 }
@@ -73,7 +73,7 @@ func (po *postorderIterator) Next() bool {
 			po.walked.Push(true)
 
 			// Add all direct children to todo.
-			walkType(*t, po.push)
+			walkType(*t, po.push, true)
 		} else {
 			// We've walked this type previously, so we now know that all
 			// children have been handled.
@@ -88,7 +88,7 @@ func (po *postorderIterator) Next() bool {
 }
 
 // walkType calls fn on each child of typ.
-func walkType(typ Type, fn func(*Type)) {
+func walkType(typ Type, fn func(*Type), reverse bool) {
 	// Explicitly type switch on the most common types to allow the inliner to
 	// do its work. This avoids allocating intermediate slices from walk() on
 	// the heap.
@@ -101,12 +101,24 @@ func walkType(typ Type, fn func(*Type)) {
 		fn(&v.Index)
 		fn(&v.Type)
 	case *Struct:
-		for i := range v.Members {
-			fn(&v.Members[i].Type)
+		if !reverse {
+			for i := range v.Members {
+				fn(&v.Members[i].Type)
+			}
+		} else {
+			for i := len(v.Members) - 1; i >= 0; i-- {
+				fn(&v.Members[i].Type)
+			}
 		}
 	case *Union:
-		for i := range v.Members {
-			fn(&v.Members[i].Type)
+		if !reverse {
+			for i := range v.Members {
+				fn(&v.Members[i].Type)
+			}
+		} else {
+			for i := len(v.Members) - 1; i >= 0; i-- {
+				fn(&v.Members[i].Type)
+			}
 		}
 	case *Typedef:
 		fn(&v.Type)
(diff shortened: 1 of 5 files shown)


btf/traversal.go
```diff
@@ -88,7 +88,7 @@ func (po *postorderIterator) Next() bool {
 }
 
 // walkType calls fn on each child of typ.
-func walkType(typ Type, fn func(*Type)) {
+func walkType(typ Type, fn func(*Type), reverse bool) {
```
lmb, line 91: Not a big fan of the bool, but the only other option I see is just duplicating the function.

btf/traversal.go
```diff
@@ -120,15 +132,24 @@ func walkType(typ Type, fn func(*Type)) {
...
+		} else {
+			for i := len(v.Params) - 1; i >= 0; i-- {
+				fn(&v.Params[i].Type)
+			}
 		}
 	case *Var:
 		fn(&v.Type)
 	case *Datasec:
-		for i := range v.Vars {
+		for i := len(v.Vars) - 1; i >= 0; i-- {
```
rgo3, line 147: Why is this always flipped and not just when reverse=true?

btf/traversal.go
```diff
@@ -120,15 +132,24 @@ func walkType(typ Type, fn func(*Type)) {
...
 			fn(&v.Vars[i].Type)
 		}
+		// for i := range v.Vars {
```
ti-mo, line 150: Leftover

btf/traversal.go
```diff
@@ -101,12 +101,24 @@ func walkType(typ Type, fn func(*Type)) {
 		fn(&v.Index)
 		fn(&v.Type)
 	case *Struct:
-		for i := range v.Members {
-			fn(&v.Members[i].Type)
+		if !reverse {
```
ti-mo, line 104: The double negative and inverted order throws me off here. To me,
```go
if reverse {
    for i := ... {}
    break
}
```

is clearer than if/else with a double negative at the top.

btf/traversal.go
ti-mo: Wild guess is this is somehow related to #954, but seems like a curious choice on the surface.

btf/traversal.go
ti-mo: Fine IMO since the function has a narrow purpose and the bool is mostly self-explanatory. I'd add it to the function doc that it only applies to the order of complex types, though.

ti-mo (changes requested):
I'm missing some context around why this is necessary and why we don't simply pick one order over the other across the board.

The commit message is also a bit misleading:

```
    [1] DATASEC a size=2 vlen=2
      type_id=7 offset=0 size=1
      type_id=4 offset=1 size=1

vs.

    [1] DATASEC a size=2 vlen=2
      type_id=4 offset=0 size=1
      type_id=7 offset=1 size=1
```

On first read, I thought we're now ordering datasec members differently, which would be an error, since the order matters and we can't just decide to shuffle them. The example being one-sized ints doesn't illustrate what actually changes in the type blob.

It also mentions:

```
The types necessary for Datasec.Vars[1] are marshaled before
the ones for Datasec.Vars[0]. Change the implementation so that
the order is reversed.
```

Yet the effect of the change isn't obvious to me; I assume this change would only cause the underlying types of id's 4 and 7 to appear in a different order in the blob, but why does this matter? Aren't type IDs allocated on spec.Add()? How come they are different after this change, and doesn't that break everything?

btf/traversal.go
lmb: My wild guess is that I should've marked this as draft to start with :D

btf/traversal.go
```diff
@@ -139,3 +162,15 @@ func walkType(typ Type, fn func(*Type)) {
 		panic(fmt.Sprintf("don't know how to walk Type %T", v))
 	}
 }
+
+func walkMembers(members []Member, mode walkMode, fn func(*Type)) {
```
rgo3, line 166: Would have been nice to somehow use this for slices of `Member`, `FuncParam` and `Var` but I see that making that work wouldn't really be worth the effort.

rgo3 (commented):
Is there a way for me to see what this change does in practice? As in where would I make use of the improved readability for humans?

btf/traversal.go
lmb: We could make it work by having all those three implement an interface, and then use that as a constraint to a generic `walkSlice[T walkable]([]T, mode walkMode, func(*Type))`. The downside is that the generic function wouldn't be inlined probably, and that `walkSlice[Member]` uses dynamic instead of static dispatch. See https://planetscale.com/blog/generics-can-make-your-go-code-slower for an explanation. Feel free to give it a try if you want to, I think it'll probably end up being a wash in terms of speed / lines of code.

lmb:
> As in where would I make use of the improved readability for humans?

It makes the verifier output more readable when loading BTF into the kernel fails for some reason. So if you printed the error from `NewHandle` to stdout it'd be easier to follow what is going on I believe.

Maybe this change isn't worth it, happy to close the PR.

PR #958, opened 2023-03-03
Fix expected fs type for /sys/kernel/debug/tracing

brycekahle:
It turns out the subfolder `/sys/kernel/debug/tracing` is mounted separately with the `tracefs` fs type.

Trimmed output of `mount`:
```
debugfs on /sys/kernel/debug type debugfs (rw,nosuid,nodev,noexec,relatime)
tracefs on /sys/kernel/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
tracefs on /sys/kernel/debug/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
```


internal/statfs.go
```diff
@@ -9,7 +9,6 @@ import (
 const (
 	BpfFSType   = 0xcafe4a11
 	TraceFSType = 0x74726163
```
lmb, line 11: Now that we're here, use `unix.TRACEFS_MAGIC` and `unix.BPF_FS_MAGIC`?

internal/statfs_test.go
```diff
@@ -0,0 +1,16 @@
+package internal
+
+import "testing"
+
+func TestFSType(t *testing.T) {
```
lmb, line 5: This test doesn't seem particularly useful, as in in doesn't test getTracefsPath. How about checking that if err != nil, neither paths exist?

lmb (changes requested):

internal/statfs_test.go
brycekahle: I can remove the test. I was mostly using it as local validation that the code returned what was expected.

lmb (approved):
Thanks!

I can't push a fixup since I can't access `datadog/ebpf`. Please create the PR from a personal repo next time.

PR #959, opened 2023-03-06
link: always include requested symbol in kprobe creation errors

### LINKED ISSUE #956: Improve error logs when retrying to attach a kprobe with an arch specific prefix
The PR https://github.com/cilium/ebpf/pull/304 added a mechanism to retry on a kprobe symbol in case it was not found, typically adding an arch-specific prefix to the symbol. It's a good idea since it allows architecture-independent symbols for syscalls while on the downside creating an implicit alias for some symbols (on amd64 you end up having `cilium/ebpf` "exposing" `__x64_sys_listen` and `sys_listen` that target the same `__x64_sys_listen` symbol).

However, the error message is sometimes a bit puzzling because you wonder when trying `__x64_sys_listen` on arm64, you end up with an error like:
```
creating perf_kprobe PMU: token __arm64___x64_sys_listen: not found: no such file or directory
```

Maybe it would be possible to include in the error message that `__x64_sys_listen` was first tried (which was the user input) and then `__arm64___x64_sys_listen`, which could be the arch-specific symbol the user was referring to.

Related to https://github.com/cilium/ebpf/issues/266.

ti-mo:
Before, when a caller would try to set a kprobe on e.g. __x64_sys_listen, the error message would look something like 'creating perf_kprobe PMU: token __arm64___x64_sys_listen: not found: no such file or directory'. This doesn't mention that the requested symbol has been attempted in addition to the prefixed symbol mentioned in the error message.

If a symbol doesn't exist, which is going to be the vast majority of cases, the caller will only see the prefixed symbol echoed in the error message, which can make them wonder if the original symbol has been attempted at all.

This patch makes PMU and tracefs errors more uniform and always mentions the requested symbol in the error string, which only used to happen for tracefs.

Fixes #956

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -91,6 +91,10 @@ func (pt probeType) PerfEventType(ret bool) perfEventType {
 // Losing the reference to the resulting Link (kp) will close the Kprobe
 // and prevent further execution of prog. The Link must be Closed during
 // program shutdown to avoid leaking system resources.
+//
+// If attaching to symbol fails, automatically retries with the running
+// platform's syscall prefix (e.g. __x64_) to support attaching to syscalls
+// in a portable fashion.
 func Kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions) (Link, error) {
 	k, err := kprobe(symbol, prog, opts, false)
 	if err != nil {
@@ -115,6 +119,10 @@ func Kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions) (Link, error
 // Losing the reference to the resulting Link (kp) will close the Kretprobe
 // and prevent further execution of prog. The Link must be Closed during
 // program shutdown to avoid leaking system resources.
+//
+// If attaching to symbol fails, automatically retries with the running
+// platform's syscall prefix (e.g. __x64_) to support attaching to syscalls
+// in a portable fashion.
 //
 // On kernels 5.10 and earlier, setting a kretprobe on a nonexistent symbol
 // incorrectly returns unix.EINVAL instead of os.ErrNotExist.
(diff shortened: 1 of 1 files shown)


mtardy (approved):
Thanks, that would be lovely! 😸

So now it will look like this:
```text
creating perf_kprobe PMU (tried '_x64_sys_listen'): token __arm64___x64_sys_listen: not found: no such file or directory
```
It still requires knowing that the lib is retrying on fail with an arch-specific prefix but at least the original symbol appears in the error which is excellent!

link/kprobe.go
```diff
@@ -199,7 +199,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
 		return tp, nil
 	}
 	if err != nil && !errors.Is(err, ErrNotSupported) {
-		return nil, fmt.Errorf("creating perf_kprobe PMU: %w", err)
+		return nil, fmt.Errorf("creating perf_kprobe PMU (tried '%s'): %w", symbol, err)
```
lmb, line 202: Maybe it's better to just return the original error?

```go
origErr := pmuKprobe(...)

fallbackErr := pmuKprobe(...)
if fallbackErr != nil { return origErr }
```

mtardy: The best would be both, being explicit about the fact that symbol1 and adaptedSymbol1 failed. Losing the fallbackErr could make the behavior not very clear I think.

mmat11 (approved):

link/kprobe.go
```diff
@@ -199,7 +207,7 @@ func kprobe(symbol string, prog *ebpf.Program, opts *KprobeOptions, ret bool) (*
...
```
lmb, line 210:
```suggestion
		return nil, fmt.Errorf("creating perf_kprobe PMU (arch specific fallback for %q): %w", symbol, err)
```

lmb (approved):

PR #960, opened 2023-03-06
collection: Add code to parse kernel config files.

eiffel-fl:
This code uses existing golang package and writes the corresponding value to .kconfig section [1].
It is mainly a translation from C of libbpf code.

Signed-off-by: Francis Laniel <flaniel@linux.microsoft.com>
[1] https://github.com/ti-mo/kconfig


collection.go
```diff
@@ -582,6 +586,180 @@ func (cl *collectionLoader) populateMaps() error {
...
+
+	str := strings.Trim(value, `"`)
+
+	// We need to trim string if the bpf array is smaller.
+	if uint32(len(str)) >= array.Nelems{
+		str = str[:array.Nelems]
+	}
+
+	// Write the string content to .kconfig.
+	for i, c := range str {
```
lmb, line 698: You can just copy: `copy(data, str)`

collection.go
```diff
@@ -582,6 +586,180 @@ func (cl *collectionLoader) populateMaps() error {
...
+	case 8:
+		internal.NativeEndian.PutUint64(data, n)
+	default:
+		return fmt.Errorf("size (%d) is not valid, expected: 1, 2, 4 or 8", size)
+	}
+
+	return nil
+}
+
+func addKconfigValue(data []byte, typ btf.Type, value string) error {
```
lmb, line 751: Nit: functions that encode into a pre-allocated slice are usually called `Put...` like `NativeEndian.PutUint16`. I think that would make sense here.

lmb (commented):
Sorry for the long delay. I think the approach is fine, we should just copy in Timo's code and improve on it. `collection.go` is already pretty large, how about you put the new code into a package `internal/kconfig` or something like that?

We can probably get away with two (exported) functions:

```go
func ParseKconfig(file string) (map[string]string, error)
func PutKconfigValue([]byte, btf.Type, value string) error
```

Have you thought about how you're going to test this? We need coverage of the parser and of programs referencing the various kinds of .kconfig (libbpf_tristate, regular bool, int, etc.).

collection.go
```diff
@@ -582,6 +586,180 @@ func (cl *collectionLoader) populateMaps() error {
...
+
+func parseKconfig() (kconfig.Kconfig, error) {
+	kernelRelease, err := getKernelRelease()
+	if err != nil {
+		return kconfig.Kconfig{}, fmt.Errorf("cannot get kernel release: %w", err)
+	}
+
+	path := fmt.Sprintf("/boot/config-%s", kernelRelease)
+
+	ret := kconfig.New()
```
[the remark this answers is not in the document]
ti-mo: Already gave my permission offline, but for the record: go ahead! :slightly_smiling_face:

collection.go
```diff
@@ -578,17 +579,55 @@ func (cl *collectionLoader) populateMaps() error {
...
+			return fmt.Errorf("cannot add kconfig patch: %w", err)
+		}
+	}
+
 	data := make([]byte, ds.Size)
 	for _, vsi := range ds.Vars {
 		v := vsi.Type.(*btf.Var)
+		n := v.TypeName()
+
+		if strings.HasPrefix(n, "CONFIG_") {
```
lmb, line 616: Why do you allow kconfig to override "specials" like `LINUX_VERSION_CODE`? I would expect it the other way around.

testdata/kconfig_config.c
```diff
@@ -0,0 +1,73 @@
+#include "common.h"
+
+char __license[] __section("license") = "MIT";
+
+// Highly inspired from libbpf test:
```
lmb, line 5: In that case the test would have to be GPL-2.0 as per the license header at the top of the file?

testdata/kconfig_config.c
```diff
@@ -0,0 +1,73 @@
...
+// https://github.com/torvalds/linux/blob/6e98b09da931a00bf4e0477d0fa52748bf28fcce/tools/testing/selftests/bpf/progs/test_core_extern.c
+// The modifications are the following:
+// 1. Removed __weak as cilium/ebpf does not support it.
```
lmb, line 8: Just to make sure we don't introduce behaviour that is incompatbile with adding __weak later: what does libbpf do when it finds a symbol that is marked weak but doesn't exist?

testdata/kconfig_config.c
```diff
@@ -0,0 +1,73 @@
...
+	val = CONFIG_ULONG;
+	bpf_map_update_elem(&array_map, &i, &val, 0);
+
+	char str[8];
+	for (i = 0; i < sizeof(CONFIG_STR); i++) {
+		str[i] = CONFIG_STR[i];
+	}
+
+	i = 6;
+	__builtin_mempcpy(&val, str, sizeof val);
```
lmb, line 64: Why can't you copy directly from `CONFIG_STR` into `val`?

testdata/kconfig_config.c
```diff
@@ -0,0 +1,73 @@
...
+	bpf_map_update_elem(&array_map, &i, &val, 0);
+
+	i = 7;
+	if (!CONFIG_MISSING)
```
lmb, line 68: How is missing different than `CONFIG_BOOL`?

lmb (changes requested):
I think the important bits are here! Some thoughts:

- Maybe I've mentioned this before, but I'm not a big fan of integration tests. They end up being cumbersome to maintain and usually don't cover a lot of the edge cases. For example, we currently don't test for min/max values of integers, all possible bools, tristates, etc. Doing that through the ELF reader is a big bore. Why not split the problem into two? First, add tests for the various `putKconfigValue...` in package `kconfig`. Make sure we handle out of bounds values correctly. Second, extend the existing kconfig ELF reader test to pull in a very common, stable kconfig like `CONFIG_HZ`. That gets rid of all of the `AdditionalKconfig` mess and also allows us to drop `PatchKconfig`.
- Can you add a small benchmark to measure how expensive parsing kconfig is? We currently parse it even if the BPF doesn't end up requiring it. I think that's the correct choice as long as parsing is not super slow.

collection.go
```diff
@@ -582,6 +586,180 @@ func (cl *collectionLoader) populateMaps() error {
...
+// We need to translate C enum to golang.
+type triState int
+
+const (
+	triNo     triState = 0
+	triYes    triState = 1
+	triModule triState = 2
+)
+
+func addKconfigValueTri(data []byte, typ btf.Type, value string) error {
```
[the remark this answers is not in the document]
lmb: I only understood what you meant after I read through the code: a kconfig can be a tristate, an integer or a string. It's therefore easiest to represent it as a string most of the time.

lmb:
> 2\. I get the following error: `load BTF: variable CONFIG_STR: getting alignment: can't calculate alignment of *btf.Array`.

That's because `alignof` doesn't understand arrays: https://github.com/cilium/ebpf/blob/d8d898ac05b27d8ea71053321d97cd92ba53ae84/btf/types.go#L655-L662 This might be enough to fix things:
```patch
diff --git a/btf/types.go b/btf/types.go
index 3fb5499..38cc76a 100644
--- a/btf/types.go
+++ b/btf/types.go
@@ -657,6 +657,8 @@ func alignof(typ Type) (int, error) {
                n = int(t.size())
        case *Int:
                n = int(t.Size)
+       case *Array:
+               return alignof(t.Type)
        default:
                return 0, fmt.Errorf("can't calculate alignment of %T", t)
        }
```

internal/kconfig/test.kconfig
```diff
@@ -0,0 +1,11 @@
+CONFIG_TRISTATE=m
```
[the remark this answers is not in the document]
lmb: This file is used in unit tests. By convention, such files are put into a `testdata` directory.

internal/kconfig/kconfig.go
```diff
@@ -0,0 +1,247 @@
...
+		return putKconfigValueNumber(data, typ, value)
+	}
+}
+
+// Golang translation of libbpf_tristate enum:
+// https://github.com/libbpf/libbpf/blob/fbd60dbff51c870f5e80a17c4f2fd639eb80af90/src/bpf_helpers.h#L169
+type TriState int
+
+const (
+	TriNo     TriState = 0
```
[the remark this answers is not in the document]
lmb: Let's keep it as it is!

In general `libbpf` has different constraints than we do. C doesn't have name spacing like Go does for example. We should have the same semantics as libbpf, but we shouldn't follow the coding style 1:1.

collection.go
```diff
@@ -578,17 +579,55 @@ func (cl *collectionLoader) populateMaps() error {
...
+	}
+
+	path := fmt.Sprintf("/boot/config-%s", kernelRelease)
+
+	config, err := kconfig.ParseKconfig(path, false)
+	if err != nil {
+		config, err = kconfig.ParseKconfig("/proc/config.gz", true)
+		if err != nil {
+			// If there is no file providing kconfig, we just default to empty one.
+			config = make(map[string]string)
```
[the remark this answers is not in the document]
lmb: I guess we end up returning an error if the instructions contain an unknown config? That should be fine.

ti-mo:
> We currently parse it even if the BPF doesn't end up requiring it. I think that's the correct choice as long as parsing is not super slow.

@lmb Unless it significantly increases code complexity, a user shouldn't need to pay for a feature they don't use. Is there a reason we can't defer parsing kconfig until a lookup is needed?

> I am wondering about reducing the memory occupancy by only having a map containing the CONFIG_* used in the eBPF code rather than all the kernel config.

Sure, but then we might as well not cache the result of the parser at all; you're basically describing populating .ksyms and throwing away the result. I'd say leave this out of scope of this PR. We can experiment with replacing the map with a minimal trie implementation that only supports inserts and lookups. (can be done in 30-40 lines of Go) This may cause more total allocs, but should result in a lower steady-state of consumed memory. Benchmarking is the only way to really know. :) I'd make sure we don't parse kconfig if the caller doesn't need it in any case.

map.go
```diff
@@ -78,6 +78,8 @@ type MapSpec struct {
 
 	// The key and value type of this map. May be nil.
 	Key, Value btf.Type
+
+	AdditionalKconfig string
```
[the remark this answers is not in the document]
lmb: 1. I think you can write an integration test without AdditionalKconfig, using `CONFIG_HZ` or similar, like I mentioned.
2. If we end up supporting something like AdditionalKconfig we shouldn't do it using `string`, but maybe a map or similar. But as long as the only use case is integration testing, we shouldn't add it at all.

collection.go
```diff
@@ -578,17 +579,32 @@ func (cl *collectionLoader) populateMaps() error {
...
 	if !ok {
 		return errors.New("map value is not a Datasec")
 	}
 
+	// If there is no file providing kconfig, we just default to empty one.
+	config := make(map[string]string)
+	scanner, err := kconfig.Find()
+	if err == nil {
+		config, _ = kconfig.Parse(scanner)
+	}
```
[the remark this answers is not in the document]
lmb: Why not return it from the function?

internal/kconfig/kconfig.go
```diff
@@ -0,0 +1,265 @@
...
+
+	"github.com/cilium/ebpf/btf"
+	"github.com/cilium/ebpf/internal"
+)
+
+// Find find a kconfig file on the host.
+// It first reads from /boot/config- of the current running kernel and tries
+// /proc/config.gz if nothing was found in /boot.
+// If none of the file provide a kconfig, it returns an error.
+func Find() (*bufio.Scanner, error) {
```
lmb, line 19: Doesn't this leak the file? I think it's better to return `*os.File` here.

internal/kconfig/kconfig.go
```diff
@@ -0,0 +1,265 @@
...
+
+	kernelRelease, err := internal.KernelRelease()
+	if err != nil {
+		return nil, fmt.Errorf("cannot get kernel release: %w", err)
+	}
+
+	path := fmt.Sprintf("/boot/config-%s", kernelRelease)
+	f, err := os.Open(path)
+	if err != nil {
+		f, err = os.Open("/proc/config.gz")
```
lmb, line 30: This would be easier to do in Parse I think, by making it take `io.ReaderAt`. Then you can pass `io.NewLimitReader(ReaderAt, math.MaxUint64)`to the gzip reader.

lmb: I pushed a commit, please take a look.

internal/kconfig/kconfig.go
```diff
@@ -0,0 +1,249 @@
...
+	if !found {
+		return fmt.Errorf("line %q does not contain separator '='", line)
+	}
+
+	if len(value) == 0 {
+		return fmt.Errorf("line %q has no value", line)
+	}
+
+	_, ok := m[key]
+	if !ok {
```
lmb, line 97: Isn't this the opposite of what kconfig does? In my experience the last value wins.

Should we allow overwriting values at all?

internal/kconfig/kconfig.go
```diff
@@ -0,0 +1,249 @@
...
+		if strings.HasPrefix(value, `"`) {
+			return putValueString(data, typ, value)
+		}
+		return putValueNumber(data, typ, value)
+	}
+}
+
+// Golang translation of libbpf_tristate enum:
+// https://github.com/libbpf/libbpf/blob/fbd60dbff51c870f5e80a17c4f2fd639eb80af90/src/bpf_helpers.h#L169
+type TriState int
```
lmb, line 122: Does this need to be exported actually?

collection.go
```diff
@@ -584,11 +585,22 @@ func resolveKconfig(m *MapSpec) error {
 		return errors.New("map value is not a Datasec")
 	}
 
+	scanner, err := kconfig.Find()
+	if err != nil {
+		return fmt.Errorf("cannot find a kconfig file: %w", err)
+	}
+
+	config, err := kconfig.Parse(scanner)
```
lmb, line 593: I think Timo is right, it would be nice if we only parsed the kconfig when required. 
I've updated the benchmark you added to use a bigger file and looked at the results very briefly: it takes around 1ms on my machine.

What if you did the following?
* in the default switch case collect the necessary config items into `map[string][]byte` or similar
* after we've iterated the items once, parse kconfig `if len(map[string][]byte) > 0` and fill in the values

That should be doable without too many code changes.

Coming back to the benchmark, most of the time is spent constructing the full `map[string]string`. That's fine for now, since it's nice and simple. But later on we could experiment using the `map[string][]byte` to only parse the lines we're actually interested in.

lmb (changes requested):
Looks great, especially the tests are really thorough! Please take a look at my suggestion to only parse when necessary and let me know what you think.

I've rebuilt all of our CI kernels so that /proc/config.gz is now a thing.

internal/kconfig/kconfig.go
lmb: Ah weird. Can you add a comment to that effect?

collection.go
```diff
@@ -613,8 +622,39 @@ func resolveKconfig(m *MapSpec) error {
...
+			configs[n] = configInfo{
+				offset: vsi.Offset,
+				typ:    v.Type,
+			}
+		}
+	}
+
+	// We only parse kconfig file if a CONFIG_* variable was found.
+	if len(configs) > 0 {
+		scanner, err := kconfig.Find()
```
lmb, line 635: What I meant earlier: this function doesn't return a `scanner` anymore, so calling this `f` or similar is clearer.

collection.go
```diff
@@ -613,8 +622,39 @@ func resolveKconfig(m *MapSpec) error {
...
+			return fmt.Errorf("cannot find a kconfig file: %w", err)
+		}
+		defer scanner.Close()
+
+		kernelConfig, err := kconfig.Parse(scanner)
+		if err != nil {
+			return fmt.Errorf("cannot parse kconfig file: %w", err)
+		}
+
+		for n := range configs {
```
lmb, line 646: Why not `n, info :=`?

lmb (changes requested):
Just some nits. Can you please update the commit messages? The first commit message describes the old API and the second one doesn't really exist :) So far we've also not used a `.` at the end of commit titles, I think it makes sense to stick to that. Can you please remove the `.`? You can squash your (first) two commits if you'd like, then there is only one message to write. Maybe you could include that we don't support weak symbols at the moment, or other things which we don't support yet?

collection.go
lmb: Makes sense, sorry if I didn't see that. I think we might also be able to cut down memory usage to a point where we can afford to keep a cached map around.

lmb:
Thanks for tidying up! I've removed the two commits since I don't have the time to review them right now, sorry. Let's get this merged and then we can iterate on the performance.

PR #965, opened 2023-03-08
perf: use the min value of maxentries and possibles cpus to create ring

chengzhycn:
When bpf program is static compiled, the perf map located in the obj file should use a static value as its max_elem value. If this value is bigger than the number of possible cpus, create a new reader will fail when calling unix.PerfEventOpen.


lmb:
Hi, you're probably setting something like `max_elem = 10`, and that breaks? If yes, you don't need to hardcode a static value. Simply keep max_elems at 0 and the library will fill in the correct number of CPUs for you. See https://github.com/cilium/ebpf/commit/c60ad69fa195c59c1a3c340c88001b82af273473 for some context why things are this way.

PR #966, opened 2023-03-08
elf_reader, linker: add kfunc support

rgo3:
```
This commit adds support for handling the following extern 
kfunc declaration in bpf programs:

extern void bpf_kfunc_foo(void *mem) __ksym;

If during relocation a call to a kfunc is found we set the corresponding 
btf.Func as Metadata on the instruction and change Instruction.Src to 
asm.PseudoKfuncCall.

fixupKfunc() has been added to the linker, to find the type id of the 
btf.Func in the running kernel and set Instruction.Constant to the found 
btf id.
```


elf_reader.go
```diff
@@ -585,6 +587,40 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 			return fmt.Errorf("asm relocation: %s: unsupported type %s", name, typ)
 		}
 
+		// If a Call instruction is found and the datasec has a btf.Func with a Name
+		// that matches the symbol name we mark the instruction as a call to a kfunc.
+		if ins.OpCode.JumpOp() == asm.Call {
```
lmb, line 592: Maybe turn this into a switch together with the `ins.OpCode.IsDWordLoad()` check below?

I think it'd also be good to label the break statements here to be more explicit.

elf_reader.go
```diff
@@ -585,6 +587,40 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
+			if ec.btf == nil {
+				break
+			}
+			var ds *btf.Datasec
```
lmb, line 596: Can you refactor this into a method on `elfCode`? `elfCode.findKfunc(name string) (btf.Type, error)` or similar. That function can return `(nil, nil)` to indicate "no BTF or not a kfunc call".

Patching of ins would still happen here though.

elf_reader.go
```diff
@@ -585,6 +587,40 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
+				ins.Metadata.Set(kfuncMeta{}, fn)
+				ins.Src = asm.PseudoKfuncCall
+				ins.Constant = -1
+
+				found = true
+				break
+			}
+
+			if !found {
+				return fmt.Errorf("kfunc %s not found in %s", rel.Name, ".ksyms")
```
lmb, line 618: I think this shouldn't be an error, otherwise we break the non `.ksyms` extern hack. We should just fall through to the end of the loop.

linker.go
```diff
@@ -185,12 +185,53 @@ func fixupAndValidate(insns asm.Instructions) error {
...
+		return fmt.Errorf("couldn't load kernel spec")
+	}
+
+	var f *btf.Func
+	if err := s.TypeByName(kfm.Name, &f); err != nil {
+		return fmt.Errorf("couldn't find kfunc in kernel spec")
+	}
+
+	// we should compare args and return type here
+	// coreAreTypesCompatible(kfm.Type, fn.Type) ?
```
lmb, line 222: Yes, I think we'll have to export that function. How about `btf.CheckCompatibility(Type, Type) error`?

The big comment at the top of the function should move into the function body, it's too much implementation detail. We'll also have to replace `errImpossibleRelocation` with a more meaningful error, maybe even an exported one.

linker.go
```diff
@@ -185,12 +185,53 @@ func fixupAndValidate(insns asm.Instructions) error {
...
+
+	id, err := s.TypeID(f)
+	if err != nil {
+		return fmt.Errorf("couldn't find type id")
```
lmb, line 226: Droppity drop

lmb (changes requested):
Very nice, things slot into place really well and it has a small scope. Thanks!

linker.go
```diff
@@ -185,12 +185,53 @@ func fixupAndValidate(insns asm.Instructions) error {
...
+	}
+
+	s, err := btf.LoadKernelSpec()
+	if err != nil {
+		return fmt.Errorf("couldn't load kernel spec")
+	}
+
+	var f *btf.Func
+	if err := s.TypeByName(kfm.Name, &f); err != nil {
+		return fmt.Errorf("couldn't find kfunc in kernel spec")
```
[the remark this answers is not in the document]
rgo3: As in `internal.UnsupportedFeatureError` or `ErrNotSupported`?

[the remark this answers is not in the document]
lmb: Thinking again, ErrNotSupported is probably better. We don't know which version the kfunc appeared in so `UnsupportedFeatureError` doesn't make sense.

elf_reader.go
rgo3: Refactoring some parts made lots of breaks obsolete.

elf_reader.go
rgo3: I opted for the `elfCode.loadKsymsSection()` which fills up the ksyms map on `elfCode`, that should go in a similar direction as you suggested I hope.

linker.go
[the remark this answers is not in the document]
rgo3: I can still return ErrNotSupported if you prefer, but just returning `err` as is gives test the option to check for `btf.ErrNotFound` as mentioned in the comment.

rgo3:
All the LibBPFCompat Tests that are failing because of STB_WEAK need to be skipped right?

lmb:
> All the LibBPFCompat Tests that are failing because of STB_WEAK need to be skipped right?

Yes. Does it make sense to return ErrNotSupported for STB_WEAK as well?

btf/btf.go
```diff
@@ -470,8 +470,7 @@ func fixupDatasec(types []Type, sectionSizes map[string]uint32, offsets map[symb
 		case ".ksyms":
 			// .ksyms describes forward declarations of kfunc signatures.
 			// Nothing to fix up, all sizes and offsets are 0.
-			// Explicitly reject for now until we officially support kfuncs.
-			return fmt.Errorf("reference to %s: %w", name, ErrNotSupported)
+			continue
```
lmb, line 473: Should we check there that the section only contains funcs?

elf_reader.go
```diff
@@ -585,20 +589,31 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
 
+		kc := ec.maps[kconfigMap]
+		ks := ec.ksyms[name]
+		switch {
+		// If a Call instruction is found and the datasec has a btf.Func with a Name
+		// that matches the symbol name we mark the instruction as a call to a kfunc.
+		case ks != nil && ins.OpCode.JumpOp() == asm.Call:
+			fn, ok := ks.(*btf.Func)
+			if !ok {
+				// I'd like to display which Type we found? %s displays a little more I think
```
lmb, line 600: Try %T?

elf_reader.go
```diff
@@ -1120,6 +1135,18 @@ func (ec *elfCode) loadVirtualDataSections() error {
 		return nil
```
lmb, line 1135: Get rid of loadVirtualDataSections and inline the calls to loadKconfigSection, etc. into the caller. The function doesn't add much except a level of indirection.

btf/btf.go
rgo3: I now check the Datasec here and assume there will only be `*btf.Func`s in there going forward.

linker.go
rgo3: How is https://github.com/cilium/ebpf/pull/896 different from exporting `coreAreTypesCompatible(Type, Type)`? cc @ti-mo

btf/core.go
```diff
@@ -837,30 +838,35 @@ func coreFindEnumValue(local Type, localAcc coreAccessor, target Type) (localVal
...
- * These rules are not set in stone and probably will be adjusted as we get
- * more experience with using BPF CO-RE relocations.
- *
- * Returns errImpossibleRelocation if types are not compatible.
- */
-func coreAreTypesCompatible(localType Type, targetType Type) error {
+// CheckTypeCompatibility checks local and target types for compatibility.
+// Only layout compatibility is checked, ignoring names of the root type.
+//
+// Returns ErrIncompatibleTypes if local and target type are not compatible.
```
lmb, line 844: `errIncompatibleTypes` is not exported currently. Two options:

1. Don't document which error is returned.
2. Export the error.

If possible I'd stick to (1)

btf/core.go
```diff
@@ -837,30 +838,35 @@ func coreFindEnumValue(local Type, localAcc coreAccessor, target Type) (localVal
...
- *   - FUNC_PROTOs are compatible if they have compatible signature: same
- *     number of input args and compatible return and argument types.
- * These rules are not set in stone and probably will be adjusted as we get
- * more experience with using BPF CO-RE relocations.
- *
- * Returns errImpossibleRelocation if types are not compatible.
- */
-func coreAreTypesCompatible(localType Type, targetType Type) error {
+// CheckTypeCompatibility checks local and target types for compatibility.
+// Only layout compatibility is checked, ignoring names of the root type.
```
lmb, line 842: Nit: newline?

btf/core.go
```diff
@@ -837,30 +838,35 @@ func coreFindEnumValue(local Type, localAcc coreAccessor, target Type) (localVal
...
- *   - TYPEDEFs/PTRs are compatible if types they pointing to are compatible;
- *   - FUNC_PROTOs are compatible if they have compatible signature: same
- *     number of input args and compatible return and argument types.
- * These rules are not set in stone and probably will be adjusted as we get
- * more experience with using BPF CO-RE relocations.
- *
- * Returns errImpossibleRelocation if types are not compatible.
- */
-func coreAreTypesCompatible(localType Type, targetType Type) error {
+// CheckTypeCompatibility checks local and target types for compatibility.
```
lmb, line 841: Maybe `... according to CO-RE rules.`? Or we just export as `COREAreTypesCompatible`?

linker.go
lmb: Ah, the reason that we don't handle Typedef in `coreAreTypesCompatible` is that the function assumes that all qualifiers and Typedefs have been stripped out using `UnderlyingType`.

> Wondering if adding *Typdef to the `case *Pointer, *Array` statement would solve this?

Unfortunately not, since we want to treat the following as equal:

```go
local: Typedef{Int}
target: Int{}
```

If we just add to the switch case, we'd still hit https://github.com/cilium/ebpf/blob/5ea5368f9823c13e71146a6c2cca84ad60a5e26a/btf/core.go#L878-L880 which would fail since `Typedef != Int`.

The best fix is to create a wrapper function I think. Leave `coreAreTypesCompatible` unexported, and instead export the wrapper. The wrapper calls `Copy(local, UnderlyingType)`, etc. and then invokes `coreAreTypesCompatible`. CO-RE relocation would still call `coreAreTypesCompatible` but external callers get the more sensible behaviour.

linker.go
```diff
@@ -185,12 +185,54 @@ func fixupAndValidate(insns asm.Instructions) error {
...
+		return nil
+	}
+
+	// check meta, if no meta return err
+	kfm, _ := ins.Metadata.Get(kfuncMeta{}).(*btf.Func)
+	if kfm == nil {
+		return fmt.Errorf("kfunc call has no kfuncMeta")
+	}
+
+	s, err := btf.LoadKernelSpec()
```
lmb, line 211: This is so expensive that we either need to land https://github.com/cilium/ebpf/pull/948 or refactor this into a separate loop over `insns asm.Instructions`, similar to https://github.com/cilium/ebpf/blob/5ea5368f9823c13e71146a6c2cca84ad60a5e26a/btf/ext_info.go#L137-L194

rgo3: fixupKfunc is called within a loop over `insns` anyway, I could pull out the kernel spec load out of the function and pass the spec as an arg. That way we are not bound to another PR landing first.

lmb: The problem is that we don't want to load the kernel spec unconditionally, since it takes 70ms! That's why I suggested a separate function with the weird double loop setup. Agreed that we should not block this on Spec.Copy improvements.

lmb (approved):

PR #969, opened 2023-03-09
add socket filter example

nijich:
Add an eBPF SocketFilter example. This example uses eBPF to filter the packets with port 80 and uses userspace program to extract the packet by the format of http, and prints the status line if the packet is the first part of http payload.


lmb:
Sorry for not replying again. I agree that your code is useful, but it doesn't really showcase additional features of the library. That said, if you decide to publish it elsewhere we could add a link to in in `examples/README.md` or similar.

PR #974, opened 2023-03-10
Add error case for creating unspecified map type.

tommyp1ckles:
When creating a new map, if the the map type is unspecified (i.e. BPF_MAP_TYPE_UNSPEC) then the call to create the map will fail with EINVAL and the following Go error is returned due to the btf file descriptor being zerod out:

...: map create: ... (without BTF k/v)

Instead, upon failure to create me, this will now check if the map type was set to unspecified and return a relevant error.

### DIFF
--- a/map.go
+++ b/map.go
@@ -454,6 +454,9 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 		if errors.Is(err, unix.EINVAL) && attr.MaxEntries == 0 {
 			return nil, fmt.Errorf("map create: %w (MaxEntries may be incorrectly set to zero)", err)
 		}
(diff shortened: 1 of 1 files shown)


map.go
```diff
@@ -334,6 +334,9 @@ func newMapWithOptions(spec *MapSpec, opts MapOptions) (_ *Map, err error) {
 // createMap validates the spec's properties and creates the map in the kernel
 // using the given opts. It does not populate or freeze the map.
 func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err error) {
+	if spec.Type == UnspecifiedMap {
```
lmb, line 337: Could you change this to happen after the map create call has failed? Basically `if err == EINVAL && spec.Type == Unspecified`. That way we avoid validating kernel arguments in userspace.

PR #976, opened 2023-03-13
Allow debugfs fstype for /sys/kernel/debug/tracing

brycekahle:
RHEL/CentOS support eBPF on 3.10 before tracefs was introduced, and thus have /sys/kernel/debug/tracing mounted with debugfs fstype.


lmb (approved):
This keeps on getting better xD

ti-mo:
Fortunately those older versions won't change fundamentally anymore. I hope.

PR #977, opened 2023-03-14
Allow passing extra kconfigs constants to `CollectionSpec`

paulcacheux:
This PR builds on https://github.com/cilium/ebpf/pull/951 and allows filling a map in `CollectionSpec` to add extra kconfig constants to be passed in the same way as `LINUX_KERNEL_VERSION`.

The main goal behind this would be to support providing a value for `LINUX_HAS_SYSCALL_WRAPPER` https://github.com/libbpf/libbpf/blob/cf46d44f0a06aa8b9400691ea3eb86ca4f066d5c/src/bpf_tracing.h#L895

This PR is obviously a PoC and maybe you have something completely different in mind to bridge the kconfig with parsed config files. If this is the case please feel free to close this


ti-mo:
This looks like a very big hammer, something like [CollectionOptions.MapReplacements](https://pkg.go.dev/github.com/cilium/ebpf#CollectionOptions) would fit much better.

lmb:
As [discussed on slack](https://cilium.slack.com/archives/C027KBX679U/p1678373554189919), Paul will attempt to implement LINUX_HAS_SYSCALL_WRAPPER directly in the lib instead.

PR #978, opened 2023-03-14
replace deprecated rand.Seed calls

### LINKED ISSUE #933: rand.Seed() is deprecated in Go 1.20
From godoc:
```
// Deprecated: Programs that call Seed and then expect a specific sequence
// of results from the global random source (using functions such as Int)
// can be broken when a dependency changes how much it consumes
// from the global random source. To avoid such breakages, programs
// that need a specific result sequence should use NewRand(NewSource(seed))
// to obtain a random generator that other packages cannot access.
```

Looks like they're pushing against using global entropy pools. Makes sense. We have a few such uses around the codebase:

```
btf/traversal_test.go
15:	rand.Seed(seed)

examples/map_in_map/main.go
50:	rand.Seed(time.Now().UnixNano())
```

The one in tests can be replaced with a helper in `testutils` that creates a new entropy pool. This needs to take care of locking since the default `rand` implementation isn't concurrency safe.

The one in examples can just be replaced with an unlocked source.

kwakubiney:
Closes #933

### DIFF
--- a/internal/testutils/seed.go
+++ b/internal/testutils/seed.go
@@ -1,0 +1,25 @@
+package testutils
+
+import (
+	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+type lockedCustomSource struct {
+	lk sync.Mutex
+	s  rand.Source
+}
+
+func Seed() {
+	src := lockedCustomSource{
+		lk: sync.Mutex{},
+		s:  rand.NewSource(time.Now().UnixMicro()),
+	}
+
+	src.lk.Lock()
+	src.s.Seed(time.Now().UnixMicro())
+	src.lk.Unlock()
+	fmt.Println("Seed is", src.s)
+}
--- a/internal/testutils/seed.go
+++ b/internal/testutils/seed.go
@@ -12,10 +12,10 @@ var randSeed struct {
 	once  sync.Once
 }
 
-func Seed() rand.Source {
+func Rand() *rand.Rand {
 	randSeed.once.Do(func() {
 		randSeed.value = time.Now().UnixMicro()
-		fmt.Printf("Seed is %d", randSeed.value)
+		fmt.Printf("Rand is %d", randSeed.value)
 	})
-	return rand.NewSource(randSeed.value)
+	return rand.New(rand.NewSource(randSeed.value))
 }
--- a/internal/testutils/seed.go
+++ b/internal/testutils/seed.go
@@ -15,7 +15,7 @@ var randSeed struct {
 func Seed() rand.Source {
 	randSeed.once.Do(func() {
 		randSeed.value = time.Now().UnixMicro()
-		fmt.Println(fmt.Sprintf("Seed is %d", randSeed.value))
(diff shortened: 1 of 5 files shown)


ti-mo (changes requested):
The changes to the map_in_map example are good.

Consider adding compatible `Intn()` and `Shuffle()` functions to package `internal` (or a new subpackage `internal/rand`)  that automatically seed a package-global rand.Source using Memoize. `testutils.Seed()` could return the value it used for seeding the Source, so tests can display it.

I'd even consider swapping out the remaining uses of `math/rand` (so, in `map.go` and `link/kprobe.go`) with a source from an internal package, since the use in `map.go` allocates a new source each time.

@lmb any thoughts?

lmb:
> automatically seed a package-global rand.Source using Memoize

We should avoid a global rand.Source, since that reintroduces the problem that the deprecation of rand.Seed is meant to fix. My preference is still `testutils.Rand` since that can take care of printing the seed once.

> I'd even consider swapping out the remaining uses of `math/rand` (so, in `map.go` and `link/kprobe.go`) with a source from an internal package, since the use in `map.go` allocates a new source each time.

Let's do this in a follow up if necessary.

internal/testutils/seed.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+)
+
+type lockedCustomSource struct {
+	lk sync.Mutex
+	s  rand.Source
+}
+
+func Seed() {
+	src := lockedCustomSource{
```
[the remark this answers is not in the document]
kwakubiney: Yeah so as discussed on Slack, I used `sync.Do()` to handle printing it once during tests and created a helper which returns the `Source` for each dedicated test to avoid using a global `Source`.

internal/testutils/seed.go
```diff
@@ -0,0 +1,21 @@
...
+	"sync"
+	"time"
+)
+
+var randSeed struct {
+	value int64
+	once  sync.Once
+}
+
+func Seed() rand.Source {
```
lmb, line 15: Why not return `*rand.Rand` here? That would make callers a bit less noisy.
```suggestion
func Rand() *rand.Rand {
```

kwakubiney: I agree. Keeping the complexity in the function is much cleaner.

lmb:
Thank you for your contribution! I fixed up your PR a little bit:

- Squashed into a single commit
- Removed duplicate `Signed-off-by:` lines. You only need one, with an email that you have validated for your GH profile.
- Added a short commit message body

Can you tick "Allow changes from maintainers" in the future? That makes it easier for us to make these little fixes!

kwakubiney:
> Thank you for your contribution! I fixed up your PR a little bit:
> 
> * Squashed into a single commit
> * Removed duplicate `Signed-off-by:` lines. You only need one, with an email that you have validated for your GH profile.
> * Added a short commit message body
> 
> Can you tick "Allow changes from maintainers" in the future? That makes it easier for us to make these little fixes!

Happy to get this merged and thanks for the clean-up! & I definitely will do that next time :)

PR #980, opened 2023-03-21
link: document AttachCgroup and flags

ti-mo:
Document some of the lessons learned through https://github.com/cilium/cilium/pull/23557.

Also removes some papercuts from `AttachCgroup` resource cloning.

### DIFF
--- a/link/cgroup.go
+++ b/link/cgroup.go
@@ -27,36 +27,33 @@ type CgroupOptions struct {
 }
 
 // AttachCgroup links a BPF program to a cgroup.
-func AttachCgroup(opts CgroupOptions) (Link, error) {
+func AttachCgroup(opts CgroupOptions) (cg Link, err error) {
 	cgroup, err := os.Open(opts.Path)
 	if err != nil {
 		return nil, fmt.Errorf("can't open cgroup: %s", err)
 	}
-
-	clone, err := opts.Program.Clone()
-	if err != nil {
+	defer func() {
+		if _, ok := cg.(*progAttachCgroup); ok {
+			// Skip closing the cgroup handle if we return a valid progAttachCgroup,
+			// where the handle is retained to implement Update().
+			return
+		}
 		cgroup.Close()
-		return nil, err
-	}
+	}()
 
-	var cg Link
-	cg, err = newLinkCgroup(cgroup, opts.Attach, clone)
+	cg, err = newLinkCgroup(cgroup, opts.Attach, opts.Program)
 	if err == nil {
-		cgroup.Close()
-		clone.Close()
 		return cg, nil
 	}
 
 	// cgroup and clone are retained by progAttachCgroup.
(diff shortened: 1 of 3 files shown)


rgo3 (approved):
Had to double check that `QueryPrograms` also returns progs attached with links just one more time :grin:

ti-mo:
> Had to double check that `QueryPrograms` also returns progs attached with links just one more time grin

Yep, we had to find that out the hard way. :cry:

link/cgroup.go
```diff
@@ -93,7 +95,17 @@ func newProgAttachCgroup(cgroup *os.File, attach ebpf.AttachType, prog *ebpf.Pro
 		return nil, fmt.Errorf("cgroup: %w", err)
 	}
 
-	return &progAttachCgroup{cgroup, prog, attach, flags}, nil
+	// Retain cgroup and program handles we own to support Update().
+	clone, err := prog.Clone()
+	if err != nil {
+		return nil, err
+	}
+	cgroup, err = os.Open(cgroup.Name())
```
lmb, line 103: Hmm, doesn't this create a TOCTTOU? The name might be pointing at at different cgroup after all. I'd prefer to not reopen the file. If you really want to you could dup the fd, but that seems a lot more cumbersome?

Also creates a hazard since we've already done AttachProgram, aka maybe replaced an existing prog. We can't undo that in this error path. So cloning, etc. must happen before attach I think.

link/cgroup.go
```diff
@@ -10,10 +10,15 @@ import (
 
 type cgroupAttachFlags uint32
 
-// cgroup attach flags
 const (
+	// Allow programs attached to sub-cgroups to override the verdict of this
```
lmb, line 14: Nice, thanks!

P.S. Probably something we should add to `sys` at some point.

lmb (changes requested):

lmb:
> Had to double check that `QueryPrograms` also returns progs attached with links

Ooh! Now I understand why you changed the doc.

link/cgroup.go
ti-mo: Went with a closure that skips Close() for `progAttachCgroup`.

link/cgroup.go
```diff
@@ -83,17 +91,23 @@ func newProgAttachCgroup(cgroup *os.File, attach ebpf.AttachType, prog *ebpf.Pro
...
+
+	err = RawAttachProgram(RawAttachProgramOptions{
 		Target:  int(cgroup.Fd()),
-		Program: prog,
+		Program: clone,
 		Flags:   uint32(flags),
 		Attach:  attach,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("cgroup: %w", err)
```
lmb, line 107: This needs clone.Close now?

ti-mo: Good catch. Leak tests don't catch it since it's in an error path.

PR #982, opened 2023-03-22
update `platformPrefix` list with libbpf values

paulcacheux:
As a first step towards being able to compute a value for `LINUX_HAS_SYSCALL_WRAPPER`, this PR updates the platform prefix list with values from libbpf and go architecture list

### DIFF
--- a/link/platform.go
+++ b/link/platform.go
@@ -6,17 +6,36 @@ import (
 )
 
 func platformPrefix(symbol string) string {
-
-	prefix := runtime.GOARCH
-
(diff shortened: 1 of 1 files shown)


lmb (approved):
Thanks! Did a small drive by refactoring. For `HAS_SYSCALL_WRAPPER` we should probably turn this into `platformPrefix() string` instead, so that we can tell if we don't know of a platform prefix.

ti-mo:
Thanks!

PR #985, opened 2023-03-27
docs: add a contributor ladder

lmb:
Spell out how contributors are organised for the library. Inspired by the cilium ladder at
https://github.com/cilium/community/blob/main/CONTRIBUTOR-LADDER.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,3 +35,41 @@ Examples:
 ./run-tests.sh 5.4 ./link
 ```
 
+# Contributor ladder
+
+If you'd like to contribute to the library more regularly, one of the
+[maintainers][ebpf-lib-maintainers] can add you to the appropriate team or mark
+you as a code owner. Just create an issue in the repository.
+
+## [ebpf-lib-contributors]
+
+Contributors support the development of the library by writing code, diagnosing
+bugs, triaging issues, answering questions, etc.
+
+* Have ["Triage"][permissions] role
+* May be asked to review certain parts of code
+* May be asked to help with certain issues
+
+## [ebpf-lib-reviewers]
+
+Code owners have contributed a certain feature or are domain experts. They support
+development by helping to review code.
+
+* Have ["Write"][permissions] role
+* Can't merge to protected branches
+* Can't create releases
+* May be asked to become CODEOWNERS
+
+## [ebpf-lib-maintainers]
+
+Code owners of last resort and responsible for the overall direction of the
+library.
+
+* Have ["Admin"][permissions] role
+* Can re-run failed CI tasks
+* May be asked to add others to [ebpf-lib-contributors] or CODEOWNERS
+
+[permissions]: https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role
+[ebpf-lib-contributors]: https://github.com/orgs/cilium/teams/ebpf-lib-contributors/members
+[ebpf-lib-reviewers]: https://github.com/orgs/cilium/teams/ebpf-lib-reviewers/members
+[ebpf-lib-maintainers]: https://github.com/orgs/cilium/teams/ebpf-lib-maintainers/members
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @cilium/ebpf-lib-maintainers


CONTRIBUTING.md
```diff
@@ -35,3 +35,41 @@ Examples:
...
+
+Contributors support the development of the library by writing code, diagnosing
+bugs, triaging issues, answering questions, etc.
+
+* Can request reviews
+* Can triage issues and pull requests
+* May be asked to review certain parts of code
+* May be asked to help with certain issues
+
+## Code owners
```
rgo3, line 55: We currently don't have a CODEOWNERS file, should we add an initial one with this PR?

lmb: I'll add

```
* @ebpf-lib-maintainers
```

as a start. If I understand correctly that would always request a review from either Timo or me, which is essentially what happens already.

There is another [bit here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners):

> The people you choose as code owners must have write permissions for the repository.

The implications of write permission are wider than I'm proposing: https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/repository-roles-for-an-organization Specifically a CODEOWNER can merge PRs (subject to branch protection rules) and create new releases. I'd prefer to restrict cutting a release to maintainers, maybe we can use [tag protection rules](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) to prevent creating `v*` tags?

Otherwise we could "emulate" code owners manually by asking individuals for review and putting responsibilites into a `.md`.

CONTRIBUTING.md
```diff
@@ -35,3 +35,41 @@ Examples:
...
+* Can't create releases
+* May be asked to become CODEOWNERS
+
+## [ebpf-lib-maintainers]
+
+Code owners of last resort and responsible for the overall direction of the
+library.
+
+* Have ["Admin"][roles] role
+* Can re-run failed CI tasks
```
lmb, line 70: Reviewers should be able to do this, but for that we probably need to migrate to GH actions

ti-mo: That's fine for now, don't recall Semaphore getting stuck jobs lately.

CONTRIBUTING.md
```diff
@@ -35,3 +35,41 @@ Examples:
 ./run-tests.sh 5.4 ./link
 ```
 
+# Contributor ladder
+
+If you'd like to contribute to the library more regularly, one of the
+[maintainers][ebpf-lib-maintainers] can add you to the appropriate team or mark
+you as a code owner.
+Just create an issue in the repository.
```
ti-mo, line 43: Nit: can be reflowed.

I'd maybe add #ebpf-go-dev as another avenue for getting ownership.

ti-mo (approved):
Agree overall, thanks! Some parts of the markdown look broken though.

CONTRIBUTING.md
lmb: I prefer issues since it leaves an "audit trail" of the decision, which we lose with the slack channel. OTOH it might increase the barrier for people if we reject their request publicly? I could also get rid of this and say "contact the maintainers"

CONTRIBUTING.md
```diff
@@ -35,3 +35,41 @@ Examples:
...
+## [ebpf-lib-contributors]
+
+Contributors support the development of the library by writing code, diagnosing
+bugs, triaging issues, answering questions, etc.
+
+* Have ["Triage"][roles] role
+* May be asked to review certain parts of code
+* May be asked to help with certain issues
+
+## [ebpf-lib-reviewers]
```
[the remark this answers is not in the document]
lmb: NB: Haven't created the team yet since I wasn't sure re permissions.

joestringer:
Is there a reason that the [org-wide contributor ladder](https://github.com/cilium/community/blob/main/CONTRIBUTOR-LADDER.md) doesn't fit well for the ebpf library as well? The _intent_ is for it to be more generic than just for cilium/cilium. While I'm not at all against having a simpler process for subprojects, I think that it could potentially introduce some confusion for contributors in future if there's two different sets of rules, tracked by different mechanisms in different places.

I'd imagine that @xmulligan and I would be happy to discuss the ways that the general ladder does/does not fit well for cilium/ebpf and help to figure out a good balance between providing a consistent contributor experience and trying to make the process as minimal as possible to still effectively manage the project.

lmb:
> Is there a reason that the [org-wide contributor ladder](https://github.com/cilium/community/blob/main/CONTRIBUTOR-LADDER.md) doesn't fit well for the ebpf library as well?

I read the ladder after you pointed it out and took inspiration. I like that is calls out responsibilities and privileges. My take is that it's designed to work for a large project. The library isn't that, so I don't want to copy it wholesale. Some specific points:

* What are voting privileges all about?
* Maintaining separate community contributors seems like overhead I'd like to avoid. We can honour non-code contributions by adding to contributors.
* The ladder lists privileges, but doesn't tie them to GH permissions. This is important to me because that makes privileges much more explicit.
* The ladder doesn't account for maintainership (maybe that's just the same as committers for cilium?)

> While I'm not at all against having a simpler process for subprojects, I think that it could potentially introduce some confusion for contributors in future if there's two different sets of rules, tracked by different mechanisms in different places.

That seems like an argument against putting the contributors ladder in our repo more than having our own ladder? To be honest, I'm not convinced the argument about confusing contributors really holds water. Many of the current library contributors (which I'd like to get on board with this ladder) don't contribute to cilium proper. So I'd say that if and when there is a significant overlap between (non-Isovalent) cilium contributors and library contributors we can see whether it really is confusing and then make changes accordingly.

There is another reason why I'm pushing back here: it seems like there is a drive to centralise governance, and I disagree with the way it's been done. I've invested a lot of my personal time into maintaining the lib and so in a way I'd like the chance to influence its destiny. So far a lot of the "org wide" changes were just decided somewhere somehow.

lmb:
Closing for now, upstream discussion wasn't very fruitful.

PR #989, opened 2023-03-28
btf: remove type alias for []Type

lmb:
Remove "type types []Type" since the only use case is the ByID method, which doesn't take split BTF into account. Use *Spec instead where necessary.

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -111,7 +111,7 @@ func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
 	file, err := internal.NewSafeELFFile(rd)
 	if err != nil {
 		if bo := guessRawBTFByteOrder(rd); bo != nil {
-			return loadRawSpec(io.NewSectionReader(rd, 0, math.MaxInt64), bo, nil, nil)
+			return loadRawSpec(io.NewSectionReader(rd, 0, math.MaxInt64), bo, nil)
 		}
 
 		return nil, err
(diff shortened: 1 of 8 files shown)


btf/btf.go
```diff
@@ -228,8 +228,19 @@ func loadSpecFromELF(file *internal.SafeELFFile) (*Spec, error) {
...
+	var (
+		baseStrings *stringTable
+		baseTypes   []Type
+		firstTypeID TypeID
+	)
+
+	if base != nil {
+		baseStrings = base.strings
+		baseTypes = base.types
+		firstTypeID = base.lastTypeID + 1
```
rgo3, line 241: Shouldn't this be:
```suggestion
		firstTypeID = TypeID(len(baseTypes))
```
Or am I missing something?

lmb: Drive by "possible bugfix" by me, sorry. I should've split this out into a separate commit. `lastTypeID` is more reliable, since `len(baseTypes)` assumes that types start at 0. That isn't correct for split BTF. Now, we don't support split BTF base, but I figured it's worth to fix this before it becomes a bug.

lmb: Made a separate commit.

lmb: It's all broken :cry:

lmb:
Removed the drive-by fix, I'll open a separate PR.

ti-mo:
Thanks!

PR #994, opened 2023-03-30
elf: use per-instruction metadata for .kconfig references

lmb:
.kconfig support is currently implemented in the ELF loader, which does
a fix up of the relevant instructions to point at a ".kconfig" map. This
map is available in CollectionSpec.Maps. This is basically how constants
(aka references to .rodata) work today. Now that we have per-instruction
metadata this isn't necessary anymore.

Instead of using the generic WithReference, introduce a new kconfig
metadata and tag instructions with that. Right now the metadata just
contains the MapSpec for .kconfig as generated by the compiler. In
the future we can create the MapSpec on the fly, which would allow
referencing kconfig via ins.WithKconfig or similar.

Prevent ".kconfig" from being accessible via CollectionSpec.Maps.
This preserves our ability to change how kconfig references work
without having to retain the original MapSpec.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -18,7 +18,12 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
 
-const kconfigMap = ".kconfig"
+type kconfigMetaKey struct{}
+
+type kconfigMeta struct {
+	Map    *MapSpec
+	Offset uint32
+}
 
 type kfuncMeta struct{}
 
@@ -33,6 +38,7 @@ type elfCode struct {
 	extInfo  *btf.ExtInfos
 	maps     map[string]*MapSpec
 	kfuncs   map[string]*btf.Func
+	kconfig  *MapSpec
 }
 
 // LoadCollectionSpec parses an ELF file into a CollectionSpec.
@@ -593,7 +599,6 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 			return fmt.Errorf("asm relocation: %s: unsupported type %s", name, typ)
 		}
 
-		kc := ec.maps[kconfigMap]
 		kf := ec.kfuncs[name]
 		switch {
 		// If a Call instruction is found and the datasec has a btf.Func with a Name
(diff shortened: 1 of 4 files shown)


linker.go
```diff
@@ -277,3 +278,60 @@ func fixupProbeReadKernel(ins *asm.Instruction) {
...
+		if insSpec == nil {
+			goto next
+		}
+
+		if insSpec != spec {
+			kconfig.Close()
+			return nil, fmt.Errorf("instruction %d: reference to multiple .kconfig maps is not allowed", iter.Index)
+		}
+
+		if err := iter.Ins.AssociateMap(kconfig); err != nil {
```
lmb, line 323: Only half joking: we could implement this without going via a map in the first place and just patch instructions directly. Something for the future maybe.

lmb:
cc @eiffel-fl

linker.go
lmb: Instead of putting `LINUX_VERSION_CODE` into the map at an offset, and referencing the map from this instruction, we could rewrite the instruction to be a `LdImm64` of LINUX_VERSION_CODE. This avoids the indirection via the map and works on older kernels. It has subtleties though when a single value is accessed multiple times, etc.

lmb:
> Also, can you please indicate me since when each instruction has metadata?

Here you go: https://github.com/cilium/ebpf/commit/e3e0e9e3f2205b3103e2caa34d90a7f28d1d8021

elf_reader.go
```diff
@@ -1134,7 +1132,7 @@ func (ec *elfCode) loadKconfigSection() error {
 	}
 
 	var ds *btf.Datasec
-	err := ec.btf.TypeByName(kconfigMap, &ds)
+	err := ec.btf.TypeByName(".kconfig", &ds)
```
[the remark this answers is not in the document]
lmb: It's only used in three places now: here in the elf reader and when fixing up datasec in btf. The one use for MapSpec.Name is purely informational, so if we get it wrong it doesn't matter much.

In general I'm not a big fan of turning string literals into a constant, since it adds a layer of indirection. When reading code it's not clear that it's a constant (could be a var) and also what it's value is. If a string literal is repeated over an over as a key I try to reduce the number of lookups instead of introducing a constant.

linker.go
```diff
@@ -277,3 +278,59 @@ func fixupProbeReadKernel(ins *asm.Instruction) {
...
+			continue
+		}
+
+		spec = meta.Map
+		goto populate
+	}
+
+	return nil, nil
+
+populate:
```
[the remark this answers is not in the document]
lmb: Yeah that's much nicer, thanks!

linker.go
lmb: I'm not going to do it now, since I don't have a use case yet. It might happen that we need something like it in cilium though. Then we can probably extend this to anything that fits in 64 bit.

ti-mo: > It has subtleties though when a single value is accessed multiple times, etc.

Indeed, but it's also a subtlety that cannot realistically be worked around by the library. It would require userspace to track all accesses to the register the map pointer was loaded into, accounting for branches and jumps, as well as the possibility of aliasing/moving, etc. This is best left to the verifier. :slightly_smiling_face:

PR #995, opened 2023-03-30
DRAFT: add support for `LINUX_HAS_SYSCALL_WRAPPER` kconfig

paulcacheux:


link/perf_event_test.go
lmb, line 1: Can you move the tests as well? Othewise it's hard to figure out which functions are only exported because of tests.

lmb (commented):
I rescued what I wrote on slack:

> so it seems like there are four different ways of detecting this, if i'm counting right.
> * CONFIG_ARCH_HAS_SYSCALL_WRAPPER in /proc/config.gz (requires some config option?)
> * check for symbol with prefix in /proc/kallsysm (needs CONFIG_KALLSYMS)
> * attach bpf_link to $prefix_sys_bpf (needs recent kernel)
> * attach tracefs kprobe (best compat?)
> 
> looking at the kprobe code, wouldn't it just be enough to create a kprobe event without attaching a program?
> 
> basically, duplicate https://github.com/cilium/ebpf/blob/5ea5368f9823c13e71146a6c2cca84ad60a5e26a/link/kprobe.go#L445-L464 to attempt creating an event for symbol $prefix_sys_bpf. if that works we know there is a wrapper, and can remove the event. this is even without attaching a program.

As far as I can tell, you've implemented probing via PMU and tracefs, which is slightly more elaborate than what I had in mind. As a result the changes required are pretty large.

Wouldn't it be sufficient to only ever probe via tracefs? This has the highest compatibility anyways and it has the nice benefit that it's just writing to files. I believe would lead to a lot smaller diff, since you only need to move the functions to create and remove a tracefs event. For that we can have `internal/tracefs`, which I think is a sensible thing to do in any case. Right now the tracefs / pmu code is so intermingled that I find it hard to keep track.

P.S. I think the `link/lite` package is a bit dangerous, since it invites us to "just reimplement the necessary bits". I don't want to end up in a situation where we have to add features twice. I'd much prefer to move as much code into `internal` with dedicated APIs so that we can "inline" the lite use cases into the callers.

lmb:
FYI unsubscribing from the PR for now to receive less notifications. Please request a new review / ping me on slack when you are ready.

internal/tracefs/kprobe.go
```diff
@@ -0,0 +1,381 @@
...
+	return fd, err
+}
+
+// TracefsProbe creates a trace event by writing an entry to <tracefs>/[k,u]probe_events.
+// A new trace event group name is generated on every call to support creating
+// multiple trace events for the same kernel or userspace symbol.
+// Path and offset are only set in the case of uprobe(s) and are used to set
+// the executable/library path on the filesystem and the offset where the probe is inserted.
+// A perf event is then opened on the newly-created trace event and returned to the caller.
+func TracefsProbe(typ ProbeType, args ProbeArgs) (*sys.FD, uint64, string, error) {
```
paulcacheux, line 49: This is the only real change from this PR, all the rest is mostly moving things around, but this basically extract the fd creation to `internal/tracefs` and keeps the `perfEvent struct` creation in `link`

internal/platform.go
```diff
@@ -1,11 +1,11 @@
-package link
+package internal
 
 import (
 	"fmt"
 	"runtime"
 )
 
-func platformPrefix(symbol string) string {
+func PlatformPrefix(symbol string) (string, bool) {
```
lmb, line 8: Maybe `PlatformPrefix() string` with `""` meaning unknown prefix is more convenient?

Nit: can you add a doc comment?

syscalls.go
```diff
@@ -262,3 +263,15 @@ var haveBPFToBPFCalls = internal.NewFeatureTest("bpf2bpf calls", "4.16", func()
 	_ = fd.Close()
 	return nil
 })
+
+var haveSyscallWrapper = internal.NewFeatureTest("syscall wrapper", "4.17", func() error {
+	testSyscallName, ok := internal.PlatformPrefix("sys_bpf")
+	if !ok {
+		return internal.ErrNotSupported
```
lmb, line 270: I think it would be more appropriate to return an error that is not ErrNotSupported here, since we're not actually able to tell whether the arch has as wrapper.

internal/tracefs/kprobe.go
```diff
@@ -0,0 +1,381 @@
...
+)
+
+var (
+	ErrInvalidInput = errors.New("invalid input")
+
+	ErrInvalidMaxActive = errors.New("can only set maxactive on kretprobes")
+)
+
+func KprobeCheckLite(symbol string, pid int) error {
+	fd, err := tryAttach(symbol, pid)
```
lmb, line 24: Why is the check here, seems like all the API is there to just move this code into `haveSyscallWrapper`?

internal/tracefs/kprobe.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+// CreateTraceFSProbeEvent creates a new ephemeral trace event.
+//
+// Returns os.ErrNotExist if symbol is not a valid
+// kernel symbol, or if it is not traceable with kprobes. Returns os.ErrExist
+// if a probe with the same group and symbol already exists. Returns an error if
+// args.RetprobeMaxActive is used on non kprobe types. Returns ErrNotSupported if
+// the kernel is too old to support kretprobe maxactive.
+func CreateTraceFSProbeEvent(typ ProbeType, args ProbeArgs) (uint64, error) {
```
lmb, line 247: Seems like this is only public to satisfy tests, could you move the tests and unexport the function?

internal/tracefs/kprobe.go
lmb: I think this is nice, and I might even go a step further: remove `OpenTracepointPerfEvent` from this function and instead of `sys.FD` return a new `type Probe struct{...}` which has a `Close()` method that calls `removeTraceFSProbeEvent`. Maybe even a finalizer, but that is TBD. I think that would make the code easier to follow and gets rid of the `pe.tracefsID != 0` checks in `perfEvent.Close`.

Nit: how about calling this `tracefs.NewProbe` instead?

lmb:
cc @mmat11 for a second opinion as well.

internal/tracefs/kprobe.go
```diff
@@ -0,0 +1,381 @@
...
+			return p.path, nil
+		}
+	}
+
+	return "", errors.New("neither debugfs nor tracefs are mounted")
+})
+
+// SanitizeSymbol replaces every invalid character for the tracefs api with an underscore.
+// It is equivalent to calling regexp.MustCompile("[^a-zA-Z0-9]+").ReplaceAllString("_").
+func SanitizeSymbol(s string) string {
```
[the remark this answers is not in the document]
paulcacheux: `TestUprobeSanitizedSymbol` was moved to `internal/tracefs` yes

internal/tracefs/kprobe.go
```diff
@@ -0,0 +1,381 @@
...
+	return group, nil
+}
+
+// IsValidTraceID implements the equivalent of a regex match
+// against "^[a-zA-Z_][0-9a-zA-Z_]*$".
+//
+// Trace event groups, names and kernel symbols must adhere to this set
+// of characters. Non-empty, first character must not be a number, all
+// characters must be alphanumeric or underscore.
+func IsValidTraceID(s string) bool {
```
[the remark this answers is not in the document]
paulcacheux: Moving this to `GetTracepointID` makes some test fail. I will investigate later, but I let it like that for now

paulcacheux:
Current status:
- I applied most of the feedback related to exported functions
- I started squashing and reorganizing commits for easier review, but there is still some work to do here
- I'm still working on the `tracefs.NewProbe` suggestion

internal/tracefs/kprobe.go
paulcacheux: Yes it's only public to satisfy tests, but those tests are not movable for now because they are using some things from `link`. It's only public but internal on the other hand. So not really exported to outside users.

lmb:
Looks nice, thanks for reorganizing the PR! I ended up dropping `NewProbe`, I think it doesn't make sense without adding a `type Event struct{}` abstraction that also handles cleanup.

Can you add a test that exercises LINUX_HAS_SYSCALL_WRAPPER? I fixed up your last commit to do `errors.Is(err, ErrNotSupported)` since feature probes never return unwrapped errors. So I think in it wouldn't have set the value correctly?

 I think we don't need the full libbpf macro magic, the following should suffice:

- Add a .c that has `extern blabla LINUX_HAS_SYSCALL_WRAPPER __kconfig` and asserts that the value is 1.
- Add a test case which loads the program.
- Running the program to check that the value is as expected would be nice, but is probably not a requirement.

paulcacheux:
Thanks ! I will add the test yes

lmb (approved):
Thanks a lot :)

PR #1000, opened 2023-04-04
elf: freeze .kconfig map

lmb:
The verifier will only treat .kconfig as constants if the map is frozen, since otherwise modification from user space is possible.

Always freeze the kconfig map.


lmb:
cc @eiffel-fl

lmb:
> As it will consider it constant it will be quicker to check it, right?

I think it actually turns the load from memory into a constant load (in the JITed BPF code) and can then use the constant value for deciding whether code is reachable or not! This might make it faster to verify a program, but the main benefit is that some programs start working at all because large chunks can be ignored by the verifier.

PR #1001, opened 2023-04-06
btf: avoid repeated kernel BTF copies in Handle.Spec

lmb:
When parsing kernel module BTF we currently obtain a copy of the vmlinux types. This is nice from an API standpoint: the user doesn't have to care that they are dealing with a module. The problem is that copying vmlinux BTF is very slow due to its size.

We need to copy vmlinux BTF since it's otherwise possible for callers of Handle.Spec to modify the internal copy of vmlinux BTF by accident.

Add an explicit base so that we can avoid copying the kernel spec.

### DIFF
--- a/btf/handle.go
+++ b/btf/handle.go
@@ -110,7 +110,10 @@ func NewHandleFromID(id ID) (*Handle, error) {
 }
 
 // Spec parses the kernel BTF into Go types.
-func (h *Handle) Spec() (*Spec, error) {
+//
+// base must contain type information for vmlinux if the handle is for
+// a kernel module. It may be nil otherwise.
+func (h *Handle) Spec(base *Spec) (*Spec, error) {
 	var btfInfo sys.BtfInfo
 	btfBuffer := make([]byte, h.size)
 	btfInfo.Btf, btfInfo.BtfSize = sys.NewSlicePointerLen(btfBuffer)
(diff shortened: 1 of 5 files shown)


btf/handle.go
```diff
@@ -110,7 +110,10 @@ func NewHandleFromID(id ID) (*Handle, error) {
 }
 
 // Spec parses the kernel BTF into Go types.
-func (h *Handle) Spec() (*Spec, error) {
+//
+// base must contain type information for vmlinux if the handle is for
+// a kernel module. It may be nil otherwise.
+func (h *Handle) Spec(base *Spec) (*Spec, error) {
```
rgo3, line 116: Do we have a policy regarding breaking an API like this?

rgo3: Do we make a note somewhere to document the change in the release notes? Yes, adding it to CONTRIBUTING is a good idea, maybe even add that breakages should always be pointed out/argued in commit messages?

rgo3 (approved):

btf/handle.go
lmb: See #1003

ti-mo: > Breaking the API in a way that causes compile failure is OK

Caveat: ... if an equivalent API is available and the migration steps are clear, e.g. in the case of deprecations.

PR #1003, opened 2023-04-06
docs: document API stability

lmb:
Write down the guidelines we've been using to decide whether a breaking change is acceptable or not.


rgo3 (approved):
Thanks!

PR #1006, opened 2023-04-10
ProgramInfo provides CreatedByUid

zachcheu:
The CreatedByUid provides some helpful context for the source of the BPF program. Making it public in ProgramInfo

### DIFF
--- a/info.go
+++ b/info.go
@@ -94,8 +94,10 @@ type ProgramInfo struct {
 	// Name as supplied by user space at load time. Available from 4.15.
 	Name string
 
-	btf   btf.ID
-	stats *programStats
+	createdByUID     uint32
+	haveCreatedByUID bool
+	btf              btf.ID
+	stats            *programStats
 
 	maps  []MapID
 	insns []byte
@@ -136,6 +138,12 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 	} else {
 		// The kernel doesn't report associated maps.
 		pi.maps = nil
+	}
+
+	// createdByUID and NrMapIds were introduced in the same kernel version.
+	if pi.maps != nil {
+		pi.createdByUID = info.CreatedByUid
+		pi.haveCreatedByUID = true
 	}
 
 	if info.XlatedProgLen > 0 {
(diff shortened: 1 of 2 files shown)


info.go
```diff
@@ -93,6 +93,8 @@ type ProgramInfo struct {
 	Tag string
 	// Name as supplied by user space at load time. Available from 4.15.
 	Name string
+	// CreatedByUid specifies the uid of the process that loaded the program.
+	CreatedByUid uint32
```
lmb, line 97: Which kernel version introduced this field? If it's after 4.9 this needs to be unexported and accessed via a getter instead, similar to `Program.ID()`. The reason is that users need to be able to tell if the information is present or not. For example, does `CreatedByUid == 0` mean that root created it, or that the kernel didn't provide that information?

zachcheu: Looks like it was [introduced in 4.15](https://elixir.bootlin.com/linux/v4.15/C/ident/created_by_uid). I don't see how a getter would help determine whether the `CreateByUid == 0` is because the root was created or because the kernel didn't populate this field. Is there a way to enforce that this function is called on kernel 4.15+?

lmb: Luckily it was introduced in the [same commit as nr_map_ids](https://github.com/torvalds/linux/commit/cb4d2b3f03d8), which we already have (feature probe) support for: https://github.com/cilium/ebpf/blob/f591f6f57a5049cb2a115188b9647037cda2b6cf/info.go#L129-L139

I suggest the following:
* Add `MapInfo.createdByUID uint32` and `MapInfo.haveCreatedByUID bool`.
* Populate that from the place I linked above. I think it needs to happen from both the `if` and the `else if` branch.
* Add `MapInfo.CreatedByUID() (uint32, bool) { return mi.createyByUID, mi.haveCreatedByUID }`

Plus some tests would be great!

zachcheu: I'm confused about why CreatedByUID fields and functions should be added under `MapInfo` since `CreatedByUID` is Program Info specific.

If fields MapInfo.createdByUID and MapInfo.haveCreatedByUID are created then how should they be populated at https://github.com/cilium/ebpf/blob/f591f6f57a5049cb2a115188b9647037cda2b6cf/info.go#L45-L53

lmb:
Fixed up the tests to work on old kernels. Thanks for the PR!

PR #1015, opened 2023-04-20
btf: move LoadKernelSpec into separate package

lmb:
vmlinux BTF turns out to be a big headache, because it is quite large at roughly 150k types. The API we currently have in place forces a copy of the whole vmlinux btf.Spec on every call to LoadKernelSpec. We copy to prevent distinct callers of LoadKernelSpec from modifying types that may be referenced by someone else. As a result, some things that require access to kernel BTF are really slow: attaching fentry/fexit hooks, kfuncs, etc.

We've optimized the copying a little bit, but we can't get around the fact that there are a lot of objects to be copied so it's unlikely to get much faster. An attempt to copy btf.Spec lazily was abandoned due to it's complexity.

Which leaves us with this approach: the library shares a single kernel Spec and must take care to avoid mutating or exposing types contained therein. Users of the library always get a fresh copy and must themselves take care to cache the spec if necessary.

Due to import cycles this means we can't expose the kernel spec from package btf anymore and instead need a separate package. This will cause compilation errors, but they should be simple to fix since the semantics don't change otherwise.

### DIFF
--- a/btf/core.go
+++ b/btf/core.go
@@ -166,11 +166,8 @@ func (k coreKind) String() string {
 // for relos[i].
 func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder) ([]COREFixup, error) {
 	if target == nil {
-		var err error
-		target, _, err = kernelSpec()
-		if err != nil {
-			return nil, fmt.Errorf("load kernel spec: %w", err)
-		}
+		// Explicitly check for nil here since the argument used to be optional.
+		return nil, fmt.Errorf("target must be provided")
 	}
 
 	if bo != target.byteOrder {
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -10,7 +10,6 @@ import (
 	"math"
 	"os"
 	"reflect"
-	"sync"
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/sys"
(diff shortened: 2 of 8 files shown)


internal/linuxint/types.go
ti-mo, line 1: What do you think of `btf/internal/linux`?

internal/linuxint/types.go
```diff
@@ -0,0 +1,107 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/cilium/ebpf/btf"
+	"github.com/cilium/ebpf/internal"
+)
+
+var kernelBTF struct {
```
ti-mo, line 14: To our discussion of earlier: in the future, instead of the anonymous struct `kernelBTF`, we could have a hypothetical `btf.Loader` here that acts as a cache and serves as the API to load vmlinux and kmods. I see this package already imports `btf/`, so no need for a global `any` var.

The `btf.Loader` stored in this package would become the internal copy, it would be flushed by `linux.FlushTypes()`, and `linux.Types()` would instead return a `btf.Loader` so the caller can have its own personal BTF cache.

With this PR, it doesn't seem like package `btf/` still relies on accessing the kernel spec. That makes things a lot less complicated!

btf/core.go
```diff
@@ -165,14 +165,6 @@ func (k coreKind) String() string {
 // Fixups are returned in the order of relos, e.g. fixup[i] is the solution
 // for relos[i].
 func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder) ([]COREFixup, error) {
-	if target == nil {
```
ti-mo, line 168: Returning error on nil here would be nice.

ti-mo (changes requested):
Few nits/comments.

internal/linuxint/types.go
lmb: > a hypothetical btf.Loader here that acts as a cache and serves as the API to load vmlinux
 
Yup! This would parse BTF more often than what we currently have though: once when the library first needs vmlinux and once for any other `btf.Loader`. The loader would make it easier to share a single copy, but would still lead to multiple copies in a single binary. Maybe that is fine though for users that care about memory pressure?

> and kmods.

This is much trickier, since kmods can come and go! Not sure yet how to deal with this.

ti-mo (approved):

PR #1025, opened 2023-04-25
allow skipping kernel version check for feature tests if test skipped via env variable

### LINKED ISSUE #822: allow silencing feature test errors
@arthurfabre reports that he'd like to be able to silence certain feature test errors, since his host kernel doesn't support CONFIG_FPROBE for example.

I'd like to keep the default behaviour of feature test tests (= when running locally) to throw an error. Otherwise developing a new feature test becomes cumbersome.

How about an environment variable that contains separated feature test names for which to ignore an error result? Basically, in https://github.com/cilium/ebpf/blob/8fceee5ca41b47a59739a14e2f6877886dc89023/internal/testutils/feature.go#L58-L62 check `tb.Name()` against an env var and call `tb.Skip()`. Brownie points if we only split the env var once or so.

This means you have to opt in to ignore the result, but you can stick that variable in your `.bashrc` or something. If your kernel changes and the test automagically isn't skipped anymore.

_Originally posted by @lmb in https://github.com/cilium/ebpf/pull/812#discussion_r990077159_

abhipranay:
closes #822

### DIFF
--- a/internal/testutils/feature.go
+++ b/internal/testutils/feature.go
@@ -3,9 +3,14 @@ package testutils
 import (
 	"errors"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/cilium/ebpf/internal"
+)
+
+const (
+	ignoreKernelVersionEnvVar = "EBPF_TEST_IGNORE_KERNEL_VERSION"
 )
 
 func CheckFeatureTest(t *testing.T, fn func() error) {
@@ -19,7 +24,11 @@ func checkFeatureTestError(t *testing.T, err error) {
 
 	var ufe *internal.UnsupportedFeatureError
 	if errors.As(err, &ufe) {
-		checkKernelVersion(t, ufe)
+		if ignoreKernelVersionCheck(t) {
+			t.Skip(ufe.Error())
+		} else {
+			checkKernelVersion(t, ufe)
+		}
 	} else {
 		t.Error("Feature test failed:", err)
 	}
@@ -108,4 +117,16 @@ func kernelVersion(tb testing.TB) internal.Version {
 		tb.Fatal(err)
 	}
 	return v
+}
+
+func ignoreKernelVersionCheck(tb testing.TB) bool {
+	if tNames := os.Getenv(ignoreKernelVersionEnvVar); tNames != "" {
+		ignored := strings.Split(os.Getenv(ignoreKernelVersionEnvVar), ",")
+		for _, n := range ignored {
+			if n == tb.Name() {
+				return true
+			}
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


internal/testutils/feature.go
```diff
@@ -19,7 +24,11 @@ func checkFeatureTestError(t *testing.T, err error) {
 
 	var ufe *internal.UnsupportedFeatureError
 	if errors.As(err, &ufe) {
-		checkKernelVersion(t, ufe)
+		if ignoreKernelVersionCheck(t) {
+			t.Skip(ufe.Error())
```
lmb, line 28: Maybe add some more context here? Otherwise it's not clear why we skipped that test.

```suggestion
			t.Skip("Ignoring error due to EBPF_TEST_IGNORE_KERNEL_VERSION:", ufe.Error())
```

internal/testutils/feature.go
```diff
@@ -109,3 +118,15 @@ func kernelVersion(tb testing.TB) internal.Version {
 	}
 	return v
 }
+
+func ignoreKernelVersionCheck(tb testing.TB) bool {
+	if tNames := os.Getenv(ignoreKernelVersionEnvVar); tNames != "" {
```
lmb, line 123: Nit: you could exit early here instead: `if tNames == "" { return false }`

internal/testutils/feature.go
```diff
@@ -109,3 +118,15 @@ func kernelVersion(tb testing.TB) internal.Version {
 	}
 	return v
 }
+
+func ignoreKernelVersionCheck(tb testing.TB) bool {
```
lmb, line 122: Could you add a test for this function?

internal/testutils/feature.go
abhipranay: I agree. I will add more context.

internal/testutils/feature.go
abhipranay: sure.

internal/testutils/feature.go
abhipranay: hmm writing test for this function seems tricky ( also less valuable ) because:
1. It uses `os.Getenv` . I can use [T.Setenv()](https://pkg.go.dev/testing#T.Setenv) but that will block us from running parallel tests.
2. I will have to inject mocked `testing.TB`. 

I can solve 2 by changing function signature to `ignoreKernelVersionCheck(tbName string)`. This new name looks more sensible to me. But 1 will still be troublesome.

lmb: TIL about T.Setenv, cool! Both 1 and 2 seem like good solutions to me.

abhipranay: I added the tests.

lmb (approved):
This looks really good! The test ended up being a lot more thorough than I expected, thanks for that :)

lmb:
Ended up squashing + adding context to the commit message.

PR #1028, opened 2023-04-27
remove unnecessary line breaks

shane965:


lmb:
Hi Shane, first off thanks for taking the time to make a PR.

As an open source project we have to be mindful what we spend resources on, in terms of review time, CI, etc. While this PR is "obviously correct" it also doesn't improve the project really, and it takes time from other activities. It's fine to fix small things like this as a drive by, but this doesn't pass the bar for a dedicated PR.

I hope this doesn't discourage you from contributing otherwise!

PR #1035, opened 2023-05-16
Add option to prevent omitting verifier logs

therealak12:
I occasionally encounter verifier errors where the output only contains a brief message, and some crucial lines are omitted, as shown below:

```
2023/05/16 13:18:52 failed to initialize bpf module, field Alb: program alb: load program: permission denied: invalid access to map value, value_size=188 off=188 size=1: R1 max value is outside of the allowed memory range (86 line(s) omitted)
```

Upon examining the code, I couldn't find a way to obtain the complete error message in the output. This PR introduces an option to fully capture and display the verifier error.

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -19,6 +19,9 @@ type CollectionOptions struct {
 	Maps     MapOptions
 	Programs ProgramOptions
 
+	// if VerboseLog is true, the verifier log lines won't be omitted
+	VerboseLog bool
+
 	// MapReplacements takes a set of Maps that will be used instead of
 	// creating new ones when loading the CollectionSpec.
 	//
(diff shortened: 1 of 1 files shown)


lmb:
Thaks for your PR! It's actually possible to get the full log, but it's arguably hard to find: https://pkg.go.dev/github.com/cilium/ebpf@v0.10.0/internal#example-VerifierError If that solves your problem we can make another PR to document this better.

therealak12:
I use `loadBpfObjects` function to load my ebpf program (which in turn uses `spec.LoadAndAssign(obj, opts)`). It doesn't matter whether I use `+` flag to print the error or not, the returned error will be printed with crucial lines omitted. And I think the cause is in the line I've changed in this PR.

BTW I'm not 100% sure I'm right. Could you please provide an example of how to rewrite this code snippet to get the full log :-?

```go
objs := bpfObjects{}
if err := loadBpfObjects(&objs, nil); err != nil {
    return nil, err
}
```

therealak12:
Perfect! Looks like the `%w` does the trick. Thanks🌸.

Issue #1036, opened 2023-05-16
Add support for XDP frags in the elf_reader


ptzianos:
Since v5.18 of the kernel, it is possible for XDP programs to process packets with MTUs larger than the size of a physical page. The feature was initially called XDP multi-buffer support and then it was renamed to XDP frags.
There are a bunch of commits related to the implementation, but one that is relevant to the codebase is this: https://github.com/torvalds/linux/commit/c2f2cdbeffda7b153c19e0f3d73149c41026c0db.
In this commit, a new flag is introduced. This flag is then added to a program section annotated with the "xdp.frags" string. Looking at the rest of the codebase, I can't see any other flags or any other changes relevant to this library, so supporting this should involve only a small amount of changes. The flag is used by the verifier to perform extra checks.

ptzianos:
@lmb I think this can be closed now right?

lmb:
Thanks :)

PR #1037, opened 2023-05-16
elf_reader: Add support for XDP frags

ptzianos:
implements #1036 

Since v5.18 of the Linux kernel, it is possible for an XDP program to process packets larger than the physical page of the system. This feature is called XDP frags.

In order for an XDP program to signal that it can process packets with large MTUs, it needs to have the "xdp.frags" annotation instead of "xdp". This new annotation causes the BPF_F_XDP_HAS_FRAGS to be added to the annotated program section of the ELF. This flag in turn causes the verifier to apply extra checks to the code. The commit along with some documentation about the feature can be seen here:
https://github.com/torvalds/linux/commit/c2f2cdbeffda7b153c19e0f3d73149c41026c0db

This commit allows the code to parse the "xdp.frags" annotation and adds the appropriate flag to the program section, along with unit tests for this behaviour. It replicated the functionality of this code: https://elixir.bootlin.com/linux/v6.0/source/tools/lib/bpf/libbpf.c#L6626


internal/unix/types_linux.go
```diff
@@ -35,6 +35,7 @@ const (
 	BPF_F_RDONLY_PROG         = linux.BPF_F_RDONLY_PROG
 	BPF_F_WRONLY_PROG         = linux.BPF_F_WRONLY_PROG
 	BPF_F_SLEEPABLE           = linux.BPF_F_SLEEPABLE
+	BPF_F_XDP_HAS_FRAGS       = linux.BPF_F_XDP_HAS_FRAGS
```
lmb, line 38: Can you add a dummy entry to `types_other.go` please?

ptzianos: sorry about that, forgot to commit the change. fixed now

lmb (approved):
LGTM, thanks for your contribution :)

PR #1040, opened 2023-05-17
program: move examples to VerifierError in main package

lmb:
Users are still struggling with VerifierError. Try to add even more documentation to point them in the right direction.


ti-mo:
Well, it being in `internal/` didn't help things. Thanks!

PR #1042, opened 2023-05-25
program: Add TestWithContext

jschwinger233:
This commit adds the TestWithContext function, which allows the acceptance of customized context for eBPF tests and returns the context modified by the eBPF program.

The input context will be cast to a `__sk_buff` struct for the tc program. For those who want to pass an skb with a special skb->mark or check the returned skb's metadata, TestWithContext can be useful for that purpose.

Signed-off-by: Zhichuan Liang <gray.liang@isovalent.com>

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -583,6 +583,41 @@ func (p *Program) Test(in []byte) (uint32, []byte, error) {
 	return ret, opts.DataOut, nil
 }
 
+// TestWithContext runs the Program in the kernel with the given data and
+// context, and gives back the data and context returned by eBPF program.
+// outLen may be zero.
+//
+// Note: the kernel expects at least 14 bytes input for an ethernet header for
+// XDP and SKB programs.
+//
+// This function requires at least Linux 4.12.
+func (p *Program) TestWithContext(data, ctx []byte) (uint32, []byte, []byte, error) {
+	// Older kernels ignore the dataSizeOut argument when copying to user space.
+	// Combined with things like bpf_xdp_adjust_head() we don't really know what the final
+	// size will be. Hence we allocate an output buffer which we hope will always be large
+	// enough, and panic if the kernel wrote past the end of the allocation.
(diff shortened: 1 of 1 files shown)


ti-mo:
@jschwinger233 Does `Program.Run()` with setting the appropriate `RunOptions` not do what you need?

jschwinger233:
@ti-mo Thank you for reply. I did notice `Run` function, my concern is this special logic for outputPad at: https://github.com/cilium/ebpf/blob/211b0e67969970c05124fe9d694e56d4aec99438/prog.go#L562-L571

Calling `Run()` directly requires to expose that magic pad to user, and I assume this is something internal that we'd like to hide from caller?

lmb:
The problem was fixed in https://github.com/torvalds/linux/commit/b5a36b1e1b138285ea0df34bf96c759e1e30fafd aka 5.0-rc1 if I read things right. You also need to call skb_adjust_head or similar to trigger it, with a non-nil `DataOut`. Does that apply in your case?

Do you want to call this from production code or from tests? If it's just from tests I'd suggest we just turn the panic into an error, and have the caller deal with it. For production we might need to solve this in `run()` instead.

jschwinger233:
@lmb I'm adding BPF test cases for cilium's IPSec feature, which requires me to check skb->mark after an execution of BPF program, please see the draft PR: https://github.com/cilium/cilium/pull/25699

In the PR above, I have to define a function `runBpfProgram` at `test/bpf_tests/bpf_test.go`, where `dataOut = make([]byte, len(data)+256+2)` looks magic. Maybe I can leave some comments there to explain the reason? What would you suggest?

lmb:
> where dataOut = make([]byte, len(data)+256+2) looks magic. 

Maybe just allocate a maximally sized `dataOut`? Seems like that's what you're already doing for data.

jschwinger233:
> Maybe just allocate a maximally sized dataOut

Sounds good, thank you!

PR #1044, opened 2023-05-25
link: add test for UDP iter

lmb:
Ensure that we can create a global UDP iterator.


lmb:
@aditighag it's possible to create an iterator, but we'll have to add support for passing a netns fd. I can take a stab at that if you have code with proposed UAPI somewhere.

aditighag:
The details around the UAPI extensions are still tbd. If the TCP/UDP iterators are getting loaded by the library as expected, then it's all good for now.

PR #1045, opened 2023-05-25
btf: avoid triggering feature probe on import

lmb:
Importing package btf currently executes a feature test since we use it to initialize a global variable. Move initialization to a function to avoid this.


brycekahle (approved):

rgo3 (approved):
LGTM, the CI failure is just a semaphore hickup I guess?

PR #1051, opened 2023-06-01
internal/kconfig: Add possibility to filter kconfig file

eiffel-fl:
Hi.

This commit introduces a new arguments to Parse(), it will be used to only add CONFIG_* corresponding to this map key to the returned map.

In this way, it permits reducing memory allocation, by only having CONFIG_* present in the eBPF code and not all the CONFIG_* set for the running kernel. This behavior mimics that of libbpf [1].

In terms of memory allocation, by filtering only with CONFIG_HZ, we reduced allocated bytes by approximately 79%:
BenchmarkParse-8         404 2484446 ns/op 1617336 B/op 9718 allocs/op
BenchmarkParseFiltered-8 751 1582447 ns/op  342272 B/op 9505 allocs/op

Best regards and thank you in advance for the review.

### DIFF
--- a/internal/kconfig/kconfig.go
+++ b/internal/kconfig/kconfig.go
@@ -99,9 +99,10 @@ func processKconfigLine(line []byte, m map[string]string, filter map[string]stru
 		return fmt.Errorf("line %q has no value", line)
 	}
 
-	k := string(key)
 	if filter != nil {
-		_, ok := filter[k]
+		// NB: map[string(key)] gets special optimisation help from the compiler
+		// and doesn't allocate. Don't turn this into a variable.
+		_, ok := filter[string(key)]
 		if !ok {
 			return nil
 		}
@@ -110,9 +111,9 @@ func processKconfigLine(line []byte, m map[string]string, filter map[string]stru
 	// This can seem odd, but libbpf only sets the value the first time the key is
 	// met:
 	// https://github.com/torvalds/linux/blob/0d85b27b0cc6/tools/lib/bpf/libbpf.c#L1906-L1908
-	_, ok := m[k]
+	_, ok := m[string(key)]
 	if !ok {
-		m[k] = string(value)
+		m[string(key)] = string(value)
 	}
 
 	return nil


internal/kconfig/kconfig.go
```diff
@@ -78,7 +78,7 @@ func Parse(source io.ReaderAt) (map[string]string, error) {
 // Golang translation of libbpf bpf_object__process_kconfig_line():
 // https://github.com/libbpf/libbpf/blob/fbd60dbff51c870f5e80a17c4f2fd639eb80af90/src/libbpf.c#L1874
 // It does the same checks but does not put the data inside the BPF map.
-func processKconfigLine(line string, m map[string]string) error {
+func processKconfigLine(line string, m map[string]string, filteredConfigs map[string]any) error {
```
lmb, line 81: This function should take `line []byte` and use `bytes.HasPrefix`, `bytes.Cut` and so on. This avoids allocating `string`s for lines we're not interested in.

internal/kconfig/kconfig.go
```diff
@@ -58,7 +58,7 @@ func Parse(source io.ReaderAt) (map[string]string, error) {
 	for s.Scan() {
 
 		line := s.Text()
-		err = processKconfigLine(line, ret)
+		err = processKconfigLine(line, ret, filteredConfigs)
```
lmb, line 61: The result map should be pre-allocated to the number of configs we are interested in:

```
ret := make(map[string]string, len(filteredConfigs))
```

This is because in the benchmark we spend a lot of time growing / copying the map.

internal/kconfig/kconfig.go
```diff
@@ -39,10 +39,10 @@ func Find() (*os.File, error) {
 }
 
 // Parse parses the kconfig file for which a reader is given.
-// All the CONFIG_* set will be put in the returned map as key with their
-// corresponding value as map value.
+// All the CONFIG_* which are in filteredConfigs and which are set set will be
+// put in the returned map as key with their corresponding value as map value.
 // If the kconfig file is not valid, error will be returned.
-func Parse(source io.ReaderAt) (map[string]string, error) {
+func Parse(source io.ReaderAt, filteredConfigs map[string]any) (map[string]string, error) {
```
lmb, line 45: `map[string]any` is kind of convenient for where we are trying to call this function from, but it's not good API since it doesn't really document itself.

I'd prefer `map[string]struct{}` or `map[string]bool` even though it means another map allocated in the caller.

internal/kconfig/kconfig.go
```diff
@@ -39,10 +39,10 @@ func Find() (*os.File, error) {
 }
 
 // Parse parses the kconfig file for which a reader is given.
-// All the CONFIG_* set will be put in the returned map as key with their
-// corresponding value as map value.
+// All the CONFIG_* which are in filteredConfigs and which are set set will be
+// put in the returned map as key with their corresponding value as map value.
```
lmb, line 43: This doesn't document what a `nil` filteredConfigs means. From the tests I gather it means "no filter".

internal/kconfig/kconfig.go
```diff
@@ -58,7 +58,7 @@ func Parse(source io.ReaderAt) (map[string]string, error) {
 	for s.Scan() {
```
lmb, line 58: We could bail out of this loop once we've found all the filteredConfigs. Upside is much faster parsing if the configs are towards the beginning of the file, downside is that perf is a lot less predicable and we might not discover some parsing error. What do you think?

```suggestion
	for s.Scan() && (filteredConfigs != nil && len(ret) != len(filteredConfigs)) {
```

internal/kconfig/kconfig_test.go
```diff
@@ -39,21 +59,41 @@ func TestParseGziped(t *testing.T) {
...
 	defer f.Close()
 
-	_, err = Parse(f)
+	_, err = Parse(f, nil)
+	if err != nil {
+		t.Fatal("Error parsing gziped kconfig: ", err)
+	}
+}
+
+func TestParseGzipedFiltered(t *testing.T) {
```
lmb, line 68: Seems like you removed the simple test case that didn't use the fedora kconfig? I liked that one since it's simple and asserts that we get the right values out. We're missing that coverage now?

I'd prefer it if we had two tests:
* TestParse against a minimal kconfig which asserts values
* TestParseFiltered against a minimal kconfig which asserts the presence of keys and size of map

lmb (changes requested):
I think this is a good approach. I know that Timo wants to experiment with using a trie, but it's always better to avoid work rather than trying to optimize more.

There are some things we can do to speed this up some more, left a couple of comments.

lmb (approved):
I pushed a small commit which gets rid of some more allocations. I think that filtered kconfig is now plenty fast:

```
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf/internal/kconfig
                 │  base.txt   │
                 │   sec/op    │
Parse-16           2.071m ± 3%
ParseFiltered-16   131.6µ ± 1%
geomean            522.1µ

                 │   base.txt   │
                 │     B/op     │
Parse-16           1.433Mi ± 0%
ParseFiltered-16   50.17Ki ± 0%
geomean            271.4Ki

                 │  base.txt   │
                 │  allocs/op  │
Parse-16           7.331k ± 0%
ParseFiltered-16    33.00 ± 0%
geomean             491.9
```

@ti-mo are you OK with this approach?

lmb:
> should I squash it with the first one?

I'd keep it separate: it has an explanation in the commit message and it makes tracking of code provenance (signed-off-by) easier.

lmb:
Here you go: https://github.com/golang/go/issues/3512

internal/kconfig/kconfig_test.go
```diff
@@ -23,13 +23,83 @@ func BenchmarkParse(b *testing.B) {
...
+	if err != nil {
+		t.Fatal("Error reading /testdata/test.kconfig: ", err)
+	}
+	defer f.Close()
+
+	filter := map[string]bool{"CONFIG_FOO": true}
+
+	config, err := Parse(f, filter)
+	if err != nil {
+		t.Fatal("Error parsing gziped kconfig: ", err)
```
ti-mo, line 96:
```suggestion
		t.Fatal("Error parsing gzipped kconfig: ", err)
```

internal/kconfig/kconfig_test.go
```diff
@@ -23,13 +23,83 @@ func BenchmarkParse(b *testing.B) {
...
+	}
+
+	expected := map[string]string{"CONFIG_FOO": `"foo"`}
+	qt.Assert(t, config, qt.DeepEquals, expected)
+}
+
 func TestParseGziped(t *testing.T) {
```
ti-mo, line 103:
```suggestion
func TestParseGzipped(t *testing.T) {
```

ti-mo (approved):
Thanks! Left a few nits.

About the approach: I think we're generally more concerned about reducing allocs than reducing CPU usage. Creating garbage has cascading effects and is expensive to clean up, and a larger steady-state memory footprint (e.g. if we parse the whole kconfig and hold onto it) is problematic for many users, given the requests we've gotten for e.g. flushing vmlinux BTF etc. If we can avoid resident memory, let's do it. :rocket:

lmb:
Please make sure to push formatted files: https://ebpf.semaphoreci.com/jobs/01c36378-27e1-413e-8f2d-d808a315f95e#L722

lmb:
Thank you!

PR #1052, opened 2023-06-06
Pre-allocate map sys.Errors

hmahmood:
This is showing up in memory profiles on some of our internal deployments. Seems like an easy gain.

### DIFF
--- a/syscalls.go
+++ b/syscalls.go
@@ -14,6 +14,15 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
 
+var (
+	// pre-allocating these here since they may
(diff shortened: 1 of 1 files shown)


syscalls.go
```diff
@@ -14,6 +14,12 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
 
+var (
```
lmb, line 17: Can you add a comment why we're preallocating them here?

lmb (changes requested):
Nice! One nit, otherwise this is good to go.

lmb (approved):
Thanks!

PR #1054, opened 2023-06-07
bpf2go: argument parsing unit test

mejedi:
Enable unit tests for argument parsing in `bpf2go`. This is in preparation for enabling bpf2go to take some arguments from the  environment.

The patch moves argument parsing from `run` into a separate function.

### DIFF
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -69,7 +69,50 @@ var targetByGoArch = map[string]target{
 }
 
 func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
-	b2g := bpf2go{
+	b2g, err := newB2G(stdout, pkg, outputDir, args)
+	switch {
+	case err == nil:
+		return b2g.convertAll()
+	case errors.Is(err, flag.ErrHelp):
+		return nil
+	default:
+		return err
+	}
+}
+
+type bpf2go struct {
+	stdout io.Writer
+	// Absolute path to a .c file.
+	sourceFile string
+	// Absolute path to a directory where .go are written
+	outputDir string
+	// Alternative output stem. If empty, identStem is used.
+	outputStem string
+	// Valid go package name.
+	pkg string
+	// Valid go identifier.
+	identStem string
+	// Targets to build for.
+	targetArches map[target][]string
+	// C compiler.
+	cc string
+	// Command used to strip DWARF.
+	strip            string
+	disableStripping bool
+	// C flags passed to the compiler.
+	cFlags          []string
+	skipGlobalTypes bool
+	// C types to include in the generatd output.
+	cTypes cTypes
+	// Build tags to be included in the output.
+	tags buildTags
+	// Base directory of the Makefile. Enables outputting make-style dependencies
+	// in .d files.
+	makeBase string
+}
+
+func newB2G(stdout io.Writer, pkg, outputDir string, args []string) (*bpf2go, error) {
+	b2g := &bpf2go{
 		stdout:    stdout,
 		pkg:       pkg,
 		outputDir: outputDir,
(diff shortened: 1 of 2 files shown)


cmd/bpf2go/main.go
```diff
@@ -75,6 +75,49 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
+	case err == nil:
+		return b2g.convertAll()
+	case errors.Is(err, flag.ErrHelp):
+		return nil
+	default:
+		return err
+	}
+}
+
+type bpf2go struct {
```
mejedi, line 89: Moving so that methods foolow the struct definition.

cmd/bpf2go/main.go
```diff
@@ -75,6 +75,49 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
+	// Absolute path to a directory where .go are written
+	outputDir string
+	// Alternative output stem. If empty, identStem is used.
+	outputStem string
+	// Valid go package name.
+	pkg string
+	// Valid go identifier.
+	identStem string
+	// Targets to build for.
+	targetArches map[target][]string
```
mejedi, line 102: Introducing `targetArches`.

cmd/bpf2go/main.go
```diff
@@ -161,40 +202,27 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
-		// Try to find a suitable llvm-strip, possibly with a version suffix derived
-		// from the clang binary.
-		if b2g.strip == "" {
-			b2g.strip = "llvm-strip"
-			if strings.HasPrefix(b2g.cc, "clang") {
-				b2g.strip += strings.TrimPrefix(b2g.cc, "clang")
-			}
-		}
-
-		b2g.strip, err = exec.LookPath(b2g.strip)
```
mejedi, line 189: Moving `LookupPath` out of `parseArgs` so that unit tests can pass arbitrary values in `-strip` flag.

cmd/bpf2go/main.go
```diff
@@ -161,40 +202,27 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
-	if len(targetArches) == 0 {
-		return fmt.Errorf("no targets specified")
-	}
-
-	targets, err := collectTargets(targetArches)
+	targetArches, err := collectTargets(strings.Split(*flagTarget, ","))
 	if errors.Is(err, errInvalidTarget) {
-		printTargets(stdout)
-		fmt.Fprintln(stdout)
+		printTargets(b2g.stdout)
```
lmb, line 207: Maybe move these into the caller in a separate `errors.Is(err, errInvalidTarget)` branch? Then we don't print at all in this function as far as I can tell.

cmd/bpf2go/main.go
```diff
@@ -75,6 +75,49 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
+	// C types to include in the generatd output.
+	cTypes cTypes
+	// Build tags to be included in the output.
+	tags buildTags
+	// Base directory of the Makefile. Enables outputting make-style dependencies
+	// in .d files.
+	makeBase string
+}
+
+func (b2g *bpf2go) parseArgs(args []string) (err error) {
```
lmb, line 120: Nit: my preference is to have constructor like function as much as possible. How would you feel about `newB2G(stdout io.Writer, pkg, outputDir string, args []string) (*b2g, error)`?

lmb (changes requested):
Looks great, thanks for the quick turn around!

cmd/bpf2go/main.go
mejedi: Done.

cmd/bpf2go/main.go
```diff
@@ -123,82 +164,63 @@ func run(stdout io.Writer, pkg, outputDir string, args []string) (err error) {
...
+		return nil, errors.New("expected at least two arguments")
 	}
 
 	b2g.identStem = args[0]
 	if !token.IsIdentifier(b2g.identStem) {
-		return fmt.Errorf("%q is not a valid identifier", b2g.identStem)
-	}
-
-	input := args[1]
-	if _, err := os.Stat(input); os.IsNotExist(err) {
```
mejedi, line 142: Moving `os.Stat` out of `newB2G` so that unit tests can pass arbitrary values.

lmb:
Fixed a nit and rebased, thanks for refactoring this!

lmb (approved):

mejedi:
@lmb thank you for the reviews! Any chance this could be merged?

PR #1055, opened 2023-06-07
internal/sys: don't panic in init()

lmb:
Remove the panic when initializing the profSet. It's causing friction when k8s tried to update the vendored dependency.


lmb:
Context: https://github.com/kubernetes/kubernetes/pull/117457

ti-mo (approved):

PR #1062, opened 2023-06-14
map: zero-allocation operations for common types

lmb:
map: zero-allocation operations for common types

    Map keys and values are currently marshaled into []byte by souped up
    versions of binary.Write and binary.Read. This allows users to be blissfully
    unaware of compiler inserted padding on the Go side. This is wasteful in
    case the Go in-memory representation matches what the kernel expects because
    we need additional allocations.

    Refactor syscall marshaling into a new package sysenc which encapsulates the
    logic we need to determine whether a Go type is safe for zero-allocation /
    zero-copy marshaling. The type must be a pointer to or a slice of:

    * A primitive type like uint32, ... or
    * An array of valid types or
    * A struct made up of valid types without any compiler
     inserted padding between fields

    Per-CPU maps don't support zero-allocation operations for now, but the new
    code already makes things a little bit cheaper.

    Structs with trailing padding also don't benefit from the optimization for
    now. Consider

        type padded struct { A uint32; B uint16 }

    Allowing such a type creates an edge case: make([]padding, 1) uses
    zero-allocation marshaling while make([]padding, 2) doesn't, due to interior
    padding. It's simpler to skip such types for now.

        goos: linux
       goarch: amd64
       pkg: github.com/cilium/ebpf
                                            │ unsafe.txt  │
                                            │   sec/op    │
       Marshaling/ValueUnmarshalReflect-16    356.1n ± 2%
       Marshaling/KeyMarshalReflect-16        368.6n ± 1%
       Marshaling/ValueBinaryUnmarshaler-16   378.6n ± 2%
       Marshaling/KeyBinaryMarshaler-16       356.2n ± 1%
       Marshaling/KeyValueUnsafe-16           328.0n ± 2%
       PerCPUMarshalling/reflection-16        1.232µ ± 1%

                                             │  unsafe.txt  │
                                            │     B/op     │
       Marshaling/ValueUnmarshalReflect-16    0.000 ± 0%
       Marshaling/KeyMarshalReflect-16        0.000 ± 0%
       Marshaling/ValueBinaryUnmarshaler-16   24.00 ± 0%
       Marshaling/KeyBinaryMarshaler-16       8.000 ± 0%
       Marshaling/KeyValueUnsafe-16           0.000 ± 0%
       PerCPUMarshalling/reflection-16        280.0 ± 0%

                                             │  unsafe.txt  │
                                            │  allocs/op   │
       Marshaling/ValueUnmarshalReflect-16    0.000 ± 0%
       Marshaling/KeyMarshalReflect-16        0.000 ± 0%
       Marshaling/ValueBinaryUnmarshaler-16   1.000 ± 0%
       Marshaling/KeyBinaryMarshaler-16       1.000 ± 0%
       Marshaling/KeyValueUnsafe-16           0.000 ± 0%
       PerCPUMarshalling/reflection-16        3.000 ± 0%

### DIFF
--- a/internal/sysenc/layout.go
+++ b/internal/sysenc/layout.go
@@ -1,0 +1,41 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found at https://go.dev/LICENSE.
+
+package sysenc
+
+import (
+	"reflect"
+	"sync"
+)
+
+var hasUnexportedFieldsCache sync.Map // map[reflect.Type]bool
+
+func hasUnexportedFields(typ reflect.Type) bool {
+	switch typ.Kind() {
+	case reflect.Slice, reflect.Array, reflect.Pointer:
+		return hasUnexportedFields(typ.Elem())
+
+	case reflect.Struct:
+		if unexported, ok := hasUnexportedFieldsCache.Load(typ); ok {
+			return unexported.(bool)
+		}
+
+		unexported := false
+		for i, n := 0, typ.NumField(); i < n; i++ {
+			field := typ.Field(i)
+			// Package binary allows _ fields but always writes zeroes into them.
+			if (!field.IsExported() && field.Name != "_") || hasUnexportedFields(field.Type) {
+				unexported = true
+				break
+			}
+		}
+
+		hasUnexportedFieldsCache.Store(typ, unexported)
+		return unexported
+
+	default:
+		// NB: It's not clear what this means for Chan and so on.
+		return false
+	}
+}
--- a/internal/sysenc/layout_test.go
+++ b/internal/sysenc/layout_test.go
@@ -1,0 +1,37 @@
+package sysenc
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+func TestHasUnexportedFields(t *testing.T) {
+	for _, test := range []struct {
+		value  any
+		result bool
+	}{
+		{struct{ A any }{}, false},
+		{(*struct{ A any })(nil), false},
+		{([]struct{ A any })(nil), false},
+		{[1]struct{ A any }{}, false},
+		{struct{ _ any }{}, false},
+		{struct{ _ struct{ a any } }{}, true},
+		{(*struct{ _ any })(nil), false},
+		{([]struct{ _ any })(nil), false},
+		{[1]struct{ _ any }{}, false},
+		{struct{ a any }{}, true},
+		{(*struct{ a any })(nil), true},
+		{([]struct{ a any })(nil), true},
+		{[1]struct{ a any }{}, true},
+		{(*struct{ A []struct{ a any } })(nil), true},
+		{(*struct{ A [1]struct{ a any } })(nil), true},
+	} {
+		t.Run(fmt.Sprintf("%T", test.value), func(t *testing.T) {
+			have := hasUnexportedFields(reflect.TypeOf(test.value))
+			qt.Assert(t, have, qt.Equals, test.result)
+		})
+	}
+}
--- a/internal/sysenc/marshal.go
+++ b/internal/sysenc/marshal.go
@@ -1,0 +1,163 @@
+package sysenc
+
+import (
+	"bytes"
+	"encoding"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"reflect"
+	"sync"
+	"unsafe"
+
+	"github.com/cilium/ebpf/internal"
+)
+
+// Marshal turns data into a byte slice using the system's native endianness.
+//
+// If possible, avoids allocations by directly using the backing memory
+// of data. This means that the variable must not be modified for the lifetime
+// of the returned [Buffer].
+//
+// Returns an error if the data can't be turned into a byte slice according to
+// the behaviour of [binary.Write].
+func Marshal(data any, size int) (Buffer, error) {
+	if data == nil {
+		return Buffer{}, errors.New("can't marshal a nil value")
+	}
+
+	var buf []byte
+	var err error
+	switch value := data.(type) {
+	case encoding.BinaryMarshaler:
+		buf, err = value.MarshalBinary()
+	case string:
+		buf = unsafe.Slice(unsafe.StringData(value), len(value))
+	case []byte:
+		buf = value
+	case int16:
+		buf = internal.NativeEndian.AppendUint16(make([]byte, 0, 2), uint16(value))
+	case uint16:
+		buf = internal.NativeEndian.AppendUint16(make([]byte, 0, 2), value)
+	case int32:
+		buf = internal.NativeEndian.AppendUint32(make([]byte, 0, 4), uint32(value))
+	case uint32:
+		buf = internal.NativeEndian.AppendUint32(make([]byte, 0, 4), value)
+	case int64:
+		buf = internal.NativeEndian.AppendUint64(make([]byte, 0, 8), uint64(value))
+	case uint64:
+		buf = internal.NativeEndian.AppendUint64(make([]byte, 0, 8), value)
+	default:
+		if buf := unsafeBackingMemory(data); len(buf) == size {
+			return newBuffer(buf), nil
+		}
+
+		wr := internal.NewBuffer(make([]byte, 0, size))
+		defer internal.PutBuffer(wr)
+
+		err = binary.Write(wr, internal.NativeEndian, value)
+		buf = wr.Bytes()
+	}
+	if err != nil {
+		return Buffer{}, err
+	}
+
+	if len(buf) != size {
+		return Buffer{}, fmt.Errorf("%T doesn't marshal to %d bytes", data, size)
+	}
+
+	return newBuffer(buf), nil
+}
+
+var bytesReaderPool = sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Reader)
+	},
+}
+
+// Unmarshal a byte slice in the system's native endianness into data.
+//
+// Returns an error if buf can't be unmarshalled according to the behaviour
+// of [binary.Read].
+func Unmarshal(data interface{}, buf []byte) error {
+	switch value := data.(type) {
+	case encoding.BinaryUnmarshaler:
+		return value.UnmarshalBinary(buf)
+
+	case *string:
+		*value = string(buf)
+		return nil
+
+	default:
+		if dataBuf := unsafeBackingMemory(data); len(dataBuf) == len(buf) {
+			copy(dataBuf, buf)
+			return nil
+		}
+
+		rd := bytesReaderPool.Get().(*bytes.Reader)
+		defer bytesReaderPool.Put(rd)
+
+		rd.Reset(buf)
+
+		return binary.Read(rd, internal.NativeEndian, value)
+	}
+}
+
+// unsafeBackingMemory returns the backing memory of data if it can be used
+// instead of calling into package binary.
+//
+// Returns nil if the value is not a pointer or a slice, or if it contains
+// padding or unexported fields.
+func unsafeBackingMemory(data any) []byte {
+	if data == nil {
+		return nil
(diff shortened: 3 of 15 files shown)


internal/sysenc/marshal.go
```diff
@@ -0,0 +1,141 @@
...
+	if data == nil {
+		return Buffer{}, errors.New("can't marshal a nil value")
+	}
+
+	layout := dataLayout{1, size, size}
+
+	var buf []byte
+	var err error
+	switch value := data.(type) {
+	case encoding.BinaryMarshaler: // TODO: Should we support this here?
```
lmb, line 36: Beep

internal/sysenc/marshal.go
```diff
@@ -0,0 +1,151 @@
...
+
+		wr := bytesBufferPool.Get().(*bytes.Buffer)
+		defer bytesBufferPool.Put(wr)
+
+		// Reinitialize the Buffer with a new backing slice since it is returned to
+		// the caller by wr.Bytes() below. Pooling is faster despite calling
+		// NewBuffer. The pooled alloc is still reused, it only needs to be zeroed.
+		*wr = *bytes.NewBuffer(make([]byte, 0, size))
+
+		err = binary.Write(wr, internal.NativeEndian, value)
```
mejedi, line 68: Please forgive me my ignorance, but won't it be much easier to verify compatibility by round-tripping a buffer of `Sizeof(T)` `FF`s via `binary.Read` followed by `binary.Write` and checking that we got the same value back?

mejedi: Well, comparing `Sizeof(T)` with the size of blob produced/consumed by `binary.Write()` / `binary.Read()` will tell whether `T` has any compiler-inserted padding.

It still remains to recursively check `T` to see if there are any struct fields that are treated as padding by `encoding/binary`, but tracking offsets/sizes won't be necessary.

lmb: You mean comparing `reflect.Type.Size()` with `binary.Size`? True, that would be easier, but it doesn't allow detecting the case where a struct has implicit trailing padding. The current code allows zero copy from something like this:

```go
struct {
    A uint32
    B uint16
    // two bytes of implicit padding
}
```

You're right that the `dataLayout` machinery is kind of involved. The reason it's this way is because I was looking at doing zero-copy marshaling for per-CPU maps as well. It turned out too big of a change at once though, but it's something I'd like to revisit at some point.

mejedi: > The current code allows zero copy from something like this:

I'm curious when this is beneficial. My assumption is that this patch doesn't change any user-visible behaviours, apart from making things faster.

`marshalBytes` checks that resulting buf has the size as expected. With trailing implicit padding it won't be the case, unless C structure is packed. Indeed, `binary.Write` will produce 6 bytes while 8 are expected. Packed structs aren't that abundant to make it worthwhile, are they?

`unmarshalBytes` on the other hand is ok if fewer bytes are consumed. Is that the case you are optimising for?

> I was looking at doing zero-copy marshaling for per-CPU maps as well

Trying to wrap my head around it, does it mean reading into correctly sized slice from bpf syscall directly? Will it work for updates as well? Presumably currently it is ok to pass a shorter slice.

lmb: > My assumption is that this patch doesn't change any user-visible behaviours, apart from making things faster.

That's the intention!

> Packed structs aren't that abundant to make it worthwhile, are they?

Okay, I think you're saying the following: given a C structure that has implicit padding, we'd declare a map that has a key / value length which includes the padding. On the Go side we'd be looking for a length of 8 instead of 6. So the fast past wouldn't apply, in fact we'd bail out of marshaling since the Go struct doesn't have the correct length.

This is true, I hadn't thought of that. My angle is that the fast path should apply for a struct even though it has trailing padding, since 6 is the "packed" / binary.Size of that struct. Keep in mind that Go and C structures don't always have to correspond 1:1, for example Cilium doesn't use bpf2go and uses hand written Go types + a linter instead.

Is there something specific you're worried about with this "optimization"? Or is this because you think we shouldn't have the layout code at all?

> Trying to wrap my head around it, does it mean reading into correctly sized slice from bpf syscall directly? Will it work for updates as well?

That's what I would like, yes. The insight is that the layout of a per-CPU map for syscalls is:

```
[nCPU]struct{ [valueSize]byte; [align(valueSize, 8)-valueSize]byte }
```

`align(valueSize, 8)` is just `stride` in the layout code.

If you store something like `uint64` in the per CPU map there is no padding and the syscall could write into Go memory directly. This is really useful for storing counters, so having them be more efficient is always nice. OTOH we've been living with the reflect based decoding for a long time, so maybe there isn't such a pressing need.

lmb: I tried your idea of using `binary.Size` instead of the custom logic in `layout.go`. It doesn't work because `binary.Size` doesn't reject unexported fields like in `struct{ a uint32 }`. We can't get rid of the custom reflect traversal unfortunately.

lmb:
@mejedi I'll be doing a code walk of this PR tomorrow 9:30 UTC+1, let me know if you'd like to join.

mejedi:
@lmb 
> @mejedi I'll be doing a code walk of this PR tomorrow 9:30 UTC+1, let me know if you'd like to join.

Thank you for the invitation! I can’t join, I’m afraid. I’m pretty much preoccupied with moving apartments this week :(

internal/sysenc/marshal.go
```diff
@@ -0,0 +1,138 @@
...
+	return newBuffer(buf), nil
+}
+
+var bytesReaderPool = sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Reader)
+	},
+}
+
+func Unmarshal(data interface{}, buf []byte) error {
```
lmb, line 75: Add doc

internal/sysenc/layout.go
```diff
@@ -0,0 +1,99 @@
...
+		isSlice = true
+	}
+
+	var size int
+	if cachedSize, ok := cachedSizes.Load(typ); ok {
+		size = cachedSize.(int)
+	} else {
+		size = sizeOfRecurse(typ)
+		if typ.Kind() == reflect.Struct {
+			cachedSizes.Store(typ, size)
```
lmb: Doc why: struct is the most expensive type due to reflect api shortcoming.

internal/sysenc/layout.go
```diff
@@ -0,0 +1,99 @@
...
+			size *= t.Len()
+		}
+		return size
+
+	case reflect.Struct:
+		offset := 0
+		for i, n := 0, t.NumField(); i < n; i++ {
+			field := t.Field(i)
+			if !field.IsExported() && field.Name != "_" {
+				return -1
```
lmb: Doc: binary.Read also allows padding but always fills it with zeroes.

internal/sysenc/layout.go
```diff
@@ -0,0 +1,99 @@
...
+
+			offset += size
+		}
+
+		if t.Size() != uintptr(offset) {
+			// There is trailing padding in the struct.
+			return -1
+		}
+
+		return -1
```
lmb: This is probably broken

lmb:
I removed the custom layout code after discussing it with Robin and Timo.

internal/sysenc/marshal.go
```diff
@@ -0,0 +1,159 @@
...
+		if elemType := value.Type().Elem(); elemType.Kind() != reflect.Slice {
+			valueSize = int(elemType.Size())
+			break
+		}
+
+		// Return a pointer to the first element, not a pointer to the slice header.
+		value = value.Elem()
+		fallthrough
+
+	case reflect.Slice:
```
ti-mo, line 130: Do we also need a `value.IsNil()` here? Or is `([]byte)(nil)` caught by the nil check on `data`?

internal/sysenc/marshal.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		if elemType := value.Type().Elem(); elemType.Kind() != reflect.Slice {
+			valueSize = int(elemType.Size())
+			break
+		}
+
+		// Return a pointer to the first element, not a pointer to the slice header.
+		value = value.Elem()
+		fallthrough
```
ti-mo, line 128: This is tricky code, I prefer a small repetition instead of relying fallthrough.

internal/sysenc/marshal.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil
+	}
+
+	// Reinterpret the pointer as a byte slice. This violates the unsafe.Pointer
+	// rules because it's very unlikely that the source data has "an equivalent
+	// memory layout". However, we can make it safe-ish because of the
+	// following reasons:
+	//  - There is no alignment mismatch since we cast to a type with an
+	//    alignment of 1.
+	//  - There are no pointers in the source type so we don't upset the GC.
```
ti-mo, line 156: Where do we validate this specifically? If my reading is correct, `hasUnexportedFields` happily chases pointers and resolves them too. binary.Size() should also tolerate structs containing pointers iirc. It would be nice to validate this in tests explicitly, I don't see a case like this at first glance.

marshalers.go
```diff
@@ -1,166 +1,55 @@
...
 }
 
 // marshalPerCPUValue encodes a slice containing one value per
 // possible CPU into a buffer of bytes.
 //
 // Values are initialized to zero if the slice has less elements than CPUs.
 //
-// slice must have a type like []elementType.
-func marshalPerCPUValue(slice interface{}, elemLength int) (sys.Pointer, error) {
+// slice must have a type like [](*)elementType.
```
ti-mo, line 51: This is a bit unclear; if either pointer types or scalars can appear in the slice, is the comment still necessary? I think this is now basically `slice must be a slice`. :smile:

marshalers.go
```diff
@@ -197,8 +86,8 @@ func marshalPerCPUValue(slice interface{}, elemLength int) (sys.Pointer, error)
 // unmarshalPerCPUValue decodes a buffer into a slice containing one value per
 // possible CPU.
 //
-// valueOut must have a type like *[]elementType
-func unmarshalPerCPUValue(slicePtr interface{}, elemLength int, buf []byte) error {
+// valueOut must have a type like *[](*)elementType.
```
ti-mo, line 89: Idem, this should be `valueOut must be a pointer to a slice` or similar.

ti-mo (commented):
Sorry for leaving this unattended for so long. I have a few questions and left a few comments. Thanks! :pray:

internal/sysenc/layout.go
```diff
@@ -0,0 +1,41 @@
...
+	"reflect"
+	"sync"
+)
+
+var hasUnexportedFieldsCache sync.Map // map[reflect.Type]bool
+
+func hasUnexportedFields(typ reflect.Type) bool {
+	switch typ.Kind() {
+	case reflect.Slice, reflect.Array, reflect.Pointer:
+		return hasUnexportedFields(typ.Elem())
```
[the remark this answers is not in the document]
rgo3: https://pkg.go.dev/reflect#Type says `Elem() Type panics if the type's Kind is not Array, Chan, Map, Pointer, or Slice.`. I am not quite sure where I would find the implementations of the `Type` interface to see an actual implementation of `Elem() Type`, but https://pkg.go.dev/reflect#Value.Elem is not what is being called.

internal/sysenc/layout_test.go
```diff
@@ -0,0 +1,33 @@
...
+		{(*struct{ A any })(nil), false},
+		{([]struct{ A any })(nil), false},
+		{struct{ _ any }{}, false},
+		{struct{ _ struct{ a any } }{}, true},
+		{(*struct{ _ any })(nil), false},
+		{([]struct{ _ any })(nil), false},
+		{struct{ a any }{}, true},
+		{(*struct{ a any })(nil), true},
+		{([]struct{ a any })(nil), true},
+		{(*struct{ A []struct{ a any } })(nil), true},
```
lmb, line 26: Add an array case

lmb:
@ti-mo this is good to go.

PR #1069, opened 2023-06-20
btf: move Spec.Add to dedicated type

lmb:
internal: add NewBuffer

    Share a single pool of bytes.Buffer across the whole library.

btf: add marshalOptions.Order

    Allow specifying which ByteOrder to encode BTF into via a new field in
    marshalOptions.

btf: use number of types to pre-size string table

    Remove the dependency on the ELF string table to pre-size the string table
    builder for large Specs. Instead we use the heuristic that for vmlinux the
    number of strings is roughly equal to the number of types.

btf: refactor to Builder

    Spec allows both querying Types as well as constructing new BTF blobs via
    Add. This has some unfortunate consequences: we need to do extra work so
    that Added types can be queried and query code needs to be aware that the
    set of types can change. The query API also allows observing side-effects of
    adding types which we'd rather hide, for example when working around datasec
    bugs. It's also difficult to make the implementation efficient for both use
    cases at once.

    Split the code for constructing new BTF blobs into a separate struct.

    NewHandle now takes a Builder instead of a Spec which is a breaking change.
    Other API changes are to functions which have not appeared in a release yet.

btf: unexport NewSpec

    NewSpec was added to be able to create BTF from scratch using Spec.Add. This
    usecase was migrated to Builder and Spec.Add doesn't exist anymore.

    Unexport NewSpec so we can still use it for testing but users only
    experience API breakage once.

btf: generalise datasecWorkaround to Builder.Add

    Extend the datasec validation workaround to all Datasec. This is safe to do
    because there are no methods on Builder which allow going from TypeID to
    Type.

### DIFF
--- a/btf/handle.go
+++ b/btf/handle.go
@@ -37,15 +37,7 @@ func NewHandle(spec *Spec) (*Handle, error) {
 	buf := getBuffer()
 	defer putBuffer(buf)
 
-	var stb *stringTableBuilder
-	if spec.strings != nil {
-		// Use the ELF string table as an estimate of the final
-		// string table size. We don't use the ELF string
-		// table since the types may have been changed in the meantime.
-		stb = newStringTableBuilder(spec.strings.Num())
-	}
-
-	err := marshalTypes(buf, spec.types, stb, kernelMarshalOptions())
+	err := marshalTypes(buf, spec.types, nil, kernelMarshalOptions())
 	if err != nil {
 		return nil, fmt.Errorf("marshal BTF: %w", err)
 	}
--- a/btf/handle.go
+++ b/btf/handle.go
@@ -22,30 +22,25 @@ type Handle struct {
 	needsKernelBase bool
 }
 
-// NewHandle loads BTF into the kernel.
+// NewHandle loads the contents of a [Builder] into the kernel.
 //
-// Returns ErrNotSupported if BTF is not supported.
-func NewHandle(spec *Spec) (*Handle, error) {
-	if spec.byteOrder != nil && spec.byteOrder != internal.NativeEndian {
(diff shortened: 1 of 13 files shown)


btf/handle.go
```diff
@@ -22,30 +22,25 @@ type Handle struct {
...
-	if spec.byteOrder != nil && spec.byteOrder != internal.NativeEndian {
-		return nil, fmt.Errorf("can't load %s BTF on %s", spec.byteOrder, internal.NativeEndian)
-	}
-
-	if spec.firstTypeID != 0 {
-		return nil, fmt.Errorf("split BTF can't be loaded into the kernel")
-	}
+// Returns an error wrapping ErrNotSupported if the kernel doesn't support BTF.
+func NewHandle(b *Builder) (*Handle, error) {
+	small := getBuffer()
```
lmb, line 29: `getByteSlice`

btf/marshal.go
```diff
@@ -36,89 +39,137 @@ type encoder struct {
...
-// that into the kernel and we don't have a use case for writing BTF
-// out again.
+// Adding the identical Type multiple times is valid and will return the same ID.
 //
-// w should be retrieved from bufferPool. opts may be nil.
-func marshalTypes(w *bytes.Buffer, types []Type, stb *stringTableBuilder, opts *marshalOptions) error {
-	if len(types) < 1 {
-		return errors.New("types must contain at least Void")
+// See [Type] for details on identity.
+func (m *Builder) Add(typ Type) (TypeID, error) {
```
lmb, line 98: `b`

rgo3 (approved):
Only nits, human linter goes brrr. Otherwise LGTM.

PR #1071, opened 2023-06-22
improv: make errors more conclusive

### LINKED ISSUE #891: `haveXYZ` feature probes are inconclusive
A Cilium contributor (@jrajahalme) tried running Cilium's BPF unit tests on an aarch64 kernel and received the following error:
```
--- FAIL: TestBPF (0.00s)
    bpf_test.go:149: new coll: map test_cilium_ipcache: map create: prealloc maps not supported (requires >= v4.6)
...
$ uname -a
Linux dev 5.4.0-124-generic #140-Ubuntu SMP Thu Aug 4 02:27:01 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

This failure is not expected, but most `haveXYZ()` probes like `haveNoPreallocMaps()` simply do `if err != nil { return internal.ErrNotSupported }`, which masks any unexpected conditions like ENOMEM or EPERM.

While this may partly be by design, I suggest we make these probes as conclusive as possible (e.g. only EINVAL -> ErrNotSupported) and return any other unexpected errors verbatim to ease debugging.

kwakubiney:
Fixes #891

### DIFF
--- a/link/syscalls.go
+++ b/link/syscalls.go
@@ -60,9 +60,11 @@ var haveProgAttachReplace = internal.NewFeatureTest("BPF_PROG_ATTACH atomic repl
 			asm.Return(),
 		},
 	})
+
 	if err != nil {
 		return internal.ErrNotSupported
 	}
+
 	defer prog.Close()
 
 	// We know that we have BPF_PROG_ATTACH since we can load CGroupSKB programs.
@@ -113,11 +115,12 @@ var haveProgQuery = internal.NewFeatureTest("BPF_PROG_QUERY", "4.15", func() err
 	}
 
 	err := sys.ProgQuery(&attr)
-	if errors.Is(err, unix.EINVAL) {
-		return internal.ErrNotSupported
-	}
+
 	if errors.Is(err, unix.EBADF) {
 		return nil
 	}
-	return err
+	if err != nil {
+		return ErrNotSupported
+	}
+	return errors.New("syscall succeeded unexpectedly")
 })
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -18,8 +18,11 @@ var (
 	uprobeRefCtrOffsetShift = 32
 	haveRefCtrOffsetPMU     = internal.NewFeatureTest("RefCtrOffsetPMU", "4.20", func() error {
 		_, err := os.Stat(uprobeRefCtrOffsetPMUPath)
-		if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
 			return internal.ErrNotSupported
+		}
+		if err != nil {
+			return err
 		}
 		return nil
 	})
--- a/map.go
+++ b/map.go
@@ -385,27 +385,6 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 		}
 	}
 
-	if spec.Flags&(unix.BPF_F_RDONLY_PROG|unix.BPF_F_WRONLY_PROG) > 0 || spec.Freeze {
-		if err := haveMapMutabilityModifiers(); err != nil {
-			return nil, fmt.Errorf("map create: %w", err)
-		}
-	}
-	if spec.Flags&unix.BPF_F_MMAPABLE > 0 {
-		if err := haveMmapableMaps(); err != nil {
-			return nil, fmt.Errorf("map create: %w", err)
-		}
-	}
-	if spec.Flags&unix.BPF_F_INNER_MAP > 0 {
-		if err := haveInnerMaps(); err != nil {
-			return nil, fmt.Errorf("map create: %w", err)
-		}
-	}
-	if spec.Flags&unix.BPF_F_NO_PREALLOC > 0 {
-		if err := haveNoPreallocMaps(); err != nil {
-			return nil, fmt.Errorf("map create: %w", err)
-		}
-	}
-
 	attr := sys.MapCreateAttr{
 		MapType:    sys.MapType(spec.Type),
 		KeySize:    spec.KeySize,
@@ -457,9 +436,35 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 		if errors.Is(err, unix.EINVAL) && spec.Type == UnspecifiedMap {
 			return nil, fmt.Errorf("map create: cannot use type %s", UnspecifiedMap)
 		}
+
+		if spec.Flags&(unix.BPF_F_RDONLY_PROG|unix.BPF_F_WRONLY_PROG) > 0 || spec.Freeze {
+			if haveFeatErr := haveMapMutabilityModifiers(); haveFeatErr != nil {
+				return nil, fmt.Errorf("map create: %w", haveFeatErr)
+			}
+		}
+
+		if spec.Flags&unix.BPF_F_MMAPABLE > 0 {
+			if haveFeatErr := haveMmapableMaps(); haveFeatErr != nil {
+				return nil, fmt.Errorf("map create: %w", haveFeatErr)
+			}
+		}
+
+		if spec.Flags&unix.BPF_F_INNER_MAP > 0 {
+			if haveFeatErr := haveInnerMaps(); haveFeatErr != nil {
+				return nil, fmt.Errorf("map create: %w", haveFeatErr)
+			}
+		}
+
+		if spec.Flags&unix.BPF_F_NO_PREALLOC > 0 {
+			if haveFeatErr := haveNoPreallocMaps(); haveFeatErr != nil {
+				return nil, fmt.Errorf("map create: %w", haveFeatErr)
+			}
+		}
+
 		if attr.BtfFd == 0 {
 			return nil, fmt.Errorf("map create: %w (without BTF k/v)", err)
 		}
+
 		return nil, fmt.Errorf("map create: %w", err)
 	}
 	defer closeOnError(fd)
@@ -991,9 +996,6 @@ func (m *Map) batchLookup(cmd sys.Cmd, startKey, nextKeyOut, keysOut, valuesOut 
 // "keys" and "values" must be of type slice, a pointer
 // to a slice or buffer will not work.
 func (m *Map) BatchUpdate(keys, values interface{}, opts *BatchOptions) (int, error) {
-	if err := haveBatchAPI(); err != nil {
-		return 0, err
-	}
 	if m.typ.hasPerCPUValue() {
 		return 0, ErrNotSupported
 	}
@@ -1035,6 +1037,9 @@ func (m *Map) BatchUpdate(keys, values interface{}, opts *BatchOptions) (int, er
 
 	err = sys.MapUpdateBatch(&attr)
 	if err != nil {
+		if haveFeatErr := haveBatchAPI(); haveFeatErr != nil {
+			return 0, haveFeatErr
+		}
 		return int(attr.Count), fmt.Errorf("batch update: %w", wrapMapError(err))
 	}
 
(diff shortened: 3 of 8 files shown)


link/syscalls.go
```diff
@@ -61,7 +61,10 @@ var haveProgAttachReplace = internal.NewFeatureTest("BPF_PROG_ATTACH atomic repl
 		},
 	})
 	if err != nil {
-		return internal.ErrNotSupported
+		if errors.Is(err, unix.EINVAL) {
```
lmb, line 60: Nit: `errors.Is` does checking for `nil`, so the following is equivalent and has less drift to the right:

```
if errors.Is(...) {
}
if err != nil {
}
```

In https://github.com/cilium/ebpf/issues/891#issuecomment-1535891636 I wrote:

> Note that errors.Is(err, EINVAL) is basically the same as checking err != nil, since EINVAL is such a common error that it's hard to infer anything.

Do you think that explicitly checking EINVAL here is meaningful?

link/uprobe.go
```diff
@@ -19,7 +20,10 @@ var (
 	haveRefCtrOffsetPMU     = internal.NewFeatureTest("RefCtrOffsetPMU", "4.20", func() error {
 		_, err := os.Stat(uprobeRefCtrOffsetPMUPath)
 		if err != nil {
-			return internal.ErrNotSupported
+			if errors.Is(err, unix.EINVAL) {
```
lmb, line 23: I'm a bit surprised this works to be honest. From `man 2 stat`:

```
ERRORS
       EACCES Search permission is denied for one of the directories in the path prefix of pathname.  (See also path_resolution(7).)

       EBADF  fd is not a valid open file descriptor.

       EBADF  (fstatat()) pathname is relative but dirfd is neither AT_FDCWD nor a valid file descriptor.

       EFAULT Bad address.

       EINVAL (fstatat()) Invalid flag specified in flags.

       ELOOP  Too many symbolic links encountered while traversing the path.

       ENAMETOOLONG
              pathname is too long.

       ENOENT A component of pathname does not exist or is a dangling symbolic link.

       ENOENT pathname is an empty string and AT_EMPTY_PATH was not specified in flags.

       ENOMEM Out of memory (i.e., kernel memory).

       ENOTDIR
              A component of the path prefix of pathname is not a directory.

       ENOTDIR
              (fstatat()) pathname is relative and dirfd is a file descriptor referring to a file other than a directory.

       EOVERFLOW
              pathname or fd refers to a file whose size, inode number, or number of blocks cannot be represented in, respectively, the types off_t, ino_t, or blkcnt_t.  This error can occur when, for  exam‐
              ple, an application compiled on a 32-bit platform without -D_FILE_OFFSET_BITS=64 calls stat() on a file whose size exceeds (1<<31)-1 bytes.
```

From this, I'd expect this to return ENOENT, for which the go equivalent is `os.ErrNotExist`. Shouldn't this be:

```
if errors.Is(err, os.ErrNotExist) { return ErrNotSupported }
```

?

syscalls.go
```diff
@@ -88,7 +88,10 @@ var haveMapMutabilityModifiers = internal.NewFeatureTest("read- and write-only m
 		MapFlags:   unix.BPF_F_RDONLY_PROG,
 	})
 	if err != nil {
-		return internal.ErrNotSupported
+		if errors.Is(err, unix.EINVAL) {
```
lmb, line 91: Same comment as elsewhere, is EINVAL more meaningful here?

syscalls.go
```diff
@@ -104,7 +107,10 @@ var haveMmapableMaps = internal.NewFeatureTest("mmapable maps", "5.5", func() er
 		MapFlags:   unix.BPF_F_MMAPABLE,
 	})
 	if err != nil {
-		return internal.ErrNotSupported
+		if errors.Is(err, unix.EINVAL) {
```
lmb, line 110: Ditto

link/syscalls.go
```diff
@@ -47,10 +47,6 @@ var haveProgAttach = internal.NewFeatureTest("BPF_PROG_ATTACH", "4.10", func() e
 })
 
 var haveProgAttachReplace = internal.NewFeatureTest("BPF_PROG_ATTACH atomic replacement of MULTI progs", "5.5", func() error {
-	if err := haveProgAttach(); err != nil {
```
lmb, line 50: Ah, this is an interesting case. Here we have to call `haveProgAttach` first since it establishes a pre-condition for the feature test to be successful. If I remember correctly it's so that we can reliably get EBADF from the syscall below.

So I think this needs to stay where it is.

link/syscalls.go
```diff
@@ -80,11 +76,19 @@ var haveProgAttachReplace = internal.NewFeatureTest("BPF_PROG_ATTACH atomic repl
...
 
 	err = sys.ProgAttach(&attr)
-	if errors.Is(err, unix.EINVAL) {
-		return internal.ErrNotSupported
-	}
-	if errors.Is(err, unix.EBADF) {
-		return nil
+	if err != nil {
+
+		if errors.Is(err, unix.EINVAL) {
```
lmb, line 81: Same nit as before, `errors.Is` doesn't have to be guarded by `if err != nil`.

map.go
```diff
@@ -460,6 +439,31 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 		if attr.BtfFd == 0 {
 			return nil, fmt.Errorf("map create: %w (without BTF k/v)", err)
 		}
+
+		if spec.Flags&(unix.BPF_F_RDONLY_PROG|unix.BPF_F_WRONLY_PROG) > 0 || spec.Freeze {
```
lmb, line 443: I think this needs to move before `if attr.BtfFd == 0`: otherwise we never execute these conditions if a map is created without BTF.

lmb (changes requested):
Looks nice, thanks for your PR! Left some comments, nothing major.

syscalls.go
kwakubiney: I believe we could just return the `err` too if `err != nil`

link/syscalls.go
kwakubiney: Not really, I believe it does not necessarily carry any extra value but I believe it's the closest error to return if that syscall was not supported by the kernel? See: [here](https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L3521)

map.go
kwakubiney: Good catch

kwakubiney: But then again, even if it comes above it ,I reckon eventually it will still enter the `if attr.BtfFd == 0 {}` block. Is there a reason why we'd prefer to throw the error at `if spec.Flags&(unix.BPF_F_RDONLY_PROG|unix.BPF_F_WRONLY_PROG) > 0 || spec.Freeze` instead?

link/syscalls.go
kwakubiney: I might have to dig into the why myself but why is it okay if it returns `EBADF`?

link/uprobe.go
lmb: > Where did you infer that it works though btw, a bit confused by that statement.

I meant: CI doesn't bail out. The reason is that on CI the path always exists, I think. So we never enter this condition.

link/syscalls.go
lmb: > Not really, I believe it does not necessarily carry any extra value

Yeah, makes sense. Maybe we stick to the current behaviour then? It means less churn.

> See: [here](https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L3521)

Nit: the EINVAL is returned here: https://elixir.bootlin.com/linux/v6.4.1/source/kernel/bpf/syscall.c#L5167

syscalls.go
lmb: As per above, let's not change this.

link/syscalls.go
kwakubiney: Ah. Got it

link/syscalls.go
kwakubiney: ^^ Agreed that it's more noisy, & adds no extra value so will change that to it's previous state as mentioned above.

map.go
kwakubiney: Not sure if that's what you meant but decided to move the `btf` check as the fallback error check because it could mask potentially more prevalent errors which arise from the feature tests.

lmb:
There is a real lint error: https://ebpf.semaphoreci.com/jobs/fd7469d0-406a-436d-ad91-19045ab94e1d#L738

lmb:
And real test errors: https://ebpf.semaphoreci.com/jobs/84c72b32-bb05-4d61-b48d-c7870fbd4ab4#L649

syscalls.go
```diff
@@ -87,9 +87,11 @@ var haveMapMutabilityModifiers = internal.NewFeatureTest("read- and write-only m
 		MaxEntries: 1,
 		MapFlags:   unix.BPF_F_RDONLY_PROG,
 	})
+
```
lmb, line 90: Nit: gratuitous whitespace change. It's fine to change whitespace to your preference when you're editing a function anyways, but this just increases the diff.

syscalls.go
```diff
@@ -119,8 +121,12 @@ var haveInnerMaps = internal.NewFeatureTest("inner maps", "5.10", func() error {
 		MaxEntries: 1,
 		MapFlags:   unix.BPF_F_INNER_MAP,
 	})
+
 	if err != nil {
-		return internal.ErrNotSupported
+		if errors.Is(err, unix.EINVAL) {
```
lmb, line 126: Ditto, do you want to keep EINVAL here after all?

syscalls.go
```diff
@@ -135,6 +141,7 @@ var haveNoPreallocMaps = internal.NewFeatureTest("prealloc maps", "4.6", func()
 		MaxEntries: 1,
 		MapFlags:   unix.BPF_F_NO_PREALLOC,
 	})
+
```
lmb, line 144: etc.

lmb (changes requested):

link/syscalls.go
```diff
@@ -60,9 +60,11 @@ var haveProgAttachReplace = internal.NewFeatureTest("BPF_PROG_ATTACH atomic repl
 			asm.Return(),
 		},
 	})
-	if err != nil {
+
+	if errors.Is(err, unix.EINVAL) {
```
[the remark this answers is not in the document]
kwakubiney: Snap, missed it when I was making the changes.

syscalls.go
kwakubiney: Fixed

link/syscalls.go
```diff
@@ -113,9 +115,7 @@ var haveProgQuery = internal.NewFeatureTest("BPF_PROG_QUERY", "4.15", func() err
 	}
 
 	err := sys.ProgQuery(&attr)
-	if errors.Is(err, unix.EINVAL) {
-		return internal.ErrNotSupported
```
lmb, line 117: This test never returns ErrNotSupported now? I guess it's technicall valid to say "I can only detect presence of a feature, not absence" but is that so useful really?

This probe is a bit weird since the "feature present" case is also an error. So the expected matrix of outcomes is:

* EBADF: syscall is supported
* nil: something weird has happened which we don't understand. This should be an error that isn't `ErrNotSupported`.
* other errors (currently EINVAL): syscall is probably not supported. This could be `ErrNotSupported`.

If you want to get rid of EINVAL here I'd do the following:

```
if errors.Is(EBADF) { return nil }
if err != nil { return ErrNotSupported }
return errors.New("syscall succeeded unexpectedly")
```

lmb (commented):
Thanks for the update and sorry for the long delay. We currently have planning seesisons going on, and last week was really busy with trying to get the release out.

One final question, then we're good to go.

link/syscalls.go
kwakubiney: I like this approach because it covers all cases. +1 on returning an error for the `nil` case

lmb (approved):
Thanks! Can you squash the commits please?

PR #1074, opened 2023-06-24
cmd/bpf2go: allow setting output dir

loshz:
## Description
Allow setting an output directory for the `bpf2go` command.

### Usage
An optional `-output` flag has been added. If not specified, then default to the current behavior of using `os.Getwd()`.

I have tested this locally on my machine using the following structure and a go.mod replace directive to point to my local fork:
```bash
$ tree
.
├── bpf
│   └── foo.bpf.c
├── build.go
├── go.mod
├── go.sum
└── pkg
    └── bpf

$ cat build.go
package build

//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -cflags "-g -O2 -Wall -Werror" -output ./pkg/bpf foo ./bpf/foo.bpf.c

$ go generate
Compiled pkg/bpf/foo_bpfel.o
Stripped pkg/bpf/foo_bpfel.o
Wrote pkg/bpf/foo_bpfel.go
Compiled pkg/bpf/foo_bpfeb.o
Stripped pkg/bpf/foo_bpfeb.o
Wrote pkg/bpf/foo_bpfeb.go
```

We could also check that the specified output dir exists, but this will be caught be `clang` so I decided to leave it out. Can add a check if necessary.

Happy to close or keep a fork if this isn't something that's deemed needed/useful.

### DIFF
--- a/cmd/bpf2go/compile_test.go
+++ b/cmd/bpf2go/compile_test.go
@@ -151,11 +151,7 @@ nothing:
 func mustWriteTempFile(t *testing.T, name, contents string) string {
 	t.Helper()
 
-	tmp, err := os.MkdirTemp("", "bpf2go")
-	if err != nil {
-		t.Fatal(err)
-	}
-	t.Cleanup(func() { os.RemoveAll(tmp) })
+	tmp := t.TempDir()
 
 	tmpFile := filepath.Join(tmp, name)
 	if err := os.WriteFile(tmpFile, []byte(contents), 0660); err != nil {
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -71,8 +71,8 @@ var targetByGoArch = map[string]target{
 	"sparc64":     {"bpfeb", "sparc"},
 }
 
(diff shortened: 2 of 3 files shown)


loshz:
The kernel 5.4 tests seem to have timed out, but I can't see the output to check if it's a transient issue or a bug in this PR.

lmb:
5.4 is a problem with our CI.

Can you explain your use case a bit more? Looks like it allows you to not put build.go next to the .c, but that seems like a small win?

loshz:
> 5.4 is a problem with our CI.

ACK.

 
> Can you explain your use case a bit more? Looks like it allows you to not put build.go next to the .c, but that seems like a small win?

It's literally just that. I thought it might be useful from a project structure perspective to allow the generated Go output to be redirected to a dir of choice. Definitely a personal opinion, but I prefer generated code to sit in a separate dir and not alongside the corresponding bpf c code.

Definitely not a critical enhancement, so I'm happy to close this if it doesn't make sense.

loshz:
I could definitely just put a `build.go` file in in the desired directory and run `go generate /path/to/build.go`, but I thought an `-output` flag _might_ be a better UX.

Issue #1073, opened 2023-06-23
Need to make socket_filter.go platform agnostic


binojmr:
**Describe the bug**
There is a reference to "github.com/cilium/ebpf/link" from project https://github.com/l3af-project/l3afd/pull/252 which is expected to run on both linux and windows. `socket_filter.go` is referencing to linux specific socket variables like `unix.SOL_SOCKET`, `unix.SO_ATTACH_BPF`, etc and causing it to fail in windows.

It is blocking the build here https://github.com/l3af-project/l3afd/actions/runs/5288892869

**To Reproduce**
Run below steps to repro the issue.

`PS E:\git> git clone https://github.com/cilium/ebpf`

`Cloning into 'ebpf'... remote: Enumerating objects: 9270, done. remote: Counting objects: 100% (2738/2738), done. remote: Compressing objects: 100% (520/520), done. remote: Total 9270 (delta 2294), reused 2275 (delta 2211), pack-reused 6532 Receiving objects: 100% (9270/9270), 17.70 MiB | 27.92 MiB/s, done. Resolving deltas: 100% (6622/6622), done. PS E:\git> go vet -tags WINDOWS ./... pattern ./...: directory prefix . does not contain main module or its selected dependencies`

`PS E:\git> cd ebpf`

`PS E:\git\ebpf> go vet -tags WINDOWS ./...`
`go: downloading github.com/frankban/quicktest v1.14.5`
`go: downloading golang.org/x/exp v0.0.0-20230224173230-c95f2b4c22f2`
`# github.com/cilium/ebpf/internal/testutils`
`internal\testutils\cgroup.go:58:21: undefined: unix.Stat_t`
`internal\testutils\cgroup.go:59:14: undefined: unix.Fstat`
`# github.com/cilium/ebpf/internal/testutils`
`vet.exe: internal\testutils\cgroup.go:58:21: undefined: unix.Stat_t`
`# github.com/cilium/ebpf/link`
`link\socket_filter.go:18:34: cannot use int(fd) (value of type int) as syscall.Handle value in argument to syscall.SetsockoptInt`
`link\socket_filter.go:34:34: cannot use int(fd) (value of type int) as syscall.Handle value in argument to syscall.SetsockoptInt`

**Expected behavior**
It should be platform agnostic. 
One option could be to add the tag `//go:build !WINDOWS` in file `socket_filter.go` so that it will be skipped for windows. Another option would be to modify the code to handle both platforms. Please suggest or help fix this issue.

[...]

PR #1075, opened 2023-06-24
Keep code multi OS aware

florianl:
Adapt code to be able to run on multiple OS.

As described in https://github.com/cilium/ebpf/issues/1073 currently there is an issue with non Linux support:

```bash
$ GOOS=windows go build ./...
# github.com/cilium/ebpf/internal/testutils
internal/testutils/cgroup.go:58:21: undefined: unix.Stat_t
internal/testutils/cgroup.go:59:14: undefined: unix.Fstat
# github.com/cilium/ebpf/link
link/socket_filter.go:18:34: cannot use int(fd) (value of type int) as syscall.Handle value in argument to syscall.SetsockoptInt
link/socket_filter.go:34:34: cannot use int(fd) (value of type int) as syscall.Handle value in argument to syscall.SetsockoptInt
```

This PR moves Linux specific code into `/internal/unix`.

Fixes https://github.com/cilium/ebpf/issues/1073

### DIFF
--- a/internal/testutils/cgroup.go
+++ b/internal/testutils/cgroup.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 
 	"github.com/cilium/ebpf/internal"
-	"golang.org/x/sys/unix"
+	"github.com/cilium/ebpf/internal/unix"
 )
 
 var cgroup2Path = internal.Memoize(func() (string, error) {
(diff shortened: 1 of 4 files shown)


lmb (approved):
Thanks for the PR!

PR #1077, opened 2023-06-26
Benchmark batch lookups vs. MapIterator, remove MapIterator.prevKey

ti-mo:
Since `MapIterator` is an interesting target for optimization through the upcoming zero-copy map ops work, add a benchmark and simplify its internal representation a bit.

```
Benchmark results at time of writing:

BenchmarkIterate/MapIterator-16         	             866	   1457681 ns/op	   64126 B/op	    4004 allocs/op
BenchmarkIterate/MapIteratorDelete-16   	             408	   2526392 ns/op	   80210 B/op	    6004 allocs/op
BenchmarkIterate/BatchLookup-16         	           59647	     19974 ns/op	   32851 B/op	       8 allocs/op
BenchmarkIterate/BatchLookupAndDelete-16         	   10000	    154757 ns/op	   32872 B/op	       8 allocs/op
BenchmarkIterate/BatchDelete-16                  	    8306	    136770 ns/op	   16416 B/op	       3 allocs/op
```

```
Change the terminology used in the iterator to revolve around the 'current'
and the 'next' key instead of considering the 'previous' key.

I found the behaviour after passing a nil `mi.prevKey` to NextKeyBytes a little
subtle. If the interface value passed to NextKeyBytes contains a nil 'any', it
will successfully detect nilness and not try to marshal the key. If the caller
passes a []byte(nil), the nil check will fail and nextKey() will try to marshal
the key.

Perform a nil check on mi.curKey instead and explicitly pass an untyped nil to
NextKeyBytes if the Map's first key needs to be looked up. This means prevKey
can be eliminated.
```

### DIFF
--- a/map.go
+++ b/map.go
@@ -1362,8 +1362,7 @@ func marshalMap(m *Map, length int) ([]byte, error) {
 // See Map.Iterate.
 type MapIterator struct {
 	target            *Map
-	prevKey           interface{}
-	prevBytes         []byte
+	curKey            []byte
 	count, maxEntries uint32
 	done              bool
 	err               error
@@ -1373,7 +1372,6 @@ func newMapIterator(target *Map) *MapIterator {
 	return &MapIterator{
 		target:     target,
 		maxEntries: target.maxEntries,
-		prevBytes:  make([]byte, target.keySize),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


map_test.go
```diff
@@ -1974,6 +1974,78 @@ func BenchmarkMap(b *testing.B) {
...
+		for i := 0; i < b.N; i++ {
+			var count uint64
+			iter := m.Iterate()
+			for iter.Next(&k, &v) {
+				count++
+			}
+			if err := iter.Err(); err != nil {
+				b.Fatal(err)
+			}
+			if count != uint64(m.MaxEntries()) {
```
lmb, line 2020: How about skipping this? Benchmarky code doesn't really benefit from such error checking imo.

lmb (approved):

map.go
```diff
@@ -1395,26 +1393,35 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
-		var nextBytes []byte
-		nextBytes, mi.err = mi.target.NextKeyBytes(mi.prevKey)
+		var nextKey []byte
+		if mi.curKey == nil {
+			// Pass nil interface to NextKeyBytes to make sure the Map's first key
+			// is returned. If we pass an uninitialized []byte instead, it'll see a
+			// non-nil interface and try to marshal it.
+			nextKey, mi.err = mi.target.NextKeyBytes(nil)
+
+			mi.curKey = make([]byte, mi.target.keySize)
```
lmb, line 1403: Maybe a bit weird: for an empty map we get `mi.curKey != nil` due to this. Not sure if it's worth fixing though.

lmb (approved):

map_test.go
ti-mo: Yeah, fair point. It was only useful while writing it.

map.go
ti-mo: Thanks for pointing it out, I think that's okay for now. No other methods depend on this state.

Issue #1068, opened 2023-06-19
btf: make loading kernel BTF cheaper


lmb:
Loading `vmlinux` BTF is currently very expensive:

```
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf/btf
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
                │      -       │
                │    sec/op    │
ParseVmlinux-16   50.47m ± ∞ ¹
¹ need >= 6 samples for confidence interval at level 0.95

                │       -       │
                │     B/op      │
ParseVmlinux-16   38.46Mi ± ∞ ¹
¹ need >= 6 samples for confidence interval at level 0.95
```

So ~50ms and 40MiB at current master, and this is after doing a bunch of optimization work. For this reason we cache vmlinux BTF in a global variable and try to reuse the same parsed Spec as much as possible. Otherwise features requiring vmlinux BTF like kconfig or ksyms are unbearably slow. I've made several attempts at improving the situation:

* https://github.com/cilium/ebpf/pull/948
* https://github.com/cilium/ebpf/pull/1015
* https://github.com/cilium/ebpf/pull/1001
* https://github.com/cilium/ebpf/commit/97e8a9fa79a6ebbe7ae99630c97ab807b4cb0bf0
* https://github.com/cilium/ebpf/commit/5b305b240a5aeea9d473631ed50b1c4c5d3af76d

Parsing is expensive, but the parsed representation is also pretty large. Go doesn't have great tools for this, but an educated guess is that parsed vmlinux takes about 25MiB of RSS. Users have of course noticed this, and we've added an escape hatch to flush the cached BTF by calling `linux.FlushCaches`. Even that causes problems and confusion however, as shown by https://github.com/cilium/ebpf/issues/1063. Tetragon was also impacted and had to chase down copies of vmlinux.

I've come to the conclusion that we're using the wrong approach. Instead of parsing the whole BTF upfront we should only "inflate" types that are needed by the user via `Spec.TypeByID()`, `Spec.TypeByName()` and so on. BTF wire format is already optimized for size, so we should refer to it as much as possible. @ti-mo suggested this a long time ago, but for reasons I can't remember I wasn't a fan. I think it's now time do to this, but getting there is going to take several refactors.

# Split writing from reading BTF

[...]

PR #1087, opened 2023-07-04
Revert "btf: avoid copies of vmlinux BTF for internal callers"

lmb:
This reverts commit c3a89fcd62038af7f958e3cea60ff4e5b9fac35d. I've come to the conclusion that a separate package isn't the right approach. Parsing vmlinux is still too expensive and will stay this way as long as we unmarshal the entire type graph when parsing BTF.

I think that the real solution will involve breaking API once more, probably in the next release. I'm backing out the change to avoid this churn. The downside is that kfunc, CO-RE and attaching programs like fentry get slower again. Since they were already slow in the last release I'm hoping that this doesn't cause too many problems.

I've sketched out what I think the future should hold in #1068.


rgo3 (approved):

brb (approved):
As a `pwru` maintainer, I'm happy about the comeback of `LoadKernelSpec()`.

ti-mo:
Makes sense to keep it this way for now, let's see what kind of gains we can achieve using some of the other approaches. :+1:

PR #1089, opened 2023-07-04
features: fix unsupported progtype/helper probes

rgo3:
features: add missing HaveProgramHelper test cases:
```
Add the missing test cases now that CI and the library support these
helpers.

Unfortunately the diff of this commit will touch all program/helper test
cases because of go formatting, while it only adds the four cases that
were pointed out in the removed comment.

Signed-off-by: Robin Gögge <r.goegge@isovalent.com>
```
features: fix unsupported progtype/helper probes:
```
By enabling feature probes for the program types Tracing, Extension, LSM
and StructOps within the HaveProgramType API, we overlooked that this
would cause false negative results when using these types with the
HaveProgramHelper API. To probe for these program types we craft
specialized ProgramSpecs, but don't do the same when probing for these
types in combination with an arbitrary helper. For now we should return
an inconclusive error from the HaveProgramHelper API until we find a
better way to probe for helpers with these program types.

Fixes: #1082

Signed-off-by: Robin Gögge <r.goegge@isovalent.com>
```

### DIFF
--- a/features/prog.go
+++ b/features/prog.go
@@ -234,6 +234,10 @@ func HaveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
 }
 
 func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
+	if ok := helperProbeNotImplemented(pt); ok {
+		return fmt.Errorf("no feature probe for %v/%v", pt, helper)
+	}
+
 	if err := HaveProgramType(pt); err != nil {
 		return err
 	}
@@ -282,4 +286,12 @@ func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
 	}
 
 	return err
+}
+
+func helperProbeNotImplemented(pt ebpf.ProgramType) bool {
+	switch pt {
+	case ebpf.Extension, ebpf.LSM, ebpf.StructOps, ebpf.Tracing:
+		return true
+	}
+	return false
 }
--- a/features/prog_test.go
+++ b/features/prog_test.go
@@ -80,3 +80,15 @@ func TestHaveProgramHelper(t *testing.T) {
 
 	}
 }
+
+func TestHelperProbeNotImplemented(t *testing.T) {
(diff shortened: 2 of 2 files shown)


features/prog.go
```diff
@@ -234,6 +234,10 @@ func HaveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
 }
 
 func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
+	if ok := helperProbeNotImplemented(pt); ok {
```
lmb, line 237: Does it make sense to add a test?

lmb (commented):
There are some real errors: https://ebpf.semaphoreci.com/jobs/5e1823a5-6494-44e4-9336-e79bb1e6e309/summary?report_id=79ed1ee4-ab67-3dab-9091-08a3b29b97c2&test_id=10b17f57-1fff-3c91-8396-622f0fbf8e4b

features/prog.go
rgo3: Wasn't sure if this needed one because we essentially would just test if a switch in go works but also felt strange to not add a test given this is a regression. I can add `TestHelperProbeNotImplemented` that calls into `haveProgramHelper` with one test case per prog type that we can't probe for at the moment.

rgo3:
Regarding the legit failures: I'm suspecting some kernel config shenanigans because `FnGetCgroupClassID` depends on `CONFIG_CGROUP_NET_CLASSID` and we don't explicitly set it in the ci kernels config... but i'ts odd that the test passes on 5.19 then :thinking:.

features/prog.go
lmb: Let's do it if it's easy to do. Also serves as a reminder to update the tests if we end up implementing the helpers.

lmb (approved):

lmb:
> but i'ts odd that the test passes on 5.19 then thinking.

Don't know exactly, but my guess is that the setting became a default yes at some point.

lmb (approved):

PR #1091, opened 2023-07-12
run-tests: support new packaging

lmb:
ci-kernels are now packaged slightly differently:
* Kernels are real tar files which also contain bpf_testmod
* Kernels contain the arch in their name
* Selftests don't contain bpf_testmod anymore

Update run-tests.sh to support the new format.

Needs: https://github.com/cilium/ci-kernels/pull/42

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -51,7 +51,7 @@ if [[ "${1:-}" = "--exec-vm" ]]; then
   fi
 
   for ((i = 0; i < 3; i++)); do
-    if ! $sudo virtme-run --kimg "${input}/bzImage" --memory 768M --pwd \
+    if ! $sudo virtme-run --kimg "${input}/boot/vmlinuz" --memory 768M --pwd \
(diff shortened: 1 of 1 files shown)


lmb:
Seems like there is a new BTF bug in the 5.4 kernel build. Since we've had working 5.4 kernels at the same point release in the past I'm going to assume that it has something to do with how we build the kernel.

lmb:
The problem is that I updated pahole from 1.23 to 1.24. The new pahole encodes enum64 which the kernel doesn't understand.

ti-mo:
@lmb Does this need our attention in the short term? I'll unsubscribe since I see it's marked as draft now.

Also, should we make this a Go tool instead of a script at some point?

lmb:
This is good to go now! I'm amazed it worked on GHA on the first try.

lmb:
@rgo3 @ti-mo PTAL. Please don't merge, since I need to merge `ci-kernels` first and drop a commit from this PR.

Issue #1034, opened 2023-05-05
ErrMapIncompatible should contain additional information


ti-mo:
In https://github.com/cilium/cilium/pull/22693, Cilium will start relying exclusively on the library to tell it when a map is incompatible with an existing pinned map, but right now `ErrMapIncompatible` is a little vague. Callers need to open the old map explicitly, inspect its info, and work out the differences with the MapSpec.

---

Updated to echo https://github.com/cilium/ebpf/pull/1093#issuecomment-1640078398.

I think there are 2 issues with the existing code:

- No clear communication (towards the user) of what 'expected' and 'got' means in (`expected type %v, got %v`). We should settle on language that clarifies which side is the existing Map and which is the incoming Spec.
- Only one error is surfaced at a time, being the first-found one.

As a bonus, it would be nice if this would return a new `ebpf.IncompatibleMapError{OldType, NewType ebpf.MapType, ...}` that `.Unwrap()`s to `ErrMapIncompatible` so it can be used like:

```
var merr *ebpf.IncompatibleMapError
if errors.As(err, &merr) {
  if merr.OldType == ebpf.ArrayOfMaps && merr.NewType == ebpf.HashOfMaps {
    // Convert the old map from ArrayOfMaps to HashOfMaps here..
  }
}
```

This would be really useful to trigger up/downgrade logic.

Also, `.Error()` should only return a diff of the map's properties, not a dump of all old/new properties. That's just something I hacked up in Cilium to move ahead with the changes there.

ti-mo:
@swastik959 There's a short document at https://github.com/cilium/ebpf/blob/master/CONTRIBUTING.md. Do you have any specific questions?

sharathsivakumar:
Hello @ti-mo, I would like to work on this if no one is actively working on this.

sharathsivakumar:
Thanks!

lmb:
Closing this for now, since it's actually quite tricky to implement. This should happen if / when we use the info from the cilium side.

PR #1093, opened 2023-07-12
Include map details when returning ErrMapIncompatible

alexandreLamarre:
Addresses #1034 

Opted for a solution that maintains compatibility with existing error checks of the form: 
```go
errors.Is(err, ErrMapIncompatible)
```

### DIFF
--- a/map.go
+++ b/map.go
@@ -219,7 +219,8 @@ type MapKV struct {
 // Compatible returns nil if an existing map may be used instead of creating
 // one from the spec.
 //
-// Returns an error wrapping [ErrMapIncompatible] otherwise.
+// Returns an error wrapping [ErrMapIncompatible] otherwise. The error also wraps [IncompatibleMapErr]
+// which contains useful information about the incompatibility.
 func (ms *MapSpec) Compatible(m *Map) error {
 	ms, err := ms.fixupMagicFields()
 	if err != nil {
@@ -238,22 +239,22 @@ func (ms *MapSpec) Compatible(m *Map) error {
 	}
 	switch {
 	case m.typ != ms.Type:
-		return fmt.Errorf("%w: %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
+		return errors.Join(ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	case m.keySize != ms.KeySize:
-		return fmt.Errorf("%w : %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
+		return errors.Join(ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	case m.valueSize != ms.ValueSize:
-		return fmt.Errorf("%w %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
+		return errors.Join(ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	case m.maxEntries != ms.MaxEntries:
-		return fmt.Errorf("%w : %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
+		return errors.Join(ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	// BPF_F_RDONLY_PROG is set unconditionally for devmaps. Explicitly allow
 	// this mismatch.
 	case !((ms.Type == DevMap || ms.Type == DevMapHash) && m.flags^ms.Flags == unix.BPF_F_RDONLY_PROG) &&
 		m.flags != ms.Flags:
-		return fmt.Errorf("%w : %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
+		return errors.Join(ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 	}
 	return nil
 }
--- a/map.go
+++ b/map.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"time"
 	"unsafe"
 
@@ -26,6 +27,42 @@ var (
 	ErrMapIncompatible  = errors.New("map spec is incompatible with existing map")
 	errMapNoBTFValue    = errors.New("map spec does not contain a BTF Value")
 )
+
+type IncompatibleMapErr struct {
+	Existing *MapInfo
+	Incoming *MapSpec
+}
+
+func NewErrIncompatibleMap(existing *MapInfo, incoming *MapSpec) *IncompatibleMapErr {
+	return &IncompatibleMapErr{
+		Existing: existing,
+		Incoming: incoming,
+	}
+}
+
+func (e *IncompatibleMapErr) Error() string {
+	return fmt.Sprintf("details(existing -> incoming) : %s", e.diff())
+}
+
+func (e *IncompatibleMapErr) diff() string {
+	diffs := []string{}
+	if e.Existing.Type != e.Incoming.Type {
+		diffs = append(diffs, fmt.Sprintf("type : %s -> %s", e.Existing.Type, e.Incoming.Type))
+	}
+	if e.Existing.Flags != e.Incoming.Flags {
+		diffs = append(diffs, fmt.Sprintf("flags : %d -> %d", e.Existing.Flags, e.Incoming.Flags))
+	}
+	if e.Existing.KeySize != e.Incoming.KeySize {
+		diffs = append(diffs, fmt.Sprintf("keySize : %d -> %d", e.Existing.KeySize, e.Incoming.KeySize))
+	}
+	if e.Existing.ValueSize != e.Incoming.ValueSize {
+		diffs = append(diffs, fmt.Sprintf("valueSize : %d -> %d", e.Existing.ValueSize, e.Incoming.ValueSize))
+	}
+	if e.Existing.MaxEntries != e.Incoming.MaxEntries {
+		diffs = append(diffs, fmt.Sprintf("maxEntries : %d -> %d", e.Existing.MaxEntries, e.Incoming.MaxEntries))
+	}
+	return strings.Join(diffs, ", ")
+}
 
 // MapOptions control loading a map into the kernel.
 type MapOptions struct {
@@ -189,24 +226,34 @@ func (ms *MapSpec) Compatible(m *Map) error {
 		return err
 	}
 
+	asInfo := func(m *Map) *MapInfo {
+		return &MapInfo{
+			Type:       m.typ,
+			KeySize:    m.keySize,
+			ValueSize:  m.valueSize,
+			MaxEntries: m.maxEntries,
+			Flags:      m.flags,
+			Name:       m.name,
+		}
+	}
 	switch {
 	case m.typ != ms.Type:
-		return fmt.Errorf("expected type %v, got %v: %w", ms.Type, m.typ, ErrMapIncompatible)
+		return fmt.Errorf("%w: %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	case m.keySize != ms.KeySize:
-		return fmt.Errorf("expected key size %v, got %v: %w", ms.KeySize, m.keySize, ErrMapIncompatible)
+		return fmt.Errorf("%w : %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	case m.valueSize != ms.ValueSize:
-		return fmt.Errorf("expected value size %v, got %v: %w", ms.ValueSize, m.valueSize, ErrMapIncompatible)
+		return fmt.Errorf("%w %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	case m.maxEntries != ms.MaxEntries:
-		return fmt.Errorf("expected max entries %v, got %v: %w", ms.MaxEntries, m.maxEntries, ErrMapIncompatible)
+		return fmt.Errorf("%w : %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 
 	// BPF_F_RDONLY_PROG is set unconditionally for devmaps. Explicitly allow
 	// this mismatch.
 	case !((ms.Type == DevMap || ms.Type == DevMapHash) && m.flags^ms.Flags == unix.BPF_F_RDONLY_PROG) &&
 		m.flags != ms.Flags:
-		return fmt.Errorf("expected flags %v, got %v: %w", ms.Flags, m.flags, ErrMapIncompatible)
+		return fmt.Errorf("%w : %w", ErrMapIncompatible, NewErrIncompatibleMap(asInfo(m), ms))
 	}
 	return nil
 }
--- a/map.go
+++ b/map.go
@@ -28,42 +28,6 @@ var (
 	errMapNoBTFValue    = errors.New("map spec does not contain a BTF Value")
 )
 
-type IncompatibleMapErr struct {
-	Existing *MapInfo
-	Incoming *MapSpec
-}
-
-func NewErrIncompatibleMap(existing *MapInfo, incoming *MapSpec) *IncompatibleMapErr {
-	return &IncompatibleMapErr{
-		Existing: existing,
-		Incoming: incoming,
-	}
-}
-
-func (e *IncompatibleMapErr) Error() string {
-	return fmt.Sprintf("details(existing -> incoming) : %s", e.diff())
-}
-
-func (e *IncompatibleMapErr) diff() string {
-	diffs := []string{}
-	if e.Existing.Type != e.Incoming.Type {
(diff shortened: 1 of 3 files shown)


ti-mo:
Sorry, I didn't really specify in https://github.com/cilium/ebpf/issues/1034 what we should do instead. I was expecting to have some form of discussion about what to do first. I think there are 2 issues with the existing code:

- No clear communication (towards the user) of what 'expected' and 'got' means in (`expected type %v, got %v`). We should settle on language that clarifies which side is the existing Map and which is the incoming Spec.
- Only one error is surfaced at a time, being the first-found one.

As a bonus, it would be nice if this would return a new `ebpf.IncompatibleMapError{OldType, NewType ebpf.MapType, ...}` that `.Unwrap()`s to `ErrMapIncompatible` so it can be used like:

```
var merr *ebpf.IncompatibleMapError
if errors.As(err, &merr) {
  if merr.OldType == ebpf.ArrayOfMaps && merr.NewType == ebpf.HashOfMaps {
    // Convert the old map from ArrayOfMaps to HashOfMaps here..
  }
}
```

This would be really useful to trigger up/downgrade logic.

By default, `.Error()` would only return a diff of the map's properties, not a dump of all old/new properties. That's just something I hacked up in Cilium to move ahead with the changes there.

alexandreLamarre:
I see the pain point in error handling those upgrade/downgrade scenarios, so I cooked up a minimalist implementation based on your feedback

There's an ugly function closure to wrap a `Map` into `MapInfo` since there seems to be no good way to convert a map into something that exports its fields

I considered a couple of alternatives here, which I wasn't too sure about:
- Introducing a new type specifically for this case, but that seemed like overkill (introduces the possibility to use struct tags for generating spec differences)
- Converting the `Map` to a partial `MapSpec`, but that also didn't seem great.
- Using `Map`'s `Info` method, but that could technically introduce new points of failure (Even though the current assumption as far as I could tell is that `Compatible` is called after a map is successfully loaded )

lmb:
@ti-mo PTAL

ti-mo:
@alexandreLamarre Thanks for taking another spin on this! This is close to what I imagined.

I'd initially typed up a bunch of feedback, but I wanted to reshuffle a few things so I thought code would be clearer. I pushed another commit to the PR that we'll squash before merging.

Summary of the changes made:
- introduced a minimal `mapInfo` type with only the fields that matter for the compatibility check
- renamed IncompatibleMapErr to IncompatibleMapError (`Err` is most commonly used as a prefix for sentinels)
- removed the use of `errors.Join` as it results in multiline error strings, which would make things less readable in this case
- folded the diff() code into `MapSpec.Compatible()` as it basically did what Compatible() was already doing
- co-located the assertions on IncompatibleMapError in the test cases itself

I'd also like to get some feedback from @lmb before merging.

ti-mo (approved):

map_test.go
```diff
@@ -1726,69 +1726,68 @@ func TestMapPinningIncompatible(t *testing.T) {
...
-		"ValueSize":  {Name: spec.Name, Type: spec.Type, KeySize: spec.ValueSize, ValueSize: 8, MaxEntries: spec.MaxEntries, Pinning: spec.Pinning},
-		"Type":       {Name: spec.Name, Type: Array, KeySize: spec.ValueSize, ValueSize: spec.ValueSize, MaxEntries: spec.MaxEntries, Pinning: spec.Pinning},
-		"MaxEntries": {Name: spec.Name, Type: spec.Type, KeySize: spec.ValueSize, ValueSize: spec.ValueSize, MaxEntries: 2, Pinning: spec.Pinning},
-		"Flags":      {Name: spec.Name, Type: spec.Type, KeySize: spec.ValueSize, ValueSize: spec.ValueSize, MaxEntries: spec.MaxEntries, Pinning: spec.Pinning, Flags: unix.BPF_F_RDONLY_PROG},
+	tests := []struct {
+		name string
+		spec *MapSpec
+		test func(*qt.C, *IncompatibleMapError)
+	}{
+		{"Type", &MapSpec{Name: "test", Type: Array, KeySize: 4, ValueSize: 4, MaxEntries: 1, Pinning: PinByName},
```
lmb, line 1747: This refactoring makes it harder to figure out which field has changed due to a lot of repetition. The previous code also has repetition but it's simpler to parse.

I'd probably go for a field `func(*MapSpec)` and then set `func(ms *MapSpec) { ms.Type = Array }`. Then it's clear what we are changing.

map.go
```diff
@@ -179,36 +180,81 @@ type MapKV struct {
...
+			Type:       ms.Type,
+			KeySize:    ms.KeySize,
+			ValueSize:  ms.ValueSize,
+			MaxEntries: ms.MaxEntries,
+			Flags:      ms.Flags,
+		},
+		diff: strings.Join(diffs, ", "),
+	}
+
+	return fmt.Errorf("%w: %w", err, ErrMapIncompatible)
```
lmb, line 257: It's not necessary to wrap ErrMapIncompatible here. We can instead implement `IncompatibleMapError.Is`:

```go
func (*IncompatibleMapError) Is(target error) bool { return target == ErrMapIncompatible }
```

map.go
```diff
@@ -28,42 +28,6 @@ var (
 	errMapNoBTFValue    = errors.New("map spec does not contain a BTF Value")
 )
 
-type IncompatibleMapErr struct {
-	Existing *MapInfo
```
lmb: @ti-mo why did you get rid of *MapInfo and *MapSpec? To me they seem like the logical types to use for this.

Maybe `Existing` should even be `*Map` so that user code can move KVs around?

map.go
```diff
@@ -216,47 +180,81 @@ type MapKV struct {
...
-			Type:       m.typ,
-			KeySize:    m.keySize,
-			ValueSize:  m.valueSize,
-			MaxEntries: m.maxEntries,
-			Flags:      m.flags,
-			Name:       m.name,
-		}
+	diffs := []string{}
+	if m.typ != ms.Type {
+		diffs = append(diffs, fmt.Sprintf("Type: %s -> %s", m.typ, ms.Type))
```
lmb, line 216: Nit: use `instead of` instead of `->` and invert the order? The arrow doesn't really tell me more.

map.go
```diff
@@ -216,47 +180,81 @@ type MapKV struct {
...
+	KeySize    uint32
+	ValueSize  uint32
+	MaxEntries uint32
+	Flags      uint32
+}
+
+type IncompatibleMapError struct {
+	Existing mapInfo
+	Incoming mapInfo
+	diff     string
```
lmb, line 194: Good idea to keep the diff separate from the metadata.

map_test.go
```diff
@@ -1726,69 +1726,68 @@ func TestMapPinningIncompatible(t *testing.T) {
...
+			func(c *qt.C, merr *IncompatibleMapError) {
+				c.Assert(merr.Existing.Type, qt.Equals, Hash)
+				c.Assert(merr.Incoming.Type, qt.Equals, Array)
+			}},
```
lmb, line 1751: Nit: somewhat wonky formatting?

lmb (commented):
Left some thoughts. My main question is: how would one use this API to implement map migration as mentioned in the c/c issue? See below on my thoughts, the TL;DR: I think this is a lot more complicated than it seems on the surface.

Here is a naive approach to slot this into the object lifecycle.

1. LoadCollection with PinPath set
2. Get IncompatibleMapError (but only for the first incompatible map)
3. Migrate the map
4. Pin the map (?)
5. Goto 1 until there are no more errors

This has the problem of being really destructive: once we migrate a single map we can't abort the migration since the pin has destroyed the old map. I guess we could work around that by using `MapOptions.MapReplacements` instead, but then we still have the issue that really we want to know _all_ the maps in need of migration at once. Our current API doesn't support that.

When I was thinking about this in the past I always thought that we'd add some sort of a `MigrateMap func(...) error` callback to `MapOptions`. The lib would call into that and any error from the callback would cause loading to fail. Once all maps are loaded we can do clever stuff behind the scenes to make upgrades as safe as possible. The downside of this is that sometimes one might want to split or merge maps, which this schema doesn't support.

map_test.go
```diff
@@ -1720,15 +1720,74 @@ func TestMapPinning(t *testing.T) {
...
 
-	spec.KeySize = 8
-	m3, err := NewMapWithOptions(spec, MapOptions{PinPath: tmp})
-	if err == nil {
-		m3.Close()
-		t.Fatalf("Opening a pinned map with a mismatching spec did not fail")
+	tests := []struct {
+		name string
+		spec *MapSpec
+		test func(*qt.C, *IncompatibleMapError)
```
lmb, line 1745: Why not `testing.T`?
```suggestion
		test func(*testing.T, *IncompatibleMapError)
```

qt.C is kind of exotic and might be removed by the next version of the package.

map.go
ti-mo: Ergonomics, mainly. `*Map` has all of its properties hidden behind accessors, so at the call site this will end up as `if merr.Existing.Type() == merr.Incoming.Type { ... }`. I have a slight preference for a symmetrical comparison.

Your suggestion of `Existing *Map` makes more sense to me than `Existing *MapInfo`. A small downside is that we'll be sending open fds to the caller that will be left for the GC to clean up, guaranteed. Would be pretty convenient for writing migration code, though.

map_test.go
ti-mo: Good idea! I feel like we have many more tests that could take this approach. :)

map_test.go
ti-mo: Sure. You'll notice I was using `c.Run()` since there are calls to `c.Assert` in the subtests. Subtests shouldn't call Fatal or Fail on the parent test scope, that leads to panics:

`testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test`

map.go
ti-mo: Good point. We still need to render the ErrMapIncompatible error string somewhere, though. With this %w removed, the error reads `creating map: use pinned map test: Type: Hash -> Array`. It doesn't actually say what the problem is.

I've included ErrMapIncompatible in `IncompatibleMapError.Error()` with a `%s` verb.

map.go
ti-mo: What about `%v changed to %v`? `instead of` also completely lacks directionality, unlike an arrow indicator (apparently IMO).

ti-mo:
> Left some thoughts. My main question is: how would one use this API to implement map migration as mentioned in the c/c issue? See below on my thoughts, the TL;DR: I think this is a lot more complicated than it seems on the surface.
> 
> Here is a naive approach to slot this into the object lifecycle.
> 
>     1. LoadCollection with PinPath set
> 
>     2. Get IncompatibleMapError (but only for the first incompatible map)
> 
>     3. Migrate the map
> 
>     4. Pin the map (?)
> 
>     5. Goto 1 until there are no more errors
> 
> 

Ok, so a bit of context here. Note that currently, c/c doesn't copy any map contents between maps during agent startup. Under 'migration', we currently understand temporarily moving an existing map pin, creating the new map, and removing the old pin if successful. New maps are always populated from scratch, contents are never copied.

The snippet I pasted in https://github.com/cilium/ebpf/issues/1034 is a potential solution for migrating per-cluster CT maps in c/c. This would be the first map that has its contents copied from an array to a hash, specifically because of the runtime impact that dropping CT maps would cause. This is a bit of a special case because the outer map is static once the agent has started, and we can safely convert it during an upgrade. The upgraded versions of the bpf progs would expect a hash map, and both old and new progs can coexist during the upgrade, since both old and new access the same inner CT maps.

> This has the problem of being really destructive: once we migrate a single map we can't abort the migration since the pin has destroyed the old map. I guess we could work around that by using `MapOptions.MapReplacements` instead, but then we still have the issue that really we want to know _all_ the maps in need of migration at once. Our current API doesn't support that.

There's no one-size-fits-all migration strategy, but temporarily moving a map pin is as close as you're going to get to a rollback path. For migrating prog arrays in particular, you don't want to unpin before all progs referring to it are gone.

> When I was thinking about this in the past I always thought that we'd add some sort of a `MigrateMap func(...) error` callback to `MapOptions`. The lib would call into that and any error from the callback would cause loading to fail. Once all maps are loaded we can do clever stuff behind the scenes to make upgrades as safe as possible. The downside of this is that sometimes one might want to split or merge maps, which this schema doesn't support.

Okay, you're coming at the problem from a specific angle: performing migrations during ELF loading. In c/c's case, doing migrations during loading is dangerous as multiple ELF loads can race, even if they're different ELFs (e.g. a bpf_host racing a bpf_overlay). In Cilium, there are some global maps shared across all ELFs, and we definitely don't want to leave the ordering up to chance, since it means a map might get recreated + populated multiple times and both operations race for creating the pin. Not necessarily a disaster, but one of the operations will fail and have to retry.

Instead, these important central maps are created, populated and pinned by individual subsystems before loading any ELFs. These subsystems are thus responsible for hosting any migration logic, where the API proposed in this PR fits without issue. Your 5-step plan above sort of assumes that all old maps would be unpinned after the ELF is loaded, but in c/c we want some pins to live until after e.g. all host devices had their tc/xdp progs replaced. A rollback might still be necessary after a successful ELF load.

map.go
lmb: > I have a slight preference for a symmetrical comparison.

Me too, but at this point we don't always have symmetrical info. MapSpec tends to have more info than what we can extract from a kernel side map due to missing APIs. I think the best course is to not obfuscate this.

> A small downside is that we'll be sending open fds to the caller that will be left for the GC to clean up, guaranteed.

Hmm good point about fds. We've worked so hard to get rid of the leaks xD. Maybe expose a function that allows loading the pinned map? Or a way to achieve something similar?

ti-mo:
Things have slightly shifted in the past few weeks. c/c is going to move away from having multiple copies of a map pinned to bpffs in favor of a single pin created by a 'commit' step after loading an ELF. This step would overwrite existing pins if all the ELF's programs have been attached.

> * Figure out how we can signal that multiple maps are incompatible at once. Maybe we can just `errors.Join` a bunch of MapIncompatibleErrors?

Yep, we can return this kind of error from the ELF loader. I don't have the use case currently, but it can be done in a follow-up if/when we see the need. Getting an `IncompatibleMapError` out of a joined error seems a little more involved, though. [The documentation on `errors.Join()` states](https://pkg.go.dev/errors#Join):

> A non-nil error returned by Join implements the Unwrap() []error method.

and 

> Unwrap only calls a method of the form "Unwrap() error". In particular Unwrap does not unwrap errors returned by [Join](https://pkg.go.dev/errors#Join).

So, currently, the caller has to do this: `uerrs := errs.(interface{ Unwrap() []error }).Unwrap()` to get an `[]error` out of an errors.Join(). :/ Fortunately, there won't be any nils in the resulting slice, but the caller still has to call `errors.As()` in a loop.

> * Figure out how to deal with the Fd leak.

We can't handwave this problem away, unfortunately. The only way to not leak them is to not return `Map`s.

If you're opposed to `struct mapInfo` and/or `IncompatibleMapError`, we can drop it for now and go with the improved diff output for now.

lmb:
> If you're opposed to struct mapInfo and/or IncompatibleMapError, we can drop it for now and go with the improved diff output for now.

Not opposed, it just has sharp corners we haven't figured out. Let's get the diff in and figure out error signaling when we need it. Maybe by that time there is some consensus what to do with multiple wrappers errors.

map.go
ti-mo: I'm warming up to the idea of returning `struct IncompatibleMapError { Incoming *MapSpec, Existing *Map}` and `errors.Join()`ing them during ELF load. It's not expected to trigger this error very often, so the impact of leaking *Maps is relatively limited. This basically implements your `MigrateMap func(...) error` idea, since it exposes a mutable reference to the offending MapSpec and allows manipulating the existing Map as well.

This would also play nicely with a potential feature proposal I have for overriding bpffs pin paths for specific maps:

```
CollectionOptions{
	Maps: MapOptions{PinPath: "/sys/fs/bpf"},
	MapPinPaths: map[string]string{
		"cilium_calls": "/sys/fs/bpf/some/nested/dir/cilium_calls",
	},
}
```

This way, the caller just receives a *Map from an `IncompatibleMapError` and doesn't need to care about where the pin was on the file system.

ti-mo:
@alexandreLamarre Thanks for working on this! We ended up merging only the changes to the `MapSpec.Compatible()` error string. I've kept a squashed + rebased version on a branch in my fork: https://github.com/ti-mo/ebpf/tree/tb/incompatible-map-error.

After discussing offline with @lmb (sorry, should've been in a public channel, it just kind of happened), we've concluded a few things:
1. Returning an IncompatibleMapError with two MapInfo structs is quite limited in terms of flexibility. It still burdens the caller with opening the Map themselves, as well as doing a lookup in CollectionSpec.Maps to find the right MapSpec to modify. Ideally, we would expose the offending MapSpec and Map for the user to manipulate directly.
1. Doing 1. through IncompatibleMapError would imply returning open Map handles to the caller, which cannot be opted out of. Not ideal.
1. A caller-provided callback like `OnIncompatibleMap func(spec *MapSpec, m *Map) error` would be more suited, since the library can stay in control of the lifecycle of the Map (closing it after executing the callback). This would be relatively unintrusive to implement in the ELF loader. If done right, the caller can completely avoid getting `ErrIncompatibleMap`, saving on verifier time.

Option 3. would tie nicely into my plans for the c/c loader infrastructure. I'll work on a proposal. I'll close this one out. Thanks again for working on this!

PR #1094, opened 2023-07-13
elf: don't clamp PerfEventArray.MaxEntries at parse time

lmb:
elf: don't clamp PerfEventArray.MaxEntries at parse time

    One key idea of the ELF reader is that it is deterministic: it should always
    produce the same output regardless of which system it runs on. The
    PerfEventArray clamping logic violates this since it substitutes the number
    of possible CPUs.

    Move the logic into map creation and adjust MapSpec.Compatible to do the
    same fixup.

map: fix MapSpec.Compatible for PerfEventArray with zero key or value size

    MapSpec.Compatible currently returns an error when comparing a
    PerfEventArray created from a MapSpec without explicit key or value sizes.
    This is because we only do the fixup just before executing the
    BPF_MAP_CREATE syscall.

    Move the fixup logic into a separate function so that we can invoke it from 
    MapSpec.Compatible.

map: fix buggy TestMapLoadReusePinned

    The test calls t.Fatal on the wrong (parent) t since it reuses qt.New from
    the parent scope. Remove qt.New.

### DIFF
--- a/map.go
+++ b/map.go
@@ -102,26 +102,26 @@ func (ms *MapSpec) Copy() *MapSpec {
 	return &cpy
 }
 
-func (ms *MapSpec) clampPerfEventArraySize() error {
-	if ms.Type != PerfEventArray {
-		return nil
-	}
-
+func fixupPerfEventArraySize(maxEntries uint32) (uint32, error) {
 	n, err := internal.PossibleCPUs()
 	if err != nil {
-		return fmt.Errorf("perf event array: %w", err)
+		return 0, fmt.Errorf("fixup perf event array: %w", err)
 	}
 
-	if n := uint32(n); ms.MaxEntries > n {
-		ms.MaxEntries = n
+	if n := uint32(n); maxEntries == 0 || maxEntries > n {
+		// MaxEntries should be zero most of the time, but there is code
+		// out there which hardcodes large constants. Clamp the number
+		// of entries to the number of CPUs at most. Allow creating maps with
+		// less than n items since some kernel selftests relied on this
+		// behaviour in the past.
+		return n, nil
 	}
 
-	return nil
+	return maxEntries, nil
 }
 
 // dataSection returns the contents and BTF Datasec descriptor of the spec.
 func (ms *MapSpec) dataSection() ([]byte, *btf.Datasec, error) {
-
 	if ms.Value == nil {
 		return nil, nil, errMapNoBTFValue
 	}
@@ -155,6 +155,15 @@ type MapKV struct {
 //
 // Returns an error wrapping [ErrMapIncompatible] otherwise.
 func (ms *MapSpec) Compatible(m *Map) error {
+	maxEntries := ms.MaxEntries
+	if ms.Type == PerfEventArray {
+		var err error
+		maxEntries, err = fixupPerfEventArraySize(maxEntries)
+		if err != nil {
+			return err
+		}
+	}
+
 	switch {
 	case m.typ != ms.Type:
 		return fmt.Errorf("expected type %v, got %v: %w", ms.Type, m.typ, ErrMapIncompatible)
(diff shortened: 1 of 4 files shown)


map.go
```diff
@@ -102,26 +102,24 @@ func (ms *MapSpec) Copy() *MapSpec {
...
+func fixupPerfEventArraySize(maxEntries uint32) (uint32, error) {
 	n, err := internal.PossibleCPUs()
 	if err != nil {
-		return fmt.Errorf("perf event array: %w", err)
+		return 0, err
 	}
 
-	if n := uint32(n); ms.MaxEntries > n {
-		ms.MaxEntries = n
+	if n := uint32(n); maxEntries == 0 || maxEntries > n {
```
rgo3, line 111: Why aren't we fixing up all possible inputs? What would `maxEntries > 0 && maxEntries < n` mean if a user uses that size?

rgo3 (approved):
Just a question on my part to better understand why we are fixing up the size of a `PerfEventArray` at all.

map.go
ti-mo: Good point. When specifying a `maxEntries < PossibleCPUs()` (let's say, 1 on a multi-core system), only bpf progs scheduled to cpu 0 will be able to successfully obtain a ring buffer to emit events to. Executions on all other CPUs will fail to emit any events.

It's not clear to me what the path forward is here. Maybe we could start rejecting all non-zero values to perf array maxentries and publish it as a breaking change?

lmb: The full context is here: https://github.com/cilium/ebpf/issues/209 TL;DR: some kernel selftests use a perf event array of size 1. Some users also do funky stuff.

> only bpf progs scheduled to cpu 0 will be able to successfully obtain a ring buffer to emit events to. Executions on all other CPUs will fail to emit any events.

This is because we set a CPU filter on perf event create? Maybe the linux selftest is just hopelessly buggy or maybe we're not doing the right thing.

> Maybe we could start rejecting all non-zero values to perf array maxentries and publish it as a breaking change? It would also mean we can remove the code dealing with the clamping and simplify the process a bit..

That's the old behaviour, which kept people creating issues.

ti-mo: Cool, perhaps worth a comment on the if statement?

lmb: Added a sentence which explains the < n case, let me know what you think

ti-mo: Looks good, thanks. :+1:

map.go
```diff
@@ -155,6 +155,15 @@ type MapKV struct {
 //
 // Returns an error wrapping [ErrMapIncompatible] otherwise.
 func (ms *MapSpec) Compatible(m *Map) error {
+	maxEntries := ms.MaxEntries
+	if ms.Type == PerfEventArray {
+		var err error
+		maxEntries, err = fixupPerfEventArraySize(maxEntries)
+		if err != nil {
+			return err
```
ti-mo, line 163: Should we wrap this? Otherwise, the caller will receive the raw error from internal.PossibleCPUs without any context.

map.go
```diff
@@ -376,12 +384,9 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 		spec.KeySize = 4
 		spec.ValueSize = 4
 
-		if spec.MaxEntries == 0 {
```
ti-mo, line 379: This will leave a 0 in spec.MaxEntries if getting possible CPUs fails. I know it's annoying, but we shouldn't do this. (also the Go tenet of not using the other value(s) if err != nil, etc..)

ti-mo (approved):
Left a few nits, thanks!

map.go
```diff
@@ -165,9 +174,8 @@ func (ms *MapSpec) Compatible(m *Map) error {
 	case m.valueSize != ms.ValueSize:
 		return fmt.Errorf("expected value size %v, got %v: %w", ms.ValueSize, m.valueSize, ErrMapIncompatible)
 
-	case !(ms.Type == PerfEventArray && ms.MaxEntries == 0) &&
-		m.maxEntries != ms.MaxEntries:
-		return fmt.Errorf("expected max entries %v, got %v: %w", ms.MaxEntries, m.maxEntries, ErrMapIncompatible)
+	case maxEntries != ms.MaxEntries:
```
ti-mo, line 177: I think this is supposed to be `m.maxEntries != maxEntries`? Comparing spec to itself otherwise.

ti-mo (changes requested):
Changing my review, think I found a bug.

map.go
lmb: Wrapped in `fixupPerfEventArraySize` instead.

lmb (commented):
Updated, should be OK now I hope. Found another bug with MapSpec compatibility which I've fixed in a separate commit.

map.go
ti-mo: Can't remember what the proposed change looked like before, not sure how I missed the copy. I guess it would've been fine.

My point was: if a function returns a non-nil error, all other return values (other than the error itself, of course) should be ignored. IIRC this said something like `spec.MaxEntries, err = ...`, so I was just pointing out the obvious.

PR #1099, opened 2023-07-19
[DRAFT] btf: handle BTF_KIND_FWD for struct/union in the type compat check

shun159:
This patch adds support for forward declarations of struct/union to enable the usage of kernel functions with a pointer to a incomplete kfunc's struct.

### DIFF
--- a/btf/core.go
+++ b/btf/core.go
@@ -902,6 +902,24 @@ func coreAreTypesCompatible(localType Type, targetType Type) error {
 		localType = UnderlyingType(*l)
 		targetType = UnderlyingType(*t)
 
+		/* For forward declarations, kflag dictates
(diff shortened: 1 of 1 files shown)


shun159:
Thank you for your reviewing, :bow: 
I removed `kind()` from types.go and use type switch instead as your comment.

catz-lw:
What conditions cause this error? Your PR fixes a very simple BTF-enabled program  that I have that works on Fedora 36, but fail on Fedora 37 and higher.

shun159:
@catz-lw Thank you for your comment. I will respond by weekend.

shun159:
Hi @catz-lw. I will revoke this PR.
I apologize for bothering you.

PR #1102, opened 2023-07-25
implementation for hinting on possible error on map create with noPrealloc flag set

### LINKED ISSUE #1072: Warn about non-supported flags in map specs
When an LRU map spec has the `BPF_F_NO_PREALLOC` flag, the library fails to load BPF programs with the error - `map create: operation not supported (without BTF k/v)`. This is misleading, as the main issue here is that the kernel doesn't support these flags for LRU type maps.

The library could give hints about such not supported configurations while loading map specs. 

It could be a candidate for good first issues, but I'll leave the judgement call to you. /cc @ti-mo

kwakubiney:
Fixes #1072

### DIFF
--- a/map.go
+++ b/map.go
@@ -422,68 +422,64 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 	}
 
 	fd, err := sys.MapCreate(&attr)
+
 	// Some map types don't support BTF k/v in earlier kernel versions.
 	// Remove BTF metadata and retry map creation.
 	if (errors.Is(err, sys.ENOTSUPP) || errors.Is(err, unix.EINVAL)) && attr.BtfFd != 0 {
 		attr.BtfFd, attr.BtfKeyTypeId, attr.BtfValueTypeId = 0, 0, 0
 		fd, err = sys.MapCreate(&attr)
 	}
-
 	if err != nil {
 		if errors.Is(err, unix.EPERM) {
 			return nil, fmt.Errorf("map create: %w (MEMLOCK may be too low, consider rlimit.RemoveMemlock)", err)
 		}
-		if errors.Is(err, unix.EINVAL) && attr.MaxEntries == 0 {
+		if errors.Is(err, unix.EINVAL) && spec.MaxEntries == 0 {
 			return nil, fmt.Errorf("map create: %w (MaxEntries may be incorrectly set to zero)", err)
 		}
 		if errors.Is(err, unix.EINVAL) && spec.Type == UnspecifiedMap {
 			return nil, fmt.Errorf("map create: cannot use type %s", UnspecifiedMap)
 		}
-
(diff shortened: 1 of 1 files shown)


kwakubiney:
The `mapCreate()` method has a lot going on in it & this makes it more bloated IMO. We could wrap the entire `switch` statement in its own function at this point.

map.go
```diff
@@ -387,6 +387,88 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 	if err != nil {
 		return nil, err
 	}
+	switch spec.Type {
```
florianl, line 390: Even if this change is still in draft mode, I want to mention the point, that map flags are subject of change. So hardcoding flag checks might leed to confusion or limit usages, especially if the version/patch status of the kernel is unknown.

[This is an example](https://github.com/torvalds/linux/commit/591fe9888d7809d9ee5c828020b6c6ae27c37229) for flag requirements for newer kernel.

What about providing hints based on the flag, if the map creation fails?

kwakubiney: I agree. I think @lmb also mentioned it could be a pain to maintain although it does help for better UX. I was thinking of a simple hint like:
```
if errors.Is(err, unix.EINVAL) && spec.Flags&unix.BPF_F_NO_PREALLOC > 0{
      return nil, fmt.Errorf("map create: %w (noPrealloc flag may be incompatible with map type %s)", err, spec.Type)
}
```

lmb: Agree with Florian, this must go into the `if err != nil {` case just after `MapCreate` (for the same reason we put the other feature detection checks there). Pulling it out into a separate function would also be nice.

map.go
```diff
@@ -430,55 +430,10 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
...
-		if attr.BtfFd == 0 {
-			return nil, fmt.Errorf("map create: %w (without BTF k/v)", err)
-		}
-
-		return nil, fmt.Errorf("map create: %w", err)
+		return nil, handleMapCreateError(attr, spec, err)
 	}
-	defer closeOnError(fd)
 
+	defer closeOnError(fd)
```
lmb, line 436: Nit: gratuitous whitespace change :P

map.go
```diff
@@ -1490,3 +1445,50 @@ func NewMapFromID(id MapID) (*Map, error) {
 
 	return newMapFromFD(fd)
 }
+
+func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) error {
```
lmb, line 1449: Nit: i think it'd be fine to put this just after `createMap`.

lmb (changes requested):
Nice! Could you split your change into two commits?
1. Move the existing code to a function
2. Add the new check you want to add

This will make it easier to do code archeology later on.

florianl (approved):
Thanks for the update! Suggested change looks good to me :+1: 
But I'll leave the final approval to the maintainers.

lmb (approved):

PR #1105, opened 2023-07-28
internal/sys: make generate-btf and regenerate types

florianl:
Update CI kernel and regenerate types.


internal/sys/types.go
```diff
@@ -535,6 +544,8 @@ type ProgInfo struct {
 	RunCnt               uint64
 	RecursionMisses      uint64
 	VerifiedInsns        uint32
+	AttachBtfObjId       uint32
```
lmb, line 547: Should either one of these be `BTFID` or `TypeID`?

florianl: does this sounds reasonable to you?
```
AttachBtfObjId BTFID
AttachBtfId       TypeID
```

lmb: I think it's the other way around actually.
* AttachBtfId should be [TypeID](https://elixir.bootlin.com/linux/v6.4.7/source/include/linux/bpf.h#L1368)
* AttachBtfObjId should be BTFID: [1](https://elixir.bootlin.com/linux/v6.4.7/source/kernel/bpf/syscall.c#L4163), [2](https://elixir.bootlin.com/linux/v6.4.7/source/kernel/bpf/btf.c#L7300)

florianl: :see_no_evil: seems I got lost in types....

I have updated the PR and replaced the types accordingly.

lmb (approved):
Thanks :)

PR #1111, opened 2023-08-09
Ringbuf read on timeout

RonFed:
The goal for this PR is describe in https://github.com/cilium/ebpf/discussions/1108

### DIFF
--- a/ringbuf/reader.go
+++ b/ringbuf/reader.go
@@ -186,7 +186,8 @@ func (r *Reader) SetDeadline(t time.Time) {
 // Read the next record from the BPF ringbuf.
 //
 // Returns os.ErrClosed if Close is called on the Reader, or os.ErrDeadlineExceeded
-// if a deadline was set.
+// if a deadline was set and no valid entry was present. A producer might use BPF_RB_NO_WAKEUP
+// which may cause the deadline to expire but a valid entry will be present.
 func (r *Reader) Read() (Record, error) {
 	var rec Record
 	return rec, r.ReadInto(&rec)
@@ -205,7 +206,8 @@ func (r *Reader) ReadInto(rec *Record) error {
 		if !r.haveData {
 			_, err := r.poller.Wait(r.epollEvents[:cap(r.epollEvents)], r.deadline)
 			if errors.Is(err, os.ErrDeadlineExceeded) && !r.ring.isEmpty() {
-				// TODO: Explain why we're ignoring this
+				// Ignoring this for reading a valid entry after timeout
+				// This can occur if the producer submitted to the ring buffer with BPF_RB_NO_WAKEUP
 				err = nil
 			}
 			if err != nil {
@@ -216,10 +218,12 @@ func (r *Reader) ReadInto(rec *Record) error {
 
 		for {
 			err := readRecord(r.ring, rec, r.header)
-			if errors.Is(err, errBusy) || errors.Is(err, errDiscard) {
+			// Not using errors.Is which is quite a bit slower
+			// For a tight loop it might make a difference
+			if err == errBusy || err == errDiscard {
 				continue
 			}
-			if errors.Is(err, errEOR) {
+			if err == errEOR {
 				r.haveData = false
 				break
 			}
--- a/ringbuf/reader.go
+++ b/ringbuf/reader.go
@@ -204,7 +204,7 @@ func (r *Reader) ReadInto(rec *Record) error {
 	for {
 		if !r.haveData {
 			_, err := r.poller.Wait(r.epollEvents[:cap(r.epollEvents)], r.deadline)
-			if err != nil {
+			if errors.Is(err, os.ErrClosed) || (errors.Is(err, os.ErrDeadlineExceeded) && r.ring.isEmpty()) {
 				return err
 			}
 			r.haveData = true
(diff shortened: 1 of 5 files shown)


ringbuf/reader.go
```diff
@@ -203,23 +204,29 @@ func (r *Reader) ReadInto(rec *Record) error {
...
+			if errors.Is(err, os.ErrDeadlineExceeded) {
+				r.haveData = !r.ring.isEmpty()
+				if !r.haveData {
+					return err
+				}
+			} else {
+				if err != nil {
+					return err
+				}
+				r.haveData = true
```
mythi, line 217: my original thinking was that something like this would be sufficient:
```suggestion
			_, err := r.poller.Wait(r.epollEvents[:cap(r.epollEvents)], r.deadline)
			if errors.Is(err, os.ErrClosed) || (errors.Is(err, os.ErrDeadlineExceeded) && r.ring.isEmpty()) {
			        return err
			}
```

would that work?

RonFed: Yeah, that is cleaner. two small comments:
1. I changed the err to be in the scope of the function: for the case where we have a timeout and there is available data - I want the err to contain the os.ErrDeadlineExceeded and won't get overriden to nil - such that the caller will know a timeout happened.
2. kept the r.haveData = true after the if statement

mythi: 1. sounds potentially problematic because when you cannot tell if the data is valid based the error?

RonFed: Right, after a second thought I agree.
Updated the new tests accordingly

ringbuf/reader.go
```diff
@@ -204,22 +204,21 @@ func (r *Reader) ReadInto(rec *Record) error {
 	for {
 		if !r.haveData {
 			_, err := r.poller.Wait(r.epollEvents[:cap(r.epollEvents)], r.deadline)
-			if err != nil {
+			if errors.Is(err, os.ErrClosed) || (errors.Is(err, os.ErrDeadlineExceeded) && r.ring.isEmpty()) {
```
lmb, line 207: Seems like we now ignore errors that are not `ErrClosed` or `ErrDeadlineExceeded`? I'd prefer to keep returning unknown errors. The if condition is also a bit hard to read. How about splitting it up like so:

```suggestion
			if errors.Is(err, os.ErrDeadlineExceeded) && !r.ring.isEmpty() {
				// TODO: Explain why we're ignoring this
				err = nil
			}
			if err != nil {
```

ringbuf/reader_test.go
```diff
@@ -234,10 +240,68 @@ func TestReaderBlocking(t *testing.T) {
...
+	lowerBound := timeoutMs * 98 / 100
+	upperBound := timeoutMs * 102 / 100
+	nonBlocking := timeoutMs * 2 / 100
+	startTime := time.Now()
+
+	rd.SetDeadline(startTime.Add(time.Duration(timeoutMs) * time.Millisecond))
+
+	record, err := rd.Read()
+
+	if time.Since(startTime).Milliseconds() < lowerBound || time.Since(startTime).Milliseconds() > upperBound {
```
lmb, line 277: In a way this tests the timeout implementation of epoll, which is not really something we control. It's also hard to make this test not-flaky across a variety of systems. Would you mind if we dropped the time checks and only made sure the semantics are correct?

ringbuf/reader_test.go
```diff
@@ -234,10 +240,68 @@ func TestReaderBlocking(t *testing.T) {
 	}
 }
 
+func TestReaderNoWakeup(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.8", "BPF ring buffer")
+
+	prog, events := mustOutputSamplesProg(t,
+		sampleMessage{size: 5, flags: unix.BPF_RB_NO_WAKEUP}, // Read after timeout
+		sampleMessage{size: 6, flags: unix.BPF_RB_NO_WAKEUP}, // Discard
+		sampleMessage{size: 7, flags: unix.BPF_RB_NO_WAKEUP}) // Read won't block
```
lmb, line 249: Nit: can you add `,` at the end?

ringbuf/reader_test.go
```diff
@@ -234,10 +240,68 @@ func TestReaderBlocking(t *testing.T) {
...
+		t.Fatal("Expected 0 as return value, got", errno)
+	}
+
+	var timeoutMs int64 = 100
+	lowerBound := timeoutMs * 98 / 100
+	upperBound := timeoutMs * 102 / 100
+	nonBlocking := timeoutMs * 2 / 100
+	startTime := time.Now()
+
+	rd.SetDeadline(startTime.Add(time.Duration(timeoutMs) * time.Millisecond))
```
lmb, line 273: This could just be
```suggestion
	rd.SetDeadline(time.Now())
```

This way we do a non-blocking read and can then check whether it returned a sample.

lmb (changes requested):
This looks really nice, thanks! `sampleMessage` is a good cleanup, we can use it to make discard behaviour more explicit. Left some comments around error handling and testing strategy.

Does it make sense to update `Read` to mention this behaviour? Or maybe this is better documented on `SetDeadline`?

RonFed:
> This looks really nice, thanks! `sampleMessage` is a good cleanup, we can use it to make discard behaviour more explicit. Left some comments around error handling and testing strategy.
> 
> Does it make sense to update `Read` to mention this behaviour? Or maybe this is better documented on `SetDeadline`?

Thanks for you comments,
I added a mention for this behaviour in `Read`, and also updated the test to remove the timing checks

ringbuf/reader_test.go
RonFed: My gofmt removes it so I assume it might break the lint checks ?

lmb (approved):
LGTM! I'll squash your changes and merge once CI has passed.

PR #1113, opened 2023-08-10
testutils: log random seed to test output

lmb:
Include the used seed in the test log so that test2json contains the full information needed to recreate a test failure. This is useful on CI which parses the output and provides a nice UI to see failures.

Also allow overriding the used seed via an environment variable.


lmb:
CI failures are due to me bungling kernel packaging again.

PR #1114, opened 2023-08-10
collection: Add Assign method

danobi:
Assign() bridges bpf2go generated code and functionality implemented in
Collection. A few things are easier to do in Collection, like iterating
through maps and programs. It is harder to guarantee invariants like map
and program initialization using reflection as well as being more
complicated.

Signed-off-by: Daniel Xu <dxu@dxuuu.xyz>

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -695,6 +695,71 @@ func LoadCollection(file string) (*Collection, error) {
 	return NewCollection(spec)
 }
 
+// Assign the contents of a Collection to a struct.
+//
+// This function bridges functionality between bpf2go generated
+// code and any functionality better implemented in Collection.
+//
+// 'to' must be a pointer to a struct. A field of the
+// struct is updated with values from Programs or Maps if it
+// has an `ebpf` tag and its type is *Program or *Map.
+// The tag's value specifies the name of the program or map as
+// found in the CollectionSpec.
+//
+//	struct {
+//	    Foo     *ebpf.Program `ebpf:"xdp_foo"`
+//	    Bar     *ebpf.Map     `ebpf:"bar_map"`
+//	    Ignored int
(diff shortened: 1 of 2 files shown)


danobi:
Looks like unrelated CI failure

collection.go
```diff
@@ -679,6 +679,56 @@ func LoadCollection(file string) (*Collection, error) {
...
+// for any successful assigns.
+func (coll *Collection) Assign(to interface{}) error {
+	// Assign() only transfers already-loaded Maps and Programs. No extra
+	// loading is done.
+	getValue := func(typ reflect.Type, name string) (interface{}, error) {
+		switch typ {
+
+		case reflect.TypeOf((*Program)(nil)):
+			if p := coll.Programs[name]; p != nil {
+				delete(coll.Programs, name)
```
lmb, line 712: From the doc:

```
// Ownership and Close()ing responsibility is transferred to `to`
// for any successful assigns.
```

This transfers ownership even if the assign fails?

lmb:
CI was my fault, sorry

collection.go
danobi: I meant to say partial assigns are possible. Seemed tricky to do rollbacks here. And also `Assign()` being called on something not generated by bpf2go seems like a rare case. So an actual error is unlikely.

WDYT? Update documentation or implement rollbacks?

lmb: Rollbacks don't have to be complicated I think, see `LoadAndAssign`. If it's okay I'd prefer rollbacks since that keeps semantics consistent.

danobi: Ack, will take a look

lmb:
Sorry for the long delay, this slipped my radar.

PR #1115, opened 2023-08-11
run-tests: support new selftest packaging

lmb:
The new selftests packaging has an additional ./ directory at the start of the filenames. Strip another path component.


lmb:
This doesn't work as a quick fix since new selftests don't contain `btf_testmod` :(

PR #1118, opened 2023-08-16
Add metadata to instructions returned by program info

dylandreimerink:
The instructions returned by program info gotten from the kernel did
not contain any metadata such as line info or function info. This
makes it harder to read the instructions when comparing the xlated
instruction in the kernel with the spec.

This PR uses the line info and function info that is available via
the program info and adds it to the returned instructions.

### DIFF
--- a/btf/ext_info.go
+++ b/btf/ext_info.go
@@ -111,6 +111,16 @@ func (ei *ExtInfos) Assign(insns asm.Instructions, section string) {
 	lineInfos := ei.lineInfos[section]
 	reloInfos := ei.relocationInfos[section]
 
+	AssignMetadataToInstructions(insns, funcInfos, lineInfos, reloInfos)
+}
+
+// Assign per-instruction metadata to the instructions in insns.
+func AssignMetadataToInstructions(
+	insns asm.Instructions,
+	funcInfos FuncInfos,
+	lineInfos LineInfos,
+	reloInfos CORERelocationInfos,
+) {
 	iter := insns.Iterate()
 	for iter.Next() {
 		if len(funcInfos.infos) > 0 && funcInfos.infos[0].offset == iter.Offset {
@@ -381,6 +391,21 @@ func newFuncInfos(bfis []bpfFuncInfo, spec *Spec) (FuncInfos, error) {
 	return fis, nil
 }
 
+// LoadFuncInfos parses btf func info in wire format.
+func LoadFuncInfos(reader io.Reader, bo binary.ByteOrder, recordNum uint32, spec *Spec) (FuncInfos, error) {
+	fis, err := parseFuncInfoRecords(
+		reader,
+		bo,
+		FuncInfoSize,
+		recordNum,
+	)
+	if err != nil {
+		return FuncInfos{}, fmt.Errorf("parsing BTF func info: %w", err)
+	}
+
+	return newFuncInfos(fis, spec)
+}
+
 // marshal into the BTF wire format.
 func (fi *funcInfo) marshal(w *bytes.Buffer, b *Builder) error {
 	id, err := b.Add(fi.fn)
@@ -510,6 +535,21 @@ type bpfLineInfo struct {
 	FileNameOff uint32
 	LineOff     uint32
 	LineCol     uint32
+}
+
+// LoadLineInfos parses btf line info in wire format.
+func LoadLineInfos(reader io.Reader, bo binary.ByteOrder, recordNum uint32, spec *Spec) (LineInfos, error) {
+	lis, err := parseLineInfoRecords(
+		reader,
+		bo,
+		LineInfoSize,
+		recordNum,
+	)
+	if err != nil {
+		return LineInfos{}, fmt.Errorf("parsing BTF line info: %w", err)
+	}
+
+	return newLineInfos(lis, spec.strings)
 }
 
 func newLineInfo(li bpfLineInfo, strings *stringTable) (*lineInfo, error) {
--- a/btf/ext_info.go
+++ b/btf/ext_info.go
@@ -16,9 +16,9 @@ import (
 // ExtInfos contains ELF section metadata.
 type ExtInfos struct {
 	// The slices are sorted by offset in ascending order.
-	funcInfos       map[string][]funcInfo
-	lineInfos       map[string][]lineInfo
-	relocationInfos map[string][]coreRelocationInfo
+	funcInfos       map[string]FuncInfos
+	lineInfos       map[string]LineInfos
+	relocationInfos map[string]CORERelocationInfos
 }
 
 // loadExtInfosFromELF parses ext infos from the .BTF.ext section in an ELF.
@@ -58,7 +58,7 @@ func loadExtInfos(r io.ReaderAt, bo binary.ByteOrder, spec *Spec, strings *strin
 		return nil, fmt.Errorf("parsing BTF function info: %w", err)
 	}
 
-	funcInfos := make(map[string][]funcInfo, len(btfFuncInfos))
+	funcInfos := make(map[string]FuncInfos, len(btfFuncInfos))
 	for section, bfis := range btfFuncInfos {
 		funcInfos[section], err = newFuncInfos(bfis, spec)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


btf/ext_info.go
```diff
@@ -130,6 +130,95 @@ func (ei *ExtInfos) Assign(insns asm.Instructions, section string) {
...
+		spec: spec,
+	}
+}
+
+// AddFuncInfos adds function info for a section to the builder.
+// The funcInfo should be in BTF wire format and native endianness.
+func (ei *ExtInfoBuilder) AddFuncInfos(section string, funcInfo []byte) error {
+	bfi, err := parseFuncInfoRecords(
+		bytes.NewReader(funcInfo),
+		internal.NativeEndian,
```
lmb, line 156: Since we're adding public API we shouldn't hardcode native endian.

lmb (changes requested):
This is a really nice idea, when we wrote the current code to pull out instructions metadata wasn't a thing yet. This reminds me that we could also use metadata to tag map load isns we get from the kernel, I should create an issue.

Some high level thoughts: seems like what you really want to do is calling `Assign()`. All of the other stuff with `ExtInfoBuilder` is just a side-show. What do you think about the following?

* Export `funcInfo` and `lineInfo`
* Export constructors for `[]FuncInfo`, `[]LineInfo`. Those should take a byte order, reader and a Spec probably. These are somewhere between `parseFuncInfos` and `parseFuncInfoRecords`: they take infos without the header and return a slice instead of a map.
* Add a function `AssignInstructionMetadata(Instructions, []FuncInfo, []LineInfo)` and call that from `ExtInfos.Assign`. The core relocation part we can keep in `ExtInfos.Assign` for now.

In general I want to avoid public APIs which assume native endian as much as possible. The same goes for the record size.

dylandreimerink:
> Some high level thoughts: seems like what you really want to do is calling Assign(). All of the other stuff with ExtInfoBuilder is just a side-show. What do you think about the following?
>
>* Export funcInfo and lineInfo
>* Export constructors for []FuncInfo, []LineInfo. Those should take a byte order, reader and a Spec probably. These are somewhere between parseFuncInfos and parseFuncInfoRecords: they take infos without the header and return a slice instead of a map.
> * Add a function AssignInstructionMetadata(Instructions, []FuncInfo, []LineInfo) and call that from ExtInfos.Assign. The core relocation part we can keep in ExtInfos.Assign for now.

Yea, that is fair, sounds like a good way to approach this. I will make it so

info.go
```diff
@@ -259,8 +282,47 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
...
+
+		spec, err := btfh.Spec(nil)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get BTF spec: %w", err)
+		}
+
+		lineInfos, err := btf.NewLineInfos(
+			bytes.NewReader(pi.lineInfos),
+			internal.NativeEndian,
+			uint32(len(pi.lineInfos))/btf.LineInfoSize,
```
lmb, line 300: This feels a bit too "raw" to me: the user shouldn't have to care what the wire format size is. How about one of these options:

1. Take `sizedReader` instead of `io.Reader` and infer number of records in `NewLineInfos`.
2. Read records from `io.Reader` until `io.EOF` and assert that `len(lineInfos) == info2.NrLineInfos` here.

I have a slight preference for option 2 since it doesn't involve defining a custom reader interface and seems more flexible overall.

btf/ext_info.go
```diff
@@ -374,6 +384,21 @@ func newFuncInfos(bfis []bpfFuncInfo, spec *Spec) ([]FuncInfo, error) {
 	return fis, nil
 }
 
+// NewFuncInfos parses btf func info in wire format.
+func NewFuncInfos(reader io.Reader, bo binary.ByteOrder, recordNum uint32, spec *Spec) ([]FuncInfo, error) {
```
lmb, line 388: Nit: other constructors from a reader in this package are `Load*`. Should these be `LoadFuncInfos`?

lmb (changes requested):

info.go
```diff
@@ -259,8 +282,47 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
 		return nil, fmt.Errorf("unmarshaling instructions: %w", err)
 	}
 
-	// Tag the first instruction with the name of the program, if available.
-	insns[0] = insns[0].WithSymbol(pi.Name)
+	if pi.btf != 0 {
+		btfh, err := btf.NewHandleFromID(pi.btf)
```
lmb, line 286: I just realised that this needs [`CAP_SYS_ADMIN`](https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L4475). That's a problem since newProgramInfoFromFd should be callable without SYS_ADMIN.

How about the following:
1. Populate `pi.funcInfos`, etc. as you currently do.
2. Move this block to `ProgramInfo.Instructions()`

That leaves the question what to do when `Instructions` is called without `CAP_SYS_ADMIN` and we get EPERM from `NewHandleFromID`?

btf/ext_info.go
```diff
@@ -111,6 +111,16 @@ func (ei *ExtInfos) Assign(insns asm.Instructions, section string) {
 	lineInfos := ei.lineInfos[section]
 	reloInfos := ei.relocationInfos[section]
 
+	AssignMetadataToInstructions(insns, funcInfos, lineInfos, reloInfos)
+}
+
+// Assign per-instruction metadata to the instructions in insns.
```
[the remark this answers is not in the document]
dylandreimerink: Did this. But I also realized that since offset is a non-exported value at the moment, any users outside of the btf package will have no way to actually sort. Would it make sense to define a type alias for `[]LineInfo` and `[]FuncInfo` which implement [`sort.Interface`](https://pkg.go.dev/sort#Interface)?

info.go
dylandreimerink: What you are describing is the current situation, is it not? The marked line which calls `btf.NewHandleFromID(pi.btf)` is inside `ProgramInfo.Instructions()` already. `newProgramInfoFromFd` can be called without `CAP_SYS_ADMIN`.

> That leaves the question what to do when Instructions is called without CAP_SYS_ADMIN and we get EPERM from NewHandleFromID?

Straight up changing could break current users. I think we can catch the `-EPERM` and fall back to old behavior. Then add to the docs that BTF info is only added when `CAP_SYS_ADMIN` is available. Would that be acceptable?

lmb: Sorry, I didn't spot that in the diff :disappointed:

> Then add to the docs that BTF info is only added when CAP_SYS_ADMIN is available. Would that be acceptable?

Yeah that sounds okay.

info.go
dylandreimerink: >  assert that len(lineInfos) == info2.NrLineInfos here

`pi.lineInfos` is created as such `pi.lineInfos = make([]byte, btf.LineInfoSize*info.NrLineInfo)` so this is already true. 

> This feels a bit too "raw" to me

Yea, I get that. What we are currently doing is re-calculating `info.NrLineInfo` from the data we do have. So I suggest just adding the number of entries to `pi` so we can skip the confusing math.

btf/ext_info.go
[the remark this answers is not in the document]
lmb: Hmm good point. New types with `sort.Interface` kind of make sense, except that I don't expect users to ever need to sort the slices, since they're sorted by the constructor. Doing it this way might lead to over sorting.

Thinking about a future API extensions: appending two `[]FuncInfo` to each other doesn't produce a valid result, since the instruction offsets are all wrong. So if we ever want to support that we'd need a wrapper type. So for now, I'd drop the documentation here. Export `type FuncInfos struct { infos []funcInfo }`, etc. instead of `FuncInfo`. That prevents users from messing with the order. Somewhere down the line we can add `FuncInfos.Append(FuncInfos)` if necessary.

What do you think?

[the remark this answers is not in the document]
dylandreimerink: Sounds like a good idea :+1:

lmb (approved):
Thanks!

PR #1125, opened 2023-08-31
elf: Add support for map struct typedef

anjmao:
Fixes https://github.com/cilium/ebpf/issues/1126


elf_reader.go
```diff
@@ -750,7 +750,12 @@ func (ec *elfCode) loadBTFMaps() error {
 			// Each Var representing a BTF map definition contains a Struct.
 			mapStruct, ok := v.Type.(*btf.Struct)
```
lmb, line 751: I think using https://pkg.go.dev/github.com/cilium/ebpf/btf#UnderlyingType here is more appropriate.

lmb (changes requested):
Does this work with `libbpf`?

anjmao:
> Does this work with `libbpf`?

Yes, typedefs are used in some tracee ebpf code places. It uses libbpf and libbpfgo cgo wrapper. It's great that this was the only needed change to migrate such large ebpf code base to cilium/ebpf.

lmb:
I forgot: can you add a test in `loader.c` for this behaviour?

PR #1128, opened 2023-09-11
elf: don't set BPF_F_MMAPABLE on .kconfig

lmb:
We don't need to set the MMAPABLE flag on .kconfig section since the library never mmaps it. This allows using .kconfig on kernel
>=5.2 instead of >=5.5.


ti-mo (approved):
Thanks!

PR #1132, opened 2023-09-13
btf: work around missing ENUM64 support

### LINKED ISSUE #1116: btf: provide work around for missing BTF_KIND_ENUM64 support
pahole from v1.24 generates a new BTF_KIND_ENUM64. The library has support for this as well, and only encodes an enum64 if necessary to represent a type. This still causes a problem when compiling an old kernel with a new pahole, since `TestRoundtripVmlinux` then tries to load the enum64 into the kernel.

This shows that it would be nice^H^H^H it's necessary to have a general workaround for this. `libbpf` does the following when enum64 is not supported:
* It clears the kind flag on BTF_KIND_ENUM and BTF_KIND_ENUM64. This is used to specify signedness and only cosmetic from what I can tell.
* It replaces BTF_KIND_ENUM64 with a union.

Details are here: https://lwn.net/ml/bpf/20220603015937.1190992-1-yhs@fb.com/

lmb:
btf: move feature tests to separate file

    No other changes intended.

btf: add feature test for BTF_KIND_ENUM64

btf: work around missing ENUM64 support on older kernels

    Replace an ENUM64 with a 64-bit integer on kernels which don't support the
    former. This behaviour is similar to libbpf, which replaces the ENUM64 with
    a union instead.

    Remove the old heuristic which encoded a 64-bit enum as a 32-bit enum if
    none of the values exceeded math.MaxUint32. The heuristic has no tests and
    doesn't take the signedness of the enum into account.

Fixes #1116

### DIFF
--- a/btf/marshal.go
+++ b/btf/marshal.go
@@ -18,6 +18,8 @@ type MarshalOptions struct {
 	Order binary.ByteOrder
 	// Remove function linkage information for compatibility with <5.6 kernels.
 	StripFuncLinkage bool
+	// Replace Enum64 with a placeholder for compatibility with <6.0 kernels.
+	ReplaceEnum64 bool
 }
 
 // KernelMarshalOptions will generate BTF suitable for the current kernel.
@@ -25,6 +27,7 @@ func KernelMarshalOptions() *MarshalOptions {
 	return &MarshalOptions{
 		Order:            internal.NativeEndian,
 		StripFuncLinkage: haveFuncLinkage() != nil,
+		ReplaceEnum64:    haveEnum64() != nil,
 	}
 }
 
@@ -328,21 +331,13 @@ func (e *encoder) deflateType(typ Type) (err error) {
 		raw.data, err = e.convertMembers(&raw.btfType, v.Members)
 
 	case *Union:
-		raw.SetKind(kindUnion)
-		raw.SetSize(v.Size)
-		raw.data, err = e.convertMembers(&raw.btfType, v.Members)
+		err = e.deflateUnion(&raw, v)
 
 	case *Enum:
-		raw.SetSize(v.size())
-		raw.SetVlen(len(v.Values))
-		raw.SetSigned(v.Signed)
-
-		if v.has64BitValues() {
-			raw.SetKind(kindEnum64)
-			raw.data, err = e.deflateEnum64Values(v.Values)
+		if v.Size == 8 {
+			err = e.deflateEnum64(&raw, v)
 		} else {
-			raw.SetKind(kindEnum)
-			raw.data, err = e.deflateEnumValues(v.Values)
+			err = e.deflateEnum(&raw, v)
 		}
 
 	case *Fwd:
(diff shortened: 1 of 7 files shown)


lmb:
libbpf implementation was added in https://lwn.net/ml/bpf/20220603015937.1190992-1-yhs@fb.com/

btf/marshal.go
```diff
@@ -464,6 +478,33 @@ func (e *encoder) deflateEnumValues(values []EnumValue) ([]btfEnum, error) {
 	return bes, nil
 }
 
+func (e *encoder) deflateEnum64(raw *rawType, enum *Enum) (err error) {
+	if e.ReplaceEnum64 {
+		// Replace an ENUM64 with an integer of the appropriate size. libbpf
+		// instead uses a union of (u)int64 which allows preserving the
+		// enum names (but not their values). Using the integer is simpler for
+		// us.
+		raw.SetKind(kindInt)
```
lmb, line 487: Use `union` instead to match libbpf behaviour, otherwise we might create BTF that is incompatible with libbpf.

lmb:
Now uses a `union` instead of the int. Please take another look.

btf/marshal.go
```diff
@@ -464,6 +483,41 @@ func (e *encoder) deflateEnumValues(values []EnumValue) ([]btfEnum, error) {
 	return bes, nil
 }
 
+func (e *encoder) deflateEnum64(raw *rawType, enum *Enum) (err error) {
+	if e.ReplaceEnum64 {
```
ti-mo, line 487: Let's add a quick explanation with some context here. (e.g. why a union and not simply an int like we discussed offline)

btf/marshal.go
```diff
@@ -443,16 +445,33 @@ func (e *encoder) convertMembers(header *btfType, members []Member) ([]btfMember
...
 		nameOff, err := e.strings.Add(value.Name)
 		if err != nil {
 			return nil, err
 		}
 
-		if value.Value > math.MaxUint32 {
-			return nil, fmt.Errorf("value of enum %q exceeds 32 bits", value.Name)
+		if enum.Signed {
+			signedValue := int64(value.Value)
+			if signedValue != int64(int32(uint32(value.Value))) {
```
ti-mo, line 468: This is too clever to be uncommented. Why compare value.Value against its sign-extended self? Wouldn't truncating it to 32 bits put it within acceptable range anyway? (I assume that's part of the trick) FYI there's now math.MaxInt32 as well, maybe just check against both instead of special-casing signed/unsigned?

ti-mo (changes requested):
Thanks! Let's make the bounds check a little clearer if possible. :pray:

btf/marshal.go
lmb: Fair point, but I'll add it to the commit message instead. I think the comment would be too large.

btf/marshal.go
lmb: Thanks, comparing against the constants is much easier to understand.

rgo3 (approved):

PR #1140, opened 2023-09-27
Update bpf2go and stringer go:generate statements across the codebase

ti-mo:
https://github.com/cilium/ebpf/pull/1047 added env support to bpf2go, so use it.

Also cleaned up the stringer statements that would otherwise pick an old version from my GOBIN.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ container-all:
 		--env CFLAGS="-fdebug-prefix-map=/ebpf=." \
 		--env HOME="/tmp" \
 		"${IMAGE}:${VERSION}" \
-		make all
+		make BPF2GO_CC="$(CLANG)" BPF2GO_FLAGS="$(CFLAGS)" all
 
 # (debug) Drop the user into a shell inside the container as root.
 container-shell:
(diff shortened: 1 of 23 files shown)


Makefile
```diff
@@ -81,9 +81,6 @@ all: format $(addsuffix -el.elf,$(TARGETS)) $(addsuffix -eb.elf,$(TARGETS)) gene
 	ln -srf testdata/loader-$(CLANG)-el.elf testdata/loader-el.elf
 	ln -srf testdata/loader-$(CLANG)-eb.elf testdata/loader-eb.elf
 
-# $BPF_CLANG is used in go:generate invocations.
-generate: export BPF_CLANG := $(CLANG)
```
lmb, line 85: Doesn't this break `make generate` outside of container?

asm/alu.go
```diff
@@ -1,6 +1,6 @@
 package asm
 
-//go:generate stringer -output alu_string.go -type=Source,Endianness,ALUOp
+//go:generate go run golang.org/x/tools/cmd/stringer@latest -output alu_string.go -type=Source,Endianness,ALUOp
```
lmb, line 3: How does this work without adding a dependency on stringer to go.mod?

Makefile
ti-mo: Not for me, it defaults to `clang` and `llvm-strip`. I also sent a patch that removed a hardcoded clang-14 somewhere.

PR #1141, opened 2023-09-28
Drop LLVM 7 and 9 in favor of 11, 14 and 17

ti-mo:
```
This is driven by a few reasons:
- To reproducibly build C examples for our documentation, we want to rely on
  the libbpf-dev Debian package for <bpf/bpf_helpers.h>. Our current llvm
  toolchain image (Buster) ships a libbpf-dev package that doesn't contain this
  header, so we are encouraged to upgrade.

- Bullseye, the next logical candidate, has LLVM packages for 11 all the way
  through 17. This means we would need to drop versions <11 from CI.

- LLVM 7 came out in 2018, LLVM 9 in 2019. Given the development pace of BPF
  and most distributions dropping these versions long ago (Debian Bookworm
  ships LLVM 14), this seems reasonable.
```

Also removed Semaphore config, made 'clang' the default binary in bpf2go tests, and made some assorted Makefile changes.


testdata/docker/VERSION
```diff
@@ -1 +1 @@
-1694533004
+1695896017
```
lmb, line 1: Pushing the image worked?

ti-mo: Yep, all good 👍

Issue #1133, opened 2023-09-14
error read auxv in vdsoMemoryAddress function in arm_32 env


hailwind:
**Describe the bug**
EOF read auxv in vdsoMemoryAddress function in arm_32 env

code position 
https://github.com/cilium/ebpf/blob/main/internal/vdso.go#L65

**To Reproduce**

```sh
#uint64
root@1c40e8105b98:/sdwan# ./testx 
2023/09/14 22:31:41 Tag: 13752409965245497377, Val: 17927356702261264
2023/09/14 22:31:41 Tag: 17592186044422, Val: 429496729617
2023/09/14 22:31:41 Tag: 281698315010051, Val: 137438953476
2023/09/14 22:31:41 Tag: 30064771077, Val: 7
2023/09/14 22:31:41 Tag: 8, Val: 286371239428105
2023/09/14 22:31:41 Tag: 11, Val: 12
2023/09/14 22:31:41 Tag: 13, Val: 14
2023/09/14 22:31:41 Tag: 23, Val: 13747605919770869785
2023/09/14 22:31:41 Tag: 26, Val: 13747607246915764255
2023/09/14 22:31:41 Tag: 13747605988490346511, Val: 0
2023/09/14 22:31:41 reading auxv entry: EOF

#uint32
root@1c40e8105b98:/tmp# ./testx 
2023/09/14 22:32:26 Tag: 33, Val: 3204325376
2023/09/14 22:32:26 aux.VAL: 3204325376

root@1c40e8105b98:/tmp# uname -a
Linux testx 5.15.130 #0 SMP Thu Sep 14 07:12:57 2023 armv7l GNU/Linux
```

troubleshooting code
```go
package main

import (
	"encoding/binary"
	"errors"
	"log"
	"os"

	"golang.org/x/sys/unix"
)

func main() {
	av, err := os.Open("/proc/self/auxv")
	if errors.Is(err, unix.EACCES) {
		log.Printf("opening auxv: %s (process may not be dumpable due to file capabilities)", err)
	}

	if err != nil {
		log.Printf("opening auxv: %s", err)
	}
	defer av.Close()

	const (
		_AT_NULL         = 0  // End of vector
		_AT_SYSINFO_EHDR = 33 // Offset to vDSO blob in process image
	)
	var NativeEndian binary.ByteOrder = binary.LittleEndian
	// Loop through all tag/value pairs in auxv until we find `AT_SYSINFO_EHDR`,
[...]

PR #1144, opened 2023-10-02
internal: fix reading auxv on 32-bit platforms

lmb:
It turns out that the auxiliary vector has a platform specific size. Adjust the code to use uintptr to approximate "unsigned long" from C.

Fixes https://github.com/cilium/ebpf/issues/1133


internal/vdso.go
```diff
@@ -16,6 +17,8 @@ var (
 	errAuxvNoVDSO = errors.New("no vdso address found in auxv")
 )
 
+const uintptrIs32bits = unsafe.Sizeof((uintptr)(0)) == 4
```
brycekahle, line 20: Is the size of `uintptr` known at build time? Is this something we can fix with build constraints?

lmb: We'd have to fix it by maintaining a list of build tags like we do here: https://github.com/cilium/ebpf/blob/main/internal/sys/ptr_64.go https://github.com/cilium/ebpf/blob/main/internal/sys/ptr_32_be.go

brycekahle: Is that something you'd want to do, or do you prefer this existing way?

lmb: I thought about this a bit, and I think the current approach is easier to maintain. With build tags there is always a part of the tree that isn't covered by regular dev environment. Any mistakes only blow up in CI.

The constant might look kind of hacky but AFAIK the compiler will eliminate the dead code. So we get the same speed with better dev experience.

brycekahle (approved):

internal/vdso.go
ti-mo: This is fine. :wink:

ti-mo:
@lmb Here's `/proc/self/auxv` from an i686 bookworm VM: [auxv32.txt](https://github.com/cilium/ebpf/files/12869790/auxv32.txt). Named .txt because GitHub.

PR #1148, opened 2023-10-03
dependabot: add configuration, ignore indirect Python dependencies

ti-mo:
What it says on the tin. We immediately got slapped with a dependabot PR after merging the documentation. Ignore security vulnerabilities in transient dependencies, we're only interested in keeping up with direct dependencies since Python code is only run during documentation build.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/docs"
+    schedule:
+      interval: "weekly"
+    allow:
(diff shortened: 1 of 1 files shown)


lmb:
Urgh I hate dependabot with a passion. Can we just turn it off?

ti-mo:
I am interested in keeping the docs' dependencies up-to-date since mkdocs-material gains nice features over time, and you just don't want to deal with a big-bang dependency upgrade every few months.

lmb:
Then let's do this monthly? It's essentially busy work for humans generated by robots. Nobody every reviews the bumped versions anyways.

PR #1156, opened 2023-10-06
collection: Freeze map right before prog load

chenhengqi:
If freeze is requested, let's do that right before prog load. The freeze property would aid the verifier process at prog load ([0]).

  [0]: https://github.com/cilium/ebpf/discussions/1143


ti-mo:
Oh no. Looks like we need to split map finalization into two stages. We should finalize all scalar maps (non-prog arrays and outer maps) before loading progs. Special-casing frozen maps is a hack, let me come up with a solution.

PR #1159, opened 2023-10-06
Freeze .rodata before trying to load programs

ti-mo:
As reported in https://github.com/cilium/ebpf/discussions/1143 and https://github.com/cilium/ebpf/pull/1156.

The change is fairly minimal, had to reindent most of `populateMaps()`.

The gist of it is `cl.loadMap()` now finalizes maps that can be finalized right after loading, aka. 'scalar' maps without references to other bpf resources. Prog maps or outer maps are skipped.

After doing a first pass over all requested Maps and Programs, a separate pass is run by the renamed `populateDeferredMaps()` that only considers prog maps and outer maps, lazy-loads dependencies and fully resolves `Spec.Contents`, then finalizes the Map.


chenhengqi (commented):
LGTM.

chenhengqi:
Testing good on my side, thanks.

elf_reader.go
```diff
@@ -1258,7 +1258,7 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
 		{"cgroup/getsockopt", CGroupSockopt, AttachCGroupGetsockopt, 0},
 		{"cgroup/setsockopt", CGroupSockopt, AttachCGroupSetsockopt, 0},
 		{"struct_ops+", StructOps, AttachNone, 0},
-		{"sk_lookup/", SkLookup, AttachSkLookup, 0},
+		{"sk_lookup", SkLookup, AttachSkLookup, 0},
```
lmb, line 1261: FWIW we aren't compatible with libbpf here, since we treat all of these names as a prefix. Might make sense to pull this into a separate PR and discuss there.

collection.go
```diff
@@ -471,6 +471,15 @@ func (cl *collectionLoader) loadMap(mapName string) (*Map, error) {
 		return nil, fmt.Errorf("map %s: %w", mapName, err)
 	}
 
+	// Finalize 'scalar' maps that don't refer to any other eBPF resources
+	// potentially pending creation. This is needed for frozen maps like .rodata
+	// that need to be finalized before invoking the verifier.
+	if !mapSpec.Type.canStoreMapOrProgram() {
+		if err := m.finalize(mapSpec); err != nil {
```
lmb, line 478: Can you remind me why we can't populate deferred maps at this point as well?

ti-mo: Map fds need to be relocated into programs before a program can be loaded, but we can't populate a prog array without creating programs first. We've split up this process into multiple stages to keep things clear. Technically, we only need to defer populating prog arrays, but we can refactor this again later if new requirements pop up.

elf_reader.go
ti-mo: Yeah, I've started noticing that as well. I'll take it out for now.

ti-mo: We should probably track this as a larger work item, changing only sk_lookup doesn't make things any more consistent.

ti-mo: Created this: https://github.com/cilium/ebpf/issues/1162.

Issue #1168, opened 2023-10-17
ProgramInfo.Instructions fails on parsing BTF line info


aibor:
**Describe the bug**

With version 0.12.0 calling `ProgramInfo.Instructions` returns:

```
parse line info: parsing BTF line info: offset 1 is not aligned with instruction size
```

With 0.11.0 the method returns the Instructions as expected.

**To Reproduce**

Given a simple XDP program in `src/src.c`, that doesn't do much, but more than just returning:

```c
__attribute__((section("xdp"), used))
int test(void *ctx) {
  return ctx == 0;
}
```

And a simple go test program: 

```go
package main

import (
	"testing"
)

//go:generate ./gobin/bpf2go -target bpfel test src/src.c

func TestInstructions(t *testing.T) {
	var objs testObjects
	err := loadTestObjects(&objs, nil)
	if err != nil {
		t.Fatalf("load: %v", err)
	}

	info, err := objs.Test.Info()
	if err != nil {
		t.Fatalf("info: %v", err)
	}

	ins, err := info.Instructions()
	if err != nil {
		t.Fatalf("instructions: %v", err)
	}

	t.Log(ins)
}
```

The test fails with:

```
=== RUN   TestInstructions
    main_test.go:29: instructions: parse line info: parsing BTF line info: offset 1 is not aligned with instruction size
--- FAIL: TestInstructions (0.00s)
```

**Expected behavior**

With version 0.11.0 the method does not fail and produces the output:

```
=== RUN   TestInstructions
    main_test.go:32: test:
        	0: MovImm dst: r0 imm: 1
        	1: JEqImm dst: r1 off: 1 imm: 0
        	2: MovImm dst: r0 imm: 0
        	3: Exit

--- PASS: TestInstructions (0.00s)
```

lmb:
cc @dylandreimerink can you take a look?

PR #1169, opened 2023-10-17
btf,info: Fix bad instruction offset when parsing infos from kernel

dylandreimerink:
When BTF ext info is encoded in ELF the instruction offsets are in bytes, but when you pass them to the kernel they must be instruction indecies. Therefor we devide the offset by the instruction size in the parsing logic for ELF.

This was missed during the initial implementation of reading back BTF ext info from the kernel. This would cause an error when loading back ext info which was not a multiple of the instruction size or a bad instruction offset if it was.

This PR adds an argument to the parsing logic to specify where the ext info is being read from. It also adds `LoadKernelLineInfos` and `LoadKernelFuncInfos` to load the ext info from the kernel, using the correct direction enum internaly. These are now used by the `info` package.

Fixes: #1168

### DIFF
--- a/btf/ext_info.go
+++ b/btf/ext_info.go
@@ -391,13 +391,14 @@ func newFuncInfos(bfis []bpfFuncInfo, spec *Spec) (FuncInfos, error) {
 	return fis, nil
 }
 
-// LoadFuncInfos parses btf func info in wire format.
+// LoadFuncInfos parses BTF func info in kernel wire format.
 func LoadFuncInfos(reader io.Reader, bo binary.ByteOrder, recordNum uint32, spec *Spec) (FuncInfos, error) {
 	fis, err := parseFuncInfoRecords(
 		reader,
 		bo,
 		FuncInfoSize,
 		recordNum,
+		false,
 	)
 	if err != nil {
 		return FuncInfos{}, fmt.Errorf("parsing BTF func info: %w", err)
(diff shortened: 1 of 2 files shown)


lmb:
So how come this wasn't caught by tests? We don't have any non-zero offsets?

dylandreimerink:
Yea, both of the test that could have caught this have very small programs where this doesn't surface:

```
=== RUN   TestProgInfoExtBTF
sched_process_exec:
         ; return 0;
        0: MovImm dst: r0 imm: 0
        1: Exit
        
 === RUN   TestProgramInstructions
test_prog:
        0: LdImmDW dst: r0 imm: -1
        2: Exit
```

Do we want a regression test for this?

lmb:
> Do we want a regression test for this?

That would be great!

dylandreimerink:
I changed the `TestProgInfoExtBTF` test that was added in the original PR to test this feature. It now loads a more complex program with multiple line infos and func infos to verify that this feature works. Should be a better test.

lmb:
@dylandreimerink super, thanks! I've pushed a fixup:

* I don't think we need the explicit `LoadKernel...`. My POV is that either we parse a full ELF, or we parse parts of the BTF from the kernel.
* I dislike `bool` arguments to functions but here it really makes the code easier to read.
* I simplified the test to not glob for the loader. There is no point in parsing the "wrong" endian files.

Let me know what you think!

dylandreimerink:
Yes, all seem sensible! If we can get CI to agree that is

Issue #1175, opened 2023-10-19
example xdp: The IP address output by xdp demo is empty


BillyChen1:
**Describe the bug**
The IP address output by [xdp demo](https://github.com/cilium/ebpf/tree/main/examples/xdp) is empty (NIL)

**To Reproduce**

Env:
```sh
~ go version
go version go1.21.1 linux/amd64
~ uname -a
Linux cqm-OptiPlex-7040 5.15.0-83-generic #92~20.04.1-Ubuntu SMP Mon Aug 21 14:00:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
~ cat /etc/issue
Ubuntu 20.04.5 LTS \n \l
```

Behavior:
```sh
xdp git:(main) ✗ sudo ./xdp enp0s31f6
2023/10/19 16:56:15 Attached XDP program to iface "enp0s31f6" (index 2)
2023/10/19 16:56:15 Press Ctrl-C to exit and remove the program
2023/10/19 16:56:16 Map contents:
        <nil> => 1
        <nil> => 1
        <nil> => 61
2023/10/19 16:56:17 Map contents:
        <nil> => 1
        <nil> => 5
        <nil> => 2
        <nil> => 1
        <nil> => 164
```

**Expected behavior**
My expectation was to see the IP address string.
I found that if I use the uint32 type in `examples/xdp/main.go` to accept the `val` attribute, and use `internal.NativeEndian.PutUint32` to convert it to a byte array, I can solve the problem.
The following is my modification to the `formatMapContents` method in[ `examples/xdp/main.go`](https://github.com/cilium/ebpf/blob/main/examples/xdp/main.go), which can solve this problem.
```golang
func formatMapContents(m *ebpf.Map) (string, error) {
	var (
		sb strings.Builder
		key uint32
		val uint32
	)
	iter := m.Iterate()
	for iter.Next(&key, &val) {
		b := make([]byte, 4)
		internal.NativeEndian.PutUint32(b, key)
		sourceIP := net.IP(b) // IPv4 source address in network byte order.
		packetCount := val
		sb.WriteString(fmt.Sprintf("\t%s => %d\n", sourceIP, packetCount))
	}
	return sb.String(), iter.Err()
}
```
After modification, the problem was solved.
```shell
➜  xdp git:(main) ✗ sudo ./xdp enp0s31f6
2023/10/19 17:08:20 Attached XDP program to iface "enp0s31f6" (index 2)
[...]

PR #1178, opened 2023-10-19
example: fix key type of xdp example

BillyChen1:
The IP address output by [xdp demo](https://github.com/cilium/ebpf/tree/main/examples/xdp) is empty. Although it can be solved by directly changing the `key` type to `netip.Addr` after go1.18 version, I still use a more general way to fix it.
Issue: #1175

### DIFF
--- a/examples/xdp/main.go
+++ b/examples/xdp/main.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/link"
 )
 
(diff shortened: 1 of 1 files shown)


rgo3:
Thanks for opening a PR! Given that the library requires Go 1.20 to be the minimum version I don't see a real benefit in going with the more general way. Just using `netip.Addr` looks cleaner and would also work if someone wants to extend the example with ipv6 support.

BillyChen1:
@rgo3 done

rgo3 (approved):
LGTM!

I've noticed that this broke once we merged 4609dc73fa846184b379ae1ee771cc9e95e2ecbc and it's likely that fixing the internals would probably prevent us from making this change necessary. But switching to `netip.Addr` seems sensible either way as the only reason we didn't use it in the first place was that 1.18 wasn't the minimum required Go version.

lmb (approved):
Interesting, `netip.Addr` works as a key since [it implements `BinaryUnmarshaler`](https://pkg.go.dev/net/netip#Addr.UnmarshalBinary)!

PR #1180, opened 2023-10-20
sysenc: special case unmarshaling into []byte

lmb:
map: use t.Cleanup in createArray

sysenc: special case unmarshaling into []byte

    The library has somewhat complicated semantics when unmarshaling map keys
    and values. It turns out that the following was supported
    (but not documented) until the zero-allocation marshaling work broke it:

        var s []byte
       m.Lookup(key, &s)

    The important thing is that the slice here is empty. The current code treats
    this as "the key should be zero length" while originally we'd assign the
    temporary buffer used by the syscall to s as a way to reduce allocations. In
    hindsight this wasn't a great idea, but it is what it is.

    Reintroduce the byte slice special case.

    Fixes https://github.com/cilium/ebpf/issues/1175

### DIFF
--- a/internal/sysenc/marshal.go
+++ b/internal/sysenc/marshal.go
@@ -11,6 +11,8 @@ import (
 	"unsafe"
 
 	"github.com/cilium/ebpf/internal"
+
+	"golang.org/x/exp/slices"
 )
 
 // Marshal turns data into a byte slice using the system's native endianness.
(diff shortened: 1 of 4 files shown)


rgo3 (approved):

ti-mo:
Interesting, but couldn't we have left it the way it was and explicitly returned an error? The old behaviour is a little surprising; you give it a pointer to a slice (allocated by you!) and the lib replaces the pointer value? That's not ideal if you're expecting to read a value from the original slice's backing memory.

lmb:
I think we can still decide to do that, but it should be a deliberate choice i think.

Issue #1157, opened 2023-10-06
map: unmarshaling silenty ignores too small keys or values


ti-mo:
**Describe the bug**
Map.BatchLookup fails silently if nextKey is undersized, leaving nextKey with a zero value.

**To Reproduce**

```go
package main

import (
	"errors"
	"log"

	"github.com/cilium/ebpf"
)

func main() {
	m, err := ebpf.NewMap(&ebpf.MapSpec{
		Type:       ebpf.LRUHash,
		KeySize:    8,
		ValueSize:  8,
		MaxEntries: 1_024_000,
	})
	if err != nil {
		log.Fatal(err)
	}
	defer m.Close()

	keys, values := make([]uint64, m.MaxEntries()), make([]uint64, m.MaxEntries())
	for i := 0; uint32(i) < m.MaxEntries(); i++ {
		keys[i] = uint64(i)
		values[i] = uint64(i)
	}

	count, err := m.BatchUpdate(keys, values, nil)
	if err != nil {
		log.Fatalf("BatchUpdate: %s", err)
	}
	log.Println("wrote", count, "entries")

	// Undersized nextKey. Only uint64 should be accepted here. (KeySize 8)
	// Should contain 1048576 at the end of program. uint32 works as well, as it
	// can represent up to 4.2B.
	var nextKey uint16

	keysOut, valuesOut := make([]uint64, m.MaxEntries()), make([]uint64, m.MaxEntries())
	count, err = m.BatchLookup(nil, &nextKey, keysOut, valuesOut, nil)
	if err != nil && !errors.Is(err, ebpf.ErrKeyNotExist) {
		log.Fatalf("BatchLookup: %s", err)
	}

	log.Printf("count: %d, nextKey: %d", count, nextKey)
}
```

**Expected behavior**

BatchLookup should fail with an unmarshaling error.

devidasjadhav:
I can reproduced this.
Some of my findings:
1. https://github.com/cilium/ebpf/blob/3a926eb8a6eeced1aa8a79dbdd3851ea1c93cf2c/map.go#L919

nextKeyOut has no explicit type.

2. As we are passing generic pointer so we cannot test this against m.keySize. 

@ti-mo 
Do you have any pointers on this ?

lmb:
I'll see whether this was introduced by #1062

lmb:
[...]

PR #1181, opened 2023-10-20
sysenc: refuse to unmarshal into undersized dst

lmb:
Unmarshal currently doesn't check that unmarshaling actually consumes to full buffer. This means that looking uint16 from an uint32 value doesn't return an error.

Fixes https://github.com/cilium/ebpf/issues/1157


rgo3 (approved):
LGTM

PR #1182, opened 2023-10-22
btf: correct printing size of Int type in bytes

Asphaltt:
It seems OK that the printing size of Int type in bits.

But while printing a struct, like 'struct sk_buff', it seems strange that the printing size units of Int type and Struct/Union type are different:

```
  : offset=1088
  : Union[fields=2]
    size: 4
    csum: offset=0
    csum: Typedef:"__wsum"[Typedef:"__u32"]
      __wsum: Typedef:"__u32"[Int:"unsigned int"]
        __u32: Int:"unsigned int"[unsigned size=32]
    : offset=0
    : Struct[fields=2]
  priority: offset=1120
  priority: Typedef:"__u32"[Int:"unsigned int"]
    __u32: Int:"unsigned int"[unsigned size=32]
```

So, it's better to make the printing size of Int type in bytes same as Struct/Union type.

### DIFF
--- a/btf/types.go
+++ b/btf/types.go
@@ -117,7 +117,7 @@ type Int struct {
 }
 
 func (i *Int) Format(fs fmt.State, verb rune) {
-	formatType(fs, verb, i, i.Encoding, "size=", i.Size*8)
+	formatType(fs, verb, i, i.Encoding, "size=", i.Size)
 }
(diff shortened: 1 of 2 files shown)


lmb (approved):
Thanks :)

PR #1191, opened 2023-10-31
btf: fix CO-RE relocations for local type id

lmb:
At some point the kernel gained the ability to dynamically allocated typed memory via a bpf_obj_new helper (see [1]). This helper receives the ID of the type to allocate as an argument.

Trying to use this helper with the library will currently fail, since we took a short-cut when implementing on the fly BTF generation. At that point in time there was no use of the local_type_id relocation in the kernel, so we just stuck with the value present in ext_infos. This is of course not correct when building a BTF blob from scratch: type IDs are reallocated, and only used types are passed to the kernel in the first place.

Fix this by allocating an ID for any type that is the target of a local_type_id relocation.

1: https://lwn.net/Articles/915404/


btf/core.go
```diff
@@ -164,7 +164,7 @@ func (k coreKind) String() string {
 //
 // Fixups are returned in the order of relos, e.g. fixup[i] is the solution
 // for relos[i].
-func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder) ([]COREFixup, error) {
+func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder, typeID func(Type) (TypeID, error)) ([]COREFixup, error) {
```
ti-mo, line 167: Would it be wrong to make CORERelocate take a Builder here? I understand this way the changes are kept minimal, but it doesn't convey clearly what (potentially) happens behind the scenes, being adding a type to the BTF blob. AFAICS, we only pass `spec.TypeID()` here in tests.

prog.go
```diff
@@ -242,14 +242,24 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+	var b btf.Builder
+	if err := applyRelocations(insns, opts.KernelTypes, spec.ByteOrder, &b); err != nil {
+		return nil, fmt.Errorf("apply CO-RE relocations: %w", err)
 	}
-	if handle != nil {
-		defer handle.Close()
 
-		attr.ProgBtfFd = uint32(handle.FD())
+	if !b.Empty() && haveProgramExtInfos() != nil {
+		// Return ErrNotSupported instead of E2BIG if CO-RE relies on BTF
```
ti-mo, line 251: It's not clear what you mean by this. Doesn't CO-RE always rely on BTF being present?

prog.go
```diff
@@ -242,14 +242,24 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+		// being present.
+		return nil, haveProgramExtInfos()
```
ti-mo, line 253: `haveProgramExtInfos()` is not guaranteed to return `ErrNotSupported` here. I'd either make the surrounding if contain `errors.Is(haveProgramExtInfos(), ErrNotSupported)` or `return nil, ErrNotSupported` here.

prog.go
```diff
@@ -242,14 +242,24 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+	}
+
+	if haveProgramExtInfos() == nil {
```
ti-mo, line 256: Maybe it's time to move this to a function? `newProgram` is getting crowded.

ti-mo (approved):
Thanks! Few questions/nits.

prog.go
lmb: I'll rephrase. The issue is that a CO-RE local type id relocation might assert that a specific type exists in the BTF blob.

prog.go
lmb: Good point.

prog.go
lmb: I'll take another look!

lmb: Yes, it needs a refactor. But I don't have the time at the moment.

PR #1192, opened 2023-10-31
Add per-CPU Map Support to Batch Operations 

### LINKED ISSUE #251: Add per-CPU Map Support to Batch Operations
Currently, batch operations do not support per-CPU maps. This is not a technically difficult thing to accomplish, but deciding on the right interface within the library presents some challenges. [See this previous PR](https://github.com/cilium/ebpf/pull/229) attempt for details.

alxn:
Fixes #251

Follow up to #229 (and I took the test code from that diff, if @nathanjsweet want's to I'll back out the test change).

There's been significant underlying changes since the previous attempt.

I will patch this locally and check my end-to-end example works with this, but the existing tests are passing.


lmb (changes requested):
As the person that wrote the per cpu marshaling code: it's some of the gnarlier bits of the library. So thanks for wading through that!

After reading https://github.com/cilium/ebpf/pull/229#issuecomment-786734823 there are some open questions for me:

- You propose `BatchLookup(keys []T, values []T)` while per CPU lookup is `Lookup(keys T, values *[]T)`. This is something we need to resolve. Probably the right approach is to support `Lookup(values []T)` in addition to what we have, and not allow `*[]T` for batch lookups.
- Why did you go with `BatchLookup(values []T)` instead of `BatchLookup(values [][]T)`? The latter seems more ergonomic to use.
- Have you benchmarked your code? My worry with the per CPU stuff was always that it's so hideously expensive that batch or no batch doesn't make a difference.

map.go
```diff
@@ -935,13 +935,49 @@ func (m *Map) BatchLookupAndDelete(prevKey, nextKeyOut, keysOut, valuesOut inter
 	return m.batchLookup(sys.BPF_MAP_LOOKUP_AND_DELETE_BATCH, prevKey, nextKeyOut, keysOut, valuesOut, opts)
 }
 
+func (m *Map) batchLookupPerCPU(cmd sys.Cmd, count int, startKey, nextKeyOut, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
+	valueBuf := make([]byte, count*int(m.fullValueSize))
```
florianl, line 939: This might maybe not directly related to the proposed change, but to reduce the number if allocations it would be great if the user could provide this slice and potentially reuse it.

alxn:
I can indeed confirm that the CPU marshaling code is hard work!

I meant to follow up on [#229 (comment)](https://github.com/cilium/ebpf/pull/229#issuecomment-786734823)

> You propose BatchLookup(keys []T, values []T) while per CPU lookup is Lookup(keys T, values *[]T). This is something we need to resolve. Probably the right approach is to support Lookup(values []T) in addition to what we have, and not allow *[]T for batch lookups.

The batch APIs `BatchLookup` and `BatchLookupAndDelete` have the following comment:

```
"keysOut" and "valuesOut" must be of type slice, a pointer to a slice or buffer will not work.
```

With `BatchUpdate`:

```
"keys" and "values" must be of type slice, a pointer to a slice or buffer will not work.
```

Whereas looking at the `Lookup()` API:

```
var v uint32
m.Lookup(uint32(0), &v)

var slice []byte
m.Lookup(uint32(0), &slice)
```

I realize you obviously cannot write in-place to a scalar passed by value, as you can with an slice passed by value.

This shakes out to `lookupPerCPU` and `lookupAndDeletePerCPU` in:

```
// slicePtr must be a pointer to a slice.
func unmarshalPerCPUValue(slicePtr any, elemLength int, buf []byte) error {
```

So supporting `Lookup(values []T)` would mean:

```
var v uint32
m.Lookup(uint32(0), &v) // Lookup a single value

var v []uint32
m.Lookup(uint32(0), v) // Lookup a per-CPU value
```

I suspect the right answer would have been to only support `BatchLookup(keys []T, values *[]T)` originally but here we are.

> Why did you go with BatchLookup(values []T) instead of BatchLookup(values [][]T)? The latter seems more ergonomic to use.

I actually agree that `[batch][cpus]T` is a lot more ergonomic to use from a user perspective, but I was wary of changing the existing pattern.

```
var v []uint32
m.Lookup(uint32(0), v) // Lookup per-CPU

var (
﻿    nextKey uint32
    lookupKeys   = make([]uint32, 2)
    lookupValues = make([]uint32, 2)
)
m.BatchLookup(nil, &nextKey, lookupKeys, lookupValues, nil) // Lookup a Batch
```

Also because of pass by value, uses will have to do the following that we'll need to validate.

```
lookupValues := make([][]uint32, batchSize)
for i, _ := range lookupValues {
    lookupValues[i] = make([]uint32, possibleCPUs)
}
```

I also partly wonder if `[]T` makes more sense because it aligns with the BPF syscall.

> Have you benchmarked your code? My worry with the per CPU stuff was always that it's so hideously expensive that batch or no batch doesn't make a difference.

No not yet. I had assumed that crossing the kernel boundary was the most expensive part of the operation, I have some near production code I wanted to use this API on.

map.go
```diff
@@ -951,13 +972,23 @@ func (m *Map) batchLookup(cmd sys.Cmd, startKey, nextKeyOut, keysOut, valuesOut
...
+
+	if m.typ.hasPerCPUValue() {
+		possibleCPUs, err := internal.PossibleCPUs()
+		if err != nil {
+			return 0, err
+		}
+		count *= possibleCPUs
+	}
 	if count != valuesValue.Len() {
 		return 0, fmt.Errorf("keysOut and valuesOut must be the same length")
```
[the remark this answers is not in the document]
alxn: I made an edit, but I wonder if it's nicer to split the message if per cpu or not.

map.go
```diff
@@ -1008,34 +1035,31 @@ func (m *Map) batchLookup(cmd sys.Cmd, startKey, nextKeyOut, keysOut, valuesOut
...
-		return 0, ErrNotSupported
-	}
-	keysValue := reflect.ValueOf(keys)
-	if keysValue.Kind() != reflect.Slice {
-		return 0, fmt.Errorf("keys must be a slice")
-	}
-	valuesValue := reflect.ValueOf(values)
-	if valuesValue.Kind() != reflect.Slice {
-		return 0, fmt.Errorf("values must be a slice")
+	if err := haveBatchAPI(); err != nil {
```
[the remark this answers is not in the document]
alxn: ok I put it back, note that `batchLookup()` does it right at the start,

map_test.go
```diff
@@ -88,6 +88,97 @@ func TestMap(t *testing.T) {
 	}
 }
 
+func TestBatchPerCPUHash(t *testing.T) {
+	if err := haveBatchAPI(); err != nil {
+		t.Skipf("batch api not available: %v", err)
+	}
+
+	m, err := NewMap(&MapSpec{
+		Type:       PerCPUHash,
```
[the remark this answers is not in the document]
alxn: Yes, but there is a quirk.

`m.BatchLookup()` will now not return `ErrKeyNotExist` because it's not a "key", it's a value in an array. (This was a head scratcher for a while for me).

So that does mean we drop the testing of `batchLookupCmd()`
```
	sysErr = wrapMapError(sysErr)
	if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
		return 0, sysErr
	}
```

marshalers.go
```diff
@@ -43,68 +43,115 @@ func makeMapSyscallOutput(dst any, length int) sysenc.Buffer {
 	return sysenc.SyscallOutput(dst, length)
 }
 
-// marshalPerCPUValue encodes a slice containing one value per
+// marshalSlice encodes a slice containing one value per
 // possible CPU into a buffer of bytes.
 //
 // Values are initialized to zero if the slice has less elements than CPUs.
-func marshalPerCPUValue(slice any, elemLength int) (sys.Pointer, error) {
+func marshalSlice(slice any, elemLength, possibleCPUs, alignedElemLength int, buf []byte) error {
```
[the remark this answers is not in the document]
alxn: I'm finding it hard to grok.

Up in `marshalBatchPerCPUValue()` I need to allocated the contiguous buffer:
```
	alignedElemLength := internal.Align(elemLength, 8)
	stride := alignedElemLength * possibleCPUs
	result := make([]byte, batchLen*stride)
```

So then I want to end up with:
```
[batch0{cpu0, cpu1}, batch1{cpu0, cpu1}...]
```

If `appendPerCPUSlice()` returns a new buffer (expanded to fit the CPUs, then I have to copy that into the existing buffer.

There should be no case where I do not have enough buffer space ?

marshalers.go
```diff
@@ -123,7 +170,35 @@ func unmarshalPerCPUValue(slicePtr any, elemLength int, buf []byte) error {
...
-	reflect.ValueOf(slicePtr).Elem().Set(slice)
+	sliceValue := reflect.ValueOf(slice)
+	possibleCPUs, err := internal.PossibleCPUs()
+	if err != nil {
+		return err
+	}
+	stride := internal.Align(alignedElemLength, 8)
+	for i := 0; i < batchLen; i++ {
+		elem := sliceValue.Slice(i*possibleCPUs, (i+1)*possibleCPUs).Interface()
+		if err := unmarshalPerCPUValue(elem, elemLength, buf[:stride]); err != nil {
```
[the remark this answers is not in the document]
alxn: Wont this mean I overwrite the internals of slice that was passed by value ?

```
reflect.ValueOf(slicePtr).Elem().Set(slice)```

File: examples/kprobe_percpu/main.go [hunk 7]
```diff
@@ -50,14 +51,27 @@ func main() {
...
 
+	batch := true
 	for range ticker.C {
 		var all_cpu_value []uint64
-		if err := objs.KprobeMap.Lookup(mapKey, &all_cpu_value); err != nil {
-			log.Fatalf("reading map: %v", err)
+		if batch {
+			var nextKeyOut uint32
+			lookupKeys := make([]uint32, 2)
+			lookupValues := make([][]uint64, 2)
```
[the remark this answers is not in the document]
alxn: I think I just spaced out and didn't checkin that file after I'd been experimenting with the other style.

```
2023/11/03 19:24:00 Waiting for events..
2023/11/03 19:24:01 sys_execve called 0 times on CPU0
2023/11/03 19:24:01 sys_execve called 1 times on CPU1
2023/11/03 19:24:01 sys_execve called 0 times on CPU2
2023/11/03 19:24:01 sys_execve called 1 times on CPU3
2023/11/03 19:24:01
2023/11/03 19:24:02 sys_execve called 0 times on CPU0
2023/11/03 19:24:02 sys_execve called 2 times on CPU1
2023/11/03 19:24:02 sys_execve called 0 times on CPU2
2023/11/03 19:24:02 sys_execve called 2 times on CPU3
2023/11/03 19:24:02
2023/11/03 19:24:03 sys_execve called 1 times on CPU0
2023/11/03 19:24:03 sys_execve called 3 times on CPU1
2023/11/03 19:24:03 sys_execve called 0 times on CPU2
2023/11/03 19:24:03 sys_execve called 2 times on CPU3
2023/11/03 19:24:03
```

I still wonder if `[][]type` is better ergonomics.

map.go
alxn: It's nice that at least this buffer is only ever held on the stack so should be easy to collect?

alxn:
```
$ go test -exec sudo  . -bench BenchmarkPerCPUMarshalling
goos: linux
goarch: arm64
pkg: github.com/cilium/ebpf
BenchmarkPerCPUMarshalling/Lookup-4         	    1766	    677450 ns/op	   98304 B/op	    4096 allocs/op
BenchmarkPerCPUMarshalling/BatchLookup-4    	    5134	    238291 ns/op	   65600 B/op	    1029 allocs/op
PASS
ok  	github.com/cilium/ebpf	3.507s
```

alxn:
Also I believe I need to solve the issue of `libbpf_num_possible_cpus()`.

lmb:
I've given this some more thought and I think we need to expand the scope of the work to get per-CPU batch ops done. And because of the increased scope we'd have to split it into multiple PRs. Something like the following:

1. Support doing a per-CPU lookup into pre-allocated slices via `Lookup(values []T)`. This will involve figuring out an equivalent of `libbpf_num_possible_cpus()` as you point out. We'll also have to stop allocating slices via reflect in the marshal / unmarshal path. Instead we should always write into the backing buffer and then have special casing for `*[]T` in `Lookup`, etc. (or as close to the top of the stack as possible).
2. On top of clean per-CPU semantics for `[]T` add batch support for `[]T * n` (more likely) or `[][]T`.

I realise that isn't what you signed up for, sorry.

> So supporting `Lookup(values []T)` would mean:
> 
> ```
> var v uint32
> m.Lookup(uint32(0), &v) // Lookup a single value
> 
> var v []uint32
> m.Lookup(uint32(0), v) // Lookup a per-CPU value
> ```

It would be a bit more complicated / subtle: `m.Lookup(uint32(0), &v)` must also still work. Semantics:

- `m.Lookup` on a per cpu map with a `*[]T` will have the current behaviour: the lib reallocates the backing slice.
- `m.Lookup` on a per cpu map with `[]T` will only write into the pre-allocated buffer.

Otherwise the two should behave the same.

> I suspect the right answer would have been to only support BatchLookup(keys []T, values *[]T) originally but here we are.

Actually, requiring `*[]T` in Lookup was the mistake. Baking allocation decisions into an API always comes back to haunt you...

> I also partly wonder if `[]T` makes more sense because it aligns with the BPF syscall.

That is a good point. It only makes a difference if we were ever to directly pass the buffer to the syscall though, which is doable but pretty tricky. Overall `[]T` will be faster though, just less nice to use.

> Also I believe I need to solve the issue of `libbpf_num_possible_cpus()`.

We already have `internal.PossibleCPUs()` but that in practice is annoying to use since if forces error checking when allocating a slice.

```go
n, err := ebpf.PossibleCPUs()
// ...
buf := make([]uint32, n)
```

This is the original reason for taking `*[]T`.

Maybe the easiest would be to add two functions:

```
func PossibleCPUs() (int, error)
func MustPossibleCPUs() int
```

The latter would panic if an error occurred.

alxn:
`Lookup(values []T)` makes a lot of sense, I'll rework it so the `*[]T` is only dealt with right at the top of the stack, and then the rest of the code will assume a correctly sized slice. (i.e. this means top level validation that the slice is correct, and not needing to grow deep in the unmarshal).

I still wonder if `BatchLookup([][]T)` is better, I *again* got burnt by that in my integration code with this patch. It would be a little annoying for the users and validation though.


I assume for `PossibleCPUs()` that I'll pull the implementation up a level and rewrite the existing uses cases (rather than put a public `Must` decorator over the `internal`).

- It should be a `sync.Once` function as well I assume.

map_test.go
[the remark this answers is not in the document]
alxn: Yeah, I think I'll fix around this. I do realize the `Batch` API is one of those interesting dual return APIs that can be confusing, i.e. `count>0` and `error` is set.

There should be a case where `batch size == number of keys` and that should not return any error I think, it's only when we try to request one more that it fails.

alxn:
> I've given this some more thought and I think we need to expand the scope of the work to get per-CPU batch ops done. And because of the increased scope we'd have to split it into multiple PRs. Something like the following:
> 
>     1. Support doing a per-CPU lookup into pre-allocated slices via `Lookup(values []T)`. This will involve figuring out an equivalent of `libbpf_num_possible_cpus()` as you point out. We'll also have to stop allocating slices via reflect in the marshal / unmarshal path. Instead we should always write into the backing buffer and then have special casing for `*[]T` in `Lookup`, etc. (or as close to the top of the stack as possible).
> 

Ok these two are in:
* bcca828
* e1bf3cc

>     2. On top of clean per-CPU semantics for `[]T` 
> 
> 
> > So supporting `Lookup(values []T)` would mean:
> > ```
> > var v uint32
> > m.Lookup(uint32(0), &v) // Lookup a single value
> > 
> > var v []uint32
> > m.Lookup(uint32(0), v) // Lookup a per-CPU value
> > ```
> 
> It would be a bit more complicated / subtle: `m.Lookup(uint32(0), &v)` must also still work. Semantics:
> 
>     * `m.Lookup` on a per cpu map with a `*[]T` will have the current behaviour: the lib reallocates the backing slice.
> 
>     * `m.Lookup` on a per cpu map with `[]T` will only write into the pre-allocated buffer.
> 
> 
> Otherwise the two should behave the same.
> 

And now all of the above is true.

> add batch support for `[]T * n` (more likely) or `[][]T`.
>
> > I suspect the right answer would have been to only support BatchLookup(keys []T, values *[]T) originally but here we are.
> 
> Actually, requiring `*[]T` in Lookup was the mistake. Baking allocation decisions into an API always comes back to haunt you...
> 
> > I also partly wonder if `[]T` makes more sense because it aligns with the BPF syscall.
> 
> That is a good point. It only makes a difference if we were ever to directly pass the buffer to the syscall though, which is doable but pretty tricky. Overall `[]T` will be faster though, just less nice to use.
>
>
> Maybe we add an allocation function for a typed slice? func AllocBatch[T any, I constraints.Int](n I) []T or similar.

So I'm just looking at this support now:
```
BatchLookup(keys []T, values []T
MakeBatch[T any, I constraints.Int](n I) []T
```

- with an `Append()` style.

But I did notice that `MakeBatch()` needs to know if it's `PerCPU` or not (and then there's an `err` for `PossibleCPU()`, unless passed in.

alxn:
> One thing I wondered about when reading your summary: in your previous PR you added a test which does a `[]byte` lookup in a regular map (IIRC). What does it currently look like if you want to look up `[]byte` in a per CPU map without batch support? That might inform our decision of `[]T` vs `[][]T`.

I didn't add a `[]byte` lookup in #1220, that was already there in `TestMap()`:
```go
var slice []byte
qt.Assert(t, m.Lookup(uint32(0), &slice), qt.IsNil)
```

I added the slice lookup for the same data:

```go
sliceVal := make([]uint32, 1)
qt.Assert(t, m.Lookup(uint32(0), sliceVal), qt.IsNil)
```
I note that the `[]byte` array does return the correct response `[42 0 0 0]`


If I make this a `PerCPUArray` and try it

```go
const valueSize = 4
m, err := NewMap(&MapSpec{
	Type:       PerCPUArray,
	KeySize:    4,
	ValueSize:  valueSize,
	MaxEntries: 2,
})
...
v := make([]byte, valueSize*possibleCPUs)
```

Passing the slice by value:
```go
if err := m.Lookup(uint32(0), v); err != nil {
```
=>
```
Can't lookup 0: per-CPU slice has incorrect length, expected 4, got 16
```

Passing the pointer:
```
Can't lookup 0: cpu 0: unmarshaling *uint8 doesn't consume all data
```

Passing `make([]uint32, possibleCPUs)` works by ref or value.

I guess `[]byte` is interesting, and the `C` viewpoint that everything is just a `char *`. It probably makes no sense to support looking up a `uint32` array using a half length sized `uint64` array.

alxn:
I finally had enough contiguous time to figure out "append()" style.

Also quick question, is it possible to get a `buf` out of `sys.Pointer{}`... I hacked an API work around so that I could write an e2e test for the marshalers. (https://github.com/cilium/ebpf/commit/061c8b7c62d6e09260fef31ec69db9f151ded624)

alxn:
```
Error: marshalers.go:8:2: package slices is not in GOROOT (/opt/hostedtoolcache/go/1.20.11/x64/src/slices)
```
Ahem. I can manually write this I guess?

lmb:
> I can manually write this I guess?

Use `x/exp/slices`.

lmb:
> Also quick question, is it possible to get a `buf` out of `sys.Pointer{}`

Not 100% sure what you mean, but sys.Pointer has no length information. So I'd be reluctant to add API which turns it into a `[]byte` if that is what you mean.

alxn:
> > Also quick question, is it possible to get a `buf` out of `sys.Pointer{}`
> 
> Not 100% sure what you mean, but sys.Pointer has no length information. So I'd be reluctant to add API which turns it into a `[]byte` if that is what you mean.

I'd written a full e2e in (https://github.com/cilium/ebpf/commit/061c8b7c62d6e09260fef31ec69db9f151ded624), but `marshalBatchPerCPUValue()` returns `sys.Pointer{buf}`... which I can't pass to `unmarshalBatchPerCPUValue()`.

map.go
alxn:
```
map.go-1007-func (m *Map) batchLookup(cmd sys.Cmd, cursor *BatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
...
map.go:1044:    valueBuf := make([]byte, count*int(m.fullValueSize))
map.go:1045:    valuePtr := sys.NewSlicePointer(valueBuf)
```
```
./map.go:1044:18: make([]byte, count * int(m.fullValueSize)) escapes to heap
```
```
map.go-1059-func (m *Map) batchLookupPerCPU(cmd sys.Cmd, count int, inBatch []byte, cursor *BatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
map.go:1060:    valueBuf := make([]byte, count*int(m.fullValueSize))
map.go:1061:    valuePtr := sys.NewSlicePointer(valueBuf)
```
```
./map.go:1060:18: make([]byte, count * int(m.fullValueSize)) escapes to heap
```

lmb:
Sorry, turns out that batch lookup was just thoroughly broken so far :(

lmb:
(I removed the `@lmb` from the commit title because GH keeps pinging me every time you push a commit)

lmb:
Seems like your commits follow chronological rather than logical order, is it ok to squash them?

alxn:
> Seems like your commits follow chronological rather than logical order, is it ok to squash them?

The first few ones are logical, but then it's chronological after.

Yeah I'd assume this would be a squash land.

alxn:
I'll squash/rebase

map_test.go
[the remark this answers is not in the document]
lmb: Take a look at the rewritten `TestMapBatch`: that now uses a `contents` map to assert that we get the correct values out. We don't test the order anymore, but for hashes that isn't defined and for arrays I'm not bothered enough to care. Net result is that the test is a lot more compact while testing both array and hash.

map_test.go
```diff
@@ -1910,38 +1998,73 @@ func BenchmarkMarshaling(b *testing.B) {
...
 		})
 		if err != nil {
 			b.Fatal(err)
 		}
 		return m
 	}
 
-	key := uint64(1)
-	val := []uint64{1, 2, 3, 4, 5, 6, 7, 8}
+	possibleCPUs, err := PossibleCPU()
```
lmb: `MustPossibleCPU`?

map_test.go
```diff
@@ -1910,38 +1998,73 @@ func BenchmarkMarshaling(b *testing.B) {
...
+	val := make([]uint64, possibleCPUs)
+	for i := range val {
+		val[i] = uint64(i + 1)
+	}
 
 	m := newMap(8)
-	if err := m.Put(key, val[0:]); err != nil {
-		b.Fatal(err)
+	for i := 0; i < batch; i++ {
+		key := uint64(i + 1)
```
lmb, line 2020: I know this is in the original code as well, but why do we start at `1`?

map_test.go
```diff
@@ -1910,38 +1998,73 @@ func BenchmarkMarshaling(b *testing.B) {
...
+				err := m.Lookup(unsafe.Pointer(&key), &value)
+				if err != nil {
+					b.Fatal("Can't get key:", err)
+				}
+			}
+		}
+	})
+	b.Run("BatchLookup", func(b *testing.B) {
+		b.ReportAllocs()
+		b.ResetTimer()
```
lmb, line 2044: Should this happen after allocating the buffers below?

alxn:
Forced, benchmark run:
```
goos: linux
goarch: arm64
pkg: github.com/cilium/ebpf
BenchmarkPerCPUMarshalling/Lookup-4         	    1722	    684530 ns/op	   98304 B/op	    4096 allocs/op
BenchmarkPerCPUMarshalling/BatchLookup-4    	    4694	    248713 ns/op	   65601 B/op	    1029 allocs/op
PASS
ok  	github.com/cilium/ebpf	3.366s```

File: map_test.go [hunk 11]
```diff
@@ -1920,27 +2009,54 @@ func BenchmarkPerCPUMarshalling(b *testing.B) {
...
+		for i := 0; i < b.N; i++ {
+			count, err := m.BatchLookup(&BatchCursor{}, keys, values, &BatchOptions{})
+			if count <= 0 {
+				if err != nil {
+					b.Fatal("Can't get batch:", err)
+				}
+			}
+			if count != batch {
+				b.Fatalf("Can't get batch: %d", count)
 			}
```
lmb, line 2060: I think this benchmark is now comparing two separate things: a direct lookup via m.Lookup and iteration via (the now badly named) `BatchLookup`. The correct thing to compare is `m.Iterate()` and `BatchLookup`, see `BenchmarkIterate`. My suggestion would be to copy that instead of modifying `BenchmarkPerCPUMarshalling`. (See also https://github.com/cilium/ebpf/pull/1254/commits/df14fdedddb5a0365781630efc588c0a690d4a0e) Crucial point: you always need to check for ErrKeyNotExist to figure out whether you are done iterating.

lmb (changes requested):
Thanks for chipping away at this and sorry that the code has been changing so much. It turns out that the BatchLookup we had in the library was pretty thoroughly broken, and that it's also a pretty poor API from the kernel side. Thoughts on the PR:

* Can you split the per-CPU an regular batch lookup paths as much as possible? In the past the regular and per-CPU code was integrated tightly and that made it hard to follow. The current approach has the same problem, for example it took me a while to figure out that `possibleCPU` is rolled into `count` in `batchLookupPerCPU`. See `m.Update`, `m.Lookup` for what I mean: there is a condition on `hasPerCPUValue` which calls into the per-CPU implementation.
* Can you add some tests which exercise the error paths in per cpu marshaling and unmarshaling? I'm thinking small slices, etc. From reading the code there is one panic we're not catching and I'm also not sure that values get unmarshaled correctly.
* The way to do roundtrip tests for the per CPU marshalers / unmarshalers is to return `[]byte` instead of `sys.Pointer`.

I'm still on the fence about `[][]T` vs. `[]T` btw. but that is something we can figure out at a later point I think. Needs some more thinking from my side.

alxn:
> The way to do roundtrip tests for the per CPU marshalers / unmarshalers is to return []byte instead of sys.Pointer.

Yeah - that's the question I was asking on https://github.com/cilium/ebpf/commit/061c8b7c62d6e09260fef31ec69db9f151ded624

I'll prequel that first.

map.go
```diff
@@ -1077,7 +1123,7 @@ func (m *Map) batchLookup(cmd sys.Cmd, cursor *BatchCursor, keysOut, valuesOut i
 	if err != nil {
 		return 0, err
 	}
-	err = sysenc.Unmarshal(valuesOut, valueBuf)
+	err = sysenc.Unmarshal(keysOut, keyBuf)
```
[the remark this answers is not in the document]
alxn: This looks like a merge conflict gone wrong.

map.go
```diff
@@ -1045,13 +1081,23 @@ func (m *Map) batchLookup(cmd sys.Cmd, cursor *BatchCursor, keysOut, valuesOut i
...
+		count *= possibleCPUs
+	}
 	if count != valuesValue.Len() {
-		return 0, fmt.Errorf("keysOut and valuesOut must be the same length")
+		return 0, fmt.Errorf("keysOut and valuesOut must be the correct relative lengths")
 	}
+	return keysValue.Len(), nil
+}
+
+func (m *Map) batchLookupCmd(cmd sys.Cmd, count int, inBatch []byte, cursor *BatchCursor, keysOut interface{}, valuePtr sys.Pointer, opts *BatchOptions) (int, error) {
```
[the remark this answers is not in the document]
alxn: Yeah this can shuffle out via `batchLookupPerCPU()` too.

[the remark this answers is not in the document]
alxn: It still needs `m.fd.Uint()` but that could come as a parameter.

marshalers.go
```diff
@@ -43,41 +44,83 @@ func makeMapSyscallOutput(dst any, length int) sysenc.Buffer {
...
-			return sys.Pointer{}, err
+			return nil, err
 		}
 
-		offset := i * alignedElemLength
-		elemBytes.CopyTo(buf[offset : offset+elemLength])
+		b := make([]byte, alignedElemLength)
+		elemBytes.CopyTo(b) // Only copies elemLength
+		buf = append(buf, b...)
+	}
```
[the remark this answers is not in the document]
alxn: I think I can just do the slice expansion after ? 

https://go.dev/play/p/yd1cw6Pr3pf

```
	// Ensure buf is full size.
	buf = buf[:alignedElemLength*possibleCPUs]```

File: marshalers.go [hunk 15]
```diff
@@ -43,41 +44,83 @@ func makeMapSyscallOutput(dst any, length int) sysenc.Buffer {
...
+		return sys.Pointer{}, err
+	}
+
+	return sys.NewSlicePointer(buf), nil
+}
+
+// marshalBatchPerCPUValue encodes a batch-sized slice of slices containing
+// one value per possible CPU into a buffer of bytes.
+//
+// Values are initialized to zero if the slice has less elements than CPUs.
```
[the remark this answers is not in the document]
alxn: > // Per-batch values are initialized to zero if that sub slice has less elements than CPUs.

map_test.go
alxn: I like to ensure non-zero values for testing.

map_test.go
alxn: Same for
```
		var value []uint64
```
in the previous, but generally should be insignificant compared to `b.N`

map_test.go
alxn: Because mostly the syscall cost is likely to be the high cost.

alxn:
```
BenchmarkPerCPUMarshalling/Lookup-4         	    1740	    677886 ns/op	   98304 B/op	    4096 allocs/op
BenchmarkPerCPUMarshalling/Iterate-4        	    1083	   1098589 ns/op	   90210 B/op	    3076 allocs/op
BenchmarkPerCPUMarshalling/BatchLookup-4    	    4807	    252948 ns/op	   65592 B/op	    1029 allocs/op```

Comment (alxn) [reply map_test.go:2060]: ok I can copy/attempt refactor of `BenchmarkIterate`

**Comment by lmb** [comment]
More conflicts... let me know if you want me to help with rebasing.

**Comment by alxn** [comment]
Rebasing is ok, for reasons I really don't want to go into again I spent the first 7 years of my career doing a *lot* of painful rebasing.

I think I'm seeing a flake on `main` though:
```
$ go test -exec sudo  . -count 100 | tee /tmp/base
--- FAIL: TestMapBatch (0.00s)
    --- FAIL: TestMapBatch/Hash (0.00s)
        quicktest.go:12:
            error:
              values are not equal
            got:
              int(1)
            want:
              int(2)
            stack:
              /Users/alun/Code/github.com/cilium/ebpf.git/map_test.go:175
                qt.Assert(t, qt.Equals(count, len(lookupKeys)))

--- FAIL: TestMapBatch (0.01s)
    --- FAIL: TestMapBatch/Hash (0.01s)
        quicktest.go:12:
            error:
              values are not equal
            got:
              int(1)
            want:
              int(2)
            stack:
              /Users/alun/Code/github.com/cilium/ebpf.git/map_test.go:175
                qt.Assert(t, qt.Equals(count, len(lookupKeys)))

FAIL
FAIL	github.com/cilium/ebpf	84.432s
FAIL
```

**Comment by alxn** [comment]
https://github.com/cilium/ebpf/pull/1259

I cannot understand this one.

**Comment by alxn** [comment]
```
goos: linux
goarch: arm64
pkg: github.com/cilium/ebpf
BenchmarkIterate/Hash/MapIterator-4         	    1346	    895241 ns/op	   24120 B/op	    1005 allocs/op
BenchmarkIterate/Hash/MapIteratorDelete-4   	     782	   1542468 ns/op	     104 B/op	       4 allocs/op
BenchmarkIterate/Hash/BatchLookup-4         	  428934	      2814 ns/op	      56 B/op	       3 allocs/op
BenchmarkIterate/Hash/BatchLookupAndDelete-4         	   22102	     54334 ns/op	      64 B/op	       3 allocs/op
BenchmarkIterate/Hash/BatchDelete-4                  	   22294	     54014 ns/op	      24 B/op	       1 allocs/op
BenchmarkIterate/PerCPUHash/MapIterator-4            	    1140	   1062411 ns/op	   56121 B/op	    2005 allocs/op
BenchmarkIterate/PerCPUHash/MapIteratorDelete-4      	     675	   1758026 ns/op	   88109 B/op	    3004 allocs/op
BenchmarkIterate/PerCPUHash/BatchLookup-4            	    6459	    188635 ns/op	   56824 B/op	    1004 allocs/op
BenchmarkIterate/PerCPUHash/BatchLookupAndDelete-4   	    4712	    246346 ns/op	   56832 B/op	    1004 allocs/op
BenchmarkIterate/PerCPUHash/BatchDelete-4            	   19084	     59482 ns/op	      24 B/op	       1 allocs/op
PASS
```

**Comment by lmb** [comment]
@alxn please ping me when you want another review.

**Comment by alxn** [comment]
> Can you split the per-CPU an regular batch lookup paths as much as
> possible?

There's many ways to take this.

- separate API (a `NewPerCPUMap()`)
- different internal implementations decorated via an interace into the concrete `Map` struct.
- some common checking then a hard `if/else` fork. (Though idomatic Go prefers `if+return`.)

> In the past the regular and per-CPU code was integrated tightly and
> that made it hard to follow. The current approach has the same
> problem, for example it took me a while to figure out that possibleCPU
> is rolled into count in batchLookupPerCPU. See m.Update, m.Lookup for
> what I mean: there is a condition on hasPerCPUValue which calls into
> the per-CPU implementation.

It's not actually rolled into count:

`batchLookupCount()` still returns the key length `keysValue.Len()`, which is the same regardless of per cpu or not.

The quirk is that values is either `==` or meant to be per cpu.

> Can you add some tests which exercise the error paths in per cpu
> marshaling and unmarshaling? I'm thinking small slices, etc. From
> reading the code there is one panic we're not catching and I'm also
> not sure that values get unmarshaled correctly.
>
> The way to do roundtrip tests for the per CPU marshalers /
> unmarshalers is to return []byte instead of sys.Pointer.

Yeah, I had locally done the `buf` change to add `TestMarshalUnmarshalBatchPerCPUValue()`.

By small slices, did you mean small batches? Or smaller than `uint32` ?

**Comment by alxn** [comment]
I'm surprised to see this on ARM

```
 === FAIL: . TestMapBatch/PerCPUArray (0.01s)
    quicktest.go:12: 
        error:
          got non-nil value
        got:
          e"batch update: not supported"
        stack:
          /home/runner/actions-runner/_work/ebpf/ebpf/map_test.go:148
            qt.Assert(t, qt.IsNil(err))
```

**Comment by lmb** [comment]
I don't think the failure is arm specific: https://github.com/cilium/ebpf/actions/runs/7176921150/job/19542631413#step:7:292 I haven't the foggiest why adding `SkipIfNotSupported` helps here though, since that just checks for batch update on `Hash`.

File: marshalers.go [hunk 16]
```diff
@@ -43,44 +44,93 @@ func makeMapSyscallOutput(dst any, length int) sysenc.Buffer {
...
-	if err != nil {
-		return sys.Pointer{}, err
+		return nil, errors.New("per-CPU value requires slice")
 	}
 
 	sliceValue := reflect.ValueOf(slice)
 	sliceLen := sliceValue.Len()
-	if sliceLen > possibleCPUs {
-		return sys.Pointer{}, fmt.Errorf("per-CPU value exceeds number of CPUs")
+	if sliceLen < possibleCPUs {
```
lmb, line 59: Reading through this again, makes me realise that this is another special case like the one you fixed in https://github.com/cilium/ebpf/commit/bcca828832442d795250ef47840514b81770d281 I'll take a look and see whether I can move the logic so that batch lookups don't inherit this.

examples/kprobe_percpu/main.go
lmb, line 1: Meta comment: I want to discourage the use of BatchLookup, do you mind if we don't change the example?

lmb (changes requested):
I'm going to push two commits:

- Reorder your map.go changes to have a smaller diff and follow the structure i mentioned. batchLookupCount is gone since it's not necessary.
- Small tweak to marshalers.go. fullValueSize is easily recalculated in unmarshalPerCPUValue. This allows minimising the places we need to care about the weird per CPU padding.

marshalers.go
```diff
@@ -43,44 +44,93 @@ func makeMapSyscallOutput(dst any, length int) sysenc.Buffer {
...
 		elemBytes, err := sysenc.Marshal(elem, elemLength)
 		if err != nil {
-			return sys.Pointer{}, err
+			return nil, err
 		}
 
-		offset := i * alignedElemLength
-		elemBytes.CopyTo(buf[offset : offset+elemLength])
+		b := make([]byte, alignedElemLength)
+		elemBytes.CopyTo(b) // Only copies elemLength
```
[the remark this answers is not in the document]
alxn: Oops, I accepted this but it doesn't zeropad the tail for alignment.

alxn:
There are a lot of changes between [141ec75](https://github.com/cilium/ebpf/pull/1192/commits/141ec75f7080b385a8e8f3d73e183f0baba7583e) and a31665caaa9a524a571d319321348eb5ec02bded that I'm not sure if I understand.

alxn:
Oh I see:
```
commit 141ec75f7080b385a8e8f3d73e183f0baba7583e
Author: Alun Evans <alun@badgerous.net>
Date:   Tue Dec 12 16:14:32 2023 +0000

    map: Add BatchPerCPU Support

    Signed-off-by: Alun Evans <alun@badgerous.net>

 examples/kprobe_percpu/main.go |  18 ++++++++-
 map_test.go                    | 319 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------------------
 marshalers.go                  | 109 +++++++++++++++++++++++++++++++++++++++++++--------
 marshalers_test.go             |  24 ++++++++++++
 4 files changed, 311 insertions(+), 159 deletions(-)
```

it completely dropped all the `map` changes.

marshalers.go
alxn: Yeah I think I got this change the wrong way during testing.

If the slice len is less than possible cpus (orig), we could/can zero pad. Honestly I think it should probably `!=` but there are tests that assume zero pad that I broke with this change so reverting.

marshalers.go
```diff
@@ -123,6 +173,33 @@ func unmarshalPerCPUValue(slice any, elemLength int, buf []byte) error {
...
+	if sliceType.Kind() != reflect.Slice {
+		return fmt.Errorf("batch requires a slice")
+	}
 
+	sliceValue := reflect.ValueOf(slice)
+	possibleCPUs, err := PossibleCPU()
+	if err != nil {
+		return err
+	}
+	if sliceValue.Len() < batchLen*possibleCPUs {
```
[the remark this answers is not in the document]
alxn: I think it should all be `:=`

lmb:
> it completely dropped all the `map` changes.

Ah sorry, that wasn't my intention!

marshalers.go
lmb: Yes, we'd still have to zero pad, but maybe I can make it zero pad somewhere else xD

marshalers.go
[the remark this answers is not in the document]
lmb: Ah I see, let me try something.

map_test.go
```diff
@@ -99,45 +99,61 @@ func TestMapBatch(t *testing.T) {
...
+	defer hashPerCpu.Close()
 
-	for _, m := range []*Map{array, hash} {
+	arrayPerCpu := mustNewMap(PerCPUArray, arrayMax)
+	defer arrayPerCpu.Close()
+
+	for _, m := range []*Map{array, hash, arrayPerCpu, hashPerCpu} {
 		t.Run(m.Type().String(), func(t *testing.T) {
+			if m.Type() == PerCPUArray {
+				testutils.SkipOnOldKernel(t, "5.15", "BatchUpdate PerCPU Array")
```
lmb, line 136: Can you add a link to the upstream commit where support was added? This can go into the commit message as well.

map_test.go
```diff
@@ -2036,140 +2063,152 @@ func BenchmarkMap(b *testing.B) {
...
+			ValueSize:  8,
+			MaxEntries: 1000,
+		})
+		if err != nil {
+			b.Fatal(err)
+		}
+		b.Cleanup(func() {
+			m.Close()
+		})
+		prefix := m.Type().String() + "/"
```
lmb, line 2079: Why not another level of `b.Run` nesting instead of manually adding `prefix`? Then filtering via `-bench` will also work correctly.

map_test.go
```diff
@@ -2036,140 +2063,152 @@ func BenchmarkMap(b *testing.B) {
...
-			}
-			b.StartTimer()
+			b.ReportAllocs()
+			b.ResetTimer()
 
-			iter := m.Iterate()
-			for iter.Next(&k, &v) {
-				if err := m.Delete(&k); err != nil {
+			for i := 0; i < b.N; i++ {
+				j := 0
```
lmb, line 2109: Why introduce `j` here?

lmb (changes requested):
Okay, implementation LGTM! I've left some small comments, squashed and added a fix of mine on top. PTAL. Final ask from my side:

* Can you write a commit message? Some things I would find interesting: `[]T` vs. `[][]T`. The behaviour change of batchdelete for per CPU arrays in 5.15.
* Can you add some tests to cover corner cases of the per cpu marshalers: slices that have smaller or larger length than what the code expects.

lmb:
Made https://github.com/cilium/ebpf/issues/1270 and https://github.com/cilium/ebpf/issues/1271

map_test.go
alxn:
```
				testutils.SkipOnOldKernel(t, "5.10", "BatchUpdate PerCPU Array")
				testutils.SkipOnOldKernel(t, "5.15", "BatchUpdate PerCPU Array")```
Ah I went by the versions were testing against, it's actually:

https://github.com/torvalds/linux/commit/f008d732ab181fd00d95c2e8a6e479d2f7c634b3 in 5.13.

File: map_test.go (hunk 21 above)
Comment (alxn) [reply map_test.go:2109]: oops some printf debugging right there.

**Comment by alxn** [comment]
>     * Can you add some tests to cover corner cases of the per cpu marshalers: slices that have smaller or larger length than what the code expects.

found you one panic, added this check in unmarshal :
```
	fullValueSize := possibleCPUs * internal.Align(elemLength, 8)
	if len(buf) != batchLen*fullValueSize {
		return fmt.Errorf("input buffer has incorrect length, expected %d, got %d",
			len(buf), batchLen*fullValueSize)
	}
```

alxn:
@lmb one question - I did switch the checks to `!=` rather than magical auto padding. Updated the commit comment with that as well.

lmb:
> I did switch the checks to `!=` rather than magical auto padding. Updated the commit comment with that as well.

You mean for batch per-CPU marshaling? Yes, I think that is the right thing to do.

lmb (approved):

lmb:
Thanks for all your hard work!

Issue #1188, opened 2023-10-30
asm: add support for ISA v4


dylandreimerink:
There has been a new addition to the eBPF ISA v4. This update adds 7 new instructions. We should add support for these to our `asm` package so we can handle programs created with this new ISA.

References:
* The original proposal https://lore.kernel.org/bpf/4bfe98be-5333-1c7e-2f6d-42486c8ec039@meta.com/
* The LLVM change https://reviews.llvm.org/D144829
* Kernel changes https://lore.kernel.org/all/20230728011156.3711870-1-yonghong.song@linux.dev/

The LLVM changes will be part of v18 which will be a few month yet, and the linux kernel changes are part of v6.6.

dylandreimerink:
I took and assembled the example instructions:
```
// CHECK: d7 01 00 00 10 00 00 00	r1 = bswap16 r1
// CHECK: d7 02 00 00 20 00 00 00	r2 = bswap32 r2
// CHECK: d7 03 00 00 40 00 00 00	r3 = bswap64 r3
r1 = bswap16 r1
r2 = bswap32 r2
r3 = bswap64 r3

// CHECK: 91 41 00 00 00 00 00 00	r1 = *(s8 *)(r4 + 0x0)
// CHECK: 89 52 04 00 00 00 00 00	r2 = *(s16 *)(r5 + 0x4)
// CHECK: 81 63 08 00 00 00 00 00	r3 = *(s32 *)(r6 + 0x8)
r1 = *(s8 *)(r4 + 0)
r2 = *(s16 *)(r5 + 4)
r3 = *(s32 *)(r6 + 8)

// CHECK: 91 41 00 00 00 00 00 00	r1 = *(s8 *)(r4 + 0x0)
// CHECK: 89 52 04 00 00 00 00 00	r2 = *(s16 *)(r5 + 0x4)
r1 = *(s8 *)(r4 + 0)
r2 = *(s16 *)(r5 + 4)

// CHECK: bf 41 08 00 00 00 00 00	r1 = (s8)r4
// CHECK: bf 52 10 00 00 00 00 00	r2 = (s16)r5
// CHECK: bf 63 20 00 00 00 00 00	r3 = (s32)r6
r1 = (s8)r4
r2 = (s16)r5
r3 = (s32)r6

// CHECK: bc 31 08 00 00 00 00 00	w1 = (s8)w3
// CHECK: bc 42 10 00 00 00 00 00	w2 = (s16)w4
w1 = (s8)w3
[...]

PR #1193, opened 2023-11-01
Add new ISAv4 instructions

dylandreimerink:
The [proposal](https://lore.kernel.org/bpf/4bfe98be-5333-1c7e-2f6d-42486c8ec039@meta.com/) for a new version of the eBPF instruction set has been accepted and merged in both LLVM 18 and linux kernel v6.6. So we might be seeing these new instructions in the near future.

This PR adds support for these new instructions. Before this PR we could still parse the instructions but did not recognize the variations added:

```
myfunc:
         0: SwapLE dst: r1 imm: 16
         1: SwapLE dst: r2 imm: 32
         2: SwapLE dst: r3 imm: 64
         3: LdXMode(128)B 
         4: LdXMode(128)H 
         5: LdXMode(128)W 
         6: LdXMode(128)B 
         7: LdXMode(128)H 
         8: MovReg dst: r1 src: r4
         9: MovReg dst: r2 src: r5
        10: MovReg dst: r3 src: r6
        11: Mov32Reg dst: r1 src: r3
        12: Mov32Reg dst: r2 src: r4
        13: InvalidJumpOp32Imm dst: r0 off: 0 imm: 4
        14: DivReg dst: r1 src: r3
        15: ModReg dst: r2 src: r4
        16: Div32Reg dst: r1 src: r3
        17: Mod32Reg dst: r2 src: r4
```

After this PR we correctly understand the new instructions:

```
myfunc:
        0: BSwap16 dst: r1 
        1: BSwap32 dst: r2 
        2: BSwap64 dst: r3 
        3: LdXMemSXB dst: r1 src: r4 off: 0 imm: 0
        4: LdXMemSXH dst: r2 src: r5 off: 4 imm: 0
        5: LdXMemSXW dst: r3 src: r6 off: 8 imm: 0
        6: LdXMemSXB dst: r1 src: r4 off: 0 imm: 0
        7: LdXMemSXH dst: r2 src: r5 off: 4 imm: 0
        8: SMov8Reg dst: r1 src: r4
        9: SMov16Reg dst: r2 src: r5
        10: SMov32Reg dst: r3 src: r6
        11: SMov8Reg32 dst: r1 src: r3
        12: SMov16Reg32 dst: r2 src: r4
        13: Ja32 imm: 4
        14: SDivReg dst: r1 src: r3
        15: SModReg dst: r2 src: r4
        16: SDivReg32 dst: r1 src: r3
        17: SModReg32 dst: r2 src: r4
```

Fixes: #1188


lmb (changes requested):
Thanks for putting this together. I think that my main concern with this approach is that it introduces so many special cases. A lot of the getters on OpCode actually don't work correctly for extended opcodes, something we've not tackled with this at all (documentation or code wise).

So far `OpCode` is a 1:1 representation of the BPF wire format. This made sense since not all of the space was allocated. This is important for `InvalidOpCode` to function correctly. Now that pretty much all space is gone there is another option we can try: divorce the representation of OpCode from the wire format. I'm thinking something along these lines:

```patch
diff --git a/asm/opcode.go b/asm/opcode.go
index 845c5521..04d1d6cf 100644
--- a/asm/opcode.go
+++ b/asm/opcode.go
@@ -66,18 +66,33 @@ func (cls Class) isJumpOrALU() bool {
 	return cls.IsJump() || cls.IsALU()
 }
 
-// OpCode is a packed eBPF opcode.
+// OpCode represents a single operation.
 //
-// Its encoding is defined by a Class value:
+// It is not a 1:1 mapping to real eBPF opcodes.
 //
-//	msb      lsb
-//	+----+-+---+
-//	| ???? |CLS|
-//	+----+-+---+
-type OpCode uint8
+// The encoding varies based on a 3-bit Class:
+//
+//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
+//	           ???           | CLS
+//
+// For ALUClass and ALUCLass32:
+//
+//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
+//	          OPC          |S| CLS
+//
+// For LdClass, LdXclass, StClass and StXClass:
+//
+//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
+//	       0       | MDE |SIZ| CLS
+//
+// For JumpClass, Jump32Class:
+//
+//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
+//	       0       |  OPC  |S| CLS
+type OpCode uint16
 
 // InvalidOpCode is returned by setters on OpCode
-const InvalidOpCode OpCode = 0xff
+const InvalidOpCode OpCode = 0xffff
 
 // rawInstructions returns the number of BPF instructions required
 // to encode this opcode.
```

Basically, make OpCode wider so that we can represent all the logical opcodes. For simplicity we'd also change the various `*Op` to be uint16. We'd then handle converting to and from OpCode in `Instruction.Marshal` and `Unmarshal`, similar to what we do for dword loads. This way `ExtendedALUOp` is not required anymore and SMod / SDiv can be ALUOp. Finally we can also guarantee that `InvalidOpCode` really is invalid.

This is basically free at runtime since we don't increase the size of `Instruction`. I even think that we can arrange it so that existing users won't break (besides having to throw in a uint16 cast maybe).

What do you think?

cmd/test/main.go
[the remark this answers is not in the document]
dylandreimerink: Oeps, that should not have been in there

dylandreimerink:
> Basically, make OpCode wider so that we can represent all the logical opcodes. For simplicity we'd also change the various *Op to be uint16. We'd then handle converting to and from OpCode in Instruction.Marshal and Unmarshal, similar to what we do for dword loads. This way ExtendedALUOp is not required anymore and SMod / SDiv can be ALUOp. Finally we can also guarantee that InvalidOpCode really is invalid.
>
>This is basically free at runtime since we don't increase the size of Instruction. I even think that we can arrange it so that existing users won't break (besides having to throw in a uint16 cast maybe).
>
>What do you think?

Yes, that feels less hacky than what I have so far :+1:

asm/instruction.go
```diff
@@ -60,6 +60,13 @@ func (ins *Instruction) Unmarshal(r io.Reader, bo binary.ByteOrder) (uint64, err
 	}
 
 	ins.Offset = int16(bo.Uint16(data[2:4]))
+
+	// For ALU Ops, the lower 8 bits of the offset field are part of the opcode
+	// within our ASM representation.
+	if ins.OpCode.Class().IsALU() {
```
lmb, line 66: I think I'd be more explicit here: if we're dealing with Mod or Div, check whether offset is 1. If it is change the opcode. Otherwise we may end up importing

asm/instruction.go
```diff
@@ -60,6 +60,13 @@ func (ins *Instruction) Unmarshal(r io.Reader, bo binary.ByteOrder) (uint64, err
...
+		ins.OpCode |= OpCode((ins.Offset & 0x00ff) << 8)
```
lmb, line 67: We should also set offset to 0 here.

asm/instruction_test.go
```diff
@@ -339,3 +339,97 @@ type testFDer int
...
+	}
+
+	var insns Instructions
+	err := insns.Unmarshal(bytes.NewReader(rawInsns), binary.LittleEndian)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	lines := []string{
+		"BSwap16 dst: r1 ",
```
lmb, line 379: It's nice to be able to see what the insns end up formatting as, but I'm a bit on the fence about testing the actual format. It's not something we want to guarantee stability for right?

Another option would be to just `t.Log` the strings and assert some things like `Swap` contains the number `16` or something like that.

Not super bothered about this though, up to you.

asm/jump.go
```diff
@@ -103,13 +103,22 @@ func (op JumpOp) Reg32(dst, src Register, label string) Instruction {
...
 	}
 
 	return OpCode(class).SetJumpOp(op).SetSource(source)
 }
 
+// LongJump returns a jump always instruction with a range of [-2^31, 2^31 - 1].
+func LongJump(label string) Instruction {
+	return Instruction{
+		OpCode:   Ja.opCode(Jump32Class, ImmSource),
+		Offset:   -1,
```
lmb, line 117: Why also make offset -1 here?

asm/alu.go
```diff
@@ -39,46 +39,60 @@ const (
...
+//	msb              lsb
+//	+-------+----+-+---+
+//	|  EXT  | OP |s|cls|
+//	+-------+----+-+---+
+type ALUOp uint16
 
-const aluMask OpCode = 0xf0
+const (
+	aluMask    OpCode = 0xfff0
+	aluOpMask  ALUOp  = 0x00f0
```
lmb, line 50: Why not just extend `aluOpMask` to be `0x01f0`? Then the setters, getters should work out of the box and `aluExtMask` isn't necessay anymore.

asm/instruction.go
```diff
@@ -107,7 +114,7 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 	}
 
 	data := make([]byte, InstructionSize)
-	data[0] = byte(ins.OpCode)
+	data[0] = byte(ins.OpCode & 0x00FF)
```
lmb, line 117: Doesn't this miss setting ins.Offset if this is SDiv, SMod?

Nit: magic number, probably better as a constant?

lmb (changes requested):

asm/instruction_test.go
dylandreimerink: This was mostly here for ease of development to make sure we got the decoding right. But I have no special love for this test, so it can go.

asm/alu.go
dylandreimerink: Because this also works with the sign extended mov which has 0x08, 0x10 and 0x20 in its upper bits. The semantic being that the upper byte of the ALUOp is the lower byte of the Offset. That makes going from one to the other easier.

asm/alu.go
```diff
@@ -39,46 +39,56 @@ const (
...
+	// SMod - signed modulo
+	SMod ALUOp = 0x0190
 	// Xor - bitwise xor
-	Xor ALUOp = 0xa0
+	Xor ALUOp = 0x00a0
 	// Mov - move value from one place to another
-	Mov ALUOp = 0xb0
+	Mov ALUOp = 0x00b0
+	// SMov8 - move lower 8 bits, sign extended upper bits of target
+	SMov8 ALUOp = 0x08b0
```
lmb, line 83: Ah, interesting! I assumed that there would be a single `SMov` instruction. I guess doing it this way actually makes this more ergonomic to use? Otherwise we'd need a function which takes Size.

For consistency with LoadMemSX, etc. this really should be `MovSX`? Signedness and sign extension are two separate things, no?

asm/instruction.go
```diff
@@ -106,6 +134,26 @@ func (ins Instruction) Marshal(w io.Writer, bo binary.ByteOrder) (uint64, error)
 		return 0, fmt.Errorf("can't marshal registers: %s", err)
 	}
 
+	if ins.OpCode.Class().IsALU() {
+		switch ins.OpCode.ALUOp() {
```
lmb, line 138: Should we return an error if Offset is not 0 here? It would indicate that someone is trying to mix extended opcodes with manually specifiying offset.

asm/instruction_test.go
lmb: Lets just keep it for now.

lmb (commented):
Two questions, otherwise LGTM.

asm/alu.go
dylandreimerink: Yea, you are totally right.

asm/instruction.go
dylandreimerink: Not sure how strict we want to be on parsing. It was sort of neat that we basically handled the ISAv4 extension already, it was just the long jump rewriting that would not have worked, but we did not error on unknown instructions. I think we should be strict on the generation side but permissive on the parsing side

asm/alu.go
lmb: Having MovSX8 etc. seems fine.

asm/instruction.go
lmb: > It was sort of neat that we basically handled the ISAv4 extension already, [...] we did not error on unknown instructions.

agreed

> I think we should be strict on the generation side but permissive on the parsing side

Not sure what you mean. I would like the following:

1. Round tripping unknown opcodes should work. Worst case you don't get pretty printing.
2. Using the `asm` package to assemble things should generate useful errors if possible.

Since we do `Offset = 0` in `Unmarshal` we know that roundtripping will work since it can't hit the `Offset != 0` case. So if we return an error here we'd make (2) possible.

dylandreimerink: > Since we do Offset = 0 in Unmarshal we know that roundtripping will work since it can't hit the Offset != 0 case. 

Right. My mind was still on Offset set to whatever we got. Yes, then it should be fine.

lmb (approved):
Renamed LoadSXMemOp to LoadMemSXOp

lmb:
Thanks!

PR #1201, opened 2023-11-03
program: fix LINUX_HAS_SYSCALL_WRAPPER use with bpf_tracing.h

lmb:
bpf_tracing.h defines LINUX_HAS_SYSCALL_WRAPPER as a boolean:

    extern _Bool LINUX_HAS_SYSCALL_WRAPPER __kconfig;

Our kconfig fixup code currently expects a uint32 and therefore bails out with the following error:

    resolve .kconfig: variable LINUX_HAS_SYSCALL_WRAPPER must be
    a 32 bits integer, got Int:"_Bool"

Fix this by changing the code to expect a one byte integer instead. A cursory search on Sourcegraph shows no open source users of uint32. This makes sense since the variable is meant to be consumed via the macros from bpf_tracing.h.


lmb:
Went for this simpler fix instead, since I think the likelihood of breaking anyone is low.

paulcacheux:
I think the fix for this should match what `libbpf` does which is supporting multiple int sizes. This would allow the fix to not break any code (current code using uint32 would still work, code using uint8/_Bool would work with the fix)

We also have no guarantee that `sizeof(_Bool) == 1` AFAIK the standard does not guarantee it.

lmb:
Do you feel like contributing the proper fix? Otherwise I'll go with this, on the basis that right now HAS_SYSCALL_WRAPPER is broken in the most common use case. I don't have time to investigate the full fix right now.

paulcacheux:
Absolutely, I will push a proper fix in a few minutes

paulcacheux:
@lmb here you go https://github.com/cilium/ebpf/pull/1202

PR #1202, opened 2023-11-03
program: fix LINUX_HAS_SYSCALL_WRAPPER use with bpf_tracing.h

paulcacheux:
We currently enforces that `LINUX_HAS_SYSCALL_WRAPPER` be an `uint32` integer. On the other hand `libbpf` [supports a larger set of integer sizes](https://github.com/libbpf/libbpf/blob/3189f70538b50fe3d2fd63f77351991a224e435b/src/libbpf.c#L1877), including support for `_Bool`.

This PR implements the same support for multiple integer size (from 1 to 8 bytes)


lmb:
Squashed your commits (so that we don't get broken tests if we ever need to bisect) and re-factored a bit to just use code from `kconfig` we already have.

paulcacheux:
That's perfect thanks !

lmb (approved):

PR #1207, opened 2023-11-07
small speedups during elf loading

paulcacheux:
This PR implements two speedups/memory usage improvements in the elf loading path.

The first one is about preallocating the instructions buffer, based on the size provided by the elf section header. Due to double width load instruction we might over allocate by a bit, but from testing on our ebpf code this over allocation is negligible compared to the pre-allocation done by `slice.append` anyway, and the win is significant in case of large instructions buffer.

The second one implements a fast path for `splitSymbols` in the case where the instructions are composed of only one symbol, which is the dominant use case from my understanding (happy to be challenged on that).

I implemented a small benchmark depicting some instructions on both ends of the spectrum (one symbol, and 2 symbols at both ends of the instructions stream).

Please let me know if you would like separate PRs for each of those.

Thanks

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -373,7 +373,7 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
 	r := bufio.NewReader(section.Open())
 
 	// Decode the section's instruction stream.
-	var insns asm.Instructions
+	insns := make(asm.Instructions, 0, section.Size/asm.InstructionSize)
 	if err := insns.Unmarshal(r, ec.ByteOrder); err != nil {
 		return nil, fmt.Errorf("decoding instructions for section %s: %w", section.Name, err)
 	}
--- a/linker.go
+++ b/linker.go
@@ -7,6 +7,8 @@ import (
 	"io"
 	"math"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/btf"
 	"github.com/cilium/ebpf/internal"
@@ -57,21 +59,33 @@ func splitSymbols(insns asm.Instructions) (map[string]asm.Instructions, error) {
 		return nil, errors.New("insns is empty")
 	}
 
-	if insns[0].Symbol() == "" {
+	currentSym := insns[0].Symbol()
+	if currentSym == "" {
 		return nil, errors.New("insns must start with a Symbol")
 	}
 
-	var name string
+	start := 0
 	progs := make(map[string]asm.Instructions)
-	for _, ins := range insns {
-		if sym := ins.Symbol(); sym != "" {
-			if progs[sym] != nil {
-				return nil, fmt.Errorf("insns contains duplicate Symbol %s", sym)
-			}
-			name = sym
+	for i, ins := range insns[1:] {
+		i := i + 1
+
+		sym := ins.Symbol()
+		if sym == "" {
+			continue
 		}
 
-		progs[name] = append(progs[name], ins)
+		// New symbol, flush the old one out.
+		progs[currentSym] = slices.Clone(insns[start:i])
+
+		if progs[sym] != nil {
+			return nil, fmt.Errorf("insns contains duplicate Symbol %s", sym)
+		}
+		currentSym = sym
+		start = i
+	}
+
+	if tail := insns[start:]; len(tail) > 0 {
+		progs[currentSym] = slices.Clone(tail)
 	}
 
 	return progs, nil
--- a/btf/ext_info.go
+++ b/btf/ext_info.go
@@ -556,21 +556,21 @@ func LoadLineInfos(reader io.Reader, bo binary.ByteOrder, recordNum uint32, spec
 	return newLineInfos(lis, spec.strings)
 }
 
-func newLineInfo(li bpfLineInfo, strings *stringTable) (*lineInfo, error) {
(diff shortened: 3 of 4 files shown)


elf_reader.go
```diff
@@ -373,7 +373,7 @@ func (ec *elfCode) loadFunctions(section *elfSection) (map[string]asm.Instructio
 	r := bufio.NewReader(section.Open())
 
 	// Decode the section's instruction stream.
-	var insns asm.Instructions
+	insns := make(asm.Instructions, 0, section.Size/asm.InstructionSize)
```
lmb, line 376: Can you add the rationale from the PR text to the commit message please? Is there a benchmark which shows the effect?

Also, this will round down if section is not an even multiple. Probably not an issue though?

linker.go
```diff
@@ -57,10 +57,24 @@ func splitSymbols(insns asm.Instructions) (map[string]asm.Instructions, error) {
...
+	onlyOneSymbol := true
+	for _, ins := range insns[1:] {
+		if ins.Symbol() != "" {
+			onlyOneSymbol = false
+			break
+		}
+	}
+	if onlyOneSymbol {
+		return map[string]asm.Instructions{
+			firstSymbol: insns,
```
lmb, line 74: Currently splitSymbols always returns a copy of insns. The new code will sometimes return a copy, sometimes not, which is a great bug magnet IMO.

Have you tried instead to optimize the loop below? I can see two big inefficiencies:

* append will do extra copies
* map lookup and assigns are also expensive

Getting rid of the two should make this a lot better without changing the alloc behaviour. You can probably track last name + index and do an exact allocation + copy whenever name changes.

lmb (changes requested):

elf_reader.go
paulcacheux: I will add the rationale yes, I don't have a benchmark for now but will add one

linker.go
paulcacheux: That's a good point, I checked that we were not modifying the result of this function but this can change in the future.

I think the main point is that we split into a map and then iterate over the map to create another array (in the only place this function is called). Would you be open to removing the map completely and replacing it with an iterator so that we skip the intermediate allocation completely ?

paulcacheux: Basically `splitSymbols` returns a map, but we never really take advantage of the fact that it's a map (except in tests), we directly iterate on it and create something else from it. What I was suggesting is that `splitSymbols` returns a lazy iterator type that would allow iteration without creating a big map with all the symbols and instructions.

I don't think it matters that much, in the latest version of the commit I improved the algorithm a lot in both the fast and slow paths.

linker_test.go
lmb, line 1: I only realised we have this when looking at BenchmarkELFReader: could you use BenchmarkNewCollectionManyProgs instead of adding a new one?

linker.go
```diff
@@ -57,21 +62,50 @@ func splitSymbols(insns asm.Instructions) (map[string]asm.Instructions, error) {
...
 	}
 
-	var name string
-	progs := make(map[string]asm.Instructions)
-	for _, ins := range insns {
+	// we first count the amount of symbols
+	// this is useful for pre-allocating slices later, but also to handle
+	// the fast path when we have only one symbol
+	symbolCounter := 1
+	for _, ins := range insns[1:] {
```
lmb, line 74: Wouldn't something like the following be a bit simpler?

```diff
diff --git a/linker.go b/linker.go
index b653b805..6eeadbf1 100644
--- a/linker.go
+++ b/linker.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"slices"
 
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/btf"
@@ -57,21 +58,31 @@ func splitSymbols(insns asm.Instructions) (map[string]asm.Instructions, error) {
 		return nil, errors.New("insns is empty")
 	}
 
-	if insns[0].Symbol() == "" {
+	currentSym := insns[0].Symbol()
+	if currentSym == "" {
 		return nil, errors.New("insns must start with a Symbol")
 	}
 
-	var name string
+	start := 0
 	progs := make(map[string]asm.Instructions)
-	for _, ins := range insns {
-		if sym := ins.Symbol(); sym != "" {
-			if progs[sym] != nil {
-				return nil, fmt.Errorf("insns contains duplicate Symbol %s", sym)
-			}
-			name = sym
+	for i, ins := range insns {
+		sym := ins.Symbol()
+		if sym == "" {
+			continue
+		}
+
+		// New symbol, flush the old one out.
+		progs[currentSym] = slices.Clone(insns[start:i])
+
+		if progs[sym] != nil {
+			return nil, fmt.Errorf("insns contains duplicate Symbol %s", sym)
 		}
+		currentSym = sym
+		start = i
+	}
 
-		progs[name] = append(progs[name], ins)
+	if tail := insns[start:]; len(tail) > 0 {
+		progs[currentSym] = slices.Clone(tail)
 	}
 
 	return progs, nil
```

Completely untested, but you get the idea. No need for another type, no need to allocate indices and syms. The single symbol case would iterate through insns once and then do a copy in the `if tail` part.

linker_test.go
paulcacheux: From what I can tell, `splitSymbols` is never called from `NewCollection` which is the function benchmarked by `BenchmarkNewCollectionManyProgs`

lmb: I added a bench that just parses manyprogs.c

lmb (approved):
Ran the new benchmark and added the results to the commit messages.

Issue #1162, opened 2023-10-10
elf: library isn't compatible with libbpf section naming


ti-mo:
We're no longer fully compatible with libbpf when it comes to ELF section naming ([see libbpf docs](https://docs.kernel.org/bpf/libbpf/program_types.html)). We treat all section name patterns as prefixes and don't obey the `+` specifier. We've simply been adding to the list without keeping up with libbpf's underlying logic.

This issue is for investigating the differences and looking how we can close the gap, ideally with minimal breakage.
```[tasklist]
### Tasks
- [ ] Extend the LibBPFCompat unit test to check that all of the parsed programs have valid ProgramSpec.Type
- [ ] Refactor `getProgType` in `elf_reader.go` to adhere to the libbpf format
```

PR #1209, opened 2023-11-07
elf: generate ELF section patterns from libbpf

lmb:
Makefile: update KERNEL_VERSION

    Update the kernel version used to generated syscall wrappers to 6.6. I
    forgot to commit this previously so there are no changes.

elf: generate ELF section patterns from libbpf

    We've historically relied on manually pulling in updated ELF section names
    from libbpf. This has led to errors creeping in as well as being constantly
    out of sync.

    Fix this by using a small awk script to pull the section definitions out of
    libbpf.c. We already rely on awk to update function names in the awk package
    so this isn't a new requirement.

    It turns out that we've strayed from libbpf in a couple of places. This
    commit adds compatibility behaviour to avoid outward visible changes for the
    section names we have tests for.

    The old generate-btf target is repurposed to do the generation and therefore
    renamed to update-kernel-deps.

    Fixes https://github.com/cilium/ebpf/issues/1162


elf_reader.go
```diff
@@ -1181,109 +1182,122 @@ func (ec *elfCode) loadKsymsSection() error {
...
 
-		return t.progType, t.attachType, t.progFlags, sectionName[len(t.prefix):]
+		return programType, attachType, flags, extra
 	}
 
 	return UnspecifiedProgram, AttachNone, 0, ""
 }
 
+func matchSectionName(sectionName, pattern string) (extra string, found bool) {
+	if strings.HasSuffix(pattern, "/") {
```
ti-mo, line 1270:
```suggestion
	// Return extra part after the slash if section name matches a prefix.
	if strings.HasSuffix(pattern, "/") {
```

elf_reader.go
```diff
@@ -1181,109 +1182,122 @@ func (ec *elfCode) loadKsymsSection() error {
...
+		if !strings.HasPrefix(sectionName, pattern) {
+			return "", false
+		}
+
+		return sectionName[len(pattern):], true
+	}
+
+	if !strings.HasSuffix(pattern, "+") {
```
ti-mo, line 1278:
```suggestion
	// Pattern is not a prefix, compare directly against section name.
	if !strings.HasSuffix(pattern, "+") {
```

elf_reader.go
```diff
@@ -1181,109 +1182,122 @@ func (ec *elfCode) loadKsymsSection() error {
...
+		return "", sectionName == pattern
+	}
+
+	// '+' means either the prefix verbatim, or prefix plus extra data
+	// delimited by a forward slash.
+	prefix := pattern[:len(pattern)-1]
```
ti-mo, line 1284: From here on down, the code is pretty hard to follow. I'd personally make the previous if clause positive: `if strings.HasSuffix(pattern, "+") {` and do `strings.SplitN(sectionName, "/", 1)` to work with a slice of strings instead of manually slicing on `len(prefix)` and `len(pattern)`.

Might be even easier to split `sectionName` at the start of the function and only work with its components from there on.

Then `return "", sectionName == pattern` can become this function's fallthrough case.

ti-mo (dismissed):
Awesome, about time this was tackled. I think `matchSectionName()`'s readability can be improved, though.

elf_reader.go
lmb: Thanks for this suggestion, this was a really great idea! Also added tests for the function.

ti-mo:
Looks like this has conflicts now, and tests are failing.

lmb:
All good to go.

PR #1211, opened 2023-11-07
btf: Optimizing BTF parsing by merging `readTypes` and `inflateRawTypes`

dylandreimerink:
After profiling the BTF parsing code, it became apparent that a lot of time was spent in `readTypes` was spent allocating rawTypes. These rawTypes are only used as an intermediate step to create the final inflated types so all of the allocation work gets thrown away.

This commit merges `readTypes` and `inflateRawTypes` into a single function. This allows us to re-use the intermediate objects and only allocate the final inflated types.

This results in the following performance improvements:

```
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
                │ before.txt  │          after.txt          │
ParseVmlinux-16   49.05m ± 1%   46.08m ± 1%  -6.06% (n=100)

                │  before.txt  │           after.txt           │
                │     B/op     │     B/op      vs base         │
ParseVmlinux-16   31.45Mi ± 0%   26.65Mi ± 0%  -15.28% (n=100)

                │ before.txt  │          after.txt           │
                │  allocs/op  │  allocs/op   vs base         │
ParseVmlinux-16   534.1k ± 0%   467.5k ± 0%  -12.48% (n=100)
```


btf/types.go
```diff
@@ -841,62 +847,112 @@ func inflateRawTypes(rawTypes []rawType, rawStrings *stringTable, base *Spec) ([
...
+
+			arr := &Array{nil, nil, bArr.Nelems}
+			fixup(bArr.IndexType, &arr.Index)
+			fixup(bArr.Type, &arr.Type)
 			typ = arr
 
 		case kindStruct:
-			members, err := convertMembers(raw.data.([]btfMember), raw.Bitfield())
+			vlen := header.Vlen()
+			if len(bMembers) < vlen {
```
lmb, line 913: I think this could just be `slices.Grow`? Same for the other occurrences.

btf/types.go
```diff
@@ -841,62 +847,112 @@ func inflateRawTypes(rawTypes []rawType, rawStrings *stringTable, base *Spec) ([
 		return members, nil
 	}
 
+	var (
+		header    btfType
+		bInt      btfInt
```
lmb, line 852: Nice! I hand't thought of this to avoid the repeated heap allocations.

lmb (changes requested):

btf/types.go
dylandreimerink: Oh, nice! I didn't know about this one yet

dylandreimerink: Hmm, doesn't quite do what we would want, but made a version that does.

lmb: It would be useful to know why it doesn't do what we want, and also what it is that we want :D

dylandreimerink: There were two main thing. 

First Grow, as the name suggest grows, so it takes a number of elements to increase by not a target size like we have, and passing a negative number is not allowed so it would not have reduced any code, we would have still hard to add an if statement around it.

Secondly, it increases capacity not length. Adding capacity is handy when doing `append` but when we pass the slice to `io.ReadFull` the length is used as an indicator of how much data to read.

So what we want is to increase the length of the array to a given target if we are currently smaller. I now put that logic in the `ensureSize` function which cleans the code up nicely

lmb: Isn't the answer to that

```
slices.Grow(s[:0], n)[:n]
```

? That also gets rid of the continuous reslicing of buf, bEnums, etc.

btf/btf_types.go
```diff
@@ -300,6 +326,17 @@ const (
 	btfIntBitsShift     = 0
 )
 
+var btfIntLen = packedSize[btfInt]()
+
+func unmarshalBtfInt(bi *btfInt, b []byte, bo binary.ByteOrder) (int, error) {
```
lmb, line 331: Why not make these methods on `btfInt` and so on?

btf/types.go
dylandreimerink: That does work! I had not though about fixing it that way, very nice!

btf/btf_types.go
dylandreimerink: That works for the single types, but for the types where we have slices such as `[]btfMember` and `[]btfParam` its nice to have a function to marshal the whole slice and do the bounds checks ect.

I thought this worked just as well.

We can also go the method route but we would have to declare a type alias for the slices to be able to add methods to them.

```
type btfMembers []btfMember

func (bm btfMembers) Marshal(b []byte, bo binary.ByteOrder) (int, error) { ....
```

But that seemed like to much added code for the same effect.

lmb: Fair enough. Medium term we need to get rid of all of these manual marshalers anyways.

lmb (approved):

PR #1206, opened 2023-11-06
bump golang.org/x/sys from 0.6.0 to v0.14.0

lifubang:
As we still use `golang.org/x/sys` with the version `0.6.0`, it makes us can't bump it to the latest version `0.14.0` in other projects depending on `github.com/cilium/ebpf`.

Please see https://github.com/opencontainers/runc/pull/4108

PR #1215, opened 2023-11-09
go.mod: update golang.org/x/sys to latest unreleased version

tklauser:
Updating golang.org/x/sys/unix to the recently released v0.14.0 will break the build because of the missing definition of BPF_F_KPROBE_MULTI_RETURN, see
https://github.com/golang/go/issues/63969 and
https://github.com/cilium/ebpf/pull/1206.

This was fixed upstream in https://go.dev/cl/540395, so manually pull in the latest unreleased version to allow modules depending on cilium/ebpf to update to later versions as well.


lmb (approved):
Thanks!

Issue #1199, opened 2023-11-02
internal: replace Memoize with sync.OnceValues (Go 1.21)


lmb:
Go 1.21 added https://pkg.go.dev/sync#OnceValues. Replace our home grown `internal.Memoize` with it.

rgo3:
Closing, as #1240 was merged.

PR #1219, opened 2023-11-13
Move PossibleCPUs to a public API

alxn:
Prequel to #1192 

Note: Also added `MustPossibleCPUs()` as a wrapper.

Considered switching to `sync.OnceValues()`, but instead referenced `internal.Memoize()`

### DIFF
--- a/cpu.go
+++ b/cpu.go
@@ -1,16 +1,28 @@
-package internal
+package ebpf
 
 import (
 	"fmt"
 	"os"
 	"strings"
+
+	"github.com/cilium/ebpf/internal"
 )
 
 // PossibleCPUs returns the max number of CPUs a system may possibly have
 // Logical CPU numbers must be of the form 0-n
-var PossibleCPUs = Memoize(func() (int, error) {
+var PossibleCPUs = internal.Memoize(func() (int, error) {
 	return parseCPUsFromFile("/sys/devices/system/cpu/possible")
 })
+
+// MustPossibleCPUs is a helper that wraps a call to PossibleCPUs and panics if
+// the error is non-nil.
+func MustPossibleCPUs() int {
+	cpus, err := PossibleCPUs()
+	if err != nil {
+		panic(err)
+	}
+	return cpus
+}
 
 func parseCPUsFromFile(path string) (int, error) {
 	spec, err := os.ReadFile(path)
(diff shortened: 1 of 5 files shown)


map_test.go
```diff
@@ -1314,7 +1314,7 @@ func TestIterateMapInMap(t *testing.T) {
 func TestPerCPUMarshaling(t *testing.T) {
 	for _, typ := range []MapType{PerCPUHash, PerCPUArray, LRUCPUHash} {
 		t.Run(typ.String(), func(t *testing.T) {
-			numCPU, err := internal.PossibleCPUs()
+			numCPU, err := PossibleCPUs()
```
lmb, line 1317: Nit: use `MustPossibleCPUs` here?

map_test.go
```diff
@@ -1372,7 +1372,7 @@ type bpfCgroupStorageKey struct {
 }
 
 func TestCgroupPerCPUStorageMarshaling(t *testing.T) {
-	numCPU, err := internal.PossibleCPUs()
+	numCPU, err := PossibleCPUs()
```
lmb, line 1375: Dito.

cpu.go
```diff
@@ -1,17 +1,29 @@
...
 	"os"
 	"strings"
+
+	"github.com/cilium/ebpf/internal"
 )
 
 // PossibleCPUs returns the max number of CPUs a system may possibly have
 // Logical CPU numbers must be of the form 0-n
-var PossibleCPUs = Memoize(func() (int, error) {
+var PossibleCPUs = internal.Memoize(func() (int, error) {
```
lmb, line 13: Can you make the variable unexported instead, and wrap it in a real function? Doing it this way makes the documentation really wonky. Example: https://pkg.go.dev/github.com/cilium/ebpf/features#pkg-variables

Let's call this `PossibleCPU` to match https://pkg.go.dev/runtime#NumCPU?

lmb (changes requested):
Thanks, `sync` changes are https://github.com/cilium/ebpf/issues/1199

cpu.go
alxn: It was super weird I was wondering.

alxn:
Note I didn't change `possibleCPUs :=` -  that will ripple out more.

lmb (approved):
Thanks :)

PR #1220, opened 2023-11-13
map: allow pre-allocating per-CPU values on lookup

alxn:
Prequel to #1192 note that this *will* conflict with #1219 and need a rebase.. oh for a stacked diff workflow on GH.

### DIFF
--- a/map.go
+++ b/map.go
@@ -684,32 +684,23 @@ func (m *Map) lookupAndDeletePerCPU(key, valueOut any, flags MapLookupFlags) err
 	return unmarshalPerCPUValue(slice, int(m.valueSize), valueBytes)
 }
 
+// ensurePerCPUSlice allocates a slice for a per-CPU value if necessary.
 func ensurePerCPUSlice(sliceOrPtr any, elemLength int) (any, error) {
-	possibleCPUs, err := PossibleCPU()
-	if err != nil {
-		return nil, err
-	}
-
 	sliceOrPtrType := reflect.TypeOf(sliceOrPtr)
 	if sliceOrPtrType.Kind() == reflect.Slice {
-		sliceValue := reflect.ValueOf(sliceOrPtr)
-		if sliceValue.Len() != possibleCPUs {
-			return nil, fmt.Errorf("per-cpu slice is incorrect length, expected %d, got %d",
-				possibleCPUs, sliceValue.Len())
-		}
-		if sliceOrPtrType.Elem().Kind() == reflect.Pointer {
-			for i := 0; i < sliceValue.Len(); i++ {
-				if !sliceValue.Index(i).Elem().CanAddr() {
-					return nil, fmt.Errorf("per-cpu slice elements cannot be nil")
-				}
-			}
-		}
-		return sliceValue.Interface(), nil
+		// The target is a slice, the caller is responsible for ensuring that
+		// size is correct.
+		return sliceOrPtr, nil
 	}
 
 	slicePtrType := sliceOrPtrType
 	if slicePtrType.Kind() != reflect.Ptr || slicePtrType.Elem().Kind() != reflect.Slice {
 		return nil, fmt.Errorf("per-cpu value requires a slice or a pointer to slice")
+	}
+
+	possibleCPUs, err := PossibleCPU()
+	if err != nil {
+		return nil, err
 	}
 
 	sliceType := slicePtrType.Elem()
--- a/map.go
+++ b/map.go
@@ -642,11 +642,15 @@ func (m *Map) LookupBytes(key interface{}) ([]byte, error) {
 }
 
 func (m *Map) lookupPerCPU(key, valueOut any, flags MapLookupFlags) error {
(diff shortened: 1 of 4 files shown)


map.go
```diff
@@ -669,11 +673,57 @@ func (m *Map) lookup(key interface{}, valueOut sys.Pointer, flags MapLookupFlags
...
+	sliceElemType := sliceType.Elem()
+	sliceElemIsPointer := sliceElemType.Kind() == reflect.Ptr
+	reflect.ValueOf(sliceOrPtr).Elem().Set(slice)
+	if !sliceElemIsPointer {
+		return slice.Interface(), nil
+	}
+	sliceElemType = sliceElemType.Elem()
+
+	for i := 0; i < possibleCPUs; i++ {
+		if sliceElemIsPointer {
```
lmb, line 720: This is always true, no?

map_test.go
```diff
@@ -75,6 +75,10 @@ func TestMap(t *testing.T) {
 		t.Error("Want value 42, got", v)
 	}
 
+	sliceVal := make([]uint32, 1)
+	qt.Assert(t, m.Lookup(uint32(0), sliceVal), qt.IsNil)
```
lmb, line 79: Hmm, this is a lookup in a normal array map right? Why does looking up a slice work here?

marshalers.go
```diff
@@ -85,37 +85,30 @@ func marshalPerCPUValue(slice any, elemLength int) (sys.Pointer, error) {
 // possible CPU.
 //
 // slicePtr must be a pointer to a slice.
```
lmb, line 87: Nit: needs updating.

map_test.go
```diff
@@ -1348,6 +1352,13 @@ func TestPerCPUMarshaling(t *testing.T) {
 			}
 
 			// Make sure unmarshaling works on slices containing pointers
+			retrievedVal := make([]*customEncoding, numCPU)
+			for i := range retrievedVal {
+				retrievedVal[i] = &customEncoding{}
```
lmb, line 1357: Can you add a case where we pass a a slice which contains a nil pointer and make sure that we don't panic?

marshalers.go
```diff
@@ -85,37 +85,30 @@ func marshalPerCPUValue(slice any, elemLength int) (sys.Pointer, error) {
...
 	sliceElemType := sliceType.Elem()
 	sliceElemIsPointer := sliceElemType.Kind() == reflect.Ptr
-	if sliceElemIsPointer {
-		sliceElemType = sliceElemType.Elem()
-	}
-
 	stride := internal.Align(elemLength, 8)
 	for i := 0; i < possibleCPUs; i++ {
 		var elem any
+		v := sliceValue.Index(i)
```
lmb, line 106: This can panic since there is no bounds check against `sliceValue.Length()`. I think it makes sense to move the check for length from ensurePerCPUSlice into this function.

lmb (changes requested):

lmb:
> oh for a stacked diff workflow on GH.

Agreed, but it's also acceptable to just push multiple commits next time ;)

lmb:
FYI I'm sending you an invite to @cilium/ebpf-go-contributors which should allow you to request re-review and the like. No pressure to accept the invitation of course :)

map.go
alxn: oh it is now - since I `return` earlier.

map_test.go
alxn: it eventually makes it to `unsafeBackingMemory()`

The key/value of map can be a non-scalar type, I thought I'd check it works with slices here. I thought `uint64[2]` would be a valid `value` in the kernel?

map_test.go
alxn:
```
--- FAIL: TestPerCPUMarshaling (0.00s)
    --- FAIL: TestPerCPUMarshaling/PerCPUHash (0.00s)
panic: reflect.Value.Addr of unaddressable value [recovered]
	panic: reflect.Value.Addr of unaddressable value```

haha... I'll add the error handling in the validation.

marshalers.go
alxn: I think I should copy the check in, though it'll mean it's impossible to test against the public API, since the validation happens at the outer level.

alxn: Added a test against the inner API too.

map_test.go
lmb: Ah yes, true. Shows how finicky this all is :(

marshalers.go
lmb: I pushed eea7482ab581845bc2f74a6e0b753e186d3ff5c0 which implements what I meant: all of the reflection specific validation happens in `unmarshalPerCPUValue`. PTAL!

alxn: oh I see you meant to inline validation as you make progress (but before the syscall), vs upfront validation. Yeah technically there is no harm here since you aren't actually doing anything.

Also I forgot about `github.com/frankban/quicktest`.

PR #1223, opened 2023-11-17
map: Introduce BatchCursor abstraction

christarazi:
See commit msgs.

* Introduce BatchCursor abstraction
* Add unit test for batch lookup with chunking
* Add warning to Batch* API operations

---

Main commit for convenience: 

> Introduce BatchCursor abstraction
> 
> The batch API in the kernel is quite tricky to get right. In order to
> pass the arguments in correctly, especially regarding the "in_batch" and
> "out_batch" arguments, it requires knowledge of how the batch map
> iteration is implemented in the kernel.
> 
> In order to avoid pushing the cognitive overhead onto users of the
> library, abstract away the details of batch map iteration into an opaque
> type, BatchCursor.
> 
> Users are expected to pass in a reference to a BatchCursor and the
> library will handle allocating the underlying buffer when needed.
> Specifically, the first time that a batch API is called using the
> cursor, the underlying buffer will be nil, which signfifies that the
> "in_batch" (aka "prevKey") should be nil to indicate the starting of a
> batching operation, and "out_batch" (aka "nextKey") is set to an
> newly-allocated buffer. Previously, users were expected to carry the
> "out_batch" reference in order to pass it in as "in_batch" upon the next
> call to the batching operation. Instead, with this abstraction, the
> cursor handles it for the user, by using the underlying buffer for both
> "in_batch" and "out_batch" in subsequent calls.
> 
> Signed-off-by: Chris Tarazi <chris@isovalent.com>
> Suggested-by: Lorenz Bauer <lmb@isovalent.com>

### DIFF
--- a/map.go
+++ b/map.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/rand"
 	"os"
 	"path/filepath"
@@ -963,14 +964,15 @@ func (m *Map) guessNonExistentKey() ([]byte, error) {
 //
 // "keysOut" and "valuesOut" must be of type slice, a pointer
 // to a slice or buffer will not work.
-// "prevKey" is the key to start the batch lookup from, it will
-// *not* be included in the results. Use nil to start at the first key.
+// "cursor" is an pointer to an opaque handle. It must be non-nil. Pass
+// "cursor" to subsequent calls of this function to continue the batching
+// operation in the case of chunking.
 //
 // ErrKeyNotExist is returned when the batch lookup has reached
 // the end of all possible results, even when partial results
 // are returned. It should be used to evaluate when lookup is "done".
-func (m *Map) BatchLookup(prevKey, nextKeyOut, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
-	return m.batchLookup(sys.BPF_MAP_LOOKUP_BATCH, prevKey, nextKeyOut, keysOut, valuesOut, opts)
+func (m *Map) BatchLookup(cursor *BatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
+	return m.batchLookup(sys.BPF_MAP_LOOKUP_BATCH, cursor, keysOut, valuesOut, opts)
 }
 
 // BatchLookupAndDelete looks up many elements in a map at once,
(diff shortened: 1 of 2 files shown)


christarazi:
cc @lorenz @ti-mo

lmb:
Sorry this took so long. I needed to make some time to dig into the kernel sources what this parameter actually does. From the selftests it looks like the type of the batch is either uint32 or uint64:
* https://elixir.bootlin.com/linux/v6.6.3/source/tools/testing/selftests/bpf/map_tests/array_map_batch_ops.c#L75
* https://elixir.bootlin.com/linux/v6.6.3/source/tools/testing/selftests/bpf/map_tests/htab_map_batch_ops.c#L79
* https://elixir.bootlin.com/linux/v6.6.3/source/tools/testing/selftests/bpf/map_tests/lpm_trie_map_batch_ops.c#L73
* https://elixir.bootlin.com/linux/v6.6.3/source/tools/testing/selftests/bpf/map_tests/map_in_map_batch_ops.c#L126

It turns out that things are even more complicated though!

- [generic_map_lookup_batch](https://elixir.bootlin.com/linux/v6.6.3/source/kernel/bpf/syscall.c#L1859) requires that batch_out is key_size bytes. This is used by array and LPM maps.
- [__htab_map_lookup_and_delete_batch](https://elixir.bootlin.com/linux/v6.6.3/source/kernel/bpf/hashtab.c#L1892) requires u64. This is used by the various hash maps.

I don't want an API which requires the user to know how batch map iteration is implemented in the kernel. So at the very least we need to abstract away the job of allocating the batch storage. Minimal change I can think of:

```go
type BatchCursor struct{
    opaque []byte // len() is max(key_size, 8) to be on the safe side
}

func (*Map) BatchLookup(cursor *BatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
    if (cursor.opaque == nil) {
        // this is the first time the lookup is called. allocate cursor.opaque and pass it as in_batch and out_batch.
    }
}
```

Usage would be something like the following:

```go
var cursor BatchCursor
count, err := m.Lookup(&cursor, ...)
```

Is that something you want to contribute? Otherwise I can fix it as well.

christarazi:
@lmb Thanks for taking a look. I'll take a stab at it. I'll repurpose this PR.

christarazi:
@lmb I've pushed "Introduce BatchCursor abstraction" commit which is the minimal implementation of your idea of the BatchCursor. Let me know if that's in line with what your thinking and I'll proceed with tightening up the godocs, etc.

map.go
```diff
@@ -964,13 +964,27 @@ func (m *Map) guessNonExistentKey() ([]byte, error) {
...
 //
 // ErrKeyNotExist is returned when the batch lookup has reached
 // the end of all possible results, even when partial results
 // are returned. It should be used to evaluate when lookup is "done".
-func (m *Map) BatchLookup(prevKey, nextKeyOut, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
-	return m.batchLookup(sys.BPF_MAP_LOOKUP_BATCH, prevKey, nextKeyOut, keysOut, valuesOut, opts)
+func (m *Map) BatchLookup(cursor *BatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {
+	if err := m.validateBatchCursor(cursor); err != nil {
+		return 0, err
+	} else if cursor.opaque == nil {
```
lmb, line 978: How about pushing this down into `m.batchLookup` instead? No need to duplicate the code.

lmb (commented):
Yeah that looks good. I'll probably add a big fat warning on the method though that it's really not a very safe or good API to use.

lmb:
One thought: the API isn't really for a lookup, since there is no way to specify a key. `BatchIterate`?

christarazi:
> One thought: the API isn't really for a lookup, since there is no way to specify a key. `BatchIterate`?

I get the sentiment but my honest opinion is that it's a bit confusing to name it differently than the kernel operation that it's doing. Ultimately, the kernel is issuing a `map_lookup_batch`. My 2 cents.

christarazi:
> prevent BatchCursor reuse across maps 

Thanks, I did not think of this case, but makes perfect sense!

PR #1226, opened 2023-11-17
test run: generate runtime.pinner method for each struct

kwakubiney:
Currently, I have been able to generate `func (*T) pin(p *runtime.Pinner)` for each `T` type. Looking at ways i could dynamically determine `Pointer` types in these structs and pin each member `Pointer` type separately.


internal/sys/types.go
```diff
@@ -490,18 +491,32 @@ type BtfInfo struct {
 	KernelBtf uint32
 }
 
+func (s *BtfInfo) Pin(p *runtime.Pinner) {
+	p.Pin(s)
+	p.Pin(&s.Btf)
```
lmb, line 496: Ah, this is subtle. We don't want to pin `&s.Btf` which is the address of the `Btf` member of this struct. Instead we need to pin the pointer it contains, which is in `s.Btf.ptr`.

How about implementing a pin function on `Pointer`? `func (p Pointer) Pin(*runtime.Pinner)`.

prog.go
```diff
@@ -314,14 +314,16 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+		var pinner runtime.Pinner
 		targetID, err := findTargetInProgram(spec.AttachTarget, spec.AttachTo, spec.Type, spec.AttachType)
 		if err != nil {
 			return nil, fmt.Errorf("attach %s/%s: %w", spec.Type, spec.AttachType, err)
 		}
 
 		attr.AttachBtfId = targetID
 		attr.AttachBtfObjFd = uint32(spec.AttachTarget.FD())
-		defer runtime.KeepAlive(spec.AttachTarget)
+		pinner.Pin(spec.AttachTarget)
```
lmb, line 325: Same as for the other sys.FD case: `KeepAlive` is appropriate here.

internal/cmd/gentypes/main.go
```diff
@@ -521,33 +568,48 @@ import (
 			return nil, fmt.Errorf("output %q: %w", goAttrType, err)
 		}
 
+		fmt.Fprintf(w, "func (s *%s) Pin(p *runtime.Pinner){ p.Pin(s)\n"+
+			"%s }\n\n", goAttrType, generatePinnedPointerTypes(s.pointerTypes))
+
 		switch s.ret {
 		case retError:
-			fmt.Fprintf(w, "func %s(attr *%s) error { _, err := BPF(%s, unsafe.Pointer(attr), unsafe.Sizeof(*attr)); return err }\n\n", s.goType, goAttrType, s.cmd)
+			fmt.Fprintf(w, "func %s(attr *%s) error {\n"+
```
lmb, line 576: Nit: maybe it's nicer to use a multi line string for these?

internal/sys/types.go
```diff
@@ -624,7 +677,14 @@ type BtfLoadAttr struct {
 	BtfLogTrueSize uint32
 }
 
+func (s *BtfLoadAttr) Pin(p *runtime.Pinner) {
+	p.Pin(s)
```
lmb, line 681: This is missing pins for `Btf` and `BtfLogBuf`?

internal/cmd/gentypes/main.go
```diff
@@ -173,6 +175,7 @@ import (
 				replace(btfID, "btf_id", "attach_btf_obj_id"),
 				replace(typeID, "attach_btf_id"),
 			},
+			[]string{"XlatedProgInsns", "MapIds", "LineInfo", "FuncInfo"},
```
lmb, line 178: I worry that maintaining these by hand is going to be quite error prone. How about this:

1. Split `outputPatchedStruct` into `patchStruct` and `outputStruct` or similar.
2. Write a new function `findPointerTypes` which takes the patched result of `patchStruct` and iterates the members to find any where `Member.Type == pointer`. Return `Member.Name` as `[]string` as now.
3. Feed the patched struct and the `[]string` of the pointer types to `outputStruct`.

lmb (changes requested):
This looks great! Left comments for some small stuff. There is one subtlety which still needs to be ironed out: we're currently not pinning the "assorted" structs, aka anything that is not passed directly to a syscall. Take a look at `sys.ObjInfo` for example: we allocate a second struct that is referenced by ObjGetInfoByFdAttr. That struct also needs to be pinned recursively. I think the way to do this is to change the `sys.Info.info` method to take a `runtime.Pinner` as well. There might be others which need this treatment.

lmb:
Can you figure out where the extra allocation is coming from?

kwakubiney:
> Can you figure out where the extra allocation is coming from?

Looks like the extra allocation is coming from the `var pinner runtime.Pinner` declaration we make on each call.

internal/cmd/gentypes/main.go
kwakubiney: Realized there's no distinct btf pointer type so I type asserted against `btf.Int` and checked if the `size == 8`

lmb: True that there is no generally valid pointer, but for our purposes comparing against `pointer` should work fine.

kwakubiney: Interesting. Maybe I am missing something here but the `replace()` function takes in a pointer which is `pointer := &btf.Int{Size: 8}` and the `Type` field of the `Member` is an interface so type asserting against `btf.Pointer` doesn't match any member.

lmb:
Great sleuthing, seems like this is more subtle than I initially thought!

Here is the syscall implementation we end up calling: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/syscall/syscall_linux.go;l=65

```go
//go:uintptrkeepalive
//go:nosplit
//go:linkname Syscall
func Syscall(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err Errno) {
```

From https://go.dev/src/runtime/HACKING
> go:uintptrkeepalive
>
> The //go:uintptrkeepalive directive must be followed by a function declaration.
>
> It specifies that the function’s uintptr arguments may be pointer values that have been converted to uintptr and must be kept alive for the duration of the call, even though from the types alone it would appear that the object is no longer needed during the call.
>
> This directive is similar to //go:uintptrescapes, but it does not force arguments to escape. [...]

So the various `attr` we use are in fact not allocated on the heap but the stack! (Kinda obvious that they have to be if we currently do map ops without heap allocations but oh well...) I guess this works because the stack of the goroutine performing a syscall is "frozen" for the duration of the syscall. This is a big win for performance and not something we can give up "just" to get move-safety.

If Go started to move memory we'd have all hell break loose:
1. Direct syscall arguments are currently allocated on the stack due to runtime magic, but there is no way to guarantee this. What if we change something in the way we invoke the syscall which causes `attr` to escape to the heap? For example we return the attr to the caller for some reason. Now we're in a situation where the syscall argument escapes but is not pinned and therefore may move in a future version of Go. But we can't unconditionally pin the memory since that forces escaping. We'd need a "pin if it escapes" API or a way to force only stack allocated arguments to syscalls.
2. The direct syscall arguments might refer to heap allocated memory. For example in the `ObjInfo` case or to point at buffers we've allocated for instructions or the verifier log. That memory might move in time.

Overall, I'm not sure any more whether `runtime.Pinner` is really applicable to our problem: from reading the proposal the problem is about dealing with memory that has to be heap allocated since it is stuffed into C memory. This isn't true in our case. Seems like even people on the Go team have similar questions: https://go-review.googlesource.com/c/sys/+/465676/5/unix/syscall_freebsd.go#286

Basically, I think that we're lacking guidance from the Go team on how to future proof syscall code for moving GC. If you're up for it you could write this up on an issue on the Go bug tracker asking for documentation. (We could do draft in a Gdoc for example.)

internal/cmd/gentypes/main.go
lmb: I mean literally comparing the variables: `Member.Type == pointer`. This asserts that the variable Member.Type matches the `pointer` variable. Since both are an interface the comparison boils down to `types equal && pointers equal`.

kwakubiney:
Thanks for the links provided and very good spot with the `//go:uintptrkeepalive` directive. I currently cannot find any good documentation stating that pointer members of these implicitly pinned higher level objects are pinned as well, I assume that is the case, and we do not need to explicitly pin them either. This has been a very interesting adventure and i have learnt a lot about `unsafe` and I guess I now know why it has that name xD

lmb:
> If you're up for it you could write this up on an issue on the Go bug tracker asking for documentation. (We could do draft in a Gdoc for example.)

Let me know if this would be interesting to you, otherwise I'll draft something at some point.

kwakubiney:
I can take a stab at this > > If you're up for it you could write this up on an issue on the Go bug tracker asking for documentation. (We could do draft in a Gdoc for example.)
> 
> 
> 
> Let me know if this would be interesting to you, otherwise I'll draft something at some point.

kwakubiney:
> > If you're up for it you could write this up on an issue on the Go bug tracker asking for documentation. (We could do draft in a Gdoc for example.)
> 
> Let me know if this would be interesting to you, otherwise I'll draft something at some point.

Interested, will draft and let you know when I am done

PR #1231, opened 2023-11-23
add kfunc benchmark

lmb:
btf: better error message for type compatibility check

    We invoke the CO-RE compatibility check while resolving, which leads to the
    following error message if something goes wrong:

        program call_kfunc: fixing up kfuncs: kfunc "bpf_skb_ct_lookup":
       type mismatch: incompatible types

    This isn't very helpful to the user. Include the type names in the error
    message:

        program call_kfunc: fixing up kfuncs: kfunc "bpf_skb_ct_lookup":
       type mismatch between Fwd:"nf_conn"[struct] and
       Struct:"nf_conn"[fields=14]: incompatible types

elf: fix in-kernel kfunc test

    We currently test that resolving kfunc works both against in-kernel as well
    as module functions. Commit 65eb006d85a2 ("bpf: Move kernel test kfuncs to
    bpf_testmod") in the Linux kernel moved the 
    bpf_kfunc_call_test_mem_len_pass1 kfunc into btf_testmod.

    Rewrite the test to use the conntrack kfuncs which were also introduced in
    5.18. They can also be in a module if nf_conntrack is a module, but our CI
    kernels and my Fedora host kernel currently compile it in.

program: add benchmark

    Add a benchmark which excercises two complicated code paths: finding a
    function in the kernel to attach to, and resolving a reference to a kfunc.

    Both currently force a new parse of vmlinux BTF which is why this is
    excruciatingly slow.


dylandreimerink (approved):
LGTM

PR #1232, opened 2023-11-23
link: rename First, Last to Head, Tail

lmb:
Timo suggested renaming First and Last to make it less likely that users assume that the position is guaranteed for the lifetime of the link.


rgo3 (approved):
Is there a way for me to retrigger the netlify deploy action, it failed to download some dependencies I think?

lmb:
I tried looking for that as well, not much luck so far.

lmb:
Not a breaking change since the API hasn't been released yet.

PR #1235, opened 2023-11-24
btf: fix slow LoadKernelSpec by making Spec.Copy lazy

### LINKED ISSUE #1068: btf: make loading kernel BTF cheaper

lmb:
btf: add benchmark for Spec.TypeByID

btf: don't access Spec.types from iterator

btf: fix slow LoadKernelSpec by making Spec.Copy lazy

    LoadKernelSpec is currently very slow, on the order of tens of milliseconds. 
    The reason for this is that we copy all ~150k types before returning the
    Spec to the caller. This is necessary since we don't want independent
    callers of LoadKernelSpec to be able to influence each other.

    So far we've put up with the slowness in case of features that are not used 
    that much yet (kfuncs) or worked around it by moving APIs into the btf 
    package where we can avoid the copy.

    I've made multiple attempts at fixing this problem. In no particlar order:

    * Have an internal non-copying API and an external copying one. Users have
     to make sure they only call LoadKernelSpec once since it's expensive.
     The problem here is that we'd still like the library internal code to use
     a single copy of the kernel Spec, but this is very difficult due to
     import cycles.

    * Decode raw BTF lazily instead of slurping all types into memory. This
     saves heap memory since we keep less inflated types around and only
     needs to do minimal upfront work. However, the upfront work still takes
     10ish ms to complete and the decoding logic becomes much more complicated
     since we need to lazily apply fixups. Both Dylan and I have written
     implementations for this, which go into the 500-600 lines of changes.

    Finally, I've decided to dust of a third approach: we still parse kernel BTF
    eagerly but make copying a Spec lazy. This makes LoadKernelSpec basically
    free, at the cost of more expensive Spec.TypeByID, etc. Doing CO-RE and
    reading split BTF also slows down, although it probably ends up faster
    overall since we save a lot by not copying types. Finally, we still pin one
    copy of the kernel spec in memory, so FlushKernelSpec() is here to stay.
    It's hard to estimate how much of an issue this really is since Go's tooling
    for heap analysis is really poor.

 Fixes #1068

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -531,16 +531,25 @@ func (s *Spec) nextTypeID() (TypeID, error) {
 // Returns an error wrapping ErrNotFound if a Type with the given ID
 // does not exist in the Spec.
 func (s *Spec) TypeByID(id TypeID) (Type, error) {
-	if id < s.firstTypeID {
+	typ, ok := s.typeByID(id)
+	if !ok {
 		return nil, fmt.Errorf("look up type with ID %d (first ID is %d): %w", id, s.firstTypeID, ErrNotFound)
+	}
+
+	return typ, nil
+}
+
+func (s *Spec) typeByID(id TypeID) (Type, bool) {
+	if id < s.firstTypeID {
+		return nil, false
 	}
 
 	index := int(id - s.firstTypeID)
 	if index >= len(s.types) {
-		return nil, fmt.Errorf("look up type with ID %d: %w", id, ErrNotFound)
+		return nil, false
 	}
 
-	return s.types[index], nil
+	return s.types[index], true
 }
 
 // TypeID returns the ID for a given Type.
@@ -671,26 +680,27 @@ func LoadSplitSpecFromReader(r io.ReaderAt, base *Spec) (*Spec, error) {
 
 // TypesIterator iterates over types of a given spec.
 type TypesIterator struct {
-	types []Type
-	index int
+	spec *Spec
+	id   TypeID
+	done bool
 	// The last visited type in the spec.
 	Type Type
 }
 
 // Iterate returns the types iterator.
 func (s *Spec) Iterate() *TypesIterator {
-	// We share the backing array of types with the Spec. This is safe since
-	// we don't allow deletion or shuffling of types.
-	return &TypesIterator{types: s.types, index: 0}
+	return &TypesIterator{spec: s, id: s.firstTypeID}
 }
 
 // Next returns true as long as there are any remaining types.
 func (iter *TypesIterator) Next() bool {
-	if len(iter.types) <= iter.index {
+	if iter.done {
 		return false
 	}
 
-	iter.Type = iter.types[iter.index]
-	iter.index++
-	return true
+	var ok bool
+	iter.Type, ok = iter.spec.typeByID(iter.id)
+	iter.id++
+	iter.done = !ok
+	return !iter.done
 }
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -29,9 +29,8 @@ var (
 // ID represents the unique ID of a BTF object.
 type ID = sys.BTFID
 
-// Spec allows querying a set of Types and loading the set into the
-// kernel.
-type Spec struct {
+// immutableTypes is a set of types which musn't be changed.
+type immutableTypes struct {
 	// All types contained by the spec, not including types from the base in
 	// case the spec was parsed from split BTF.
 	types []Type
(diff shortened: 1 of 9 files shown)


dylandreimerink (approved):
:+1:

ti-mo (approved):

PR #1239, opened 2023-11-28
elf: support netkit prog

zf1575192187:


zf1575192187:
@lmb We want to use netkit based on the latest cilium/ebpf version, but netkit features 6.7-rc1 and later, there is no kernel for this version in the [](https://github.com/cilium/ci-kernels), so some checks are wrong. I will mention a pr. Add the 6.7-rc2 version of the kernel

borkmann:
Thanks for working on this @zf1575192187. Also BPF link support would be nice for the lib.

hemanthmalla:
@zf1575192187 here's the [PR](https://github.com/cilium/ebpf/pull/1257) to support netkit devices with BPF links.

PR #1252, opened 2023-12-05
doc: use Sourcegraph query for list of importers

lmb:
The pkg.go.dev importers tab is not very useful to gauge how widely the library is used because it includes lots of duplicates due to forks.

Use a Sourcegraph query instead which also includes repo stars.

### DIFF
--- a/docs/ebpf/users.md
+++ b/docs/ebpf/users.md
@@ -4,10 +4,9 @@ Below is a non-comprehensive list of open-source software built with {{ proj }},
 just for inspiration or to gain a better understanding of how to tackle certain
 problems using eBPF.
 
-A live list of package importers can be found at [:fontawesome-brands-golang: {{
-proj }} package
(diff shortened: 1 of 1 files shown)


docs/ebpf/users.md
```diff
@@ -4,10 +4,9 @@ Below is a non-comprehensive list of open-source software built with {{ proj }},
 just for inspiration or to gain a better understanding of how to tackle certain
 problems using eBPF.
 
-A live list of package importers can be found at [:fontawesome-brands-golang: {{
-proj }} package
-importers](https://pkg.go.dev/github.com/cilium/ebpf?tab=importedby). If you'd
-like to include a project on this page, feel free to open a pull request.
+A list of :fontawesome-brands-golang: {{ proj }} importers can be found on
+[Sourcegraph].
```
ti-mo, line 8: Nit: this paragraph can be reflowed. I use https://marketplace.visualstudio.com/items?itemName=stkb.rewrap to do this while authoring.

PR #1254, opened 2023-12-05
map: avoid allocations during batch lookup of common types

lmb:
Extend the optimization from 4609dc73 ("map: zero-allocation operations for common types") to BatchLookup. This means that types without implicit padding will not incur allocations.

    core: 1
    goos: linux
    goarch: amd64
    pkg: github.com/cilium/ebpf
                        │   base.txt   │              opt.txt               │
                        │    sec/op    │   sec/op     vs base               │
    Iterate/BatchLookup   4317.0n ± 2%   771.6n ± 1%  -82.13% (p=0.002 n=6)

                        │   base.txt    │              opt.txt              │
                        │     B/op      │    B/op     vs base               │
    Iterate/BatchLookup   16432.00 ± 0%   48.00 ± 0%  -99.71% (p=0.002 n=6)

                        │  base.txt  │              opt.txt              │
                        │ allocs/op  │ allocs/op   vs base               │
    Iterate/BatchLookup   4.000 ± 0%   2.000 ± 0%  -50.00% (p=0.002 n=6)

### DIFF
--- a/internal/sysenc/buffer.go
+++ b/internal/sysenc/buffer.go
@@ -32,6 +32,7 @@ func UnsafeBuffer(ptr unsafe.Pointer) Buffer {
 
 // SyscallOutput prepares a Buffer for a syscall to write into.
 //
+// size is the length of the desired buffer in bytes.
 // The buffer may point at the underlying memory of dst, in which case [Unmarshal]
 // becomes a no-op.
 //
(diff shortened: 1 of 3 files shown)


lmb:
cc @christarazi

ti-mo:
Nice. Can we also enforce this via `testing.AllocsPerRun()`? (TIL! Was looking for something like this for the longest time.)

christarazi (approved):
:rocket:

ti-mo (approved):

PR #1257, opened 2023-12-06
Add support for netkit device

hemanthmalla:
Adds support for attaching bpf programs to [netkit devices](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=22360fad5889cbefe1eca695b0cc0273ab280b56) using bpf links.

Generated code is currently based on `6.7-rc4`

TODO : 

- [x] Decide on a way to create `netkit` device in CI for testing.
- [x] Re-gen code once `6.7` is out and validate for any changes.

### DIFF
--- a/link/netkit.go
+++ b/link/netkit.go
@@ -1,0 +1,71 @@
+package link
+
+import (
+	"fmt"
+	"runtime"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+type NetkitOptions struct {
+	// Index of the interface to attach to.
+	Interface int
+	// Program to attach.
+	Program *ebpf.Program
+	// One of the AttachNetkit* constants.
+	Attach ebpf.AttachType
+	// Attach relative to an anchor. Optional.
+	Anchor Anchor
+	// Only attach if the expected revision matches.
+	ExpectedRevision uint64
+	// Flags control the attach behaviour. Specify an Anchor instead of
+	// F_LINK, F_ID, F_BEFORE, F_AFTER and R_REPLACE. Optional.
+	Flags uint32
+}
+
+func AttachNetkit(opts NetkitOptions) (Link, error) {
+	if opts.Interface < 0 {
+		return nil, fmt.Errorf("interface %d is out of bounds", opts.Interface)
+	}
+
+	if opts.Flags&anchorFlags != 0 {
+		return nil, fmt.Errorf("disallowed flags: use Anchor to specify attach target")
+	}
+
+	attr := sys.LinkCreateNetkitAttr{
+		ProgFd:           uint32(opts.Program.FD()),
+		AttachType:       sys.AttachType(opts.Attach),
+		TargetIfindex:    uint32(opts.Interface),
+		ExpectedRevision: opts.ExpectedRevision,
+		Flags:            opts.Flags,
+	}
+
+	if opts.Anchor != nil {
+		fdOrID, flags, err := opts.Anchor.anchor()
+		if err != nil {
+			return nil, fmt.Errorf("attach netkit link: %w", err)
+		}
+
+		attr.RelativeFdOrId = fdOrID
+		attr.Flags |= flags
+	}
+
+	fd, err := sys.LinkCreateNetkit(&attr)
+	runtime.KeepAlive(opts.Program)
+	runtime.KeepAlive(opts.Anchor)
+	if err != nil {
+		if haveFeatErr := haveNetkit(); haveFeatErr != nil {
+			return nil, haveFeatErr
+		}
+		return nil, fmt.Errorf("attach netkit link: %w", err)
+	}
+
+	return &netkitLink{RawLink{fd, ""}}, nil
+}
+
+type netkitLink struct {
+	RawLink
+}
+
+var _ Link = (*netkitLink)(nil)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   TMPDIR: /tmp
-  CI_MAX_KERNEL_VERSION: '6.6'
+  CI_MAX_KERNEL_VERSION: '6.7'
   CI_MIN_CLANG_VERSION: '11'
   go_version: '~1.22'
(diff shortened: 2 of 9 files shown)


chent1996:
Thanks for this great work!

borkmann:
@hemanthmalla happy new year! Given 6.7 is officially released now, could you move forward with the PR? Thanks! :)

hemanthmalla:
@lmb I remember reading somewhere, that `cilium/ebpf` prefers not to have libs like `vishvananda/netlink` as dependencies. Do you have thoughts on how we should go about creating a `netkit` device in CI for tests?
Is shelling out to do something like `ip link add nk0 type netkit` the best option here ?

ti-mo:
> @lmb I remember reading somewhere, that `cilium/ebpf` prefers not to have libs like `vishvananda/netlink` as dependencies. Do you have thoughts on how we should go about creating a `netkit` device in CI for tests? Is shelling out to do something like `ip link add nk0 type netkit` the best option here ?

Interesting, I never realized we attached to ifindex 1 in the XDP tests, but that obviously won't work for netkit. I think newer versions of Go automatically prune testing-only module dependencies, so technically there wouldn't be any downside to importing e.g. `jsimonetti/rtnetlink`. (e.g. vendoring the lib doesn't automatically pull in quicktest either, afaik)

I think we can soften our stance on importing a netlink lib for testing if this is the case. Not sure if we can do the same for examples. I think shelling out to `ip` would be strictly worse, since we'll need to deal with the various failure cases then. Using a library would make it easier to degrade gracefully (e.g. skip a test if creating a netkit dev returns EINVAL etc.) without resorting to screen scraping `ip`. I'd like to avoid that.

link/netkit.go
```diff
@@ -0,0 +1,68 @@
...
+
+	attr := sys.LinkCreateNetkitAttr{
+		ProgFd:           uint32(opts.Program.FD()),
+		AttachType:       sys.AttachType(opts.Attach),
+		TargetIfindex:    uint32(opts.Interface),
+		ExpectedRevision: opts.ExpectedRevision,
+		Flags:            opts.Flags,
+	}
+
+	if opts.Anchor != nil {
```
lmb, line 44: I think this is shared with TCX, could you factor this out into a common function? Something like `detailsFromAnchor(opts.Anchor, opts.Flags) (fdOrId, flags, error)`.

link/netkit_test.go
```diff
@@ -0,0 +1,89 @@
...
+	var err error
+	nk := &netlink.Netkit{
+		LinkAttrs: netlink.LinkAttrs{
+			Name: "nk1",
+		},
+		Mode:       netlink.NETKIT_MODE_L2,
+		Policy:     netlink.NETKIT_POLICY_FORWARD,
+		PeerPolicy: netlink.NETKIT_POLICY_BLACKHOLE,
+	}
+	err = netlink.LinkAdd(nk)
```
lmb, line 71: How much work would it be to use the library Timo mentioned, or something else deriving from https://pkg.go.dev/github.com/mdlayher/netlink? visvhananda/netlink is quite low quality.

lmb (changes requested):
Looks fine, only some small requests. Thanks!

link/netkit_test.go
hemanthmalla: I don't think `mdlayher/netlink` has support for netkit yet though ? Will need to add support for it first IIUC.

hemanthmalla: If we strongly feel about this, I can take a look at adding support for this.

ti-mo: > If we strongly feel about this

We do. :) The goal is to start moving parts of Cilium off of v/netlink once we've done some refactoring because it keeps causing problems.

link/netkit.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		attr.RelativeFdOrId = fdOrID
+		attr.Flags |= flags
+	}
+
+	fd, err := sys.LinkCreateNetkit(&attr)
+	runtime.KeepAlive(opts.Program)
+	runtime.KeepAlive(opts.Anchor)
+	if err != nil {
```
rgo3, line 57: Could we add a feature check here, akin to https://github.com/cilium/ebpf/blob/38c3af9ffeeddd1d4f5fecf3c80babc7d1cd17e0/link/tcx.go#L58
This allows callers to distinguish between cases where the netkit API is not available in the kernel and EINVAL was returned for a different reason.

link/netkit_test.go
hemanthmalla: Marking this resolved since its now tracked with https://github.com/cilium/ebpf/issues/1350

link/netkit.go
hemanthmalla: @rgo3 Added a new feature test called `haveNetkit()`. Does this look okay now ?

link/netkit.go
hemanthmalla: Hmm. Are you thinking of just abstracting the call to `anchor()` method ? We cannot move opts.Anchor nil check into the function without complicating the error check later. We could abstract updating the flag value, but we'd still need to deal with updating `RelativeFdOrId` conditionally. Sounds to me like it would end up being a wrapper over anchor() method with no other benefit. Am I missing something here ?

link/netkit.go
rgo3: Yes, looks good! Could you add a unit test like https://github.com/cilium/ebpf/blob/38c3af9ffeeddd1d4f5fecf3c80babc7d1cd17e0/link/syscalls_test.go#L25 as well?

rgo3 (approved):
LGTM, just one small nit.

link/netkit.go
hemanthmalla: Done

hemanthmalla:
Test failures seem flaky to me. Previously only `ci / Run tests on previous stable Go` failed, re-triggered the tests on same commit changes and now ` Run tests on pre-built kernel (6.7)` seems to fail. How do I re-trigger just the failed test ?

link/netkit.go
rgo3: Awesome, thanks.

PR #1259, opened 2023-12-06
map: Fix MapBatch test for BatchLookupAndDelete case

alxn:
I've been seeing this test flake on `main`:

```
$ go test -exec sudo  . -count 100
--- FAIL: TestMapBatch (0.00s)
    --- FAIL: TestMapBatch/Hash (0.00s)
        quicktest.go:12:
            error:
              values are not equal
            got:
              int(1)
            want:
              int(2)
            stack:
              /Users/alun/Code/github.com/cilium/ebpf.git/map_test.go:175
                qt.Assert(t, qt.Equals(count, len(lookupKeys)))

--- FAIL: TestMapBatch (0.00s)
    --- FAIL: TestMapBatch/Hash (0.00s)
        quicktest.go:12:
            error:
              values are not equal
            got:
              int(1)
            want:
              int(2)
            stack:
              /Users/alun/Code/github.com/cilium/ebpf.git/map_test.go:175
                qt.Assert(t, qt.Equals(count, len(lookupKeys)))

--- FAIL: TestMapBatch (0.01s)
    --- FAIL: TestMapBatch/Hash (0.01s)
        quicktest.go:12:
            error:
              values are not equal
            got:
              int(1)
            want:
              int(2)
            stack:
              /Users/alun/Code/github.com/cilium/ebpf.git/map_test.go:175
                qt.Assert(t, qt.Equals(count, len(lookupKeys)))

FAIL
FAIL	github.com/cilium/ebpf	75.904s
FAIL```

It seems that `BatchLookupAndDelete()` can actually sometimes return `1` element (verified by iterating over the map afterwards, and there were 3 items remaining).


lmb:
Something very strange happened. I checked out your PR via the GH VScode extension and then amended the commit. For some reason the checkout only did a shallow clone of your repo (how does that even work?!). Can you push you original change again? I can't push to the branch anymore.

alxn:
https://github.com/torvalds/linux/blob/master/kernel/bpf/syscall.c#L1771

I guess it could be the retry case after getting 1 element, but it was a surprise to me that it returned only 1, after being asked for 2.

lmb:
This is because the batch lookup semantics are really complicated for hash tables. It doesn't use `generic_map_lookup_batch`.

Issue #1258, opened 2023-12-06
CI: automatically build kernels


lmb:
CI relies on pre built kernels for testing. Those kernels live in the https://github.com/cilium/ci-kernels repository and are stored in large file storage. Right now adding a new kernel is a manual process:

1. Edit `make.sh`
2. Run `make`
3. Commit the resulting `.tgz` into the repo
4. Push, PR, merge

We should automate this and store the result as a container image on ghcr.io. The current build process already uses Docker behind the scenes, with the appropriate [image targets](https://github.com/cilium/ci-kernels/blob/f05416d8f33f86246ce1f319029e77d5bf6ad415/Dockerfile.binaries#L53-L61).

Rough idea for a workflow:

1. Edit `make.sh`
2. Commit, Push
3. GHA builds kernel via existing Dockerfiles
4. Merge
5. GHA builds and pushes images from `main` (vmlinux, selftests)

This will require modifying `run-tests.sh` to pull kernel + selftests from ghcr.io instead of the `.tgz`. @tb says that containers are just tars, so maybe we can get away with some additional bash magic to go from an image url to a tgz we can use. If that is too complicated we can maybe reuse docker somehow to just do an image download for us?

This new workflow has one downside: images created from a branch build are not available anywhere, so executing `./run-tests.sh` against them isn't straight forward. One solution I can think of is to just move the build tooling into the `ebpf` repo itself. `ci-kernels` really exists to hold the .tgz so once we stop doing that there is no need to keep it around anymore. We already have docker stuff in `testdata/docker` after all.

```[tasklist]
### Tasks
- [x] Build kernel images on GHA
- [x] Publish kernel images from `main`
- [x] Allow using container as kernel source in `run-tests.sh`
```

PR #1264, opened 2023-12-08
run-tests: fetch kernels and selftests from containers

lmb:
Use the new GH Actions built containers as a source for run-tests.sh.

Fixes: https://github.com/cilium/ebpf/issues/1258

### DIFF
--- a/docs/ebpf/contributing/index.md
+++ b/docs/ebpf/contributing/index.md
@@ -15,7 +15,7 @@ go test -exec sudo ./...
 ```
 
 To test the current package with a different kernel version you can use the [run-tests.sh] script.
-It requires [virtme] and qemu to be installed.
+It requires [virtme], qemu and docker to be installed.
 Unfortunately virtme is not well maintained at the moment, so we recommend installing
 a known working version:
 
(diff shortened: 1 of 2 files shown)


lmb:
@ti-mo this is now good to go.

rgo3 (approved):
LGTM!

rgo3:
I have a general question for docker images, not sure if this can happen at all, or only locally or also in CI. If I run `run-tests.sh 6.6` for the first time, I pull the docker image for 6.6 which maps to lets say 6.6.5. That image will stay in my local image cache unless I clean it up right? 
Now if there's a kernel release to 6.6.6, and run the tests again locally but have an "older" 6.6 image around that still points to the 6.6.5 image, will it pull the newer one or run the tests against the old image?

ti-mo (approved):
Thanks! With the `manifest inspect` removed, I guess CONTAINER_ENGINE can go as well.

ti-mo:
> I have a general question for docker images, not sure if this can happen at all, or only locally or also in CI. If I run `run-tests.sh 6.6` for the first time, I pull the docker image for 6.6 which maps to lets say 6.6.5. That image will stay in my local image cache unless I clean it up right? Now if there's a kernel release to 6.6.6, and run the tests again locally but have an "older" 6.6 image around that still points to the 6.6.5 image, will it pull the newer one or run the tests against the old image?

That behaviour is usually controlled by a `pull` flag, specified to `buildx` here: `--pull                          Always attempt to pull all referenced images`. On every run, this will contact the registry to check where the tag points to and pull in the necessary changes.

rgo3:
Ah, thanks for the pointer :pray:

PR #1266, opened 2023-12-11
Feat: Implement Poll mechanism for Ring Buffer

vietanhduong:
Hi everyone,

As the title suggests, the primary aim of this pull request is to implement the `Poll` function to work similarly to the `epoll_wait` syscall.

- [x] Test
- [x] Add Example


vietanhduong:
Hi @ti-mo, @lmb could you please take a look? Sorry because I can't request your review.

mythi:
> implement the `Poll` function to work similarly to the `epoll_wait` syscall.

AFAIUI, the existing `SetDeadline()` + `Read()/ReadInto()` works like this. What's the missing logic?

vietanhduong:
Hi @mythi , the `Read` function reads data from the ring one by one, even if there are more than one record. In contrast, the `Poll` function polls all records in the ring if possible. It works similarly to the `ring_buffer__poll` in libbpf.

lmb (commented):
Thanks for opening your first PR! I agree with Mikko, why do you want to add this to the library? It seems like you can implement Poll outside of the library, it doesn't need access to internals. We try to avoid adding public API as much as possible, see https://ebpf-go.dev/contributing/new-feature/

As an aside, the current implementation of Poll allocates quite a bit. You can improve it by using ReadInto, see https://github.com/cilium/ebpf/blob/14adc787359b2a2f948773ed286cfee2e7b3bffe/ringbuf/reader_test.go#L378-L387 for an example.

vietanhduong:
Hi @lmb, you're correct. I can handle this outside the library, but it would require an extra wrapper for just a Poll function. So, I thought having a dedicated Poll function would be helpful. However, I'm okay with your answer. I think we can close this PR. Thank you, guys

PR #1267, opened 2023-12-12
cmd/bpf2go: rephrase GOPACKAGE error message

lmb:
The current error message makes people think that bpf2go can only be invoked via go generate. In fact we just expect to find the package name in GOPACKAGE.

Reword the error and add a test to make sure that the error message mentions the variable.


lmb:
Based on https://github.com/cilium/ebpf/discussions/1265

PR #1268, opened 2023-12-13
GH: use an issue form for bug reports

lmb:
Lots of bug reports currently just omit information which we'd like to have. Usually this means we have to ask follow up question.

Let's try using an issue form in the hopes of getting more complete answers.


lmb:
You can see the rendered version of the file at https://github.com/cilium/ebpf/blob/0571a00178cd20dd60fbd3f548887e946a011c8e/.github/ISSUE_TEMPLATE/bug_report.yml

ti-mo (approved):

PR #1269, opened 2023-12-13
Add support for uprobe multi link

olsajiri:
Adding support for uprobe multi link.


olsajiri:
@lorenz @ti-mo I still need to test this with tetragon, but any early feedback would be great, thanks

link/uprobe.go
```diff
@@ -159,6 +159,48 @@ func (ex *Executable) load(f *internal.SafeELFFile) error {
 	return nil
 }
 
+// address calculates the address of a symbol in the executable.
+//
+// opts must not be nil.
+func (ex *Executable) addressMulti(symbols []string) ([]uint64, error) {
```
lmb, line 165: Why can't this call `ex.address` in a loop instead of copying the body?

link/uprobe_multi.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+type UprobeMultiOptions struct {
+	Path          string
```
lmb, line 17: These could use some doc comments.

link/uprobe_multi.go
```diff
@@ -0,0 +1,148 @@
...
+	RefCtrOffsets []uint64
+	Cookies       []uint64
+}
+
+func (ex *Executable) UprobeMulti(symbols []string, prog *ebpf.Program, opts *UprobeMultiOptions) (Link, error) {
+	if opts == nil {
+		opts = &UprobeMultiOptions{}
+	}
+	var err error
+	if len(symbols) != 0 {
```
lmb, line 28: Is it valid to not specify any symbols? And why do we need to avoid calling addressMulti if there aren't any?

link/uprobe_multi.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+type UprobeMultiOptions struct {
+	Path          string
+	Offsets       []uint64
```
lmb, line 18: Is this an offset or an address? Judging from `KprobeMultiOptions` I'm guessing `Address`.

link/uprobe_multi.go
```diff
@@ -0,0 +1,148 @@
...
+	if refCtrOffsets != 0 {
+		attr.RefCtrOffsets = sys.NewPointer(unsafe.Pointer(&opts.RefCtrOffsets[0]))
+	}
+	if cookies != 0 {
+		attr.Cookies = sys.NewPointer(unsafe.Pointer(&opts.Cookies[0]))
+	}
+
+	fd, err := sys.LinkCreateUprobeMulti(attr)
+	if errors.Is(err, unix.ESRCH) {
+		return nil, fmt.Errorf("XXX: %w", os.ErrNotExist)
```
lmb, line 78: ?

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"os/exec"
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal/testutils"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+func TestUprobeMulti(t *testing.T) {
+	testutils.SkipIfNotSupported(t, haveBPFLinkUprobeMulti())
```
lmb, line 14: Nit: it's cleaner to do this on the `err` returned from UprobeMulti.

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,133 @@
...
+func TestUprobeMultiErrors(t *testing.T) {
+	testutils.SkipIfNotSupported(t, haveBPFLinkUprobeMulti())
+
+	prog := mustLoadProgram(t, ebpf.Kprobe, ebpf.AttachTraceUprobeMulti, "")
+
+	// Wrong offset
+	um, err := bashEx.UprobeMulti([]string{}, prog, &UprobeMultiOptions{Offsets: []uint64{1<<64 - 1}})
+	if !errors.Is(err, unix.EINVAL) {
+		um.Close()
+		t.Skipf("need kernel change, skipping for now, expected EINVAL, got: %s", err)
```
lmb, line 57: What does this mean? If there is a kernel limitation it doesn't make much sense to me to test it in the library.

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func TestUprobeMultiInput(t *testing.T) {
+	testutils.SkipIfNotSupported(t, haveBPFLinkUprobeMulti())
+
+	prog := mustLoadProgram(t, ebpf.Kprobe, ebpf.AttachTraceUprobeMulti, "")
+
+	// One of Symbols or Addresses must be given.
+	_, err := bashEx.UprobeMulti([]string{}, prog, nil)
+	if !errors.Is(err, errInvalidInput) {
```
lmb, line 34: Can you use `qt.Assert(t, qt.ErrorIs(...))`?

link/kprobe_test.go
```diff
@@ -337,6 +337,71 @@ func newUpdaterMapProg(t *testing.T, typ ebpf.ProgramType, attach ebpf.AttachTyp
 	return m, p
 }
 
+func newUpdaterMapIncProg(t *testing.T, typ ebpf.ProgramType, attach ebpf.AttachType) (*ebpf.Map, *ebpf.Program) {
```
lmb, line 340: Can you change newUpdaterMapProg instead of copying the implementation? It should be fine to just increment the value always.

lmb (changes requested):
I think the implementation looks fine, but it would be nice if the exposed API was closer to `KprobeMulti`. Specifically, I think we should just have `UprobeMultiOptions.Symbols` like we do for kprobes.

link/kprobe_test.go
olsajiri: yep, will change

link/uprobe.go
olsajiri: yes, we can do that, will change

link/uprobe_multi.go
olsajiri: copy/paste error, will remove, thanks

link/uprobe_multi_test.go
olsajiri: ok, will change

link/uprobe_multi_test.go
olsajiri: yes, there's posted fix: https://lore.kernel.org/bpf/20231213141234.1210389-1-jolsa@kernel.org/

I can remove the test

link/uprobe_multi_test.go
olsajiri: ok, will change

link/uprobe_multi.go
olsajiri: so both uprobe and uprobe_multi work on top of offsets within the binary file,
perhaps we should rename ex.address to ex.offset

link/uprobe_multi.go
olsajiri: yes, it's valid because all the necessary info to attach can be provided by UprobeMultiOptions

link/uprobe_multi.go
olsajiri: ok

link/uprobe_multi_test.go
olsajiri: looks like this needs to be the first line of the test after all.. I'm getting error on 4.9 kernel because of the unknown attach type.. so the test will fail on loading the test program and fail

link/uprobe_multi.go
mmat11: uprobe's `Address` was initially called `Offset`, it was renamed for consistency with the Kprobe's API, more context here: https://github.com/cilium/ebpf/commit/5713a12044f2f0f2c4bdf6bd2bd1b9a8cfb77b05.

In the case of `UprobeMulti`, I guess naming it `Offsets` would be a bit confusing (even if from the kernel API perspective is more correct), probably naming it `Addresses` would be better(?)

link/uprobe.go
```diff
@@ -162,11 +162,7 @@ func (ex *Executable) load(f *internal.SafeELFFile) error {
 // address calculates the address of a symbol in the executable.
 //
 // opts must not be nil.
```
mmat11, line 164: comment is outdated

link/uprobe.go
```diff
@@ -198,8 +194,20 @@ func (ex *Executable) address(symbol string, opts *UprobeOptions) (uint64, error
 		return 0, fmt.Errorf("cannot resolve %s library call '%s': %w "+
 			"(consider providing UprobeOptions.Address)", ex.path, symbol, ErrNotSupported)
 	}
+	return address, nil
+}
 
-	return address + opts.Offset, nil
+// addressMulti calculates the array of addresses from symbols array in the executable.
+func (ex *Executable) addressMulti(symbols []string) ([]uint64, error) {
```
mmat11, line 201: nit: rename to `uprobeMultiAddresses`, for consistency?

link/uprobe_multi.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+// UprobeMultiOptions defines additional parameters that will be used
+// when opening a UprobeMulti Link.
+type UprobeMultiOptions struct {
```
mmat11, line 18: I see the kernel API has a `pid` argument, was that left out because not needed (or will be added in the future), or it doesn't work like normal uprobes' pid?

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,129 @@
...
+	qt.Assert(t, qt.ErrorIs(err, errInvalidInput))
+
+	// One offset, two cookies.
+	_, err = bashEx.UprobeMulti([]string{}, prog, &UprobeMultiOptions{
+		Offsets: []uint64{1},
+		Cookies: []uint64{2, 3},
+	})
+	qt.Assert(t, qt.ErrorIs(err, errInvalidInput))
+
+	// Two offset, ont refctr offset.
```
mmat11, line 43:
```suggestion
	// Two offset, one refctr offset.
```

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,129 @@
...
+	ex, err := OpenExecutable(elf)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Open UprobeMulti on the executable for the given symbol
+	// and attach it to the ebpf program created above.
+	um, err := ex.UprobeMulti([]string{"main", "_start"}, p, nil)
+	if errors.Is(err, ErrNoSymbol) {
+		// Assume bash::main and go::main.main always exists
```
mmat11, line 90:
```suggestion
		// Assume bash::main and bash::_start always exists
```

link/uprobe.go
olsajiri: ah right, will remove the opts, thanks

link/uprobe.go
olsajiri: hm addressMulti is not uprobe specific, it just returns offsets for symbols array,
there's:
```
  func (ex *Executable) address(symbol string) (uint64, error) {
  func (ex *Executable) addressMulti(symbols []string) ([]uint64, error) {
```

which are not probe sprecific and:

```
func (ex *Executable) uprobeAddress(symbol string, opts *UprobeOptions) (uint64, error) {
```
which returns address + opts.Offset which is specific for uprobe

uprobe multi does not provide any symbol + relative offset functionality,
it's either symbol or offset

link/uprobe_multi.go
olsajiri: ah it's leftover, I'll add it, it's the sae functionality as for uprobe, thanks

link/uprobe_multi.go
```diff
@@ -0,0 +1,164 @@
...
+	if cookies != 0 {
+		attr.Cookies = sys.NewPointer(unsafe.Pointer(&opts.Cookies[0]))
+	}
+
+	fd, err := sys.LinkCreateUprobeMulti(attr)
+	if errors.Is(err, unix.ESRCH) {
+		return nil, fmt.Errorf("XXX: %w", os.ErrNotExist)
+	}
+	if errors.Is(err, unix.EINVAL) {
+		return nil, fmt.Errorf("%w (missing kernel symbol or prog's AttachType not AttachTraceUprobeMulti?)", err)
```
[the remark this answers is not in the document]
olsajiri: right, thanks

link/uprobe_multi_test.go
olsajiri: thanks

link/uprobe_multi_test.go
olsajiri: will fix, thanks

link/uprobe_multi.go
olsajiri: I was wondering why thanks ;-) using opts.Addresses seems better in order to stay consistent with uprobe opts

olsajiri: hm, I wonder we could also add opts.Offsets array  with the same functionality as for uprobe's opt.Offset,
at the end the kernel interface is just and offset array

link/uprobe.go
mmat11: hmm ok, I had the impression that the only non-specific common variant was the `address(...)` one 👍🏻

olsajiri:
using it in tetragon https://github.com/cilium/tetragon/pull/1914 , seems ok ;-)

link/kprobe_multi_test.go
```diff
@@ -106,9 +106,10 @@ func TestKprobeMultiProgramCall(t *testing.T) {
 
 	// Trigger ebpf program call.
 	unix.Getpid()
+	unix.Gettid()
 
 	// Assert that the value at index 0 has been updated to 1.
```
lmb, line 111: Can you remove the comment? It's outdated and doesn't add much.

link/uprobe_multi.go
lmb: This is still missing?

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+	// from the executable.
+	// Mutually exclusive with symbols UprobeMulti argument.
+	Addresses []uint64
+
+	// Offsets into functions provided by symbols array in UprobeMulti
+	// For example to set uprobes to main+5 and _start+10 call UprobeMulti
+	// with:
+	//     symbols: "main", "_start"
+	//     opt.Offset: 5, 10
+	Offsets []uint64
```
lmb, line 32: TODO: This doesn't exist for KprobeMulti. Can you add KprobeMultiOptions.Offsets in a follow up?

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+func (ex *Executable) uprobeMultiAddresses(symbols []string, opts *UprobeMultiOptions) ([]uint64, error) {
+	var addresses []uint64
+
+	offsets := len(opts.Offsets)
+	for idx, symbol := range symbols {
+		address, err := ex.address(symbol)
+		if err != nil {
+			return []uint64{0}, err
+		}
+		if offsets != 0 {
```
lmb, line 53: Why are you checking len(opts.Offsets) == 0 here? IMO, if `opts.Offsets` is not nil and length doesn't match `symbols` we should return an error.

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+//
+// opts must not be nil.
+func (ex *Executable) uprobeMultiAddresses(symbols []string, opts *UprobeMultiOptions) ([]uint64, error) {
+	var addresses []uint64
+
+	offsets := len(opts.Offsets)
+	for idx, symbol := range symbols {
+		address, err := ex.address(symbol)
+		if err != nil {
+			return []uint64{0}, err
```
lmb, line 51: Why `[]uint64{0}`? `nil` is more common in such cases.

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	syms := uint32(len(symbols))
+	addrs := uint32(len(opts.Addresses))
+	offsets := uint32(len(opts.Offsets))
+
+	if addrs != 0 && (offsets != 0 || syms != 0) {
+		return nil, fmt.Errorf("Address and Offsets/symbols are mutually exclusive: %w", errInvalidInput)
+	}
+	if offsets != 0 && offsets != syms {
```
lmb, line 73: Ah, I think this check belongs in `uprobeMultiAddresses`.

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+func (ex *Executable) UprobeMulti(symbols []string, prog *ebpf.Program, opts *UprobeMultiOptions) (Link, error) {
+	if opts == nil {
+		opts = &UprobeMultiOptions{}
+	}
+
+	syms := uint32(len(symbols))
+	addrs := uint32(len(opts.Addresses))
+	offsets := uint32(len(opts.Offsets))
+
+	if addrs != 0 && (offsets != 0 || syms != 0) {
```
lmb, line 70: For the regular uprobe we require passing `symbol` but allow overriding it with `.Address`. Why are the two mutually exclusive here?

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+		// Translate symbols to offsets
+		opts.Addresses, err = ex.uprobeMultiAddresses(symbols, opts)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return ex.uprobeMulti(prog, opts, 0)
+}
+
+func (ex *Executable) UretprobeMulti(symbols []string, prog *ebpf.Program, opts *UprobeMultiOptions) (Link, error) {
```
lmb, line 88: This method completely ignores `symbols`? Also, is it even valid to specify offsets for Uretprobe?

link/uprobe_multi.go
lmb: bump

link/uprobe_multi.go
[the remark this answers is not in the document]
lmb: bump

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+
+var haveBPFLinkUprobeMulti = internal.NewFeatureTest("bpf_link_uprobe_multi", "6.6", func() error {
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Name: "probe_upm_link",
+		Type: ebpf.Kprobe,
+		Instructions: asm.Instructions{
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+		AttachType: ebpf.AttachTraceUprobeMulti,
```
lmb, line 171: Why is it not enough that the AttachType is accepted by the kernel to determine that uprobe multi is supported?

link/uprobe_multi.go
```diff
@@ -0,0 +1,199 @@
...
+	RefCtrOffsets []uint64
+
+	// Optional, array of associated BPF cookies.
+	Cookies []uint64
+}
+
+// addressMulti calculates the array of addresses from symbols array in the executable.
+//
+// opts must not be nil.
+func (ex *Executable) uprobeMultiAddresses(symbols []string, opts *UprobeMultiOptions) ([]uint64, error) {
```
lmb, line 44: Can you make this like `uprobeAddress` and handle `.Addresses`, `.Symbols` and `.Offsets` in this method?

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/cilium/ebpf/internal/testutils"
+	"github.com/go-quicktest/qt"
+)
+
+func TestUprobeMulti(t *testing.T) {
+	testutils.SkipIfNotSupported(t, haveBPFLinkUprobeMulti())
+
+	prog := mustLoadProgram(t, ebpf.Kprobe, ebpf.AttachTraceUprobeMulti, "")
+
+	um, err := bashEx.UprobeMulti([]string{"bash_logout", "main"}, prog, nil)
```
lmb, line 18: Nit: Can you reuse / add to `bashSym` instead please?

link/uprobe_multi_test.go
```diff
@@ -0,0 +1,159 @@
...
+	})
+	qt.Assert(t, qt.ErrorIs(err, errInvalidInput))
+}
+
+func TestUprobeMultiAddresses(t *testing.T) {
+	addrSym1, err := bashEx.address("main")
+	if err != nil {
+		t.Fatal(err)
+	}
+	addrSym2, err := bashEx.address("_start")
```
lmb, line 62: Why this instead of `bash_logout` for example?

lmb (changes requested):

link/kprobe_multi_test.go
olsajiri: ok

link/uprobe_multi.go
olsajiri: true, will remove that

link/uprobe_multi.go
olsajiri: yes, will change

link/uprobe_multi.go
olsajiri: we could have CONFIG_UPROBES switched off

link/uprobe_multi_test.go
olsajiri: ok

link/uprobe_multi_test.go
olsajiri: I felt comfortable it'd be always there.. bash_logout sounds good as well

link/uprobe_multi.go
olsajiri: do you mean to move the Stmbols/Offsets dimensions in uprobeMultiAddresses ?

link/uprobe_multi.go
olsajiri: ugh yep, sorry, will add it

link/uprobe_multi_test.go
olsajiri: actually bash_logout is not called for 'bash --help' .. will keep main and _start

link/uprobe_multi.go
olsajiri: actually, offsets are optional.. so it's either defined and have same dimension as symbols, or it's not defined and we use just symbols array.. but we need to check on that

link/uprobe_multi.go
olsajiri: yes, it can have same setup as entry probe, so it needs to do same resolving.. I fixed that in current version

link/uprobe_multi.go
olsajiri: ok

link/uprobe_multi.go
olsajiri: I did not realized we allow that for regular uprobe.. it feels more natural to allow to make that mutually exclusive,
I'd think user won't pass opts.Addresses array together with symbols/opts.Offsets if only Addresses will be used and the latter will be ignored.. but I can relax that check for sure

olsajiri:
@lmb @mmat11 I tried to address all the comments, hopefully I did not miss anything now, sry

there's still open question of this change  https://github.com/cilium/ebpf/pull/1269#discussion_r1450718975  I can make the change if it's needed

link/tracepoint_test.go
```diff
@@ -74,7 +74,7 @@ func TestTracepointProgramCall(t *testing.T) {
 	unix.Getpid()
 
 	// Assert that the value at index 0 has been updated to 1.
```
ti-mo, line 76: See previous comments.

link/uprobe_multi.go
```diff
@@ -0,0 +1,230 @@
...
+	})
+	switch {
+	case errors.Is(err, unix.EBADF):
+		return nil
+	case err != nil:
+		return internal.ErrNotSupported
+	}
+	// should not happen
+	fd.Close()
+	return internal.ErrNotSupported
```
ti-mo, line 229: This somewhat violates the convention of our feature probe return values. `ErrNotSupported` means a conclusive 'unsupported' verdict, but successfully attaching a uprobe to `/` seems highly unexpected. I would return a bespoke `errors.New("successfully attached uprobe_multi to /, kernel bug?")` here to help the user. Also, we tend to return `ErrNotSupported` only on return codes we recognize:

```go
	switch {
	case errors.Is(err, unix.EBADF):
		return nil
	case err != nil:
		return err
	}

	fd.Close()
	return errors.New("successfully attached uprobe_multi to /, kernel bug?")
```

`kprobe_multi.go` also handles this case:
```
	// If CONFIG_FPROBE isn't set.
	case errors.Is(err, unix.EOPNOTSUPP):
```

Does that apply here as well?

ti-mo (changes requested):
Thanks for the work! I haven't looked at the uprobe test suite in-depth, but I already left a few suggestions for the changes to kprobe tests.

link/uprobe.go
```diff
@@ -271,6 +264,21 @@ func (ex *Executable) Uretprobe(symbol string, prog *ebpf.Program, opts *UprobeO
 	return lnk, nil
 }
 
+// uprobeAddress calculates the address of a symbol in the executable
+// tuned with uprobe opts
+//
+// opts must not be nil.
+func (ex *Executable) uprobeAddress(symbol string, opts *UprobeOptions) (uint64, error) {
```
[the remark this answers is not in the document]
olsajiri: hum, so I separated `ex.address` to be used later in `ex.uprobeMultiAddresses` ... it seemed ok to have matching behaviour for uprobe and uprobe_multi ... I guess I can keep `ex.address` with `&UprobeOptions` argument and call it with nil from `ex.uprobeMultiAddresses`  but it seems strange

link/uprobe_multi.go
olsajiri: right, makes sense.. will change it, thanks

link/kprobe_multi_test.go
```diff
@@ -106,9 +106,9 @@ func TestKprobeMultiProgramCall(t *testing.T) {
 
 	// Trigger ebpf program call.
 	unix.Getpid()
+	unix.Gettid()
 
-	// Assert that the value at index 0 has been updated to 1.
-	assertMapValue(t, m, 0, 1)
+	assertMapValueGE(t, m, 0, 2)
```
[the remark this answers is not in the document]
olsajiri: right, that would indicate we got hit from both probes.. but we'd need to call the helper to get the cookie value, which might not be supported on older kernels, but yes it's doable.. any chance this could be a follow up? ;-)

link/kprobe_test.go
```diff
@@ -264,7 +264,7 @@ func TestKprobeProgramCall(t *testing.T) {
 	unix.Getpid()
 
 	// Assert that the value at index 0 has been updated to 1.
```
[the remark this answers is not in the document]
olsajiri: ok

link/tracepoint_test.go
olsajiri: ok

link/uprobe.go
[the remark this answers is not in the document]
lmb: I've pushed a change which kind of addresses this.

link/uprobe_multi.go
lmb: Also addressed in the code I pushed.

lmb (commented):
In the interest of getting this merged I'm going to push some fix up commits. One small change (get rid of unused Path) and one larger change to the way address calculation happens. Symbols and Addresses are no longer mutually exclusive, since they also aren't for Uprobes. It's valid to do `Uprobe(sym, {Addr: 0})` since we can use the symbol name to give better error messages.

PTAL.

lmb (approved):
LGTM.

link/kprobe_multi_test.go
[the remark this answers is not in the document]
ti-mo: > we'd need to call the helper to get the cookie value, which might not be supported on older kernels

Since we're testing kprobe-multi anyways, don't those kernels have cookie support already?

link/uprobe_multi.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+// UprobeMultiOptions defines additional parameters that will be used
+// when opening a UprobeMulti Link.
+type UprobeMultiOptions struct {
+	// Symbol addresses. If set, overrides the address eventually parsed
```
ti-mo, line 19:
```suggestion
	// Symbol addresses. If set, overrides the addresses eventually parsed
```

link/uprobe_multi.go
```diff
@@ -0,0 +1,225 @@
...
+	// from the executable.
+	// Mutually exclusive with symbols UprobeMulti argument.
+	Addresses []uint64
+
+	// Offsets into functions provided by symbols array in UprobeMulti
```
ti-mo, line 24:
```suggestion
	// Offsets into functions provided by UprobeMulti's symbols argument.
```

link/uprobe_multi.go
```diff
@@ -0,0 +1,225 @@
...
+	// For example to set uprobes to main+5 and _start+10 call UprobeMulti
+	// with:
+	//     symbols: "main", "_start"
+	//     opt.Offset: 5, 10
+	Offsets []uint64
+
+	// Optional, array of associated ref counter offsets.
```
ti-mo, line 31:
```suggestion
	// Optional, slice of associated ref counter offsets.
```

ti-mo (approved):
I'm a bit pressed for time, but only spotted a few nits in the docstrings. Thanks a lot for the awesome contribution! :rocket:

sancppp:
Can you contribute some usage examples to this PR?

PR #1272, opened 2023-12-13
link: fix TestUprobeExtWithOpts address

rgo3:
In c035e6ac we changed the address we pass when creating a perf_event uprobe to the first byte of the second page to support arm64. This however fails consistenly on my local setup (older intel machine). After some investigation, it turns out arm64 doesn't require the address to not be in the first page, but to be aligned with a multiple of 4.


lmb (approved):

lmb:
Thanks!

PR #1275, opened 2023-12-15
program: fix raw_tracepoint run repeat check bug

mtardy:
The `bpf_prog_test_run_raw_tp` function has a lazy check on `bpf_attr->test.repeat`: it checks that it's equal to 0 while it could check for <= 1.

```C
/* doesn't support data_in/out, ctx_out, duration, or repeat */
if (kattr->test.data_in || kattr->test.data_out ||
    kattr->test.ctx_out || kattr->test.duration ||
    kattr->test.repeat || kattr->test.batch_size)
	return -EINVAL;
```

Program.run was hard setting the repeat value to 1 in case of 0, preventing running any raw_tracepoint program. Now it handles special cases with repeat <= 0, and does not touch the value, letting the kernel handle the 0 value.

### DIFF
--- a/prog_test.go
+++ b/prog_test.go
@@ -136,6 +136,37 @@ func TestProgramRunWithOptions(t *testing.T) {
 	}
 }
 
+func TestProgramRunRawTracepoint(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.10", "RawTracepoint test run")
+
+	ins := asm.Instructions{
+		// Return 0
+		asm.LoadImm(asm.R0, 0, asm.DWord),
+		asm.Return(),
+	}
+
+	prog, err := NewProgram(&ProgramSpec{
+		Name:         "test",
+		Type:         RawTracepoint,
+		Instructions: ins,
+		License:      "MIT",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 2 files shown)


mtardy:
Looks like I need a `testutils.SkipOnOldKernel(t, "X.Y", "something")` but not sure on which version.

prog_test.go
```diff
@@ -91,7 +91,7 @@ func TestProgramRun(t *testing.T) {
 }
 
 func TestProgramRunWithOptions(t *testing.T) {
-	testutils.SkipOnOldKernel(t, "5.15", "XDP ctx_in/ctx_out")
+	testutils.SkipOnOldKernel(t, "5.10", "RawTracepoint test run")
```
lmb, line 94: This needs to go in TestProgramRunRawTracepoint :)

mtardy: yes yes I just realized I modified the wrong one ahah

mtardy: https://github.com/cilium/ebpf/compare/35b0c1b6cedb6c5f92c0eb04e2ee5eaae955d853..f810c0a9dec0414378dbfb63cdb6e4188fa62619

lmb (approved):
Thanks!

PR #1277, opened 2023-12-15
fix: add testcast for PutInteger

atgane:
I've implemented three parts to catch errors.

1. Receive an error if the len(data) is less than integer.size.
2. if integer.Encoding is not signed, throw error if it exceeds the maximum value of the unsigned size.
3. if integer.Encoding is signed, throw error when n is in a range that cannot be converted from uint64.

image

For example, the red part below throws an error.

### DIFF
--- a/internal/kconfig/kconfig.go
+++ b/internal/kconfig/kconfig.go
@@ -263,6 +263,21 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
 		return fmt.Errorf("invalid boolean value: %d", n)
 	}
 
+	if len(data) < int(integer.Size) {
+		return fmt.Errorf("byte array size %d is less than the integer size %d", len(data), integer.Size)
+	}
+
+	upperBound := uint64(1<<(8*integer.Size) - 1)
+	lowerBound := uint64(math.MaxUint64)
+	if integer.Encoding == btf.Signed {
+		upperBound = uint64(1<<(8*integer.Size-1) - 1)
+		lowerBound = lowerBound - upperBound - 1
+	}
+
+	if !(n <= upperBound || lowerBound < n) {
+		return fmt.Errorf("%d exceeded the given byte range %d bytes size", n, integer.Size)
+	}
+
 	switch integer.Size {
 	case 1:
 		data[0] = byte(n)
--- a/internal/kconfig/kconfig.go
+++ b/internal/kconfig/kconfig.go
@@ -264,26 +264,24 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
 	}
 
 	if len(data) < int(integer.Size) {
-		return fmt.Errorf("byte array size %d is less than the integer size %d", len(data), integer.Size)
-	}
-
-	upperBound := uint64(1<<(8*integer.Size) - 1)
-	lowerBound := uint64(math.MaxUint64)
-	if integer.Encoding == btf.Signed {
-		upperBound = uint64(1<<(8*integer.Size-1) - 1)
-		lowerBound = lowerBound - upperBound - 1
-	}
-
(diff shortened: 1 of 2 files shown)


internal/kconfig/kconfig.go
```diff
@@ -263,6 +263,21 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
 		return fmt.Errorf("invalid boolean value: %d", n)
 	}
 
+	if len(data) < int(integer.Size) {
+		return fmt.Errorf("byte array size %d is less than the integer size %d", len(data), integer.Size)
```
ti-mo, line 267: Nit: technically, data is a byte slice, not an array. Also, it's always better to tell the user what the problem is. Sure the byte slice is smaller, but why is that an issue?
```suggestion
		return fmt.Errorf("can't fit an integer of size %d into a byte slice of length %d", integer.Size, len(data))
```

internal/kconfig/kconfig.go
```diff
@@ -263,6 +263,21 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
...
+	}
+
+	upperBound := uint64(1<<(8*integer.Size) - 1)
+	lowerBound := uint64(math.MaxUint64)
+	if integer.Encoding == btf.Signed {
+		upperBound = uint64(1<<(8*integer.Size-1) - 1)
+		lowerBound = lowerBound - upperBound - 1
+	}
+
+	if !(n <= upperBound || lowerBound < n) {
```
ti-mo, line 277: This is hard to read. Try:
```suggestion
	if n > upperBound || n < lowerBound {
```

internal/kconfig/kconfig.go
```diff
@@ -263,6 +263,21 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
 		return fmt.Errorf("invalid boolean value: %d", n)
 	}
 
+	if len(data) < int(integer.Size) {
+		return fmt.Errorf("byte array size %d is less than the integer size %d", len(data), integer.Size)
+	}
+
+	upperBound := uint64(1<<(8*integer.Size) - 1)
+	lowerBound := uint64(math.MaxUint64)
```
ti-mo, line 271: Why is this called `lowerBound` if the value used is `math.MaxUint64`? Wouldn't this typically be used as an upper bound? Also, there's no need to perform this check on unsigned ints if we enforce some kind of minimum length of `data`.

internal/kconfig/kconfig.go
```diff
@@ -263,6 +263,21 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
...
+		upperBound = uint64(1<<(8*integer.Size-1) - 1)
+		lowerBound = lowerBound - upperBound - 1
+	}
+
+	if !(n <= upperBound || lowerBound < n) {
+		return fmt.Errorf("%d exceeded the given byte range %d bytes size", n, integer.Size)
+	}
+
 	switch integer.Size {
 	case 1:
```
ti-mo, line 282: Instead of the `upperBound`/`lowerBound` stuff from above, what about something like:

```suggestion
	case 1:
		if integer.Encoding == btf.Signed && (n > math.MaxInt8 || n < math.MinInt8) {
			return fmt.Errorf("...")
		}
```

This makes it stupidly obvious what the intention is and avoids the subtle math. Go defines all these constants, so we should just make use of them.

ti-mo (changes requested):

internal/kconfig/kconfig.go
atgane: That looks good, but math.MinInt8 is not unsigned value, so condition can be 

`(n > math.MaxInt8 && n < math.MaxUint64 - math.MaxInt8)`

And I'll change to return the error message only if it's signed conditions.

PR #1278, opened 2023-12-15
map: Make the Examples all testable examples.

alxn:
Noticed there were a few `// Output:` sections missing.

For `PerCPU`, it's impossible to generate variable `Output:` check, so ensure to bring back to a deterministic value regardless of CPUs.


map_test.go
```diff
@@ -2213,12 +2221,15 @@ func ExampleMap_perCPU() {
 		for _, n := range values {
 			sum += n
 		}
-		fmt.Printf("Sum of %d: %d\n", key, sum)
+		fmt.Printf("Mean of %d: %d\n", key, sum/uint32(possibleCpus))
```
ti-mo, line 2224: Not sure if calculating the mean is more illustrative of what per-cpu maps are generally used for. Sure, this test needs to run on machines of any core count, but then we should probably stick to updating only the first two CPUs like the example originally did.

Per-cpu maps are typically used for keeping packet/event counters without requiring atomics on the bpf side, and their contents are usually summed to gather total results. I don't think this new proposal is strictly clearer for a newcomer. Per-cpu maps are also seldom written to, so introducing `MustPossibleCPU()` in this context may be a little distracting.

map_test.go
```diff
@@ -2275,17 +2286,17 @@ func ExampleMap_NextKey() {
...
 		panic(err)
 	}
 
-	var nextKey string
-	if err := hash.NextKey(firstKey, &nextKey); err != nil {
+	if err := hash.NextKey(keys[0], &keys[1]); err != nil {
 		panic(err)
 	}
-
-	// Order of keys is non-deterministic due to randomized map seed
```
ti-mo, line 2288: Ack on the sort, but don't remove the comment.

map_test.go
```diff
@@ -2316,37 +2327,101 @@ func ExampleMap_Iterate() {
 		entries = hash.Iterate()
 	)
 
+	m := make(map[string]uint32)
```
ti-mo, line 2330: Nit: we use `m` for `*Map` all over the place, I did a double take when I read this.
```suggestion
	values := make(map[string]uint32)
```

map_test.go
```diff
@@ -2316,37 +2327,101 @@ func ExampleMap_Iterate() {
...
+			ValueSize:  4,
+			MaxEntries: 2,
+		},
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 10,
+	}
+	arrayOfMaps, err := NewMap(spec)
+	if err != nil {
+		if errors.Is(err, internal.ErrNotSupported) {
```
ti-mo, line 2370: Calls to `errors.Is()` don't need to be gated by `err != nil`. Move this block up a line in front of `err != nil`.

ti-mo: Why do we need to handle this case? This would need an ancient kernel to fail with ErrNotSupported. Remember, these are supposed to be minimal examples.

map_test.go
```diff
@@ -2316,37 +2327,101 @@ func ExampleMap_Iterate() {
...
-	// we are done.
-	defer m.Close()
+	if err := inner.Put(uint32(0), uint32(1)); err != nil {
+		panic(err)
+	}
+	if err := inner.Put(uint32(1), uint32(2)); err != nil {
+		panic(err)
+	}
+
+	if err := arrayOfMaps.Put(uint32(0), inner); err != nil {
```
ti-mo, line 2398: I think it could be nice to insert the inner map through `spec.Contents` instead. WDYT?

map_test.go
```diff
@@ -2316,37 +2327,101 @@ func ExampleMap_Iterate() {
...
 
 	for entries.Next(&key, &m) {
+		// Make sure that the iterated map is closed after
+		// we are done.
+		defer m.Close()
+
 		// Order of keys is non-deterministic due to randomized map seed
-		fmt.Printf("key: %v, map: %v\n", key, m)
+		fmt.Printf("key: %v\n", key)
+		items := m.Iterate()
```
ti-mo, line 2409: Move this closer to where it's used, e.g. `items := m.Iterate(); for items.Next() { ... }`, or at least make `items`, `var` and `for` a contiguous block.

map_test.go
```diff
@@ -2316,37 +2327,101 @@ func ExampleMap_Iterate() {
...
+			Type:       Array,
+			KeySize:    4,
+			ValueSize:  4,
+			MaxEntries: 2,
+		},
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 10,
+	}
+	arrayOfMaps, err := NewMap(spec)
```
ti-mo, line 2368: I'd name this `outer` as opposed to the `inner` handle you're creating below.

map_test.go
```diff
@@ -2316,37 +2327,101 @@ func ExampleMap_Iterate() {
...
+		}
+		panic(err)
+	}
+	defer arrayOfMaps.Close()
+	inner, err := NewMap(spec.InnerMap)
+	if err != nil {
+		panic(err)
+	}
+	defer inner.Close()
 	var (
```
ti-mo, line 2385: Move these closer to where they're used, right before `for entries.Next()`.

ti-mo (changes requested):
Thanks for picking this up! This code is all relatively old, and over the years we've established more consistent naming conventions and idioms for working with our own APIs, so I think it's important to reflect this in our examples as well. Hence, my comments are mostly about style, so please bear with me. :)

map_test.go
alxn: > Per-cpu maps are typically used for keeping packet/event counters

hah, yes that's exactly what I'm using them for.

Instead of the mean trick, I was thinking I could emit the *last* value, which could be the same as the first value for cpus==1

alxn: But otherwise it's very hard to generate testable output here.

map_test.go
alxn: I also discovered this is possible:

```
// Unordered output:
```

map_test.go
alxn: Fails on CI otherwise due to old kernel testing.

map_test.go
alxn: Yeah that looks nice, if you mean:
```
Contents: []MapKV{
	{uint32(0), uint32(1)},
	{uint32(1), uint32(2)},
},
```

I don't think I can get hold of `inner` until I've created the outer?

map_test.go
ti-mo: Okay, to illustrate the width of the value then, let's use PossibleCPUs() to set all cpus' slots to the same value, and in the nested loop ( currently `for _, n := range values {`), we can assert that all slots contain the same value instead of summing them. A bit contrived, but relatively straightforward.

I'm on the fence about limiting the example to only a single key, but you can make that call. We could add a comment about only using a single key for demonstration purposes in case there's any doubt.

map_test.go
ti-mo: > ```
> Contents: []MapKV{
> 	{uint32(0), uint32(1)},
> 	{uint32(1), uint32(2)},
> },
> ```

Note that this may create the illusion that all the outer map's inner maps will implicitly carry these values, which is not the case. `Contents` is just a convenience field for inserting values immediately after creation, and only occurs if you create a map using the same exact spec with those `Contents`. Maybe this is going into bikeshedding territory, I do think using `Contents` is here good.

> I don't think I can get hold of `inner` until I've created the outer?

You can! Just need to move things around a bit by declaring the inner spec ahead of the outer spec:

```go
inner := &MapSpec{
	...
	Contents: []MapKV{
		{uint32(0), uint32(1)},
		{uint32(1), uint32(2)},
	},
}

im, err := NewMap(innerSpec)
...
defer im.Close()

outer := &MapSpec{
	InnerMap: inner,
	Contents: []MapKV{
		{uint32(0), im},
	},
}
```

etc.

map_test.go
ti-mo: Ah, yes, probably the reason why we never committed to making these testable. :)

map_test.go
alxn: It does mean we can't use the sum in the output.

alxn:
@ti-mo I think it's ready for review again.

ti-mo (changes requested):
Thanks for taking another stab at this! I've made a few more observations, maybe we can avoid sorting in a few places to keep things focused.

map_test.go
```diff
@@ -2275,17 +2298,18 @@ func ExampleMap_NextKey() {
 		panic(err)
 	}
 
-	var firstKey string
-	if err := hash.NextKey(nil, &firstKey); err != nil {
+	keys := make([]string, 2)
```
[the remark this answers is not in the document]
alxn: I had to do some bad bad stuff here:

```go
	for err = hash.NextKey(nil, &next); ; err = hash.NextKey(cur, &next) {
		if errors.Is(err, ErrKeyNotExist) {
			break
		}
		if err != nil {
			panic(err)
		}
		keys = append(keys, next)
		cur = next
	}
```

map_test.go
```diff
@@ -2335,57 +2357,111 @@ func ExampleMap_Iterate() {
...
-	defer m.Close()
-
 	for entries.Next(&key, &m) {
+		// Make sure that the iterated map is closed after
+		// we are done.
+		defer m.Close()
+
 		// Order of keys is non-deterministic due to randomized map seed
-		fmt.Printf("key: %v, map: %v\n", key, m)
+		fmt.Printf("key: %v\n", key)
```
ti-mo:
```suggestion
		fmt.Printf("outerKey: %v\n", key)
```

map_test.go
```diff
@@ -2335,57 +2357,111 @@ func ExampleMap_Iterate() {
...
+		ValueSize:  4,
+		MaxEntries: 10,
+		Contents: []MapKV{
+			{uint32(0), im},
+		},
+	}
+	arrayOfMaps, err := NewMap(outer)
+	if errors.Is(err, internal.ErrNotSupported) {
+		// Fake the output if on very old kernel.
+		fmt.Println("key: 0")
```
ti-mo:
```suggestion
		fmt.Println("outerKey: 0")
```

map_test.go
```diff
@@ -2335,57 +2357,111 @@ func ExampleMap_Iterate() {
...
+		if err := items.Err(); err != nil {
+			panic(fmt.Sprint("Inner Iterator encountered an error:", err))
+		}
 	}
 
 	if err := entries.Err(); err != nil {
 		panic(fmt.Sprint("Iterator encountered an error:", err))
 	}
+	// Output:
+	// key: 0
```
ti-mo:
```suggestion
	// outerKey: 0
```

ti-mo (approved):
Thanks for sticking with this! I've pushed a small fixup to rename `key` to `outerKey` in the last example.

PR #1280, opened 2023-12-18
Support specifying output directory and package name

chent1996:
Support specifying output directory and package name

Signed-off-by: tangchen <tangchen.1@bytedance.com>


chent1996:
As discussed in https://github.com/cilium/ebpf/discussions/1265#discussioncomment-7841576. This PR adds some modifications to allow bpf2go to be used as a proper tool in Bazel

1. allow user to specify the output directory as bazel will check if all the outputs show up in it's output directory.
2.  allow user to specify the package name instead of setting the GOPACKAGE env, in bazel actions, user only allows to use the default shell env or set all env manually, if GOPACKAGE env must be added manually, then all default shell env will be lost.

cmd/bpf2go/main.go
```diff
@@ -114,11 +114,11 @@ type bpf2go struct {
 	makeBase string
 }
 
-func newB2G(stdout io.Writer, pkg, outputDir string, args []string) (*bpf2go, error) {
+func newB2G(stdout io.Writer, pkg, defaultOutDir string, args []string) (*bpf2go, error) {
```
lmb, line 117: How about just dropping `outputDir`, and then doing the Getwd lookup in here if `*outDir == ""`?

Same for `pkg`, we can reuse `getEnv` as for `BPF2GO_CC`, etc.

cmd/bpf2go/main.go
```diff
@@ -315,7 +322,17 @@ func (b2g *bpf2go) convert(tgt target, arches []string) (err error) {
 		stem = fmt.Sprintf("%s_%s_%s", outputStem, tgt.clang, tgt.linux)
 	}
 
-	objFileName := filepath.Join(b2g.outputDir, stem+".o")
+	absOutPath, err := filepath.Abs(b2g.outputDir)
+	if err != nil {
+		return err
+	}
+
+	_, err = os.Stat(absOutPath)
```
lmb, line 330: What's the pupose of the extra Stat?

lmb (changes requested):

lmb:
Can you extend the unit tests in `main_test.go`?

cmd/bpf2go/main.go
chent1996: That sounds reasonable, can you help check it again?
I also add some unit tests.

cmd/bpf2go/main.go
chent1996: I removed this

chent1996:
> Can you extend the unit tests in `main_test.go`?

Sure, I have modified all relevant test cases and added more for testing the new flags

chent1996:
@lmb 
Hi, I just found there is a conflict with a new merged request, and I resolved the conflict.
If everything is right, could you help me proceed with this PR?

lmb:
Sorry for the delay!

PR #1288, opened 2024-01-02
fix for "Use of the internal package is not allowed caused by github.com/cilium/ebpf/internal"

antoniomerlin:

PR #1289, opened 2024-01-02
examples: avoid depending on internal packages

ti-mo:
Technically this builds fine in CI, but these are supposed to function from outside the repo's ebpf module as well. These internal references were added after we removed the separate Go module from examples/.

Supersedes #1288

Thanks @antoniomerlin for reporting!

### DIFF
--- a/examples/tcprtt/main.go
+++ b/examples/tcprtt/main.go
@@ -24,7 +24,6 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/cilium/ebpf/internal"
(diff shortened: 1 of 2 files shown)


ti-mo:
Turns out `binary.NativeEndian` was only merged in Go 1.21. Converting to draft until we drop support for 1.20.

paulcacheux:
@ti-mo I think this should be good now

lmb:
@ti-mo can you fix the DCO violation?

lmb (approved):

PR #1294, opened 2024-01-05
link: add feature test for tcx

rgo3:
As with many of the other existing APIs, this adds a feature test for tcx to indicate to the user if the underlying kernel APIs are available in the running kernel.


link/syscalls.go
```diff
@@ -125,3 +125,24 @@ var haveProgQuery = internal.NewFeatureTest("BPF_PROG_QUERY", "4.15", func() err
 	}
 	return errors.New("syscall succeeded unexpectedly")
 })
+
+var haveTCX = internal.NewFeatureTest("tcx", "6.6", func() error {
```
ti-mo, line 129: Nit: this needs a test in `syscalls_test.go`

PR #1296, opened 2024-01-07
cmd/bpf2go: Make LoongArch a supported target

chenhengqi:
In #975, someone added support for the loong64 GOARCH, but did't provide a Linux string for it. Since the upstream support ([0]) has already landed, let's add the missing Linux string and make LoongArch a supported target.

  [0]: https://github.com/torvalds/linux/commit/00883922ab40


lmb (approved):
Thanks!

PR #1300, opened 2024-01-08
Add CO-RE support for kernel modules

brycekahle:
re #705 

- Adds `KernelModule string` to `ProgramSpec`. This is automatically populated by `tracefs.KernelModule()` for kprobe/fentry programs, by parsing `available_filter_functions`. 
- Adds `KernelModuleTypes *btf.Spec` to `ProgramOptions`, so user can provide their own kmod BTF.

The usage of a `*Spec` in parts of CO-RE had to be abstracted to avoid copying/merging of `[]Type` slices.

This does work (tested with `nf_conntrack` kmod and `__nf_conntrack_hash_insert` kprobe), but I imagine you have thoughts about how to adapt this approach.

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -9,12 +9,10 @@ import (
 	"io"
 	"math"
 	"os"
-	"path/filepath"
 	"reflect"
 	"sync"
 
 	"github.com/cilium/ebpf/internal"
-	"github.com/cilium/ebpf/internal/kallsyms"
 	"github.com/cilium/ebpf/internal/sys"
 )
 
@@ -190,10 +188,6 @@ type Spec struct {
 
 	// String table from ELF.
 	strings *stringTable
-}
-
-func init() {
-	kernelModuleBTF.spec = make(map[string]*Spec)
 }
 
 // LoadSpec opens file and calls LoadSpecFromReader on it.
@@ -400,173 +394,6 @@ func indexTypes(types []Type, firstTypeID TypeID) (map[Type]TypeID, map[essentia
 	}
 
 	return typeIDs, typesByName
-}
-
-// LoadKernelSpec returns the current kernel's BTF information.
-//
-// Defaults to /sys/kernel/btf/vmlinux and falls back to scanning the file system
-// for vmlinux ELFs. Returns an error wrapping ErrNotSupported if BTF is not enabled.
-func LoadKernelSpec() (*Spec, error) {
-	spec, _, err := kernelSpec()
-	if err != nil {
-		return nil, err
-	}
-	return spec.Copy(), nil
-}
-
-// LoadKernelModuleSpec returns the BTF information for the named kernel module.
-//
-// Defaults to /sys/kernel/btf/<module>.
-// Returns an error wrapping ErrNotSupported if BTF is not enabled.
-func LoadKernelModuleSpec(module string) (*Spec, error) {
-	dir, file := filepath.Split(module)
-	if dir != "" || filepath.Ext(file) != "" {
-		return nil, fmt.Errorf("invalid module name %q", module)
-	}
-	spec, err := kernelModuleSpec(module)
-	if err != nil {
-		return nil, err
-	}
-	return spec.Copy(), nil
-}
-
-var kernelBTF struct {
-	sync.RWMutex
-	spec *Spec
-	// True if the spec was read from an ELF instead of raw BTF in /sys.
-	fallback bool
-}
-
-var kernelModuleBTF struct {
-	sync.RWMutex
-	spec map[string]*Spec
-}
-
-// FlushKernelSpec removes any cached kernel type information.
-func FlushKernelSpec() {
-	kernelModuleBTF.Lock()
-	defer kernelModuleBTF.Unlock()
-	kernelBTF.Lock()
-	defer kernelBTF.Unlock()
-
-	kernelBTF.spec, kernelBTF.fallback = nil, false
-	kernelModuleBTF.spec = make(map[string]*Spec)
-
-	kallsyms.FlushKernelModuleCache()
-}
-
-func kernelSpec() (*Spec, bool, error) {
-	kernelBTF.RLock()
-	spec, fallback := kernelBTF.spec, kernelBTF.fallback
-	kernelBTF.RUnlock()
-
-	if spec == nil {
-		kernelBTF.Lock()
-		defer kernelBTF.Unlock()
-
-		spec, fallback = kernelBTF.spec, kernelBTF.fallback
-	}
-
-	if spec != nil {
-		return spec, fallback, nil
-	}
-
-	spec, fallback, err := loadKernelSpec()
-	if err != nil {
-		return nil, false, err
-	}
-
-	kernelBTF.spec, kernelBTF.fallback = spec, fallback
-	return spec, fallback, nil
-}
-
-func kernelModuleSpec(module string) (*Spec, error) {
-	kernelModuleBTF.RLock()
-	spec := kernelModuleBTF.spec[module]
-	kernelModuleBTF.RUnlock()
-
-	if spec == nil {
-		kernelModuleBTF.Lock()
-		defer kernelModuleBTF.Unlock()
-
-		spec = kernelModuleBTF.spec[module]
-	}
-
-	if spec != nil {
-		return spec, nil
-	}
-
-	spec, err := loadKernelModuleSpec(module)
-	if err != nil {
-		return nil, err
-	}
-
-	kernelModuleBTF.spec[module] = spec
-	return spec, nil
-}
-
-func loadKernelSpec() (_ *Spec, fallback bool, _ error) {
-	fh, err := os.Open("/sys/kernel/btf/vmlinux")
-	if err == nil {
-		defer fh.Close()
-
-		spec, err := loadRawSpec(fh, internal.NativeEndian, nil)
-		return spec, false, err
-	}
-
-	file, err := findVMLinux()
-	if err != nil {
-		return nil, false, err
-	}
-	defer file.Close()
-
-	spec, err := LoadSpecFromReader(file)
-	return spec, true, err
-}
-
-func loadKernelModuleSpec(module string) (*Spec, error) {
-	base, _, err := kernelSpec()
-	if err != nil {
-		return nil, err
-	}
-
-	fh, err := os.Open(filepath.Join("/sys/kernel/btf", module))
-	if err != nil {
-		return nil, err
-	}
-	defer fh.Close()
-
-	return loadRawSpec(fh, internal.NativeEndian, base)
-}
-
-// findVMLinux scans multiple well-known paths for vmlinux kernel images.
-func findVMLinux() (*os.File, error) {
-	release, err := internal.KernelRelease()
-	if err != nil {
-		return nil, err
-	}
-
-	// use same list of locations as libbpf
-	// https://github.com/libbpf/libbpf/blob/9a3a42608dbe3731256a5682a125ac1e23bced8f/src/btf.c#L3114-L3122
-	locations := []string{
-		"/boot/vmlinux-%s",
-		"/lib/modules/%s/vmlinux-%[1]s",
-		"/lib/modules/%s/build/vmlinux",
-		"/usr/lib/modules/%s/kernel/vmlinux",
-		"/usr/lib/debug/boot/vmlinux-%s",
-		"/usr/lib/debug/boot/vmlinux-%s.debug",
-		"/usr/lib/debug/lib/modules/%s/vmlinux",
-	}
-
-	for _, loc := range locations {
-		file, err := os.Open(fmt.Sprintf(loc, release))
-		if errors.Is(err, os.ErrNotExist) {
-			continue
-		}
-		return file, err
-	}
-
-	return nil, fmt.Errorf("no BTF found for kernel version %s: %w", release, internal.ErrNotSupported)
 }
 
 func guessRawBTFByteOrder(r io.ReaderAt) binary.ByteOrder {
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -9,10 +9,12 @@ import (
 	"io"
 	"math"
 	"os"
+	"path/filepath"
 	"reflect"
 	"sync"
 
 	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/kallsyms"
 	"github.com/cilium/ebpf/internal/sys"
 )
 
@@ -188,6 +190,10 @@ type Spec struct {
 
 	// String table from ELF.
 	strings *stringTable
+}
+
+func init() {
+	kernelModuleBTF.spec = make(map[string]*Spec)
 }
 
 // LoadSpec opens file and calls LoadSpecFromReader on it.
@@ -408,6 +414,22 @@ func LoadKernelSpec() (*Spec, error) {
 	return spec.Copy(), nil
 }
 
+// LoadKernelModuleSpec returns the BTF information for the named kernel module.
+//
+// Defaults to /sys/kernel/btf/<module>.
+// Returns an error wrapping ErrNotSupported if BTF is not enabled.
+func LoadKernelModuleSpec(module string) (*Spec, error) {
+	dir, file := filepath.Split(module)
+	if dir != "" || filepath.Ext(file) != "" {
+		return nil, fmt.Errorf("invalid module name %q", module)
+	}
+	spec, err := kernelModuleSpec(module)
+	if err != nil {
+		return nil, err
+	}
+	return spec.Copy(), nil
+}
+
 var kernelBTF struct {
 	sync.RWMutex
 	spec *Spec
@@ -415,12 +437,22 @@ var kernelBTF struct {
 	fallback bool
 }
 
+var kernelModuleBTF struct {
+	sync.RWMutex
+	spec map[string]*Spec
+}
+
 // FlushKernelSpec removes any cached kernel type information.
 func FlushKernelSpec() {
+	kernelModuleBTF.Lock()
+	defer kernelModuleBTF.Unlock()
 	kernelBTF.Lock()
 	defer kernelBTF.Unlock()
 
 	kernelBTF.spec, kernelBTF.fallback = nil, false
+	kernelModuleBTF.spec = make(map[string]*Spec)
+
+	kallsyms.FlushKernelModuleCache()
 }
 
 func kernelSpec() (*Spec, bool, error) {
@@ -448,6 +480,31 @@ func kernelSpec() (*Spec, bool, error) {
 	return spec, fallback, nil
 }
 
+func kernelModuleSpec(module string) (*Spec, error) {
+	kernelModuleBTF.RLock()
+	spec := kernelModuleBTF.spec[module]
+	kernelModuleBTF.RUnlock()
+
+	if spec == nil {
+		kernelModuleBTF.Lock()
+		defer kernelModuleBTF.Unlock()
+
+		spec = kernelModuleBTF.spec[module]
+	}
+
+	if spec != nil {
+		return spec, nil
+	}
+
+	spec, err := loadKernelModuleSpec(module)
+	if err != nil {
+		return nil, err
+	}
+
+	kernelModuleBTF.spec[module] = spec
+	return spec, nil
+}
+
 func loadKernelSpec() (_ *Spec, fallback bool, _ error) {
 	fh, err := os.Open("/sys/kernel/btf/vmlinux")
 	if err == nil {
@@ -465,6 +522,21 @@ func loadKernelSpec() (_ *Spec, fallback bool, _ error) {
 
 	spec, err := LoadSpecFromReader(file)
 	return spec, true, err
+}
+
+func loadKernelModuleSpec(module string) (*Spec, error) {
+	base, _, err := kernelSpec()
+	if err != nil {
+		return nil, err
+	}
+
+	fh, err := os.Open(filepath.Join("/sys/kernel/btf", module))
+	if err != nil {
+		return nil, err
+	}
+	defer fh.Close()
+
+	return loadRawSpec(fh, internal.NativeEndian, base)
 }
 
 // findVMLinux scans multiple well-known paths for vmlinux kernel images.
--- a/btf/core.go
+++ b/btf/core.go
@@ -164,6 +164,33 @@ func (k coreKind) String() string {
 	}
 }
 
+type mergedSpec []*Spec
+
+func (s mergedSpec) TypeByID(id TypeID) (Type, error) {
+	for _, sp := range s {
+		t, err := sp.TypeByID(id)
+		if err != nil {
+			if errors.Is(err, ErrNotFound) {
+				continue
+			}
+			return nil, err
+		}
+		return t, nil
+	}
+	return nil, fmt.Errorf("look up type with ID %d (first ID is %d): %w", id, s[0].imm.firstTypeID, ErrNotFound)
+}
+
+func (s mergedSpec) NamedTypes(name essentialName) []TypeID {
+	var typeIDs []TypeID
+	for _, sp := range s {
+		namedTypes := sp.imm.namedTypes[name]
+		if len(namedTypes) > 0 {
+			typeIDs = append(typeIDs, namedTypes...)
+		}
+	}
+	return typeIDs
+}
+
 // CORERelocate calculates changes needed to adjust eBPF instructions for differences
 // in types.
 //
@@ -177,17 +204,26 @@ func (k coreKind) String() string {
 //
 // Fixups are returned in the order of relos, e.g. fixup[i] is the solution
 // for relos[i].
-func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder, resolveLocalTypeID func(Type) (TypeID, error)) ([]COREFixup, error) {
-	if target == nil {
-		var err error
-		target, _, err = kernelSpec()
+func CORERelocate(relos []*CORERelocation, targets []*Spec, kmodName string, bo binary.ByteOrder, resolveLocalTypeID func(Type) (TypeID, error)) ([]COREFixup, error) {
+	if len(targets) == 0 {
+		kernelTarget, _, err := kernelSpec()
 		if err != nil {
 			return nil, fmt.Errorf("load kernel spec: %w", err)
 		}
-	}
+		targets = append(targets, kernelTarget)
 
-	if bo != target.imm.byteOrder {
-		return nil, fmt.Errorf("can't relocate %s against %s", bo, target.imm.byteOrder)
+		if kmodName != "" {
+			kmodTarget, err := kernelModuleSpec(kmodName)
+			if err != nil {
+				return nil, fmt.Errorf("load kernel module spec: %w", err)
+			}
+			targets = append(targets, kmodTarget)
+		}
+	}
+	for _, target := range targets {
+		if bo != target.imm.byteOrder {
+			return nil, fmt.Errorf("can't relocate %s against %s", bo, target.imm.byteOrder)
+		}
 	}
 
 	type reloGroup struct {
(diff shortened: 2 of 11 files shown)


brycekahle:
It looks like some examples are failing because tracefs is not mounted. I could make `KernelModule` always return `""` in face of an error, but I'm not sure that is a good idea.

btf/core.go
```diff
@@ -169,19 +169,31 @@ func (k coreKind) String() string {
...
 		if err != nil {
 			return nil, fmt.Errorf("load kernel spec: %w", err)
 		}
 	}
-
 	if bo != target.byteOrder {
 		return nil, fmt.Errorf("can't relocate %s against %s", bo, target.byteOrder)
 	}
 
+	if kmodTarget == nil && kmodName != "" {
```
lmb, line 184: This is to avoid copies of the BTF I imagine? Otherwise `kmodName` seems unused.

lmb (changes requested):
Nice, I was worried that this would be a lot more complicated!

I still have my old questions re semantics and corner cases:
* As mentioned, do we have to deal with module reload / unload? It would invalidate the list of probeable functions for sure. It might change kmod BTF as well.
* Using the fn name to figure out the kmod seems a bit error prone.
* You propose that CO-RE for a probe attached to a kprobe in a kernel module will look at `set of vmlinux types` + `set of containing kmod types`. What about a kprobe on a vmlinux function which needs access to kmod types? I'm thinking of data behind `void *ctx` and similar. (There is also a related case of probe on kmod which needs access to other kmod types, but that is probably rare.)

For the implementation, I was thinking that instead of introducing `mergedSpec` we'd operate on `[]*Spec`. `ProgramOptions.KernelTypes` would become `[]*Spec` as well. `CORERelocate` would iterate the `[]*Spec` and try to find a valid set of relocations for each target. The logic to find the best possible match would have to be adjusted accordingly: https://github.com/cilium/ebpf/blob/a8be855d547788e79c77db7b3f65ca142b6e4495/btf/core.go#L276-L294 This also decouples the CO-RE logic from the decision which set of types to use.

prog.go
```diff
@@ -127,6 +127,9 @@ type ProgramSpec struct {
 
 	// The byte order this program was compiled for, may be nil.
 	ByteOrder binary.ByteOrder
+
+	// The kernel module the AttachTo is contained in, if applicable based on Type and AttachType.
+	KernelModule string
```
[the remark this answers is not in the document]
brycekahle: There needs to be some way to determine if a function is within a kernel module or not. The same function can have different answers depending on the kernel and how it was compiled. I was trying to make it as easy as possible for the user, but we could force some extra work for kmods. I do think we should supply an easy-to-use function to make the determination (like the `KernelModule` function I added).

btf/core.go
brycekahle: `kmodTarget` can `nil` in both the non-kmod function and kmod function cases if they are expecting to use default BTF. `kmodName` is used to disambiguate between those cases.

brycekahle: `kmodName` is also passed to `kernelModuleSpec`

btf/btf.go
```diff
@@ -321,6 +370,21 @@ func loadKernelSpec() (_ *Spec, fallback bool, _ error) {
 	return spec, true, err
 }
 
+func loadKernelModuleSpec(module string) (*Spec, error) {
+	base, _, err := kernelSpec()
+	if err != nil {
+		return nil, err
+	}
+
+	fh, err := os.Open("/sys/kernel/btf/" + module)
```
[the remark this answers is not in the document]
brycekahle: It seems it only searches the one path: https://github.com/libbpf/libbpf/blob/f81eef23b33c0dbf923e863a72ce51ea4d32e291/src/btf.c#L4973

elf_reader.go
```diff
@@ -325,6 +326,17 @@ func (ec *elfCode) loadProgramSections() (map[string]*ProgramSpec, error) {
 
 		progType, attachType, progFlags, attachTo := getProgType(sec.Name)
 
+		kernelModule := ""
+		if attachTo != "" {
+			if progType == Kprobe ||
```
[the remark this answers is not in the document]
brycekahle: That makes sense. I did it here so the user would know the kernel module name if needed to load non-default BTF (like from BTFHub). It seems we may have to place that burden upon the user to correctly provide a kernel module name, if required. I think we can provide a helper function for determining if a function is within a kernel module, but I don't see a way to make it transparent AND also support non-default BTF since that is supplied during kernel load.

btf/btf.go
```diff
@@ -262,19 +266,39 @@ func LoadKernelSpec() (*Spec, error) {
...
 var kernelBTF struct {
 	sync.RWMutex
 	spec *Spec
 	// True if the spec was read from an ELF instead of raw BTF in /sys.
 	fallback bool
 }
 
+var kernelModuleBTF struct {
+	sync.RWMutex
+	spec map[string]*Spec
```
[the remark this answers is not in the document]
brycekahle: I'm not super familiar with dkms, but it seems that it is mostly concerned with rebuilding non-source tree modules when the kernel changes, rather than any kind of dynamic loading/unloading.

internal/tracefs/kprobe.go
```diff
@@ -357,3 +359,34 @@ func KprobeToken(args ProbeArgs) string {
 
 	return po
 }
+
+// KernelModule returns the kernel module, if any, a probe-able function is contained in.
+func KernelModule(fn string) (string, error) {
+	path, err := sanitizeTracefsPath("available_filter_functions")
```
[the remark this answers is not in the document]
brycekahle: It looks like using `/proc/kallsyms` is the safer and more-inclusive approach here.

brycekahle:
> For the implementation, I was thinking that instead of introducing `mergedSpec` we'd operate on `[]*Spec`. `ProgramOptions.KernelTypes` would become `[]*Spec` as well. `CORERelocate` would iterate the `[]*Spec` and try to find a valid set of relocations for each target.

Since the type IDs of kernel modules conflict with each other, you are only ever going to be dealing with vmlinux plus one optional kmod BTF. libbpf makes this same assumption.

brycekahle:
> * You propose that CO-RE for a probe attached to a kprobe in a kernel module will look at `set of vmlinux types` + `set of containing kmod types`. What about a kprobe on a vmlinux function which needs access to kmod types?

With the changes you requested, you could accomplish this by setting `KernelModule` in the program spec (and supplying `KernelModuleTypes` in program options, if you want). That field doesn't affect the attachment in any way, just the BTF load and CO-RE relocation.

brycekahle:
> * do we have to deal with module reload / unload?

I don't see a way to know if a module has been unloaded and then loaded with different contents, other than maybe comparing sizes from `/proc/modules`. This would be unlikely for a kernel module in the source tree right?

btf/btf.go
```diff
@@ -262,19 +266,39 @@ func LoadKernelSpec() (*Spec, error) {
...
+}
+
 var kernelBTF struct {
 	sync.RWMutex
 	spec *Spec
 	// True if the spec was read from an ELF instead of raw BTF in /sys.
 	fallback bool
 }
 
+var kernelModuleBTF struct {
```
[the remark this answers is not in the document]
brycekahle: Combining them introduced a deadlock because `loadKernelModuleSpec` calls `kernelSpec`

btf/btf.go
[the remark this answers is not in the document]
lmb: I know you are using btfhub, what's the situation there? I'm happy to start with just built-in BTF fwiw.

prog_test.go
```diff
@@ -1152,3 +1152,29 @@ func dupFD(tb testing.TB, fd int) int {
 
 	return dup
 }
+
+func TestKernelModule(t *testing.T) {
+	kallsyms := []byte(`0000000000000000 t hid_generic_probe	[hid_generic]
```
lmb, line 1157: What about all these other symbol types?

```
$ grep -F [ /proc/kallsyms | cut -d' ' -f2 /proc/kallsyms | sort -u
a
A
b
B
d
D
r
R
t
T
V
w
W
```

lmb (commented):
Okay, went over the PR again. Here is what I would propose, let me know what you think.

* Skip `ProgramSpec.KernelModule`. Instead, `ProgramOptions.KernelTypes` becomes `[]*btf.Spec` as discussed. If `KernelTypes` is nil, the library tries to find the appropriate BTF to use, via means of parsing kallsyms / available_filter_functions and doing a lookup of ProgramSpec.AttachTo. Most of the time this will be just vmlinux. Some of the time this will be vmlinux + kmod via LoadKernelModuleSpec.
* kallsyms / available_filter_functions: which one is more appropriate to use? I'd also prefer to not export `KernelModule` for now, since it ties us into an API. Probably best to stick this into `internal/kallsyms`. Need to figure out whether this needs to be cached, and if yes how to invalidate the cache when the set of loaded modules changes. Something in `proc` that lists loaded modules maybe?
* Merge `FlushKernelModuleCache` into `btf.FlushKernelSpec()`. There should be a single function to clean up cached info, and in a way this is all related to btf stuff anyways. We could rename the function if need be.
* CORERelocate will work against `[]*Spec` as discussed. I think it makes sense to reject target type ID lookups for kmods as long as there is no canonical way to refer to these IDs from the kernel side. This might need some finessing in CORERelocate, I can probably take a look at that once the first implementation is in.

lmb (commented):
Okay, went over the PR again. Here is what I would propose, let me know what you think.

* Skip `ProgramSpec.KernelModule`. Instead, `ProgramOptions.KernelTypes` becomes `[]*btf.Spec` as discussed. If `KernelTypes` is nil, the library tries to find the appropriate BTF to use, via means of parsing kallsyms / available_filter_functions and doing a lookup of ProgramSpec.AttachTo. Most of the time this will be just vmlinux. Some of the time this will be vmlinux + kmod via LoadKernelModuleSpec.
* kallsyms / available_filter_functions: which one is more appropriate to use? I'd also prefer to not export `KernelModule` for now, since it ties us into an API. Probably best to stick this into `internal/kallsyms`. Need to figure out whether this needs to be cached, and if yes how to invalidate the cache when the set of loaded modules changes. Something in `proc` that lists loaded modules maybe?
* Merge `FlushKernelModuleCache` into `btf.FlushKernelSpec()`. There should be a single function to clean up cached info, and in a way this is all related to btf stuff anyways. We could rename the function if need be.
* CORERelocate will work against `[]*Spec` as discussed. I think it makes sense to reject target type ID lookups for kmods as long as there is no canonical way to refer to these IDs from the kernel side. This might need some finessing in CORERelocate, I can probably take a look at that once the first implementation is in.

btf/btf.go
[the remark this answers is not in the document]
brycekahle: when using btfhub, we always provide the BTF through `ProgramOptions` and do not rely upon the loader to find it.

brycekahle:
> * Merge `FlushKernelModuleCache` into `btf.FlushKernelSpec()`. There should be a single function to clean up cached info, and in a way this is all related to btf stuff anyways

`FlushKernelModuleCache` is flushing the cache of `kernel function` -> `module name`, not BTF related.

brycekahle:
> * If `KernelTypes` is nil, the library tries to find the appropriate BTF to use, via means of parsing kallsyms / available_filter_functions and doing a lookup of ProgramSpec.AttachTo. Most of the time this will be just vmlinux. Some of the time this will be vmlinux + kmod via LoadKernelModuleSpec.

This moves the loading of kernel/kmod BTF up from `CORERelocate` (which is a public func) to `newProgramWithOptions` because it needs access to `ProgramSpec.AttachTo`. That seems like a breaking change if folks expect `CORERelocate` to load kernel BTF for them.

brycekahle:
I made some changes that are close to what you want, but had to fill in some gaps. I would like to see `KernelModule` function exported.

prog_test.go
lmb: Yes, why though? We're only interested in functions?

btf/btf.go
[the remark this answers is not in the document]
lmb: And on btfhub module btf is merged into kernel btf? Or supplied as separate files? I remember you discussion this somewhere.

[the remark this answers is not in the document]
brycekahle: We decided on separate files. It mirrors the kernel behavior, and it means we don't have to deal with re-numbering all the types.

brycekahle:
@lmb can you take another look?

internal/kallsyms/kallsyms_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/go-quicktest/qt"
+)
+
+func TestKernelModule(t *testing.T) {
+	kallsyms := []byte(`0000000000000000 t hid_generic_probe	[hid_generic]
```
lmb, line 11: Repeating my earlier comment: https://github.com/cilium/ebpf/pull/1300#discussion_r1453494363

According to https://stackoverflow.com/questions/39120818/what-is-the-difference-between-t-and-t-in-proc-kallsyms and `man nm` my /proc/kallsyms also includes symbols which are not functions. Why not filter those out?

btf/btf.go
```diff
@@ -262,19 +267,41 @@ func LoadKernelSpec() (*Spec, error) {
...
+	sync.RWMutex
+	spec map[string]*Spec
+}
+
 // FlushKernelSpec removes any cached kernel type information.
 func FlushKernelSpec() {
 	kernelBTF.Lock()
+	kernelModuleBTF.Lock()
```
lmb, line 297: Nit: there is another deadlock here if loading a kmod spec means taking kernelBTF.Lock? The locking order there would be kernelModuleBTF, kernelBTF.

btf/btf.go
```diff
@@ -302,6 +326,31 @@ func kernelSpec() (*Spec, bool, error) {
...
+	if spec != nil {
+		return spec, nil
+	}
+
+	spec, err := loadKernelModuleSpec(module)
+	if err != nil {
+		return nil, err
+	}
+
+	kernelModuleBTF.spec[module] = spec
```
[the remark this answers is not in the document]
lmb: It's what we do for the vmlinux BTF as well. Otherwise you could have two concurrent parses of the spec going on, no?

btf/btf.go
```diff
@@ -321,6 +373,21 @@ func loadKernelSpec() (_ *Spec, fallback bool, _ error) {
 	return spec, true, err
 }
 
+func loadKernelModuleSpec(module string) (*Spec, error) {
+	base, _, err := kernelSpec()
+	if err != nil {
+		return nil, err
+	}
+
+	fh, err := os.Open("/sys/kernel/btf/" + module)
```
lmb, line 382: This currently allows path traversal attacks with a module like `../foo`. Can we sanitize the name somehow? I'm guessing that Linux doesn't allow arbitrary names either.

btf/core.go
```diff
@@ -169,17 +169,26 @@ func (k coreKind) String() string {
 //
 // Fixups are returned in the order of relos, e.g. fixup[i] is the solution
 // for relos[i].
-func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder, resolveLocalTypeID func(Type) (TypeID, error)) ([]COREFixup, error) {
-	if target == nil {
-		var err error
-		target, _, err = kernelSpec()
+func CORERelocate(relos []*CORERelocation, targets []*Spec, kmodName string, bo binary.ByteOrder, resolveLocalTypeID func(Type) (TypeID, error)) ([]COREFixup, error) {
+	if len(targets) == 0 {
```
lmb, line 173: With https://github.com/cilium/ebpf/commit/4ad6b8c1eed691c5628580c922c3534509d67a70 in place this code can go into `applyRelocations` in package ebpf, since calling LoadKernelSpec is cheap again. That means we don't need to deal with kmodName here.

btf/btf.go
```diff
@@ -694,3 +761,55 @@ func (iter *TypesIterator) Next() bool {
...
+	}
+	return 0, ErrNotFound
+}
+
+func (s mergedSpec) NamedTypesIterate(name essentialName) *namedTypesIterator {
+	iter := &namedTypesIterator{specIndex: 0, typeIndex: 0}
+	for _, sp := range s {
+		namedTypes := sp.namedTypes[name]
+		if len(namedTypes) > 0 {
+			iter.types = append(iter.types, namedTypes)
```
lmb, line 786: This is not going to work any more with lazy Spec.Copy going in. Iterating all the named types is "expensive" since it now causes copying.

btf/core.go
```diff
@@ -221,14 +230,14 @@ func CORERelocate(relos []*CORERelocation, target *Spec, bo binary.ByteOrder, re
...
+	mergeTarget := mergedSpec(targets)
 	for localType, group := range relosByType {
 		localTypeName := localType.TypeName()
 		if localTypeName == "" {
 			return nil, fmt.Errorf("relocate unnamed or anonymous type %s: %w", localType, ErrNotSupported)
 		}
 
-		targets := target.namedTypes[newEssentialName(localTypeName)]
-		fixups, err := coreCalculateFixups(group.relos, target, targets, bo)
+		fixups, err := coreCalculateFixups(group.relos, &mergeTarget, mergeTarget.NamedTypesIterate(newEssentialName(localTypeName)), bo)
```
lmb, line 240: Merging specs up front is going to be more complicated with lazy copy going in. How about the following pseudo go:

```go
for localType, group := range relosByType {
    var bestFixup
    for _, target := range targets {
        fixup, err := coreCalculateFixup(...)
        
        if fixup > bestFixup {
            bestFixup = fixup
            // other checks
        }
    }
    if bestFixup == nil {
        // generate poison all
    }
}
```

* Instead of merging the specs, we add a loop here which iterates over `targets` and tries to find a valid fixup.
* We factor out the code to do scoring and call if from here in addition to coreCalculateFixups https://github.com/cilium/ebpf/blob/f95957d1669c9d2ee4d6dc34a6dfa25c4130003d/btf/core.go#L276-L294
* We move generating the poison all fixup to after the loop over targets in this function: https://github.com/cilium/ebpf/blob/f95957d1669c9d2ee4d6dc34a6dfa25c4130003d/btf/core.go#L298-L309

lmb (commented):
Going back to semantics: your current implementation means that a type defined in a kernel module can "shadow" a vmlinux type if it's better according to our scoring. I'm on the fence whether it'd be better to always prefer vmlinux if there is a matching target, no matter how good a kmod type is. Do you have an opinion on that?

lmb:
> Ugh. Since we need different `KernelTypes []*btf.Spec` per-program, it kind of excludes us from using `NewCollection/NewCollectionWithOptions` because that only takes one set of options for all the programs.

Ah, that is annoying :( Why do you want to avoid multiple NewCollection calls?

> One idea would be `KernelTypes map[string]*btf.Spec` with either a sentinel or empty string value for the base kernel BTF.

That could work. I'd probably use `vmlinux` as a sentinel. OTOH we could also keep the current `KernelTypes` and add `KernelModuleTypes map[string]*Spec`, which means we don't break that part of the API.

> * both the user and library would have to do the `func->kmod` lookup for each program.

This is because you don't want to load all kmod by default? What if you only parsed kmod for all loaded modules instead?

> * the map only makes sense for a collection, not for a single program, so it would be in `CollectionOptions`

Hmm, I don't follow, sorry. We could still stick the map in ProgramOptions I think? Not sure how we would make use of it when loading `ebpf.Map` for example.

> Another idea is a callback function `func(kmodName string) (*btf.Spec, error)`

I'm a bit wary of callbacks. If we add this it begs the question whether we'd use this for vmlinux as well? Does it deprecate `KernelTypes`? Gut feeling only, but I'd prefer to not use a callback.

brycekahle:
> Why do you want to avoid multiple NewCollection calls?

We have a bunch of eBPF programs in the same object file. That is currently a single collection. Anything that needs kernel modules would need to be extracted out and dealt with separately, probably as individual program loads. That makes shared map handling way more complicated. It isn't impossible, but is a smell about the API maybe being wrong.

brycekahle:
> This is because you don't want to load all kmod by default? What if you only parsed kmod for all loaded modules instead?

If you mean modules already loaded in the kernel, that can still be 100 or more. That is a lot of CPU and memory used for probably no reason.

brycekahle:
> Hmm, I don't follow, sorry. We could still stick the map in ProgramOptions I think?

We could put it in `ProgramOptions` but it doesn't make a lot of sense, because a single eBPF program can have max 2 BTFs, the kernel and a kernel module.

brycekahle:
> your current implementation means that a type defined in a kernel module can "shadow" a vmlinux type if it's better according to our scoring.

How would that happen? kmod BTF is generated usingn the vmlinux BTF so that it doesn't include anything already in vmlinux BTF.

brycekahle:
> A single eBPF program could use several structs defined in different kernel modules, so that could be more than 2 BTFs.

Can you give a real example? It is worth noting that I don't think libbpf can accomplish this at the moment.

brycekahle:
> so it can't be known at program-loading time which module owns the functions to attach to

We have code to look up the module (if any) using `/proc/kallsyms` based on the function they are trying to attach to. Is that sufficient for this use case?

brycekahle:
> I'd prefer if that worked, but I understand we have to balance complexity and usefulness from real use cases.

I'd want to make sure it is actually possible before trying to solve it. I don't think we could automagically know which additional modules to load BTF for, so the user would have to provide/specify that somehow. We can only automagically know the module for the function you are trying to attach to.

btf/core.go
brycekahle: Isn't that a behavior change? Since `CORERelocate` is public, calling with `targets = nil` means it will auto-load kernel BTF if available. Moving this code into `applyRelocations` would change that.

btf/core.go
brycekahle: I'm not actually merging them, just storing pointers to each Spec and iterating through them with wrapper functions. I updated to handle your changes in main. Let me know what you think.

brycekahle:
> No, I don't have a real example with structs. I'd prefer if that worked, but I understand we have to balance complexity and usefulness from real use cases.

I thought about this some more. With kernel module dependencies if A depends on B, it appears that A's BTF contains all the necessary types it is using from B.

brycekahle:
@lmb updated using a map for the module types.

btf/core.go
lmb: Yes, but it's one I'm comfortable with. Either way we'll break the API. We've gone back and forth on this a couple of times already xD

lmb:
Sorry for the long radio silence. I've been busy hashing out cilium/cilium work.

@brycekahle 

> We could put it in `ProgramOptions` but it doesn't make a lot of sense, because a single eBPF program can have max 2 BTFs, the kernel and a kernel module.

I don't see the problem to be honest. `ProgramOptions` is already not scoped to a specific program.

> How would that happen? kmod BTF is generated using the vmlinux BTF so that it doesn't include anything already in vmlinux BTF.

By using distinct types with the same name in multiple compilation units. Try grepping for `ring_buffer` in the kernel sources.

Which ring_buffer do we prefer? The kmod one? Or the kernel? Which one of the kernel ones?

> It is worth noting that I don't think libbpf can accomplish [relocating types from multiple kmod] at the moment.

Does that mean we can't do better? ;P

> I don't think we could automagically know which additional modules to load BTF for, so the user would have to provide/specify that somehow.

We can't figure out which modules we need up front, but we could do something like the following:

- Define an ordering in which we visit kmod (one reason I want []*Spec as argument to CORERelocate: it gives us an order)
- Keep iterating specs until we've found a target for all relocations (logic for this will be tricky)

These are the correct semantics to me (rather than just vmlinux + kmod) but I understand that the immediate need is simpler. I still want to keep the door open for this later on though.

> With kernel module dependencies if A depends on B, it appears that A's BTF contains all the necessary types it is using from B.

That's interesting! Can you show me an example or tell me how to reproduce that?

@alban 

> I'd prefer if [merged kmod btf] worked, but I understand we have to balance complexity and usefulness from real use cases.

Thanks for giving your view point and giving me confirmation that I'm not completely off xD

internal/kallsyms/kallsyms_test.go
```diff
@@ -0,0 +1,34 @@
...
+0000000000000000 T tcp_connect`)
```
lmb, line 12: Please add some not `t|T` lines.

btf/core.go
lmb: Okay, works for me. I'm not a big fan of `mergedSpec`, it seems too special purpose for me. Can you move it to `core.go` so that it's close to where it'll be used?

lmb (changes requested):
Looks fine, please make the changes to `CORERelocate` as we discussed. If targets is nil it can just return an error, which is what it was doing before.

I'm a bit concerned about the dead lock potential between kernel and kmod loading, I'd love there to  only be a single lock. Oh well.

Finally, the bit I'm still missing is how you're going to use this with external kmod BTF?

* You don't want to parse all loaded modules, since that might be in the 100s and could still be too slow.
* There is no feedback mechanism for you to figure out which modules the lib wants.

I see two solutions: replace the map with the callback you proposed or expose `programKernelModule` on `ProgramSpec`. The latter seems fine, and would also allow dropping the export of `ebpf.KernelModule`. WDYT?

brycekahle:
> Finally, the bit I'm still missing is how you're going to use this with external kmod BTF?
> 
> * You don't want to parse all loaded modules, since that might be in the 100s and could still be too slow.
> * There is no feedback mechanism for you to figure out which modules the lib wants.

You can see how I've implemented it at the moment: https://github.com/DataDog/ebpf-manager/commit/8b77c9a990a93b77762fd8a3c7deef2f5209eee1

brycekahle:
> If targets is nil it can just return an error, which is what it was doing before.

This is not what it was doing. It tries to auto-load the kernel spec: 

https://github.com/cilium/ebpf/blob/b3432fb1eca90dae072e53d428f248aecec65d18/btf/core.go#L173-L179

lmb:
> This is not what it was doing. It tries to auto-load the kernel spec:

I meant this: https://github.com/cilium/ebpf/blob/4267cbc577606bee83f21f5232c18154848292e7/btf/core.go#L169-L170

I'm looking at merging this but realised that you merged main into your branch. Can you please undo that and rebase + squash on top of main?

brycekahle:
> Can you please undo that and rebase + squash on top of main?

Done. Let me know if there are any other changes you'd like me to make.

lmb (approved):
@brycekahle I ended up making the changes myself.

- Don't export FlushKallsysm separately.
- Don't pass kmodName to CORERelocate.
- Remove the deadlock risk from having two locks protecting vmlinux and modules

I feel like these came up during the review before, but I'm not sure what happened.

Sorry it took so long to get this over the finish line!

lmb:
Going to merge this tomorrow morning UK time.

btf/kernel.go
```diff
@@ -0,0 +1,158 @@
...
+	kernelBTF.RUnlock()
+
+	if spec != nil {
+		return spec.Copy(), nil
+	}
+
+	base, err := LoadKernelSpec()
+	if err != nil {
+		return nil, fmt.Errorf("load kernel spec: %w", err)
+	}
```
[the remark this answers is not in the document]
lmb: That would cause a deadlock: we already hold kernelBTF.Lock at that point, so acquiring it in LoadKernelSpec will block indefinitely.

btf/kernel.go
```diff
@@ -0,0 +1,158 @@
...
+// Defaults to /sys/kernel/btf/vmlinux and falls back to scanning the file system
+// for vmlinux ELFs. Returns an error wrapping ErrNotSupported if BTF is not enabled.
+func LoadKernelSpec() (*Spec, error) {
+	kernelBTF.RLock()
+	spec := kernelBTF.kernel
+	kernelBTF.RUnlock()
+
+	if spec == nil {
+		kernelBTF.Lock()
+		defer kernelBTF.Unlock()
```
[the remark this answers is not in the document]
lmb: Yeah, the code is a bit hard to follow. Reason we do it this way is that dropping the read lock and acquiring the write lock are not atomic. Another goroutine might have preempted us and populated kernelBTF.kernel.

PR #1301, opened 2024-01-09
fix missing buildInfo.Main

chent1996:
In go 1.18, buildInfo.Main will not be set if the bpf2go is built with file arguments

### DIFF
--- a/cmd/bpf2go/tools.go
+++ b/cmd/bpf2go/tools.go
@@ -85,7 +85,8 @@ func toUpperFirst(str string) string {
 
 func currentModule() string {
 	bi, ok := debug.ReadBuildInfo()
-	if !ok {
+	// In go 1.18, buildInfo.Main will not be set if the bpf2go is built with file arguments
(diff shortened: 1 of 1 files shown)


chent1996:
As discussed in https://github.com/cilium/ebpf/discussions/1298.

This PR returns the default module name when **debug.ReadBuildInfo()** returns an empty result

ti-mo:
See Lorenz' comment: https://github.com/cilium/ebpf/discussions/1298#discussioncomment-8063008.

It feels like a bad idea to work around the Go toolchain like this. It's best to build the package, not `package/*.go`.

chent1996:
> See Lorenz' comment: [#1298 (reply in thread)](https://github.com/cilium/ebpf/discussions/1298#discussioncomment-8063008).
> 
> It feels like a bad idea to work around the Go toolchain like this. It's best to build the package, not `package/*.go`.

Alright, I agree it's better to do it this way

PR #1303, opened 2024-01-10
link: fix tcx feature test

rgo3:
The `BPF_LINK_CREATE` syscall code in the kernel uses the type of the program that is being attached with bpf_link to determine the actual link type, in this case TCX.

This commit ensures we pass a program of type SchedCLS to the `BPF_LINK_CREATE` syscall to actually feature test the TCX link type. Without it, the `haveTCX` feature test only probes for the availability of the `BPF_LINK_CREATE` syscall which is already covered by `haveBPFLink` and isn't sufficient here.

When passing an invalid FD to the syscall, the returned error won't actually be EBADF, but ENODEV.

### DIFF
--- a/link/syscalls.go
+++ b/link/syscalls.go
@@ -127,18 +127,32 @@ var haveProgQuery = internal.NewFeatureTest("BPF_PROG_QUERY", "4.15", func() err
 })
 
 var haveTCX = internal.NewFeatureTest("tcx", "6.6", func() error {
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type:    ebpf.SchedCLS,
+		License: "MIT",
+		Instructions: asm.Instructions{
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+	})
+
+	if err != nil {
+		return internal.ErrNotSupported
+	}
+
+	defer prog.Close()
 	attr := sys.LinkCreateTcxAttr{
 		// We rely on this being checked during the syscall.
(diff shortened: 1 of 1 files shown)


rgo3:
Sorry for the mess! Should have caught all of this with the original PR in #1294 :disappointed:

link/syscalls.go
```diff
@@ -127,18 +127,32 @@ var haveProgQuery = internal.NewFeatureTest("BPF_PROG_QUERY", "4.15", func() err
...
+	defer prog.Close()
 	attr := sys.LinkCreateTcxAttr{
 		// We rely on this being checked during the syscall.
-		// With an otherwise correct payload we expect EBADF here
+		// With an otherwise correct payload we expect ENODEV here
 		// as an indication that the feature is present.
-		ProgFd:        ^uint32(0),
 		TargetIfindex: ^uint32(0),
+		ProgFd:        uint32(prog.FD()),
 		AttachType:    sys.AttachType(ebpf.AttachTCXIngress),
```
lmb, line 150: The reason is that AttachType is checked after ifindex / progfd is checked?

rgo3: First the program type is checked, then the AttachType. See https://elixir.bootlin.com/linux/v6.7/source/kernel/bpf/syscall.c#L4991. `tcx_link_attach` will then return `ENODEV` because of the invalid ifindex. On kernels without tcx support `EINVAL` will be returned because `BPF_LINK_CREATE` doesn't know how to create links for programs with type `SchedCLS`.

lmb (approved):

Issue #1218, opened 2023-11-10
btf: move wire type definitions into sys


lmb:
btf_types.go contains a bunch of hand written structs that map to types defined in the kernel's UAPI. We should replace these with auto generated types in sys package and use those from btf. This guarantees that we match the kernel 1:1 and also ensures that there is no implicit padding present.

- btfType
- btfEnum
- btfArray
- btfMember
- btfParam
- btfVar
- btfVarSecInfo
- btfDeclTag
- btfEnum64

The changes need to be made here: https://github.com/cilium/ebpf/blob/0acd95c6c83cabeff32e94594cd4b7de6e7df4a2/internal/cmd/gentypes/main.go#L157-L208

amiremohamadi:
I see bunch of helper functions and methods in `btf_types.go `. are we gonna keep them inside that file and just auto generate the structs inside `sys` package?

lmb:
We could move those into a separate file in `sys`. That way we can have manually written methods on auto generated types.

Mahdi-BZ:
What is the status of this? the PR is in draft for a year

PR #1304, opened 2024-01-10
btf/cmd/genbtftypes: (#1218) a generator that generates btf_types.go

OmarTariq612:
This is guaranteed to match the kernel struct defenitions.

I've tried to generate those structs in the `sys` package (`sys/types.go`) but there are private methods so I've tried to write those method manually after generating the types but there were other dependencies (constants like `btfMagic`, types like `FwdKind`).

Related to #1218


ti-mo:
@OmarTariq612 What's the benefit of having these be auto-generated? This adds 10x the amount of code to maintain to generate a handful of structs with a few fields, and these rarely change. I think @lmb's comment in the issue was just about moving the struct definitions into `sys`, not auto-generating them. Not all code in `sys` is autogenerated.

lmb (changes requested):
Hi @OmarTariq612, thanks for your PR :) Contrary to what Timo says, I did mean to auto generate the types. I think he has a point that having a separate binary to do the generation is not necessary.

The code to generate the types can just live in `gentypes`, and the generated types can be move to and exported from `internal/sys` instead of `btf`. The btf types can just go into `sys/types.go` as well. Is that something you're up for?

OmarTariq612:
I closed the PR by mistake, I'm still working on this, and I will reopen it when I'm done

OmarTariq612:
@ti-mo the migration is not easy, generating those types in `internal/sys` forces us to implement the methods there and those methods depend on (ex: `btf.FwdKind`, `internal.DiscardZeroes`, ...), trying to use those from `internal/sys` introduces a cycle.

OmarTariq612:
@ti-mo the migration is not easy, generating those types in `internal/sys` forces us to implement the methods there and those methods depend on (ex: `btf.FwdKind`, `internal.DiscardZeroes`, ...), trying to use those from `internal/sys` introduces a cycle.

PR #1308, opened 2024-01-15
For s390x arch the auto generated go object's name needs to have s390x prefix

### LINKED ISSUE #1307: bpf2go issues with s390x arch
### Describe the bug

```markdown
I have an application that build amd64,arm64,ppc64le and s390x arch the bpf2go generate is successful and the auto generated files are created for all arch

here is the command I used

//go:generate bpf2go -cc $BPF_CLANG -cflags $BPF_CFLAGS -target amd64,arm64,ppc64le,s390x

here is list of binaries created

ls  pkg/ebpf/
bpf_bpfeb_s390.go  bpf_bpfel_arm64.go  bpf_bpfel_powerpc.go  bpf_bpfel_x86.go
bpf_bpfeb_s390.o   bpf_bpfel_arm64.o   bpf_bpfel_powerpc.o   bpf_bpfel_x86.o

Now when I build the userspace golang for the different arch it builds for all arch except s390x
it fails when it tries to reference the auto generate go package for s390x


### How to reproduce

The quickest way to reproduce this 
- git clone https://github.com/netobserv/netobserv-ebpf-agent
- GOARCH=s390x  make compile

### Debugging
when turn on go build -x -v options and compare working and none working

-arm64
EOF
cat >/tmp/go-build3473788767/b145/embedcfg << 'EOF' # internal
{
	"Patterns": {
		"bpf_bpfel_arm64.o": [
			"bpf_bpfel_arm64.o"
		]
	},
	"Files": {
		"bpf_bpfel_arm64.o": "/home/mmahmoud/go/src/netobserv-ebpf-agent/pkg/ebpf/bpf_bpfel_arm64.o"
	}
}EOF
/usr/local/go/pkg/tool/linux_amd64/compile -o $WORK/b145/_pkg_.a -trimpath "$WORK/b145=>" -p github.com/netobserv/netobserv-ebpf-agent/pkg/ebpf -lang=go1.20 -complete -buildid HNBc_4x3RVDmb4k0dkDf/HNBc_4x3RVDmb4k0dkDf -goversion go1.20.10 -c=4 -nolocalimports -importcfg $WORK/b145/importcfg -embedcfg $WORK/b145/embedcfg -pack ./pkg/ebpf/bpf_bpfel_arm64.go ./pkg/ebpf/tracer.go

-s390x

/usr/local/go/pkg/tool/linux_amd64/compile -o $WORK/b147/_pkg_.a -trimpath "$WORK/b147=>" -p github.com/netobserv/netobserv-ebpf-agent/pkg/ebpf -lang=go1.20 -complete -buildid I8IGR_iTcXeniUkIzrN6/I8IGR_iTcXeniUkIzrN6 -goversion go1.20.10 -c=4 -nolocalimports -importcfg $WORK/b147/importcfg -pack ./pkg/ebpf/tracer.go

its clear for s390x the right dependency pkg isn't there that is why it failed

### Version information
go version go1.20.10 linux/amd64
github.com/cilium/ebpf v0.12.3

### PR DESCRIPTION
**Pull request by msherif1234**
`s390` is not a valid `GOARCH` changing yje auto generated go file to use `s390x` arch instead.
fix #1307

### DIFF
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -309,7 +309,12 @@ func (b2g *bpf2go) convert(tgt target, goarches []goarch) (err error) {
 	}
 	stem := fmt.Sprintf("%s_%s", outputStem, tgt.clang)
 	if tgt.linux != "" {
-		stem = fmt.Sprintf("%s_%s_%s", outputStem, tgt.clang, tgt.linux)
+		// for s390 target GOARCH has to be s390x
+		if tgt.linux == "390" {
+			stem = fmt.Sprintf("%s_%s_%s", outputStem, tgt.clang, "s390x")
+		} else {
+			stem = fmt.Sprintf("%s_%s_%s", outputStem, tgt.clang, tgt.linux)
+		}
 	}
 
 	objFileName := filepath.Join(b2g.outputDir, stem+".o")
--- a/cmd/bpf2go/main_test.go
+++ b/cmd/bpf2go/main_test.go
@@ -187,6 +187,12 @@ func TestCollectTargets(t *testing.T) {
 			},
 		},
 		{
+			[]string{"s390x"},
+			map[target][]goarch{
+				{"bpfeb", "s390"}: linuxArchesBE["s390"],
+			},
+		},
+		{
 			[]string{"native"},
 			nativeTarget,
 		},

### CODE REVIEW

File: cmd/bpf2go/main.go [hunk 1]
```diff
@@ -66,7 +66,7 @@ var targetByGoArch = map[string]target{
 	"mips64p32":   {"bpfeb", ""},
 	"ppc64":       {"bpfeb", "powerpc"},
 	"s390":        {"bpfeb", "s390"},
-	"s390x":       {"bpfeb", "s390"},
+	"s390x":       {"bpfeb", "s390x"},
```
lmb, line 69: This is not correct since the second entry in `target` is used to define [`__TARGET_ARCH_s390`](https://github.com/libbpf/libbpf/blob/f81eef23b33c0dbf923e863a72ce51ea4d32e291/src/bpf_tracing.h#L11C34-L11C34), etc. There is no `__TARGET_ARCH_s390x` from what I can tell.

lmb (commented):
Can you try simply passing `-target s390x` instead of `-target s390`? I think that will do the right thing. Ignore the file name, the important bit is the build constraint contained within.

msherif1234:
> Can you try simply passing `-target s390x` instead of `-target s390`? I think that will do the right thing. Ignore the file name, the important bit is the build constraint contained within.

using `-target s390x` also generate files with _390.* which fail go build

cmd/bpf2go/main.go
msherif1234: u mean the define in bpf_tracing.h right I think the same target `_s390` seems to be used for both arch 32bits and 64bits at least that is what I saw in https://github.com/iovisor/bcc/commit/7d7b6e888a1fe0af1db037bed2300c16cf32994c

msherif1234: u are right reworked the diffs PTAL

lmb:
Can you copy and paste the error you get when using `-target s390x` please?

msherif1234:
> Can you copy and paste the error you get when using `-target s390x` please?

the auto generated go file not getting embedded when build so GOARCH=s390x because the generated file is _s390.go instead of _s390x.go

```sh
GOARCH=s390x  make compile
### Compiling project
# github.com/netobserv/netobserv-ebpf-agent/pkg/ebpf
pkg/ebpf/tracer.go:55:22: undefined: BpfObjects
pkg/ebpf/tracer.go:84:15: undefined: LoadBpf
pkg/ebpf/tracer.go:401:48: undefined: BpfFlowId
pkg/ebpf/tracer.go:401:60: undefined: BpfFlowMetrics
pkg/ebpf/tracer.go:405:23: undefined: BpfFlowId
pkg/ebpf/tracer.go:405:35: undefined: BpfFlowMetrics
pkg/ebpf/tracer.go:406:9: undefined: BpfFlowId
pkg/ebpf/tracer.go:407:16: undefined: BpfFlowMetrics
pkg/ebpf/tracer.go:452:78: undefined: BpfObjects
pkg/ebpf/tracer.go:512:18: undefined: BpfObjects
pkg/ebpf/tracer.go:407:16: too many errors
make: *** [Makefile:131: compile] Error 1
```

I shared the build logs in issue #1307

lmb:
Okay, in that case the issue is that `s390` is a valid build tag. The go toolchain interprets `_s390.go` as an additional constraint to satisfy, which of course doesn't work. Oh boy.

msherif1234:
Yes GO doesn't consider `s390` as vaild GOARCH https://go.dev/doc/install/source#environment
that is why in my changes I tried to replace _s390.* extenstion with _s390x.* to get things working

florianl:
Just my thought looking into this a bit more:
The resulting Go filename is created by the following statement:
https://github.com/cilium/ebpf/blob/bfbacc1c6144d3c5beec82f2cdb451dc4204ee5f/cmd/bpf2go/main.go#L310

Here the element `clang` is used from the struct `target` which fits most use cases. But for GOARCH `s390x` it doesn't work, as `clang` can't handle the suffix `x` and so there is a difference between clangs `__TARGET_ARCH_xzy` and GOARCH.

At the moment, I think, it is not possible to determine GOARCH from the value combinations of `target`. So maybe one simple approach might be to extend the `target` struct with the GOARCH information and adapt creation of `goFileName` via `stem` accordingly.

lmb:
@florianl take a look at https://github.com/cilium/ebpf/commit/00aa2c3db9b0f8f7e9bdd5f364b3f952d2bb8ac9 which hopefully clarifies goarch vs clang target vs linux target. The way files are generated are as follows:

- The file name suffix is `_$linux_$clang.go` or `_$clang.go`. This is because any `(clang, linux)` target tuple can have multiple GOARCH associated with it.
- The actual build constraints are embedded in the file using `// go:build` since that is more flexible than just the file suffix tag that the Go toolchain also supports.

The problem is that the clang target `s390` is also a valid GOARCH, and the toolchain interprets the filename as adding a build constraint on `GOARCH=s390`. This of course fails.

I think the fix is to change the way we name files, which unfortunately is a breaking change. I'll try to prep a PR.

msherif1234:
Thanks @lmb pls cc me in your new PR

msherif1234:
this PR no longer needed a new PR by @lmb will be generated to fix this issue

PR #1312, opened 2024-01-17
cmd/bpf2go: fix s390x target

lmb:
The current format of output filenames for a specific GOARCH is

    <stem>_<clang target>_<linux target>.go

The s390x GOARCH has a linux target of "s390", which leads to a file name like

    foo_bpfeb_s390.go

The problem is that "_s390" is interpreted as an additional build constraint by the Go toolchain. In effect, the will only be included in the build if GOARCH is s390 and s390x which is of course not possible.

Fix this by changing the output filename format and add some code which removes files with the old naming schema. There is a small risk that this might in the future delete files that we haven't written, but the chances of that seem low enough.

Extend the bpf2go tests to ensure that we've generated the right build tags for arm64, amd64 and s390x.

Fixes https://github.com/cilium/ebpf/issues/1307


lmb:
@msherif1234 can you give this a try please?

cc @florianl

msherif1234:
sure do u mind share wih me the best way to run ur version of bpf2go 
I tried to update my go.mod to point to ur fork:branch 
`replace github.com/cilium/ebpf v0.12.3 => ../lmb-ebpf`
but that still not pointing to the new bpf2go , can u suggest a better way ?

lmb:
Try `go mod edit -replace=github.com/cilium/ebpf=github.com/lmb/ebpf@bpf2go-s390-fix && go mod tidy`.

msherif1234:
tried the above but no luck I think pls proceed with the merge and I will test from latest and let u know if there are any issues, Thanks for the quick fix!!

cmd/bpf2go/main.go
```diff
@@ -307,9 +307,17 @@ func (b2g *bpf2go) convert(tgt target, goarches []goarch) (err error) {
...
 	}
+
+	// The output filename must not match any of the following patterns:
+	//
+	//     *_GOOS
+	//     *_GOARCH
+	//     *_GOOS_GOARCH
+	//
+	// Otherwise it is interpreted as a build constraint by the Go toolchain.
 	stem := fmt.Sprintf("%s_%s", outputStem, tgt.clang)
```
florianl, line 318: With `tgt.clang` filling in the `_GOARCH` value https://github.com/cilium/ebpf/issues/1307 will not be resolved.
While `go tool dist list` does not list `s390` as supported architecture, [internals of Go](https://github.com/golang/go/blob/f19f31f2e7c136a8dae03cbfe4f8ebbb8b54569b/src/internal/goarch/goarch.go#L39) still are `s390` aware and make a difference between `s390` and `s390x`.
As `tgt.clang` should be `s390` at this point, it will generate a file with the according name and Gos build constraints will not consider it for `s390x`.

lmb: No, clang target is only ever `bpfel` or `bpfeb`. This will break if we ever need to support clang targets other than those two which then end up matching go constraints, but why would we?

lmb: List of clang targets:
```
$ clang -print-targets

  Registered Targets:
    aarch64     - AArch64 (little endian)
    aarch64_32  - AArch64 (little endian ILP32)
    aarch64_be  - AArch64 (big endian)
    amdgcn      - AMD GCN GPUs
    arm         - ARM
    arm64       - ARM64 (little endian)
    arm64_32    - ARM64 (little endian ILP32)
    armeb       - ARM (big endian)
    avr         - Atmel AVR Microcontroller
    bpf         - BPF (host endian)
    bpfeb       - BPF (big endian)
    bpfel       - BPF (little endian)
    hexagon     - Hexagon
    lanai       - Lanai
    loongarch32 - 32-bit LoongArch
    loongarch64 - 64-bit LoongArch
    mips        - MIPS (32-bit big endian)
    mips64      - MIPS (64-bit big endian)
    mips64el    - MIPS (64-bit little endian)
    mipsel      - MIPS (32-bit little endian)
    msp430      - MSP430 [experimental]
    nvptx       - NVIDIA PTX 32-bit
    nvptx64     - NVIDIA PTX 64-bit
    ppc32       - PowerPC 32
    ppc32le     - PowerPC 32 LE
    ppc64       - PowerPC 64
    ppc64le     - PowerPC 64 LE
    r600        - AMD GPUs HD2XXX-HD6XXX
    riscv32     - 32-bit RISC-V
    riscv64     - 64-bit RISC-V
    sparc       - Sparc
    sparcel     - Sparc LE
    sparcv9     - Sparc V9
    systemz     - SystemZ
    thumb       - Thumb
    thumbeb     - Thumb (big endian)
    ve          - VE
    wasm32      - WebAssembly 32-bit
    wasm64      - WebAssembly 64-bit
    x86         - 32-bit X86: Pentium-Pro and above
    x86-64      - 64-bit X86: EM64T and AMD64
    xcore       - XCore
```

florianl: sorry for coming back on this just now.
I tested this PR again with the issue described in https://github.com/cilium/ebpf/issues/1307. And I think, I now also got the origin of my confusion.  As I tested this PR with code with the linked repository from https://github.com/cilium/ebpf/issues/1307, I did not clean up the and removed the checked-in compiled binary objects. So I ended up with a strange mix that led to my confusion.
Testing it again now, I can confirm, that this PR works as intended.

florianl (approved):

cmd/bpf2go/main.go
florianl: @msherif1234  I tested this PR by checking out this PR, compiling bpf2go and replace the bpfgo in the generate instruction in https://github.com/netobserv/netobserv-ebpf-agent/blob/504add4c32d1259afabe2d2ab24664a684109d5c/pkg/ebpf/tracer.go#L29 with the compiled version from this PR

msherif1234:
verified the fix with the new version and it worked as expected thanks @lmb !!

PR #1314, opened 2024-01-19
add kernel 6.7

lmb:
automatically regenerate tests when kernel deps change

    Invoke make container-all after updating kernel dependencies. This means
    there is one less step necessary to support a new kernel.

    Fix make generate so that syscall wrappers are always generated first.
    Otherwise go generate on examples may fail due to missing symbols in
    internal/sys.

CI: switch from 6.6 to 6.7

    Test against the latest mainline release instead of the latest stable. This
    will allow testing netkit devices.


lmb:
cc @hemanthmalla this is what's necessary to bump the kernel. I've streamlined the process a little as well.

PR #1318, opened 2024-01-26
Create merged coverage to html

### LINKED ISSUE #1256: CI: store merged coverage as an artifact
Running our test suite creates coverage information in `coverage.out`:

```
$ ./run-tests.sh 6.1
Fetching linux-6.1-amd64.tgz
Fetching linux-6.1-amd64-selftests-bpf.tgz
...
$ ls -lh coverage.out 
-rw-r--r--. 1 lorenz lorenz 521K Dec  6 14:12 coverage.out
```

That can be turned into a html report like so:

```
$ go tool cover -html=coverage.out -o coverage.html
$ ls -lh coverage.html 
-rw-r--r--. 1 lorenz lorenz 899K Dec  6 14:56 coverage.html
```

We should integrate this into our CI:

- Upload each generated `coverage.out` as an artifact for that job
- Collect all `coverage.out` from the various jobs and merge them using `go tool covdata merge`
- Generate a single .html which shows overall test coverage and upload attach that to the build somehow
- Brownie points if its possible to look at the coverage html by just clicking on a link instead of downloading + extracting

amit-62:
It generates`coverage.out` as an artifact and merges it to `coverage.html`.
fixes #1256


.github/workflows/ci.yml
```diff
@@ -197,3 +197,31 @@ jobs:
...
+    name: Generate code coverage
+    runs-on: ubuntu-latest
+    needs: build-and-lint
+    timeout-minutes: 10   
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Run Tests
+        run: ./run-tests.sh 6.1
```
ti-mo, line 211: This doesn't make much sense. `vm-test` has a test matrix that runs this against a wide range of kernel versions, and we need test coverage for each of them.

The point is to add coverage reporting to the existing `vm-test`, `test-on-prev-go` and `test-on-arm64` jobs. Ideally, this would all be done in a single step to keep the size of the yaml a bit manageable:

```yaml
      - run: go tool covdata merge -o coverage-merged.out
      - run: go tool cover -html=coverage-merged.out -o coverage.html

      - name: Upload Coverage Report
        uses: actions/upload-artifact@v3
        with:
          name: coverage-report
          path: |
            ./coverage-merged.out
            ./coverage.html
```

ti-mo (changes requested):

.github/workflows/ci.yml
amit-62: Will I have to add this to all three job or it can be done in another separate job?

ti-mo: If we want coverage for all three jobs (we do), then yes, it needs to be in all three. :slightly_smiling_face:

lmb:
Closing this for now.

Issue #1166, opened 2023-10-12
internal: stop parsing /proc/self/auxv once (Go 1.21)


lmb:
From Go 1.21 is possible to extract the auxiliary vector from the runtime: https://cs.opensource.google/go/x/sys/+/refs/tags/v0.13.0:cpu/runtime_auxv_go121.go This will allow us to get rid of a bunch of custom code which parses `/proc/self/auxv` once 1.21 is the minimum version required by the lib.

We can just vendor the linked file and call `runtime_getAuxv`.

kwakubiney:
Could take a stab at this

paulcacheux:
I'm going to draft a PR for this if it's OK for you

PR #1319, opened 2024-01-26
add support for reading auxv from Go runtime

paulcacheux:
Fixes https://github.com/cilium/ebpf/issues/1166

This PR extracts auxv related code to its own file. It also creates a common interface to read auxv pairs. The new go 1.21 runtime function to fetch auxv is then used to create a new auxv provider, and the old file based one is kept around for tests.

This PR also adds a test ensuring the file and runtime based auxv reader actually read the same pairs.

If we decide we can remove `TestAuxvVDSOMemoryAddress` and `TestAuxvNoVDSO`, I would be open to simplifying drastically this PR and keeping only the runtime auxv provider. Open to feedback especially on this.

### DIFF
--- a/internal/auxv.go
+++ b/internal/auxv.go
@@ -1,0 +1,60 @@
+package internal
+
+import (
+	"errors"
+	"io"
+	_ "unsafe"
+)
+
+type auxvPairReader interface {
+	Close() error
+	ReadAuxvPair() (uint64, uint64, error)
+}
+
+// See https://elixir.bootlin.com/linux/v6.5.5/source/include/uapi/linux/auxvec.h
+const (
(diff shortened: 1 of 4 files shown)


lmb:
@paulcacheux you could rebase this now if you want.

lmb (changes requested):
I really like the idea of testing the runtime provided values against reading from /proc/auxv. As far as I understand we don't rely on reading from /proc except in tests now? How about moving the proc reader into a `_test.go` file?

paulcacheux:
Yes good idea, the first version of this PR kept support for older version of Go, meaning that the file provider was actually used. But now that we support only go 1.21 and up I will move it to a test file

PR #1323, opened 2024-01-31
fix: add testcast for PutInteger

atgane:
Recreated the PR for [#1277](https://github.com/cilium/ebpf/pull/1277).


internal/kconfig/kconfig.go
```diff
@@ -263,12 +263,25 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
 		return fmt.Errorf("invalid boolean value: %d", n)
 	}
 
+	if len(data) < int(integer.Size) {
+		return fmt.Errorf("can't fit an integer of size %d into a byte slice of length %d", integer.Size, len(data))
+	}
+
 	switch integer.Size {
 	case 1:
+		if integer.Encoding == btf.Signed && (n > math.MaxInt8 && n < math.MaxUint64-math.MaxInt8) {
```
ti-mo, line 272: Sorry, but again I don't understand why it needs to be so complicated:
```suggestion
		if integer.Encoding == btf.Signed && (n > math.MaxInt8 || n < math.MinInt8) {
```

internal/kconfig/kconfig.go
```diff
@@ -263,12 +263,25 @@ func PutInteger(data []byte, integer *btf.Int, n uint64) error {
...
+			return fmt.Errorf("can't represent %d as a signed integer of size %d", n, integer.Size)
+		}
 		internal.NativeEndian.PutUint16(data, uint16(n))
 	case 4:
+		if integer.Encoding == btf.Signed && (n > math.MaxInt32 && n < math.MaxUint64-math.MaxInt32) {
+			return fmt.Errorf("can't represent %d as a signed integer of size %d", n, integer.Size)
+		}
 		internal.NativeEndian.PutUint32(data, uint32(n))
 	case 8:
 		internal.NativeEndian.PutUint64(data, uint64(n))
```
ti-mo, line 287: This also needs a check, since it's possible for n to overflow/underflow an int64.

ti-mo (changes requested):

internal/kconfig/kconfig.go
atgane: I'll add some test code for the boundary value of an int32.

internal/kconfig/kconfig.go
atgane: This is because n is of type uint and MinInt8 is of type int, so there was a compiler error in the comparison operation. However, when I cast n to int64, I saw that there was no problem when n exceeded MaxInt64 of int64. I will create a pull request with a similar format by converting n to int64.

ti-mo: > I will create a pull request with a similar format by converting n to int64.

Please, no more pull requests. Just amend and force-push to this PR's branch, please.

atgane: I forced push.

internal/kconfig/kconfig.go
ti-mo: Never mind this comment, wasn't thinking straight.

ti-mo (approved):

ti-mo:
@atgane I've squashed your commits and cut down on the size of the test suite. Try to avoid testing error strings, it's of questionable benefit and makes errors annoying to modify.

I split off a few negative tests into TestPutIntegerError.

atgane:
> @atgane I've squashed your commits and cut down on the size of the test suite. Try to avoid testing error strings, it's of questionable benefit and makes errors annoying to modify.
> 
> I split off a few negative tests into TestPutIntegerError.

Thank you for your feedback

PR #1332, opened 2024-02-07
dependabot: onboard github actions upgrades

paulcacheux:
We currently have a lot of github actions missing upgrades, resulting in some 
```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-go@v4 ...
```
[warnings](https://github.com/cilium/ebpf/actions/runs/7817377529).

This PR onboards the github actions workflows into dependabot, which should make our life updating those easier.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,7 @@ updates:
       # dependencies only appearing in Pipfile.lock.
       - dependency-name: "*"
         dependency-type: "direct"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


brycekahle:
Careful of this issue: https://github.com/actions/checkout/issues/1590

paulcacheux:
I think this issue only affects workflows that run in a container, which we are not doing. In all cases if we think we do not want a bump we can simply close the dependabot PR

ti-mo (approved):
@lmb Any objections?

lmb:
Ph boy :sweat_smile: Is there a way to group these into one PR?

paulcacheux:
Hopefully this is only happening because we need to catch up. Yes there is a way to group PR, let me take a look at the doc and open a PR

PR #1345, opened 2024-02-15
Adding additional tests for parseCPUsFromFile to cover parseCPUs test…

zedGGs:
Why is this important ? Why to add additional tests ? 

goal is to add better test coverage for `parseCPUsFromFile` from actual file, 
test cases used are from tests in `parseCPUs`

gentle ping for @lmb, if adding these tests make sense I can make required updates to your suggestions, 
thank you !

### DIFF
--- a/cpu_test.go
+++ b/cpu_test.go
@@ -1,6 +1,8 @@
 package ebpf
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


ti-mo:
Hi @zedGGs, does this address an existing issue? I think adding a single test for `parseCPUsFromFile` would be sufficient here, the test cases you added look copied from `TestParseCPUs`. There's no need to exercise `os.ReadFile`, that one already has plenty of tests and is known to work. :slightly_smiling_face:

lmb:
Closing for now.

PR #1347, opened 2024-02-16
fix trusted "Tag breaking changes" workflow

paulcacheux:
This has been broken by https://github.com/cilium/ebpf/commit/4d619bbc8b492e0d2f2a37c369f444da39c9a8db, and the bump to a version higher than v5 (see https://github.com/actions/github-script?tab=readme-ov-file#v5).

Please note: this fix is pretty difficult to test since the job itself uses the workflow definition from the main branch to run

### DIFF
--- a/.github/workflows/trusted.yml
+++ b/.github/workflows/trusted.yml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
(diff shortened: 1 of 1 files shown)


lmb:
Thanks, agreed that GH Actions are a pain this way.

lmb (approved):

PR #1349, opened 2024-02-19
MapIterator support for Queues/Stacks + further testing + example

s41m0n:
Dear community, with this PR I would like to introduce the following contributions.

I'll be available for further changes/updates.

## MapIterator support for Queue/Stack maps

Ther `MapIterator.Next` would normally fail with maps of type Queue or Stack for two reasons:

1. Lookup for Queue/Stack acts as a peek method, returning always the head value without removing it. It additionally fails when provided with a non-nil key
1. LookupAndDelete is the method for retrieving values in a Queue/Stack map.

With that in mind, instead of adding further workarounds and control to each individual call to the previous `MapIterator.Next` function or to each `Map.Lookup` (potentially a non-negligible overhead), I'd propose splitting it into two internal methods. In this case, the program logic of calling all the other functions (e.g., Lookup, BatchLookup that is not supported) to a Queue/Stack is preserved, throwing the expected error.

The proposed solution preserves a maximum amount of retrieved items, to avoid an infinite loop as the Queue/Stack may be continuously populated while values are retrieved.

## Extended tests for Queue/Stack maps

1. Add further tests to the Lookup method (peek) with Queue maps, previously missing
1. Add full tests for Stack maps, previously missing

## Example of using a Queue map

I'd introduce an example that demonstrates the usage of a Queue map within an XDP program. The program parses IPv4 packets, retrieving the source address and pushing it, alongside a computed `ktime_ns` into the Queue. In the userspace program, the Queue is periodically emptied, formatting its content into a human-readable string containing the IPv4 address and the timestamp into a go `Time` struct.

## Further Proposal

Would it be useful to introduce a function to parse a `ktime_ns` value into `Time` struct as a utility function of this library? In case, please don't hesitate to point me to the right file where this change could be performed and I can do it. I'd use a similar method that retrieves the boot uptime time once (to prevent continuous syscalls) and simply return it incremented by the `ktime_ns` argument value.


examples/queue/main.go
```diff
@@ -0,0 +1,115 @@
...
+// an HashMap with a small __u64 arrays associated to each IPv4 address (key).
+// In both the two ways it is possible to lose some packet if (a) queue is not large
+// enough or the packet processing time is slow or (b) if the associated array is
+// smaller than the actual received packet from an address.
+// The userspace program (Go code in this file) prints the contents
+// of the map to stdout every second, parsing the raw structure into a human-readable
+// IPv4 address and Unix timestamp.
+// It is possible to modify the XDP program to drop or redirect packets
+// as well -- give it a try!
+// This example depends on bpf_link, available in Linux kernel version 5.7 or newer.
```
dylandreimerink, line 15: Like you mention yourself, this isn't a prime example of queues or stacks. I understand you want an example to showcase those map types. A good example is small and doesn't add to much extra info. I honestly don't think it adds much to your PR, I suggest dropping it.

map.go
```diff
@@ -1506,17 +1506,42 @@ func newMapIterator(target *Map) *MapIterator {
...
+
+	// For Queue/Stack map block the iteration after maxEntries
+	// to avoid potential infinite loops
+	// (values can be pushed to map while doing pop)
+	if mi.count == mi.maxEntries {
+		mi.err = fmt.Errorf("%w", ErrIterationAborted)
+		return false
+	}
+
+	mi.count++
```
dylandreimerink, line 1527: I don't think this check works as intended. If a user indeed uses a stack or queue as kernel to userspace mechanism, and is constantly adding new objects, then you would expect to iterate for more than the maxEntries of the map.

I think we can just leave this out.

map.go
```diff
@@ -1575,6 +1600,26 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+//
+// Returns false if there are no more entries. You must check
+// the result of Err afterwards.
+//
+// See Map.Get for further caveats around valueOut.
+func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
+	if mi.target.typ.isQueueStack() {
+		return mi.nextQueueMap(valueOut)
+	}
+	return mi.next(keyOut, valueOut)
```
dylandreimerink, line 1620: I think that we should implement this slightly differently. Now that we have multiple iterator implementations (and I expect more to come), it makes sense to start using an interface with multiple implementations instead.

I suggest renaming `MapIterator`, and making it unexported. Then add a `MapIterator` interface with the `Err() error` and `Next(keyOut, valueOut interface{}) bool` methods. And add your logic as an new stack/queue iterator struct. 

Then make `Map.Iterate` pick the correct iterator.

map_test.go
```diff
@@ -839,6 +859,74 @@ func TestMapQueue(t *testing.T) {
...
+		t.Error("Want value 42, got", v)
+	}
+
+	if err := m.LookupAndDelete(nil, &v); !errors.Is(err, ErrKeyNotExist) {
+		t.Fatal("Lookup and delete on empty Stack:", err)
+	}
+
+	if err := m.Lookup(nil, &v); !errors.Is(err, ErrKeyNotExist) {
+		t.Fatal("Lookup (Peek) on empty Stack:", err)
+	}
```
dylandreimerink, line 929: I feel like this test is more about testing how the map type is implemented than it does the library. Something that is more at home in the kernel selftests. Same goes for the additions in the queue test. I would simply not bother with these.

On the other hand, I see no tests for the new iteration code. That is what I am interested in seeing, to check that the iterator actually works and stays working.

dylandreimerink (changes requested):
Thank you very much for putting in the work. Though I would like to request some changes.

map_test.go
s41m0n: Agree, proceeding this way :+1: 

I would just leave one additional lookup (peek) test on the Queue for completeness.

examples/queue/main.go
s41m0n: I see your point and I agree.
Let me know if the `ktime -> Unix time` conversion utility function could be useful somehow. I'd leave it up to the user and wouldn't insert it as part of the library.

map.go
s41m0n: I see. In that way, is the user that decides whether to stop after certain iterations or not.

map.go
s41m0n: That is perfectly fine by me.  I'm going to create two MapIterator implementations, one for key-value maps `indexedMapIterator`  and one for value-only maps `bufferedMapIterator` . Correct me if the naming is horrible (I also thought about `keylessMapIterator` and `keyedMapIterator`)

Concerning the PerCpu maps, I'd leave as it is, so that the old `MapIterator` (hence the newly `indexedMapIterator`) handles this case.

s41m0n:
> Thank you very much for putting in the work. Though I would like to request some changes.

Many thanks again Dylan for the feedback!

map.go
```diff
@@ -1486,10 +1490,34 @@ func marshalMap(m *Map, length int) ([]byte, error) {
...
+	//
+	// The method must be called after Next returns nil.
+	//
+	// For key-value maps, returns ErrIterationAborted if
+	// it wasn't possible to do a full iteration.
+	Err() error
+
+	// Next decodes the next key and value.
+	//
+	// In case of a value-only map (Queue and Stack), the key
```
ti-mo, line 1508: This is an implementation detail, not something the caller needs to take into account explicitly to correctly use the interface. Semantically, `keyOut` doesn't seem different from regular kv maps, so I'd drop this paragraph.

map.go
```diff
@@ -1486,10 +1490,34 @@ func marshalMap(m *Map, length int) ([]byte, error) {
...
+// MapIterator is the interface defining methods to iterate
+// through the content of the target map.
+//
+// See Map.Iterate.
+type MapIterator interface {
+	// Err returns any encountered error.
+	//
+	// The method must be called after Next returns nil.
+	//
+	// For key-value maps, returns ErrIterationAborted if
```
ti-mo, line 1502: I'd remove the 'for key-value maps' part. The interface shouldn't document the implementation. Generally, `ErrIterationAborted` means a full iteration wasn't possible.

map.go
```diff
@@ -1499,24 +1527,59 @@ type MapIterator struct {
 	err               error
 }
 
-func newMapIterator(target *Map) *MapIterator {
-	return &MapIterator{
+// keylessMapIterator iterates a Map defined with only values.
+// (Queue and Stack)
+//
+// See Map.Iterate.
+type keylessMapIterator struct {
```
ti-mo, line 1534: I'm not 100% convinced that making MapIterator an interface is currently worth it. Only `next()` is fundamentally different, and the kv vs. keyless split feels somewhat arbitrary. If we were planning to add many more different implementations, it would make more sense, but currently this is not being discussed.

See my comment on keylessMapIterator.next() below.

map.go
```diff
@@ -1499,24 +1527,59 @@ type MapIterator struct {
...
-// Next decodes the next key and value.
-//
-// Iterating a hash map from which keys are being deleted is not
-// safe. You may see the same key multiple times. Iteration may
-// also abort with an error, see IsIterationAborted.
+// next decodes the next value from a map of type Queue or Stack.
 //
 // Returns false if there are no more entries. You must check
 // the result of Err afterwards.
+func (mi *keylessMapIterator) next(_, valueOut interface{}) bool {
```
ti-mo, line 1557: I'd make this `MapIterator.pop()` and make `MapIterator.Next()` call `next()` or `pop()` depending on the outcome of `isQueueStack()`.

types.go
```diff
@@ -102,6 +102,12 @@ func (mt MapType) hasPerCPUValue() bool {
 	return mt == PerCPUHash || mt == PerCPUArray || mt == LRUCPUHash || mt == PerCPUCGroupStorage
 }
 
+// isQueueStack returns true if the Map is a Queue (BPF_MAP_TYPE_QUEUE)
+// or Stack (BPF_MAP_TYPE_STACK)
+func (mt MapType) isQueueStack() bool {
```
ti-mo, line 107: This is a shortcut for the map's type instead of its semantics. Maybe `needsPop()` would be more descriptive?

ti-mo (changes requested):
Thanks for the patches! I'm not completely convinced we want to make MapIterator an interface (yet).

map.go
lmb: I think that was how the code was structured originally, until Dylan suggested using an interface (after discussing it with me). I guess it doesn't make much difference either way.

lmb (commented):
I have a question regarding semantics:
- What happens when you do a plain Lookup on a queue or stack? Does that function like a peek? Or does it return an error?
- Are there other map types which support LookupAndDelete? Are there ones which support passing a Key argument?

map.go
s41m0n: I agree, probably too many comments aren't needed in this case.

s41m0n:
Thanks to all the reviewers for the feedback. 
I guess the only thing left to decide is whether to proceed with one of the two proposed solutions:

1. Creating a simple method `pop` that would be called instead of `next`, as for the solution proposed in the previous commit.
2.  Creating an interface to support multiple MapIterators, as for the current commit under review.

lmb:
Have you had the time to look into my questions in https://github.com/cilium/ebpf/pull/1349#pullrequestreview-1951908908 ? I think they will help us decide what the implementation looks like.

s41m0n:
> Have you had the time to look into my questions in [#1349 (review)](https://github.com/cilium/ebpf/pull/1349#pullrequestreview-1951908908) ? I think they will help us decide what the implementation looks like.

@lmb  I apologize, I missed the message.

1. According to https://docs.kernel.org/bpf/map_queue_stack.html and some quick tests of the code, Queues/Stack maps support the following operations.
    -  lookup (peek): returns the value at the head without removing it. If no value exists, `ErrKeyNotExist` is returned.
    -  update (push)
    -  delete (pop)
In addition, `batch operations` are not supported. Also, `NextKey` is not supported: it either returns `can't marshal key: <type> doesn't marshal to 0 bytes` or `next key: invalid argument` when called with `nil` (these maps have a key of size 0).
1.  On the other hand, other maps like `Array` and `Hash` support `LookupAndDelete`,  and the method requires a non-null `key` as a parameter.

dylandreimerink:
> Queues/Stack maps support the following operations.

Clarification, for stacks/queues:
`BPF_MAP_LOOKUP_ELEM` -> peek
`BPF_MAP_LOOKUP_AND_DELETE_ELEM` -> pop
`BPF_MAP_UPDATE_ELEM` -> push

> On the other hand, other maps like Array and Hash support LookupAndDelete, and the method requires a non-null key as a parameter.

Only hash maps(and variants: LRU, PER_CPU, LRU_PER_CPU) have support for  `BPF_MAP_LOOKUP_AND_DELETE_ELEM`  since v5.14. Array maps don't seem to support it since you can't delete from an array map.

Also interesting is that `BPF_MAP_LOOKUP_AND_DELETE_BATCH` which is the batch version is only supported by hash maps and not by stacks or queues and was added before the non-batch version was supported in v5.6.

> Are there ones which support passing a Key argument?

Yes, the hashmap supports passing a key value. Actually, even before v5.14, a valid pointer must be provided to the key field. The kernel always checks it and copies it into kernel memory, just to discard it. After v5.14 it of course starts to actually get passed to the map implementation.

lmb:
Okay! That was very helpful, thanks to both of you. The peek operations only return the top of the stack / first item in the queue? And currently doing `.Iterate()` on a queue or stack doesn't work at all since we pass a key?

I'm currently leaning towards adding a new method `Map.Drain() *MapIterator`. Initially this could only support queues / stacks but later on we may be able to extend it to hash maps. The benefit of a new method is that it makes it clearer that we're actually deleting items here. We can put the implementation into the same `MapIterator` struct for now as Timo suggested. I think that we should take  `MapIterator.Next(nil, &value)` when draining (and not ignore the key argument as currently done.)

dylandreimerink:
> The peek operations only return the top of the stack / first item in the queue? And currently doing .Iterate() on a queue or stack doesn't work at all since we pass a key?

Yes, peeking only returns the top of the stack. Any key provided will be ignored as far as I understand. But as @s41m0n mentioned the `BPF_MAP_GET_NEXT_KEY` op would return an error if using the normal iterator.

s41m0n:
Thanks for the info-gathering and brainstorming. I just pushed a possible implementation of what has been discussed.
Let me know your opinions :)

map.go
```diff
@@ -1545,12 +1557,16 @@ func newMapIterator(target *Map) *MapIterator {
...
 
-// Next decodes the next key and value.
+// Next decodes the next key and value. If the iterator is created
+// through the Map.Drain API, the key and value are also removed.
 //
 // Iterating a hash map from which keys are being deleted is not
 // safe. You may see the same key multiple times. Iteration may
 // also abort with an error, see IsIterationAborted.
 //
+// Iterating a queue/stack map returns ErrIterationAborted, as only
```
lmb, line 1567: Does it make sense to document this on `Map.Iterate` instead?

map.go
```diff
@@ -1545,12 +1557,16 @@ func newMapIterator(target *Map) *MapIterator {
 	}
 }
 
-// Next decodes the next key and value.
+// Next decodes the next key and value. If the iterator is created
```
lmb, line 1560: Nit: do you think it's necessary to call this out again? It's already on Drain after all.

map.go
```diff
@@ -1559,10 +1575,17 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+}
 
-	// For array-like maps NextKey returns nil only after maxEntries
-	// iterations.
-	for mi.count <= mi.maxEntries {
+func (mi *MapIterator) nextIterate(keyOut, valueOut interface{}) bool {
+	// For array-like maps NextKey returns nil only after maxEntries iterations.
+	// For maps with keySize equal to 0 (Queue/Stack) we return ErrIterationAborted,
+	// since NextKey returns an error. In this case, Map.Drain should be used instead.
+	for mi.target.keySize != 0 && mi.count <= mi.maxEntries {
```
lmb, line 1588: I'm a bit unsure about using keySize like this, it's a very roundabout way of referring to queue / stack. What if there are maps in the future without a key size?

My personal preference would be to drop the special case and make sure that whatever error we get from `Queue.Drain().Next()` due to next key failing is sensible.

map.go
```diff
@@ -1614,6 +1637,85 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+	} else if mi.err != nil {
+		mi.err = fmt.Errorf("get next key: %w", mi.err)
+		return false
+	}
+
+	// falling back to sequential Lookup -> Delete in case LookupAndDelete
+	// is not supported (e.g., kernel < 5.14).
+	if mi.fallback {
+		mi.count++
+		mi.err = mi.target.Lookup(mi.cursor, valueOut)
```
lmb, line 1682: This doesn't have the same semantics as LookupAndDelete, since that guarantees that a key / value combo is only returned once if there are concurrent traversal. Those semantics seem useful to me. Would you mind just dropping the fallback?

map.go
```diff
@@ -1614,6 +1637,85 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+
+	// Always retrieve first key in the map. This should ensure that
+	// the whole map is traversed, despite concurrent insertion.
+	// The expected ordering might differ:
+	// - initial keys in map: `a -> b -> c`
+	// - call MapIterator.Next and retrieve key `a`
+	// - insert key `d` in map
+	// - retrieve all the remaining keys `d -> b -> c`
+	mi.err = mi.target.NextKey(nil, mi.cursor)
+	if errors.Is(mi.err, ErrKeyNotExist) {
```
lmb, line 1669: What error does a NextKey on a queue or stack return?

map.go
```diff
@@ -1614,6 +1637,85 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
 	return false
 }
 
+func (mi *MapIterator) nextDrain(keyOut, valueOut interface{}) bool {
+	// Check for maps without a key (e.g., Queue/Stack). In this case, when there is no
+	// more data, ErrKeyNotExist arise, but we gently stop the retrieval with no error.
+	if mi.target.keySize == 0 {
+		mi.err = mi.target.LookupAndDelete(keyOut, valueOut)
```
lmb, line 1644: This allows a user to provide a non-nil key even when we know that there probably shouldn't be one. I'm on the fence whether this is a good idea. Maybe we should validate that `key == nil` instead?

map.go
```diff
@@ -1614,6 +1637,85 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+		if mi.err != nil && errors.Is(mi.err, ErrNotSupported) || errors.Is(mi.err, unix.EINVAL) {
+			mi.fallback = true
+			return mi.nextDrain(keyOut, valueOut)
+		}
+		mi.count++
+	}
+
+	if errors.Is(mi.err, ErrKeyNotExist) {
+		// Same as in MapIterator.nextIterate.
+		return mi.nextDrain(keyOut, valueOut)
```
lmb, line 1703: Why does this recurse? To avoid the for loop? I think the for loop is easier to understand here because one just has to look for a break or return statement to figure out how the loop exits. For me that is more difficult with recursive functions.

map_test.go
```diff
@@ -1044,6 +1056,49 @@ func TestIterateEmptyMap(t *testing.T) {
...
+	for _, mapType := range []MapType{
+		Hash,
+		SockHash,
+		Queue,
+	} {
+		t.Run(mapType.String(), func(t *testing.T) {
+			keySize := uint32(4)
+
+			if mapType == Queue {
+				testutils.SkipOnOldKernel(t, "4.20", "map type queue")
```
lmb, line 1069: After dropping the fallback the whole test can just be 5.15 and above I guess?

map_test.go
```diff
@@ -1044,6 +1056,49 @@ func TestIterateEmptyMap(t *testing.T) {
 	}
 }
 
+func TestDrainEmptyMap(t *testing.T) {
```
lmb, line 1059: Is this test necessary? You could just try another Drain at the end of the other tests below.

map_test.go
```diff
@@ -1090,6 +1145,109 @@ func TestMapIterate(t *testing.T) {
...
+		}
+		if keys[0] != "hello" {
+			t.Error("Expected index 0 to be hello, got", keys[0])
+		}
+		if keys[1] != "world" {
+			t.Error("Expected index 1 to be hello, got", keys[1])
+		}
+	})
+
+	t.Run(Queue.String(), func(t *testing.T) {
```
lmb, line 1197: What about stack?

map_test.go
lmb, line 1: Can you take a look at how we use `qt.Assert`? Especially `qt.DeepEquals` for comparing slices. I think that'll let you shorten the tests a bit.

map_test.go
smagnani96: > What about stack?

As suggested in https://github.com/cilium/ebpf/pull/1349#discussion_r1502624529, probably Stack vs Queue it's a matter of implementation in kernel space, don't know if it's really needed for us to test it here.

map_test.go
smagnani96: I wanted to keep the same separation as for `TestIterateEmptyMap` https://github.com/cilium/ebpf/blob/c6205528d1f7c29552021af9bcd2f049dd92cabe/map_test.go#L1003

Let me know.

map.go
```diff
@@ -1573,26 +1588,28 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+	var key interface{}
 
-	// For array-like maps NextKey returns nil only after maxEntries
-	// iterations.
+	// For array-like maps NextKey returns nil only after maxEntries iterations.
 	for mi.count <= mi.maxEntries {
 		if mi.cursor == nil {
 			// Pass nil interface to NextKey to make sure the Map's first key
 			// is returned. If we pass an uninitialized []byte instead, it'll see a
 			// non-nil interface and try to marshal it.
```
dylandreimerink, line 1605: This comment should be updated, doesn't line up with the code anymore

dylandreimerink (changes requested):
I like the end result, but between the commits a lot happens that is canceled out later. So please squash all commits into one (or cleanup the git history some other way)

smagnani96:
This PR has been rebased to https://github.com/cilium/ebpf/pull/1559 so that the current PR preserve only the needed material for the `Map.Drain` API. The only commit that would remain and which we should refer to is https://github.com/cilium/ebpf/pull/1559.

smagnani96:
I removed from this API the support for the fallback sequential `Lookup` -> `Deletete`. 
The Drain API against  BPF_MAP_TYPE_HASH, BPF_MAP_TYPE_PERCPU_HASH, BPF_MAP_TYPE_LRU_HASH and BPF_MAP_TYPE_LRU_PERCPU_HASH is not expected to work with kernel versions < 5.14.

smagnani96:
* Moved functions so that GH code diffs are more readable.
* Added comments to newly helper functions.
* Removed TestDrainEmptyMap: added this check into `TestMapDrain` before filling the map.

map.go
```diff
@@ -1622,12 +1649,56 @@ func newMapIterator(target *Map) *MapIterator {
...
 }
 
+// cursorToKeyOut copies the current value held in the cursor to the
+// provided argument. In case of errors, returns false and sets a
+// non-nil error in the MapIterator.
+func (mi *MapIterator) cursorToKeyOut(keyOut interface{}) bool {
+	buf := mi.cursor.([]byte)
+	if ptr, ok := keyOut.(unsafe.Pointer); ok {
+		copy(unsafe.Slice((*byte)(ptr), len(buf)), buf)
+	} else {
```
dylandreimerink, line 1659: Sorry if this is a bit off topic. @ti-mo Do you recall why we have this exception here? This is just moving the existing code added in https://github.com/cilium/ebpf/commit/4609dc73fa846184b379ae1ee771cc9e95e2ecbc. But we seem to take so much care to ensure safety elsewhere, including in `sysenc.Unmarshal`. But not here.

If for some reason the user passes in an unsafe pointer in `keyOut` that is smaller than `buf` this causes us to write into unknown parts of the heap.

dylandreimerink (approved):
I am happy with this iteration :wink:.

map.go
```diff
@@ -1622,12 +1649,56 @@ func newMapIterator(target *Map) *MapIterator {
...
+		mi.err = fmt.Errorf("get next key: %w", mi.err)
+	}
+
+	return false
+}
+
+// drainMapEntry removes and returns the key held in the cursor
+// from the underlying map.
+func (mi *MapIterator) drainMapEntry(valueOut interface{}) bool {
```
ti-mo, line 1684: This is just a wrapper around LookupAndDelete:
```suggestion
func (mi *MapIterator) lookupAndDeleteInto(valueOut interface{}) bool {
```

map.go
```diff
@@ -1622,12 +1649,56 @@ func newMapIterator(target *Map) *MapIterator {
...
+	mi.err = mi.target.LookupAndDelete(mi.cursor, valueOut)
+	if mi.err == nil {
+		mi.count++
+		return true
+	}
+
+	if errors.Is(mi.err, ErrKeyNotExist) {
+		mi.err = nil
+	} else {
```
ti-mo, line 1693: Nitpicking, but IMO this is important none the less: try to avoid `else` near the end of a block, it's often unnecessary.

In this case, it's also a little misleading when mixed in with early returns. The `else` clause would be executed if `!errors.Is(..., ErrNotExist)`, but also when `err == nil`, and the latter case is already checked above.

It's usually clearer to be explicit and let things fall through naturally:
```go
	if errors.Is(mi.err, ErrKeyNotExist) {
		mi.err = nil
		return false
	}

	mi.err = fmt.Errorf("lookup and delete next key: %w", mi.err)
	return false
```

map.go
```diff
@@ -1622,12 +1649,56 @@ func newMapIterator(target *Map) *MapIterator {
...
+		mi.err = fmt.Errorf("lookup_and_delete key: %w", mi.err)
```
ti-mo, line 1694: Prefer natural language in error strings, these will be seen by the user:
```suggestion
		mi.err = fmt.Errorf("lookup and delete next key: %w", mi.err)
```

map.go
```diff
@@ -1636,6 +1707,38 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+	// is traversed, despite concurrent operations (ordering of items might differ).
+	for mi.err == nil && mi.fetchNextKey(nil) {
+		if mi.drainMapEntry(valueOut) {
+			return mi.cursorToKeyOut(keyOut)
+		}
+	}
+	return false
+}
+
+func (mi *MapIterator) nextIterate(keyOut, valueOut interface{}) bool {
```
ti-mo, line 1740:
```suggestion
func (mi *MapIterator) next(keyOut, valueOut interface{}) bool {
```

map_test.go
```diff
@@ -1174,6 +1174,139 @@ func TestMapIteratorAllocations(t *testing.T) {
...
+					err = m.Put(uint32(i), uint32(i))
+				}
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
+
+			iter := m.Drain()
+			allocs := testing.AllocsPerRun(int(m.MaxEntries()-1), func() {
+				if !iter.Next(keyPtr, &value) {
```
ti-mo, line 1300:
```suggestion
				qt.Assert(t, qt.IsTrue(iter.Next(keyPtr, &value)))
```

Or maybe qt.Assert allocates? Feel free to ignore.

map_test.go
```diff
@@ -1174,6 +1174,139 @@ func TestMapIteratorAllocations(t *testing.T) {
...
+			qt.Assert(t, qt.ErrorIs(err, nil))
+			defer m.Close()
+
+			for i := 0; i < int(m.MaxEntries()); i++ {
+				if keySize == 0 {
+					err = m.Put(nil, uint32(i))
+				} else {
+					err = m.Put(uint32(i), uint32(i))
+				}
+				if err != nil {
```
ti-mo, line 1293:
```suggestion
				qt.Assert(t, qt.IsNil(err))
```

map.go
```diff
@@ -1636,6 +1707,38 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+		return mi.nextDrain(keyOut, valueOut)
+	}
+	return mi.nextIterate(keyOut, valueOut)
+}
+
+func (mi *MapIterator) nextDrain(keyOut, valueOut interface{}) bool {
+	// Handle value-only maps (ex. queue).
+	if !isKeyValueMap(mi.target) {
+		if keyOut != nil {
+			mi.err = fmt.Errorf("non-nil keyOut provided for map without a key, must be nil instead")
```
ti-mo, line 1720:
```suggestion
			mi.err = fmt.Errorf("non-nil keyOut provided for map without a key")
```

map.go
```diff
@@ -1636,6 +1707,38 @@ func (mi *MapIterator) Next(keyOut, valueOut interface{}) bool {
...
+			return false
+		}
+		return mi.drainMapEntry(valueOut)
+	}
+
+	if mi.cursor == nil {
+		mi.cursor = make([]byte, mi.target.keySize)
+	}
+
+	// Always retrieve first key in the map. This should ensure that the whole map
```
ti-mo, line 1730: I don't think this is the case. Consider the following scenario:

- drainer A: NextKey(0) yields a
- drainer B: NextKey(0) yields a
- drainer B: LookupAndDelete(a) succeeds
- drainer A: LookupAndDelete(a) fails

The L&D failing can happen at any point due to a concurrent delete, also when e.g. a bpf program deletes a key, which is fairly common. In the current implementation, I think iteration halts at that point, while there may still be keys left in the map.

ti-mo (changes requested):
Still some changes to be made here, and I think some more testing needs to be done to understand the behaviour in the face of concurrent deletes.

It's also important to understand that LookupAndDelete may keep failing after a successful NextKey on very busy maps, and we may need to protect the application from getting into runaway CPU usage. I vaguely recall some existing code in the library that gives up after x attempts, but I can't remember where exactly.

Please reach out offline if you'd like some help with this.

We'll ship this in a 0.17.1 after the xmas break.

ti-mo:
:wave: @smagnani96 I've converted this to a draft for now, please address the feedback and ping when you want another review.

smagnani96:
> 👋 @smagnani96 I've converted this to a draft for now, please address the feedback and ping when you want another review.

Closing this, will reopen in case I'll find some slots to finish it 🙏🏼

Issue #1348, opened 2024-02-18
Can't load CO-RE eBPF code that accesses enum values


orishuss:
### Describe the bug

When using `bpf_core_enum_value_exists` on an enum and within it `bpf_core_enum_value` on the same enum value, when the value doesn't exist, the output go binary fails to load.
When loading the compiled object file with libbpf (bpftool) it loads successfully.

The code is in the attached file, but the problem can be seen in:
```c
if (bpf_core_enum_value_exists(enum cgroup_subsys_id, cpuset_cgrp_id_lublub)) {
    __attribute__((unused)) const volatile int val =
            bpf_core_enum_value(enum cgroup_subsys_id, cpuset_cgrp_id_lublub);
}
```
where the kernel enum `cgroup_subsys_id` is queried for type `cpuset_cgrp_id_lublub` (which doesn't really exist). The expected behaviour, which happens when loading the object with `bpftool`, is that the program passes the verifier's checks because the code inside the `if` shouldn't be evaluated.

The second value in the enum in the attached code, `cpuset_cgrp_id`, indeed exists, so using it instead of the fictitious value `cpuset_cgrp_id_lublub` works with both cilium and libbpf.

### How to reproduce

The code:
[enum-error.tar.gz](https://github.com/cilium/ebpf/files/14321981/enum-error.1.tar.gz)

build with `make` and run with `./main`.

Loading with bpftool (succeeds):
```shell
bpftool prog load bpf_bpf.o /sys/fs/bpf/bpf_bpf
rm /sys/fs/bpf/bpf_bpf # to remove
```

### Version information

- `github.com/cilium/ebpf v0.12.3`
- `go version go1.21.5 linux/amd64`
- `Linux 5.15.0`

dylandreimerink:
I was able to repo the issue and did some digging, found the following:

When I dump the program before CO-RE relocation logic the program looks like:
```
kprobe_tcp_close:
         ; void BPF_KPROBE(kprobe_tcp_close)
        0: LdImmDW dst: r1 imm: 1
         ; if (bpf_core_enum_value_exists(enum cgroup_subsys_id, cpuset_cgrp_id_lublub))
        2: JEqImm dst: r1 off: 3 imm: 0
        3: LdImmDW dst: r1 imm: 1
         ; __attribute__((unused)) const volatile int val =
        5: StXMemW dst: rfp src: r1 off: -4 imm: 0
         ; void BPF_KPROBE(kprobe_tcp_close)
[...]

PR #1351, opened 2024-02-19
btf: Take instruction size into account when handling poison relocation

dylandreimerink:
A CO-RE relocation might be poisoned, in such cases we write a "bogus" instruction to the relocation target. This can still result in working code if that instruction turns out to be in a dead code branch due to other CO-RE logic.

Currently this bogus instructions is always a function call to a non-existing function. This is a problem when the original instruction is a dword load immediate, which takes 2 instructions worth of space. When we replace this with a function call, we shrink the instruction stream which throws off offsets and instruction metadata.

So this commit makes makes the CO-RE logic check if we are dealing with a dword load immediate, and if so, we replace it with a dword imm load of 0xbad2310 to R10, which is illigal and will trip the verifier if evaluated.

Fixes: #1348


lmb:
Thanks for diagnosing this!

- Does libbpf do the same write into R10 trick, or something different?
- Can you add a test?

dylandreimerink:
> Does libbpf do the same write into R10 trick, or something different?

I came up with this myself, I didn't think to look at libbpf's approach. They seem to emit 2 call instructions in this case:
```
if (res->poison) {
poison:
	/* poison second part of ldimm64 to avoid confusing error from
	 * verifier about "unknown opcode 00"
	 */
	if (is_ldimm64_insn(insn))
		bpf_core_poison_insn(prog_name, relo_idx, insn_idx + 1, insn + 1);
	bpf_core_poison_insn(prog_name, relo_idx, insn_idx, insn);
	return 0;
}
```

We could just copy that.

> Can you add a test?

Yea, ofcourse.

> Does the jump instruction in the ELF carry a relocation to an LBB_ block?

It does, but I think the current fix is good enough?

```
bpf_bpf.o:      file format elf64-bpf

Disassembly of section kprobe/tcp_close:

0000000000000000 <kprobe_tcp_close>:
       0:       18 01 00 00 01 00 00 00 00 00 00 00 00 00 00 00 r1 = 0x1 ll
       2:       15 01 03 00 00 00 00 00 if r1 == 0x0 goto +0x3 <LBB0_2>
       3:       18 01 00 00 01 00 00 00 00 00 00 00 00 00 00 00 r1 = 0x1 ll
       5:       63 1a fc ff 00 00 00 00 *(u32 *)(r10 - 0x4) = r1

0000000000000030 <LBB0_2>:
       6:       b7 00 00 00 00 00 00 00 r0 = 0x0
       7:       95 00 00 00 00 00 00 00 exit
```

lmb:
> We could just copy that.

Yeah let's do that. The call instruction is already confusing enough.

lmb (approved):
Thanks!

lmb:
From discussion on Slack: going with the write to RFP trick, since that means we can avoid breaking API.

PR #1352, opened 2024-02-19
add example using Linux TC hook for network flow monitoring

s41m0n:
While an example program already exists for the incoming traffic hook with XDP, an example for a TC program was missing. 

This example shows how to load an eBPF program that monitors both incoming and outgoing TCP/UDP/IP flows identified with the 5-tuple session identifier (IP addresses, L4 ports, IP protocol).

The statistics are periodically displayed, and the content of the map is erased after a given number of iterations to allow potential new flows to be monitored.

The common header file is updated accordingly, to introduce a few shrunk data structures used during the packet processing logic (useful also for future examples).


lmb (commented):
Thanks for this nice example! I find that examples work best when they are quite narrow in scope: they show how to do one particular thing well. What would you say is that particular thing here?

I see a couple of candidates:
- Use of TCX
- Sharing code between XDP and TC
- Subtleties of LRU maps
- Batch operations
- Something else?

My personal preference is to choose one of them and strip the example down to the essentials. For example, rework this to showcase how to use the TCX link for flow monitoring on both ingress and egress. Drop the things that increase the lines of code needed beyond that: no XDP, no map clearing, no batch ops. WDYT?

ti-mo:
@s41m0n Thanks for the ping. I think Lorenz was suggesting to pick one of the items in the list to showcase in this PR. Given the focus of this example is tcx, I'd go for that one. All the other bullets can be examples of their own.

I'll convert to a draft for now, please mark as ready to review when done. Thanks!

s41m0n:
> @s41m0n Thanks for the ping. I think Lorenz was suggesting to pick one of the items in the list to showcase in this PR. Given the focus of this example is tcx, I'd go for that one. All the other bullets can be examples of their own.
> 
> I'll convert to a draft for now, please mark as ready to review when done. Thanks!

Hi @ti-mo and thanks for the feedback. If I'm not mistaken, the second commit should have removed everything except the "Use of TCx" from the bullet list. The example now is composed by:
1. Data plane: attached to TC Ingress and Egress to monitor connections (5-tuple session identifier) and count in/out packets
2. Control plane: periodically outputs the map's content

If you'd like to keep it even simpler, let me know :)

s41m0n:
> > @s41m0n Thanks for the ping. I think Lorenz was suggesting to pick one of the items in the list to showcase in this PR. Given the focus of this example is tcx, I'd go for that one. All the other bullets can be examples of their own.
> > I'll convert to a draft for now, please mark as ready to review when done. Thanks!
> 
> Hi @ti-mo and thanks for the feedback. If I'm not mistaken, the second commit should have removed everything except the "Use of TCx" from the bullet list. The example now is composed by:
> 
> 1. Data plane: attached to TC Ingress and Egress to monitor connections (5-tuple session identifier) and count in/out packets
> 2. Control plane: periodically outputs the map's content
> 
> If you'd like to keep it even simpler, let me know :)

@ti-mo @lmb With respect to the previous proposal, I think the example is now simpler, dealing with only Ingress/Egress flow monitoring through TC. What are your opinions about this? Thanks in advance for the feedback.

lmb:
Hi @s41m0n, sorry for the long delay. I've written up what kind of examples fit into the library here: https://ebpf-go.dev/contributing/new-example/ I think it makes sense to have an example for `tcx`. Would you be open to make the BPF side even simpler? I think just counting the number of ingress / egress packets would be sufficient.

s41m0n:
> Hi @s41m0n, sorry for the long delay. I've written up what kind of examples fit into the library here: https://ebpf-go.dev/contributing/new-example/ I think it makes sense to have an example for `tcx`. Would you be open to make the BPF side even simpler? I think just counting the number of ingress / egress packets would be sufficient.

@lmb That's amazing, thanks for the guide! I just simplified the example to only count ingress and egress packets, as you suggested. Let me know :)

examples/tcx/tcx.c
```diff
@@ -0,0 +1,47 @@
+//go:build ignore
+
+#include "common.h"
+
+char __license[] SEC("license") = "Dual MIT/GPL";
+
+/* Define an ARRAY map for storing ingress packet count */
+struct {
```
ti-mo, line 8: It's a shame we haven't worked on an API to interact with static variables defined in an ELF. That would've made this example much more elegant.

lmb (approved):

examples/tcx/tcx.c
s41m0n: Reaching you out soon on Slack to discuss this :)

PR #1357, opened 2024-02-22
btf: fix race due to concurrent read access

lmb:
Up until the introduction of lazy copying, reading from a Spec concurrently was safe. Now a read may trigger a copy and a write into the Spec, therefore causing a race on mutableTypes.

Fix this by introducing a mutex which protects access to the mutable state. We need to be a bit careful here: copying in mutableTypes.add happens piecemeal, so we need to take a lock for the whole duration of modifyGraph.

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -66,8 +66,9 @@ func (s *immutableTypes) typeByID(id TypeID) (Type, bool) {
 // mutableTypes is a set of types which may be changed.
 type mutableTypes struct {
 	imm           immutableTypes
+	mu            *sync.RWMutex   // protects copies below
 	copies        map[Type]Type   // map[orig]copy
-	copiedTypeIDs map[Type]TypeID //map[copy]origID
+	copiedTypeIDs map[Type]TypeID // map[copy]origID
 }
 
 // add a type to the set of mutable types.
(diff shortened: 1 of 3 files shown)


lmb:
@eiffel-fl the locking ended up being more subtle than what you had in your PR. Can you give this a whirl?

btf/btf.go
```diff
@@ -66,15 +66,30 @@ func (s *immutableTypes) typeByID(id TypeID) (Type, bool) {
...
 // Copies type and all of its children once. Repeated calls with the same type
 // do not copy again.
 func (mt *mutableTypes) add(typ Type, typeIDs map[Type]TypeID) Type {
+	mt.mu.RLock()
+	cpy, ok := mt.copies[typ]
+	mt.mu.RUnlock()
+
+	if ok {
+		// Fast path: the type has been copied before.
+		return cpy
```
ti-mo, line 85: How come we didn't have this before? :thinking:

ti-mo (approved):
Thanks for picking this up! We should release 0.13.1 to address this.

btf/btf.go
lmb: Because the same thing happens when entering the closure passed to modifyGraphPreorder.

PR #1359, opened 2024-02-23
link, internal: generate full linkinfo for union subtypes

rgo3:
The current generic abstraction of `bpf_link_info` makes it harder to implement features like #1295. Instead of splitting LinkInfo into a generic part and having extra structs to represent the fields of union members in `bpf_link_info`, with this PR we generate the full LinkInfo struct for each member the union in `bpf_link_info`.

### DIFF
--- a/link/link.go
+++ b/link/link.go
@@ -1,8 +1,6 @@
 package link
 
 import (
-	"bytes"
-	"encoding/binary"
 	"fmt"
 
 	"github.com/cilium/ebpf"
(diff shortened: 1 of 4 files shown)


link/link.go
```diff
@@ -355,32 +353,68 @@ func (l *RawLink) Info() (*Info, error) {
 	var extra interface{}
```
olsajiri, line 353: hum, so the idea is to run the link info syscall twice, first to get the link info type and second to read it into specific type?

could we just store the link type, like to have Link interface function that returns that? seems wasteful to do 2 same syscalls just to get it into convenient type

olsajiri: @rgo3 @lmb I updated the missed counts interface changes https://github.com/cilium/ebpf/pull/1295 on top of this change, still need to update the tetragon code on top of that, but so far looks good

lmb: We can move the `info.Type` specific code from this function to new `Info()` methods on the concrete link types, like you did with `KprobeMulti.Info()` before. Then the only caller of `RawLink.Info()` is `wrapRawLink` which only really cares about `info.Type` anyways. This way `Info()` only does a single call in the common case (caveat: sometimes we need two calls anyways to retrieve buffer sizes, etc.).

We can make that change in this PR, or as a follow up, or once you've landed your kprobe missed count changes. Up to you and @rgo3.

rgo3: How would having separate `Info()` methods on all concrete link types as well look in practice when loading a pinned link from bpffs and get specific information from it?

1. Load pinned link
2. Call generic Info method
3. Check type of returned object that satisfies the `Link` interface
4. Assert concrete link type
5. Call type dependent Info method

lmb: That sounds right to me. 1-4 happen behind the scenes in LoadPinnedLink. 5 happens when a user does `KprobeMulti.Info()`.

olsajiri (approved):

lmb:
@olsajiri I think @rgo3 is head deep in IPSec work, when do you need this?

rgo3:
I've talked to @olsajiri yesterday that I would take a look today. I think the fastest solution would be merging this PR, jiri can base his work on top and then I look into refactoring after? That we we enable Jiri to continue and I get a little more time :grin:
If you agree I can rebase this PR and get it ready.

lmb:
SGTM

mmat11 (approved):

rgo3:
Rebased on top of the netkit changes and did the same refactoring needed for that link type, otherwise no changes.

lmb:
Thanks :pray:

Issue #1355, opened 2024-02-20
loader: handle missing kfunc gracefully


dylandreimerink:
While looking at `__weak` support I came across the following pattern used in selftests:
```
void invalid_kfunc(void) __ksym __weak;

#define bpf_ksym_exists(sym) \
	({ \
		_Static_assert(!__builtin_constant_p(!!sym), #sym " should be marked as __weak"); \
		!!sym; \
	})

SEC("xdp") int xdp_prog(struct xdp_md *ctx) {
	if (bpf_ksym_exists(invalid_kfunc))
		invalid_kfunc();

	return 0;
}
```

So, essentially, this allows programs to handle cases where a given kfunc isn't available such as on older kernels or on kernels missing certain KCONFIG.

The `__weak` attribute is required for this to work so initially this results in a `relocating instruction: asm relocation: invalid_kfunc: unsupported binding: STB_WEAK` error. But assuming this is resolved by #466. We get the error: `fixing up kfuncs: kfunc "invalid_kfunc": not supported`.

The instructions for the example program are:
```
xdp_prog:
         ; if (bpf_ksym_exists(invalid_kfunc))
        0: LdImmDW dst: r1 imm: 0 <invalid_kfunc>
        2: JEqImm dst: r1 off: 1 imm: 0
         ; invalid_kfunc();
        3: Call Kfunc(-1) <invalid_kfunc>
         ; return 0;
        4: MovImm dst: r0 imm: 0
        5: Exit
```

The way libbpf handles this is by simply doing nothing. Since instruction 0 has a imm of 0 by default the code branch with the call is dead code and never trips the verifier.

There is a secondary issue, that being that we only relocate `Call` instructions, even if the kfunc were valid, instruction 0 would be ignored. So if we can fix both, we effectively allow CO-RE with kfuncs.

PR #1364, opened 2024-03-02
Add support for weak kfuncs

dylandreimerink:
This PR allows users to mark their kfunc defintions as `__weak`. Weak kfuncs do not cause an error during loading if the kfunc in question can't be found. Instead, a poison value is written which will cause the verifier to bail out if the instruction is evaluated.

In addition, this PR relocates LDIMM64 instructions with kfunc relocation entries. When relocated, the BTF id of the kfunc is written to the instruction's immediate field.

Together these two changes allow users to write eBPF programs which can gracefully handle the absence of a kfunc. To do so the `bpf_ksym_exists` macro from `bpf_helpers.h` can be used to check if a kfunc is present.

```
void invalid_kfunc(void) __ksym __weak;

__section("tp_btf/task_newtask") int weak_kfunc_missing(void *ctx) {
	if (bpf_ksym_exists(invalid_kfunc)) {
		invalid_kfunc();
		return 0;
	}

	return 1;
}
```

In this example the kfunc is always invalid, yet the program can still load since the branch calling the kfunc is unreachable.

So this effectively provides CO-RE capabilities for kfuncs. Especially important since kfuncs are less stable than helpers, their availability depending on kernel version, kernel compilation flags, or kernel modules being loaded or not.

Fixes: #1355


testdata/kfunc.c
```diff
@@ -29,3 +30,41 @@ __section("fentry/bpf_fentry_test2") int benchmark() {
 	// the kernel when benchmarking the loader.
 	return bpf_fentry_test1(0);
 }
+
+#define __ksym __attribute__((section(".ksyms")))
```
lmb, line 34: Nit: put these in `common.h`?

linker.go
```diff
@@ -259,6 +260,8 @@ func fixupAndValidate(insns asm.Instructions) error {
 	return nil
 }
 
+const kfuncCallPoisonBase = 2002000000
```
lmb, line 263: Can you add a link with the source of the magic number? Guessing that this is from libbpf.

linker.go
```diff
@@ -292,24 +295,45 @@ fixups:
...
+		metadata := ins.Metadata.Get(kfuncMetaKey{})
+		if metadata == nil {
 			if !iter.Next() {
 				// break loop if this was the last instruction in the stream.
 				break
 			}
 			continue
 		}
 
+		if !ins.IsKfuncCall() && !ins.OpCode.IsDWordLoad() {
```
lmb, line 307: Nit: can you fold this into the `if ins.IsKfuncCall()` check + fixup below? Seems more natural there and we don't duplicate the conditions.

elf_reader.go
```diff
@@ -610,12 +615,22 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
+				Func:    kf,
+				Binding: bind,
+			})
+
+			if ins.OpCode.JumpOp() == asm.Call {
+				ins.Src = asm.PseudoKfuncCall
+				ins.Constant = -1
+			}
+
+			if ins.OpCode.IsDWordLoad() {
```
lmb, line 631: Nit: `else if`?

lmb (changes requested):
Nice work! Just some nits.

lmb (approved):

PR #1375, opened 2024-03-13
ringbuf: fix rare event corruption by reading event header atomically

paulcacheux:
I have been investigating some very rare event corruption in the output sample data coming from the eBPF ringbuffer. Looking at [libbpf](https://github.com/libbpf/libbpf/blob/2778cbce609aa1e2747a69349f7f46a2f94f0522/src/ringbuf.c#L249), and [aya (rust equivalent)](https://github.com/aya-rs/aya/blob/c302f8370d34568df513a80036c3c5c44d41dad4/aya/src/maps/ring_buf.rs#L390) it seems we should be reading the len field in the header atomically.

I have been able to confirm that this fixes my issue.


internal/unix/types_other.go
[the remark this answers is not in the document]
paulcacheux: you're right, no need for that. reverted

ringbuf/reader.go
```diff
@@ -22,12 +21,9 @@ var (
 	errBusy    = errors.New("sample not committed yet")
 )
 
-var ringbufHeaderSize = binary.Size(ringbufHeader{})
-
 // ringbufHeader from 'struct bpf_ringbuf_hdr' in kernel/bpf/ringbuf.c
 type ringbufHeader struct {
-	Len   uint32
-	PgOff uint32
```
lmb, line 30: Nit: do `_ uint32` instead to make it clear that the struct in the kernel is larger? Otherwise the comment is incorrect. Also doesn't match `BPF_RINGBUF_HDR_SZ`.

ringbuf/ring.go
```diff
@@ -109,6 +109,19 @@ func (rr *ringReader) remaining() int {
...
 
+func (rr *ringReader) readHeader() (ringbufHeader, error) {
+	prod := atomic.LoadUint64(rr.prod_pos)
+
+	// ensure we do not go over the prod position
+	if prod-rr.cons < unix.BPF_RINGBUF_HDR_SZ {
+		return ringbufHeader{}, io.EOF
+	}
+
+	len := atomic.LoadUint32((*uint32)((unsafe.Pointer)(&rr.ring[rr.cons&rr.mask])))
```
lmb, line 120: Can you add a comment to explain why we need to read the header atomically?

lmb (changes requested):
Thanks, very interesting! Do you understand how user space can end up with the busy bit missing? From my perspective:

1. BPF reserves space in the ringbuf. This [sets the BPF_RINGBUF_BUSY_BIT](https://github.com/torvalds/linux/blob/eb26cbb1a754ccde5d4d74527dad5ba051808fad/kernel/bpf/ringbuf.c#L444). This set is not atomic!
2. The kernel does an an [atomic store to prod_pos](https://github.com/torvalds/linux/blob/eb26cbb1a754ccde5d4d74527dad5ba051808fad/kernel/bpf/ringbuf.c#L448).
3. User space does an [atomic load of prod_pos](https://github.com/cilium/ebpf/blob/e1dc68abf05a02803a12b36810528c6dfa67f679/ringbuf/ring.go#L113).
4. User space copies len (non atomically). BPF_RINGBUF_BUSY_BIT is not present.

The only way I can see this happening is if:

- The assignment of `len = size | BUSY_BIT` is decomposed into `len = size; len |= BUSY` in assembly.
- The store to prod_pos is reordered to happen after `len = size` and before `len |= BUSY`.

The problem with this theory is that I don't understand why reading len atomically would fix this problem, since there is no atomic write to len on the kernel side? (At least until we hit the `xchg` in `bpf_ringbuf_commit`)

lmb:
Follow up question: does happen on arm64 and amd64?

lmb:
Here is `len = size | BUSY_BIT` on x86-64: https://gist.github.com/lmb/df04ba01510f344f4e18bbe90aca7fe4#file-disas-asm-L86-L93. Doesn't seem like the store could be reordered so that the busy bit isn't present.

ringbuf/ring.go
paulcacheux: Done

ringbuf/reader.go
paulcacheux: Done

paulcacheux:
I just checked and I only see this corruption on arm64 hosts

lmb:
Fixed up the second commit to handle `io.EOF` correct and added some context to the commit message. I ended up "validating" this fix by writing a litmus test against the Linux Kernel Memory Model.

PR #1379, opened 2024-03-14
ringbuf: simplify ringReader

lmb:
The code to correctly access the ring is currently split across two files an multiple functions which makes it hard to grasp what is going on. Strip out the overly generic Reader interface and only expose ringReader.readRecord. Also move discarding of samples into readRecord.

This removes the amount of code and cuts down on the number of places where we call into sync/atomic.


paulcacheux (approved):
Looks much better, thanks for cleaning this up

mythi:
:+1:

ti-mo:
@lmb Thanks!

PR #1381, opened 2024-03-19
internal: add support for loong64

zhangwenlong8911:


internal/endian_be.go
```diff
@@ -1,4 +1,4 @@
-//go:build armbe || arm64be || mips || mips64 || mips64p32 || ppc64 || s390 || s390x || sparc || sparc64
```
paulcacheux, line 1: Looking at https://cs.opensource.google/go/x/sys/+/refs/tags/v0.18.0:cpu/byteorder.go;l=44, it seems loong64 is actually a little endian arch

lmb:
@zhangwenlong8911 I'm closing this for now. Feel free to open a new PR with the correct changes.

PR #1384, opened 2024-03-20
allow using binary.NativeEndian

lmb:
use binary.NativeEndian

    Remove the internal implementation of NativeEndian in favour of 
    binary.NativeEndian. There is one hiccup: it's not possible to use a simple
    comparison to figure out whether NativeEndian is the same as for example
    LittleEndian. The IsNativeEndian helper takes care of this.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -935,14 +935,19 @@ func (iter *InstructionIterator) Next() bool {
 type bpfRegisters uint8
 
 func newBPFRegisters(dst, src Register, bo binary.ByteOrder) (bpfRegisters, error) {
+	buf := make([]byte, 2)
+	val := uint16(dst&0x0F)<<8 | uint16(src&0x0F)
 	switch bo {
 	case binary.LittleEndian:
-		return bpfRegisters((src << 4) | (dst & 0xF)), nil
+		binary.LittleEndian.PutUint16(buf, val)
 	case binary.BigEndian:
-		return bpfRegisters((dst << 4) | (src & 0xF)), nil
+		binary.BigEndian.PutUint16(buf, val)
+	case binary.NativeEndian:
+		binary.NativeEndian.PutUint16(buf, val)
 	default:
 		return 0, fmt.Errorf("unrecognized ByteOrder %T", bo)
 	}
+	return bpfRegisters(buf[0]<<4 | buf[1]), nil
 }
 
 // IsUnreferencedSymbol returns true if err was caused by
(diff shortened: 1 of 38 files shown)


lmb:
Best reviewed with white space hidden.

lmb:
Discussed with Timo: seems like a lot of pain for little gain. We'll discuss next week.

asm/instruction.go
```diff
@@ -935,14 +935,19 @@ func (iter *InstructionIterator) Next() bool {
 type bpfRegisters uint8
 
 func newBPFRegisters(dst, src Register, bo binary.ByteOrder) (bpfRegisters, error) {
+	buf := make([]byte, 2)
+	val := uint16(dst&0x0F)<<8 | uint16(src&0x0F)
 	switch bo {
 	case binary.LittleEndian:
```
paulcacheux, line 941: why not replacing this whole switch with `bo.PutUint16` directly ?

ti-mo: That's the annoying bit: calling an interface method cannot be inlined, plus it allocates. This function is called a ton, so it starts adding up.

We have some amount of discovery to do, because now binary.NativeEndian is a thing, we can probably eliminate these kind of switches/branches and replace them with `binary.NativeEndian.PutUint16`, since marshaling bpf insns is usually done before loading into the host kernel, and we'd almost always do that in native endianness. 

`Instruction.Marshal()` is exported API, though, so we should have a look at public callers on Sourcegraph to see if anyone depends on marshaling asm.Instructions in arbitrary endiannesses (e.g. for converting ELFs from LE to BE).

paulcacheux: ah yeah, that's too bad

lmb: > `Instruction.Marshal()` is exported API, though, so we should have a look at public callers on Sourcegraph to see if anyone depends on marshaling asm.Instructions in arbitrary endiannesses

I'd rather not get into the habit of hardcoding NativeEndian, it leads to not thinking about endianness at callsites and that tends to cause subtle bugs. Compared to that the type switch isn't bad at all imo.

ti-mo:
Coming back to this, I find our `internal.NativeEndian` implementation more elegant than what they did with `binary.NativeEndian`. Shall we keep things as-is for now?

lmb:
yeah makes sense. i think the real solution is to stop using `binary.ByteOrder` to represent endianness in `Collection`, etc. We should have made up our own constants.

PR #1386, opened 2024-03-21
miscellaneous test fixes

lmb:
cmd/bpf2go: don't require internet connection in tests

    Disable the checksum db and the module proxy. They aren't needed for tests
    and disabling means we don't need internet access.

link: disable uprobe tests on v4.9

    Running v4.9 in vimto, trying to create a perf event via perf_event_open 
    returns EIO. I've tracked this down to a check in uprobe_register:

        if (!inode->i_mapping->a_ops->readpage &&
    !shmem_mapping(inode->i_mapping))
           return -EIO;

    See
    https://elixir.bootlin.com/linux/v4.9.337/source/kernel/events/uprobes.c#L880

    My hunch is that this is related to how vimto sets up the overlayfs mount:
    if the probed binary is below the overlayfs the setup fails.

    Instead of fixing vimto, stop testing these code paths on v4.9. That kernel 
    is already EOL so there is not much use in investing more time.

### DIFF
--- a/cmd/bpf2go/main_test.go
+++ b/cmd/bpf2go/main_test.go
@@ -90,6 +90,8 @@ func main() {
 			goBuild.Env = append(os.Environ(),
 				"GOOS=linux",
 				"GOARCH="+arch,
+				"GOPROXY=off",
+				"GOSUMDB=off",
 			)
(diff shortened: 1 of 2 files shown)


lmb:
Broken out from #1321

PR #1392, opened 2024-03-23
link: add Iterator

mpastyl:
Add an iterator for links that loops over all links in the system.

This then can be used to get information about attached programs, e.g. to [find which XDP program attached to an interface](https://github.com/cilium/ebpf/discussions/1380).

The implementations closely mirrors that of the  [HandleIterator](https://github.com/cilium/ebpf/blob/65fb774468cbc51047f7115a57623832e8388eff/btf/handle.go#L201)

### DIFF
--- a/link/link.go
+++ b/link/link.go
@@ -1,7 +1,9 @@
 package link
 
 import (
+	"errors"
 	"fmt"
+	"os"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/btf"
@@ -46,14 +48,34 @@ type Link interface {
 
 // NewLinkFromFD creates a link from a raw fd.
 //
-// You should not use fd after calling this function.
+// Deprecated: use [NewFromFD] instead.
 func NewLinkFromFD(fd int) (Link, error) {
+	return NewFromFD(fd)
+}
+
+// NewFromFD creates a link from a raw fd.
+//
+// You should not use fd after calling this function.
+func NewFromFD(fd int) (Link, error) {
 	sysFD, err := sys.NewFD(fd)
 	if err != nil {
 		return nil, err
 	}
 
 	return wrapRawLink(&RawLink{fd: sysFD})
+}
+
+// NewFromID returns the link associated with the given id.
+//
+// Returns ErrNotExist if there is no link with the given id.
+func NewFromID(id ID) (Link, error) {
+	getFdAttr := &sys.LinkGetFdByIdAttr{Id: id}
+	fd, err := sys.LinkGetFdById(getFdAttr)
+	if err != nil {
+		return nil, fmt.Errorf("get link fd from ID %d: %w", id, err)
+	}
+
+	return wrapRawLink(&RawLink{fd, ""})
 }
 
 // LoadPinnedLink loads a link that was persisted into a bpffs.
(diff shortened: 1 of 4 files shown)


link/link.go
```diff
@@ -446,3 +461,74 @@ func (l *RawLink) Info() (*Info, error) {
...
+		l, err := NewLinkFromID(id)
+		if errors.Is(err, os.ErrNotExist) {
+			// Couldn't load the link fast enough. Try next ID.
+			continue
+		} else if err != nil {
+			it.err = fmt.Errorf("get link for ID %d: %w", id, err)
+			break
+		}
+
+		if it.Link != nil {
```
mmat11, line 501: the iterator code lgtm, but I don't understand why are links being closed? wouldn't this break existing links?

mpastyl: Good point. The link lifecycle is not super clear to me. If you have resources on this they are highly appreciated.
From what I can gather from [here](https://kernel.org/doc/html/v6.9-rc1/userspace-api/ebpf/syscall.html) and [here](https://ebpf-go.dev/concepts/object-lifecycle/). 

> BPF_LINK_GET_FD_BY_ID
...
Return
A new file descriptor (a nonnegative integer), or -1 if an error occurred (in which case, errno is set appropriately).

and at the end (although this refers to maps and programs, not specifically to links)
> An eBPF object is deallocated only after all file descriptors referring to the object have been closed and no references remain pinned to the filesystem or attached (for example, bound to a program or device).

So my rough understanding is that only the new file descriptor is closed on `Close()`. If there are still references from other sources the link will remain active.

I wonder if there is an easy way to validate it in unit tests. 
Calling `testLink(t, &linkCgroup{*tLink}, prog)` on the original (Raw) link, after the iterator is done seems to work.
Not sure if this is enough though...

mmat11: > An eBPF object is deallocated only after all file descriptors referring to the object have been closed and no references remain pinned to the filesystem or attached

ok, I see 👍🏻

mmat11 (approved):
lgtm!

link/link.go
```diff
@@ -56,6 +58,19 @@ func NewLinkFromFD(fd int) (Link, error) {
 	return wrapRawLink(&RawLink{fd: sysFD})
 }
 
+// NewLinkFromID returns the link associated with the given id.
+//
+// Returns ErrNotExist if there is no link with the given id.
+func NewLinkFromID(id ID) (Link, error) {
```
lmb, line 64: Nit: this reads as `link.NewLinkFromID`, maybe just `link.NewFromID` is nicer?

link/link_test.go
```diff
@@ -88,6 +88,67 @@ func TestRawLinkLoadPinnedWithOptions(t *testing.T) {
...
+	}
+	if it.Link != nil {
+		t.Fatal("Next doesn't clean up link on last iteration")
+	}
+	if prev != it.ID {
+		t.Fatal("Next changes ID on last iteration")
+	}
+	if foundLink == nil {
+		t.Fatal("Original link not found")
+	}
```
lmb, line 139: Nit: can you `defer foundLink.Close()` here so that we don't forget to clean up on the error paths below?

lmb (changes requested):
Looks great, just two small nits. Thanks!

link/link.go
mmat11: I think this was named after `link.NewLinkFromFD` for consistency; should we change both?

mpastyl: Sounds good, I changed both.

lmb: Made it so that we keep the old function around but `Deprecate` it. No need to break users in this simple case.

lmb (approved):
Thanks!

PR #1399, opened 2024-03-26
Enhance kernel version detection by using uname

Vyom-Yadav:
* Refactored the detectKernelVersion function to not solely rely on vdsoVersion. Added a fallback mechanism using the KernelRelease function, which utilizes the uname system call to get the kernel release string.

---

I had `vDSO` disabled for tapping into some syscalls, so this is something good to have.

### DIFF
--- a/internal/version.go
+++ b/internal/version.go
@@ -2,13 +2,15 @@ package internal
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/cilium/ebpf/internal/unix"
 )
 
 const (
-	// Version constant used in ELF binaries indicating that the loader needs to
+	// MagicKernelVersion constant used in ELF binaries indicating that the loader needs to
 	// substitute the eBPF program's version with the value of the kernel's
 	// KERNEL_VERSION compile-time macro. Used for compatibility with BCC, gobpf
 	// and RedSift.
@@ -18,7 +20,7 @@ const (
 // A Version in the form Major.Minor.Patch.
 type Version [3]uint16
 
-// NewVersion creates a version from a string like "Major.Minor.Patch".
+// NewVersion creates a version from a string like "KernelVersion.Major.Minor".
 //
 // Patch is optional.
 func NewVersion(ver string) (Version, error) {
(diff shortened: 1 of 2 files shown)


brycekahle:
We moved away from `uname` because it is notoriously hard to parse, and often doesn't contain enough information to match the value the kernel expects (for kernels < 5.0).

You can always set `ProgramSpec.KernelVersion` if you don't want the library to automatically set the value.

Vyom-Yadav:
> We moved away from uname because it is notoriously hard to parse, and often doesn't contain enough information to match the value the kernel expects (for kernels < 5.0).

@brycekahle I added this as a fallback mechanism. Setting the flag manually v/s at least having something to get the kernel version 🤷🏼‍♂️

brycekahle:
I understand, but this is more code to maintain for what sounds like a very narrow use case (vDSO disabled). See the code we removed in #500 for the old parsing code.

Maybe the maintainers feel differently, but my vote is to not add more maintenance burden.

Vyom-Yadav:
I can see why the team took that decision. I was using Tetragon, and it isn't working with vDSO disabled, so I was trying to work around that somehow. I don't program using eBPF so I don't know how to set this up using some external Tetragon config, if possible.

Vyom-Yadav:
Indeed, it's not pretty, but that is just a backup option in case vdso is disabled.

It is not reducing any functionality, just enhancing it. It does not break anything. 

I have vDSO disabled to tap into time syscalls. This would help in debugging unreproducible builds, so that's the reason behind it.

lmb:
Discussed this with @ti-mo, we're not going to bring back uname parsing. That said, I think we should allow running without the vDSO.

In fact, the current version of the library shouldn't really rely on the vDSO any more since https://github.com/cilium/ebpf/commit/fbff7dbd617fccbf0bea0ea8c383f162088c9df5 Can you check which version of the library you are using at the moment? My guess is that you don't have the commit above. If that isn't true, please open a bug report that includes the error you are seeing and which version of the library you are using. We might be able to work around this in some other fashion.

Vyom-Yadav:
@lmb Created https://github.com/cilium/ebpf/issues/1408

PR #1404, opened 2024-03-28
add WakeupEvents support to perf Reader

brycekahle:
follow up to #524 with only the `WakeupEvents` portion.

### DIFF
--- a/perf/ring.go
+++ b/perf/ring.go
@@ -22,12 +22,12 @@ type perfEventRing struct {
 	ringReader
 }
 
-func newPerfEventRing(cpu, perCPUBuffer, watermark int, overwritable bool) (*perfEventRing, error) {
-	if watermark >= perCPUBuffer {
+func newPerfEventRing(cpu, perCPUBuffer int, opts ReaderOptions) (*perfEventRing, error) {
+	if opts.Watermark >= perCPUBuffer {
 		return nil, errors.New("watermark must be smaller than perCPUBuffer")
 	}
 
-	fd, err := createPerfEvent(cpu, watermark, overwritable)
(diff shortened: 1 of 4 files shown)


perf/ring.go
```diff
@@ -98,13 +98,20 @@ func (ring *perfEventRing) Close() {
 	ring.mmap = nil
 }
 
-func createPerfEvent(cpu, watermark int, overwritable bool) (int, error) {
-	if watermark == 0 {
-		watermark = 1
+func createPerfEvent(cpu int, opts ReaderOptions) (int, error) {
+	wakeup := 0
+	bits := 0
```
florianl, line 103: As the documentation for `WakeupEvents` states, it should be mutually exclusive with `Watermark`.
Can we add a check for the case where both, `WakeupEvents` and `Watermark`, are set  and return something like `fmt.Errorf("WakeupEvents and Watermark are mutually exclusive: %w", unix.EINVAL)`.

brycekahle: Like this one? https://github.com/cilium/ebpf/pull/1404/files#diff-7b72e423487d9bde0d352acf2654d5e70f0c10898776b867fc3bac2b38f9ceafR199-R201

florianl: ah - good point! placing the check in `NewReaderWithOptions()` before doing allocations make sense!

florianl (approved):

dylandreimerink:
Hi! It seems this PR introduced a flaky test, see https://github.com/cilium/ebpf/issues/1419. @brycekahle could you take a look? So far I have not been able to find the root cause, will update the issue if I do.

PR #1412, opened 2024-04-02
gh: try to group dependabot updates

lmb:
Make dependabot less spammy.


lmb:
Bleh, no way to easily validate dependabot config. @ti-mo please take a look once you're back.

Issue #1413, opened 2024-04-02
Allow changing line info data in btf.Line


MarcusWichelmann:
### Describe the bug

I'd like to change the `line` field in the `btf.Line` structure for a single instruction of a program, but cannot find a way to do that.

#### Background

I'm templating an eBPF program by replacing a placeholder function in its instructions with some custom instructions. Then I use `WithMetadata()` (see https://github.com/cilium/ebpf/pull/832) to copy over the metadata of the first placeholder instruction (that has the symbol) to the new one. 

This works fine, but I'd like to also change the BTF line info for the first instruction to contain some information about the inserted logic so it can be recognized in an instruction dump / verifier log.

I could use `.WithSource(asm.Comment("ABC"))` for that, but this causes a `missing bpf_line_info for 1 funcs starting from func#1` verifier error, because the line info seems to go missing then.

What I probably need to do is, to retrieve the existing `bpf.Line` structure using `.Source().(*btf.Line)` and change its fields, but these are unexported, so there doesn't seem to be a way to do that.

I'd like to discuss what would be the preferred solution here, before I make a pull request.

### How to reproduce

Load an eBPF program and call this on the first instruction of a function (the one with the symbol):

```
insns[i] = insns[i].WithMetadata(insn.Metadata).WithSource(asm.Comment("ABC"))
```

Then try to load the program and check the verifier error.

### Version information

github.com/cilium/ebpf v0.14.0

lmb:
I'm using the same functionality in the library for https://github.com/cilium/ebpf/pull/1402/commits I think we could just export the fields in `Line` as you suggested. cc @dylandreimerink

dylandreimerink:
Yes. I think that would be alright. There isn't any guarding of values needed, so just exposing fields should be safe. Are you thinking of just exposing `Line` or the rest of the fields as well (seem odd otherwise).

MarcusWichelmann:
Wow, you're fast.

@dylandreimerink Yeah would probably make sense to export all of them then.

Should I make a PR?

MarcusWichelmann:
What would be your preferred solution? Are there other ways how invalid values can get into that structure which would make a validation before marshalling necessary instead of just checking in the setters? Or where would you put that validation?

MarcusWichelmann:
Also there seems to be some validation already:

https://github.com/cilium/ebpf/blob/main/btf/ext_info.go#L594

Isn't that sufficient?

lmb:
[...]

PR #1417, opened 2024-04-03
btf: synthesise instruction comments into line info 

MarcusWichelmann:
When an asm.Comment is added to an instruction, existing source information like btf.Line will be replaced. But in some cases, instructions without line information won't pass the verifier. The expected behavior would be, that the comment gets synthesized into a line info where everything but the line string is zero. This PR implements this.

Fixes https://github.com/cilium/ebpf/issues/1413

### DIFF
--- a/btf/ext_info.go
+++ b/btf/ext_info.go
@@ -142,12 +142,13 @@ func AssignMetadataToInstructions(
 
 // MarshalExtInfos encodes function and line info embedded in insns into kernel
 // wire format.
+//
+// If an instruction has an [asm.Comment], it will be synthesized into a mostly
+// empty line info.
 func MarshalExtInfos(insns asm.Instructions, b *Builder) (funcInfos, lineInfos []byte, _ error) {
 	iter := insns.Iterate()
 	for iter.Next() {
-		_, ok := iter.Ins.Source().(*Line)
-		fn := FuncMetadata(iter.Ins)
-		if ok || fn != nil {
+		if iter.Ins.Source() != nil || FuncMetadata(iter.Ins) != nil {
 			goto marshal
 		}
 	}
@@ -167,7 +168,16 @@ marshal:
 			}
 		}
 
-		if line, ok := iter.Ins.Source().(*Line); ok {
+		if source := iter.Ins.Source(); source != nil {
+			var line *Line
+			if l, ok := source.(*Line); ok {
+				line = l
+			} else {
+				line = &Line{
+					line: source.String(),
+				}
+			}
+
 			li := &lineInfo{
 				line:   line,
 				offset: iter.Offset,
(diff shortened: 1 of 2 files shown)


btf/ext_info.go
```diff
@@ -142,12 +142,15 @@ func AssignMetadataToInstructions(
...
+//
+// If an instruction has an [asm.Comment], it will be synthesized into a mostly
+// empty line info.
 func MarshalExtInfos(insns asm.Instructions, b *Builder) (funcInfos, lineInfos []byte, _ error) {
 	iter := insns.Iterate()
 	for iter.Next() {
-		_, ok := iter.Ins.Source().(*Line)
+		source := iter.Ins.Source()
 		fn := FuncMetadata(iter.Ins)
-		if ok || fn != nil {
```
MarcusWichelmann, line 150: I could simplify the check here, because `.Source()` returns a `fmt.Stringer` and the code below now works with any non-nil `fmt.Stringer`, not just `*Line`s.

btf/ext_info.go
```diff
@@ -142,12 +142,15 @@ func AssignMetadataToInstructions(
 
 // MarshalExtInfos encodes function and line info embedded in insns into kernel
 // wire format.
+//
+// If an instruction has an [asm.Comment], it will be synthesized into a mostly
+// empty line info.
 func MarshalExtInfos(insns asm.Instructions, b *Builder) (funcInfos, lineInfos []byte, _ error) {
```
MarcusWichelmann, line 148: There doesn't seem to be a test for this function that I could extend and it also doesn't seem trivial to add one, since I'd have to fake the `Builder` argument. But at least this gets partially tested by the `TestCOREPoisonLineInfo` test.

dylandreimerink: Yea, we don't have a specific test. But pretty much every other test that loads a program with BTF into the kernel invokes this, so we have a lot of indirect coverage. If we break anything in here it will be caught.

dylandreimerink (changes requested):

MarcusWichelmann:
I'm not sure why the `arm64` CI job fails and if this could be related to this PR. 

>  The job running on runner 9cc722ffc9bb5c1dff1132904f63cda3c36d7769 has exceeded the maximum execution time of 10 minutes.

dylandreimerink:
Yea, its a bit odd. It almost looks like the test runner gets stuck, perhaps a struck test. It seems flaky, we had the same on the initial push, which was resolved after a retry. 

Unfortunately the CI action gets killed before the test framework itself, so we don't get any info about which test might be stuck. The default go timeout is 10m as well, so I will see if we can extend the runner timeout to something like 15m so that the go tests can exit on its own accord on a stuck test and give us some clue about its source.

dylandreimerink (approved):

lmb (approved):
Thanks!

PR #1420, opened 2024-04-08
when ringbuffer map size is not a power-of-2 multiple of page size, prompt users in a user-friendly way

hengyoush:
example:
```c
struct {
	__uint(type, BPF_MAP_TYPE_RINGBUF);
	__uint(max_entries, 4097);
} rb SEC(".maps");
```
then output: 
```
program dev_hard_start_xmit: map rb: EINVAL (ring map size 4097 not a power of two page size(pageSize 4096))
```

### DIFF
--- a/map.go
+++ b/map.go
@@ -490,6 +490,15 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 			return fmt.Errorf("map create: %w", haveFeatErr)
 		}
 	}
+	// BPF_MAP_TYPE_RINGBUF's max_entries must be a power-of-2 multiple of kernel's page size.
+	if errors.Is(err, unix.EINVAL) &&
+		(attr.MapType == sys.BPF_MAP_TYPE_RINGBUF || attr.MapType == sys.BPF_MAP_TYPE_USER_RINGBUF) {
+		pageSize := uint32(os.Getpagesize())
+		maxEntries := attr.MaxEntries
+		if maxEntries%pageSize != 0 || !internal.IsPow(maxEntries) {
+			return fmt.Errorf("map create: %w (ring map size %d not a multiple of page size %d)", err, maxEntries, pageSize)
+		}
+	}
 	if attr.BtfFd == 0 {
 		return fmt.Errorf("map create: %w (without BTF k/v)", err)
 	}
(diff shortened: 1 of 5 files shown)


map.go
```diff
@@ -432,6 +432,17 @@ func (spec *MapSpec) createMap(inner *sys.FD, opts MapOptions) (_ *Map, err erro
 
 	fd, err := sys.MapCreate(&attr)
 
+	// BPF_MAP_TYPE_RINGBUF's max_entries must be a power-of-2 multiple of kernel's page size.
```
florianl, line 435: Thanks for your contribution!

Can we move this error check to `handleMapCreateError()`? `handleMapCreateError()` is a central place to extend the returned error with more information.

hengyoush: done

florianl (approved):

lmb:
I've pushed a small refactoring to reuse an existing implementation of `pow` we already had in the code base.

map.go
```diff
@@ -490,6 +490,16 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 			return fmt.Errorf("map create: %w", haveFeatErr)
 		}
 	}
+	// BPF_MAP_TYPE_RINGBUF's max_entries must be a power-of-2 multiple of kernel's page size.
+	if errors.Is(err, unix.EINVAL) &&
+		(attr.MapType == sys.BPF_MAP_TYPE_RINGBUF || attr.MapType == sys.BPF_MAP_TYPE_USER_RINGBUF) {
+		pageSize := uint32(os.Getpagesize())
+		maxEntries := attr.MaxEntries
+		nPages := maxEntries / pageSize
+		if maxEntries%pageSize != 0 || !internal.IsPow(nPages) {
```
lmb, line 499: @hengyoush I think kernel actually checks that `maxEntries` is a power of two, not `nPages`: https://elixir.bootlin.com/linux/v6.8.4/source/kernel/bpf/ringbuf.c#L194 Do you think that is correct?

lmb (changes requested):

map.go
hengyoush: Thank you for the correction.

dylandreimerink (approved):

PR #1421, opened 2024-04-08
example: add tcp option address example

ytlm:


lmb:
Hi @ytlm, thanks for your contribution. I tried diffing your example with the tcprtt example, and it seems that you want to show how the BPF side code works?

We try to only include examples where the bulk of the "interesting code" is related to how to use the library, so I currently think that this repo is not the right place for this example.

PR #1422, opened 2024-04-09
btf: refuse reloTypeIDTarget for kmod types

lmb:
reloTypeIDTarget is used to substitute the ID of an equivalent type in vmlinux. This is problematic when dealing with kmod types, since their IDs are defined to be sequential with vmlinux. If the last vmlinux type ID is 99, the first type in kmod a and b has ID 100. To disambiguate between these we need a (BTF ID, Type ID) tuple, which we currently don't support.

Poison any relocation which tries to use the target ID of a kmod type. This also gets rid of mergedSpec which didn't take overlapping ID ranges into account.

### DIFF
--- a/btf/core.go
+++ b/btf/core.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


btf/core.go
```diff
@@ -265,15 +246,29 @@ func CORERelocate(relos []*CORERelocation, targets []*Spec, bo binary.ByteOrder,
...
 		}
 
-		targets := mergeTarget.NamedTypes(newEssentialName(localTypeName))
-		fixups, err := coreCalculateFixups(group.relos, &mergeTarget, targets, bo)
+		essentialName := newEssentialName(localTypeName)
+
+		var targetTypes []Type
+		for _, target := range targets {
+			namedTypeIDs := target.imm.namedTypes[essentialName]
+			targetTypes = slices.Grow(targetTypes, len(namedTypeIDs))
```
dylandreimerink, line 260: Why do we need to grow the slice manually? Why not let the append take care of that?

lmb: It's an unbenchmarked optimisation based on my gut feeling. Let me know if you prefer I removeit.

dylandreimerink: Right, i think its going to help much, but it also don't think it will hurt :+1:

dylandreimerink (approved):

PR #1423, opened 2024-04-09
perf: fix TestPerfReaderWakeupEvents

lmb:
The WakeupEvents limit is per ring, one per CPU. And when we execute BPF_PROG_RUN multiple times, we sometimes execute on different CPUs. This means that samples don't all go into a single ring.

Fix this by forcing the producer to run on a single CPU.

Co-developed-by: Dylan Reimerink <dylan@isovalent.com>

### DIFF
--- a/internal/testutils/cpu.go
+++ b/internal/testutils/cpu.go
@@ -1,0 +1,32 @@
+package testutils
+
+import (
+	"runtime"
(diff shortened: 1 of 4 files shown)


lmb:
I'll go ahead and merge this since it makes CI pretty flaky.

PR #1425, opened 2024-04-10
link: allow retrieving perf event file descriptor

lmb:


lmb:
Use would be something like:

```
kp, _ := Kprobe()

if pe, ok := kp.(link.PerfEvent); ok {
   file, err := pe.PerfEvent()
}

// Same for Uprobe
```

* Naming of the method isn't great :(
* I'll document this in the exported callers of `attachPerfEvent`, but still hard to discover. Probably OK since this is a power user thing anyways?

olsajiri:
using it in here https://github.com/cilium/tetragon/pull/1941/commits/9098f1ef886d052b3e3317915648d6ec886f5947 , seems to work nicely, thanks

link/perf_event.go
```diff
@@ -78,6 +79,18 @@ func (pe *perfEvent) Close() error {
 	return nil
 }
 
+// PerfEvent is implemented by some Link types which use a perf even under
```
mmat11, line 82:
```suggestion
// PerfEvent is implemented by some Link types which use a perf event under
```

mmat11 (approved):

olsajiri:
great, thanks a lot!

PR #1426, opened 2024-04-10
Don't mention inexistent outLen parameter in (*Program).Test godoc

tklauser:
The outLen parameter to (*Program).Test was removed in commit e45ae4ab8640 ("detect kernel support for BPF_PROG_TEST_RUN and work around unsafe API"). Remove it from the godoc comment as well.


lmb (approved):
Thanks!

PR #1429, opened 2024-04-10
 perf: flush Record when deadline is exceeded

lmb:
epoll: simplify timeout calculation

epoll: fix eventFd.add

    eventFd.add hardcoded the value to add to 1 instead of using the argument.
    This hasn't caused a bug since we never call add with a different value.

perf: opportunistically check all rings for data

    Waking up user space from the kernel is expensive and so the perf reader 
    allows adjusting the rate at which wakeups happen. This saves CPU at the 
    cost of latency: some data will remain in the buffer for longer.

    The reader is an abstraction over multiple ring buffers. It only reads from
    a ring buffer if it has received a wakeup from the kernel. This is wasteful
    because wakeups are expensive (due to context switching and so on) but
    checking a ring for contents is cheap (just an atomic load).

    Change the behaviour so that we read data from any ready ring buffer 
    regardless of why we were woken up.

perf: flush Record when deadline is exceeded

    Watermark and WakeupEvents allow reducing the rate at which user space is
    woken up. This creates an edge case where Read() returns 
    os.ErrDeadlineExceeded even though there is data in one of the rings.

    This is also a problem in the ringbuf package, where we've solved this by
    checking whether there is any data when encountering an ErrDeadlineExceeded.
    Adopt the same approach for the perf reader.

    TestPerfReaderWakeupEvents has to change since it relies on the old 
    behaviour of not checking for pending data. Rework the test so that it
    doesn't test the ErrDeadlineExceeded error: it's not so important that we
    don't wake up early.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -327,7 +327,11 @@ func (pr *Reader) SetDeadline(t time.Time) {
 //
 // Calling Close interrupts the function.
 //
-// Returns os.ErrDeadlineExceeded if a deadline was set.
+// Returns [os.ErrDeadlineExceeded] if a deadline was set and the perf ring buffer
+// was empty. Otherwise returns a record and no error, even if the deadline was
+// exceeded.
+//
+// See [Reader.ReadInto] for a more efficient version of this method.
 func (pr *Reader) Read() (Record, error) {
 	var r Record
 
@@ -336,7 +340,7 @@ func (pr *Reader) Read() (Record, error) {
 
 var errMustBePaused = fmt.Errorf("perf ringbuffer: must have been paused before reading overwritable buffer")
 
-// ReadInto is like Read except that it allows reusing Record and associated buffers.
+// ReadInto is like [Reader.Read] except that it allows reusing Record and associated buffers.
 func (pr *Reader) ReadInto(rec *Record) error {
 	pr.mu.Lock()
 	defer pr.mu.Unlock()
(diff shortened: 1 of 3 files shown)


perf/reader.go
```diff
@@ -374,6 +374,11 @@ func (pr *Reader) ReadInto(rec *Record) error {
 					pr.epollRings = append(pr.epollRings, ring)
 				}
 			}
+
+			if len(pr.epollRings) == 0 {
+				// Woken up but no data, deadline exceeded.
+				return fmt.Errorf("read: %w", os.ErrDeadlineExceeded)
```
albe19029, line 380: This return without checking errors.Is(err, os.ErrDeadlineExceeded) is invalid.
As in you patch you replace pr.epollEvents to one single event. In case there are 2 signaled files, we will read both of them. And when we will call _, err := pr.poller.Wait(events, pr.deadline) one more time - it will return second signaled file. But it was read. In this case we will return ErrDeadlineExceeded which was not triggered.

brycekahle: I think @albe19029 is correct. From https://man7.org/linux/man-pages/man2/epoll_wait.2.html :
> If more than maxevents file descriptors are ready when epoll_wait() is called, then successive epoll_wait() calls will round robin through the set of ready file descriptors

lmb: Ah interesting, basically a spurious wakeup? I'll have a think!

lmb: I pushed a fixed version, thanks for pointing this out. Important caveat in the current implementation (which applies to `ringbuf` as well):

```go
// Read ...
//
// Returns [os.ErrDeadlineExceeded] if a deadline was set and the perf ring buffer
// was empty. Otherwise returns a record and no error, even if the deadline was
// exceeded.
```

This means we will never return ErrDeadlineExceeded as long as there is data in the rings. Kind of crucial for the purpose of the change, but also weird? Not sure.

florianl: Thinking more about this, should we continue with this draft?
It is a slightly change in the API (behaviour) but most users should benefit, I think.

lmb: How are we going to solve this? Do you think this behaviour is OK?

> This means we will never return ErrDeadlineExceeded as long as there is data in the rings.

florianl: Yeah - thats the API (behaviour) change im refering to. Without a bigger change in the API, I think it is not possible to solve this.
But I also think, this behaivour is ok and should benefit most users.

lmb:
Reworked a little bit and fixed test failures. PTAL.

florianl (approved):

PR #1430, opened 2024-04-11
Use latest Go version for ARM64 and VM tests in CI

tklauser:
The test-on-arm64 and vm-test jobs in CI are using the previous stable Go version (currently 1.21) instead of the latest stable version (currently 1.22). The previous stable Go version is already covered by the test-on-prev-go job, so switch the others to use the latest.


lmb (approved):
Yep, seems to be a copy pasta error. Thanks for spotting this!

Issue #1436, opened 2024-04-16
kernel module btf support breaks programs when kernel misconfigured


orishuss:
### Describe the bug

```markdown
full description in a minute
```

### How to reproduce

```shell
uploading a zip in a minute
```

### Version information

github.com/cilium/ebpf v0.14.0

brycekahle:
Yeah, that would probably work. If the program needs the module BTF then the CO-RE relocations will fail and poison the instructions. That error is a little harder to debug, but will still fail regardless.

lmb:
The fix is going to be here somewhere: https://github.com/cilium/ebpf/blob/7719d2ffbbb8b3a595ef747b016939c8c7419e56/linker.go#L149-L155

- LoadKernelModuleSpec should return a discernible error when the load fails due to the file not existing. Something like `errors.Is(err, os.ErrNotFound)`.
- The code above should ignore the error if it is due to not being found.

@orishuss are you up for contributing a fix? We can probably test this by using a bogus `kmodName` and asserting that we don't get an error.

PR #1440, opened 2024-04-17
program: don't return error when kmod BTF is disabled

orishuss:
In kernels where the flag `CONFIG_DEBUG_INFO_BTF_MODULES` is not set, including kernels 5.10 and below (because the flag was introduced in 5.11), loading a program that attaches to kernel module functions and relied on CORE failed because the module's BTF is not available.

This fix allows the program to run, obviously only as long as it only relies on the kernel's BTF and not on the specific module's BTF.

Fixes: https://github.com/cilium/ebpf/issues/1436. Tested on the same example included in the issue: when using the fixed package, the program attaches successfully.

### DIFF
--- a/btf/kernel.go
+++ b/btf/kernel.go
@@ -63,6 +63,7 @@ func LoadKernelSpec() (*Spec, error) {
 //
 // Defaults to /sys/kernel/btf/<module>.
 // Returns an error wrapping ErrNotSupported if BTF is not enabled.
+// Returns an error wrapping fs.ErrNotExist if BTF for the specific module doesn't exist.
 func LoadKernelModuleSpec(module string) (*Spec, error) {
 	kernelBTF.RLock()
 	spec := kernelBTF.modules[module]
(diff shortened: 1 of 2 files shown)


dylandreimerink (approved):
Changes look good to me. Note that the commit isn't signed off

lmb:
Could you extend `testdata/manyprogs.c` to include one section for `kprobe/ctnetlink_fill_info` or similar? Then we should get coverage of this on our 5.10 CI run.

lmb:
I've had to drop the test in manyprogs.c since the file is only used from benchmarks. I tried to come up with a way to test this in CI, but it's not straight forward :( I also updated your first commit to include the useful PR description.

Unrelated to this, I want to make sure that the new semantics actually make sense.

- If your kernel lacks kmod BTF you will be able to load a kprobe with an AttachTo in a kmod. This is necessary to keep the old behaviour.
- You won't get an error from the library if your code tries to use types from the module. Instead you'll get some harder-to-diagnose error from the verifier. This is because we don't know upfront whether a program does rely on types from the kmod or not.

lmb:
Thanks for your contribution!

PR #1441, opened 2024-04-17
[go.mod] Change `go` directive to `1.21`

mx-psi:
Relates to #1438

Signed-off-by: Pablo Baeyens <pablo.baeyens@datadoghq.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cilium/ebpf
 
-go 1.21.0
+go 1.21
 
 require (
 	github.com/go-quicktest/qt v1.101.0


mx-psi:
@lmb would you mind approving the workflows?

lmb:
@mx-psi sorry, I got sidetracked!

lmb (approved):
Seems like things work! Not sure what broke originally. Mind adding some context to the commit message?

mx-psi:
I amended the commit to include a more descriptive commit message :)

lmb:
So, I've figured out why I didn't do it this way from the start: it breaks toolchain forward compat. Trying to run a command inside the module with go < 1.21 doesn't work since 1.21 is not a valid toolchain.

I'm not exactly sure what the consequences of that are, but overall it makes me think that the original change was the right one.

mx-psi:
>  Trying to run a command inside the module with go < 1.21 doesn't work since 1.21 is not a valid toolchain.

Can you give me an example of the command you are trying to run? If it causes trouble, I surely don't want to do this on my project either!

lmb:
This is for a different repo, but the same rules apply:

```
$ go1.21.0 build
$ go mod edit -go=1.22
$ grep -F 1.22 go.mod
go 1.22
$ go1.21.0 build
go: downloading go1.22 (linux/amd64)
go: download go1.22 for linux/amd64: toolchain not available
$ go mod edit -go=1.22.0
$ go1.21.0 build
go: downloading go1.22.0 (linux/amd64)
$ 
```

I guess `toolchain go1.22.0` would also fix this, but that seems more cumbersome for little (no?) gain.

mx-psi:
@lmb Looks like this is getting fixed upstream: golang/go@27ed85d4d1702e868730ab6ea2ad6326988c615c. The fix is also being backported to Go 1.21 (golang/go/issues/67235) and Go 1.22 (golang/go/issues/67236).

lmb:
Thanks for digging this out!

PR #1443, opened 2024-04-20
perf: do not overwrite named return

florianl:
When calling `newPerfEventRing()` a new variable named `err` is created which conflicts with the named return of the function `NewReaderWithOptions()`. As a consequence the defer function to clean up, that works on the named return of the function, can not work correctly if an error is returned by `newPerfEventRin()` and so may leak unclosed FDs.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -231,23 +231,24 @@ func NewReaderWithOptions(array *ebpf.Map, perCPUBuffer int, opts ReaderOptions)
 	// Hence we have to create a ring for each CPU.
 	bufferSize := 0
 	for i := 0; i < nCPU; i++ {
-		ring, err := newPerfEventRing(i, perCPUBuffer, opts)
-		if errors.Is(err, unix.ENODEV) {
+		ring, ringErr := newPerfEventRing(i, perCPUBuffer, opts)
+		if errors.Is(ringErr, unix.ENODEV) {
 			// The requested CPU is currently offline, skip it.
 			rings = append(rings, nil)
 			pauseFds = append(pauseFds, -1)
 			continue
 		}
 
-		if err != nil {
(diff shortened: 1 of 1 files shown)


lmb (commented):
Is this something that bit you in prod or did you find this when reading the code?

Looking at it some more, I think the problem is that the line `poller, err := epoll.New()` before the defer shadows the `err` returned from the function.

Overall this pattern is very brittle, maybe we can / should reuse the `closeOnError` pattern?

florianl:
> Is this something that bit you in prod or did you find this when reading the code?

I did find it by reading the code.

> Overall this pattern is very brittle, maybe we can / should reuse the `closeOnError` pattern?

I'm not sure, how the `closeOnError` pattern could help here.  When looking for the `closeOnError` use, we have the shadowing of the named return `err` in other places as well - e.g.
- https://github.com/cilium/ebpf/blob/20ddfe1e69f60db0852e2f84a859c2cabb0af235/linker.go#L293
- https://github.com/cilium/ebpf/blob/20ddfe1e69f60db0852e2f84a859c2cabb0af235/linker.go#L430
- https://github.com/cilium/ebpf/blob/20ddfe1e69f60db0852e2f84a859c2cabb0af235/map.go#L316
- https://github.com/cilium/ebpf/blob/20ddfe1e69f60db0852e2f84a859c2cabb0af235/map.go#L418

florianl:
I got bitten by the implicit assignment of defer statements and the code is working as expected, so closing this issue.

When reading the Go spec again and the section about [Defer statements](https://go.dev/ref/spec#Defer_statements), I have learned about the assignment that happens before the Defer statement is executed. Within the scope of `NewReaderWithOptions()` return is always called with the expected return parameters. These parameters are then assigned to the named return and the Defer statement can access these named returns.

PR #1444, opened 2024-04-23
perf: clean up error handling and bare fds

lmb:
The perf reader was written before we had the sys.FD abstraction and therefore still uses plain int to keep track of fds. This means we don't get leak tracking during CI. It also relies on a subtle interaction between defer and named return values which is easy to break by accidentally shadowing err:

    func ... (err error) {
        poller, err := poller.New(...)

        defer func() {
            if err != nil {} // Refers to the second err
        }()
    }

Rework the code to use sys.FD and the closeOnError idiom used elsewhere in the library.

### DIFF
--- a/perf/reader.go
+++ b/perf/reader.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/epoll"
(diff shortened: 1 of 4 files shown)


lmb:
@florianl your intuition was right that the error handling path in the perf reader is wonky. It's really old code by now, I think this should be better.

florianl (approved):
Thanks for the cleanup! I think it is easier now to follow.

PR #1448, opened 2024-04-26
Make some bpf2go output types/funcs public

connorworley:
This makes the necessary types/funcs public in order to implement a workaround for https://github.com/cilium/ebpf/discussions/1142. I will write a script to run these against an already built BPF ELF.


lmb:
Hi @connorworley, sorry for the long wait. It's been difficult to decide what to do about bpf2go. Please take a look at #1494 and let me know if that would fulfill your needs.

Issue #1451, opened 2024-04-26
TestHaveProgramType/Extension fails on kernels >6.7


lmb:
### Describe the bug

Running the testsuite against new kernels results in an error:

```
--- FAIL: TestHaveProgramType (0.07s)
    --- FAIL: TestHaveProgramType/Extension (0.00s)
        feature.go:30: Feature 'Extension' isn't supported even though kernel is newer than v5.6
FAIL
FAIL	github.com/cilium/ebpf/features	0.670s
FAIL
```

### How to reproduce

Using newest vimto:

```
vimto -kernel :stable-selftests -- go test -short -count 1 ./features
```

### Version information

main

dylandreimerink:
I dumped the error which causes the probe to fail which is:
```
load program: invalid argument:
	Cannot replace static functions
	processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```

Logic related to this error was changed in this patch set: https://lore.kernel.org/all/20231215011334.2307144-1-andrii@kernel.org/

dylandreimerink:
I think I figured it out. [This patch](https://lore.kernel.org/all/20231215011334.2307144-3-andrii@kernel.org/) adds the following logic:
```
		/* if main BPF program has associated BTF info, validate that
		 * it's matching expected signature, and otherwise mark BTF
		 * info for main program as unreliable
		 */
		if (env->prog->aux->func_info_aux) {
			ret = btf_prepare_func_args(env, 0);
			if (ret || sub->arg_cnt != 1 || sub->args[0].arg_type != ARG_PTR_TO_CTX)
				env->prog->aux->func_info_aux[0].unreliable = true;
		}
```

So if a program is a main program (it is directly attached / an entrypoint) then its BTF must have exactly 1 argument, that being a pointer to CTX. But in our current probe the func info doesn't include a parameter. This causes the first XDP program to be marked `unreliable`, which causes the `Cannot replace static functions` error when trying to load the extension. Fix PR incoming

PR #1452, opened 2024-04-26
CI: Update kernel from v6.7 to v6.8

dylandreimerink:
This PR updates the kernels we test against in CI to the latest stable kernel, v6.8. However, the feature probe for the extension program type fails on this kernel, so we also include a fix to the probe.

In kernel v6.8, logic was added that restricts valid func info BTF for programs that can be replaced with an BPF_PROG_TYPE_EXT program. Before this change the parameters of the extension program had to be the same as the original program. However, now the verifier asserts that the BTF func info of "main" programs must have exactly 1 parameter of a pointer type to the context struct, even if the code doesn't use the context like in our case.

Fixes: #1451


lmb:
Thanks for diagnosing this! Can you add a link to the upstream commit in the commit message? You'll also have to update the kernel versions we test against in CI.

rgo3 (approved):
Thank Dylan!

PR #1453, opened 2024-04-30
Add get methods for name of program and map

aibor:
When working with objects loaded from fds, it can be useful to have access to their names. This is already possible via their *Info structs. The objects themself have a name field, though, which are provisioned with the *Info's value. So, the name is already present in the objects, but not exposed yet.

This commit adds methods to Program and Map to access the value of their name field.

### DIFF
--- a/map.go
+++ b/map.go
@@ -541,6 +541,11 @@ func (m *Map) String() string {
 	return fmt.Sprintf("%s#%v", m.typ, m.fd)
 }
 
+// Name returns the name of the map.
+func (m *Map) Name() string {
+	return m.name
+}
+
 // Type returns the underlying type of the map.
 func (m *Map) Type() MapType {
 	return m.typ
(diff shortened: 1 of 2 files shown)


aibor:
I agree that exposing basically the same value at two locations is not ideal. Also, the name string length limitation and potential for surprise potential are definitely not ideal as well.

When retrieving name via Info(), those limitations apply as well, though. But you also need to handle the error that Info() may return. There is a reason why this field exists in the object structs and is not always retrieved via Info(). So, since the field is already there, it appeared natural to expose it as well.

In case the change is approved, I should probably add documentation for the the limitations (length and kernel version requirement) to avoid surprises before it is merged.

lmb:
I agree with Timo, exposing this information in addition to `.Info()` is a bit problematic. Timo already mentioned the problems with old kernels, but there are additional wrinkles with maps and programs retrieved from a pin. Also `NewMapFromFD` and similar.

aibor:
Fair enough. Thanks for the review and the explanation. Appreciate it. :)

Issue #1350, opened 2024-02-19
Add test suite for netkit devices


ti-mo:
https://github.com/cilium/ebpf/pull/1257 proposes to add support for attaching netkit programs, but we don't want to take on a dependency to vishvananda/netlink, even for testing. We've decided to drop the netkit tests for now, since the patchset is small and largely based on existing code, and kernel support isn't shipping in major distros for a while.

This issue is for 2 things:
- add netkit support to https://github.com/jsimonetti/rtnetlink
- add tests for the behaviour introduced in https://github.com/cilium/ebpf/pull/1257

ti-mo:
> @ti-mo I can work on adding netkit support to `jsimonetti/rtnetlink`

Perfect! Any ETA?

brlbil:
Not sure, I will try to push a PR upstream as soon as time allows.

brlbil:
@ti-mo we will need to wait for a new version of `golang.org/x/sys` package which should be released soon, at least according to their release history. Netkit related symbols are there but not in a tagged version yet.

lmb:
Hi @brlbil! Any news?

brlbil:
Hi @lmb We need some significant changes upstream. I will open an issue to discuss what would be the best way to proceed upstream.

brlbil:
@lmb @ti-mo I have created an issue upstream for discussion https://github.com/jsimonetti/rtnetlink/issues/216

brlbil:
The PR to merge drivers including `netkit`, https://github.com/jsimonetti/rtnetlink/pull/221

PR #1463, opened 2024-05-10
link: add test suite for netkit

brlbil:
The netkit link support was added without a test suite due to a lack of support for driver-specific extractions in the `github.com/jsimonetti/rtnetlink` package. The driver support has been added to `rtnetlink`.  This PR adds a test suite for the netkit link and pulls in the `github.com/jsimonetti/rtnetlink/v2` package for testing.

Fixes: #1350


link/netkit_test.go
```diff
@@ -0,0 +1,103 @@
...
+func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachType) (Link, int) {
+	var err error
+	conn, err := rtnetlink.Dial(nil)
+	qt.Assert(tb, qt.IsNil(err))
+
+	layer2 := driver.NetkitModeL2
+	blackhole := driver.NetkitPolicyDrop
+	err = conn.Link.New(&rtnetlink.LinkMessage{
+		Family: unix.AF_UNSPEC,
+		Index:  1000,
```
lmb, line 71: This means we can't call the function more than once per test or concurrently.

Is there a way to get the kernel to allocate an ID + retrieve the ID from the netlink answer somehow?

lmb (commented):
Left some questions, thanks for putting in the work to contribute the support upstream!

@ti-mo remind me, adding a dependency to tests doesn't cause downstream consumers to depend on rtnetlink?

link/netkit_test.go
```diff
@@ -0,0 +1,103 @@
...
+	iface, err := net.InterfaceByName("nk1")
+	qt.Assert(tb, qt.IsNil(err))
+
+	link, err := AttachNetkit(NetkitOptions{
+		Program:   prog,
+		Attach:    attachType,
+		Interface: iface.Index,
+	})
+	qt.Assert(tb, qt.IsNil(err))
+	tb.Cleanup(func() {
```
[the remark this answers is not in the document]
brlbil: I see, will fix it.

link/netkit_test.go
brlbil: If we do not provide the index the kernel will assign the index, the same goes for the name, just like the ip command line tool without giving any arguments. We are getting the link with `net.InterfaceByName` so we do not need to provide the index. But even if we get rid of the index we still have a hard-coded interface name. I am not sure what we can do about that. If we do not give the name the kernel will assign names starting from nk0 and nk1, and whatever available next.

brlbil: Could we get the interface index from the calling test function? It would be the caller's responsibility to give a unique index for every call. Also, we would not need to call `net.InterfaceByName` nor return the index from `net.InterfaceByName`. Also, conn can come from the caller so that if we need to call the function more than once we would not need to create new connections every time. What do you think?

lmb: > Could we get the interface index from the calling test function? It would be the caller's responsibility to give a unique index for every call.

I like this, seems straightforward enough.

>  Also, conn can come from the caller so that if we need to call the function more than once we would not need to create new connections every time. What do you think?

That would make it more cumbersome to call (because we need another func to dial the conn). I'd just dial a new conn as needed until it turns out that doesn't work for some reason.

ti-mo: Why not generate a random string at the start of the function and use it as the interface name?

lmb: Can do, ultimately this isn't going to happen often I'd guess. So manually allocating indices is probably not too much of a hassle.

brlbil: Push changes.

lmb: Ended up adding a small sequence generator which allocates ids automatically.

Issue #406, opened 2021-09-07
ci: uprobes integration test on Go binaries fails on kernel < 4.14


mmat11:
[This test](https://github.com/cilium/ebpf/blob/master/link/uprobe_test.go#L338) should pass once this get fixed

ti-mo:
@mmat11 Is there a plan to move forward with this? If you don't have the resources to investigate this in the near future, maybe best to add a comment to the API and close this up.

mmat11:
I've retested on a debian 9 VM and it works:

```
bpf prog
SEC("uprobe/traceme")
int uprobe_traceme(struct pt_regs *ctx) {
	bpf_printk("OK");
	return 0;
}
```
```
tracer
...
ex, err := link.OpenExecutable("/tmp/tracee")
if err != nil {
    log.Fatalf("opening executable: %s", err)
}

up, err := ex.Uretprobe("main.traceme", objs.UprobeTraceme, nil)
if err != nil {
    log.Fatalf("creating uprobe: %s", err)
}
defer up.Close()
...

tracee
func main() {
	for {
		time.Sleep(time.Second)
		traceme(1, 2)
	}
}

//go:noinline
func traceme(a, b int) {
	fmt.Println(a + b)
}
```
```
matt@debian9-univ:~$ /tmp/tracee
3
3
3
^C

matt@debian9-univ:~$ sudo cat /sys/kernel/debug/tracing/trace_pipe
[sudo] password for matt: 
tracee-7677  [006] ....  7341.527519: : OK
tracee-7677  [006] ....  7342.527641: : OK
tracee-7677  [006] ....  7343.527723: : OK

matt@debian9-univ:~$ uname -r
4.9.0-13-amd64

matt@debian9-univ:~$ cat /boot/config-4.9.0-13-amd64 |grep -e BPF -e UPROBE -e FUNCTION_TRACER
CONFIG_ARCH_SUPPORTS_UPROBES=y
CONFIG_BPF=y
CONFIG_BPF_SYSCALL=y
# CONFIG_BPF_JIT_ALWAYS_ON is not set
CONFIG_UPROBES=y
CONFIG_NETFILTER_XT_MATCH_BPF=m
CONFIG_NET_CLS_BPF=m
CONFIG_NET_ACT_BPF=m
CONFIG_BPF_JIT=y
CONFIG_HAVE_EBPF_JIT=y
CONFIG_HAVE_FUNCTION_TRACER=y
[...]

PR #1464, opened 2024-05-10
perf_event_open cpu should be -1 for valid PID

brycekahle:
This will match libbpf behavior. Relevant snippet from `perf_event_open` [manpage](https://man7.org/linux/man-pages/man2/perf_event_open.2.html#DESCRIPTION)

>        pid > 0 and cpu == -1
>              This measures the specified process/thread on any CPU.

### DIFF
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -274,7 +274,11 @@ func pmuProbe(args tracefs.ProbeArgs) (*perfEvent, error) {
 		}
 	}
 
-	rawFd, err := unix.PerfEventOpen(&attr, args.Pid, 0, -1, unix.PERF_FLAG_FD_CLOEXEC)
+	cpu := 0
+	if args.Pid != perfAllThreads {
+		cpu = -1
+	}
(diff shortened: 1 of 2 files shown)


mmat11 (approved):
lgtm, this only affected Uprobes with PID>0, right?

lmb:
I'm missing context, sorry. What is currently broken? Can you extend the commit message to include the explanation? And is it possible to add a test?

brycekahle:
Theoretically, if you specify a PID then the arguments given mean it is applying only to CPU 0. In practice, it may not be affecting anything because the perf <-> tracepoint/kprobe linkage has some gaps.

lmb:
Hmm, could this be the cause of https://github.com/cilium/ebpf/issues/406 ?

mmat11:
> Hmm, could this be the cause of #406 ?

I believe not, in that test `opts` is nil so we trace all pids (and the cpu value is correct in that case)

brycekahle:
Updated commit message

PR #1465, opened 2024-05-11
Add RecursionMisses to access ProgramInfo recursion misses counter

olsajiri:
Adding RecursionMisses to access ProgramInfo recursion misses counter.

It does not depend on whether collection of statistics is enabled, the value is always counted.

### DIFF
--- a/info.go
+++ b/info.go
@@ -83,6 +83,9 @@ type programStats struct {
 	runtime time.Duration
 	// Total number of times the program was called.
 	runCount uint64
+	// Total number of times the programm was NOT called.
+	// Added in commit 9ed9e9ba2337 ("bpf: Count the number of times recursion was prevented").
+	recursionMisses uint64
 }
 
 // ProgramInfo describes a program.
@@ -125,8 +128,9 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 		Name: unix.ByteSliceToString(info.Name[:]),
 		btf:  btf.ID(info.BtfId),
 		stats: &programStats{
-			runtime:  time.Duration(info.RunTimeNs),
-			runCount: info.RunCnt,
+			runtime:         time.Duration(info.RunTimeNs),
+			runCount:        info.RunCnt,
+			recursionMisses: info.RecursionMisses,
 		},
 	}
 
@@ -257,6 +261,16 @@ func (pi *ProgramInfo) Runtime() (time.Duration, bool) {
 		return pi.stats.runtime, true
 	}
 	return time.Duration(0), false
+}
+
+// RecursionMisses returns the total number of times the program was NOT called.
+// This can happen when another bpf program is already running on the cpu, which
+// is likely to happen for example when you interrupt bpf program execution.
+func (pi *ProgramInfo) RecursionMisses() (uint64, bool) {
+	if pi.stats != nil {
+		return pi.stats.recursionMisses, true
+	}
+	return 0, false
 }
 
 // Instructions returns the 'xlated' instruction stream of the program
--- a/info_test.go
+++ b/info_test.go
@@ -262,6 +262,14 @@ func TestStats(t *testing.T) {
 		t.Errorf("expected a runtime of 0ns but got %v", rt)
 	}
 
+	rm, ok := pi.RecursionMisses()
+	if !ok {
+		t.Errorf("expected recursion misses info to be available")
+	}
+	if rm != 0 {
+		t.Errorf("expected a recursion misses of 0 but got %v", rm)
+	}
+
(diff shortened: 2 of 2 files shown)


info.go
```diff
@@ -259,6 +262,15 @@ func (pi *ProgramInfo) Runtime() (time.Duration, bool) {
 	return time.Duration(0), false
 }
 
+// RecursionMisses returns the total number of times the program was NOT called.
+//
+// It is INDEPENDENT on whether collection of statistics is enabled or not, but
+// it keeps the return values pattern as the other stats accessors and returns
+// bool (allways true) together with the counter value.
+func (pi *ProgramInfo) RecursionMisses() (uint64, bool) {
+	return pi.stats.recursionMisses, true
```
lmb, line 271: Which kernel did this appear on though? The boolean is meant to indicate whether the kernel exposes this flag in general, not whether stats were enabled (IIRC).

olsajiri: it got in v5.11 `9ed9e9ba2337 bpf: Count the number of times recursion was prevented`

I'm actually not sure how is the availability detected, both Runtime and RunCount
return values if `pi.stats != nil ` which seems to be the case always  AFAICS

I can keep the same checks as for the Runtime and RunCount

info.go
```diff
@@ -259,6 +262,15 @@ func (pi *ProgramInfo) Runtime() (time.Duration, bool) {
 	return time.Duration(0), false
 }
 
+// RecursionMisses returns the total number of times the program was NOT called.
```
ti-mo, line 265: What could lead to a program not being called? Sounds important to briefly call out here.

info.go
```diff
@@ -259,6 +262,15 @@ func (pi *ProgramInfo) Runtime() (time.Duration, bool) {
...
+//
+// It is INDEPENDENT on whether collection of statistics is enabled or not, but
```
ti-mo, line 267: I'd remove the caps and make the sentence more straightforward:
```suggestion
// This value is always measured on supported kernels, regardless of statistics being enabled or not.
```

Also drop the part about the bool as we're discussing below. If the bool is always true and a real probe is not planned, we may as well drop it.

info.go
ti-mo: > I'm actually not sure how is the availability detected, both Runtime and RunCount
return values if pi.stats != nil which seems to be the case always AFAICS

Can't remember who contributed it, but iirc they were planning to add a probe, and the bool was added to not have to break API later. Looks like the probe was never implemented, and looking back it hardly seems worth it to load a prog, call prog_run and verify if the run counter increases just so we can return that bool correctly.

In the case of RecursionMisses, that one feels unrealistic to try to reproduce reliably at all. I'd just drop the bool and add a comment about it being 5.11+.

ti-mo (dismissed):
Thanks! Left some feedback. Let's just drop the bool, fix up the comment and call it a day.

info.go
lmb: That breaks when using newProgramInfoFromProc. So unless we deprecate that we should just stick with the established pattern of returning a bool.

olsajiri: ok, keeping the bool then, thanks

lmb:
Pushed a fixup, PTAL. Looking at the upstream commit it's not clear to me where the idea that this is related to program statistics comes from? Seems easier to me to just treat these things as completely independent.

Just returning `pi.stats.recursionMisses, true` doesn't work due to `newProgramInfoFromProc`. Reworked to just use a separate field which works around this problem.

info.go
lmb: pi.stats is nil and therefore the line panics.

olsajiri:
hm, it's per program stats collected under same struct in kernel together with run count and run time

```
struct bpf_prog_stats {
        u64_stats_t cnt;
        u64_stats_t nsecs;
        u64_stats_t misses;
        struct u64_stats_sync syncp;
} __aligned(2 * sizeof(u64));
```

it's incremented through `bpf_prog_inc_misses_counter` call any time there's recursion detected on link and we can't run the program

it makes sense to me to place it under `programStats`

lmb:
I can see your point, moving it out of the struct doesn't really add value. I've changed it back, fixed the panic and added a comment.

info.go
```diff
@@ -83,6 +83,9 @@ type programStats struct {
 	runtime time.Duration
 	// Total number of times the program was called.
 	runCount uint64
+	// Total number of timer the programm was NOT called.
```
mmat11, line 86:
```suggestion
	// Total number of times the program was NOT called.
```

lmb:
CI failures are due to unrelated upstream changes.

lmb (approved):

PR #1488, opened 2024-06-14
map: add Map.Handle() method for retrieving a Map's type information

ti-mo:
What it says on the tin. A do-over from Program.Handle().

cc @olsajiri

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -11,6 +11,7 @@ import (
 	"unsafe"
 
 	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/btf"
 	"github.com/cilium/ebpf/internal"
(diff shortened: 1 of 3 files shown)


map_test.go
```diff
@@ -1791,6 +1792,34 @@ func TestMapPinning(t *testing.T) {
...
 
+func TestMapHandle(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "4.18", "btf_id in map info")
+
+	kv := &btf.Int{Size: 4}
+	m, err := NewMap(&MapSpec{
+		Type:       Hash,
+		KeySize:    kv.Size,
+		ValueSize:  kv.Size,
+		Key:        kv,
```
lmb, line 1803: I'm amazed this all just works!

PR #1491, opened 2024-06-18
Add Flush to manually unblock Read/ReadInto

brycekahle:
When using `WakeupEvents` for perf buffers or `BPF_RB_NO_WAKEUP` for ring buffers, you can enqueue data that does not immediately trigger a reader blocked in `Read/ReadInto` to wakeup.

This PR adds a function `Flush` to manually cause those blocked readers to wakeup. Further, a sentinel error `FlushCompleteError` is returned when all the rings have been read, indicating all the data already queued when `Flush` was called has been read.

### DIFF
--- a/internal/epoll/poller.go
+++ b/internal/epoll/poller.go
@@ -1,16 +1,20 @@
 package epoll
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"os"
 	"runtime"
+	"slices"
 	"sync"
 	"time"
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/unix"
 )
+
+var ErrFlushed = errors.New("data was flushed")
 
 // Poller waits for readiness notifications from multiple file descriptors.
 //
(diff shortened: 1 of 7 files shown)


florianl (approved):
Thanks for the update! :pray:

perf/reader.go
```diff
@@ -22,6 +22,8 @@ var (
 	errEOR    = errors.New("end of ring")
 )
 
+type FlushCompleteError = internal.FlushCompleteError
```
lmb, line 25: I wrote this on slack:
> i'd probably drop FlushCompleteError in favour of ErrFlushed or something along those lines. Flush is different than an exceeded deadline after all.

Why add a dedicated error type?

perf/reader.go
```diff
@@ -374,7 +378,7 @@ func (pr *Reader) ReadInto(rec *Record) error {
 			if errors.Is(err, os.ErrDeadlineExceeded) {
 				// We've hit the deadline, check whether there is any data in
 				// the rings that we've not been woken up for.
-				deadlineWasExceeded = true
+				pr.pendingErr = &FlushCompleteError{Err: os.ErrDeadlineExceeded}
```
lmb, line 381: Like I mentioned before, the two errors are not the same. You probably want this so that checking for `flush || deadline` is just `errors.As(FlushCompleteError)`? I think `errors.Is(ErrDeadlineExceeded) || errors.Is(ErrFlushed)` is more appropriate.

perf/reader_test.go
```diff
@@ -68,6 +68,82 @@ func TestReaderSetDeadline(t *testing.T) {
...
+
+	wait := make(chan int)
+	go func() {
+		wait <- 0
+		_, rem := checkRecord(t, rd)
+		wait <- rem
+	}()
+
+	<-wait
+	time.Sleep(10 * time.Millisecond)
```
lmb, line 129: Why sleep?

internal/epoll/poller.go
```diff
@@ -154,13 +176,22 @@ func (p *Poller) Wait(events []unix.EpollEvent, deadline time.Time) (int, error)
...
 				// we'll keep getting this wakeup until Close() acquires the
 				// lock and sets p.epollFd = -1.
 				return 0, fmt.Errorf("epoll wait: %w", os.ErrClosed)
 			}
+			if int(event.Fd) == p.flushEvent.raw {
+				// read event to prevent it from continuing to wake
+				p.flushEvent.read()
+				events = slices.Delete(events, i, i+1)
+				n -= 1
+				continue
```
lmb, line 192: So the behaviour is that `Wait` will return `0, nil` for a flush. But what about a case where another ring becomes ready simultaneously with flushEvent? In that case we'd get two events from epoll, would discard one and then `return 1, nil`. In that case we drop an `ErrFlushed` from `Reader.Read` which the user is expecting.

How about removing the event here, and doing `return n, ErrFlushed` from `Wait` instead? Then `ringbuf` and `perf` re-export `epoll.ErrFlushed` and the rest can mostly stay the same.

lmb (changes requested):

perf/reader_test.go
brycekahle: because we need the reader to be blocked in `Read` before calling `Flush`

internal/epoll/poller.go
brycekahle: > In that case we drop an ErrFlushed from Reader.Read which the user is expecting.
I don't follow. `Flush` is the one that sets up a pending error of `ErrFlushed`, which will still be returned after all the rings have been read.

The `n` return value from `Wait` isn't even used. Only if an error is returned does a different code path happen. We want `Flush` to behave as if a regular ring epoll event happened, which is what it is doing.

brycekahle: I think we actually want to ensure we read all the events in case there both a `Flush` and `Close` in the same wakeup.

brycekahle:
The test failures on Go 1.21 do not seem to be related to my PR

lmb:
> The test failures on Go 1.21 do not seem to be related to my PR

yeah i think it's due to sometimes getting a slow GH runner or something along those lines. are you able to re-run failed runs? if not we might need to fiddle with permissions.

brycekahle:
> are you able to re-run failed runs?

It didn't seem like I could.

lmb:
@brycekahle both Flush implementations were racing with Read. I've pushed a commit which fixes this, by returning ErrFlushed from the poller as I mentioned. PTAL.

brycekahle:
Looks good

PR #1492, opened 2024-06-20
Fix ringbuffer ring size reporting

usamasaqib:
The data pages of a ring buffer are mapped twice in a contiguous virtual region to allow easy wrap around when writing data. The code returns the ring size as the capacity of this mmaped region. However, this is wrong since the kernel [double maps the underlying pages](https://elixir.bootlin.com/linux/latest/source/kernel/bpf/ringbuf.c#L99).

This PR fixes this by reporting the ring size as half the capacity of the underlying mmap region.

### DIFF
--- a/ringbuf/ring.go
+++ b/ringbuf/ring.go
@@ -77,8 +77,9 @@ func (rr *ringReader) isEmpty() bool {
 	return prod == cons
 }
 
-// The data pages in ring buffers are mapped twice in a single contiguous virtual region
-// Therefore the true size is half the size of the mmaped region
+// To be able to wrap around data, data pages in ring buffers are mapped twice in
+// a single contiguous virtual region.
+// Therefore the returned usable size is half the size of the mmaped region.
 func (rr *ringReader) size() int {
 	return cap(rr.ring) / 2
 }
--- a/ringbuf/ring.go
+++ b/ringbuf/ring.go
@@ -77,8 +77,10 @@ func (rr *ringReader) isEmpty() bool {
 	return prod == cons
 }
 
+// The data pages in ring buffers are mapped twice in a single contiguous virtual region
+// Therefore the true size is half the size of the mmaped region
 func (rr *ringReader) size() int {
-	return cap(rr.ring)
+	return cap(rr.ring) / 2
 }
 
 // Read a record from an event ring.
--- a/ringbuf/reader_test.go
+++ b/ringbuf/reader_test.go
@@ -59,7 +59,7 @@ func TestRingbufReader(t *testing.T) {
 			}
 			defer rd.Close()
 
-			if uint32(rd.BufferSize()) != 2*events.MaxEntries() {
+			if uint32(rd.BufferSize()) != events.MaxEntries() {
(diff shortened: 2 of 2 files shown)


ringbuf/ring.go
```diff
@@ -77,8 +77,10 @@ func (rr *ringReader) isEmpty() bool {
 	return prod == cons
 }
 
+// The data pages in ring buffers are mapped twice in a single contiguous virtual region
+// Therefore the true size is half the size of the mmaped region
 func (rr *ringReader) size() int {
-	return cap(rr.ring)
+	return cap(rr.ring) / 2
```
brycekahle, line 83: I think you need to fix `TestRingbufReader` to accommodate this change.

florianl:
Here we `mmap()` memory, that is later reported via `(rr *ringReader) size() int {}` and is publicly accessable via `(r *Reader) BufferSize() int {}`
https://github.com/cilium/ebpf/blob/c959fd7bf295edeb5da0261c258ab64a501c3de7/ringbuf/ring.go#L27

I'm not yet sure about the proposed change. Maybe updating the documentation to `(r *Reader) BufferSize() int {}` could help, as an alternative?

brycekahle:
As the person who added `BufferSize` in https://github.com/cilium/ebpf/pull/1167/files, the intent was to return the total amount of data the user could write into the ring buffer. We use this primarily to determine ring buffer utilization in tandem with `Remaining`.

usamasaqib:
> Here we `mmap()` memory, that is later reported via `(rr *ringReader) size() int {}` and is publicly accessable via `(r *Reader) BufferSize() int {}`
> 
> https://github.com/cilium/ebpf/blob/c959fd7bf295edeb5da0261c258ab64a501c3de7/ringbuf/ring.go#L27
> 
> I'm not yet sure about the proposed change. Maybe updating the documentation to `(r *Reader) BufferSize() int {}` could help, as an alternative?

Hey @florianl, in the mmap call we set the size as `PAGE_SIZE + 2*size`, where size is the max entries specified by the user. The call doubles it because of the unique memory layout of the buffer discussed above. However, the capacity infact is still `size`.

As Bryce mentioned, `BufferSize` is used as the _total amount of data the buffer can hold_, so it should be adjusted according to this.

brycekahle (approved):

ringbuf/ring.go
```diff
@@ -77,8 +77,10 @@ func (rr *ringReader) isEmpty() bool {
 	return prod == cons
 }
 
+// The data pages in ring buffers are mapped twice in a single contiguous virtual region
+// Therefore the true size is half the size of the mmaped region
```
florianl, line 81: Maybe we can be more explicit here?
```suggestion
// To be able to wrap around data, data pages in ring buffers are mapped twice in
// a single contiguous virtual region.
// Therefore the returned usable  size is half the size of the mmaped region.
```

brycekahle: @florianl Usama is on PTO for the next week or so. Do you have permissions to make changes on his branch? This doc change looks good to me, although there is an extra space between `usable` and `size`.

florianl (approved):

lmb:
Thanks all!

PR #1494, opened 2024-06-21
bpf2go: export core functionality

lmb:
bpf2go: export binding generator

    Moves the code to generate go bindings to its own dedicated package and
    export the necessary functions to use it in bpf2go.

    This change will make it possible to reuse this code in other tools, without
    having to compile the C programs.

    The behaviour of bpf2go is unchanged.

    [ Lorenz: move code and backed out changes around module and identifier.
     These can come later, with tests. ]

    Co-developed-by: Lorenz Bauer <lmb@isovalent.com>
    Signed-off-by: Wouter Dullaert <wouter.dullaert@exoscale.ch>

bpf2go: export compilation

    Move the code necessary to compile a C to an ELF. The behaviour of bpf2go is
    unchanged.

    The code to fix up make-style depfiles remains in bpf2go since it has little
    tests, and is probably used only seldomly.

bpf2go: export targets

    Move target and force users to use one of the predefined targets. Also
    export logic to generate build contraints from goarches.


cmd/bpf2go/internal/module.go
```diff
@@ -0,0 +1,14 @@
...
+
+// CurrentModule returns the module containing the main package.
+func CurrentModule() string {
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		// Fall back to constant since bazel doesn't support BuildInfo.
+		return "github.com/cilium/ebpf"
+	}
+
+	return bi.Main.Path
```
wdullaer, line 13: https://pkg.go.dev/runtime/debug#BuildInfo

This means that any outside program trying to use the code template will insert its own package instead of `github.com/cilium/ebpf`.
So even though the template functions are now public, it is not possible to use them outside of this repo and generate functional bindings.

lmb: Ugh, good point. I guess removing the parameter outright it is.

lmb: Testing this properly is a headache as well.

lmb: Goodbye BuildInfo, it was probably a bad idea from the start.

PR #1496, opened 2024-06-23
allow to pin kprobe and kprobe_multi links

olsajiri:
we need to pin kprobe and kprobe_multi links, ATM I'm using this, but I guess there's better way at least for standard kprobes

I'm not sure if the issue described in Pin function comment is also valid  for kprobes created via perf_link,
but the pinning seems to work, and that's what we'd like to use (we don't need to re-open the link)

note the wrapRawLink for perf event kprobe needs to be done properly somehow

cc: @lmb @ti-mo

### DIFF
--- a/link/link.go
+++ b/link/link.go
@@ -119,7 +119,7 @@ func wrapRawLink(raw *RawLink) (_ Link, err error) {
 	case UprobeMultiType:
 		return &uprobeMultiLink{*raw}, nil
 	case PerfEventType:
-		return nil, fmt.Errorf("recovering perf event fd: %w", ErrNotSupported)
+		return &perfEventLink{*raw, nil}, nil
 	case TCXType:
 		return &tcxLink{*raw}, nil
 	case NetfilterType:
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -99,28 +99,14 @@ type perfEventLink struct {
 
 func (pl *perfEventLink) isLink() {}
 
-// Pinning requires the underlying perf event FD to stay open.
-//
-// | PerfEvent FD | BpfLink FD | Works |
-// |--------------|------------|-------|
-// | Open         | Open       | Yes   |
-// | Closed       | Open       | No    |
-// | Open         | Closed     | No (Pin() -> EINVAL) |
-// | Closed       | Closed     | No (Pin() -> EINVAL) |
-//
-// There is currently no pretty way to recover the perf event FD
-// when loading a pinned link, so leave as not supported for now.
-func (pl *perfEventLink) Pin(string) error {
-	return fmt.Errorf("perf event link pin: %w", ErrNotSupported)
-}
-
-func (pl *perfEventLink) Unpin() error {
-	return fmt.Errorf("perf event link unpin: %w", ErrNotSupported)
-}
-
 func (pl *perfEventLink) Close() error {
 	if err := pl.fd.Close(); err != nil {
 		return fmt.Errorf("perf link close: %w", err)
+	}
+
+	// when created from pinned link
+	if pl.pe == nil {
+		return nil
 	}
 
 	if err := pl.pe.Close(); err != nil {
@@ -136,6 +122,11 @@ func (pl *perfEventLink) Update(prog *ebpf.Program) error {
 var _ PerfEvent = (*perfEventLink)(nil)
 
 func (pl *perfEventLink) PerfEvent() (*os.File, error) {
+	// when created from pinned link
+	if pl.pe == nil {
+		return nil, ErrNotSupported
+	}
+
 	fd, err := pl.pe.fd.Dup()
 	if err != nil {
 		return nil, err
--- a/link/kprobe_multi.go
+++ b/link/kprobe_multi.go
@@ -130,14 +130,6 @@ func (kml *kprobeMultiLink) Update(prog *ebpf.Program) error {
 	return fmt.Errorf("update kprobe_multi: %w", ErrNotSupported)
 }
 
-func (kml *kprobeMultiLink) Pin(string) error {
-	return fmt.Errorf("pin kprobe_multi: %w", ErrNotSupported)
-}
-
-func (kml *kprobeMultiLink) Unpin() error {
-	return fmt.Errorf("unpin kprobe_multi: %w", ErrNotSupported)
-}
-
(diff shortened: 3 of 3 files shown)


link/perf_event.go
```diff
@@ -99,30 +99,31 @@ type perfEventLink struct {
 
 func (pl *perfEventLink) isLink() {}
 
-// Pinning requires the underlying perf event FD to stay open.
-//
-// | PerfEvent FD | BpfLink FD | Works |
-// |--------------|------------|-------|
-// | Open         | Open       | Yes   |
-// | Closed       | Open       | No    |
```
lmb, line 107: So why is this OK now? Did something change wrt. this table? If I understand correctly, doing the following will not do the intuitive thing of "keep the attachment alive"?
```
link.Pin()
link.Close() // destroys the perf event fd
```

Maybe I misunderstood what "works" means in this context?

olsajiri: so Pin/Close works for kprobe created via perf_link, because Pin increases the link ref, so it won't go away with the Close.. also the perf link increases ref of the kprobe perf event, so it stays when the original fd is closed

but it won't  work for kprobe via perf ioctl, which is fine

lmb: > so Pin/Close works for kprobe created via perf_link, because Pin increases the link ref, so it won't go away with the Close.. also the perf link increases ref of the kprobe perf event, so it stays when the original fd is closed

But we still can't retrieve the FD and so can't load a pinned link?

> but it won't work for kprobe via perf ioctl, which is fine

How does it fail then? Silently?

link/perf_event.go
```diff
@@ -99,30 +99,31 @@ type perfEventLink struct {
...
-// | Open         | Closed     | No (Pin() -> EINVAL) |
-// | Closed       | Closed     | No (Pin() -> EINVAL) |
-//
-// There is currently no pretty way to recover the perf event FD
-// when loading a pinned link, so leave as not supported for now.
-func (pl *perfEventLink) Pin(string) error {
-	return fmt.Errorf("perf event link pin: %w", ErrNotSupported)
+func (pl *perfEventLink) Pin(fileName string) error {
```
lmb, line 102: I think you can just remove this method + `Unpin` and then you'll get the implementation from `RawLink`.

link/link.go
```diff
@@ -119,7 +119,7 @@ func wrapRawLink(raw *RawLink) (_ Link, err error) {
 	case UprobeMultiType:
 		return &uprobeMultiLink{*raw}, nil
 	case PerfEventType:
-		return nil, fmt.Errorf("recovering perf event fd: %w", ErrNotSupported)
+		return &perfEventLink{*raw, nil}, nil
```
lmb, line 122: How do we know this is `perfEventLink` and not `perfEventIoctl`?

Also, `perfEventLink.PerfEvent` doesn't check for `pe == nil` right now and crashes.

link/perf_event.go
lmb: Ok, can we migrate the contents of comment to where it makes sense instead of outright deleting it? Probably on `perfEventIoctl` or similar.

lmb (commented):
Please correct me if I'm wrong @mmat11:

- ioctl attach doesn't have pinning, so we don't allow it.
- bpf_link attach does have pinning. so far we don't allow it because we can't recover `perfEventLink.pe`.
- `perfEventLink.pe` is used by `Close()` and returned by `PerfEvent()`.
- contrary to ioctl attach, closing the perf event when doing bpf_link attach is not an issue.

If all of these are true, why do we have `perfEvent.pe` in the first place? We could remove the field. To replace `PerfEvent()` we expose an API that explicitly attaches via bpf_link (we've done similar for cgroup IIRC) and returns `(Link, perfEvent *os.File, error)` or similar.

Thoughts?

mmat11:
> Please correct me if I'm wrong @mmat11:
> 
>     * ioctl attach doesn't have pinning, so we don't allow it.
> 
>     * bpf_link attach does have pinning. so far we don't allow it because we can't recover `perfEventLink.pe`.
> 
>     * `perfEventLink.pe` is used by `Close()` and returned by `PerfEvent()`.
> 
>     * contrary to ioctl attach, closing the perf event when doing bpf_link attach is not an issue.
> 
> 
> If all of these are true, why do we have `perfEvent.pe` in the first place? We could remove the field. To replace `PerfEvent()` we expose an API that explicitly attaches via bpf_link (we've done similar for cgroup IIRC) and returns `(Link, perfEvent *os.File, error)` or similar.
> 
> Thoughts?

should all be correct, I think we can remove `perfEventLink.pe` if the kernel takes care of the perf event fd via bpf_link (does this FD get closed once the refcount gets to 0?)

olsajiri:
I'm trying the pushed code, but I'm not sure where to put the comments with fd/link states.. @lmb

mmat11:
> @mmat11 do you think they are still useful? Seems like the table doesn't match the reality of bpf_link attached with a perf_event (why?) and also not for the ioctl case (since that doesn't have a link fd in the first place).

I think we can remove the comment, it refers only to the bpf_link implementation which now has pinning

> Seems like the table doesn't match the reality of bpf_link attached with a perf_event (why?)

These are some quick tests I made a while ago ("works" refers to the link, not the pinning) so I don't exactly remember how I got to these results, and that's what I could observe at the time. The only line that doesn't match should be `// | Closed       | Open       | No    |`, one explanation could be that I forgot to Pin() before closing the perf event fd (?)

> and also not for the ioctl case (since that doesn't have a link fd in the first place).

This table refers to bpf_link only

lmb:
Makes me think we should see whether we can return `ErrNotSupported` from `RawLink.Pin` directly instead of hardcoding it in a bunch of places.

lmb:
What about uprobe_multi, can that be pinned as well?

olsajiri:
> What about uprobe_multi, can that be pinned as well?

yep, added ;-) thanks

olsajiri:
hm, I guess we could actually just remove Pin/Unpin methods from kprobe/uprobe multi links as well.. it'll be taken care of via raw link

mmat11 (approved):

lmb:
Also removed the shadowed Pin/Unpin methods from kprobe multi, they aren't necessary.

PR #1500, opened 2024-07-02
program, btf: probe correct log buffer size

lmb:
ci: fix golangci-lint

    For some reason the Action decided to break now, after being updated a
    couple of weeks ago.

program, btf: probe correct log buffer size

    The kernel exposes a log of operations to aid debugging a program or BTF
    load. Until recently there was no way to know the size of that buffer, so
    the library forces the user to specify a buffer size. From a cursory survey
    on sourcegraph.com it seems that users either hardcode a large fixed buffer
    or probe the correct size by doubling the buffer size when hitting ENOSPC.

    Since commit 47a71c1f9af0 ("bpf: Add log_true_size output field to return
    necessary log buffer size") the kernel does provide a hint to user space.

    Move probing of the correct log buffer size into the library. On recent (>=
    6.4) kernels this is guaranteed to work in a single call. On older kernels
    we use the doubling strategy employed by cilium and tetragon.

### DIFF
--- a/prog_test.go
+++ b/prog_test.go
@@ -463,31 +463,12 @@ func TestProgramVerifierOutput(t *testing.T) {
 	}
 }
 
-// Test all scenarios where the VerifierError.Truncated flag is expected to be
-// true, marked with an x. LL means ProgramOption.LogLevel.
-//
-// |      | Valid | Invalid |
-// |------|-------|---------|
-// | LL=0 |       |    x    |
-// | LL>0 |   x   |    x    |
-func TestProgramVerifierLogTruncated(t *testing.T) {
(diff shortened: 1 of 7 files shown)


prog_test.go
```diff
@@ -969,8 +945,8 @@ func TestProgramLoadErrors(t *testing.T) {
 		want error
 	}{
 		{"poisoned_single", errBadRelocation},
-		{"poisoned_double", errBadRelocation},
-		{"poisoned_kfunc", errUnknownKfunc},
+		// {"poisoned_double", errBadRelocation},
```
florianl, line 948: Is this a left over?

dylandreimerink (approved):
LGTM other than the bit Florian already pointed out

PR #1501, opened 2024-07-02
ci: replace golangci-lint with staticcheck

lmb:
golangci-lint is not better than staticcheck for our purposes, in fact it includes a copy of the latter's ruleset.
It also just broke the build by somehow breaking a tagged version of the GitHub Action.

Just replace it with staticcheck.


ti-mo:
@lmb Isn't staticcheck a completely different linter with coverage for different aspects? Not sure if a bug shipped in the GH Action justifies getting rid of it.

lmb:
It is, I also believe that golanci-lint is lower quality than staticcheck.

ti-mo:
Doing some PR housekeeping. I'm certainly open to changing this if there are clear benefits, so far the reasons are a little unsubstantiated. Thanks!

PR #1435, opened 2024-04-15
link: add Info() method to all link types

rgo3:
This PR pulls out retrieving linktype-specific metadata from `RawLink.Info()`. This will reduce the amount of syscalls when calling `RawLink.Info()` directly. If needed the inktype-specific metadata can now be retrieved through linktype-specific `Info()` methods.

PR #1507, opened 2024-07-10
link: fix nil pointer dereference in AttachXDP

gavinbunney:
With the change in #1435, failures to attach to a XDP link cause a nil pointer dereference. This change ensures any error returned from `AttachRawLink` is handled correctly.


lmb (approved):
Thanks, added a basic test.

PR #1508, opened 2024-07-10
kprobe: attach kprobe / kretprobe based on the SectionName

### LINKED ISSUE #1497: program: make it harder to attach Kretprobe via link.Kprobe and vice versa
Kprobe and Kretprobe both have the same program type, so the library can't guard against using the wrong attachment hooks (see https://github.com/cilium/ebpf/discussions/1490#discussioncomment-9821116). Figuring out whether a program is a kretprobe currently involves doing a string match on `ProgramSpec.SectionName`. It would be nice to do one at least of the following.

- Make it so that `link.Kprobe` returns an error when passed a Kretprobe program. This requires either figuring out whether a program is a Kretprobe (hard, because programs can be loaded from a pin) or communicating it to `link.Kprobe` somehow. Maybe we can shoehorn this into `ProgramSpec.AttachTo` somehow, which is often passed directly to `link.Kprobe` etc. But does this create other breakage?
- Make it easier to figure out if a ProgramSpec is a Kretprobe somehow (though a method on ProgramSpec seems really ugly?)
- Add an example which shows how to attach kprobe / kretprobe based on the SectionName.

Ghostbaby:
Signed-off-by: zhuhuijun <zhuhuijunzhj@gmail.com>

fixed #1497


ti-mo:
Hi, sorry for the slow response.

I'm a little confused by this patch. What problem does it solve? You've added some code that relies on ProgramInfo.Name (which gets truncated to 20 characters), but I'm not sure why. May I assume you want to 'fix' the issue by documenting it as an example?

I think the library can do more to be more helpful, but on the other hand, this all feels like a bit of a non-issue to me. There's nothing that semantically differentiates a kprobe from a kretprobe program, so trying to prevent the user from attaching something to the 'wrong' hook feels a bit counterproductive.

@lmb WDYT as you initially created the issue. Do you have more context about the situation this popped up in?

PR #1509, opened 2024-07-10
bpf2go: enable ebpf code reuse across go packages

mejedi:
Extract imports from "bpf2go.hfiles.go" in the output dir, scan packages for header files, and expose headers to clang. C code consumes headers by providing a go package path in include directive, e.g.

```
bpf2go.hfiles.go:
  package awesome
  import (
    _ "example.org/foo"
  )

frob.c:
  #include "example.org/foo/foo.h"
```

It is handy for sharing code between multiple ebpf blobs within a project. Even better, it enables sharing ebpf code between multiple projects using go modules as delivery vehicle.

By listing build dependencies in a .go file, we ensure that they are properly reflected in go.mod.

### DIFF
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -4,6 +4,9 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"go/build"
+	"go/parser"
+	"go/token"
 	"io"
 	"os"
 	"os/exec"
@@ -11,6 +14,7 @@ import (
 	"regexp"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/cilium/ebpf"
@@ -273,6 +277,10 @@ func (b2g *bpf2go) convertAll() (err error) {
 		}
 	}
 
+	if err := b2g.addHeaders(); err != nil {
+		return fmt.Errorf("adding headers: %w", err)
+	}
+
 	for target, arches := range b2g.targetArches {
 		if err := b2g.convert(target, arches); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


cmd/bpf2go/main.go
```diff
@@ -282,6 +290,77 @@ func (b2g *bpf2go) convertAll() (err error) {
...
+			return err
+		}
+		if pkg.Dir == "" {
+			return fmt.Errorf("%s is missing locally: consider 'go mod download'", path)
+		}
+		if len(hfiles(pkg)) == 0 {
+			fmt.Fprintf(b2g.stdout, "Package doesn't contain .h files: %s\n", path)
+			continue
+		}
+		pkgs = append(pkgs, pkg)
```
derlaft, line 332: What if instead of looking for a statically named file `bpf2go.hfiles.go` you looked for all the imports of the current package? Would it be an improvement? Motivation: If you are writing a library which contains both eBPF-code and golang code, you already likely importing it in another file.

```suggestion
	curPkg, err := build.Import(".", b2g.outputDir, 0)
	if err != nil {
		return err
	}

	var pkgs []*build.Package
	for _, importedPath := range curPkg.Imports {
		if build.IsLocalImport(importedPath) {
			return fmt.Errorf("local imports are not supported: %s", importedPath)
		}
		pkg, err := build.Import(importedPath, b2g.outputDir, 0)
		if err != nil {
			return err
		}
		if pkg.Dir == "" {
			return fmt.Errorf("%s is missing locally: consider 'go mod download'", importedPath)
		}
		if len(hfiles(pkg)) == 0 {
			continue
		}

		fmt.Fprintf(b2g.stdout, "Detected a package containing .h files: %s\n", importedPath)
		pkgs = append(pkgs, pkg)
```

mejedi: > What if instead of looking for a statically named file bpf2go.hfiles.go you looked for all the imports of the current package?

This would be an improvement in the ease-of-use department.
I am reluctant to make this change for the following reasons (big to small):
 * It IS hard to reason about a system with circular dependencies. With proposed changes, package in the `outputDir` is both consumed by `bpf2go` (scanning for imports) AND augmented with generated output. Note: there might be further `go:generate` directives in the package beyond `bpf2go`.
 * One must not forget to `go generate` every time an import is added or removed anywhere in the package. (The problem is not exactly new, and if you've invested in CI it can catch your hand.)
 * Git conflict markers break Golang syntax. Imagine we are dealing with a merge conflict. First, we resolve conflicts in non-generated files. Then we `go:generate`, expecting that it will take care of the generated files. Still some packages are not valid Go yet, throwing `build.Import` off. (Note: it doesn't parse past imports, so chances of catching a Git conflict marker are low. Still, some resilience doesn't hurt. It gets interesting if we consider dependencies: `go generate` is oblivious to dependencies, hence a user can get generated before a dependency which might still have Git conflict markers. If this ever turns out to be a practical issue, we can use `build.FindOnly` flag to prevent `build.Import` from looking at the package code. It is not an option though if `build.Import` is used to determine all imports of a package as proposed.)

On the other hand, everyone accepts that if we split Go code into several files, each file needs `import`s, even though the same thing might have been already imported somewhere else in the current package.

cmd/bpf2go/vfs.go
```diff
@@ -0,0 +1,107 @@
...
+			os.Remove(temp.Name())
+		}
+	}()
+
+	if err = json.NewEncoder(temp).Encode(vfs); err != nil {
+		return "", err
+	}
+
+	return temp.Name(), nil
+}
```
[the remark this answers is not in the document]
mejedi: This is a relatively minor thing that I would like to address in a follow up.

mejedi:
Superseded by #1625

PR #1510, opened 2024-07-10
perf: do not add offline CPUs

florianl:
Adding nil to rings and eventFds for offline CPUs can cause issues when iterating over these elements.
Therefore skip adding to rings and eventFds if the CPU is offline at the time of initalization.


perf/reader.go
brycekahle, line 1: I think there is another nil check in `Close()`

florianl: thanks for the catch! fixed in https://github.com/cilium/ebpf/pull/1510/commits/0ba9749c40a8a1017f43fedcb8be16b49d62a206

brycekahle (approved):

lmb (approved):
Would you mind squashing?

florianl:
:white_check_mark: squashed the commits

PR #1512, opened 2024-07-11
info: expose additional `bpf_prog_info` fields in `ProgramInfo`

tyrone-wu:
Hello! 👋 

I'm implementing a Telegraf plugin that extracts metadata information from the `bpf_prog_info` object. There's some additional fields that I'd like to include in the metric, that aren't exposed in `ProgramInfo`.

The additional fields that are exposes are:
- `jited_prog_len`
- `xlated_prog_len`
- `load_time`
- `verified_insns`

### DIFF
--- a/info.go
+++ b/info.go
@@ -133,9 +133,12 @@ type ProgramInfo struct {
 	haveCreatedByUID bool
 	btf              btf.ID
 	stats            *programStats
+	loadTime         time.Duration
 
-	maps  []MapID
-	insns []byte
+	maps                 []MapID
+	insns                []byte
+	jitedSize            uint32
+	verifiedInstructions uint32
 
 	lineInfos    []byte
 	numLineInfos uint32
@@ -164,6 +167,9 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 			runCount:        info.RunCnt,
 			recursionMisses: info.RecursionMisses,
 		},
+		jitedSize:            info.JitedProgLen,
+		loadTime:             time.Duration(info.LoadTime),
+		verifiedInstructions: info.VerifiedInsns,
 	}
 
 	// Start with a clean struct for the second call, otherwise we may get EFAULT.
@@ -391,6 +397,29 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
 	return insns, nil
 }
 
+// JitedSize returns the size of the program's JIT-compiled machine code in bytes, which is the
+// actual code executed on the host's CPU. This field requires the BPF JIT compiler to be enabled.
+//
+// Available from 4.13. Reading this metadata requires CAP_BPF or equivalent.
+func (pi *ProgramInfo) JitedSize() (uint32, error) {
+	if pi.jitedSize == 0 {
+		return 0, fmt.Errorf("insufficient permissions, unsupported kernel, or JIT compiler disabled: %w", ErrNotSupported)
+	}
+	return pi.jitedSize, nil
+}
+
+// TranslatedSize returns the size of the program's translated instructions in bytes, after it has
+// been verified and rewritten by the kernel.
+//
+// Available from 4.13. Reading this metadata requires CAP_BPF or equivalent.
+func (pi *ProgramInfo) TranslatedSize() (int, error) {
+	insns := len(pi.insns)
+	if insns == 0 {
+		return 0, fmt.Errorf("insufficient permissions or unsupported kernel: %w", ErrNotSupported)
+	}
+	return insns, nil
+}
+
 // MapIDs returns the maps related to the program.
 //
 // Available from 4.15.
@@ -398,6 +427,25 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
 // The bool return value indicates whether this optional field is available.
 func (pi *ProgramInfo) MapIDs() ([]MapID, bool) {
 	return pi.maps, pi.maps != nil
+}
+
+// LoadTime returns when the program was loaded since boot time.
+//
+// Available from 4.15.
+//
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) LoadTime() (time.Duration, bool) {
+	// loadTime and NrMapIds were introduced in the same kernel version.
+	return pi.loadTime, pi.loadTime > 0
+}
+
+// VerifiedInstructions returns the number verified instructions in the program.
+//
+// Available from 5.16.
+//
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) VerifiedInstructions() (uint32, bool) {
+	return pi.verifiedInstructions, pi.verifiedInstructions > 0
 }
 
 func scanFdInfo(fd *sys.FD, fields map[string]interface{}) error {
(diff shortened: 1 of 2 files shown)


lmb:
Hi @tyrone-wu, thanks for your contribution! Please take a look at the comment I wrote in https://github.com/cilium/ebpf/pull/1515 which explains how to extend the Info API. Please also add tests for the API you're introducing.

ti-mo (changes requested):
@tyrone-wu Ping, are you still interested in merging this? This is good to go IMO with one fixup. I'll pull it out of draft so it's more visible.

info.go
```diff
@@ -147,6 +150,9 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 			runCount:        info.RunCnt,
 			recursionMisses: info.RecursionMisses,
 		},
+		jitedSize:            info.JitedProgLen,
+		loadTime:             time.Duration(info.LoadTime),
```
florianl, line 154: Is there a risk of wrong information, as Go's `time.Duration` is of type `int64` while `LoadTime` from the kernel is of type `uint64`?

info.go
```diff
@@ -383,6 +411,25 @@ func (pi *ProgramInfo) MapIDs() ([]MapID, bool) {
...
 }
 
+// LoadTime returns when the program was loaded since boot time.
+//
+// Available from 4.15.
+//
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) LoadTime() (time.Duration, bool) {
+	// loadTime and NrMapIds were introduced in the same kernel version.
+	return pi.loadTime, pi.maps != nil
```
florianl, line 421: Maybe just a personal point of view, but I would not mix `maps` and `loadTime` at this point. At the top of my head, I can't think or know a reason, why a valid `loadTime` should be `0`.  
```suggestion
	return pi.loadTime, pi.loadTime > 0
```

tyrone-wu:
Sorry for the delay, I've been caught in some other things. I can definitely get this done by the end of today. 😺

info.go
tyrone-wu: You're right, `loadTIme` would be 0 if the field isn't available.

info.go
tyrone-wu: Ah, I suppose there could be a risk of it overflowing, although that would mean the program has been loaded for over ~290 years. 😅
Would this be an okay assumption that programs wouldn't be loaded for that long?

dylandreimerink (approved):
Looks good to me, thank you!

florianl (approved):
thanks for the update :+1:

info.go
ti-mo: Don't know if uptime contests are still a thing, but I guess we'll find out in 290 years! ;) Of course this is fine.

info.go
```diff
@@ -391,6 +397,29 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
...
 
+// JitedSize returns the size of the program's JIT-compiled machine code in bytes, which is the
+// actual code executed on the host's CPU. This field requires the BPF JIT compiler to be enable.
+//
+// Available from 4.13. Reading this metadata requires CAP_SYS_ADMIN or equivalent.
+//
+// The bool return value indicates whether the program has been JIT-compiled, or whether this
+// field is available.
+func (pi *ProgramInfo) JitedSize() (uint32, bool) {
+	return pi.jitedSize, pi.jitedSize > 0
```
ti-mo, line 408: If this could be a permissions error, I would follow the same convention as TranslatedSize:

```suggestion
	return 0, fmt.Errorf("insufficient permissions or unsupported kernel: %w", ErrNotSupported)
```

info.go
```diff
@@ -391,6 +397,29 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
 	return insns, nil
 }
 
+// JitedSize returns the size of the program's JIT-compiled machine code in bytes, which is the
+// actual code executed on the host's CPU. This field requires the BPF JIT compiler to be enable.
```
ti-mo, line 401:
```suggestion
// actual code executed on the host's CPU. This field requires the BPF JIT compiler to be enabled.
```

info.go
```diff
@@ -391,6 +397,29 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
...
+func (pi *ProgramInfo) JitedSize() (uint32, bool) {
+	return pi.jitedSize, pi.jitedSize > 0
+}
+
+// TranslatedSize returns the size of the program's translated instructions in bytes, after it has
+// been verified and rewritten by the kernel.
+//
+// Available from 4.13. Reading this metadata requires CAP_SYS_ADMIN or equivalent.
+func (pi *ProgramInfo) TranslatedSize() (int, error) {
+	var err error
```
ti-mo, line 416: This looks a bit atypical, I would write:
```suggestion
	insns := len(pi.insns)
	if insns == 0 {
		return 0, fmt.Errorf("insufficient permissions or unsupported kernel: %w", ErrNotSupported)
	}
	return insns, nil
```

info.go
```diff
@@ -391,6 +397,29 @@ func (pi *ProgramInfo) Instructions() (asm.Instructions, error) {
...
+// The bool return value indicates whether the program has been JIT-compiled, or whether this
+// field is available.
+func (pi *ProgramInfo) JitedSize() (uint32, bool) {
+	return pi.jitedSize, pi.jitedSize > 0
+}
+
+// TranslatedSize returns the size of the program's translated instructions in bytes, after it has
+// been verified and rewritten by the kernel.
+//
+// Available from 4.13. Reading this metadata requires CAP_SYS_ADMIN or equivalent.
```
ti-mo, line 414: If I understand the code comments earlier in the file correctly, `pi.insns` should be populated if the process has CAP_BPF. CAP_SYS_ADMIN is only required for fetching the program's BTF blob, which isn't needed in TranslatedSize().

ti-mo (changes requested):
I thought I'd already left a code review earlier, sorry for the delay.

info.go
tyrone-wu: Ah, you're right. I was originally basing this off of the commit https://github.com/torvalds/linux/commit/1e270976908686ec25fb91b8a34145be54137976#diff-9fb056de263cc4a10925f1474cc0371518419d22bdb4b26d8cfcc7043bfa7883R1292, but it looks like on newer kernels, either `CAP_BPF` or `CAP_SYS_ADMIN` are permitted https://elixir.bootlin.com/linux/v6.1/source/kernel/bpf/syscall.c#L3954.

info.go
tyrone-wu: I believe if JIT compiler is disabled, `jited_prog_len` could also be zero.
It passed the `vimto` test for kernel `v4.14`, however, on mainline Ubuntu kernels below `v4.15`, I found the JIT compiler to be disabled by default.
I included "JIT compiler disabled" in the error message as well, although, I'm not sure it would go along with 
`ErrNotSupported`.

ti-mo: Perfect, thank you!

info.go
ti-mo: Awesome, thanks for checking!

ti-mo (approved):
Thanks for the patch! I gave this a quick rebase due to https://github.com/cilium/ebpf/pull/1563, CI needed to have its toolchain image upgraded.

PR #1519, opened 2024-07-18
map: do not allocate on lookup when key doesn't exist

### LINKED ISSUE #1516: map Lookup allocates when key does not exist
### Describe the bug

In a hot loop of Lookup where a lot of keys do not exist, the `fmt.Errorf` at https://github.com/cilium/ebpf/blob/fbb9ed862764a07745604fc608dc09534e76cd36/map.go#L698 can allocate a decent chunk of memory

### How to reproduce

1. Create map
2. Do a bunch of lookups for keys that do not exist
3. Profile allocations

### Version information

github.com/cilium/ebpf v0.15.1-0.20240709101333-5976561b28aa

brycekahle:
Fixes #1516

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -245,6 +245,17 @@ func TestMapLookupKeyTooSmall(t *testing.T) {
 	qt.Assert(t, qt.IsNotNil(m.Lookup(uint32(0), &small)))
 }
 
+func TestMapLookupKeyNotFoundAllocations(t *testing.T) {
+	m := createArray(t)
+	defer m.Close()
+	var key, out uint32 = 3, 0
+
+	allocs := testing.AllocsPerRun(5, func() {
+		_ = m.Lookup(&key, &out)
+	})
+	qt.Assert(t, qt.Equals(allocs, float64(0)))
+}
+
 func TestBatchAPIMapDelete(t *testing.T) {
 	if err := haveBatchAPI(); err != nil {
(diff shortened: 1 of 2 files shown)


florianl (approved):

map_test.go
```diff
@@ -245,6 +245,13 @@ func TestMapLookupKeyTooSmall(t *testing.T) {
 	qt.Assert(t, qt.IsNotNil(m.Lookup(uint32(0), &small)))
 }
 
+func TestMapLookupKeyNotFoundAllocations(t *testing.T) {
+	allocs := testing.AllocsPerRun(5, func() {
+		_ = wrapMapLookupError(unix.ENOENT)
```
lmb, line 250: Can we do this on Map.Lookup instead?

syscalls.go
```diff
@@ -145,6 +145,13 @@ var haveNoPreallocMaps = internal.NewFeatureTest("prealloc maps", "4.6", func()
 	return nil
 })
 
+func wrapMapLookupError(err error) error {
```
lmb, line 148: Nit: it'd be ok to inline this into `Map.lookup` I think.

map_test.go
brycekahle: Done.

syscalls.go
brycekahle: Done.

lmb (approved):

PR #1520, opened 2024-07-23
program: deprecate LogSize and VerifierError.Truncated

lmb:
We removed LogSize and VerifierError.Truncated in a previous commit, but this causes lots of headache when trying to update the dependency in cilium/cilium and cilium/tetragon. We won't be the only ones suffering, so deprecate the field instead of removing it outright.


lmb:
This undoes the breaking change in https://github.com/cilium/ebpf/pull/1500

PR #1523, opened 2024-07-23
Drop internal.NativeEndian, use binary.NativeEndian

kolyshkin:
Since go 1.21 we can use binary.NativeEndian directly. The main benefit is there's no need to maintain own definition of it.

This change is mostly automated using:

	gofmt -w -r 'internal.NativeEndian -> binary.NativeEndian' .
	goimports .


lmb:
Take a look at https://github.com/cilium/ebpf/pull/1384 where I tried a similar thing. NativeEndian is not comparable which really throws a spanner into the works. We decided that for now there is little benefit to making this change.

If we want to get rid of internal.NativeEndian we probably need to define our own type and then use that for CollectionSpec.ByteOrder.

```go
type Endian int

const (
  BigEndian Endian = iota
  // ...
)
```

kolyshkin:
Oh I see, thanks for details.

PR #1525, opened 2024-07-24
examples: add rewrite_constants example

sancppp:
Add rewrite_constants example.

The function of this sample program is to monitor the number of calls to the target system call made by the process every second. The target system call id is rewritten at load time.


ti-mo:
Hey, thanks for the patch! I'm planning to publish a page on https://ebpf-go.dev soon that demonstrates rewriting constants, as well as some related things about section naming for .rodata etc.

For the purpose of demonstrating rewriting constants, I think this example is much too broad. It interacts with various other subsystems like tracepoints that, while it provides a useful example, detracts a lot from the core concept (const rewrites) you're aiming to highlight here.

Instead, could we minimize this to a less-practical/useful bpf program (I know, it sounds dumb) that simply returns the const, and let the user space part call RewriteConstants + Program.Run()? Afterwards, it simply displays the return code of the bpf program. The example could take an integer argument to substitute into the bpf program.

The more focused the example, the faster the user can learn from it.

sancppp:
> Hey, thanks for the patch! I'm planning to publish a page on https://ebpf-go.dev soon that demonstrates rewriting constants, as well as some related things about section naming for .rodata etc.
> 
> For the purpose of demonstrating rewriting constants, I think this example is much too broad. It interacts with various other subsystems like tracepoints that, while it provides a useful example, detracts a lot from the core concept (const rewrites) you're aiming to highlight here.
> 
> Instead, could we minimize this to a less-practical/useful bpf program (I know, it sounds dumb) that simply returns the const, and let the user space part call RewriteConstants + Program.Run()? Afterwards, it simply displays the return code of the bpf program. The example could take an integer argument to substitute into the bpf program.
> 
> The more focused the example, the faster the user can learn from it.

I think the sample program needs to show what can be achieved with rewrite constant, just like the other examples. 
Now, the rewrite_constant sample program functions the same as https://github.com/cilium/ebpf/tree/main/examples/kprobe, but uses the RewriteConstants function.

ti-mo:
Hi @sancppp, I'm working on deprecating RewriteConstants in favor of `CollectionSpec.Variables` (in progress here https://github.com/ti-mo/ebpf/tree/tb/ebpf-variables). It will come with a manual page on ebpf-go.dev and an example, so I'll close this one out. Thank you for your patience!

PR #1532, opened 2024-08-05
docs: make timo code owner

lmb:
Add Timo as the codeowner of the docs/ folder, which mostly gets automated PRs from dependabot.


ti-mo:
Oh, I thought we already did this. Thanks!

PR #1533, opened 2024-08-05
feat(ringbuf): Add AvailableBytes()

dave-tucker:
Adds AvailableBytes(), which is equivalent to ring__avail_data_size in libbpf.

### DIFF
--- a/ringbuf/reader.go
+++ b/ringbuf/reader.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"sync"
 	"time"
+	"unsafe"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal/epoll"
@@ -25,6 +26,8 @@ type ringbufHeader struct {
 	Len uint32
 	_   uint32 // pg_off, only used by kernel internals
 }
+
+const ringbufHeaderSize = int(unsafe.Sizeof(ringbufHeader{}))
 
 func (rh *ringbufHeader) isBusy() bool {
 	return rh.Len&sys.BPF_RINGBUF_BUSY_BIT != 0
@@ -195,4 +198,11 @@ func (r *Reader) BufferSize() int {
 // until you receive a ErrFlushed error.
 func (r *Reader) Flush() error {
 	return r.poller.Flush()
+}
+
+// AvailableBytes returns the amount of data available to read in the ring buffer in bytes.
+func (r *Reader) AvailableBytes() int {
+	// Don't need to acquire the lock here since the implementation of AvailableBytes
+	// performs atomic loads on the producer and consumer positions.
+	return int(r.ring.AvailableBytes())
 }
--- a/ringbuf/reader_test.go
+++ b/ringbuf/reader_test.go
@@ -50,6 +50,15 @@ func TestRingbufReader(t *testing.T) {
 				15: {1, 2, 3, 4, 4, 3, 2, 1, 1, 2, 3, 4, 4, 3, 2},
 			},
 		},
+		{
+			name:     "send five samples, every even is discarded",
+			messages: []sampleMessage{{size: 5}, {size: 10}, {size: 15}, {size: 20}, {size: 25}},
+			want: map[int][]byte{
+				5:  {1, 2, 3, 4, 4},
+				15: {1, 2, 3, 4, 4, 3, 2, 1, 1, 2, 3, 4, 4, 3, 2},
+				25: {1, 2, 3, 4, 4, 3, 2, 1, 1, 2, 3, 4, 4, 3, 2, 1, 1, 2, 3, 4, 4, 3, 2, 1, 1},
+			},
+		},
 	}
 	for _, tt := range readerTests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -60,6 +69,8 @@ func TestRingbufReader(t *testing.T) {
 				t.Fatal(err)
 			}
 			defer rd.Close()
+
+			qt.Assert(t, qt.Equals(rd.AvailableBytes(), 0))
 
 			if uint32(rd.BufferSize()) != events.MaxEntries() {
 				t.Errorf("expected %d BufferSize, got %d", events.MaxEntries(), rd.BufferSize())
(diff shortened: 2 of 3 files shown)


ti-mo:
Hi @dave-tucker, could you give an example of a practical use case? A test would be nice to have as well. Thanks!

dave-tucker:
@ti-mo added some assertions to the tests to exercise this API.
Use case is for doing ring buffer tuning work. For example, you may have [some logic in your bpf program](https://github.com/torvalds/linux/blob/eb5e56d1491297e0881c95824e2050b7c205f0d4/tools/testing/selftests/bpf/progs/ringbuf_bench.c#L23-L32) that decides whether to use `BPF_RB_NO_WAKEUP` or `BPF_RB_FORCE_WAKEUP` in `bpf_ringbuf_submit`

By periodically calling `AvailableData()` we can see from userspace how far ahead the producer is from the consumer and whether we're at risk of the ringbuf "wrapping around" and overwriting unread records.

florianl (approved):

brycekahle:
Does #1167 accomplish what you need?

dave-tucker:
@brycekahle I can try it out and compare both.
The difference between the two approaches is that `record.Remaining` is calculated once every time a record is read.
`AvailableData()` will always show the difference between the consumer/producer values at any point in time, so it would handle the case of the producer pointer increasing since last read, but we haven't been woken up to read an event yet.

dave-tucker:
> @brycekahle I can try it out and compare both.

I tried it out.

- `AvailableData` was really easy to wire up to a `prometheus.GaugeFunc` as it easily be queried once every scrape interval.
- `Record.Remaining` was less easy. I used a `prometheus.Guage` and `Set` the value within my ringbuf reading function after a record was read.

You can see this on the graphs below - Top is `AvailableData()`, Bottom is `Record.Remaining`
Screenshot 2024-08-07 at 14 35 42

The gap in the middle was when I restarted my process to only use `BPF_RB_FORCE_WAKEUP` when there are 10 entries in the ringbuffer - each entry is 80 bytes.

The top graph is more useful for my purposes of evaluating what ☝️ does to the ring buffer.

I find the bottom graph less useful for my purposes right now, but thinking it through it might be better suited to creating a histogram.

TL;DR I think both are useful

ringbuf/reader.go
```diff
@@ -195,3 +195,11 @@ func (r *Reader) BufferSize() int {
 func (r *Reader) Flush() error {
 	return r.poller.Flush()
 }
+
+// AvailableData returns the amount of data available to read in the ring buffer in bytes.
+// This is racy during ongoing operations but is still useful for overall information on how a ringbuffer is behaving.
```
lmb, line 200: To me this line raises more questions than it answers. "racy" implies it's somehow not safe to call? I'd probably just drop the line.

ringbuf/reader_test.go
```diff
@@ -75,6 +87,13 @@ func TestRingbufReader(t *testing.T) {
 				t.Fatal("Expected 0 as return value, got", errno)
 			}
 
+			var availableData int
+			for _, m := range tt.messages {
+				// size (aligned to 8 bytes) + 8 bytes for the ringbuf header
+				availableData += int(math.Ceil(float64(m.size)/8)*8) + 8
```
lmb, line 93: Could you use `internal.Align`? ringbuf header size could also be a const?

ringbuf/reader_test.go
```diff
@@ -269,12 +288,18 @@ func TestReaderNoWakeup(t *testing.T) {
...
+
 	ret, _, err := prog.Test(internal.EmptyBPFContext)
 	testutils.SkipIfNotSupported(t, err)
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	availableData := 3 * 16 // 8 bytes + 8 bytes for the ringbuf header
+
+	qt.Assert(t, qt.Equals(rd.AvailableData(), uint64(availableData)))
```
lmb, line 301: Could you do the following instead, for all places where you compare with a constant?
```suggestion
	qt.Assert(t, qt.Equals(rd.AvailableData(), 3* 16))
```

That way we get the constant in the output of the test failure.

lmb (changes requested):
Thanks for checking, very informative, especially with the graphs! You're right that putting .Remaining in a gauge isn't going to be very helpful. Did you end up trying with a histogram?

What did you decide to change based on the metrics? Increase buffer size? Wake up less frequently?

The upside of AvailableData is that it's simple to export and graph. The downside is that we can't extend the concept to `perf`, since that wraps multiple buffers. So far we've always tried to keep the API of both packages very similar, but that is not a deal breaker.

Left some nits, overall I agree with Florian that this is fine to merge.

ringbuf/reader.go
```diff
@@ -195,3 +195,11 @@ func (r *Reader) BufferSize() int {
...
+func (r *Reader) AvailableData() uint64 {
```
lmb, line 201: Can you make this return `int` for consistency with the existing size related fields?

dave-tucker:
> Thanks for checking, very informative, especially with the graphs! You're right that putting .Remaining in a gauge isn't going to be very helpful. Did you end up trying with a histogram?

I played around with it and it looked promising but I ended up bailing on it because I couldn't find a satisfactory size for the buckets.

> What did you decide to change based on the metrics? Increase buffer size? Wake up less frequently?

All the above. I'm still working on it, but I plan to write up what I've learned soon and I'll share it on Slack when I'm done.

> The upside of AvailableData is that it's simple to export and graph. The downside is that we can't extend the concept to `perf`, since that wraps multiple buffers. So far we've always tried to keep the API of both packages very similar, but that is not a deal breaker.
> 
> Left some nits, overall I agree with Florian that this is fine to merge.

Addressed! Thanks!

ringbuf/reader.go
```diff
@@ -195,3 +195,10 @@ func (r *Reader) BufferSize() int {
 func (r *Reader) Flush() error {
 	return r.poller.Flush()
 }
+
+// AvailableData returns the amount of data available to read in the ring buffer in bytes.
+func (r *Reader) AvailableData() int {
```
ti-mo, line 200:
```suggestion
func (r *Reader) AvailableBytes() int {
```

ti-mo (approved):
@dave-tucker Thanks for the context you've provided so far! Left one nit, it's common in our lib (and in Go in general) to name functions after the unit of data they return. 'Data' is redundant, since what else would a ring buffer contain? :slightly_smiling_face: 'Bytes' would be more useful here. Thanks!

dave-tucker:
Thanks for the review and suggestions @ti-mo. I've incorporated them and force-pushed an update.

ti-mo:
Pushed a small fixup, will merge when CI is green.

PR #1537, opened 2024-08-06
internal/sys: add constants

lmb:
internal/sys: generate constants from unnamed enums

    bpf.h defines a bunch of constants as unnamed enums. This makes them tricky
    to pull out of vmlinux BTF, so instead we rely on the internal/unix shim.
    This causes a small maintenance burden whenever we need a new flag, but more
    importantly it makes some upcoming cross platform work harder.

    Pull the constant definitions out of vmlinux BTF instead of relying on 
    package unix. Unfortunately a bunch of the flags are re-used. This is most
    noticeable for MapFlags, for which the enum in the kernel has become a bit
    of a grab bag. Instead of trying hard to wrap everything into a neat type we
    just generate untyped constants and get rid of MapFlags.

internal/sys: add DEFINEd constants

    Some constants in bpf.h are defined using the C preprocessor and are 
    therefore not available in BTF. Add them manually.


features/map.go
```diff
@@ -227,7 +227,7 @@ func init() {
 }
 
 // MapFlags document which flags may be feature probed.
-type MapFlags = sys.MapFlags
```
lmb, line 230: This is the breaking change the linter is complaining about. I don't think this should make a difference in practice.

ti-mo: Nope, should be fine since external callers never had access to `sys.MapFlags` anyway.

ti-mo (approved):
One suggestion, looks fine otherwise!

internal/cmd/gentypes/main.go
```diff
@@ -104,6 +102,60 @@ import (
...
+			continue
+		}
+
+		unnamedEnums = append(unnamedEnums, e)
+	}
+
+	var consts []btf.EnumValue
+nextConstant:
+	for _, firstValueName := range []string{
+		"BPF_ANY",           // BPF_MAP_UPDATE_ELEM flags
```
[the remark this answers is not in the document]
lmb: Neat idea, thanks! PTAL.

PR #1540, opened 2024-08-07
internal: split out linux specific bits into a separate package

lmb:
internal: move Pin and Unpin to sys

    Pin and Unpin implement library-specific behaviour for BPF object pinning.
    Move it to package sys since that abstracts away BPF things.

internal: move linux specific code to separate package

    Move some linux specific code out of package internal into package linux. As
    a rule of thumb, code probably belongs in this package if it calls into
    package unix.

internal/kconfig: move Find to package linux

    Move Find to package linux, since it depends on deriving a file name from
    the current linux kernel version.

    This means that package kconfig only implements parsing of the kconfig
    format.

examples: add linux build tag to uretprobe, ringbuf, socket_dist

    The two examples call into package unix which is not available on non-Unix
    platforms. Add build tags so that

        GOOS=windows go build ./...

    works without errors.

program: use unix.SIGUSR1 in tests

link: use internal/unix in netkit tests

CI: cross build for Windows instead of Darwin

    CI cross compiles the package and its tests for Darwin to make sure that
    it's possible to edit the source code on other Unix like platforms.

    Up the game a little and cross build for Windows instead of Darwin: this
    flushes out accidental uses of x/sys/unix instead of internal/unix.


link/netkit_test.go
```diff
@@ -5,12 +5,13 @@ import (
...
 	"testing"
 
-	"github.com/cilium/ebpf"
-	"github.com/cilium/ebpf/internal/testutils"
 	"github.com/go-quicktest/qt"
 	"github.com/jsimonetti/rtnetlink/v2"
 	"github.com/jsimonetti/rtnetlink/v2/driver"
-	"golang.org/x/sys/unix"
+
+	"github.com/cilium/ebpf"
```
rgo3, line 12: Should we add a linter that checks if imports are formatted correctly? Happy to open an issue!

dylandreimerink (approved):
:rocket:

rgo3 (approved):

link/netkit_test.go
rgo3: Skipped the issue and just opened https://github.com/cilium/ebpf/pull/1541

PR #1543, opened 2024-08-08
bpf2go: type-safe ebpf.Variable wrappers

smagnani96:
## Status

- [ ] https://github.com/cilium/ebpf/pull/1610: land basic support in bpf2go.
- [ ] Discuss and implement type-safe wrappers for `ebpf.Variable`, `ebpf.VariableSpec` (wish-to-have also for `ebpf.Map` and `ebpf.MapSpec`) 

## Proposal

Supposing to have an ebpf program that defines the following two global variables:

```c
__u64 pkt_count			= 0;
const __u32 ro_variable	= 1;
char var_msg[]			= "Hello World!";
```

The steps performed during the Go code generation are as follows:

1. An alias with the variable name is generated in the Go file:

```go
type PktCountType 	= uint64
type RoVariable 	= uint32
type VarMsgType 	= [16]int8
```

2. A new type definition (non-alias, need new methods) for the variables is performed:

```go
type PktCount	ebpf.Variable
type RoVariable	ebpf.Variable
type VarMsg		ebpf.Variable
```

3. The `.Get` methods for all types are generated:

```go
func (v *PktCount) Get() (PktCountType, error) {
	var ret PktCountType
	return ret, (*ebpf.Variable)(v).Get(&ret)
}

func (v *RoVariable) Get() (RoVariableType, error) {
	var ret RoVariableType
	return ret, (*ebpf.Variable)(v).Get(&ret)
}

func (v *VarMsg) Get() (VarMsgType, error) {
	var ret VarMsgType
	return ret, (*ebpf.Variable)(v).Get(&ret)
}
```

4. The `.Set` methods for the non read-only variables are generated:

```go
func (v *PktCount) Set(val PktCountType) error {
	return (*ebpf.Variable)(v).Set(val)
}

func (v *VarMsg) set(val VarMsgType) error {
	return (*ebpf.Variable)(v).Set(ret)
}
```

4. In case the type alias is supported by the atomic package, and then
    also supported by `ebpf.Variable` (int32, uint32, int64, uint64), then
    an additional `AtomicRef` method is created for the non read-only
	variables to get an reference to the underlying data.

```go
func (v *PktCount) AtomicRef() (*ebpf.Uint64, error) {
	ret, _ := (*ebpf.Variable)(v).AtomicUint64()
	return ret
}
```

From the user perspective, ignoring the error catching, the following
operations can be performed:

```go
...
objs := bpfObjects{}
err := loadBpfObjects(&objs, nil)
...

// ref kept and can be used multiple times
aPktCount := objs.PktCount.AtomicRef()
vPktCount := aPktCount.Load()
aPktCount.Store(rand.Uint64())
...

vRoVariable, _ := objs.RoVariable.Get()
...

varMsg, _ := objs.VarMsg.Get()
varMsg[0] = 'B'
err := objs.VarMsg.Set(varMsg)
```

Suggestions/Comments are very welcome.


smagnani96:
This PR has been rebased on https://github.com/ti-mo/ebpf/tree/tb/ebpf-variables.
After discussing, @ti-mo proceeds with the `Map.Memory()` api and the definition of `ebpf.Variable`, while me (this PR) should provide the needed support for `bpf2go`.

1. https://github.com/cilium/ebpf/commit/99358ae8053c10264aaf5e909fb2bbd91aa7f59b is required but should be introduced in ti-mo updates (open for discussion).
2. https://github.com/cilium/ebpf/commit/9cf6f9fe28cfad3e0758798f4f15059ad1f4ce3c is the commit under review.

mejedi:
It probably makes sense to land basic `bpf2go` variable support first:
  * `VariableSpec`s included in generated `xxxSpecs`; do we need to segregate them similar to today's `xxxProgramSpecs`/`xxxMapSpecs`?
  * `Variable`s included in generated `xxxObjects`.

I see huge value in a type-safe `get`/`set` API you are proposing. It would be nice to have not only `Variable`s covered, but probably `VariableSpec`s, and `Map`s, and `MapSpec`s as well.

Is it possible to devise an approach that works with `Variable`s initially, but can be extended for other object kinds later? Can we generate less code by using generics?

smagnani96:
> It probably makes sense to land basic `bpf2go` variable support first:
> 
> * `VariableSpec`s included in generated `xxxSpecs`; do we need to segregate them similar to today's `xxxProgramSpecs`/`xxxMapSpecs`?
> * `Variable`s included in generated `xxxObjects`.
> 
> I see huge value in a type-safe `get`/`set` API you are proposing. It would be nice to have not only `Variable`s covered, but probably `VariableSpec`s, and `Map`s, and `MapSpec`s as well.
> 
> Is it possible to devise an approach that works with `Variable`s initially, but can be extended for other object kinds later? Can we generate less code by using generics?

Do you suggest to change this PR to provide basic VariableSpecs support in bpf2go or should I open a new one?

Concerning the type-safe API, I can keep exploring solutions. I tried with generics with no luck, but maybe there's still a way...AFAIU CollectionSpec would also need to be a generic holding maps and variables each of different generic types.

mejedi:
@smagnani96 
> Do you suggest to change this PR to provide basic VariableSpecs support in bpf2go or should I open a new one?

It is probably worthwhile to keep this one around as a RFC and to showcase the end goal. One way to handle it is to structure commits such that basic support is introduced first, followed by type-safe APIs. Once @ti-mo changes are merged,  I'd suggest doing a separate PR to introduce basic variable support in `bpf2go`. Keep rebasing this one as needed.

> Concerning the type-safe API, I can keep exploring solutions. I tried with generics with no luck, but maybe there's still a way...

A clever use of `reflect.Value.ConvertibleTo` by the way:)

 It looks like `bpf2go` won't have to generate `Set` and `Get` methods if we could rely on generic `TypedVariable` in `ebpf` package, e.g.:

```go
// in ebpf package
type TypedVariable[T any] Variable
func (v *TypedVariable[T]) Set(val T) error { return (*Variable)(v).Set(val) }
...
```

> AFAIU CollectionSpec would also need to be a generic holding maps and variables each of different generic types.

What I had in mind was `xxxSpecs` `bpf2go` generates. It is similar to `xxxObjects`, but for the specs. `VariableSpec`s should be included in generated `xxxSpecs`. They are useful to initialise constants, and it would be nice to do it in a type-safe manner as well.

smagnani96:
> It is probably worthwhile to keep this one around as a RFC and to showcase the end goal. One way to handle it is to structure commits such that basic support is introduced first, followed by type-safe APIs. Once @ti-mo changes are merged, I'd suggest doing a separate PR to introduce basic variable support in `bpf2go`. Keep rebasing this one as needed.

Perfect, I'll keep this one open just as reference and will track https://github.com/cilium/ebpf/pull/1572. Once merged, I'll push basic support for `VariableSpec`s and `Variable`s.
 
> It looks like `bpf2go` won't have to generate `Set` and `Get` methods if we could rely on generic `TypedVariable` in `ebpf` package, e.g.:
> 
> ```go
> // in ebpf package
> type TypedVariable[T any] Variable
> func (v *TypedVariable[T]) Set(val T) error { return (*Variable)(v).Set(val) }
> ...
> ```

A wrapper would certainly be a good choice, so we can statically define and test methods rather than generating them at runtime. An approach could be as follows, and then in `bpf2go` we just output the type accordingly.

```go
type ROVariable[T any] struct {
	Variable
}

type RWStruct[T any] struct {
	ROVariable[T]
}

type AtomicVariable[T int32 | int64 | uint64 | uint32] struct {
	RWVariable[T]
}

func (a *ROVariable[T]) Get() T {
	...
}

func (a *RWVariable[T]) Set(val T) {
	...
}

func (a *AtomicVariable[T]) AtomicReference() atomic.T { //somehow
	...
}
```

Would you consider such a wrapper also for `VariableSpec`s (and in case it can be applied also to `MapSpec`s) as well?

ti-mo:
@smagnani96 I've pushed up a few changes to https://github.com/cilium/ebpf/pull/1572 today. Note that all 'atomic' methods are gone. It wasn't a great design to begin with and @mejedi suggested mmapping over the Go heap so we can hand out arbitrary pointers instead. The pointer stuff is not part of the initial proposal, this work is currently in a separate branch here: https://github.com/ti-mo/ebpf/commits/tb/atomic-memory.

tl;dr:
```
// VariablePointer returns a pointer to a value of type T that has its
// underlying memory mapped to the memory representing the Variable. Accessing a
// constant Variable is not supported.
//
// T must be a fixed-size type according to [binary.Size]. Types containing Go
// pointers are not valid.
//
// When accessing structs, embedding [structs.HostLayout] may help ensure the
// layout of the Go struct matches the one in the BPF C program.
func VariablePointer[T any](v *Variable) (*T, error)
```

This works for atomics, structs, scalars, etc. Currently no types containing pointers are accepted.

Note that currently, the plan is to _not_ populate `Collection.Variables` in kernels <5.5, which would be a problem if the 'all objects' struct generated by bpf2go includes Variables by default. We can decide to keep these in a separate object, but that would mean the caller needs to load the whole Collection and then Assign() the variables in a separate step, which is not ideal. Curious to hear your opinions on this.

The alternative is to emit Variables, but disable Set, Get and VariablePointer (make them return errors), which is also far from ideal. If we go down this route, we'd need to introduce a compatibility layer that implements Get() using a map lookup, but then we need to carry a `*Map` in Variable. This makes the implementation quite a bit more complex.

mejedi:
@ti-mo 
> Note that currently, the plan is to not populate Collection.Variables in kernels <5.5, which would be a problem if the 'all objects' struct generated by bpf2go includes Variables by default. We can decide to keep these in a separate object, but that would mean the caller needs to load the whole Collection and then Assign() the variables in a separate step, which is not ideal. Curious to hear your opinions on this.

Globals landed in 5.2. Call me biased, but my preference would be to inconvenience (a presumably small number of) users requiring at least 5.2 that can't bump the minimum version to 5.5. As of today, `bpf2go` emits `xxxObjects` made of `xxxPograms` and `xxxMaps`. Should `xxxVariables` join the party, they can declare and use their own `xxxMyObjects` embedding just `xxxPrograms` and `xxxMaps`.

ti-mo:
@mejedi That sounds reasonable, thanks for sharing your perspective. The downside is this is kind of a breaking change; rebuilding the same code base with a newer version of b2g and ebpf-go will no longer work as-is on some systems.

mejedi:
@smagnani96
Not sure about inheritance, will make the job of `CollectionSpec.LoadAndAssign` harder. With the latest changes from @ti-mo every rw `Variable` is capable of giving out the pointer, so there are fewer variations left:
 * `ROVariable` — `Get`
 * `Variable` — `Get`, `Set`, `Addr`

(Not sure how we should name these generics, `TypedVariable` or `VariableT` or something else?)

The overlap is just the `Get` method, we can afford to implement it twice for 2 distinct types.
```go
type ROVariableT[T any] Variable
type VariableT[T any] Variable
``` 
In return, `CollectionSpec.LoadAndAssign` can use `reflect.Value.ConvertibleTo` and avoid further complication.

Note: `.Addr` is how `reflect` [names](https://pkg.go.dev/reflect#Value.Addr) a method to get a pointer to the `Value`. If you ask me, `.AtomicReference` is way too long.

Couple further questions to consider:

1. Should `Get` return `T` or `(T, error)`? Similarly for `Set`.

    It will be more convenient if user didn't have to handle errors. Errors happen because `T` fails to marshal but in  `bpf2go` we are generating these types and we know that they will marshal cleanly.

2. A subset of types has a matching type in `sync/atomic`, such as `atomic.Uin64`, etc. Should `.Addr` return a pointer to type from `sync/atomic` if available?

```go
var objects struct { foo *VariableT[uint64, atomic.Uint64] `ebpf:"foo"` }
objects.foo.Addr().Load()
```
v.s.
```go
var objects struct { foo *VariableT[uint64] `ebpf:"foo"` }
atomic.LoadUint64(objects.foo.Addr())
```
First option guarantees that value is accessed atomically.  It also involves less typing.

Second option is more consistent. `Get`, `Set` and `Addr` types are consistent. Memory is accessed the same whether it is a standalone value or a compound type:
```go
type compound { bar uint64 }
var objects struct { foo *VariableT[compound] `ebpf:"foo"` }
atomic.LoadUint64(objects.foo.Addr().bar)
```

> Would you consider such a wrapper also for `VariableSpec`s (and in case it can be applied also to `MapSpec`s) as well?

Should be easier since we have just one flavour of `VariableSpec`.

ti-mo:
I've updated https://github.com/cilium/ebpf/pull/1572 to emit a Variable to the Collection for every VariableSpec in the CollectionSpec. Get(), Set() and VariablePointer() will return ErrNotSupported on kernels without BPF_F_MMAPABLE. This ensures LoadAndAssign will continue to work as usual after upgrades, but users opting in to interactions with `Variable` can only expect that to work on newer kernels. I think that's more user-friendly than failing LoadAndAssign.

ti-mo:
@smagnani96 FYI https://github.com/cilium/ebpf/pull/1572 has been merged, and https://github.com/cilium/ebpf/pull/1607 was opened to keep track of the off-heap Go pointers, though keep in mind this might take a while to be resolved upstream. Then it's at least another 6 months of Go releases until ebpf-go can start depending on the behaviour.

I suggest we shelve the atomics use case in bpf2go for now, or we special-case `atomic.(U)int*` in `Variable.Set()` and `.Get()`.  Only loads and stores would be supported, but those may cover the majority of use cases. If really needed, we can consider implementing `Variable.Add()`, but that gets weird when used on structs. Let's start with the basics, we can expand as we go.

smagnani96:
Apologies for the absence.
Thanks both @ti-mo and @mejedi for landing the new API.
I like that the `ErrNotSupported` solution for older kernels, no changes to the usual workflow.

## bpf2go basic support 

I just opened  https://github.com/cilium/ebpf/pull/1610. 
It introduces simple support to emit `Variable`s and `VariableSpec`s, no type-safe methods (will continue the discussion here).
There seems to be a lot of changes, but they're mainly due to the third commit (run `make`), where I regenerated the targets for the further testing and update of the tcx example.
Let me know if anything else is missing, or if we prefer to split the PR.

ti-mo:
@smagnani96 I'm planning to continue with https://github.com/cilium/ebpf/pull/1607, but putting it behind a `CollectionOptions.UnsafeMemory` flag. It will take a while before manual memory cleanup is a thing in the Go runtime, and even longer until it trickles down to a Go version targeted by ebpf-go.

I'm not sure if it really makes sense for bpf2go to emit accessors for these, but maybe we can reevaluate if you give this a rebase.

ti-mo:
Closing this for now, since:
- the direct memory access feature will remain experimental (behind a build tag) until the Go runtime supports off-heap finalizers
- users need to decide when/where they want to take out a variable pointer using e.g. `ebpf.VariablePointer[uint32](objs.MyUint32)`, it's already short and expressive, and I don't know what extra functionality bpf2go can provide here
- there's no explicit concept of an atomic variable; users can simply take out e.g. `atomic.Uint32` pointers if they want atomic access out of the box

Let's see where the Go proposal goes and how the experiment is used in the wild before we add something to bpf2go.

PR #1548, opened 2024-08-20
btf: export DeclTag and TypeTag

dylandreimerink:
```
This commit makes the TypeTag BTF type exported. It had not been
exported because it was lumped in with the DeclTag type. However,
no modification to the codebase seems to be needed to accommodate
the usage of TypeTags.
```

```
This commit makes the DeclTag BTF type exported. The DeclTag has been
unexported because its encoding is a bit tricky.

DeclTags on the wire refer to a type by its ID, and an component index.
Unlike other BTF types, the type ID isn't always the actual "target",
instead if the component index is not `-1` it targets a child object of
the type. This commit adds a `Tags` field to all types that can have
DeclTags including Members.

Exception is function parameters. Due to the func -> funcProto ->
parameter encoding, it is possible that BTF deduplication reuses the
same funcProto for multiple functions. The declTags point to the func to
annotate the arguments. Thus, adding a `Tags` field to Param would
cause inaccurate BTF marshalling. Instead, a `ParamTags` field is added
the func as `[][]string` to match the write format more closely to
avoid issues caused by BTF deduplication. This is a compromise between
technical feasibility and user experience. Users need to manually match
up the `ParamTags` with the `Params`. This shouldn't be a huge issue
in most cases however.

This means we now effectively have less "types" in our Go BTF
representation then we have on the wire since declTags become strings
on other types. This might leave "holes" in our BTF type numbering
if we were to remove them. So instead we leave the `declTag` types in
the BTF type list, but we hide them from the user during iteration.
Technically, users can still access them by ID if they know the ID, but
like before this commit, the returned value can only be printed.

Marshalling relies on the iteration of the BTF types. When go to load
a BTF blob from a collection, we iterate from a collection of root types
which are directly used. This causes us to load only BTF info that is
actually in use. This commit introduces a trick into the `children`
logic used during iteration. When we iterate over a type with declTags,
we create internal `declTag` types at-hoc with the correct component
index. So even though the tags are just strings on other types, when
iterating they appear as actual types. This causes only declTags that
are on types which are in use to be marshalled.
```


btf/traversal.go
```diff
@@ -59,17 +59,47 @@ func children(typ Type, yield func(child *Type) bool) bool {
...
+				if !yield(&tag) {
+					return false
+				}
+			}
+		}
+		for _, t := range v.Tags {
+			var tag Type = &declTag{t, v, -1}
+			if !yield(&tag) {
+				return false
+			}
```
mejedi, line 73: I'm curious if it is worthwhile to compress the code by having `var tags []string` before the type switch, moving the iteration all the way down to `return true` below, and replacing these 6 lines (and other similar occurrences) with `tags = v.Tags`.

mejedi: Or a `func yieldTags(typ Type, yield func(child *Type) bool, int component, tags []string) bool` helper.

dylandreimerink: If you look at the `&declTag{...}` bit, the last field is -1 or positive number depending on if its a tag on the main type or a sub-element. In the case of function, we produce both declTags for the main function and function params. So you can do:

```
var tags []*declTag

switch (...){
case *btf.Func:
  ...
  for _,t := v.Tags {
    tags = append(tags, &declTag{t, v, -1})
  }
  ...
  for i := funcParams {
    for_, t := range v. ParamTags[i] {
      tags = append(tags, &declTag{t, v, i})
     }
  }
  ...
}

for _, t := range tags {
  if ! yield(tag) {
    return false
   }
}
```

But then we are not winning much, and I personally find it harder to follow

mejedi: I had a slightly different thing in mind:
```go
var tags []string

switch (...){
case *btf.Func:
  ...
  tags := v.Tags // <-- trade 5 lines for 1
  ...
  for i := funcParams {
    for_, t := range v. ParamTags[i] {
      tags = append(tags, &declTag{t, v, i})
     }
  }
  ...
}

for _, t := range tags { // <-- iterate tags in a common block at the func end
  if !yield(&declTag{t, v, -1}) {
    return false
  }
}
```
It bothers me a little that we have the same 5 line block replicated again and again. But maybe you are right and it is easier to follow the way it is now.

dylandreimerink: > I had a slightly different thing in mind

Right, that makes more sense. Let my try this, that would remove 5 loops, so I think that is better.

btf/btf.go
```diff
@@ -695,5 +695,11 @@ func (iter *TypesIterator) Next() bool {
 	iter.Type, ok = iter.spec.typeByID(iter.id)
 	iter.id++
 	iter.done = !ok
+	if !iter.done {
+		// Skip declTags
+		if _, ok := iter.Type.(*declTag); ok {
+			return iter.Next()
+		}
+	}
```
mejedi, line 703: Should we modify `indexTypes` as well?

How about `fixup` in `readAndInflateTypes`, should we allow (invalid) references to `DeclTag`s in contexts where proper TypeID is expected (e.g. a struct member referencing `DeclTag` instead of a type)?

It is not clear to me whether this filtering is done for user convenience OR to avoid dealing with pre-existing `DeclTag`s during marshalling.

btf/traversal.go
```diff
@@ -40,8 +40,9 @@ func children(typ Type, yield func(child *Type) bool) bool {
 	// Explicitly type switch on the most common types to allow the inliner to
 	// do its work. This avoids allocating intermediate slices from walk() on
 	// the heap.
+	var tags []string
 	switch v := typ.(type) {
-	case *Void, *Int, *Enum, *Fwd, *Float:
+	case *Void, *Int, *Enum, *Fwd, *Float, *declTag:
```
mejedi, line 45: If we have to deal with pre-existing `declTag`s, we should probably recurse into `.Type`.

*Pre-existing*: a user obtains a reference to a `declTag` somehow and submits it for marshalling.

btf/types.go
```diff
@@ -1142,26 +1148,69 @@ func readAndInflateTypes(r io.Reader, bo binary.ByteOrder, typeLen uint32, rawSt
...
+				t.ParamTags[dt.Index] = append(t.ParamTags[dt.Index], dt.Value)
+				continue
+			}
+
+			if dt.Index == -1 {
+				t.Tags = append(t.Tags, dt.Value)
+				continue
+			}
+
+			return nil, fmt.Errorf("type %s: decl tag for type %s has invalid component idx", dt, t)
```
mejedi, line 1212: Minor: can we reject `dt.Index < -1` before entering the type switch? If yes, the code can be simplified
```suggestion
			if dt.Index >= len(fp.Params) {
				return nil, fmt.Errorf("type %s: component idx %d exceeds params of %s", dt, dt.Index, t)
			}

			if len(t.ParamTags) < dt.Index {
				t.ParamTags = make([][]string, len(fp.Params))
			}

			if dt.Index >= 0 {
				t.ParamTags[dt.Index] = append(t.ParamTags[dt.Index], dt.Value)
			} else {
				t.Tags = append(t.Tags, dt.Value)
			}
```

mejedi:
It looks like `.copy` methods on nodes should copy `.Tags`. Apparently, `children` function powers both marshalling and deep copying. In the context of deep copying, creating and yielding `declTag`s is not strictly necessary.

btf/btf.go
dylandreimerink: The situation as it stands is that the iterator returns `*declTag` since its is a `Type` though you can't do anything with it outside of printing it. Because this PR causes decl tags to be merged into the types they refer to, we essentially make holes in the type list:
```
BTF blob
0: void
1: int
2: declTag -> int
3: struct

Go:
0: void
1: int + declTag
3: struct
```

We pretty much represent a BTF blob as `[]Type` everywhere, where the ID of the type is the index into the slice, which in turn match the ID in lets say BTF.ext. So when we fold declTags into other types we need to still fill the `[]Type` slice with something to not mess with the indexes/IDs. I discovered this when initially trying to just not emit a `Type` for declTags. But its causes problem, for anything outside BTF to reference IDs such as relocations or func info.

I tried just making `nil` entries, but we assume every `Type` in the slice is always non-nil, so I was afraid that this would introduce unseen bugs and nil-dereferences. 

So, I came up with 3 ways to work around it:

1. Keep emitting `*declTag` , but just hide them during iteration. That is the smallest change. Users can technically still access them via the TypeByID functions ect, but there is nothing that references declTag IDs so that doesn't seem like a big deal
2. Introduce a new `Type` implementation like "hiddenType" or something, so its more explicit. But given we have declTag already, It seemed excessive
3. Emit a `Void` for every *declTag. Normally only at index 0, it basically means "nothing". However, without similar modifications to the iterator you get a whole bunch of `Void`s which seems really odd and unexpected.

I went for number 1, but if you see a better way of dealing with this then I am open to suggestions.

btf/traversal.go
dylandreimerink: The thought is that we essentially make *declTag an internal type, it shouldn't exist since all tags are part of other types. We create them during traversal as if they hang under the other types (inverted from how they are encoded).

This way, we only traverse decl tags of types that are actually in use then we go an marshal BTF blob

btf/types.go
dylandreimerink: I am sorry, but I do not feel like this is an improvement. It does not save me meaningful amount of lines. It makes the error you will get for cases where -1 is the only valid answer less specific ("numbers below -1 are invalid" vs "-1 is the only valid value") and to me having all of the checks together on one screen makes it easier to read than having to scroll to validate all cases are handled.

mejedi: Sure, I see your point. It would still read easier without `continue` (`switch` `case` maybe?) but up to you.

Could you please help me understand why the following code is necessary?
```go
if len(t.ParamTags) < len(fp.Params) {
	t.ParamTags = append(t.ParamTags, make([][]string, len(fp.Params)-len(t.ParamTags))...)
}
```
The number of parameters does not change. We need to allocate a properly-sized array once.

btf/traversal.go
dylandreimerink: Yes. But since a user can't create a  `declTag` from scratch, they would have to first marshal a BTF blob, then guess the correct ID, get the `btf.Type` and add it to the builder. That would result in a `no ID for type` panic. But that would likely happen with any type you do this with that references another.

ti-mo: We could consider returning an error if the requested ID is for a declTag, but I think that would cause unnecessary confusion and complexity. As long as the happy path is reasonably free of footguns, it's fine, and I think that's the case here.

btf/btf.go
ti-mo: Option 1 seems like the best compromise to me. We're doing our best to abstract over any warts and make things as user-friendly as possible, but if an object has a type ID, it would be surprising if it didn't in our implementation. We can try to limit what the user can do with a `declTag`, but hiding it completely by relegating it to the BTF encoding would require significant internal (and external?) changes.

Essentially, this would mean we need to tolerate holes in a type ID sequence, which goes against many assumptions made by the code. Let's go with this for now. Callers can still pull a `declTag` out of a Spec if they really want to, e.g. for debugging purposes, but it won't be of much use. This was the case before as well.

Let's see if any other declTag-like things get merged down the line, maybe there will be more of these internal types that need special treatment. We can reevaluate if/when that time comes.

btf/testdata/tags.c
```diff
@@ -0,0 +1,33 @@
...
+struct s tagD s1;
+union u tagE u1;
+td tagA t1;
+
+int tagA tagB fwdDecl(char tagC x, char tagD y);
+
+int tagE normalDecl(char tagB x, char tagC y) {
+	return fwdDecl(x, y);
+}
+
```
ti-mo, line 30: Would it make sense to declare two functions with an identical prototype here that could one day be deduplicated by llvm? Might be useful as a canary.

btf/traversal.go
```diff
@@ -40,8 +40,9 @@ func children(typ Type, yield func(child *Type) bool) bool {
...
 		// No children to traverse.
```
ti-mo, line 46: I'd add this:
```suggestion
		// No children to traverse.
		// declTags is declared as a leaf type since it's parsed into .Tags fields of other types
		// during unmarshaling.
```

btf/traversal.go
```diff
@@ -86,6 +102,20 @@ func children(typ Type, yield func(child *Type) bool) bool {
 		if !yield(&v.Type) {
 			return false
 		}
+		tags = v.Tags
```
ti-mo, line 105: Move this to the end of the `case` clause for consistency? I was wondering why we were ignoring v.Tags for Func. :stuck_out_tongue:

btf/types.go
```diff
@@ -540,19 +547,19 @@ func (dt *declTag) copy() Type {
 	return &cpy
 }
 
-// typeTag associates metadata with a type.
-type typeTag struct {
+// TypeTag associates metadata with a type.
```
ti-mo, line 550: Now we're exporting this, could we elaborate in a few sentences what these are used for?

btf/types.go
```diff
@@ -400,9 +404,11 @@ func (r *Restrict) copy() Type {
...
 // Func is a function definition.
 type Func struct {
-	Name    string
-	Type    Type
-	Linkage FuncLinkage
+	Name      string
+	Type      Type
+	Linkage   FuncLinkage
+	Tags      []string
+	ParamTags [][]string
```
ti-mo, line 411: This needs a bit of inline documentation around the expectations of the length of the outer/inner arrays. And when will ParamTags be != nil?

btf/types.go
```diff
@@ -1142,26 +1149,69 @@ func readAndInflateTypes(r io.Reader, bo binary.ByteOrder, typeLen uint32, rawSt
...
 		case *Func:
 			fp, ok := t.Type.(*FuncProto)
 			if !ok {
 				return nil, fmt.Errorf("type %s: %s is not a FuncProto", dt, t.Type)
 			}
 
-			if dt.Index >= len(fp.Params) {
-				return nil, fmt.Errorf("type %s: index %d exceeds params of %s", dt, dt.Index, t)
+			// Ensure the number of argument tag lists equals the number of arguments
+			if len(t.ParamTags) < len(fp.Params) {
```
ti-mo, line 1195: Could this be simplified/clarified a bit? For example, could this be initialized once?
```go
if len(t.ParamTags) == 0 {
  t.ParamTags = make([][]string, 0, len(t.Params))
  for range len(t.Params) {
    t.ParamTags = append(t.ParamTags, make([]string))
  }
}
```

Would that be sufficient? After reading 20 times, I still don't understand what `len(fp.Params)-len(t.ParamTags)` is supposed to do.

ti-mo (commented):
Awesome! Left a few comments, haven't looked at the tests in-depth.

ti-mo (approved):

mejedi:
> It looks like .copy methods on nodes should copy .Tags. Apparently, children function powers both marshalling and deep copying. In the context of deep copying, creating and yielding declTags is not strictly necessary.

Submitted a follow-up to handle `.Tags` in `.copy`: https://github.com/cilium/ebpf/pull/1582

PR #1550, opened 2024-08-22
btf: Add decl tags for functions

olsajiri:
I need declaration tags for functions to be part of program's BTF, I'm using this POC change and it seems to work for me
I guess it could be written much nicer, happy to change that or test another change, thanks


ti-mo:
This should land in https://github.com/cilium/ebpf/pull/1548, @dylandreimerink could you confirm?

olsajiri:
awesome.. will test it #1548 with my changes, thanks

PR #1552, opened 2024-08-26
Support wasm architecture

alban:
cilium/ebpf currently does not compile to wasm with tinygo. This patch makes it compile.

I am working on a website where users can submit ebpf binaries (either in ELF format or as an Inspektor Gadget export) and the website will parse the ebpf binary client-side with wasm. The wasm code is written in Go, using cilium/ebpf and is compiled with tinygo.

My wasm code uses ebpf.LoadCollectionSpecFromReader() to display information about the ebpf binary. But it will not call ebpf.NewCollection() because the wasm/javascript environment in the browser cannot interact with the Linux kernel.

---

The website PoC looks like this:
- https://github.com/alban/bpfanalyser
- https://alban.github.io/bpfanalyser/

image

Since cilium/ebpf is running in the browser, I can use Chrome DevTools to profile the performance:

Screenshot-Chrome-DevTools-cilium-ebpf

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -632,8 +632,16 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
 		wanted = typPtr.Elem().Type()
 	}
 
-	if !wanted.AssignableTo(typeInterface) {
-		return fmt.Errorf("%T does not satisfy Type interface", typ)
+	switch wanted {
+	case reflect.TypeOf((**Datasec)(nil)).Elem():
+		// Those types are already assignable to Type. No need to call
+		// AssignableTo. Avoid it when possible to workaround
+		// limitation in tinygo:
+		// https://github.com/tinygo-org/tinygo/issues/4277
+	default:
+		if !wanted.AssignableTo(typeInterface) {
+			return fmt.Errorf("%T does not satisfy Type interface", typ)
+		}
 	}
 
 	types, err := s.AnyTypesByName(name)
--- a/internal/page_wasm.go
+++ b/internal/page_wasm.go
@@ -1,0 +1,8 @@
+//go:build wasm
+
+package internal
+
+func Getpagesize() int {
+	// A WebAssembly page has a constant size of 65,536 bytes, i.e., 64KiB
+	return 64 * 1024
+}
--- a/syscalls.go
+++ b/syscalls.go
@@ -27,7 +27,14 @@ var (
 // invalidBPFObjNameChar returns true if char may not appear in
 // a BPF object name.
 func invalidBPFObjNameChar(char rune) bool {
-	dotAllowed := objNameAllowsDot() == nil
+	var dotAllowed bool
+	if runtime.GOOS == "js" {
+		// In Javascript environments, BPF objects are not going to be
+		// loaded to a kernel, so we can use dots without testing.
+		dotAllowed = true
+	} else {
+		dotAllowed = objNameAllowsDot() == nil
+	}
 
 	switch {
 	case char >= 'A' && char <= 'Z':
--- a/internal/endian_le.go
+++ b/internal/endian_le.go
@@ -1,4 +1,4 @@
-//go:build 386 || amd64 || amd64p32 || arm || arm64 || loong64 || mipsle || mips64le || mips64p32le || ppc64le || riscv64
+//go:build 386 || amd64 || amd64p32 || arm || arm64 || loong64 || mipsle || mips64le || mips64p32le || ppc64le || riscv64 || wasm
 
 package internal
 
--- a/internal/io.go
+++ b/internal/io.go
@@ -28,7 +28,7 @@ func NewBufferedSectionReader(ra io.ReaderAt, off, n int64) *bufio.Reader {
 	// of 4096. Allow arches with larger pages to allocate more, but don't
 	// allocate a fixed 4k buffer if we only need to read a small segment.
 	buf := n
-	if ps := int64(os.Getpagesize()); n > ps {
+	if ps := int64(Getpagesize()); n > ps {
 		buf = ps
 	}
 
--- a/internal/page_others.go
+++ b/internal/page_others.go
@@ -1,0 +1,9 @@
+//go:build !wasm
+
+package internal
+
+import "os"
+
+func Getpagesize() int {
+	return os.Getpagesize()
+}
--- a/map.go
+++ b/map.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"os"
 	"path/filepath"
 	"reflect"
 	"slices"
@@ -507,7 +506,7 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 	// BPF_MAP_TYPE_RINGBUF's max_entries must be a power-of-2 multiple of kernel's page size.
 	if errors.Is(err, unix.EINVAL) &&
 		(attr.MapType == sys.BPF_MAP_TYPE_RINGBUF || attr.MapType == sys.BPF_MAP_TYPE_USER_RINGBUF) {
-		pageSize := uint32(os.Getpagesize())
+		pageSize := uint32(internal.Getpagesize())
 		maxEntries := attr.MaxEntries
 		if maxEntries%pageSize != 0 || !internal.IsPow(maxEntries) {
 			return fmt.Errorf("map create: %w (ring map size %d not a multiple of page size %d)", err, maxEntries, pageSize)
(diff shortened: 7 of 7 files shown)


alban:
I cleaned up the code and marked the PR as ready.

Wasm is little endian and has a page size of 64 KiB according to its spec.

internal/page_wasm.go
```diff
@@ -0,0 +1,8 @@
+//go:build wasm
+
+package internal
+
+func Getpagesize() int {
+	// A WebAssembly page has a constant size of 65,536 bytes, i.e., 64KiB
+	return 64 * 1024
```
ti-mo, line 7: I don't know much about tinygo, but some Googling reveals that it implements a subset of the Go "os" package, and it looks like it implements os.Getpagesize() as well for the arches it targets. Why is it hardcoded here? This is the crux of the PR but isn't mentioned anywhere.

In any case, I'd like this to go into the respective stdlib, either in Google's implementation or tinygo if needed.

syscalls.go
```diff
@@ -27,7 +27,14 @@ var (
 // invalidBPFObjNameChar returns true if char may not appear in
 // a BPF object name.
 func invalidBPFObjNameChar(char rune) bool {
-	dotAllowed := objNameAllowsDot() == nil
+	var dotAllowed bool
+	if runtime.GOOS == "js" {
+		// In Javascript environments, BPF objects are not going to be
+		// loaded to a kernel, so we can use dots without testing.
+		dotAllowed = true
```
ti-mo, line 34: Why isn't this logic in `objNameAllowsDot` instead? Also, instead of hardcoding a check on `runtime.GOOS`, we should let the syscall fail and conclude that it means a dot can be used.

ti-mo (changes requested):
Hi Alban, very cool project! Note that we don't test wasm builds so this can regress at any time in the future. Feel free to add wasm builds to CI, and maybe create an issue to get (part of) the test suite to run on wasm.

For ebpf-on-windows, we're looking at the opposite: there are no CollectionSpecs on windows, so no ELF loader needed. (They ship bpf in signed winPE instead) Tests that load an ELF and put it into the kernel need to be skipped there. We'll need some generic mechanism to skip tests like we do for kernel versions etc.

internal/page_wasm.go
alban: I get the following error if I try to use os.Getpagesize:
```
$ tinygo build -o wasm.wasm main.go
# github.com/cilium/ebpf/internal
../../cilium/ebpf/internal/page_wasm.go:8:12: undefined: os.Getpagesize
```

According to https://pkg.go.dev/github.com/tinygo-org/tinygo/src/os#Getpagesize, it was added in tinygo 0.24.0 and I use 0.30.0, but it still does not work.

Issue: https://github.com/tinygo-org/tinygo/issues/1285

I agree it should ideally be fixed in tinygo or Go... I could try to make a PR there. But I would like to have a workaround in cilium/ebpf meanwhile. Btw, `syscall.Getpagesize()` works, presumably thanks to https://github.com/tinygo-org/tinygo/issues/2856; could that be an option to use that in cilium/ebpf as a workaround?

In https://github.com/tinygo-org/tinygo/issues/1285, they suggest to return -1... That would not work for cilium/ebpf when used in `unix.Mmap` but we wouldn't call those methods from wasm context. And for `internal.NewBufferedSectionReader(n=-1)`, it seems the implementation would use a buffer of 16 bytes when given a negative number, so that would work (but might be slow).

syscalls.go
ti-mo: I did some reading on this last week, apparently ENOSYS is returned by much of the tinygo syscall package. This would be a good sentinel to conclude that a syscall can't be made and we need to fall back to a sane default. We can make things on the Windows side behave similarly, so high-level logic only needs to check for ENOSYS.

cc @lmb

internal/page_wasm.go
ti-mo: > I could try to make a PR there. But I would like to have a workaround in cilium/ebpf meanwhile.

If we merge a workaround, there's no guarantee you will get around to making that upstream contribution. :wink: I really don't want to hardcode non-bpf-related platform constants, that's definitely the runtime/stdlib's responsibility.

> `syscall.Getpagesize()` works

Underlyingly, `os.Getpagesize()` simply calls `syscall.Getpagesize()` (https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/types.go;l=13), so sounds like a trivial fix in tinygo? Might be a good exercise to understand the underlying tech a bit better.

I prefer to make only the absolute minimum changes to the lib to force us to address issues in the underlying layers. This is good for the health of the platform and potentially enables more software to run correctly out-of-the-box.

lmb:
I blown away that this works, it's a very cool idea!

> Note that we don't test wasm builds so this can regress at any time in the future. Feel free to add wasm builds to CI, and maybe create an issue to get (part of) the test suite to run on wasm.

Before going down this route: adding wasm to CI kind of implies that we will start maintaining a wasm "port". I personally don't have the capacity to do that. I really like this kind of work, but adding wasm support feels like a very steep price to pay to (essentially) get access to the elf reader.

Maybe a more maintainable solution would be to split the elf reader into a separate, cross-platform package? This is in line with the design goals we have for it. It would also mean moving a lot of code around unfortunately, and might get stuck in circular dependency hell.

ti-mo:
> Before going down this route: adding wasm to CI kind of implies that we will start maintaining a wasm "port". I personally don't have the capacity to do that. I really like this kind of work, but adding wasm support feels like a very steep price to pay to (essentially) get access to the elf reader.

If it largely works as-is with a few lines of changes, is there a real downside? If we get some tests to run on tinygo (you just added the infra for it! :wink:), that could be a good way of keeping the elf reader hermetic. (e.g. no touching /proc during ELF->CollectionSpec parsing, see recent example: https://github.com/cilium/ebpf/commit/7f04cae1edd4549ebe890500b3608cfdab451058, this would've broken Alban's use case as well)

This pales in comparison to, let's say, adding ebpf-for-windows support. :wink: 

> Maybe a more maintainable solution would be to split the elf reader into a separate, cross-platform package? This is in line with the design goals we have for it.

Not sure what you mean in practice, but we could potentially move some ELF reader logic into platform-specific internal packages. If you mean moving CollectionSpec and friends into a subpackage, that ship has long sailed I think.

internal/page_wasm.go
lmb: Seems like os.Getpagesize now works!

btf/btf.go
```diff
@@ -632,8 +632,16 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
 		wanted = typPtr.Elem().Type()
 	}
 
-	if !wanted.AssignableTo(typeInterface) {
-		return fmt.Errorf("%T does not satisfy Type interface", typ)
+	switch wanted {
+	case reflect.TypeOf((**Datasec)(nil)).Elem():
```
lmb, line 636: This is a bummer I think. How would we prevent this list from ever expanding as the ELF reader changes?

syscalls.go
```diff
@@ -27,7 +27,14 @@ var (
 // invalidBPFObjNameChar returns true if char may not appear in
 // a BPF object name.
 func invalidBPFObjNameChar(char rune) bool {
-	dotAllowed := objNameAllowsDot() == nil
+	var dotAllowed bool
```
ti-mo, line 30: @alban This code is now gone, you can drop this change.

ti-mo:
Hi @alban, any progress here? Are the Getpagesize() changes still required?

lmb:
Getpagesize is fixed, the reflection changes are the only blocker. I'd be happy to swap our `GOOS=darwin` compile check in CI to wasm if we could somehow make that less painful.

alban:
I have not worked on this topic since opening the PR, and we will likely not need the wasm support in cilium/ebpf anymore, sorry. Feel free to either close this PR or take it over.

@KapilSareen is taking over the [work](https://mentorship.lfx.linuxfoundation.org/project/a6d66c40-3d12-4fa4-88bf-18574f6b4ec0) on analysing ELF BPF programs using cilium/ebpf (https://github.com/inspektor-gadget/inspektor-gadget/pull/4259) but that is done without using WASM.

> I'm a newbie, but this was all that's needed to make GOOS=js GOARCH=wasm go build ./... work:
> [changes in variables_bpfel.go and endian_le.go]

It seems reasonable to make a PR with just those 2 changes to make it work for the official Go implementation. And leave the tinygo-specific changes for later, if someone ever wants to revisit the tinygo support.

> Just curious, what's the benefit of using tinygo for wasm? Is it more mature than the official Go implementation?

It was more mature before, but I am not sure it is still the case. It still produces smaller binaries. But the official Go implementation improved a lot. In Inspektor Gadget, we recently switched to the official Go implementation (https://github.com/inspektor-gadget/inspektor-gadget/pull/4130) because it is faster than tinygo. But IG uses WASM for things unrelated to cilium/ebpf.

ti-mo:
@alban Thanks for the update! Closing this in favor of https://github.com/cilium/ebpf/pull/1757.

PR #1551, opened 2024-08-24
internal: clean up use of golang.org/x/exp

liggitt:
Removes last remaining use of golang.org/x/exp and adds test for new integer types

Fixes https://github.com/cilium/ebpf/issues/1095

PR #1557, opened 2024-09-03
internal: remove dependency on x/exp/constraints

ti-mo:
Temporarily copy the definition of the Integer constraint until the constraints package is moved out of x/exp into the stdlib.

Add a test that makes sure Integer contains all of Go's integer types.

Supersedes https://github.com/cilium/ebpf/pull/1551.


dims:
thanks @ti-mo

PR #1558, opened 2024-09-03
map: remove MapSpec.Freeze field

ti-mo:
This has been an ambiguous flag since its inception. Internally, we would allow maps to be created with BPF_F_RDONLY_PROG, but not freeze them afterwards. Conversely, we would allow MapSpecs with Freeze set to 'true', but without the rdonly map flag set.

The latter case has caused subtle bugs in the past, since BPF_MAP_FREEZE only blocks further modifications from user space, but doesn't actually mark the map as readonly for the verifier. This will prevent code pruning and other optimizations from taking place.

This commit removes the Freeze field in favor of the map flag and adds a helper to keep existing call sites simple. Sourcegraph code search yields no references to this field in open-source code.

### DIFF
--- a/map.go
+++ b/map.go
@@ -73,9 +73,6 @@ type MapSpec struct {
 	// The initial contents of the map. May be nil.
 	Contents []MapKV
 
-	// Whether to freeze a map after setting its initial contents.
-	Freeze bool
-
 	// InnerMap is used as a template for ArrayOfMaps and HashOfMaps
 	InnerMap *MapSpec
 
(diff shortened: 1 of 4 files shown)


map.go
```diff
@@ -1400,7 +1401,7 @@ func (m *Map) finalize(spec *MapSpec) error {
 		}
 	}
 
-	if spec.Freeze {
+	if spec.readOnly() {
```
lmb, line 1404: I don't think this is correct: it is valid for a map to be BPF_F_RDONLY_PROG but mutable from user space. Your change breaks this behaviour.

ti-mo: Ack, will change this to a name-based trigger like libbpf instead.

PR #1559, opened 2024-09-03
map tests queues and refactor

smagnani96:
This PR introduces a new test for `MapIterator` as well as small refactoring to some other tests.
With this, the aim is primarily to ensure that `MapIterator` fails when called on keyless maps such as queues/stacks: this is because the current implementation relies on a key (nextKey) lookup from the map, which is then used to retrieve its respective value.
Commit messages should contain each the explanation, feel free to reach me for further q&a :smiley:

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -838,6 +838,14 @@ func TestMapQueue(t *testing.T) {
 	}
 
 	var v uint32
+	if err := m.Lookup(nil, &v); err != nil {
+		t.Fatal("Lookup (Peek) on Queue:", err)
+	}
+	if v != 42 {
+		t.Error("Want value 42, got", v)
+	}
+	v = 0
+
 	if err := m.LookupAndDelete(nil, &v); err != nil {
 		t.Fatal("Can't lookup and delete element:", err)
 	}
@@ -855,6 +863,10 @@ func TestMapQueue(t *testing.T) {
 
 	if err := m.LookupAndDelete(nil, &v); !errors.Is(err, ErrKeyNotExist) {
 		t.Fatal("Lookup and delete on empty Queue:", err)
+	}
+
+	if err := m.Lookup(nil, &v); !errors.Is(err, ErrKeyNotExist) {
+		t.Fatal("Lookup (Peek) on empty Queue:", err)
 	}
 }
 
--- a/map_test.go
+++ b/map_test.go
@@ -1062,7 +1062,7 @@ func TestIterateEmptyMap(t *testing.T) {
 
 			var key string
 			var value uint64
-			if entries.Next(&key, &value) != false {
+			if entries.Next(&key, &value) {
 				t.Error("Empty hash should not be iterable")
 			}
 			if err := entries.Err(); err != nil {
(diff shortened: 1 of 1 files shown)


dylandreimerink (approved):
Changes all look good to me

map_test.go
```diff
@@ -1091,38 +1103,50 @@ func TestMapIterate(t *testing.T) {
...
+	var (
+		key   string
+		value uint32
+		keys  []string
+		data  = []string{"hello", "world"}
+	)
 
-	if err := hash.Put("world", uint32(42)); err != nil {
-		t.Fatal(err)
+	for i, k := range data {
```
ti-mo, line 1113: Nit: for clarity's sake, keep variable declarations close to where they're used:
```suggestion
	data := []string{"hello", "world"}
	for i, k := range data {
```

(keep key, value, keys where they are)

map_test.go
```diff
@@ -1091,38 +1103,50 @@ func TestMapIterate(t *testing.T) {
...
+		if err := hash.Put(k, uint32(i)); err != nil {
+			t.Fatal(err)
+		}
 	}
 
-	var key string
```
ti-mo, line 1102: Feel free to declare these in a var() block, but this didn't really need changing.

map_test.go
```diff
@@ -1091,38 +1103,50 @@ func TestMapIterate(t *testing.T) {
...
+
+	m, err := NewMap(&MapSpec{
+		Type:       Queue,
+		ValueSize:  4,
+		MaxEntries: 2,
+	})
+	qt.Assert(t, qt.IsNil(err))
+	defer m.Close()
+
+	for _, v := range []uint32{0, 1} {
```
ti-mo, line 1140: Can this go into `MapSpec.Contents` with entries like `MapKV{nil, 0}` instead?

map_test.go
```diff
@@ -1091,38 +1103,50 @@ func TestMapIterate(t *testing.T) {
...
 
-	if n := len(keys); n != 2 {
-		t.Fatal("Expected to get 2 keys, have", n)
-	}
-	if keys[0] != "hello" {
-		t.Error("Expected index 0 to be hello, got", keys[0])
-	}
-	if keys[1] != "world" {
-		t.Error("Expected index 1 to be hello, got", keys[1])
+func TestIterateWrongMap(t *testing.T) {
```
ti-mo, line 1129: I'm not sure if I understand the point of this test. Isn't the idea to make MapIterator support queues and stacks transparently later on? Why do we need to test here that it _doesn't_ work?

ti-mo (changes requested):
Thanks for splitting this up into smaller bits. Please try not to move stuff (like the variable declarations) around without a clear benefit. It causes churn and makes things more time-consuming to review.

map_test.go
smagnani96: The support for queues and stacks will be provided by the `Map.Drain` API, not `Map.Iterate` -- I would also like to later introduce the `TestDrainWrongMap` to expect an error when calling `Map.Drain` with a wrong map (i.e. an array map that doesn't support `LookupAndDelete`) --

If this test seems redundant I can remove it. However, it could help clarifying what is the expected behavior against specific types of maps. Let me know :smiley:

ti-mo: Sorry! That makes a lot of sense, haven't had the opportunity to look at the other patches. In that case, testing that Map.Iterate() never returns any data and always returns an error is perfect. :+1:

smagnani96: Np, thanks for the feedback, a double-check is always appreciated :wink:

ti-mo (approved):

PR #1560, opened 2024-09-06
feat(gen): Make Identifier function configurable

wdullaer:
The standard Identifier function used by the generator results in very unidiomatic go identifier names when used with my ebpf object. With this change, the Identifier function becomes configurable, allowing me to pass in something that produces better outputs in my context. If no function is passed in, we keep using the default, this makes the change entirely backwards compatible.
Furthermore, a test has been added to ensure the configuration is correctly passed through.

### DIFF
--- a/cmd/bpf2go/gen/output.go
+++ b/cmd/bpf2go/gen/output.go
@@ -90,10 +90,15 @@ type GenerateArgs struct {
 	ObjectFile string
 	// Output to write template to.
 	Output io.Writer
+	// Function which transforms the input into a valid go identifier. Uses the default behaviour if nil
+	Identifier func(string) string
 }
 
 // Generate bindings for a BPF ELF file.
 func Generate(args GenerateArgs) error {
+	if args.Identifier == nil {
+		args.Identifier = internal.Identifier
+	}
 	if !token.IsIdentifier(args.Stem) {
 		return fmt.Errorf("%q is not a valid identifier", args.Stem)
 	}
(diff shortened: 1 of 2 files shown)


cmd/bpf2go/gen/output.go
```diff
@@ -90,10 +90,15 @@ type GenerateArgs struct {
 	ObjectFile string
 	// Output to write template to.
 	Output io.Writer
+	// Function which transforms the input into a valid go identifier (internal.Identifier if left nil)
```
ti-mo, line 93: Note: referring to internal symbols in godoc is a bit strange, since they don't appear in the godoc UI. Something like 'uses the default behaviour if nil' is sufficient here.

ti-mo (commented):
I don't know where we stand regarding using bpf2go as a library, or if this fits the vision for the project. Deferring to @mejedi and @lmb for that.

@wdullaer Could you provide some examples of unusual output of `internal.Identifier`? Maybe we could improve the default instead, which could benefit all users.

wdullaer:
We use ebpf programs written in rust, which means the identifiers follow rust convention. Specifically map names are all caps and field names have underscores.

We're using this function:
```go
// Identifier turns a C style type or field name into an exportable Go equivalent.
func Identifier(s string) string {
	s = strings.TrimSpace(s)
	if s == "" {
		return s
	}

	capNext := true
	prevIsCap := false
	return strings.Map(func(v rune) rune {
		vIsCap := unicode.IsUpper(v)
		if capNext {
			if unicode.IsLower(v) {
				v = unicode.ToUpper(v)
			}
			vIsCap = true
		} else if prevIsCap && unicode.IsUpper(v) {
			v = unicode.ToLower(v)
		}
		prevIsCap = vIsCap

		if unicode.IsLetter(v) {
			capNext = false
			return v
		} else if unicode.IsNumber(v) {
			capNext = true
			return v
		} else {
			capNext = v == '_' || unicode.IsSpace(v) || v == '-' || v == '.'
			return -1
		}
	}, s)
}
```

I would be careful with changing the default here, since this might cause churn in clients.
I understand this is very much an aesthetic thing and the original bikeshed :-)

cmd/bpf2go/gen/output_test.go
```diff
@@ -81,3 +82,23 @@ func TestPackageImport(t *testing.T) {
...
+		Package:    "foo",
+		Stem:       "bar",
+		ObjectFile: "frob.o",
+		Output:     &buf,
+		Programs:   []string{"do_thing"},
+		Identifier: identifier,
+	}
+	err := Generate(args)
+	qt.Assert(t, qt.IsNil(err))
+	qt.Assert(t, qt.IsTrue(called))
```
mejedi, line 103: `Called` feels a little artificial, can we use `strings.ToUpper` as `Identifier` and check that `buf` contains `DO_THING` instead?

PR #1572, opened 2024-10-01
Collection: add Variables field to interact with global variables

ti-mo:
See individual commits for more detailed information.

- [ ] Extend Global Variables documentation


internal/unix/types_linux_arm.go
```diff
@@ -0,0 +1,23 @@
...
+
+	linux "golang.org/x/sys/unix"
+)
+
+// TODO: Remove this file and wrap linux.MmapPtr in types_linux.go when upgrading to Go 1.23.
+func MmapPtr(fd int, offset int64, addr unsafe.Pointer, length uintptr, prot int, flags int) (ret unsafe.Pointer, err error) {
+	page := uintptr(offset / 4096)
+	if offset != int64(page)*4096 {
+		return unsafe.Pointer(nil), EINVAL
+	}
```
mejedi, line 16: Doesn't the kernel implement the same checks?

memory.go
```diff
@@ -0,0 +1,227 @@
...
+	// page-aligned on most architectures anyway.
+	size = internal.Align(size, os.Getpagesize())
+
+	// Get the address of the backing array of the slice and check if it's
+	// page-aligned to make sure it works with MAP_FIXED.
+	ptr := unsafe.Pointer(unsafe.SliceData(make([]byte, size)))
+	addr := int(uintptr(ptr))
+	if internal.Align(addr, os.Getpagesize()) != addr {
+		return nil, 0, fmt.Errorf("allocated memory is not page-aligned: %d", addr)
+	}
```
mejedi, line 132: It is surprising that we *reliably* get a page-aligned piece of memory here. Runtime manages various size classes differently, and thresholds are [subject to change](https://go.dev/doc/go1.22#runtime).

Should we get an unaligned chunk, we could retry with +1 page size and re-slice.

memory.go
```diff
@@ -0,0 +1,227 @@
...
+	// randomly decide to move objects around, in which case the mapped region
+	// will not move along with it.
+	if heapObjectsCanMove() {
+		return nil, 0, errors.New("this Go runtime has a moving garbage collector")
+	}
+
+	// Request at least a full page from the runtime, otherwise the allocated span
+	// is likely not page-aligned, risking mapping over objects on another page.
+	// Since we use MAP_FIXED, the starting address of the mapping must be
+	// page-aligned on most architectures anyway.
```
mejedi, line 123: Pardon my ignorance, how does page-misaligned memory mapping work on those architectures?

memory.go
```diff
@@ -0,0 +1,227 @@
...
+
+	if p == nil {
+		return 0, fmt.Errorf("input buffer p is nil")
+	}
+
+	if off < 0 || off >= int64(len(mm.b)) {
+		return 0, fmt.Errorf("read offset out of range")
+	}
+
+	n := copy(p, mm.b[off:])
```
mejedi, line 195: It would be nice if `ReadAt` copied properly aligned chunks in word-size units. Imagine I copy a `struct` with counters that is concurrently updated by bpf program. Should the `copy` decide to go one byte at a time, values can be seen that never were.

I suppose builtin `copy` is probably doing exactly that, albeit non-contractual. (Very minor and can wait until later.)

internal/unix/types_linux_arm.go
ti-mo: Not sure, need to look at kernel source for this. I borrowed this from the stdlib and will be removed after the next Go release.

memory.go
ti-mo: Here's the current table of size classes: https://github.com/golang/go/blob/master/src/runtime/sizeclasses.go. I can't find this mentioned anywhere explicitly, but isn't this just an extension of typical struct alignment rules? If this slice would be a 4k struct instead, it would need to be 4k-aligned to make sure any pointers within the struct would be correctly 8-byte aligned by extension. I don't think an allocator is allowed to return allocations at unaligned offsets, but I could be wrong of course.

We could test this further by allocating some ballast and/or fragmenting the heap before running the tests a large number of times.

memory.go
ti-mo: I'm paraphrasing `man mmap(2)` here.

```
MAP_FIXED
              Don't interpret addr as a hint: place the mapping at
              exactly that address.  addr must be suitably aligned: for
              most architectures a multiple of the page size is
              sufficient; however, some architectures may impose
              additional restrictions. [...]
```

I briefly came across one such additional restriction somewhere, but I forgot, sorry.

Here's another snippet: https://pubs.opengroup.org/onlinepubs/009604499/functions/mmap.html (this seems to be some kind of POSIX spec) 
```
[EINVAL]
    The addr argument (if MAP_FIXED was specified) or off is not a multiple of the page size as returned by sysconf(), or is considered invalid by the implementation.
```

tl;dr - `off` within the target fd must also be page-aligned. We're not using the offset parameter.

mejedi:
```suggestion
	// Since we use MAP_FIXED, the starting address of the mapping must be
	// page-aligned anyway.
```
The original wording can be interpreted as some architectures being fine with non-page-aligned mmap starting address.

memory.go
ti-mo: > Should the copy decide to go one byte at a time, values can be seen that never were.

Without using atomics, the memory model(s) of the architectures we run on don't make any guarantees about the consistency of values under concurrent access. To my limited understanding, on x64, the smallest unit of memory that can be shared between cores/threads is a cache line, representing 8 bytes. If kernel space is writing and user space is reading, these cache lines can be stale by the time the reader fetches them. Afaik, this is only a real issue if a value spans multiple cache lines, like a float/complex128 or a buffer of arbitrary length. In general, struct alignment rules make sure values stay contained within a single cache line.

I'm not sure what change you're proposing to our use of copy(). I haven't looked at copy's implementation in detail, but ultimately we have very little control over how the compiler chooses to shuffle memory acesses around through several optimization passes, unless copy() contains some kind of memory barrier. (unlikely, need to check) If you're suggesting to batch our calls to copy(), that sounds ineffectual to me.

If you're worried about reading a uint64 that has 4 out of its 8 bytes gone stale, afaik that's not possible as long as you stay away from packed structs. Batching copy() won't save your bacon here either. If you need synchronization, pair `sync/atomic` accesses in Go with `__sync_` in C.

memory.go
ti-mo: Sure, the starting address being page-aligned for MAP_FIXED sounds universal indeed.

memory.go
mejedi: > If you're worried about reading a uint64 that has 4 out of its 8 bytes gone stale, afaik that's not possible as long as you stay away from packed structs.

This is exactly what I'd like to prevent (an **aligned** `uint64` appearing as if it was partially updated when fetched via `ReadAt`).

If copy goes one byte at a time, it could copy the first 4 bytes, get preempted, and get back to copying the rest later. That would be a conformant `copy` implementation, albeit not a very fast one.

> I'm not sure what change you're proposing to our use of copy().

No change proposed :) We might need to implement our own copy if it turns out that builtin misbehaves on some platforms. [amd64](https://github.com/golang/go/blob/master/src/runtime/memmove_amd64.s#L35) seems to be ok.

memory.go
mejedi: >this slice would be a 4k struct instead, it would need to be 4k-aligned to make sure any pointers within the struct would be correctly 8-byte aligned by extension.

I don’t get why it needs to be 4k aligned. 8 byte alignment should be sufficient.

ti-mo: Never mind my contrived example. :) If you request an 8-byte allocation, it needs to be aligned to 8 bytes. If you request a 4k allocation, it needs to be aligned to 4k bytes, etc.

memory.go
ti-mo: The more I think about it, the more I'm considering removing byte-wise reads/writes from Variable altogether. I think `Memory` still needs them to be able to initialize slabs of memory after creating an arena map, since it's not practical to use the syscall map API for it if it works at all.

I hadn't considered the preemption case for some reason. Newer versions of the runtime can preempt after any instruction aside from some 'critical sections', which I haven't found a definition for. Disabling preemption within a function boundary (https://github.com/golang/go/issues/65874) is currently not possible, but even if we manage to prevent runtime preemption, there's still the OS scheduler.

For shared memory accesses to be sound, they need to occur in a single instruction.

> [amd64](https://github.com/golang/go/blob/master/src/runtime/memmove_amd64.s#L35) seems to be ok.

What's your conclusion exactly from reading this memmove implementation?

mejedi: > What's your conclusion exactly from reading this memmove implementation?

Memory is not read a byte at a time unless absolutely necessary. Implementation picks a transfer unit based on the total size. Could be 8 byte at a time (general-purpose registers), could be 16 (SSE).

It doesn't care whether source or destination is aligned. Assuming transfer size of 8 was picked: it will do a 8 byte load at `+0`, another load at `+8`, etc. Values are consistent as long as loads don't cross a cache-line boundary. Of course, there are no ordering guarantees, but we never see partially updated values.

An interesting case is when the total size is NOT a multiple of the transfer size ([`move_9through16`](https://github.com/golang/go/blob/master/src/runtime/memmove_amd64.s#L192)):
```
| 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |
|***|***|***|***|***|***|***|***|   |   | <-- first load
|   |   |***|***|***|***|***|***|***|***| <-- second load
                                |
                                | could be a cache line boundary
| A | A | b | b | b | b | b | b | c | c | cells annotated with "epochs"
```
Some loads are performed at `+8`, etc. relative to block **end**, creating some overlap.

For the purposes of this discourse, let's assume that the start is 8-byte aligned. Then there *might* be a cache line boundary breaking our block into 8 and 2 byte chunks. The second load *might* cross a cache line boundary. I believe it is specified somewhere; CPU fetches 2 cache lines separately and internally combines the data. So after copy completes, the first 2 bytes are from epoch `A`, followed by 6 bytes from epoch `b`, followed by 2 bytes from epoch `c`.

C structure layout rules defines structure alignment as the max alignment between the members. Structure size in turn is a multiple of the structure alignment. Should the total size of the structure be 10, it indicates that the struct alignment was 2 or 1. There were no members larger than `u16`. Hence no member received a partially updated value. (Note: so far we assumed 8 byte aligned block. We should consider +2, +4, and +6 misaligned blocks for the complete proof.)

To summarise, it seems to be ok to use `copy` to capture structures from concurrently modified memory on amd64. There are no relative ordering guarantees, but we never see a partially updated field.

> I hadn't considered the preemption case for some reason. [...] For shared memory accesses to be sound, they need to occur in a single instruction.

Well, we can't guarantee that we obtain a consistent "snapshot" of a variable, but a weaker guarantee of "no partially updated fields" is also useful in practice.

> The more I think about it, the more I'm considering removing byte-wise reads/writes from Variable altogether.

It's reasonable to be somewhat concerned :) That's why I wanted to learn more about `copy` implementation. Note: Linux is facing a similar challenge, when doing `bpf_map_lookup_elem` through the syscall, value is memory-copied w/o obtaining any locks.

I think we should keep reads for read-only variables, the alternative is to expose pointers that will fault on write (AFAIK go can't recover from SIGBUS).

Reads/writes are also nice to have for the symmetry with maps api; objects used as K/V can define their own custom serialisation methods.

ti-mo: Thank you for the analysis, that was really helpful! I have a few more points to add before we can reach consensus:

> but a weaker guarantee of "no partially updated fields" is also useful in practice.

Of course, consistency across fields of a struct can never be guaranteed without some form of synchronization, but if there's a chance for either side to see a partially-written value, that's a problem. Fortunately, struct alignment rules are on our side here. I'll make sure to document that packed structs should be avoided at all costs and summarize the hazards we flagged in this discussion. Now we know how memmove works at a high level, I'm more comfortable leaving Variable.Get() and .Set() in the proposal.

I briefly looked at the arm and arm64 implementations as well, and both of them unsurprisingly try to copy as much data in a single instruction as possible, with the pathological case being unaligned or odd-sized (e.g. copying 3 bytes) as expected.

It looks like arm32 can fill 32 bytes of memory into 8 4-byte registers (`MOVM.DB.W (FROM), [R0-R7]`) in a single instruction. My only doubt here is that this seems to be limited to 32 and 16 bytes, notably missing the 8-byte case for the common u64 and bpf pointers. These seem to be handled using 2 4-byte chunks in a loop, which could be prone to preemption.

I'm not sure if we should let this force our hand, as I'm not even sure how much of bpf actually works on arm32 nowadays. When I tried on an old Pi a few years ago, it simply didn't, but I've heard some people use it.

> the alternative is to expose pointers that will fault on write

I don't think we should allow that, it's too big of a footgun. Someone changes a C global to a const and their application gets terminated. A Go program failing like this is highly surprising, so I'd rather not allow it. For consts, Variable.Get() works fine and is safe, and I'd expect most consts to only be written to from user space; after all the application itself is the source of truth for these.

---

tl;dr - the current proposal is sufficient, modulo documentation around memory alignment? The manual still needs to be written anyway, so I can keep it in mind from the start.

memory.go
ti-mo: I asked on golang-nuts here: https://groups.google.com/g/golang-nuts/c/LzUMguyouXM/m/Gcngw8OuAAAJ. Will update here if any responses come in.

memory.go
mejedi: Please feel free to mark as resolved.

memory.go
ti-mo: tl;dr outcome of the thread:
- as expected, "why would anyone do this?"
- don't depend on any allocs being page-aligned, request an extra page to make sure we can align

The latter makes a lot of sense and is an easy change.

lmb:
I'll make some time next week to review this.

ti-mo:
I've added a small section on Variable and Collection.Variables, should be enough for now. More will be added with the MemoryPointer and VariablePointer follow-up PR.

lmb (changes requested):
I really like how modifying the variable via the getter and setter work, I feel like it integrates with the rest of the API very well.

As you can imagine I'm a little bit hesitant about `Memory`. From reading the discussion on Slack it seems like the core idea behind it is to (at some later point) facilitate atomic access for things like uint64 and to maybe make arena memory usable like Go memory.

The latter is an intriguing idea but I worry that it'll cause us to paint ourselves into a corner. Once we add an API which "transmutes" the off-heap BPF arena memory into a Go pointer "tracked" by the GC we lose all ability to fix things later on. There is pretty much just a single way of implementing this. And while the implementation works right now (on the platforms we test) there is no guarantee that it will continue to do so. Worse, if the package is considered sufficiently important it might even prevent or constrain future improvements on the Go runtime side. We've seen this happen with the recent go:linkname debacle and I think we need to take this into account. Because it's not at all clear what sharing an arena is useful for I think we shouldn't shape our implementation based on this.

This leaves the atomic access to variables. I can't find the old branch, but I think in a previous iteration there were a bunch of wrapper types like `ebpf.Uint64` etc. that abstracted the atomic access away. That added a lot of boilerplate and probably also makes code generation via bpf2go or similar more complicated. Now the plan is (if I understand correctly) to just directly access typed memory via the `atomic` package. I think there might be a third way, by mediating atomic access via `Get` and `Set`.

```
func (v *Variable) Set(value any) error {
	switch value := value.(type) {
	case uint64:
		if v.size != unsafe.Sizeof(value) && v.offset % unsafe.Alignof(value) {
			return errors.New("bad!")
		}
		atomic.StoreUint64(v.addr(), value)
		return nil

	// ...
	}
}
```

This only covers the most basic atomic accesses, but that is probably fine for 90% of the cases? If we ever need more access we could add a function like the following:

```
// VariableMemory allows accessing a variable as a pointer.
//
// It is not valid to access ptr outside of the closure it is passed into.
//
// Returns an error if T is not compatible with the variable.
func VariableMemory[T any](v *Variable, func(ptr *T) error) error
```

Behind the scenes each Variable would refer to a `Memory` (though I'd probably unexport that for now) which is used to unmap the memory with a finalizer.

internal/unix/types_linux_noarm.go
```diff
@@ -0,0 +1,18 @@
...
+package unix
+
+import (
+	"unsafe"
+
+	linux "golang.org/x/sys/unix"
+)
+
+// TODO: Remove this file and wrap linux.MmapPtr in types_linux.go when upgrading to Go 1.23.
+func MmapPtr(fd int, offset int64, addr unsafe.Pointer, length uintptr, prot int, flags int) (ret unsafe.Pointer, err error) {
```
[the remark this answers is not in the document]
ti-mo: Welp, for some reason I thought we needed to wait for a new Go release to use this.. Bumping x/sys to >0.22 does the job.

variable.go
```diff
@@ -111,3 +111,92 @@ func (s *VariableSpec) copy(cpy *CollectionSpec) *VariableSpec {
...
+}
+
+// Size returns the size of the variable.
+func (v *Variable) Size() uint64 {
+	return v.size
+}
+
+// Constant returns true if the Variable represents a variable that is read-only
+// after loading the Collection into the kernel.
+func (v *Variable) Constant() bool {
```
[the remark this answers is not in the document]
ti-mo: The VariableSpec equivalent is already called `Constant()`, but we can still change that as it hasn't been released. The signatures should be consistent. Which do you prefer?

memory.go
```diff
@@ -0,0 +1,280 @@
...
+// Putting the finalizer on the backing array was a conscious decision to avoid
+// having to maintain a reference to the Memory at all times, which is hard to
+// guarantee if the actual access is done through another object, e.g. a
+// sync.Atomic synthesized around a slice of the backing array.
+//
+// In the current implementation, accessing a Memory (or any derived pointers)
+// is guaranteed to access bpf map memory. When all references are gone, the
+// allocation's finalizer will run and unmap the bpf map's memory.
+
+//go:linkname heapObjectsCanMove runtime.heapObjectsCanMove
```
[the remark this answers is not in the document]
ti-mo: This won't land us in the hall of fame. :) `go:linkname` will stop working for importers (a 'pull') in a future release for stdlib symbols that don't have an explicit  `go:linkname` (a 'push'). `heapObjectsCanMove()` is explicitly pushed for callers to import, without the disclaimer that it's not allowed.

Also, I don't really mind being called out for these things. Hyrum's law etc. If importing certain symbols is undesirable, that's pressure that might set things in motion for making them stable exported API, leading to more useful software. For many of these internal symbols, there's no immediate workaround/alternative.

variable.go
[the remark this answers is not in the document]
ti-mo: I've mulled it over a bit, and I think it makes sense to call this `Variable{,Spec}.Constant()`, even if non-constant VariableSpec would turn into a constant Variable at load time. It feels a bit unnatural to say 'a read-only variable'. 'Read-only memory', yes.

Let me know if you feel strongly about this.

variable.go
```diff
@@ -111,3 +111,92 @@ func (s *VariableSpec) copy(cpy *CollectionSpec) *VariableSpec {
...
+func (v *Variable) Get(out any) error {
+	if v.mm == nil {
+		return fmt.Errorf("variable %s: direct access requires Linux 5.5 or later: %w", v.name, ErrNotSupported)
+	}
+
+	if int(v.offset+v.size) > v.mm.Size() {
+		return fmt.Errorf("offset %d(+%d) for variable %s is out of bounds", v.offset, v.size, v.name)
+	}
+
+	b := make([]byte, v.size)
```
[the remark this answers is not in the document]
ti-mo: Sorry, what do you suggest we do instead? We need a byte slice for ReadAt to write into.

memory.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if size == 0 {
+		return nil, errors.New("size must be greater than 0")
+	}
+
+	// Request at least two pages of memory from the runtime to ensure we can
+	// align the requested allocation to a page boundary. This is needed for
+	// MAP_FIXED and makes sure we don't mmap over some other allocation on the Go
+	// heap.
+	size = internal.Align(size+os.Getpagesize(), os.Getpagesize())
```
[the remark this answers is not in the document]
ti-mo: No, this wastes only one page. If the bpf map is one page, we request two pages from the runtime. On a 16k page architecture, this would 'waste' 16k, not 32k. The other 16k would be mapped to kernel memory. If the slice's allocation is page-aligned (which it should be under the current Go allocator), the second page will never be touched and will technically never get dirtied, only 'consuming' virtual memory.

memory.go
```diff
@@ -0,0 +1,280 @@
...
+	// alias the initial pointer if it was already page-aligned.
+	aligned := unsafe.Pointer(internal.Align(uintptr(alloc), uintptr(os.Getpagesize())))
+	runtime.KeepAlive(alloc)
+
+	// Return an aligned pointer into the backing array, losing the original
+	// reference. The runtime.SetFinalizer docs specify that its argument 'must be
+	// a pointer to an object, complit or local var', but this is still somewhat
+	// vague and not enforced by the current implementation.
+	//
+	// Currently, finalizers can be set and triggered from any address within a
```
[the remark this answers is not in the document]
ti-mo: I included this bit since we'll probably switch the whole lib over to `runtime.AddCleanup` as soon as we can use it. I assume the underlying implementation is unlikely to change in the meantime, given the future API will make this explicit.

memory.go
```diff
@@ -0,0 +1,280 @@
...
+	// mmap layout in the process' address space (0xc000000000 is a span of Go
+	// heap), visualized using pmap:
+	//
+	// Address           Kbytes     RSS   Dirty Mode  Mapping
+	// 000000c000000000    1824     864     864 rw--- [ anon ]
+	// 000000c0001c8000       4       4       4 rw-s- [ anon ]
+	// 000000c0001c9000    2268      16      16 rw--- [ anon ]
+	//
+	// This will break up the Go heap, but as long as the runtime doesn't try to
+	// move our allocation around, this is safe for as long as we hold a reference
```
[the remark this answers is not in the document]
ti-mo: You're right, though I haven't been able to observe the runtime using anything other than `anon|private`. Haven't audited the allocator, though, and of course this is not mentioned in the spec anywhere.

Anyway, I (un)fortunately shelved the mapping-over-the-heap mechanism.

memory.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	// Allocate a page-aligned span of memory on the Go heap.
+	alloc, err := allocate(size)
+	if err != nil {
+		return nil, fmt.Errorf("allocating memory: %w", err)
+	}
+
+	flags := unix.PROT_READ | unix.PROT_WRITE
+	if ro {
```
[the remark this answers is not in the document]
ti-mo: Makes sense, done.

memory.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+}
+
+// Size returns the size of the memory-mapped region in bytes.
+func (mm *Memory) Size() int {
+	return len(mm.b)
+}
+
+// Readonly returns true if the memory-mapped region is read-only.
+func (mm *Memory) Readonly() bool {
```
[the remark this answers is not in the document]
ti-mo: I've only found one such spot (LoadPinOptions), but why not. Done.

variable.go
```diff
@@ -111,3 +111,92 @@ func (s *VariableSpec) copy(cpy *CollectionSpec) *VariableSpec {
...
+	if v.Constant() {
+		return fmt.Errorf("variable %s: %w", v.name, ErrReadOnly)
+	}
+
+	buf, err := sysenc.Marshal(in, int(v.size))
+	if err != nil {
+		return fmt.Errorf("marshaling value %s: %w", v.name, err)
+	}
+
+	if int(v.offset+v.size) > v.mm.Size() {
```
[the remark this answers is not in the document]
ti-mo: Good idea. :+1:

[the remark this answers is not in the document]
ti-mo: Regarding WriteAt, it does return the necessary errors for this. The interface spec prescribes EOF needs to be returned if the write goes out of bounds. If off is out of bounds, a generic error is returned.

variable.go
```diff
@@ -111,3 +111,92 @@ func (s *VariableSpec) copy(cpy *CollectionSpec) *VariableSpec {
...
+func (v *Variable) Constant() bool {
+	if v.mm == nil {
+		return true
+	}
+	return v.mm.Readonly()
+}
+
+// Type returns the BTF type of the variable. It contains the [btf.Var] wrapping
+// the underlying variable's type.
+func (v *Variable) Type() btf.Type {
```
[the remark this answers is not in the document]
ti-mo: Sure, makes sense. This is the case on VariableSpec as well, I'll change it there.

collection.go
```diff
@@ -361,19 +374,31 @@ func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (*Co
 		}
 	}
 
+	for varName := range spec.Variables {
+		_, err := loader.loadVariable(varName)
+		if errors.Is(err, ErrNotSupported) {
```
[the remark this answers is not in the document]
ti-mo: Good catch, this was a decision made later on in the cycle. We'll indeed always emit a `coll.Variables["foo"]` for a `spec.Variables["foo"]`.

variable.go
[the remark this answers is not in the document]
ti-mo: Okay. it's actually more surprising for `VariableSpec.Constant()` to return false and its corresponding `Variable.Constant()` to return true after loading. I'll name it `Variable.ReadOnly()` after all.

variable.go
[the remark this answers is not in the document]
ti-mo: Done, thanks for the suggestion!

ti-mo:
I've addressed all feedback, I think I'll merge this so we can move on. It feels like we need a bit more support from the runtime to make native variable pointers a reality, but in the meantime the branch lives here: https://github.com/ti-mo/ebpf/commits/tb/atomic-memory.

I've experimented with the closure suggestion and it's indeed incredibly clunky, though it does move our closer to our goal. It does guarantee the Variable/Memory lifetime for the duration of the variable access, and it's not terribly prone to misuse, even though it can be. It's not a solution I'm comfortable pushing towards our users, though. For now, I think the current proposal is shippable and adds good value, so let's not hold it up any further.

In the meantime, the code Variable/MemoryPointer with heap mmapping lives here: https://github.com/ti-mo/ebpf/commits/tb/atomic-memory. I've created a Go proposal (https://github.com/golang/go/issues/70224) and it's received at least some form of positive traction. Let's see how it plays out.

PR #1574, opened 2024-10-03
link: uprobe multi: add context to EINVAL, don't overflow pid_t in tests

ti-mo:
As documented in the code, Linux commit torvalds/linux@46ba0e49b642 added an extra check on the pid field, which it interprets as a signed pid_t value, returning EINVAL if it's negative.

Take care not to overflow the value in tests so we can keep receiving ESRCH.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,7 +192,8 @@ jobs:
     strategy:
       matrix:
         tag:
-          - "mainline"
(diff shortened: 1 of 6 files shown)


mmat11 (approved):

ti-mo:
@brycekahle @paulcacheux Heads up, kprobing unknown symbols through tracefs returns EINVAL as of 6.11.

PR #1576, opened 2024-10-03
info: expose ksym info and func info by ProgramInfo

Asphaltt:
In [pwru](https://github.com/cilium/pwru), I'm planning to improve the tracing info of bpf prog by mapping ksym addresses to bpf prog info, i.e. the function name of subprograms.

Therefore, it requires more prog info exposed from ProgramInfo, including ksym info and func info.

`bpf_prog_info` supports ksym info since commit torvalds/linux@dbecd7388476 ("bpf: get kernel symbol addresses via syscall"), and supports func info since commit torvalds/linux@838e96904ff3 ("bpf: Introduce bpf_func_info").

### DIFF
--- a/info.go
+++ b/info.go
@@ -531,6 +531,39 @@ func (pi *ProgramInfo) KsymAddrs() ([]uintptr, bool) {
 	return addrs, pi.numKsymInfos > 0
 }
 
+// FuncInfos returns the offset and function information of all (sub)programs in
+// a BPF program.
+//
+// Available from 5.0.
+//
+// Requires CAP_SYS_ADMIN or equivalent for reading BTF information. Returns
+// ErrNotSupported if the program was created without BTF or if the kernel
+// doesn't support the field.
+func (pi *ProgramInfo) FuncInfos() (btf.FuncOffsets, error) {
+	id, ok := pi.BTFID()
+	if pi.numFuncInfos == 0 || !ok {
+		return nil, fmt.Errorf("program created without BTF or unsupported kernel: %w", ErrNotSupported)
+	}
+
+	h, err := btf.NewHandleFromID(id)
+	if err != nil {
+		return nil, fmt.Errorf("get BTF handle: %w", err)
+	}
+	defer h.Close()
+
+	spec, err := h.Spec(nil)
+	if err != nil {
+		return nil, fmt.Errorf("get BTF spec: %w", err)
+	}
+
+	return btf.LoadFuncInfos(
+		bytes.NewReader(pi.funcInfos),
+		internal.NativeEndian,
+		pi.numFuncInfos,
+		spec,
+	)
+}
+
 func scanFdInfo(fd *sys.FD, fields map[string]interface{}) error {
 	fh, err := os.Open(fmt.Sprintf("/proc/self/fdinfo/%d", fd.Int()))
 	if err != nil {
--- a/info.go
+++ b/info.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"syscall"
 	"time"
-	"unsafe"
 
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/btf"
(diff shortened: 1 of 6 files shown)


info.go
```diff
@@ -508,6 +518,34 @@ func (pi *ProgramInfo) VerifiedInstructions() (uint32, bool) {
 	return pi.verifiedInstructions, pi.verifiedInstructions > 0
 }
 
+// KsymAddrs returns the ksym addresses of bpf programs, including subprograms,
+// of current bpf program. The addresses are same as the ones in /proc/kallsyms.
+//
+// Available from 4.18.
+//
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) KsymAddrs() ([]uint64, bool) {
```
ti-mo, line 527: This should return `[]uintptr`. (makes it even easier to feed into MultiKprobe!)

info.go
```diff
@@ -508,6 +518,34 @@ func (pi *ProgramInfo) VerifiedInstructions() (uint32, bool) {
...
+	return pi.ksymInfos, pi.numKsymInfos > 0
+}
+
+type ProgramFunctionInfo = sys.FuncInfo
```
ti-mo, line 531: `sys` is internal and autogenerated, we don't expose this as it's subject to change outside of our control.

info.go
```diff
@@ -508,6 +518,34 @@ func (pi *ProgramInfo) VerifiedInstructions() (uint32, bool) {
...
+
+// FuncInfos returns the function information of bpf programs, including
+// subprograms, of current bpf program.
+//
+// Available from 5.0.
+//
+// The bool return value indicates whether this optional field is available.
+func (pi *ProgramInfo) FuncInfos() ([]ProgramFunctionInfo, bool) {
```
ti-mo, line 539: I'm curious about the specific use case for this function, since this information is already attached as asm.Instruction metadata to the output of `ProgramInfo.Instructions()`.

This function should at least return something like `[]struct{off uint64, fi btf.FuncInfo}`, otherwise the caller needs to do the work of getting the btf handle and pulling out the right IDs, as demonstrated by the test you added.

ti-mo (changes requested):
Thanks for the patch! Left a few comments.

ti-mo:
(By the way, a common format for referring to a commit in Linux is `dbecd7388476 ("bpf: get kernel symbol addresses via syscall")` or `torvalds/linux@dbecd7388476 ("bpf: get kernel symbol addresses via syscall")` to make it clickable on GH.

info.go
Asphaltt: `[]struct{off uint64, fi btf.FuncInfo}` seems better.

So, `ProgramFunctionInfo` will be:

```Go
type ProgramFunctionInfo struct {
	Off		uint64
	FuncInfo	*btf.Func
}
```

ti-mo: Sorry, yes, `btf.FuncInfo` doesn't exist, I meant `btf.Func`. Let's do:
```go
type ProgramFunctionInfo struct {
	InstructionOffset	uint64
	FuncInfo		*btf.Func
}
```

Asphaltt: > the specific use case for this function

When to run `pwru --output-stack`, it's really helpful to show the exact bpf prog info instead of `bpf_prog_f9e178771bf9c08b_subprog3   [bpf]` for its ksym address, especially for the case there are multiple `bpf_prog_f9e178771bf9c08b_subprog3   [bpf]` for different ksym addresses.

As for my plan, it's to show `subprog3:entry2[type:SCHED_CLS][ID:290][bpf]` for subprog's ksym addresses, `entry2[type:SCHED_CLS][ID:290][bpf]` for non-subprog's ksym addresses. ID is required here because a tc-bpf code can run for multiple pods in k8s env, whose name, tag, type are same but ID.

dylandreimerink (approved):
These changes all make sense to me

ti-mo (approved):

PR #1578, opened 2024-10-04
feat: add support for untyped extern __ksym variables

patrickpichler:
It is now possible to use the `__ksym` attribute to load the address of the ksym for every use of the variable. It works by searching the name of any external ksym variable in the `/proc/kallsyms` file and patching any load operation on that external variable to load the address as an immediate value.


elf_reader.go
```diff
@@ -647,6 +659,15 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
...
 			ins.Constant = 0
 
+		case ksymsVar != nil && ins.OpCode.IsDWordLoad():
+			if bind != elf.STB_GLOBAL {
+				return fmt.Errorf("asm relocation: %s: %w: %s", name, errUnsupportedBinding, bind)
+			}
+			println("==> set")
+			ins.Metadata.Set(ksymMetaKey{}, &ksymMeta{
+				rel.Name,
+			})
```
dylandreimerink, line 669: `__ksym`s can also be `__weak` meaning that we should not error when we can't find them. Can we imediatly add support for this? Please look at the `case`s above this one to see how we handled it for kfuncs.

linker.go
```diff
@@ -457,3 +459,41 @@ func resolveKconfigReferences(insns asm.Instructions) (_ *Map, err error) {
...
+				return fmt.Errorf("error whild trying to load kallsyms: %v", err)
+			}
+		}
+
+		addr, found := ksymMap[meta.name]
+		if !found {
+			if !slices.Contains(missingKsyms, meta.name) {
+				missingKsyms = append(missingKsyms, meta.name)
+			}
+		}
```
dylandreimerink, line 488: Once we have binding info, please make a case here for weak ksyms. If we can't find them, set its value to `0`.

This allows users to write CO-RE code using `__ksym`s
```
extern void some_symbol __ksym __weak;

if (&some_symbol) {
  // we have a kernel with this symbol, use it.
} else {
  // we are running on a kernel without this symbol, execute fallback code.
}
```

internal/kallsyms/kallsyms.go
```diff
@@ -72,3 +78,55 @@ func loadKernelModuleMapping(f io.Reader) (map[string]string, error) {
...
+
+	ksyms, err = loadSymbolAddresses(f)
+	if err != nil {
+		return nil, err
+	}
+
+	return ksyms, nil
+}
+
+func loadSymbolAddresses(f io.Reader) (map[string]uint64, error) {
```
dylandreimerink, line 106: I am not yet 100% convinced this is the way to go. This essentially loads all symbols into memory, persistently. A quick dump on my local machine indicates `/proc/kallsyms` is about 17MiB (it will vary depending on the amount of modules loaded),  we are removing a bit due to having the address as uint64 and adding a bit due to map overhead. So it likely isn't super far off.

My thinking is, `/proc/kallsyms` is a virtual file, so reading it is likely faster than reading an actual file. Memory allocation is slow, especially maps and especially when you can't provide a capacity. Might it not be better to search the `/proc/kallsyms` instead of pre-indexing everything. Both time wise and memory wise.

I would at least like to do a bit of research before committing to one or the other approach.

testdata/ksym.c
dylandreimerink, line 1: Could you also commit the pre-generated .elf files, we tend to commit these so running tests is quicker and easier.

dylandreimerink (changes requested):
Thank you for the work! I have a few things to address though.

internal/kallsyms/kallsyms.go
patrickpichler: That is a good point. I had another look how libbpf is handling this and they are patching the address in the symbols directly. I guess a better approach would be to simply collect all used symbols into a map, iterate over the data of `/proc/kallsym` and then patch the addresses from there. What do you think?

dylandreimerink: I did some quick testing.

Added benchmarks for the map indexing and searching approach:
```
func BenchmarkLoadSymbolAddresses(b *testing.B) {
	for i := 0; i < b.N; i++ {
		f, err := os.Open("/proc/kallsyms")
		if err != nil {
			b.Fatal(err)
		}

		_, err = loadSymbolAddresses(f)
		if err != nil {
			b.Fatal(err)
		}
	}
}

func BenchmarkFindSymbolAddresses(b *testing.B) {
	for i := 0; i < b.N; i++ {
		f, err := os.Open("/proc/kallsyms")
		if err != nil {
			b.Fatal(err)
		}

		_, err = findSymbolAddresses(f, []string{
			"fixed_percpu_data",
			"__per_cpu_start",
			"cpu_debug_store",
			"irq_stack_backing_store",
			"cpu_tss_rw",
			"gdt_page",
			"exception_stacks",
			"entry_stack_storage",
			"espfix_waddr",
			"espfix_stack",
		})
		if err != nil {
			b.Fatal(err)
		}
	}
}

func TestFindSymbolAddresses(t *testing.T) {
	f, err := os.Open("/proc/kallsyms")
	if err != nil {
		t.Fatal(err)
	}

	_, err = findSymbolAddresses(f, []string{
		"fixed_percpu_data",
		"__per_cpu_start",
		"cpu_debug_store",
		"irq_stack_backing_store",
		"cpu_tss_rw",
		"gdt_page",
		"exception_stacks",
		"entry_stack_storage",
		"espfix_waddr",
		"espfix_stack",
	})
	if err != nil {
		t.Fatal(err)
	}

	t.Fail()
}
```

Implemented a rough version which minimizes allocations:
```
func findSymbolAddresses(f io.Reader, syms []string) ([]uint64, error) {
	r := bufio.NewReader(f)
	result := make([]uint64, len(syms))

	const bufSize = 1024 * 16
	buf := make([]byte, bufSize)
	var err error
	i := 0
	n := 0
	for {
		newLine := bytes.IndexByte(buf[i:n], '\n')
		if newLine == -1 {
			rem := buf[i:n]
			copy(buf, rem)

			n, err = r.Read(buf[len(rem):])
			if err != nil {
				if err == io.EOF {
					return result, nil
				}
				return nil, err
			}
			n = n + len(rem)
			i = 0

			continue
		}

		line := buf[i : i+newLine]

		sIdx := bytes.IndexByte(line, ' ')
		if sIdx == -1 {
			continue
		}
		addrBytes := line[:sIdx]
		line = line[sIdx+1:]

		sIdx = bytes.IndexByte(line, ' ')
		if sIdx == -1 {
			continue
		}
		sIdx = bytes.IndexByte(line, ' ')
		symBytes := line
		if sIdx != -1 {
			symBytes = line[:sIdx]
		}

		for i, sym := range syms {
			if string(symBytes) == sym {
				addr, err := strconv.ParseUint(string(addrBytes), 16, 64)
				if err != nil {
					return nil, err
				}
				result[i] = addr
			}
		}

		i += newLine + 1
	}
}
```

Results on my machine are:
```
go test -benchmem -run=^$ -bench=".*" ./internal/kallsyms 
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf/internal/kallsyms
BenchmarkLoadSymbolAddresses-20                2         631965570 ns/op        101882808 B/op   2936085 allocs/op
BenchmarkFindSymbolAddresses-20                8         126794984 ns/op           20682 B/op          6 allocs/op
PASS
ok      github.com/cilium/ebpf/internal/kallsyms        3.067s
```

The benchmark searches 10 symbols, but varying this does not seem to change much performance wise. A pprof shows 99% in the read syscall. So this is near the best we can do by re-parsing every time.

So there is an interesting tradeoff. Pre-indexing starts to become faster after 5 calls (regardless of the amount of symbols). At 4 calls CPU time is even, but then re-parsing still wins on memory usage. Makes me wonder if it would be possible to collect the symbols for all programs in a collection and fixup all programs at once?

patrickpichler: Thanks for sharing those result! 

I've updated the code and now parse `/proc/kallsyms` once during loading of the collection. I didn't update the parser function for the individual lines yet, but would love to hear what you think about the updated solution.

elf_reader.go
patrickpichler: I've added support for `__weak` as well 👍

internal/kallsyms/kallsyms.go
dylandreimerink: Thank you! This approach makes more sense to me. We are still a bit slow in `loadSymbolAddresses`, that is simply because the bufio scanner, and scanf do a lot of allocation internally. Lets deal with that some other time (another PR). 

However, even with improved performance walking the kallsyms is slow read wise. Can we make calling `LoadSymbolAddresses` conditional, so we only take the hit if we have at least 1 ksym to resolve?

patrickpichler: I will add the check 👌 We can also return as soon as we have found all symbols, no need to iterate over potentially the full file.

dylandreimerink: > We can also return as soon as we have found all symbols, no need to iterate over potentially the full file.

Yea, also a very good idea. :+1:

patrickpichler: I've implemented the optimization and cleanup the commit history. I will now also turn this back to a PR from the draft 👍 

Thanks for the nice pointers 👌

dylandreimerink (approved):
Thanks again for the work!

internal/kallsyms/kallsyms.go
```diff
@@ -72,3 +74,56 @@ func loadKernelModuleMapping(f io.Reader) (map[string]string, error) {
...
+	ksyms, err := loadSymbolAddresses(f, symbols)
+	if err != nil {
+		return nil, err
+	}
+
+	return ksyms, nil
+}
+
+func loadSymbolAddresses(f io.Reader, symbols []string) (map[string]uint64, error) {
+	scan := bufio.NewScanner(f)
```
ti-mo, line 97: Nit: declare this closer to where it's used, e.g. 
```go
s := bufio.NewScanner(f)
for s.Scan() {
...
```

internal/kallsyms/kallsyms.go
```diff
@@ -72,3 +74,56 @@ func loadKernelModuleMapping(f io.Reader) (map[string]string, error) {
...
+
+	ksyms, err := loadSymbolAddresses(f, symbols)
+	if err != nil {
+		return nil, err
+	}
+
+	return ksyms, nil
+}
+
+func loadSymbolAddresses(f io.Reader, symbols []string) (map[string]uint64, error) {
```
ti-mo, line 96: Is this faster than using a map? e.g.
`func loadSymbolAddresses(f io.Reader, syms map[string]uintptr) error`, looking up every symbol read from kallsyms against syms and setting the value if it's present.

prog.go
```diff
@@ -337,6 +337,11 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 	}
 	defer kconfig.Close()
 
+	err = resolveKsymReferences(insns)
```
ti-mo, line 340: Nit:
```suggestion
	if err := resolveKsymReferences(insns); err != nil {
```

linker.go
```diff
@@ -457,3 +458,33 @@ func resolveKconfigReferences(insns asm.Instructions) (_ *Map, err error) {
 
 	return kconfig, nil
 }
+
+func resolveKsymReferences(insns asm.Instructions) error {
```
ti-mo, line 462: I had a few comments on this function, so I played around with it a bit and made it closer to the style of the code base:
```go

func resolveKsymReferences(insns asm.Instructions) error {
	var missing []string

	iter := insns.Iterate()
	for iter.Next() {
		ins := iter.Ins
		meta, _ := ins.Metadata.Get(ksymMetaKey{}).(*ksymMeta)
		if meta == nil {
			continue
		}

		v := meta.VarSpec
		if v.Addr != 0 {
			ins.Constant = int64(v.Addr)
			continue
		}

		if meta.Binding == elf.STB_WEAK {
			ins.Constant = 0
			continue
		}

		name := v.Var.TypeName()
		if !slices.Contains(missing, name) {
			missing = append(missing, name)
		}
	}

	if len(missing) > 0 {
		return fmt.Errorf("missing ksyms: %s", strings.Join(missing, ","))
	}

	return nil
}
```

- shorter lines and shorter var names
- moving things closer to where they're used
- don't check for tautological conditions, e.g. (a == 0 followed by a != 0)
- inverted slices.Contains, it would append all occurrences otherwise

internal/kallsyms/kallsyms.go
```diff
@@ -72,3 +74,56 @@ func loadKernelModuleMapping(f io.Reader) (map[string]string, error) {
 	}
 	return mods, nil
 }
+
+func LoadSymbolAddresses(symbols []string) (map[string]uint64, error) {
```
ti-mo, line 78: I don't think this function should return all symbols, as currently we'll happily take an address of any kernel symbol (like a function) and stuff it into an instruction. I hope the verifier rejects that, but we should filter kallsyms output to catch this earlier.

I can't find a reference for the kallsyms format, but this seems to be a variable: `00000000000000E0 V hv_root_partition`. Perhaps `LoadSymbolAddresses` and `loadSymbolAddresses` should take a string to match the type against and filter the output if the string is not `""`.

testdata/ksym.c
```diff
@@ -0,0 +1,39 @@
...
+char __license[] __section("license") = "Dual MIT/GPL";
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 2);
+	__type(key, uint32_t);
+	__type(value, uint64_t);
+} array_map __section(".maps");
+
+static long (*bpf_map_update_elem)(void *map, const void *key, const void *value, uint64_t flags) = (void *)2;
```
ti-mo, line 12: This is already in `common.h` as `map_update_elem`. I'll send a quick patch to remove this from the other testdata files that do the same so the copypasta stops.

ti-mo (changes requested):
Thanks for working on this! This is not a trivial contribution, I commend you for taking on the challenge. :slightly_smiling_face: Most of my comments are around style and readability.

We also need to make sure we reject non-variable references, unless there's a bpf use case for that, of course. Thanks again!

internal/kallsyms/kallsyms.go
dylandreimerink: > Is this faster than using a map?

Technically, using a map is faster, especially for large amounts of symbols. But if you go an pprof this, then 99% of time is spent allocating in the scanner.Text and the read syscall. So we can essentially go whatever way we find easier to use, doesn't make a practical difference.

internal/kallsyms/kallsyms.go
patrickpichler: I have to say, here I mostly went with what libbpf is doing ([source in kernel 5.15](https://github.com/torvalds/linux/blob/v5.15/tools/lib/bpf/libbpf.c#L6586)). 

Unless I overlooked something, they also do not check the `sym_type` at all. 

Is there any problem you think will happen if loading of any symbol is allowed?

internal/kallsyms/kallsyms.go
patrickpichler: I did some quick benchmarks and on my machine the slice is faster (input of 19 elements, where one of the requested symbols is the last in `/proc/kallsyms`):
```
BenchmarkLoadSymbolAddresses/map_output-4                      7         162487483 ns/op
BenchmarkLoadSymbolAddresses/in_out-4                          7         165968776 ns/op
```

Here is the benchmark code:
This is the new function
```go
func loadSymbolAddressesInOut(f io.Reader, symbols map[string]uint64) error {
	result := make(map[string]uint64)

	scan := bufio.NewScanner(f)
	for scan.Scan() {
		var (
			addr   uint64
			t      rune
			symbol string
		)

		line := scan.Text()

		_, err := fmt.Sscanf(line, "%x %c %s", &addr, &t, &symbol)
		if err != nil {
			return err
		}
		if _, found := symbols[symbol]; found {
			result[symbol] = addr
		}
		// We got all the symbols we need, no need to process any more lines.
		if len(result) == len(symbols) {
			break
		}
	}

	if scan.Err() != nil {
		return scan.Err()
	}

	return nil
}
```
This are the benchmarks:
```go
var globalMap map[string]uint64

func BenchmarkLoadSymbolAddresses(b *testing.B) {
	kallsyms, err := os.ReadFile("/proc/kallsyms")
	if err != nil {
		b.Fatal(err)
	}

	b.ResetTimer()

	b.Run("map output", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			krdr := bytes.NewBuffer(kallsyms)
			result, err := loadSymbolAddresses(krdr, []string{
				"bpf_fd_inode_storage_update_elem",
				"kprobe_insn_page",
				"el1t_64_sync",
				"el1t_64_irq",
				"el1t_64_fiq",
				"el1t_64_error",
				"el1h_64_sync",
				"el1h_64_irq",
				"el1h_64_fiq",
				"el1h_64_error",
				"el0t_64_sync",
				"el0t_64_irq",
				"el0t_64_fiq",
				"el0t_64_error",
				"el0t_32_sync",
				"el0t_32_irq",
				"el0t_32_fiq",
				"el0t_32_error",
			})
			if err != nil {
				b.Fatal(err)
			}
			globalMap = result
		}
	})

	b.Run("in out", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			krdr := bytes.NewBuffer(kallsyms)
			result := map[string]uint64{
				"bpf_fd_inode_storage_update_elem": 0,
				"kprobe_insn_page":                 0,
				"el1t_64_sync":                     0,
				"el1t_64_irq":                      0,
				"el1t_64_fiq":                      0,
				"el1t_64_error":                    0,
				"el1h_64_sync":                     0,
				"el1h_64_irq":                      0,
				"el1h_64_fiq":                      0,
				"el1h_64_error":                    0,
				"el0t_64_sync":                     0,
				"el0t_64_irq":                      0,
				"el0t_64_fiq":                      0,
				"el0t_64_error":                    0,
				"el0t_32_sync":                     0,
				"el0t_32_irq":                      0,
				"el0t_32_fiq":                      0,
				"el0t_32_error":                    0,
			}
			err := loadSymbolAddressesInOut(krdr, result)
			if err != nil {
				b.Fatal(err)
			}
			globalMap = result
		}
	})

}
```

elf_reader.go
```diff
@@ -44,6 +57,7 @@ type elfCode struct {
 	maps     map[string]*MapSpec
 	vars     map[string]*VariableSpec
 	kfuncs   map[string]*btf.Func
+	ksymVars map[string]*ksymVarSpec
```
ti-mo, line 60: We can just call these 'ksyms' if they can represent any kind of kernel symbol (vars, functions, ..)

```suggestion
	ksyms map[string]uint64 // see comment above
```

elf_reader.go
```diff
@@ -32,6 +33,18 @@ type kfuncMeta struct {
...
 }
 
+type ksymMetaKey struct{}
+
+type ksymMeta struct {
+	Binding elf.SymBind
+	Var     *ksymVarSpec
+}
+
+type ksymVarSpec struct {
```
ti-mo, line 43: I'd do away with this type altogether, I'm not sure what value it adds. elfCode only needs to memoize the symbol name and offset.

```go
type ksymMeta struct {
	Binding elf.SymBind
	Addr    uint64
	Name    string
}

type elfCode struct {
	...
	kfuncs   map[string]*btf.Func
	ksyms    map[string]uint64
	kconfig  *MapSpec
}
```

In loadKsymsSection():
```go
       var ksyms []string
        for _, v := range ds.Vars {
                switch t := v.Type.(type) {
...
                case *btf.Var:
                       ec.ksymVars[t.TypeName()] = &ksymVarSpec{
                               Var: t,
                       }
                       ksyms = append(ksyms, t.TypeName())
...

       ksymAddrs, err := kallsyms.LoadSymbolAddresses(ksyms)
...
```

internal/kallsyms/kallsyms.go
```diff
@@ -52,6 +55,8 @@ func FlushKernelModuleCache() {
 	kernelModules.kmods = nil
 }
 
+var ErrKsymIsUnambiguous = errors.New("ksym is unambiguous")
```
ti-mo, line 58: I think this should read `ksym is ambiguous`.

This sentinel is only checked from within the package itself, there's no need for this to be exported. An external caller wouldn't need to react to it at runtime, for example. Maybe they could, but it would involve trying to load a different ELF that doesn't refer to the symbol. If this ever becomes relevant, it can be exported, but unexporting a symbol users depend on is more difficult to justify.

internal/kallsyms/kallsyms_test.go
```diff
@@ -7,18 +7,21 @@ import (
...
+00000000000000B0 B empty_zero_page
+00000000000000C0 D kimage_vaddr
+00000000000000D0 R __start_pci_fixups_early
+00000000000000E0 V hv_root_partition
+00000000000000F0 W calibrate_delay_is_known
+A0000000000000AA a nft_counter_seq	[nft_counter]
+A0000000000000BA b bootconfig_found
+A0000000000000CA d __func__.10
+A0000000000000DA r __ksymtab_LZ4_decompress_fast
+A0000000000000EA t writenote`)
```
ti-mo, line 22: Nit: Is it typical for duplicate symbols to have the same address? Typically, they come from separate compilation units, so their addresses shoud differ. Not important for the sake of the test, but maybe we could tolerate duplicates as long as their addresses match.

ti-mo (changes requested):
Thanks! Just gave this a second review. Feel free to squash any amends into the first commit, I don't usually look at fixup commits separately. Re-request review when ready.

linker.go
```diff
@@ -457,3 +458,38 @@ func resolveKconfigReferences(insns asm.Instructions) (_ *Map, err error) {
...
+			continue
+		}
+
+		v := meta.Var
+		if v.Addr != 0 {
+			ins.Constant = int64(v.Addr)
+			continue
+		}
+
+		if meta.Binding == elf.STB_WEAK {
```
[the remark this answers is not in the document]
patrickpichler: AFAIK since this anyway is a simple constant load operation, we can just set te constant to `0`. I guess `0` should be fine, as this is also what libbpf seems to be doing ([link](https://github.com/torvalds/linux/blob/v5.15/tools/lib/bpf/libbpf.c#L6857)).

I will add a comment though, haha.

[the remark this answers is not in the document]
ti-mo: Agreed that it's good for clarity, thanks for the link!

ti-mo:
@patrickpichler Thank you for addressing all of my feedback so far. Something came up in https://github.com/cilium/ebpf/pull/1584 that made me realize using fmt scanning is really quite slow for something that potentially happens on every ELF load. The result of the parsing operation also isn't cached.

Since we have some work to do on providing an efficient kallsyms parser, is it okay if we change this to a draft for now? The surrounding code is fine and doesn't need to change anymore, it just needs to use a common parser implementation with the kallsyms kmod logic.

patrickpichler:
@ti-mo after thinking some more about it, would it maybe be possible to still merge it? 😅 
It is an opt-in change, meaning there will be no performance regression of existing users and I would kinda need the functionality for another project 😅 

It would also make things easier to work on BTF support (otherwise I would need to layer PRs) 😅

If not, it is also no problem though 😄

ti-mo:
> @ti-mo after thinking some more about it, would it maybe be possible to still merge it? 😅 It is an opt-in change, meaning there will be no performance regression of existing users and I would kinda need the functionality for another project 😅

Agreed, no harm in merging as-is. Will make it somewhat easier to introduce the new reader and making sure it works for both use cases.

I've pushed a small fixup and will merge when tests are green. Thanks again for picking this up!

ti-mo (approved):

dylandreimerink:
> Something came up in https://github.com/cilium/ebpf/pull/1584 that made me realize using fmt scanning is really quite slow ...

This is true, the scanner allocates a lot, which makes it slow. I mentioned this in https://github.com/cilium/ebpf/pull/1578#discussion_r1787820015 as well. You can 4x it with a more bespoke implementation, after which you are bottleneck-ed by the read syscall. But you only hit that path if you use ksyms.

> ... for something that potentially happens hundreds of thousands of times on every ELF load. 

We specifically change the logic so we collect all symbols for a whole collection and do this op once per ELF/Collection load.

> The result of the parsing operation also isn't cached, which isn't ideal.

We also discussed the advantages and disadvantages of caching the parsed kallsyms in https://github.com/cilium/ebpf/pull/1578#discussion_r1787820015. From my benchmarks, its not worth it. Unless you plan on loading more than 5 collections and can spare the RSS.

ti-mo:
> We also discussed the advantages and disadvantages of caching the parsed kallsyms in [#1578 (comment)](https://github.com/cilium/ebpf/pull/1578#discussion_r1787820015). From my benchmarks, its not worth it. Unless you plan on loading more than 5 collections and can spare the RSS.

Yep, just read through some of the comments earlier. Shame I didn't see them before, I did the same implementation exercise for the other PR.

What _could_ be a real improvement is having *both* a kallsyms parser that allocates minimally *and* a pull-through cache that only caches the symbols that were requested by the user.

I'd wager a guess that 99% of ELF loads over the lifetime of an application are the same objects over and over (looking at Cilium especially) for resources like containers that come and go. If parsing becomes cheap, we can cache less aggressively, and we won't need to worry about ballooning memory consumption.

With this PR, we can technically refer to any symbol (not only `tT`), so with the current shotgun approach we'd need to cache everything, which is not realistic. I'll try and come up with a small parser that can be reused between kmods and ksyms.

dylandreimerink:
> What could be a real improvement is having both a kallsyms parser that allocates minimally and a pull-through cache that only caches the symbols that were requested by the user.
>
> I'd wager a guess that 99% of ELF loads over the lifetime of an application are the same objects over and over (looking at Cilium especially) for resources like containers that come and go. The set of symbols is going to be fairly static across all of those loads, since in the ksyms case these are baked into the variable definitions, and most people would be running something like bpf2go/prebuilt ELFs/CORE. If parsing becomes cheap, we can cache less aggressively, and we won't need to worry about ballooning memory consumption.

Yea, I would also make that wager, good point.

PR #1582, opened 2024-10-10
btf: Handle Tags when making copies

mejedi:
d4b2f8cc introduced Tags on different types implementing btf.Type. Tags arrays are initialized at load time from declTag-s referring to the respective type.

Type-s are explicitly copied with btf.Copy, also implicitly when copying a btf.Spec. Semantically, it is a deep copy, hence Tags must be cloned as well.


dylandreimerink (approved):
Oh, seems like I forgot something :facepalm:. Nice catch, thank you.

PR #1584, opened 2024-10-14
kallsyms: optimise parsing

alban:
In tests with pprof in Inspektor Gadget, I see allocations getting reduced from 34MB to 12MB.

I also added a standalone benchmark.

tl;dr: 12% faster, 42% less allocations in bytes, 48% less allocations

```
$ go test -run=BenchmarkKernelModuleMemory -bench=. ./internal/kallsyms/... -count 10 | tee main.bench
$ go test -run=BenchmarkKernelModuleMemory -bench=. ./internal/kallsyms/... -count 10 | tee patched.bench
$ benchstat main.bench patched.bench
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf/internal/kallsyms
cpu: Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
                     │ main.bench  │            patched.bench            │
KernelModuleMemory-4   299.9m ± 3%   262.1m ± 6%  -12.59% (p=0.000 n=10)

                     │  main.bench  │            patched.bench             │
KernelModuleMemory-4   37.70Mi ± 0%   21.67Mi ± 0%  -42.53% (p=0.000 n=10)

                     │ main.bench  │            patched.bench            │
KernelModuleMemory-4   432.1k ± 0%   221.9k ± 0%  -48.64% (p=0.000 n=10)
```

xref https://github.com/inspektor-gadget/inspektor-gadget/pull/3571#issuecomment-2410532116

cc @burak-ok

### DIFF
--- a/internal/kallsyms/kallsyms.go
+++ b/internal/kallsyms/kallsyms.go
@@ -60,7 +60,12 @@ func loadKernelModuleMapping(f io.Reader) (map[string]string, error) {
 	mods := make(map[string]string)
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
-		fields := bytes.Fields(scanner.Bytes())
+		line := scanner.Bytes()
+		// bytes.Fields is expensive, so filter for '[' first
+		if !bytes.ContainsRune(line, '[') {
+			continue
+		}
+		fields := bytes.Fields(line)
 		if len(fields) < 4 {
 			continue
 		}
--- a/internal/kallsyms/kallsyms_test.go
+++ b/internal/kallsyms/kallsyms_test.go
@@ -2,6 +2,7 @@ package kallsyms
 
 import (
 	"bytes"
+	"runtime"
 	"testing"
 
 	"github.com/go-quicktest/qt"
@@ -61,4 +62,16 @@ func TestLoadSymbolAddresses(t *testing.T) {
 	}
 	err := loadSymbolAddresses(b, ksyms)
 	qt.Assert(t, qt.ErrorIs(err, errKsymIsAmbiguous))
+}
+
+func BenchmarkKernelModuleMemory(b *testing.B) {
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		FlushKernelModuleCache()
+		runtime.GC()
+
+		b.StartTimer()
+		KernelModule("foo")
+	}
 }


florianl (approved):
great optimization! :+1:

internal/kallsyms/kallsyms.go
```diff
@@ -56,7 +56,12 @@ func loadKernelModuleMapping(f io.Reader) (map[string]string, error) {
 	mods := make(map[string]string)
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
-		fields := bytes.Fields(scanner.Bytes())
+		line := scanner.Bytes()
+		// bytes.Fields is expensive, so filter for '[' first
+		if !bytes.ContainsRune(line, '[') {
```
burak-ok, line 61: I am not that sure (since I don't know the format that well), but if the closing bracket `]` is always at the end of the line we could utilize that to improve performance even further

burak-ok (approved):
Thank you Alban for making it less memory hungry and faster :)

internal/kallsyms/kallsyms.go
alban: Do you mean replacing `bytes.ContainsRune` by something like this?
```go
if strings.HasSuffix(line, "]")
```

That should filter the same amount of lines as `bytes.ContainsRune`, so it's a matter of if `strings.HasSuffix` is faster than `bytes.ContainsRune`. I have not tested it but I suspect it is similar for short lines like the ones in kallsyms, and in any cases would not change the amount of memory allocations.

The printfs for /proc/kallsyms are implemented here:
https://github.com/torvalds/linux/blob/eca631b8fe808748d7585059c4307005ca5c5820/kernel/kallsyms.c#L446-L465

The last change was 3 years ago to add the buildid inside the brackets:
https://github.com/torvalds/linux/commit/9294523e3768030ae8afb84110bcecc66425a647#diff-9538b26d3e082f233e6adac664cd2c14cbf2d510d5d7f286eef329c58de87eadR451
This is only for kernels with `CONFIG_STACKTRACE_BUILD_ID` (which is not set for my kernel). But the way cilium/ebpf parses the line is fine with regards to that change.

So I would say the format seems more or less stable but can change. And testing "contains-[" should be more stable than testing "suffix-]", in case Linux adds new fields at the end.

burak-ok:
> so it's a matter of if strings.HasSuffix is faster than bytes.ContainsRune

That is what I meant. Instead of looping through every rune in the string it only needs to check the last one.

> And testing "contains-[" should be more stable than testing "suffix-]", in case Linux adds new fields at the end.

That is what I had feared

ti-mo:
This is a nice band-aid, but conflicts with what https://github.com/cilium/ebpf/pull/1578 is trying to do. Reading this PR here made me realize the proposal in #1578 is even slower and more expensive, that will need to be improved, and we shouldn't have multiple ways of scanning kallsyms.

As an experiment, I wrote a naive barebones line parser for kallsyms that calls `bytes.IndexAny(line, " ")` repeatedly and re-slices the line:
```
func parseLine(line []byte) (addr uint64, name string, mod string, error)
```

I got your results down to this, basically cutting down another 3/4 of the size and number of allocations:
```
BenchmarkKernelModuleMemory-16    	       5	 220576087 ns/op	13158904 B/op	  148693 allocs/op
```

Note: this similarly skips lines containing `[` and looks for `tT` symbol types only. 

Perhaps we can write a simple kallsyms iterator on top of `bufio.ReadLine` (not using `bufio.Scanner()` since it allocates a ton). I'm not sure if we need to deal with utf-8 for splitting on space and \t. Having one iterator that can jump to the next line or word boundary would be ideal.

```
func (s *Scanner) NextLine() bool
func (s *Scanner) NextWord() bool

// These only return the current word, never a full line.
func (s *Scanner) WordBytes() []byte
func (s *Scanner) WordText() string
```

This avoids having to allocate an individual `bufio.ScanWords` scanner around a `bytes.Reader` for every line, since we have many short lines we need to parse into a `(uint64, string)` or `(uint64, string, string)` if it involved a kmod. If we need to filter out certain symbol types, we can call `s.NextLine()` if we're not interested in the remainder of the line. What we need to avoid is allocating slices and letting them escape to the heap, which is something `bytes.Fields` does in spades. Returning `[][]byte` should always result in 4 or 5 slice headers and a backing array for the outer slice being put on the heap.

alban:
I rebased my PR to fix the conflict with #1578.

Your idea of custom scanner/iterator looks interesting. Is it something you want to write after this PR gets merged?

ti-mo:
@alban Closing this in favor of https://github.com/cilium/ebpf/pull/1588. It reduces allocs more or less to the same degree.

PR #1585, opened 2024-10-14
prog: don't search for kmod when not needed

alban:
When using btfgen (to provide BTF information for kernels without BTF support), spec.LoadAndAssign() is called with opts.KernelTypes but without opts.KernelModuleTypes.

In this scenario, newProgramWithOptions() never makes use of kmod. Therefore we don't need to read /proc/kallsyms to find out the kmod implementing the symbol.

This patch removes the call to spec.KernelModule() in that case. This saves us about 10MB of memory allocation.

Related optimization: #1584

cc @burak-ok

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -276,9 +276,15 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 	insns := make(asm.Instructions, len(spec.Instructions))
 	copy(insns, spec.Instructions)
 
-	kmodName, err := spec.KernelModule()
-	if err != nil {
-		return nil, fmt.Errorf("kernel module search: %w", err)
+	var kmodName string
+	var err error
+
+	// applyRelocations() only uses kmodName if len(targets) == 0
+	if opts.KernelTypes == nil || len(opts.KernelModuleTypes) != 0 {
+		kmodName, err = spec.KernelModule()
+		if err != nil {
+			return nil, fmt.Errorf("kernel module search: %w", err)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


alban:
@ti-mo While you are working on kallsyms optimizations, do you mind reviewing this PR as well?

ti-mo:
Another worthwhile optimization, but the wrong way to go about it. This is already somewhat of a spaghetti and this PR doesn't make it much better. `newProgramWithOptions` shouldn't make too many assumptions about what `applyRelocations` is doing.

I'd like to change the caching strategy of kallsyms to make `FlushKernelModuleCache()` obsolete. We should never cache all of kallsyms, only the values we actually pull out of it. To get there, NewCollection and LoadAndAssign could prime the kallsyms cache for all ProgramSpec.AttachTo values it knows about, so `ProgramSpec.KernelModule()` becomes essentially free.

Then, downstream code can just call the method without worrying too much about whether or not it will impact performance.

ti-mo:
@alban Please give https://github.com/cilium/ebpf/pull/1590 a try instead. `KernelModuleTypes` is kind of bolted on as it is and I'd rather not touch it unless it's for refactoring (e.g. moving more of it into `applyRelocations` or giving it its own function with a clear purpose. Without this patch of yours, /proc/kallsyms is still parsed, but only once for the whole Collection, and only the lookups are cached, not the whole thing.

I still need to fix up some of the tests for older kernels, but this should work already.

ti-mo:
Superseded by https://github.com/cilium/ebpf/pull/1590.

PR #1587, opened 2024-10-16
Explicitly check ksym address was found during relocations

patrickpichler:
Running bpf2go as non root user is not that uncommon. This caused the ksym relocation check to fail, as it only checked if the returned address was not the default value, not if it is present in the ksym map.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -638,7 +638,7 @@ func (ec *elfCode) relocateInstruction(ins *asm.Instruction, rel elf.Symbol) err
 		}
 
 		kf := ec.kfuncs[name]
-		ksymAddr := ec.ksyms[name]
+		ksymAddr, ksymAddrFound := ec.ksyms[name]
 
 		switch {
 		// If a Call / DWordLoad instruction is found and the datasec has a btf.Func with a Name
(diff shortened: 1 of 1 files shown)


patrickpichler:
@dylandreimerink @ti-mo I noticed that ksyms relocation do not work when running e.g. bpf2go as non root.

ti-mo:
@patrickpichler Sorry, what's the relevance of bpf2go in this context? This commit doesn't touch bpf2go code, and ksym resolution is not a compile-time thing.

It's true that reading /proc/kallsyms as an unprivileged user will show zeroes for all addresses. Are you trying to load bpf as an unprivileged user?

patrickpichler:
@ti-mo bpf2go will load the collection ([link](https://github.com/cilium/ebpf/blob/9356805af830c669f5bc9a9bf8eb2e654cf9c76a/cmd/bpf2go/main.go#L360)), which in turn will try to apply the relocations. Before a ksym of `0` would be skipped, which then would cause the case after handling ksyms to be hit (I guess it is specific to my current program, as I also use some external `kconfig` variables). This then cause `bpf2go` to fail, with `relocating instruction: kconfig socket_file_ops not found in .kconfig`. 

I hope this clears it up 😅

ti-mo:
bpf2go does not load a Collection, it loads a CollectionSpec. My bad, I didn't catch this during review. We shouldn't be opening /proc/kallsyms at all while loading CollectionSpec. Resolving the actual address needs to be deferred to NewCollection/LoadAndAssign (so, in resolveKsymReferences).

ti-mo:
@patrickpichler This should be addressed by https://github.com/cilium/ebpf/pull/1590, please give that PR a try and let me know if that doesn't fix it for you.

patrickpichler:
Ah yeah, good point regarding `kconfig`, I misunderstood the code, as I assumed it does validate the existing `kconfig` values available, but it only checks for the existing variables. 👌

PR #1588, opened 2024-10-17
 kallsyms: introduce 'reader', an efficient /proc/kallsyms line parser

ti-mo:
This commit introduces kallsyms.reader, meant to address the performance
concerns caused by fmt.Sscanf and bytes.Fields. Both of these options
allocate heavily and aren't flexible in terms of avoiding unnecessary
work like skipping the remainder of a line if we're not interested in
the remainder of its contents.

This approach borrows the implementation for reader.Word() from
bufio.ScanWords, paying attention to handling utf-8 characters correctly.
Kernel modules (out-of-tree, at least) can contain international symbols
in exported global kernel variables, which show up in /proc/kallsyms. We
need to err on the side of caution since this is technically user input
that's beyond our control.

This commit renames kallsyms.LoadSymbolAddresses to AssignAddresses and
kallsyms.KernelModule to SymbolModule respectively. Both implementations
have been changed to use kallsyms.reader.

```
                   │   old.txt   │              new.txt               │
SymbolKmods-16       246.6m ± 1%   281.5m ± 2%  +14.16% (p=0.002 n=6)
AssignAddresses-16   734.9m ± 1%   299.4m ± 1%  -59.26% (p=0.002 n=6)
geomean              425.7m        290.3m       -31.80%

                   │    old.txt    │               new.txt               │
                   │     B/op      │     B/op      vs base               │
SymbolKmods-16        39.81Mi ± 0%   15.81Mi ± 0%  -60.27% (p=0.002 n=6)
AssignAddresses-16   60.622Mi ± 0%   9.881Mi ± 0%  -83.70% (p=0.002 n=6)
geomean               49.12Mi        12.50Mi       -74.55%

                   │   old.txt    │              new.txt               │
                   │  allocs/op   │  allocs/op   vs base               │
SymbolKmods-16        477.4k ± 0%   266.5k ± 0%  -44.18% (p=0.002 n=6)
AssignAddresses-16   2653.0k ± 0%   462.7k ± 0%  -82.56% (p=0.002 n=6)
geomean               1.125M        351.2k       -68.80%
```

The SymbolKmods-16 wall clock time took a small hit because bytes.Fields
has an ASCII fast-path. Still worth it in the bigger picture since the
allocation pressure dropped by a significant amount to compensate, and
future work will change how the caching layer works for both AssignAddresses
and SymbolModule.

cc @alban @patrickpichler

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -1313,7 +1313,7 @@ func (ec *elfCode) loadKsymsSection() error {
 		}
 	}
 
-	if err := kallsyms.LoadSymbolAddresses(ec.ksyms); err != nil {
+	if err := kallsyms.AssignAddresses(ec.ksyms); err != nil {
 		return fmt.Errorf("error while loading ksym addresses: %w", err)
 	}
 
--- a/elf_reader_test.go
+++ b/elf_reader_test.go
@@ -798,7 +798,7 @@ func TestKsym(t *testing.T) {
 		"tty_fops":        0,
 	}
 
-	qt.Assert(t, qt.IsNil(kallsyms.LoadSymbolAddresses(ksyms)))
+	qt.Assert(t, qt.IsNil(kallsyms.AssignAddresses(ksyms)))
 
 	var value uint64
 	qt.Assert(t, qt.IsNil(obj.ArrayMap.Lookup(uint32(0), &value)))
--- a/internal/kallsyms/kallsyms.go
+++ b/internal/kallsyms/kallsyms.go
@@ -1,14 +1,17 @@
 package kallsyms
 
 import (
-	"bufio"
-	"bytes"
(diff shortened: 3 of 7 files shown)


alban:
Hi @ti-mo,
Do you think the kallsyms package could be moved outside of "internal"?
I'm checking if the reader can be reused for some parts in [github.com/inspektor-gadget/inspektor-gadget/pkg/kallsyms](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/kallsyms/kallsyms.go).

ti-mo:
@alban I'm considering it, I'd like to use it for another project that I currently have no time to work on. Feel free to copy it in the meantime, you'll hear about it if we ever end up exporting it.

PR #1589, opened 2024-10-17
btf: LoadSpecFromReader: add filter param to optimize memory

alban:
This is a proof-of-concept. The code is not ready for merging but it shows it is possible to significantly reduce the memory consumption (by 27MB).

---

In the new LoadSpecFromReader, I added a filter parameter to only parse required kernel btf types.

```go
type TypeFilter struct {
       Names map[string]bool
}

func LoadSpecFromReader(rd io.ReaderAt, filter *TypeFilter) (*Spec, error)
```

Users of the cilium/ebpf package can get the list of relocations they programs require and give it to LoadKernelSpecWithFilter:
```go
var filter *btf.TypeFilter
if len(programs) > 0 {
        filter = &btf.TypeFilter{
                Names: map[string]bool{},
        }
        for _, p := range programs {
                iter := p.Instructions.Iterate()
                for iter.Next() {
                        if relo := btf.CORERelocationMetadata(iter.Ins); relo != nil {
                                fmt.Printf("relo %s\n", relo.String())
                                filter.Names[relo.TypeName()] = true
                        }
                }
        }
}
s, err := btf.LoadKernelSpecWithFilter(filter)
...
opts := ebpf.CollectionOptions{
	Programs: ebpf.ProgramOptions{
		KernelTypes: s,
	},
}
...
if err := spec.LoadAndAssign(objs, &opts); err != nil {
```

---

In my tests in Inspektor Gadget (with this [branch](https://github.com/alban/inspektor-gadget/commits/alban_btf_mem/)), pprof flamegraph inuse_space shows a reduction of memory usage from 40MB to 12MB: btf.LoadKernelSpec() is completely removed from the flamegraph. It used to take 27MB.

Before:

image

After:

image

---

It works with parsing the BTF file in two passes:
1. read sequentially the BTF file to get offsets of each BTF type and mark those we want to keep
2. parse the types we want to keep at the offsets found in the previous steps and follow the type dependency graph.

@burak-ok @mauriciovasquezbernal @flyth

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -187,24 +187,24 @@ func LoadSpec(file string) (*Spec, error) {
 	}
 	defer fh.Close()
 
-	return LoadSpecFromReader(fh)
+	return LoadSpecFromReader(fh, nil)
 }
 
 // LoadSpecFromReader reads from an ELF or a raw BTF blob.
 //
 // Returns ErrNotFound if reading from an ELF which contains no BTF. ExtInfos
 // may be nil.
-func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
+func LoadSpecFromReader(rd io.ReaderAt, filter *TypeFilter) (*Spec, error) {
 	file, err := internal.NewSafeELFFile(rd)
 	if err != nil {
 		if bo := guessRawBTFByteOrder(rd); bo != nil {
-			return loadRawSpec(io.NewSectionReader(rd, 0, math.MaxInt64), bo, nil)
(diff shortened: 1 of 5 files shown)


alban:
I will work on refactoring this PoC to remove the breaking changes and to make it automatic so users don't have to specify the btf file in `ebpf.CollectionOptions`.

anjmao:
@alban Awesome optimization. I tried your brunch and got `reference to invalid type id: 13429` error.

ti-mo:
Realistically, we're not going to go this route. Closing this out in favor of lazy BTF decoding: https://github.com/cilium/ebpf/pull/1763.

PR #1590, opened 2024-10-17
 kallsyms: change Modules caching strategy, cache address lookups

ti-mo:
See individual commits for context.

At a high level:
- add caching for __ksym reference resolving
- stop caching all text symbols that have an associated kmod
- for both addresses and modules, cache only the requested symbols (pull-through cache)
- at the beginning of Collection creation, scan all ProgramSpecs and instructions for module and symbol references and request them in two batches
- perform individual symbol lookups (from the populated cache) in `newProgramWithOptions` onwards
- don't open /proc/kallsyms during CollectionSpec loading


collection.go
```diff
@@ -408,6 +413,45 @@ func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) (*collec
...
+	// Look up associated kernel modules for all symbols referenced by
+	// ProgramSpec.AttachTo for program types that support attaching to kmods.
+	mods := make(map[string]string)
+	for _, p := range progs {
+		if p.AttachTo != "" && p.targetsKernelModule() {
+			mods[p.AttachTo] = ""
+		}
+	}
+	if len(mods) != 0 {
+		if err := kallsyms.AssignModules(mods); err != nil {
```
patrickpichler, line 429: I tried this branch locally and it fails for me with an ambiguous error for kernel module `load_elf_phdrs`. If I understood the  old code correctly, it didn't care about duplicate symbols in `/proc/kallsyms`.

dylandreimerink (approved):
This all makes sense. Very nice

alban:
@ti-mo Thanks, I tested this PR in Inspektor Gadget, and it uses between 8 and 10MB less memory than before.

collection.go
ti-mo: Correct, it would ignore kallsyms entries without modules, which is not correct. If a symbol is ambiguous between the core kernel and a module, we should not attach to it. The old code might've picked the address of the base symbol to attach to, but provided the BTF of the symbol in the kernel module.

ti-mo: @patrickpichler Is this a breaking change for you, or were you just testing out ambiguous symbol handling during kmod attach?

ti-mo: Yes, if you'd actually make use of CO:RE things, build your BPF C program against headers of one kernel, deploy your app to another host, then be unlucky enough for some of the fields to have changed between versions, you would've likely hit a verifier error at runtime.

'Unfortunately', in most cases like `load_elf_phdrs`, it's actually the same function with the same signature and underlying types that gets included into multiple compilation units (by way of including a common header) and then gets merged into the same binary during linking of the kernel binary. Things would've kept working fine in your case since it's likely an exact duplicate, but you would run into funky cases where some code paths in the kernel invoke one version of the symbol, while other paths invoke the other, meaning you'd miss events since you're only attaching to one of the symbols.

Let's see if any other skeletons fall out of the ceiling when we ship this in a release, but this has been a bugbear for bpf since the beginning.

PR #1592, opened 2024-10-18
internal: add a way to write portable feature tests

lmb:
Feature tests are used to enable new features in a backwards compatible manner. They exercise very specific code paths in the kernel. It makes sense to assume that the existing feature tests will not work as intended when porting to a new platform. Change the existing NewFeatureTest constructor to return a new sentinel error when invoked on non-Linux OS.

Add a new constructor NewPortableFeatureTest which allows indicating on which platforms and versions a feature test is reliable.

### DIFF
--- a/internal/feature.go
+++ b/internal/feature.go
@@ -3,15 +3,25 @@ package internal
 import (
 	"errors"
 	"fmt"
+	"runtime"
+	"strings"
 	"sync"
 )
 
-// ErrNotSupported indicates that a feature is not supported by the current kernel.
+// ErrNotSupported indicates that a feature is not supported.
 var ErrNotSupported = errors.New("not supported")
+
+// ErrNotSupportedOnOS indicates that a feature is not supported on the current
+// operating system.
+var ErrNotSupportedOnOS = fmt.Errorf("%w on %s", ErrNotSupported, runtime.GOOS)
 
 // UnsupportedFeatureError is returned by FeatureTest() functions.
 type UnsupportedFeatureError struct {
-	// The minimum Linux mainline version required for this feature.
+	// The minimum version required for this feature.
+	//
+	// On Linux this refers to the mainline kernel version, on other platforms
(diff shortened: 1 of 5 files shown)


internal/feature.go
```diff
@@ -3,15 +3,25 @@ package internal
...
+// operating system.
+var ErrNotSupportedOnOS = fmt.Errorf("%w on %s", ErrNotSupported, runtime.GOOS)
+
 // UnsupportedFeatureError is returned by FeatureTest() functions.
 type UnsupportedFeatureError struct {
-	// The minimum Linux mainline version required for this feature.
+	// The minimum version required for this feature.
+	//
+	// On Linux this refers to the mainline kernel version, on other platforms
+	// to the version of the runtime.
```
ti-mo, line 23: Nit: careful to distinguish between 'bpf runtime' and 'Go runtime' when phrasing things like this.

internal/feature.go
```diff
@@ -68,6 +89,33 @@ func NewFeatureTest(name, version string, fn FeatureTestFn) func() error {
 	return ft.execute
 }
 
+// NewPortableFeatureTest creates a cross-platform [FeatureTest].
+//
+// versions are in the format "GOOS:Major.Minor[.Patch]"
+func NewPortableFeatureTest(name string, fn FeatureTestFn, versions ...string) func() error {
```
ti-mo, line 95: I'm not sure why this needs to have a separate constructor. Couldn't we take `Linux:6.11` and `6.11` to mean the same thing for backwards compat? (or just rewrite all the ones we have) Making `version` variadic in `NewFeatureTest` isn't a huge problem as it only has a few dozen invocations. Refactoring should be fairly straightforward.

lmb:
For posterity: I think I accidentally deleted this code when correcting my mistake of force pushing to main.

lmb:
See https://github.com/cilium/ebpf/pull/1592

PR #1602, opened 2024-10-25
feat: export netkit and tcx probes

borkmann:
Export the two feature probes for users. These are quite useful and in particular the one for netkit is needed for Cilium to throw an error earlier.

Link: https://github.com/cilium/cilium/issues/33878


borkmann:
Discussed with Robin it's better if we probe inside Cilium instead of exporting.

Issue #963, opened 2023-03-07
btf: Pointer.Size is incorrect on arches with 32 bit pointers


lmb:
We currently hardcode pointer size:
https://github.com/cilium/ebpf/blob/9447ec825af0802d63ea2d74523536f99622bb59/btf/types.go#L149

The kernel uses [`sizeof(void*)` instead](https://elixir.bootlin.com/linux/v6.2.2/source/kernel/bpf/btf.c#L1960). That size depends on which [data model](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models) is used.The correct size of `Pointer` in Go land is actually `unsafe.Sizeof(uintptr(0))`, but we can't use that directly. We may read BTF generated on a 32 bit pointer arch on a 64 bit pointer arch and vice versa.

It would be nice if we could reject parsing BTF with a different pointer size, or add "proper" support for differing pointer sizes. (What would that look like though?)

ti-mo:
Since this is mainly (only?) used for CO-RE/btf marshaling, and the kernel's `btf_type_id_size` is only called during btf unmarshaling and map creation, `unsafe.Sizeof(uintptr(0))` may be the right thing to do here.

Since the pointer size isn't encoded in the BTF wire format, what else would we return?

ti-mo:
My thoughts exactly, not sure what the state of bpf on 32-bit is atm.

PR #1607, opened 2024-11-07
[WIP] variable: return native Go pointers to underlying bpf map values

ti-mo:
Pending the proposal in https://github.com/golang/go/issues/70224, here's the current API proposal for returning Go pointers to 'off-heap' bpf map memory:

```go
func MemoryPointer[T any](mm *Memory, off uint64) (*T, error)
func VariablePointer[T any](v *Variable) (*T, error)
```

We need a bit of help from the Go runtime to manage the lifecycle of the underlying memory mapping. Currently, this mmaps over a piece of Go heap to allow the GC to track pointers into the bpf map memory, obviating the need for tying the mmap's lifetime to an `ebpf.Memory`. Instead, any Go pointers into the mmap will keep it alive, removing the risk of use-after-free.


variable.go
```diff
@@ -232,3 +233,37 @@ func (v *Variable) Get(out any) error {
...
+// not valid.
+//
+// When accessing structs, embedding [structs.HostLayout] may help ensure the
+// layout of the Go struct matches the one in the BPF C program.
+func VariablePointer[T any](v *Variable) (*T, error) {
+	if err := checkVariable[T](v); err != nil {
+		return nil, fmt.Errorf("variable pointer %s: %w", v.name, err)
+	}
+	return memoryPointer[T](v.mm, v.offset)
+}
```
mejedi, line 269: It doesn't look like it is possible to support slices within the same API, is it?

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+	vs, bs := uint64(size), uint64(len(mm.b))
+	if off+vs > bs {
+		return fmt.Errorf("%d-byte value at offset %d exceeds mmap size of %d bytes", vs, off, bs)
+	}
+
+	return nil
+}
+
+// reinterp reinterprets a pointer of type In to a pointer of type Out.
+func reinterp[Out any, In any](in *In) *Out {
```
lmb, line 238: Nit: just do the cast directly?

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if mm.ro {
+		return ErrReadOnly
+	}
+	if !mm.heap {
+		return fmt.Errorf("memory region is not heap-mapped, see CollectionOptions.UnsafeVariableExperiment: %w", ErrNotSupported)
+	}
+
+	var t T
+	size := binary.Size(t)
```
lmb, line 219: Why is this binary.Size and not unsafe.Sizeof?

lmb: This also doesn't take `unsafe.Alignof` into account.

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+	// Return an aligned pointer into the backing array, losing the original
+	// reference. The runtime.SetFinalizer docs specify that its argument 'must be
+	// a pointer to an object, complit or local var', but this is still somewhat
+	// vague and not enforced by the current implementation.
+	//
+	// Currently, finalizers can be set and triggered from any address within a
+	// heap allocation, even individual struct fields or arbitrary offsets within
+	// a slice. In this case, finalizers set on struct fields or slice offsets
+	// will only run when the whole struct or backing array are collected. The
+	// accepted runtime.AddCleanup proposal makes this behaviour more explicit and
```
lmb, line 126: Does all of this still work when switching to AddCleanup?

variable.go
```diff
@@ -232,3 +233,37 @@ func (v *Variable) Get(out any) error {
...
 	return nil
 }
+
+func checkVariable[T any](v *Variable) error {
+	if v.ReadOnly() {
+		return ErrReadOnly
+	}
+
+	var t T
+	size := binary.Size(t)
```
lmb, line 243: Would it be clearer to check that `unsafe.Sizeof() == binary.Size()`?

variable_test.go
```diff
@@ -182,3 +185,113 @@ func TestVariableFallback(t *testing.T) {
...
+		AddAtomic   *Program `ebpf:"add_atomic"`
+		CheckStruct *Program `ebpf:"check_struct"`
+		CheckArray  *Program `ebpf:"check_array"`
+
+		Atomic *Variable `ebpf:"var_atomic"`
+		Struct *Variable `ebpf:"var_struct"`
+		Array  *Variable `ebpf:"var_array"`
+	}{}
+
+	qt.Assert(t, qt.IsNil(spec.LoadAndAssign(&obj, &CollectionOptions{UnsafeVariableExperiment: true})))
```
lmb, line 206: Please use new `loadAndAssign` helper.

variable_test.go
```diff
@@ -182,3 +185,113 @@ func TestVariableFallback(t *testing.T) {
...
+	t.Cleanup(func() {
+		obj.AddAtomic.Close()
+		obj.CheckStruct.Close()
+		obj.CheckArray.Close()
+	})
+
+	// Bump the value by 1 using a bpf program.
+	want := uint32(1338)
+	a32, err := VariablePointer[atomic.Uint32](obj.Atomic)
```
lmb, line 215: So binary.Size accepts atomic.Uint32?

variable_test.go
```diff
@@ -182,3 +185,113 @@ func TestVariableFallback(t *testing.T) {
...
+	// Pull out Program handle and Variable pointer so obj reference is dropped.
+	prog := obj.AddAtomic
+	t.Cleanup(func() {
+		prog.Close()
+	})
+
+	a32, err := VariablePointer[atomic.Uint32](obj.Atomic)
+	qt.Assert(t, qt.IsNil(err))
+
+	// No obj references past this point. Trigger multiple GC cycles to ensure
```
lmb, line 283: A single one is not enough?

variable_test.go
```diff
@@ -182,3 +185,113 @@ func TestVariableFallback(t *testing.T) {
...
+	// obj is collected.
+	runtime.GC()
+	runtime.GC()
+	runtime.GC()
+
+	// Trigger prog and read memory multiple times to ensure reference is still
```
lmb, line 289: There is no test which ensures that GC actually happens.

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+// This obviates the need to maintain a reference to the *Memory at all times,
+// which is difficult for the caller to achieve if the variable access is done
+// through another object (like a sync.Atomic) that can potentially be passed
+// around the Go application. Accidentally losing the reference to the *Memory
+// would result in hard-to-debug segfaults, which are always unexpected in Go.
+
+//go:linkname heapObjectsCanMove runtime.heapObjectsCanMove
+func heapObjectsCanMove() bool
+
+var ErrInvalidType = errors.New("invalid type")
```
lmb, line 43: Nit: doc why is is invalid?

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	// Set a finalizer on the heap allocation to undo the mapping before the span
+	// is collected and reused by the runtime. This has a few reasons:
+	//
+	//  - Avoid leaking memory/mappings.
+	//  - Future writes to this memory should never clobber a bpf map's contents.
+	//  - Some bpf maps are mapped read-only, causing a segfault if the runtime
+	//    reallocates and zeroes the span later.
+	runtime.SetFinalizer((*byte)(addr), unmap(size))
```
lmb, line 167: Would it be possible to set the finalizer on the original "real allocation" as well?

memory_unsafe_test.go
lmb, line 1: I think there are lots of assumptions that aren't explored yet:
* GC behaviour on mappings which need to be aligned and therefore don't alias the original allocation
* What happens when the heap mapping is dropped while the mapmap still exists
* What happens when a user sets a finalizer on type-cast memory
* Test that things are actually freed at the right point in time
* Test that a pointer at an offset to the original allocation keeps the whole thing alive

collection.go
```diff
@@ -33,6 +33,15 @@ type CollectionOptions struct {
 	// The given Maps are Clone()d before being used in the Collection, so the
 	// caller can Close() them freely when they are no longer needed.
 	MapReplacements map[string]*Map
+
+	// UnsafeVariableExperiment enables the unsafe variable experiment, allowing
+	// the use of [VariablePointer] for direct memory access and atomic operations
+	// on global bpf variables.
+	//
+	// Experimental: enabling this may cause segfaults or compromise the memory
```
lmb, line 41: We should also document that this may go away at any point in time if it is necessary for the health of the ecosystem. Go runtime >> this feature.

lmb (commented):
Can either of you refresh my mind again what the killer feature for this is? Seems like @mejedi is interested in using this for slices?

memory_unsafe.go
lmb: (I think that might be easier to reason about)

variable.go
ti-mo: Not in the current proposal, no. I think supporting slices is a hazard, and we should probably document that clearly. (`append()` will reallocate and copy, unsharing the backing array)

Of course the caller can always slice the array, but we should encourage passing around the array pointer instead.

memory_unsafe.go
ti-mo: Need to check, but afaik AddCleanup is SetFinalizer modulo object resurrection.

memory_unsafe.go
ti-mo: Sure. The initial proposal had dozens of these, so this made things less error-prone.

memory_unsafe.go
ti-mo: I think this came up in a review on the initial patch set in another PR. To do what you suggest, we'd need to return the initial []byte from `allocate()` in addition to the unsafe.Pointer, which could end up pointing to a different offset within the backing array than the `unsafe.Pointer`, since the latter is potentially aligned differently. It seems like a worse idea to return two (potentially different) pointers to the same backing array.

In practice, this does not make a difference, here's a relevant playground: https://go.dev/play/p/FKVnnU85C8b. This behaviour (supporting multiple finalizers per alloc) is unlikely to change. If it does (and 1 span == 1 finalizer/cleanup), I expect a panic due to there being an existing finalizer, much like setting multiple finalizers on the same address does today. Either way, we only set one finalizer per alloc, although potentially at an offset larger than zero. This is fine for the course of the experiment IMO.

variable.go
ti-mo: I think that would make things more complex, not less. The reader would have to be aware of the intricacies and the differences between both, and we'd still need the comparison to `v.size`, so it wouldn't eliminate any code.

variable_test.go
ti-mo: Yep, it's a fixed-size type.

variable_test.go
ti-mo: See https://go.dev/play/p/FKVnnU85C8b and https://pkg.go.dev/runtime@master#SetFinalizer:

> When the garbage collector finds an unreachable block with an associated finalizer, it clears the association and runs finalizer(obj) in a separate goroutine. This makes obj reachable again, but now without an associated finalizer. Assuming that SetFinalizer is not called again, the next time the garbage collector sees that obj is unreachable, it will free obj.

memory_unsafe_test.go
ti-mo: > * GC behaviour on mappings which need to be aligned and therefore don't alias the original allocation
> * What happens when a user sets a finalizer on type-cast memory

https://go.dev/play/p/FKVnnU85C8b

> * Test that a pointer at an offset to the original allocation keeps the whole thing alive

Could this question be nuanced? This seems to be inherent to how GC works.

> * What happens when the heap mapping is dropped while the mapmap still exists

Not sure how this can happen without a moving GC and while there are still references to a span.

> * Test that things are actually freed at the right point in time

Let me come up with something.

variable_test.go
ti-mo: I ended up only needing a single one after making the main goroutine wait for a channel closed by a finalizer.

ti-mo:
> Can either of you refresh my mind again what the killer feature for this is? Seems like @mejedi is interested in using this for slices?

I think the most pressing use case is atomics, which is currently not supported at all. Array variables currently work through Get() and Set(), but there's always a marshaling cost. Also, simply having the convenience of manipulating BPF variables like they are Go variables is nice.

collection.go
ti-mo: The goal is to never let things get to this point. The path has been charted using `AddForeignCleanup`, so if this workaround ends up painting the runtime into a corner, that's all the more reason to get foreign cleanup shipped ASAP.

variable_test.go
lmb: I should've said: it also only contains unexported fields. Passing the type to binary.Read or Write will cause an error. Is that the correct choice to make here?

memory_unsafe.go
lmb: I missed the detail that the finalizer doesn't just unmap, but rather remaps the sub-mapping.

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+// Used as a finalizer in [newMemory], split off into a separate function for
+// testing and to avoid accidentally closing over the unsafe.Pointer to the
+// memory region, which would cause a cyclical reference.
+//
+// The resulting function panics if the mmap operation returns an error, since
+// it would mean the integrity of the Go heap is compromised.
+func unmap(size int) func(*byte) {
+	return func(a *byte) {
+		// Create another mapping at the same address to undo the original mapping.
+		// This will cause the kernel to repair the slab since we're using the same
```
lmb, line 187: What would happen if the kernel didn't "repair the slab" for some reason, and kept a separate mapping instead? Nothing in the man pages suggests that this is required behaviour.

memory_unsafe.go
```diff
@@ -0,0 +1,256 @@
...
+
+	if size == 0 {
+		return nil, errors.New("size must be greater than 0")
+	}
+
+	// Request at least two pages of memory from the runtime to ensure we can
+	// align the requested allocation to a page boundary. This is needed for
+	// MAP_FIXED and makes sure we don't mmap over some other allocation on the Go
+	// heap.
+	size = internal.Align(size+os.Getpagesize(), os.Getpagesize())
```
lmb, line 109: Something we discussed on the previous PR I think: this will be expensive on platforms that have large page sizes like arm64. Not sure there is anything we can do about it, but might make sense to call it out in the docs.

memory_unsafe.go
ti-mo: > What would happen if the kernel didn't "repair the slab" for some reason, and kept a separate mapping instead?

I just documented this as an observation; you're right that this isn't documented behaviour, it's just how Linux behaves. There would be no behavioural change if the kernel doesn't repair the slab (basically, if the kernel no longer defragments contiguous page table entries/vmas). munmap takes addr and length, so if the runtime decides to release the mspan, it would munmap the whole range it thought it owned, including the intermediate mappings created by our hack (https://pubs.opengroup.org/onlinepubs/009696799/functions/munmap.html):

> The munmap() function shall remove any mappings for those entire pages containing any part of the address space of the process starting at addr and continuing for len bytes.

variable_test.go
ti-mo: > I should've said: it also only contains unexported fields. Passing the type to binary.Read or Write will cause an error. 

Ah, gotcha.

> Is that the correct choice to make here?

What choice? :sweat_smile: With atomics being a major use case, this is important to have tests for, if that's what you mean.

lmb: I mean: should you allow types which have unexported fields in the first place. Should I be able to place a `sync.Mutex` in there?

memory_unsafe.go
```diff
@@ -0,0 +1,303 @@
...
+	return nil
+}
+
+// checkType recursively checks if the given type is supported for memory
+// mapping. Only fixed-size types are supported: bools, floats, (u)int[8-64],
+// arrays, and structs containing them.
+//
+// Doesn't check for loops since it rejects pointers. Should that ever change, a
+// visited set would be needed.
+func checkType(name string, t reflect.Type) error {
```
lmb, line 253: I have a hunch that you'll have to allow `uintptr` because that is the natural way to represent a C pointer in Go. That only works as long as user space and kernel space are both 64 bit of course xD

lmb (commented):
Thanks for adding all those tests, they look excellent to me! Left one thought around how to enable / disable this, feel free to deal with that as you wish. I have concerns around what types should be allowed, where I'd appreciate it if we can be even more conservative for a start.

collection.go
```diff
@@ -35,6 +35,15 @@ type CollectionOptions struct {
...
+	// integrity of your Go application or the bpf maps representing Variables.
+	// Use at your own risk.
+	UnsafeVariableExperiment bool
```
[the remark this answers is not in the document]
ti-mo: Yep, didn't manage to get that done last week. Initially, I was tempted to extend the experiment to `MemoryPointer()` as well, but while implementing the type checker, I realized `MemoryPointer()` for arenas would be relatively useless without allowing pointers, so that needs a bit more work.

memory_unsafe.go
```diff
@@ -0,0 +1,303 @@
...
+	case reflect.Struct:
+		for i := range t.NumField() {
+			at := t.Field(i).Type
+			if err := checkType(fmt.Sprintf("%s.%s", name, at.String()), at); err != nil {
+				return err
+			}
+		}
+
+	default:
+		if t.String() == t.Kind().String() {
```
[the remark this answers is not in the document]
ti-mo: Sure, I thought it was clear from the `Output:` examples in the following line.

For built-in types, `reflect.Type.String()` and `reflect.Kind.String()` are the same, so the fallthrough case would print something like `int value int not supported`, which is a little confusing.

memory_unsafe_test.go
```diff
@@ -0,0 +1,104 @@
...
+	rd, err := mustMmapableArray(t, sys.BPF_F_RDONLY_PROG).unsafeMemory()
+	qt.Assert(t, qt.IsNil(err))
+
+	// BPF_F_RDONLY_PROG flag, so the Memory should be read-only.
+	qt.Assert(t, qt.IsTrue(rd.ReadOnly()))
+
+	// Frozen maps can't be mapped rw either.
+	frozen := mustMmapableArray(t, 0)
+	qt.Assert(t, qt.IsNil(frozen.Freeze()))
+	fz, err := frozen.Memory()
```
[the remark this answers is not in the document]
ti-mo: EBUSY: https://elixir.bootlin.com/linux/v6.13.7/source/kernel/bpf/syscall.c#L2182

variable_test.go
ti-mo: > There are atomic free functions that you can use without relying on the atomic types.

I feel like this has been discussed and considered at nauseam. Within the context of this global variable API, this would require the user to take out a normal *uint32 that they must then very carefully guard with atomic accessors. This is essentially what `atomic.Uint32` and friends already do, so I don't see much point in removing this option for users. The [documentation](https://pkg.go.dev/sync/atomic#AddInt32) also consistently warns about using the free functions:

> Consider using the more ergonomic and less error-prone [Int32.Add](https://pkg.go.dev/sync/atomic#Int32.Add) instead.

It also mentions 32-bit related bugs, but again not clear what the state of bpf on 32-bit is today, so that doesn't matter much to us.

> Why would they want to though? ... and I'm convinced that allowing sync.Mutex is a bad idea.

Exactly, bpf doesn't have mutexes, so I don't really see the point you're trying to make. If it fits in the same space in memory and access is aligned, we don't really care.

I don't hear any concrete suggestions; do you want to special-case and reject `sync.Mutex`? What about a `sync.Once`? What makes these so special as to warrant this limitation? Feels quite arbitrary to me.

Same the other way around; why would we accept `atomic.Bool` but not other structs of the same width? (e.g. a user-made atomic type with some extra behaviours, maybe for tracing?) This feels equally arbitrary.

> Only structs without unexported fields

AFAIU, that limitation exists to allow the user to omit certain fields from marshaling operations and from comparisons in go-cmp, etc. In the context of this API, any given type is projected into C memory in its entirety, nothing can be omitted. As such, this restriction doesn't make sense here.

> Containing structs.HostLayout

I think it's reasonable to offer this as a hint, e.g. in an ErrInvalidType error message. Or perhaps we could propagate a btf.Type into the Variable, which, if present, align-checks with the Go struct and issues a similar warning if needed. (We want to put Cilium's alignchecker into the lib at some point)

I don't think we should ever make HostLayout a hard requirement, though, since it would require us to special-case atomics, which is a slippery slope as mentioned above.

memory_unsafe.go
ti-mo: Good point, we can allow uintptr if it's 8 bytes.

Reminder: we still treat btf.Pointer as 64 bits https://github.com/cilium/ebpf/issues/963.

variable_test.go
lmb: > I don't hear any concrete suggestions; do you want to special-case and reject `sync.Mutex`? What about a `sync.Once`? What makes these so special as to warrant this limitation?

They aren't special, just useful examples. My concrete suggestion is: only allow structs with all-exported fields.

> why would we accept atomic.Bool but not other structs of the same width? (e.g. a user-made atomic type with some extra behaviours, maybe for tracing?) This feels equally arbitrary.

We'd allow them because there are ergonomic building blocks, as you point out. The user could still have their own type wrapping them if they desire:

```
type myAtomicFancyBool struct{
   Value atomic.Bool
}
```

> because cross-package assignments to unexported fields would be surprising. Within the context of this API, any given type is projected into C memory in its entirety, nothing can be omitted. It doesn't cross a package boundary.

If you replace "package boundary" with "abstraction layer" or "implementation boundary" I'd say it does: `sync.Cond` escaping into C memory leaks the implementation / abstraction. Hence it makes sense to me to apply the same restrictions here.

The current approach is "any type you can get a pointer to is fair game and may be mapped into shared memory". I think that goes against the spirit of the Go type system, package reflect, etc.

> I don't think we should ever make HostLayout a hard requirement, though, since it would require us to special-case atomics, which is a slippery slope as mentioned above.

Then that leaves the question what to do if / when the compiler does reorder layouts? I guess you could always add that requirement then...

ti-mo: > We'd allow them because there are ergonomic building blocks, as you point out. The user could still have their own type wrapping them if they desire:

No this is kind of my whole point: atomic.Bool contains an unexported field, so it would be rejected according to the rules you propose (the rules need to be applied recursively for obvious reasons). This will lock out many good use cases like this, where the user doesn't necessarily control the type they want to use as a variable.

ti-mo: Discussed offline: we'll require `structs.HostLayout` in all structs, special-case `atomic` types, and don't require all fields to be exported.

Also, exploring whether we can use a build tag instead of an env.

PR #1609, opened 2024-11-11
prog, map: add nil checks for value marshaller

aibor:
In case of programs or maps as map value types, the value marshalling results in a panic in case of a nil pointer. For other types this is not a problem since sysenc.Marshal does a nil check.

This commit adds a nil check to the internal helper for marshalling programs and maps.


ti-mo:
Thanks! Pushed a small fixup.

aibor:
Great, thanks for the quick fixup and merge.

PR #1610, opened 2024-11-13
land bpf2go basic support for Variable and VariableSpec

smagnani96:
Commit messages should (hopefully) describe the introduced changes. In brief:

1. `add bpf2go basic support to emit Variables and VariableSpecs`: as discussed in the reference PR*, this commit adds the minimal support to bpf2go for emitting `Variables` and `VariableSpecs`. Introduced in https://github.com/cilium/ebpf/pull/1572, `Variables` are always emitted for every `VariableSpec` in the `CollectionSpec`, despite its usage being supported or not (kernel >=v5.5).
2. `fix typo in comment when bpf2go emits bpfProgramSpecs`: fixed a typo in a bpf2go-generated comment (`bpfSpecs` rather than `bpfProgramSpecs`)
3. `regenerate all TARGETS in main`: plain execution of `make`, to regenerate all the targets after (1).
4. `add bpf2go test emit-and-load specs+bpf objects`: extended `api_test.go` to check also for Variables to be loaded, and all the `{Map,Variable,Program}Specs` to be loaded. Introduced new test in `output_test.go` to verify that the generated bpf2go output is as expected for every spec and bpf object.
5. `simplify and update tcx example to use Variable API`: showcase the new Variable API in the TCX example. Back then when we introduced it, we wanted to use global variables, but there were no support for that, so we used bpf maps for counting packets.

*Reference PR with the discussion: https://github.com/cilium/ebpf/pull/1543


smagnani96:
Had to repeat commit n.3, regenerating targets due to other changes.
In the meantime I mistakenly pushed local examples I was keeping.
Finally, I had to make a change in commit n.5.

Now should be good to review. Actual line changes should not be that much, the others are all auto-generated stuff :crossed_fingers:

mejedi (approved):

mejedi:
We should consider generating types exposed via variables by default. Right now, maps key and value types are emitted. E.g.
```
struct struct_type variable;
```
Explicitly adding `-type struct_type` feels redundant.

smagnani96:
> Right now, maps key and value types are emitted. We should consider generating types exposed via variables by default. E.g.
> 
> ```
> struct struct_type variable;
> ```
> 
> Explicitly adding `-type struct_type` feels redundant.

In that regards, I just pushed 3 new commits for which I'd love to receive further feedback :pray: 

I was in doubt where to put the `shouldEmit` function https://github.com/cilium/ebpf/pull/1610/files#diff-4613c0526058156a97d6786beea6b8499c231f81ca1fc4e6f4d3663ff578f0d1R53-R61: I didn't want to insert that logic in `btf/format.go`, since I believe that it is strictly related to `bpf2go` and not `btf` itself.
Let me explain why I needed this piece of code:
1. take btf.Type iterating through btf.Datasec would lead to a mismatch when checking an ebpf.VariableSpec type against the generated one.
2. passing as it is a btf.Var would cause our GoFormatter to emit whatever type it receives, leading to potentially redundant generated code: I wouldn't expect us to generate a simple `type t uint32`, similarly as we do for maps.

For this reason, I'd like to prevent this behavior from bpf2go perspective.
Let me know if it sounds good or if there's a better way to address this.

smagnani96:
> Right now, maps key and value types are emitted. We should consider generating types exposed via variables by default. E.g.
> 
> ```
> struct struct_type variable;
> ```
> 
> Explicitly adding `-type struct_type` feels redundant.

In that regards, I just pushed 3 new commits for which I'd love to receive further feedback 🙏

I was in doubt where to put the shouldEmit function https://github.com/cilium/ebpf/pull/1610/files#diff-4613c0526058156a97d6786beea6b8499c231f81ca1fc4e6f4d3663ff578f0d1R53-R61: I didn't want to insert that logic in btf/format.go, since I believe that it is strictly related to bpf2go and not btf itself.
Let me explain why I needed this piece of code:

take btf.Type iterating through btf.Datasec would lead to a mismatch when checking an ebpf.VariableSpec type against the generated one.
passing as it is a btf.Var would cause our GoFormatter to emit whatever type it receives, leading to potentially redundant generated code: I wouldn't expect us to generate a simple type t uint32, similarly as we do for maps.
For this reason, I'd like to prevent this behavior from bpf2go perspective.
Let me know if it sounds good or if there's a better way to address this.

ti-mo:
I've been polishing this a bit to get it into shape for the release. Please hold off on pushing any changes here.

ti-mo:
@smagnani96 I've done some polishing and squashed this down into a few commits. In the future, please regenerate testdata in the same commit as when codegen/testdata changes are made, otherwise commits can't be individually reverted and rebasing becomes painful. I've mostly rewritten the functionality related to CollectGlobalTypes and added some tests. I had to get hands-on to familiarize myself with the problem, only to realize it's all very finnicky and the existing code was sort of broken, so doing this through code review would've been too time-consuming.

@mejedi PTAL so this can go into this week's release, see individual commits. Upon closer inspection, I found several flaws in `CollectGlobalTypes` and decided to give it a rework. It's still a bit hard to follow, but I've tried separating concerns as much as I could.

mejedi:
@ti-mo 
> @mejedi PTAL so this can go into this week's release, see individual commits. Upon closer inspection, I found several flaws in `CollectGlobalTypes` and decided to give it a rework. It's still a bit hard to follow, but I've tried separating concerns as much as I could.

Congrats on the release! I love the changes in `CollectTypes`.

mejedi:
@ti-mo 
I ended up rewriting `sortTypes` in #1636

PR #1612, opened 2024-11-14
Unprivileged `objNameAllowsDot`, return *btf.Var from `Variable{Spec}.Type()`

ti-mo:
Fixes for a few papercuts encountered from implementing the API into the Cilium agent.

Changed my mind about returning btf.Type from .Type(), this should just be *btf.Var. No need in obscuring the Var, it's the easiest way to get to decl tags.


ti-mo:
Removed breaking-change label as this API is yet to appear in a release.

syscalls.go
```diff
@@ -201,10 +209,19 @@ var objNameAllowsDot = internal.NewFeatureTest("dot in object names", func() err
 		MapName:    sys.NewObjName(".test"),
 	}
 
+	// Tolerate EPERM, otherwise MapSpec.Name has its dots removed when run by
```
lmb, line 212: @ti-mo Just found this as I was rebasing other work. I think the fact that the elf reader changes based on the host system is the problem here, not that the test doesn't account for not having permissions. I'll see whether I can untangle my mess.

PR #1619, opened 2024-12-05
Optimize `epoll.Wait` by using sentinel errors

def:
Batched perfmap readers with 10ms deadlines showed `fmt.Errorf` consuming ~4% of CPU. This PR reduces CPU usage by replacing repeated formatting with package-level error variables.

Screenshot 2024-12-05 at 10 16 24


ti-mo (approved):
Awesome, thanks!

PR #1625, opened 2024-12-12
bpf2go: enable ebpf code reuse across go packages

mejedi:
New -import command line option specifies golang packages to pull headers from. Scan packages for header files, and expose headers to clang. C code consumes headers by providing a go package path in include directive, e.g.

frob.c:
  #include "example.org/foo/foo.h"

It is handy for sharing code between multiple ebpf blobs within a project. Even better, it enables sharing ebpf code between multiple projects using go modules as delivery vehicle.

"Build" dependencies are recorded in generated files so that "go mod tidy" can see them. Further, we don't check if -import refers to internal packages, but "go build" will.

### DIFF
--- a/cmd/bpf2go/README.md
+++ b/cmd/bpf2go/README.md
@@ -34,6 +34,15 @@ up-to-date list.
 disable this behaviour using `-no-global-types`. You can add to the set of
 types by specifying `-type foo` for each type you'd like to generate.
 
+## eBPF packages
+
+`bpf2go` can pull header files from other Go packages. List packages you
+wish to pull headers from using `-import` command line option, e.g.:
+`-import example.org/foo`.
+```
+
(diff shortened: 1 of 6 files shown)


mejedi:
Change to listing packages explicitly on command line with `-import` option.

Prior iteration (#1509) pulled them from the "magic" `bpf2go.hfiles.go` in the output directory.

ti-mo:
I don't think we want bpf2go to do C dependency management using Go packages as a vehicle. It's a great idea, but again not something the majority of users will need to leverage. This kind of stuff is also bound to be specific to the  the user's code/package layout, so I don't see this translate to the masses.

Personally, I'd write a separate tool that can pull C headers from Go packages and roll it into a vfs artifact. (which is basically a tar) Then, call it using a `//go:generate` statement preceding `bpf2go` and specify the vfs path as a c flag.

PR #1626, opened 2024-12-12
pin: new package for loading bpf pins and walking bpffs directories

ti-mo:
See individual commits.

The API added in this PR:

```go
type Pinner interface {
	Pin(string) error
}

// Load retrieves a pinned object from a bpf virtual filesystem. It returns one
// of [ebpf.Map], [ebpf.Program], or [link.Link].
//
// Trying to open anything other than a bpf object is an error.
func Load(path string, opts *ebpf.LoadPinOptions) (Pinner, error)

// WalkDirFunc is the type of the function called for each object visited by
// [WalkDir]. It's identical to [fs.WalkDirFunc], but with an extra [Pinner]
// argument. If the visited node is a directory, obj is nil.
//
// err contains any errors encountered during bpffs traversal or object loading.
type WalkDirFunc func(path string, d fs.DirEntry, obj Pinner, err error) error

// WalkDir walks the file tree rooted at path, calling bpffn for each node in
// the tree, including directories. Running WalkDir on a non-bpf filesystem is
// an error. Otherwise identical in behavior to [fs.WalkDir].
//
// See the [WalkDirFunc] for more information.
func WalkDir(root string, bpffn WalkDirFunc) error
```

### DIFF
--- a/internal/sys/fd.go
+++ b/internal/sys/fd.go
@@ -81,10 +81,13 @@ func (fd *FD) Close() error {
 		return nil
 	}
 
-	return unix.Close(fd.disown())
+	return unix.Close(fd.Disown())
 }
 
-func (fd *FD) disown() int {
+// Disown destroys the FD and returns its raw file descriptor without closing
+// it. After this call, the underlying fd is no longer tied to the FD's
+// lifecycle.
+func (fd *FD) Disown() int {
 	value := fd.raw
 	fdtrace.ForgetFD(value)
 	fd.raw = -1
@@ -118,7 +121,7 @@ func (fd *FD) File(name string) *os.File {
 		return nil
 	}
 
-	return os.NewFile(uintptr(fd.disown()), name)
+	return os.NewFile(uintptr(fd.Disown()), name)
 }
 
 // ObjGetTyped wraps [ObjGet] with a readlink call to extract the type of the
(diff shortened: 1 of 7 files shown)


mtardy (approved):
I love this!

internal/sys/fd.go
```diff
@@ -81,10 +81,13 @@ func (fd *FD) Close() error {
 		return nil
 	}
 
-	return unix.Close(fd.disown())
+	return unix.Close(fd.Disown())
 }
 
-func (fd *FD) disown() int {
+// Disown destroys the FD and returns its raw file descriptor without closing
```
rgo3, line 87: Nittiest of nits: This comment gets added with the last commit instead of the first.

rgo3 (approved):
:very_nice:

internal/sys/fd.go
ti-mo: MVP comment, looks like I messed up the fixup, thanks!

dylandreimerink (approved):
Looking good to me :rocket:

mtardy:
This is so much better, look at the improvement https://github.com/cilium/tetragon/pull/3206/commits/4c31e33e447a025ac0ad16ddd7a291ee67d402a6. It removed sketchy code. Thanks again!

ti-mo:
> This is so much better, look at the improvement https://github.com/cilium/tetragon/pull/3206/commits/ef53ab5dfe6ab37eda9342246f04a05c53a84121. It removed sketchy code. Thanks again!

Thanks for the feedback, enjoy!

PR #1627, opened 2024-12-12
map: automatically set CPUMap MaxEntries based on possible CPUs

learnitall:
This commit modifies the `MapSpec.fixupMagicFields` method to support the following for the CPU map:

  * Validation of KeySize and ValueSize
  * Setting KeySize, ValueSize and MaxEntries when set to zero
  * Clamping MaxEntries to the number of possible CPUs

This enables to write more portable code when utilizing the CPU map, since kernel-specific and node-specific arguments no longer have to be supplied at compile-time.

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -981,6 +981,15 @@ func TestPerfEventArray(t *testing.T) {
 	}
 }
 
+func TestCPUMap(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "4.15", "cpu map")
+
+	m, err := NewMap(&MapSpec{Type: CPUMap, KeySize: 4, ValueSize: 4})
+	qt.Assert(t, qt.IsNil(err))
+	qt.Assert(t, qt.Equals(m.MaxEntries(), uint32(MustPossibleCPU())))
+	m.Close()
+}
+
 func createMapInMap(t *testing.T, typ MapType) *Map {
 	t.Helper()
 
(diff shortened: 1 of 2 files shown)


map_test.go
```diff
@@ -21,16 +21,14 @@ import (
 	"github.com/go-quicktest/qt"
 )
 
-var (
```
ti-mo, line 24: No arbitrary style changes, please.

ti-mo (changes requested):
Only the PossibleCPU() changes, please.

The key/val change doesn't make much sense IMO.

- Changing the key or value size without the knowledge of the bpf program will result in the program potentially reading past the end of the value on kernels where the value is 4 bytes, resulting in a verifier error. In order to remain portable, an application (the Cilium agent, in this case) needs to target the lowest common denominator of any given feature. When support for older kernels is dropped, the 8-byte value can be used (should be fine for Cilium since we target 5.4+), unless the program is compiled specifically for the running kernel (which we're trying to get rid of in Cilium).

- Kernel versions are meaningless in bpf land; vendors (mostly RH) backport tons of features to older kernels, while disabling others for security reasons. Reacting to the availability of certain features must be done using probes, e.g. in this case by trying to create a map of a particular size. Tests are the only exception, since we use only mainline kernels and need some way of defining when a test needs to pass and when it can be skipped.

Also, for our purposes, a simple map creation test will do with a check on Map.MaxEntries(). Don't overthink it. :wink:

map_test.go
learnitall: My apologies. Turns out my editor was using `gofumpt` instead of `gofmt` which caused changes like this to sneak in.

learnitall:
Done deal @ti-mo, thanks for the explanation!

ti-mo (approved):
Thanks!

PR #1634, opened 2024-12-19
ci: reduce bpf2go and gentypes verbosity, mount gomodcache when using podman

ti-mo:
- Improve podman QoL by mounting the user's go module cache.
- Reduce verbosity of gentypes
- Reduce verbosity of bpf2go behind -debug flag or `V=1` env

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,10 @@ UIDGID := $(shell stat -c '%u:%g' ${REPODIR})
 # Prefer podman if installed, otherwise use docker.
 # Note: Setting the var at runtime will always override.
 CONTAINER_ENGINE ?= $(if $(shell command -v podman), podman, docker)
-CONTAINER_RUN_ARGS ?= $(if $(filter ${CONTAINER_ENGINE}, podman), --log-driver=none, --user "${UIDGID}")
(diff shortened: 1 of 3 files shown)


lmb:
Ack on the podman changes.

Verbose `gentypes` is useful because it gives context which type failed when generating, try changing one of the patches to return an error and you'll see what I mean. I think it'd be nice to keep that, outputting a number doesn't really tell me anything useful imo. When is it too verbose for you?

I'm neutral on the bpf2go change. Does the tool print anything now by default? I always ask myself whether things worked when there is no output at all.

ti-mo:
> Verbose `gentypes` is useful because it gives context which type failed when generating, try changing one of the patches to return an error and you'll see what I mean. 

I'll take another look and maybe add a verbose flag like bpf2go has now.

> outputting a number doesn't really tell me anything useful imo

> Does the tool [bpf2go] print anything now by default? I always ask myself whether things worked when there is no output at all.

I find these statements a bit contradictory. :stuck_out_tongue: The number being non-zero means it worked and the tool did something. Over time we get a sense of how many types we'd expect to see there, and we can catch things that are out of the ordinary.

PR #1636, opened 2024-12-20
bpf2go: ergonomic enums

mejedi:
Constant names emitted for enum elements end up quite unwieldy. They are
    prefixed with enum name because in C struct, union and enum names are in
    a separate namespace, allowing for overlaps with identifiers, e.g:

    enum E { E };

While such overlaps are possible in *theory*, people usually don't do
    it. If a typicall C naming convention is followed, we get

    enum something {
        SOMETHING_FOO, SOMETHING_BAR
    };

 generating `<STEM>SomethingSOMETHING_FOO` and `<STEM>SomethingSOMETHING_BAR`.

In addition to "safe" long names, generate shorter ones if the
    respective name is not taken. `<STEM>SOMETHING_FOO` and
    `<STEM>SOMETHING_BAR` are much nicer to work with.

### DIFF
--- a/cmd/bpf2go/gen/output.go
+++ b/cmd/bpf2go/gen/output.go
@@ -173,6 +173,17 @@ func Generate(args GenerateArgs) error {
 	gf := &btf.GoFormatter{
 		Names:      nameByType,
 		Identifier: args.Identifier,
+		ShortEnumIdentifier: func(_, element string) string {
+			elementName := args.Stem + args.Identifier(element)
+			if _, nameTaken := typeByName[elementName]; nameTaken {
+				return ""
+			}
+			if _, nameReserved := reservedNames[elementName]; nameReserved {
+				return ""
+			}
+			reservedNames[elementName] = struct{}{}
+			return elementName
+		},
 	}
 
 	ctx := struct {
@@ -201,7 +212,7 @@ func Generate(args GenerateArgs) error {
 
 	var buf bytes.Buffer
 	if err := commonTemplate.Execute(&buf, &ctx); err != nil {
-		return fmt.Errorf("can't generate types: %s", err)
+		return fmt.Errorf("can't generate types: %v", err)
 	}
 
 	return internal.WriteFormatted(buf.Bytes(), args.Output)
--- a/cmd/bpf2go/gen/output.go
+++ b/cmd/bpf2go/gen/output.go
@@ -7,7 +7,6 @@ import (
 	"go/build/constraint"
 	"go/token"
 	"io"
-	"sort"
 	"strings"
 	"text/template"
 	"unicode"
@@ -141,21 +140,38 @@ func Generate(args GenerateArgs) error {
 		programs[name] = args.Identifier(name)
 	}
 
-	typeNames := make(map[btf.Type]string)
-	for _, typ := range args.Types {
-		// NB: This also deduplicates types.
-		typeNames[typ] = args.Stem + args.Identifier(typ.TypeName())
+	tn := templateName(args.Stem)
+	reservedNames := map[string]struct{}{
+		tn.Specs():         {},
+		tn.MapSpecs():      {},
+		tn.ProgramSpecs():  {},
+		tn.VariableSpecs(): {},
+		tn.Objects():       {},
+		tn.Maps():          {},
+		tn.Programs():      {},
+		tn.Variables():     {},
 	}
 
-	// Ensure we don't have conflicting names and generate a sorted list of
-	// named types so that the output is stable.
-	types, err := sortTypes(typeNames)
-	if err != nil {
-		return err
+	typeByName := map[string]btf.Type{}
+	nameByType := map[btf.Type]string{}
+	for _, typ := range args.Types {
+		// NB: This also deduplicates types.
+		name := args.Stem + args.Identifier(typ.TypeName())
+		if _, reserved := reservedNames[name]; reserved {
+			return fmt.Errorf("type name %q is reserved", name)
+		}
+		if otherType, ok := typeByName[name]; ok {
+			if otherType == typ {
+				continue
+			}
+			return fmt.Errorf("type name %q is used multiple times", name)
+		}
+		typeByName[name] = typ
+		nameByType[typ] = name
 	}
 
 	gf := &btf.GoFormatter{
-		Names:      typeNames,
+		Names:      nameByType,
 		Identifier: args.Identifier,
 	}
 
@@ -168,8 +184,7 @@ func Generate(args GenerateArgs) error {
 		Maps        map[string]string
 		Variables   map[string]string
 		Programs    map[string]string
-		Types       []btf.Type
-		TypeNames   map[btf.Type]string
+		Types       map[string]btf.Type
 		File        string
 	}{
 		gf,
@@ -180,8 +195,7 @@ func Generate(args GenerateArgs) error {
 		maps,
 		variables,
 		programs,
-		types,
-		typeNames,
+		typeByName,
 		args.ObjectFile,
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/bpf2go/gen/output.go
```diff
@@ -141,22 +140,50 @@ func Generate(args GenerateArgs) error {
...
+		}
+		typeByName[name] = typ
+		nameByType[typ] = name
 	}
 
 	gf := &btf.GoFormatter{
-		Names:      typeNames,
+		Names:      nameByType,
 		Identifier: args.Identifier,
+		ShortEnumIdentifier: func(_, element string) string {
```
ti-mo, line 176: Could this be made a part of `EnumIdentifier` instead? Not sure why this deserves its own hook, it's just another step during enum identifier generation.

cmd/bpf2go/gen/output_test.go
```diff
@@ -12,58 +13,6 @@ import (
 	"github.com/cilium/ebpf/cmd/bpf2go/internal"
 )
 
-func TestOrderTypes(t *testing.T) {
```
ti-mo, line 15: Basically same question as above. Do we have another way of ensuring stability in the generated output? Was this always unnecessary?

cmd/bpf2go/gen/output.go
```diff
@@ -141,22 +140,50 @@ func Generate(args GenerateArgs) error {
 		programs[name] = args.Identifier(name)
 	}
 
-	typeNames := make(map[btf.Type]string)
-	for _, typ := range args.Types {
-		// NB: This also deduplicates types.
-		typeNames[typ] = args.Stem + args.Identifier(typ.TypeName())
+	tn := templateName(args.Stem)
+	reservedNames := map[string]struct{}{
```
ti-mo, line 144: At first sight, I'd model this as a property to be specified during construction of the GoFormatter. Everytime the formatter generates an identifier (struct or otherwise), it can check uniqueness and push it to the set of seen names. I think it's a good feature to have in the formatter to prevent it from emitting invalid code, it doesn't need to be b2g-specific.

This way, you also avoid having to close over `reservedNames` and `typeByName` to get it into an `ShortEnumIdentifier` function.

ti-mo (changes requested):
While yes, this makes for shorter identifiers, it also makes it less deterministic who will get the prefix-less one in case of conflicts. The first-declared one? The one appearing first alphabetically? Is it random? This needs to be clearly defined.

I also kind of don't really see the problem, could you give a real-world example of an identifier that gets unwieldy because of this? There's a case to be made for anonymous enums, which I'm not sure we currently support; at least I couldn't find any examples.

I'm also not sure about adding GoFormatter.ShortEnumIdentifier, we already have EnumIdentifier that could probably be extended.

cmd/bpf2go/gen/output_test.go
```diff
@@ -116,3 +65,46 @@ func TestObjects(t *testing.T) {
...
+			if conflict {
+				qt.Assert(t, qt.Not(pred))
+			} else {
+				qt.Assert(t, pred)
+			}
+		})
+	}
+}
+
+func wsSeparated(terms ...string) *regexp.Regexp {
```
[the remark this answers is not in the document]
mejedi: >  is the amount of whitespace not deterministic?
The amount of whitespace depends on the length of other identifiers in the set since formatter inserts extra spaces for alignment. It doesn't improve readability indeed, but makes it easier to update test code in the future.

cmd/bpf2go/gen/output.go
```diff
@@ -180,44 +206,18 @@ func Generate(args GenerateArgs) error {
...
-		types,
-		typeNames,
+		typeByName,
 		args.ObjectFile,
 	}
 
 	var buf bytes.Buffer
 	if err := commonTemplate.Execute(&buf, &ctx); err != nil {
-		return fmt.Errorf("can't generate types: %s", err)
+		return fmt.Errorf("can't generate types: %v", err)
```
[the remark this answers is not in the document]
mejedi: I actually meant [%w](https://pkg.go.dev/fmt#Errorf).

cmd/bpf2go/gen/output.go
mejedi: >  Not sure why this deserves its own hook, it's just another step during enum identifier generation.

I propose generating BOTH long and short identifiers to maintain backwards compatibility.

cmd/bpf2go/gen/output.go
```diff
@@ -180,44 +206,18 @@ func Generate(args GenerateArgs) error {
...
 	}
 
 	return internal.WriteFormatted(buf.Bytes(), args.Output)
 }
 
-// sortTypes returns a list of types sorted by their (generated) Go type name.
-//
-// Duplicate Go type names are rejected.
-func sortTypes(typeNames map[btf.Type]string) ([]btf.Type, error) {
```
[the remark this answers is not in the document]
mejedi: https://pkg.go.dev/text/template#hdr-Actions
> {{range pipeline}} T1 {{end}}
	The value of the pipeline must be an array, slice, map, or channel.
	If the value of the pipeline has length zero, nothing is output;
	otherwise, dot is set to the successive elements of the array,
	slice, or map and T1 is executed. If the value is a map and the
	keys are of basic type with a defined order, the elements will be
	visited in sorted key order.

cmd/bpf2go/gen/output_test.go
mejedi: When we `range` over a map in `text/template`, items are first ordered by key.

mejedi:
@ti-mo 
> I also kind of don't really see the problem, could you give a real-world example of an identifier that gets unwieldy because of this?

```c
enum reassm_rc {
  REASSM_RC_SUCCESS,
  REASSM_RC_INTERNAL_ERROR,
  REASSM_RC_TOO_MANY_REASSEMBLIES,
  ...
};
```
yields
```go
const (
    frReassmRcREASSM_RC_SUCCESS               frReassmRc = 0
    frReassmRcREASSM_RC_INTERNAL_ERROR        frReassmRc = 1
    frReassmRcREASSM_RC_TOO_MANY_REASSEMBLIES frReassmRc = 2
    ....
```
This code is meant to be reused across several project, therefore I can't simply have `SUCCESS`, `INTERNAL_ERROR` and `TOO_MANY_REASSEMBLIES`.

mejedi:
@ti-mo 
> it also makes it less deterministic who will get the prefix-less one in case of conflicts. The first-declared one? The one appearing first alphabetically? Is it random? This needs to be clearly defined.

Let's talk about conflicts. Assuming a C compiler, the only conflict possible is with a type name. E.g.
```c
struct foo {};
enum E {foo};
```
Enum item name MUST be unique. The following program fails to compile:
```c
enum A {foo};
enum B {foo}; // error: redefinition of enumerator 'foo'
```
In case of conflict, a type name wins.

<hr>

In theory, the input object file could have duplicate enum identifiers. It is possible with non-C languages or multiple C compilation units linked together with `bpftool`. It is a niche use case as of today.

Types are visited in alphabetical order, items in a enum are visited in BTF order which typically matches declaration order.

ti-mo:
@mejedi Thanks for your patience! As we've discussed offline, I'm reluctant to accept the patch as-is since it introduces subtle behaviour that makes things less deterministic and could be surprising for the user. In general, we want to keep changes minimally invasive.

I suggest we head towards a model where you can introduce all the hooks you need to implement niche logic like this in in-house tooling. The GoFormatter already has a few of these points, and you're introducing `ShortEnumIdentifier` here, so I'm wondering if you could submit another proposal that allows you to implement and maintain the logic you want on your side. bpf2go is also similarly (but not completely) modular to allow consuming it as a library, but I have no idea what's missing there.

Please start a discussion in #ebpf-go-dev to get this moving, that will get a bit more eyes on it than just reviewers.

PR #1640, opened 2025-01-06
fix kernelModule AttachTo condition

brycekahle:
It should've been `ps.AttachTo != ""`, but `targetsKernelModule` already checks that condition, so just remove it completely.


ti-mo (changes requested):
Thanks! If this made it through, that means test coverage is lacking. Maybe now's a good opportunity to add one? You can assume bpf_testmod is loaded in CI if that helps.

brycekahle:
> You can assume bpf_testmod is loaded in CI

I think that is only available on 5.11+?

brycekahle:
@ti-mo added some tests

ti-mo (approved):

brycekahle:
@ti-mo any chance of getting a minor release for this fix? I don't think kernel module programs will work without it.

ti-mo:
> @ti-mo any chance of getting a minor release for this fix? I don't think kernel module programs will work without it.

@brycekahle Yep, aiming for Mon/Tues.

PR #1641, opened 2025-01-08
features: make HaveProgramHelper more accurate

lmb:
Parse the verifier output to figure out whether a helper call is really not supported. This makes the result more reliable and allows us to get rid of asm.BuiltinFunc.Max().


features/prog.go
```diff
@@ -267,30 +263,65 @@ func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
...
-		// TODO: possibly we need to check verifier output here to be sure
-		err = ebpf.ErrNotSupported
+		// https://github.com/torvalds/linux/blob/09a0fa92e5b45e99cf435b2fbf5ebcf889cf8780/kernel/bpf/verifier.c#L10663
+		if logContainsAll(verr.Log, "invalid func", helperTag) {
+			return ebpf.ErrNotSupported
+		}
+
+		// https://github.com/torvalds/linux/blob/09a0fa92e5b45e99cf435b2fbf5ebcf889cf8780/kernel/bpf/verifier.c#L10668
+		if logContainsAll(verr.Log, "program of this type cannot use helper", helperTag) {
+			return ebpf.ErrNotSupported
```
florianl, line 296: I think it would help users to distinguish between the different kinds of `ErrNotSupported`. So maybe extend it with something like this?
```suggestion
			return fmt.Errorf("program of this type cannot use helper: %w", ebpf.ErrNotSupported)
```

features/prog.go
```diff
@@ -267,30 +263,63 @@ func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
...
+		wrongProgramType = wrongProgramType || logContainsAll(verr.Log, "unknown func")
+		if wrongProgramType {
+			return fmt.Errorf("program of this type cannot use helper: %w", ebpf.ErrNotSupported)
+		}
 	}
 
 	return err
 }
 
+func logContainsAll(log []string, needles ...string) bool {
```
rgo3, line 307: Nit, as it should be functionally the same but removes need for the goto label
```suggestion
func logContainsAll(log []string, needles ...string) bool {
	first := max(len(log)-5, 0) // Check last 5 lines.
	return slices.ContainsFunc(log[first:], func(line string) bool {
		for _, needle := range needles {
			if !strings.Contains(line, needle) {
				return false
			}
		}
		return true
	})
}
```

rgo3 (approved):
Cool, I think we thought about doing this since the beginning but never deemed it necessary. Did you find an instance where just checking EINVAL results in a false-negative?

lmb:
> Did you find an instance where just checking EINVAL results in a false-negative?

No, the motivation is to get rid of `Max()` for the Windows work.

florianl (approved):

PR #1646, opened 2025-01-14
ebpf: encode platform in type constants

lmb:
ebpf: encode platform in type constants

    Linux and Windows share the concept of program types, map types and so on. 
    For example, XDP is supported on both platforms. Unfortunately the constant 
    values used by both platforms are different. On Linux, XDP has value 0x6 
    while on Windows it is 0x1.

    This problem extends to a CollectionSpec parsed from an ELF. Because of the 
    different values a Windows PerCPUHash aliases with a Linux ProgramArray. 
    This causes all sorts of mayhem that is hard to avoid without making too 
    many changes to the code base.

    Introduce the notion of a supported platform. External users only see this
    as a string, like "linux". Internally each platform has a corresponding tag
    which is used to encode the platform into a constant. This means that there
    are distinct constants for each platform. Instead of overloading the XDP
    constant we will introduce WindowsXDP. This will allow the ELF reader to
    remain platform independent and makes generating stringers a lot easier.

    The value of the platform tag is chosen so that on Linux the constant values
    do not change and are identical with the value in upstream headers. On
    Windows the constants have a fixed (but implementation defined) offset.

    The downside of this approach is that Windows constants will all require a
    prefix of some sort to distinguish them from Linux ones. This is probably 
    most onerous for map types, because those tend to be created manually more 
    frequently than programs, for example. We can add some behind the scenes 
    translation of Linux map types to Windows ones if need be.

    Another downside is that constant values are now limited to 28 bits since we
    steal the top 4 bits to store the platform. The constants we're applying 
    this to are all sequentially numbered from 0, so this is hopefully enough.


asm/instruction.go
```diff
@@ -43,9 +44,18 @@ type Instruction struct {
...
 func (ins *Instruction) Unmarshal(r io.Reader, bo binary.ByteOrder) (uint64, error) {
+	err := ins.UnmarshalForPlatform(r, bo, internal.NativePlatform)
+	if err != nil {
+		return 0, err
+	}
+
+	return ins.Size(), nil
+}
+
+func (ins *Instruction) UnmarshalForPlatform(r io.Reader, bo binary.ByteOrder, plat internal.Platform) error {
```
ti-mo, line 55: This can be moved to the root package if need be.

lmb:
@ti-mo I ended up using `string` instead of a new dedicated type. Usage within the library doesn't really change much. The upside is that we don't need to fiddle around with multiple package to hide `LinuxTag`. Downside is that the code is theoretically slower. I think it doesn't make a difference in practice.

lmb:
Making this a draft again, I discovered that adding Windows CI is pretty straight forward. https://github.com/cilium/ebpf/pull/1650 is a better first PR.

PR #1647, opened 2025-01-20
linker: do not copy instructions when there is no reference in `flattenInstructions`

paulcacheux:
When resolving the bpf-to-bpf calls we need to copy the instructions from the program to flatten the references.

But when there is no reference there is nothing to do so we can skip the copy operation altogether.


lmb (dismissed):
Could you add a test using testing.AllocsPerRun to make sure this doesn't regress?

ti-mo (approved):
Thanks!

PR #1651, opened 2025-01-21
pin: fix a bunch of fd leaks in tests

lmb:
The new pin package didn't import fdtrace.TestMain until now, which is why the leaks went by unnoticed.

### DIFF
--- a/pin/load.go
+++ b/pin/load.go
@@ -2,6 +2,7 @@ package pin
 
 import (
 	"fmt"
+	"io"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal/sys"
@@ -11,6 +12,7 @@ import (
 // Pinner is an interface implemented by all eBPF objects that support pinning
 // to a bpf virtual filesystem.
 type Pinner interface {
+	io.Closer
 	Pin(string) error
 }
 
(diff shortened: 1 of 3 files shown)


lmb:
It's very easy to accidentally leak objects when using WalkDir, and kind of awkward to do the right thing (because obj can be nil?) This is why https://pkg.go.dev/github.com/cilium/ebpf/btf#HandleIterator.Take exists for handles. It might be good to do another pass over this API.

cc @ti-mo @mtardy @dylandreimerink @rgo3

mtardy:
Mmmh I'm slightly confused now, this it the typical use we have in Tetragon:
- https://github.com/cilium/tetragon/blob/d72f98aefb8e88c06caec7cfbc687bbeabf97a5f/cmd/tetra/debug/progs.go#L408-L445
- https://github.com/cilium/tetragon/blob/d72f98aefb8e88c06caec7cfbc687bbeabf97a5f/pkg/bugtool/maps.go#L148-L176

Maybe having an example in the doc would make thing more clear?

lmb:
Timo pointed out that an issue is probably a better place for this: https://github.com/cilium/ebpf/issues/1652

ti-mo:
I knew I forgot something. Thanks for the patch!

PR #1659, opened 2025-01-27
internal/unix: add Errno wrapper for Windows

lmb:
eBPF for Windows has historically aimed to be source compatible with libbpf API. As part of this, POSIX error codes are found in errno.h are used to indicate errors of all kinds.

eBPF for Windows so far aimed to be source compatible with libbpf. This includes using the same error constants (EINVAL, etc.) as libbpf. It does this using the errno.h header supplied by the Microsoft C runtime. Unfortunately, Windows error codes do not always use the same value as Linux does.

Introduce a dedicated Errno type on Windows which maps errors to the correct constant for the platform. This means that syscall.Errno is not an alias for unix.Errno any more.


internal/unix/errno_windows.go
```diff
@@ -0,0 +1,78 @@
...
+		return e == EEXIST || e == ENOTEMPTY
+	case os.ErrNotExist:
+		return e == ENOENT
+	case errors.ErrUnsupported:
+		return e == ENOSYS || e == ENOTSUP || e == EOPNOTSUPP
+	}
+	return false
+}
+
+func (e Errno) Temporary() bool {
```
ti-mo, line 72: Are you using these in other Windows-specific code? I don't think an external caller can get ahold of these methods since this is `internal/unix.Errno`. Unless we need them to satisfy some interface, I propose we remove them.

ti-mo (approved):
Looks good, one nit.

internal/unix/errno_windows.go
lmb: We expose these indirectly because we wrap original errno from `bpf()` and friends. It'd be surprising / a footgun if those errors behaved different between platforms, I think.

PR #1665, opened 2025-02-04
internal/linux: use `unix.Auxv`

tklauser:
Use the newly added `Auxv` wrapper added in `golang.org/x/sys/unix` v0.30.0 instead of manually go:linknaming `runtime.getAuxv`. This also allows to get rid of the check for malformed auxv data as it is already performed by `unix.Auxv`.


tklauser:
Sorry, opened this a bit too quickly. Will reopen once this is ready.

ti-mo (approved):
Cool, thank you! `[][2]uintptr` also makes a lot more sense, indeed. :)

PR #1666, opened 2025-02-04
windows: testutils changes

lmb:
internal: add IsWindows constant

    Add an IsWindows constant and move the existing OnLinux constant to a new
    internal package. This avoids a circular dependency between testutils.

internal/fdtrace: rename to testmain and add windows trace logging

    The eBPF for Windows runtime logs traces and error messages to the Windows
    event subsystem. Being able to correlate test failures with this trace is
    very useful to diagnose problems.

    On Windows, start an event tracing session before starting the tests, and
    dump a summary to stderr after tests have finished.

    The summary looks roughly like this:

          base       _get_map_descriptor_properties returned success
         entry-exit _ebpf_core_protocol_map_delete_element
       ! base       _ebpf_core_protocol_map_delete_element returned error  
    Error=19

    The exclamation mark in the first column indicates an error.

internal/testutils: make feature test helpers platform agnostic


internal/testutils/testmain/windows.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return ts.stop(os.DevNull)
+}
+
+func (ts *traceSession) stop(file string) error {
+	if ts.session == "" {
+		return nil
+	}
+
+	wpr := exec.Command("wpr.exe", "-stop", file, "-instancename", ts.session)
```
rgo3, line 95: Is the wpr.exe executable available by default on windows systems? There is at least one other shell-out in this PR, the same question applies there.

rgo3 (approved):
Just a couple of nits and questions, otherwise LGTM.

Can you refresh my memory on if we already have a guide to test windows changes as a Linux user? I would be intrigued to follow it once I find some cycles in the next couple of weeks.

internal/testutils/testmain/windows.go
lmb: Good question, I think `wpr.exe` does ship with Windows:
```
PS C:\Users\lorenz> (get-command wpr.exe).Path
C:\Windows\system32\wpr.exe
```

lmb: wpr.exe does come out of the box: https://learn.microsoft.com/en-us/windows-hardware/test/wpt/introduction-to-wpr#wpr-command-line-and-user-interface

> WPR.exe ships with Windows OS (Windows 8.1 or later) and you do not need additional installation.

Netsh seems to be the `ip` equivalent: https://learn.microsoft.com/en-us/windows-server/networking/technologies/netsh/netsh

```
PS C:\Users\lorenz> (get-command netsh.exe).Path
C:\Windows\system32\netsh.exe
```

lmb:
> Can you refresh my memory on if we already have a guide to test windows changes as a Linux user?

I have one written up on a local branch, but I don't want to merge it before I'm ready to pronounce basic Windows support to be there.

Issue #914, opened 2023-01-23
proposal: sys: make UAPI wrappers type safe(er)


lmb:
The BPF uapi uses a special type `__aligned_u64` to store pointers:

```C
	struct { /* anonymous struct used by BPF_OBJ_* commands */
		__aligned_u64	pathname;
		__u32		bpf_fd;
		__u32		file_flags;
	};
```

The reason for this is explained in types.h:

```
/*
 * aligned_u64 should be used in defining kernel<->userspace ABIs to avoid
 * common 32/64-bit compat problems.
 * 64-bit values align to 4-byte boundaries on x86_32 (and possibly other
 * architectures) and to 8-byte boundaries on 64-bit architectures.  The new
 * aligned_64 type enforces 8-byte alignment so that structs containing
 * aligned_64 values have the same alignment on 32-bit and 64-bit architectures.
 * No conversions are necessary between 32-bit user-space and a 64-bit kernel.
 */
#define __aligned_u64 __u64 __attribute__((aligned(8)))
```

This poses a problem for our uapi wrappers: pointers in Go vary in size depending on the platform. We need a way to emulate `__aligned_u64`. So far, we've used a wrapper around `unsafe.Pointer` with a per-platform definition.

```go
// on 64 bit platforms
type Pointer struct {
	ptr unsafe.Pointer
}

// on 32 bit little-endian platforms
type Pointer struct {
	ptr unsafe.Pointer
	pad uint32
}
```

This ensures that `Pointer` is always 64 bit, however it doesn't enforce that Pointer is aligned to 8 bytes (Go doesn't provide control over alignment). In addition, the generated uapi structs lack type information, the caller needs to know what to put where.

```go
type ProgLoadAttr struct {
        ...
	FuncInfo           Pointer
	FuncInfoCnt        uint32
	LineInfoRecSize    uint32
	LineInfo           Pointer
        ...
}
```

# Proposal: make sys.Pointer a generic type

When I originally came up with `Pointer` we didn't have generics available, so `unsafe.Pointer` was the simplest choice. We can do better now.

```go
type Pointer[T any] struct { ... }
```

`ProgLoadAttr` would now become:

```go
type ProgLoadAttr struct {
        ...
	FuncInfo           Pointer[FuncInfo]
        ...
}
```

We can do more neat things with this, for example introduce a pointer type for a C string:

```go
type StringPointer Pointer[byte]

// NewStringPointer allocates a null-terminated backing slice for str and returns
// a pointer to it.
[...]

PR #1670, opened 2025-02-06
sys: add type safe Pointer wrapper

lmb:
Add TypedPointer which is a type safe wrapper around Pointer. This relieves users of the generated bindings from knowing which types are expected. This should make it less likely that we cause a hard to debug panic or crash due to the kernel overwriting runtime memory.

sys.Pointer still exists because there are some cases where we legitimately don't know what type to point at, for example the link info APIs. Map APIs are also unchanged for now because they don't benefit much from converting to TypedPointer[byte].

### DIFF
--- a/internal/cmd/gentypes/main.go
+++ b/internal/cmd/gentypes/main.go
@@ -70,20 +70,46 @@ func run(args []string) error {
 
 func generateTypes(spec *btf.Spec) ([]byte, error) {
 	objName := &btf.Array{Nelems: 16, Type: &btf.Int{Encoding: btf.Char, Size: 1}}
+	mapID := &btf.Int{Size: 4}
+	progID := &btf.Int{Size: 4}
 	linkID := &btf.Int{Size: 4}
 	btfID := &btf.Int{Size: 4}
 	typeID := &btf.Int{Size: 4}
-	pointer := &btf.Int{Size: 8}
 	logLevel := &btf.Int{Size: 4}
+
+	rawPointer := &btf.Pointer{}
+	stringPointer := &btf.Pointer{}
+	stringSlicePointer := &btf.Pointer{}
+	bytePtr := &btf.Pointer{}
+	int32Ptr := &btf.Pointer{}
+	uint32Ptr := &btf.Pointer{}
+	uint64Ptr := &btf.Pointer{}
+	uintptrPointer := &btf.Pointer{}
+	mapIDPtr := &btf.Pointer{}
+	progIDPtr := &btf.Pointer{}
+	linkIDPtr := &btf.Pointer{}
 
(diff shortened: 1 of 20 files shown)


lmb:
The breaking change is due to replacing ebpf.MapID and ProgramID with a type alias.

kolyshkin:
I like the approach!

It will be better to separate the actual overrun fix from the new type system.

I've also updated the fix in #1660 to not break the existing API nor return truncated pointers, PTAL @lmb

dylandreimerink (approved):
I like this! More type safety more better

ti-mo:
@lmb Looks like this implements https://github.com/cilium/ebpf/issues/914?

internal/cmd/gentypes/main.go
```diff
@@ -70,19 +70,44 @@ func run(args []string) error {
...
+	mapID := &btf.Int{Size: 4}
+	progID := &btf.Int{Size: 4}
 	linkID := &btf.Int{Size: 4}
 	btfID := &btf.Int{Size: 4}
 	typeID := &btf.Int{Size: 4}
-	pointer := &btf.Int{Size: 8}
+	rawPointer := &btf.Int{Size: 8}
+	stringPointer := &btf.Int{Size: 8}
+	stringSlicePointer := &btf.Int{Size: 8}
+	bytePtr := &btf.Int{Size: 8}
```
ti-mo, line 81: Do the target type sizes need to be correct here? It's a bit confusing regardless.

ti-mo (dismissed):
Wouldn't this allow us to unexport `sys.Pointer`? Ie. if it needs to be used outside of the package, it needs to be wrapped in an (Unsafe/Typed/String/XYZ)Pointer. Then we make sure UAPI no longer takes any `sys.Pointer`s.

I still see a bunch of `sys.Pointer` uses in the map API; I'd argue those should all be `sys.UnsafePointer`s instead, to be converted to typed pointers before returning.

internal/cmd/gentypes/main.go
lmb: The size needs to be equivalent to the size of a pointer. Using `Pointer` is a bit cumbersome because it requires specifying a target. I'll add a comment / try to work around this somehow.

lmb:
> I still see a bunch of `sys.Pointer` uses in the map API; I'd argue those should all be `sys.UnsafePointer`s instead, to be converted to typed pointers before returning.

Discussed face to face: UnsafePointer is a constructor, not a type. We still need `sys.Pointer` to account for endianness differences on 32 bit arches.

internal/cmd/gentypes/main.go
lmb: Turns out that using Pointer works just fine an is a bit clearer.

PR #1673, opened 2025-02-08
btf: fix data race in `(*mutableTypes).copy`

paulcacheux:
Move the creation of the mutableTypes copy under the read lock, so that the access to the len of the source types is protected when pre-allocating the maps.

```
$ go test -exec "sudo -E" -race -run TestSpecConcurrentAccess ./btf
==================
WARNING: DATA RACE
Read at 0x00c0016fd800 by goroutine 12:
  github.com/cilium/ebpf/btf.(*mutableTypes).copy()
      /home/vagrant/Projects/ebpf/btf/btf.go:105 +0x58
  github.com/cilium/ebpf/btf.(*Spec).Copy()
      /home/vagrant/Projects/ebpf/btf/btf.go:540 +0xdc
  github.com/cilium/ebpf/btf.TestSpecConcurrentAccess.func1()
      /home/vagrant/Projects/ebpf/btf/btf_test.go:542 +0x120

Previous write at 0x00c0016fd800 by goroutine 13:
  runtime.mapaccessK()
      /usr/local/go/src/runtime/map.go:519 +0x1ec
  github.com/cilium/ebpf/btf.copyType()
      /home/vagrant/Projects/ebpf/btf/types.go:737 +0xb0
  github.com/cilium/ebpf/btf.copyType.func1()
      /home/vagrant/Projects/ebpf/btf/types.go:744 +0x64
  github.com/cilium/ebpf/btf.children()
      /home/vagrant/Projects/ebpf/btf/traversal.go:126 +0xea0
  github.com/cilium/ebpf/btf.copyType()
      /home/vagrant/Projects/ebpf/btf/types.go:743 +0x1c0
  github.com/cilium/ebpf/btf.copyType.func1()
      /home/vagrant/Projects/ebpf/btf/types.go:744 +0x64
  github.com/cilium/ebpf/btf.children()
      /home/vagrant/Projects/ebpf/btf/traversal.go:104 +0x4d4
  github.com/cilium/ebpf/btf.copyType()
      /home/vagrant/Projects/ebpf/btf/types.go:743 +0x1c0
  github.com/cilium/ebpf/btf.(*mutableTypes).add()
      /home/vagrant/Projects/ebpf/btf/btf.go:93 +0x134
  github.com/cilium/ebpf/btf.(*mutableTypes).anyTypesByName()
      /home/vagrant/Projects/ebpf/btf/btf.go:167 +0x34c
  github.com/cilium/ebpf/btf.(*Spec).AnyTypesByName()
[...]

PR #1674, opened 2025-02-11
ci: add test run with race detector

paulcacheux:
As discussed in https://github.com/cilium/ebpf/pull/1673, this PR adds a test run with the race detector enabled.

In the thread it was discussed to re-use the "previous go version" run and to add `-race` to it, but it's not possible because the race detector requires CGo, this makes the test binary not static anymore, and vimto is having a hard time running it.

This PR choses the easy path and just creates a new test run closer to the arm64 one (no vimto)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,47 @@ jobs:
           name: Test Results (previous stable Go)
           path: junit.xml
 
+  test-race:
+    name: Run tests with race detector
+    runs-on: ubuntu-latest
+    needs: build-and-lint
+    timeout-minutes: 15
+    env:
+      CI_KERNEL_SELFTESTS: '/usr/src/linux/tools/testing/selftests/bpf'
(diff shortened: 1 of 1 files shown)


ti-mo (approved):
Thanks! Not sure why this and the gotestsum PR were drafts. :wink: I gave this a rebase and fixed up the gotestsum version.

ti-mo:
Looks like running with the race detector enabled crashes the GH runner (?). The logs aren't showing, but they did stream in when I started the job, and RoundtripVMLinux was the last test it was able to start.

Locally, this works fine, but there may be some additional constraints on these runners.

ti-mo:
Looks like this doesn't work on GHA, closing for now. @paulcacheux Feel free to reopen if you want to pursue this further.

ti-mo:
This may be an issue with the 24.04 kernel crashing. Let's run -race tests in vimto as well.

lmb:
Superseded by https://github.com/cilium/ebpf/pull/1770

PR #1677, opened 2025-02-12
ebpf: implement PossibleCPU on Windows

lmb:
Move the Linux implementation into internal/linux and add a wrapper around the Windows GetMaximumProcessorCount function.


ti-mo:
Thanks!

PR #1678, opened 2025-02-12
btf: disable handle API on windows

lmb:
Windows currently has no / limited BTF support. Disable the API for now.


btf/core_reloc_test.go
```diff
@@ -120,15 +120,14 @@ func TestLD64IMMReloc(t *testing.T) {
...
 	coll, err := ebpf.NewCollection(spec)
+	testutils.SkipIfNotSupportedOnOS(t, err)
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer coll.Close()
 }
 
 func TestCOREPoisonLineInfo(t *testing.T) {
-	testutils.SkipOnOldKernel(t, "5.0", "program ext_infos")
```
dylandreimerink, line 130: I don't understand why this can be removed. We still have <5.0 kernels in the test matrix. Has this to do with the `FORCE_BTF` bit you added?

dylandreimerink (approved):
Changes in general seems good. Just something I don't fully understand.

btf/core_reloc_test.go
lmb: Yep. Now that we force BTF the SkipIfNotSupported below takes care of this.

PR #1679, opened 2025-02-12
allow verifier log size to reach max to prevent infinite loop

brycekahle:
In the case of:
- needing the max verifier log size
- kernel does not support `log_true_size` yet

https://github.com/cilium/ebpf/blob/29fedc5e2401a1edb9cb89753aaa72592ea0996e/prog.go#L453-L455

maxVerifierLogSize is `1073741823`, so integer divided by 2 = `536870911`. Multiplying by 2 equals `1073741822`. Thus we never reach `maxVerifierLogSize` in the loop. The kernel keeps returning `ENOSPC` but the size never changes and you get an infinite loop.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -51,11 +51,6 @@ const (
 // verifier log.
 const minVerifierLogSize = 64 * 1024
 
-// maxVerifierLogSize is the maximum size of verifier log buffer the kernel
-// will accept before returning EINVAL. May be increased to MaxUint32 in the
-// future, but avoid the unnecessary EINVAL for now.
-const maxVerifierLogSize = math.MaxUint32 >> 2
-
 // ProgramOptions control loading a program into the kernel.
 type ProgramOptions struct {
 	// Bitmap controlling the detail emitted by the kernel's eBPF verifier log.
@@ -411,7 +406,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 
 	// The caller requested a specific verifier log level. Set up the log buffer
 	// so that there is a chance of loading the program in a single shot.
-	logSize := internal.Between(opts.LogSizeStart, minVerifierLogSize, maxVerifierLogSize)
+	logSize := internal.Between(opts.LogSizeStart, minVerifierLogSize, math.MaxUint32)
 	var logBuf []byte
 	if !opts.LogDisabled && opts.LogLevel != 0 {
 		logBuf = make([]byte, logSize)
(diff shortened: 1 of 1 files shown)


brycekahle:
If it wasn't clear, we are seeing the infinite loop in one of our tools.

prog.go
```diff
@@ -450,9 +450,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 
 		// Make an educated guess how large the buffer should be by multiplying.
 		// Ensure the size doesn't overflow.
-		const factor = 2
-		logSize = internal.Between(logSize, minVerifierLogSize, maxVerifierLogSize/factor)
-		logSize *= factor
+		logSize = internal.Between(logSize*2, minVerifierLogSize, maxVerifierLogSize)
```
lmb, line 453: Can you bring back the overflow check that used to be here?

https://github.com/cilium/ebpf/commit/e8b05c5#diff-348ab86651d6b4babd4e62cbe504d85db37f9aea172e2b72a570b8d5d73b8bc5L443-L445

brycekahle: I don't think it is possible to overflow? `maxVerifierLogSize` is `2^30`, so `2^31` is the max it can be, which fits in a `uint32`.

lmb: Good point. There are two problems with the current code:

- The maxVerifierLogSize may change at some point, as the comment points out. It'd be nice to not have this problem again.
- There is no protection against an infinite loop, as you've discovered.

As I wrote the code originally the intention was that:

1. Passing a too large number for the buffer would give EINVAL, aborting the loop.
2. The buffer exceeding uint32 would also abort it.

(1) doesn't happen anymore since we clamp to the maximum, (2) was removed. IMO I'd prefer bringing back the original code, which doesn't clamp to a max value (because hardcoding other system's limit makes code brittle) and checks for overflow. We can keep the LogSizeStart logic, although that might need adjustment because with `LogLevel == 0` we actually use `LogSizeStart * 2` for the first call.

brycekahle: Updated, please take a look.

brycekahle: Actually, I thought about this some more. Having the knowledge about the kernel max allows us to work up to that value before erroring with `EINVAL`. If we just keep doubling, then we could blow right by a value that would actually function.

brycekahle: Maybe we should do that only for older kernels where `log_true_size` is not available?

ti-mo:
Ugh, the amount of time I've spent on these 30 lines of code over the years is getting ridiculous.. Discussed this offline with Lorenz, we came up with the following steps:

- Roll back the `LogSizeStart` field (again..) in favor of a `LogSizeOverride` field that disables the retry logic and does a one-shot load. Downside is that we don't have VerifierError.Truncated anymore, so it's not as clear for the user. We need an escape hatch anyway, though.
- Extract the retry decision-making logic into a function that can be tested independently without making the buffer allocations to catch silly mistakes like this rounding error.
- Limit the loop to 32 retries (1 for each bit in the LogSize field) to prevent this from going nuclear.

That should, hopefully, make this a bit more robust. I'll come up with something this week.

brycekahle:
> I'm not sure if reaching the max log size is important here

I believe it is, because if you use the max size the kernel will not error, even if it actually needed more space.

ti-mo:
> > I'm not sure if reaching the max log size is important here
> 
> I believe it is, because if you use the max size the kernel will not error, even if it actually needed more space.

Hmm, interesting. It's a shame this is so hard to test due to the memory requirements and having to produce a program large and complex enough to trigger the limit.. Anyway, I tried covering as many cases as possible in my PR, please give it a try.

PR #1680, opened 2025-02-13
elf: do not remove dots from names

lmb:
elf: do not remove dots from names

    The ELF reader currently strips dots from object names if the kernel doesn't
    support them. This is against the goal that the ELF reader should produce
    the same result no matter which environment it is run in.

    Change the setup around slightly, so that the ELF reader now always allows
    dots in names. Instead we strip dots if necessary when making the syscalls.

    This will break users which:

    * use object names with dots in them
    * on kernels which don't support dots
    * which are pinned into bppfs

Revert "features: tolerate EPERM in haveObjName and objNameAllowsDot"

    This reverts commit cfd8515687f75ac9e95dcfe75db1e3ef0d8f6601.

    Now that the ELF reader doesn't rely on feature tests anymore we can undo
    the change to treat EPERM specially.


lmb:
@ti-mo any concerns about this one?

florianl (approved):
Not Timo here but this would be a great change :+1:

ti-mo:
My main concern is we should've thought this through before we merged the initial version. :stuck_out_tongue: Also, I'd really like to unexport `SanitizeName`; it should've been an implementation detail, but I understand why we provided it.

Also, I found this: https://sourcegraph.com/github.com/cloudflare/xdpcap/-/blob/hook.go?L23-25. We need to give existing users a workaround if we can. Unfortunately, xdpcap substitutes with `_` instead of removing the character, which is different from our default. (IMO another reason to unexport the function)

I have a few Qs:

- Does bpffs also reject those unsupported characters in file names on older kernels?
- Could we extend the stripping behaviour to `Pin()` and `LoadPinned*()` to make it so the user doesn't need to care? (ignoring corner cases like breaking os.Remove for a second)

lmb:
As far as I know bpffs doesn't have restrictions on what files can be named (outside of the usual rules which apply to all filesystems). The only limitation is on map and program names as passed to the create syscalls. I think that rule came about because programs end up in /proc/kallsyms, but I'm not sure.

I think we shouldn't sanitize in Pin and LoadPinned since bpffs doesn't have these restrictions.

> We need to give existing users a workaround if we can. Unfortunately, xdpcap substitutes with _ instead of removing the character, which is different from our default.

You're saying: this change may break xdpcap because on old kernels `SanitizeName` won't replace `.` with `_`? That's a good point, in the case of xdpcap it's going to be fine I think. `Name` is set as a debug aid. On old kernels which don't allow dots, a name like `foo.bar` would become `foobar` instead of `foo_bar`.

> I'd really like to unexport `SanitizeName`;

Let's tackle that separately if need be?

PR #1681, opened 2025-02-13
sys: hint that structs require host layout

lmb:
Go 1.23 introduced structs.HostLayout to tell the toolchain that a struct should be laid out in memory as it is laid out in code. Add the annotation to the generated structs.


ti-mo (commented):
Looks good, thanks! We should probably also emit these in bpf2go types.

PR #1682, opened 2025-02-13
btf: use iterators

lmb:
Now that we're on 1.23 we can use iterators. The existing iterators don't really need to be modified. Allocations don't change, run time seems a little bit faster.

    core: 1
    goos: linux
    goarch: amd64
                                        │   base.txt   │              iter.txt              │
                                        │    sec/op    │    sec/op     vs base              │
    PostorderTraversal/single_type           49.79n ± ∞ ¹   48.68n ± ∞ ¹  -2.23% (p=0.029 n=4)
    PostorderTraversal/cycle(1)              146.4n ± ∞ ¹   135.8n ± ∞ ¹  -7.24% (p=0.029 n=4)
    PostorderTraversal/cycle(10)             1.928µ ± ∞ ¹   1.861µ ± ∞ ¹  -3.45% (p=0.029 n=4)
    PostorderTraversal/gov_update_cpu_data   1.895m ± ∞ ¹   1.907m ± ∞ ¹       ~ (p=0.057 n=4)
    geomean                                  2.271µ         2.201µ        -3.12%
    ¹ need >= 6 samples for confidence interval at level 0.95

### DIFF
--- a/btf/marshal.go
+++ b/btf/marshal.go
@@ -243,28 +243,27 @@ func (b *Builder) addString(str string) (uint32, error) {
 	return b.strings.Add(str)
 }
 
-func (e *encoder) allocateIDs(root Type) (err error) {
-	visitInPostorder(root, e.visited, func(typ Type) bool {
+func (e *encoder) allocateIDs(root Type) error {
+	for typ := range postorder(root, e.visited) {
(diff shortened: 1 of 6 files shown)


dylandreimerink (approved):
Nice! Neat to see how close the existing implementation already was

PR #1685, opened 2025-02-14
sys: add windows support

lmb:
More Windows pieces. Note that package ebpf, btf, link need follow up work.

efw: add wrappers for eBPF for Windows runtime

    Add basic infrastructure to call into the eBPF for Windows runtime. This
    uses runtime dynamic linking to call into ebpfapi.dll instead of relying on
    CGo. Ultimately this boils down to using syscall.SyscallN to call into
    specific functions in the DLL and looks very similar to what we do on Linux.

    The most important function is a wrapper for the bpf() function in 
    ebpfapi.dll. This is an ABI compatible re-implementation of the Linux bpf()
    syscall. Of course, not all of the syscall is implemented yet.

    In addition to that there are a variety of helper functions that do not have
    a bpf() equivalent and therefore need to be called directly.

sys: add windows support

    The sys package is at the core of the library, so make it work on Windows to
    minimise changes to the rest of the code base.

    The first important piece is a custom FD abstraction. On Windows, operating
    system resources are identified by handles. The differences are large enough
    that the efW runtime decided to wrap them in a file descriptor abstraction,
    most of which is provided by the Universal C Runtime on Windows. Change our
    FD type to call into the efW runtime.

    Second, Windows does not have the equivalent of bpffs. Instead objects are
    stored in a global table, keyed by a string. This means that we can't use
    file system APIs to manipulate pinning.

    Third, the bpf() syscall is emulated by calling into ebpfapi.dll. This is
    the key piece which allows most code to work unchanged.


internal/sys/fd.go
```diff
@@ -39,49 +38,21 @@ func (fd *FD) finalize() {
...
-	if value != 0 {
-		return fd, nil
-	}
-
-	dup, err := fd.Dup()
-	_ = fd.Close()
-	return dup, err
-}
-
-func (fd *FD) String() string {
```
lmb, line 62: Move back to common code.

internal/sys/fd_windows.go
```diff
@@ -0,0 +1,58 @@
...
+// NewFD wraps a raw fd with a finalizer.
+//
+// You must not use the raw fd after calling this function.
+func NewFD(value int) (*FD, error) {
+	if value == invalidFd {
+		return nil, fmt.Errorf("invalid fd %d", value)
+	}
+
+	if value == 0 {
+		// The bpf() syscall API can't deal with zero fds but we can't dup because
```
lmb, line 26: Update comment: efW runtime never emits 0-2 as fd.

internal/sys/fd_windows.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return NewFD(int(dup))
+}
+
+// File panics on Windows.
+func (fd *FD) File(name string) *os.File {
+	panic("FD.File is not implementable on Windows")
```
lmb, line 57: Make it return an error.

internal/sys/syscall_ext_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+	"unsafe"
+
+	"github.com/go-quicktest/qt"
+
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+func TestMap(t *testing.T) {
```
lmb, line 13: TestBPFSyscall and only make sure we can create an array.

internal/sys/ptr.go
```diff
@@ -29,12 +29,12 @@ func NewSlicePointerLen(buf []byte) (Pointer, uint32) {
...
 // NewStringPointer creates a 64-bit pointer from a string.
 func NewStringPointer(str string) Pointer {
-	p, err := unix.BytePtrFromString(str)
+	slice, err := unix.ByteSliceFromString(str)
 	if err != nil {
 		return Pointer{}
 	}
 
-	return Pointer{ptr: unsafe.Pointer(p)}
+	return Pointer{ptr: unsafe.Pointer(&slice[0])}
```
ti-mo, line 37: There's a 'supported' equivalent nowadays: `unsafe.SliceData(slice)`.

ti-mo (approved):
Thanks! Left a few nits.

internal/sys/ptr.go
lmb: This ends up being longer, because it doesn't return an unsafe Pointer: `unsafe.Pointer(unsafe.SliceData(slice))`

PR #1688, opened 2025-02-17
internal/testutils: make TempBPFFS portable

lmb:
Object pinning on Windows isn't backed by a file system, instead the path is just used as an identifier. Generate a random prefix to use for tests on Windows.

### DIFF
--- a/internal/testutils/bpffs_windows_test.go
+++ b/internal/testutils/bpffs_windows_test.go
@@ -1,0 +1,72 @@
+package testutils_test
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/testutils"
+	"github.com/go-quicktest/qt"
+)
+
+func TestTempBPFFS(t *testing.T) {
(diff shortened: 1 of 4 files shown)


internal/testutils/bpffs_windows_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/testutils"
+	"github.com/go-quicktest/qt"
+)
+
+func TestTempBPFFS(t *testing.T) {
```
ti-mo, line 16: I wonder if this test is really necessary; could we cover this with existing tests that run on both platforms? I get that this is more of a smoke test, but if it wouldn't work correctly, tests in the root package would probably also fail, especially given that t.Cleanup() looks very strict around wanting to remove all pins explicitly.

lmb: We don't really have tests which pin and then check that the function removes pins, I discovered this the hard way. Mostly because on Linux the cleanup is just `os.RemoveAll`, where here it's more involved. We could make this a test which runs on Linux as well, the benefit there is minimal IMO.

lmb: That is to say: I think it's worth having this test.

PR #1689, opened 2025-02-17
all: encode platform in type constants

lmb:
Linux and Windows share the concept of program types, map types and so on. For example, XDP is supported on both platforms. Unfortunately the constant values used by both platforms are different. On Linux, XDP has value 0x6 while on Windows it is 0x1.

This problem extends to a CollectionSpec parsed from an ELF. Because of the different values a Windows PerCPUHash aliases with a Linux ProgramArray. This causes all sorts of mayhem that is hard to avoid without making too many changes to the code base.

Introduce the notion of a supported platform. External users only see this as a string, like "linux". Internally each platform has a corresponding tag which is used to encode the platform into a constant. This means that there are distinct constants for each platform. Instead of overloading the XDP constant we will introduce WindowsXDP. This will allow the ELF reader to remain platform independent and makes generating stringers a lot easier.

The value of the platform tag is chosen so that on Linux the constant values do not change and are identical with the value in upstream headers. On Windows the constants have a fixed (but implementation defined) offset.

The downside of this approach is that Windows constants will all require a prefix of some sort to distinguish them from Linux ones. This is probably most onerous for map types, because those tend to be created manually more frequently than programs, for example. We can add some behind the scenes translation of Linux map types to Windows ones if need be.

Another downside is that constant values are now limited to 28 bits since we steal the top 4 bits to store the platform. The constants we're applying this to are all sequentially numbered from 0, so this is hopefully enough.


lmb:
```
github.com/cilium/ebpf/asm
  Incompatible changes:
  - (*Instruction).Unmarshal: changed from func(io.Reader, encoding/binary.ByteOrder) (uint64, error) to func(io.Reader, encoding/binary.ByteOrder, string) error
  - (*Instructions).Unmarshal: removed
  - BuiltinFunc: changed from int32 to uint32
```

internal/platform/constants.go
```diff
@@ -0,0 +1,57 @@
...
+		return ""
+	}
+}
+
+// Encode a platform and a value into a tagged constant.
+//
+// The platform tag is stored in the 4 most significant bits. The tags value is
+// one less than the platform constant so that Linux constants remain the same.
+//
+// Returns an error if either p or c are out of bounds.
```
ti-mo, line 39: Nit: outdated comment.

internal/platform/constants.go
```diff
@@ -0,0 +1,57 @@
...
+	case LinuxTag:
+		return Linux
+	default:
+		return ""
+	}
+}
+
+// Encode a platform and a value into a tagged constant.
+//
+// The platform tag is stored in the 4 most significant bits. The tags value is
```
ti-mo, line 36: What is meant by the `tags` value? The platform constant I assume is LinuxTag in this case? Ack on `LinuxTag` being zero for backwards compat, but the 'one less than ..' throws me off. Maybe the comment just hasn't evolved with the rest of the code.

linker.go
```diff
@@ -272,9 +273,11 @@ func fixupAndValidate(insns asm.Instructions) error {
...
 }
 
-// POISON_CALL_KFUNC_BASE in libbpf.
-// https://github.com/libbpf/libbpf/blob/2778cbce609aa1e2747a69349f7f46a2f94f0522/src/libbpf.c#L5767
-const kfuncCallPoisonBase = 2002000000
+// A constant used to poison calls to non-existent kfuncs.
+//
+// Similar POISON_CALL_KFUNC_BASE in libbpf, except that we use a value lower
+// than 2^28 to fit into a tagged constant.
+const kfuncCallPoisonBase = 1<<28 - 1
```
ti-mo, line 280: Would it make sense to export `platform.platformMask` for this purpose? This is basically `^platformMask`. I'm also not sure if all 1s retains its debuggability properties. We could also pick something else that fits in 28 bits like `0xdedc0de`, and it would be relevant, too. :wink:

linker.go
```diff
@@ -331,8 +334,11 @@ fixups:
 				// If the kfunc call is weak and not found, poison the call. Use a recognizable constant
 				// to make it easier to debug. And set src to zero so the verifier doesn't complain
 				// about the invalid imm/offset values before dead-code elimination.
-				ins.Constant = kfuncCallPoisonBase
-				ins.Src = 0
```
ti-mo, line 335: I assume zeroing Src was redundant to begin with, but just checking. Comment may need a little touch-up if that's the case.

types.go
```diff
@@ -182,6 +193,14 @@ const (
 	Netfilter             = ProgramType(sys.BPF_PROG_TYPE_NETFILTER)
 )
 
+func ProgramTypeForPlatform(plat string, value uint32) (ProgramType, error) {
```
ti-mo, line 196: Reminder to change this to whatever we decide for MapType.

types.go
```diff
@@ -252,6 +271,14 @@ const (
 	AttachNetkitPeer                 = AttachType(sys.BPF_NETKIT_PEER)
 )
 
+func AttachTypeForPlatform(plat string, value uint32) (AttachType, error) {
```
ti-mo, line 274: Reminder to change this to whatever we decide for MapType.

asm/instruction.go
```diff
@@ -529,29 +543,24 @@ type FDer interface {
...
 
-// Unmarshal unmarshals an Instructions from a binary instruction stream.
-// All instructions in insns are replaced by instructions decoded from r.
-func (insns *Instructions) Unmarshal(r io.Reader, bo binary.ByteOrder) error {
-	if len(*insns) > 0 {
-		*insns = nil
-	}
-
+// AppendInstructions decodes [Instruction] from r and appends them to insns.
+func AppendInstructions(insns Instructions, r io.Reader, bo binary.ByteOrder, platform string) (Instructions, error) {
```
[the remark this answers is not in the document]
lmb: The common operation is to encode into a new slice. With a free function that is:

```
AppendInstructions(nil, ...)
```

As a method it requires an extra line:
```
var insns Instructions
insns.Append(...)
```

linker.go
lmb: Went with 0xdedc0de, thanks!

linker.go
lmb: No, src had to change because that indicates that the call is a kfunc call. This is now taken care of by by `Call`.

types.go
```diff
@@ -104,6 +105,16 @@ const (
 	Arena
 )
 
+// MapTypeForPlatform returns a platform specific map type.
+func MapTypeForPlatform(plat string, typ uint32) (MapType, error) {
```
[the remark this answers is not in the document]
lmb: That's not how it's intended to be used at the moment. It solves the problem of what to do if a new map type is introduced which the library doesn't have a constant for (yet). Then you can do:

```
mapType, err := MapTypeForPlatform("windows", 0xfeed)
// use mapType in calls that expect MapType
```

Remember that HashOfMaps on linux does not have the same numerical value as HashOfMaps on windows.

types.go
```diff
@@ -104,6 +105,16 @@ const (
...
+	return platform.EncodeConstant[MapType](plat, typ)
+}
+
+// Decode a constant to its platform-dependent value.
+func (mt MapType) Decode() (string, uint32) {
```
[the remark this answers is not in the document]
lmb: SGTM

PR #1690, opened 2025-02-17
Load kernel BTF spec once when creating a new collection

patrickpichler:
When creating a new collection without specifying the `KernelTypes` in the `ProgramOptions`, the kernel BTF is implicitly loaded. For this the `btf.LoadKernelSpec()` helper is used. Even though the spec is only loaded once, this operation cause a lot of memory churn, as the kernel spec is being copied each time the function get called. This becomes an issue when a collection with lots of programs is loaded in a resource limited environment.

This commit initializes an missing `KernelTypes` field with the BTF kernel spec, getting rid of the subsequent calls of `BTF.Copy()`.

This greatly reduces the number of allocations/op when running the `BenchmarkNewCollectionManyProgs` benchmark. Here are the results:

Previous:
BenchmarkNewCollectionManyProgs-4  12 98443414 ns/op 116779863 B/op 403964 allocs/op

With optimization:
BenchmarkNewCollectionManyProgs-4 184  5807742 ns/op   4134444 B/op      17325 allocs/op

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -252,7 +252,11 @@ func (cs *CollectionSpec) Assign(to interface{}) error {
 // Returns an error if any of the fields can't be found, or
 // if the same Map or Program is assigned multiple times.
 func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions) error {
-	loader, err := newCollectionLoader(cs, opts)
+	if opts == nil {
+		opts = &CollectionOptions{}
+	}
+
+	loader, err := newCollectionLoader(cs, *opts)
 	if err != nil {
 		return err
 	}
@@ -351,7 +355,7 @@ func NewCollection(spec *CollectionSpec) (*Collection, error) {
 // Omitting Collection.Close() during application shutdown is an error.
 // See the package documentation for details around Map and Program lifecycle.
 func NewCollectionWithOptions(spec *CollectionSpec, opts CollectionOptions) (*Collection, error) {
-	loader, err := newCollectionLoader(spec, &opts)
+	loader, err := newCollectionLoader(spec, opts)
 	if err != nil {
 		return nil, err
 	}
@@ -405,9 +409,13 @@ type collectionLoader struct {
 	vars     map[string]*Variable
 }
 
-func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) (*collectionLoader, error) {
-	if opts == nil {
-		opts = &CollectionOptions{}
+func newCollectionLoader(coll *CollectionSpec, opts CollectionOptions) (*collectionLoader, error) {
+	if opts.Programs.KernelTypes == nil {
+		kernelSpec, err := btf.LoadKernelSpec()
+		if err != nil {
+			return nil, fmt.Errorf("cannot load kernel spec: %w", err)
+		}
+		opts.Programs.KernelTypes = kernelSpec
 	}
 
 	// Check for existing MapSpecs in the CollectionSpec for all provided replacement maps.
@@ -423,7 +431,7 @@ func newCollectionLoader(coll *CollectionSpec, opts *CollectionOptions) (*collec
 
 	return &collectionLoader{
 		coll,
-		opts,
+		&opts,
 		make(map[string]*Map),
 		make(map[string]*Program),
 		make(map[string]*Variable),


patrickpichler:
@lmb @dylandreimerink I know the test for kernels without BTF support are failing. I guess this can be fixed by lazy loading the kernel spec only once. Before I start refactoring the code though, is this approach something you would merge? 

A different approach might be to refactor the `btf.Spec` type and somehow introduce a immutable version of it. I guess this is not really possible without introducing breaking changes though.

lmb:
> For this the `btf.LoadKernelSpec()` helper is used. Even though the spec is only loaded once, this operation cause a lot of memory churn, as the kernel spec is being copied each time the function get called.

IIRC we don't really copy the full spec. It's a shallow, lazy copy. Is that still too expensive?

> This commit initializes an missing `KernelTypes` field with the BTF kernel spec, getting rid of the subsequent calls of `BTF.Copy()`.

The problem with this is that it always causes a parse of kernel BTF, even in cases where the load wouldn't have needed one.

patrickpichler:
> IIRC we don't really copy the full spec. It's a shallow, lazy copy. Is that still too expensive?

Ok after looking at the flame graphs again, I think the issue is not directly the copying of the kernel BTF spec, but since each time relocations get applied, they call `spec.TypeByID` [here](https://github.com/cilium/ebpf/blob/d2f8fcb4a31f6ad9fa4b25cd1499971d64bd2456/btf/core.go#L271), which if the BTF spec is shared will short circuit, as the type has already been copied, but for new BTF specs, it will always end up copying the same type over and over again. I misunderstood the copy logic the first time I look at this, causing me to suspect it is at fault, but since the kernels BTF spec always returns a copy, it should never have to deep copy copied types (there should be no copied types). 

An alternative could be, to use the type directly from the `imm` part of the spec. This greatly reduces allocations, but I guess could be dangerous, as changes to types found in the instruction metadata would ripple through. Though I am not sure if there is a valid use case for this (just because I cannot think of it, of course doesn't mean there isn't 😅). I prototyped this [here](https://github.com/patrickpichler/ebpf-go/commit/334a87e3904e630afdd691eccadf0f7aeb8e05e3).

What do you think about this solution?

lmb:
Sorry for the late reply. The culprit for you is always CO-RE relocation?

I think we could extend your approach to add an unexported method `Spec.readOnlyTypeByID` and use that from CO-RE. That function would either return an already copied type, or use the immutable type directly. It's not safe to export this since the caller has to pinky promise that they really won't modify the type. We also need to audit the CO-RE code to ensure that it doesn't modify the type somehow and in the best case even write a test for this. (Gut feeling: the test will be the hardest part.)

lmb:
> I guess the `Spec.readOnlyTypeByID` will in the end just be a wrapper around `imm.typeByID`, or do you have anything more ambitious in mind?

It would have to:

1. Check `mutableTypes.copies` for a copied type and return that if it exists.
2. Return the immutable type.

> create a forced deep copy of the BTF spec (I guess a new method is needed here), load the program and then compare if the specs are the same.

I think that could work, but maybe it's enough to copy `Spec.types` before and then compare only that slice? Not sure we need to check all the other members of Spec.

patrickpichler:
Sounds good!
I will try to produce a PoC in the next comming days and report back with performance numbers 👍

patrickpichler:
Sorry for the long radio silence, the last weeks were busier than thought. Figuring out if the specs have changed is harder than expected. One thing I noticed though, I believe there is a potential problem with how `Type.copy` is implemented for types that reference other types. 

Take the implementation of the Array type for example ([source](https://github.com/cilium/ebpf/blob/334a87e3904e630afdd691eccadf0f7aeb8e05e3/btf/types.go#L163)):
```go
func (arr *Array) copy() Type {
	cpy := *arr
	return &cpy
}
```

This will cause the `Index`/`Type` types to be shallow copied. Meaning if one changes a mutable array type, it will also update the immutable type (unless I am overlooking something, but I've checked the way types are copied in the [copyType](https://github.com/cilium/ebpf/blob/334a87e3904e630afdd691eccadf0f7aeb8e05e3/btf/types.go#L725) function and form what I can tell, it only calls the `Type.copy` function).

@lmb do you think this is an issue worth addressing?

lmb:
See https://github.com/cilium/ebpf/pull/1778

PR #1693, opened 2025-02-20
prog: avoid verifier loop of death

ti-mo:
Supersedes https://github.com/cilium/ebpf/pull/1679.

```
prog: factor retry decision into separate testable function

This commit special-cases LogDisabled into a clause without a retry loop since
it doesn't need a buffer allocated. It also simplifies the logic somewhat.

A follow-up commit will add another separate clause for a new LogSizeOverride
CollectionOption to factor out even more complexity related to buffer sizing.

The main purpose of this commit is to extract the decision-making logic around
retries. Instead of directly manipulating loop control flow, a function is
introduced that issues a retry verdict, and it can be tested indendently of
the program load action.
```

```
prog: give up when max verifier log size is reached

Previously, loading a program would loop forever if the log didn't fit
in the kernel's max log size.
```

```
prog: prevent prog_load loop of death

To limit the damage of any future regressions, limit the attempts we'll take
to load a program.
```

@lmb I eventually decided against deprecating LogSizeStart because it has utility beyond just break-glass cases, for example to reduce the amount of attempts the Cilium verifier tests take to obtain the full buffer, which is often quite large. We don't want to hardcode a buffer size there because the size varies significantly across programs and the environment is somewhat resource-constrained. It also ended up being more code; as of this proposal it's quite concise and testable, and I'm happy with it.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -409,59 +409,39 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 		}
 	}
 
-	// The caller requested a specific verifier log level. Set up the log buffer
-	// so that there is a chance of loading the program in a single shot.
-	logSize := internal.Between(opts.LogSizeStart, minVerifierLogSize, maxVerifierLogSize)
 	var logBuf []byte
-	if !opts.LogDisabled && opts.LogLevel != 0 {
-		logBuf = make([]byte, logSize)
-		attr.LogLevel = opts.LogLevel
-		attr.LogSize = uint32(len(logBuf))
-		attr.LogBuf = sys.NewSlicePointer(logBuf)
-	}
+	var fd *sys.FD
 
-	for {
-		var fd *sys.FD
+	if opts.LogDisabled {
+		// Loading with logging disabled should never retry.
 		fd, err = sys.ProgLoad(attr)
 		if err == nil {
-			return &Program{unix.ByteSliceToString(logBuf), fd, spec.Name, "", spec.Type}, nil
+			return &Program{"", fd, spec.Name, "", spec.Type}, nil
+		}
+	} else {
+		// Only specify log size if log level is also specified. Setting size
+		// without level results in EINVAL. Level will be bumped to LogLevelBranch
+		// if the first load fails.
+		if opts.LogLevel != 0 {
+			attr.LogLevel = opts.LogLevel
+			attr.LogSize = internal.Between(opts.LogSizeStart, minVerifierLogSize, maxVerifierLogSize)
 		}
 
-		if opts.LogDisabled {
-			break
+		for {
+			if attr.LogLevel != 0 {
+				logBuf = make([]byte, attr.LogSize)
+				attr.LogBuf = sys.NewSlicePointer(logBuf)
+			}
+
+			fd, err = sys.ProgLoad(attr)
+			if err == nil {
+				return &Program{unix.ByteSliceToString(logBuf), fd, spec.Name, "", spec.Type}, nil
+			}
+
+			if !retryLogAttrs(attr, opts.LogSizeStart, err) {
+				break
+			}
 		}
-
-		if attr.LogTrueSize != 0 && attr.LogSize >= attr.LogTrueSize {
-			// The log buffer already has the correct size.
-			break
-		}
-
-		if attr.LogSize != 0 && !errors.Is(err, unix.ENOSPC) {
-			// Logging is enabled and the error is not ENOSPC, so we can infer
-			// that the log buffer is large enough.
-			break
-		}
-
-		if attr.LogLevel == 0 {
-			// Logging is not enabled but loading the program failed. Enable
-			// basic logging.
-			attr.LogLevel = LogLevelBranch
-		}
-
-		// Make an educated guess how large the buffer should be by multiplying.
-		// Ensure the size doesn't overflow.
-		const factor = 2
-		logSize = internal.Between(logSize, minVerifierLogSize, maxVerifierLogSize/factor)
-		logSize *= factor
-
-		if attr.LogTrueSize != 0 {
-			// The kernel has given us a hint how large the log buffer has to be.
-			logSize = attr.LogTrueSize
-		}
-
-		logBuf = make([]byte, logSize)
-		attr.LogSize = logSize
-		attr.LogBuf = sys.NewSlicePointer(logBuf)
 	}
 
 	end := bytes.IndexByte(logBuf, 0)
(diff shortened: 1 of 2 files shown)


prog.go
```diff
@@ -511,6 +503,52 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+		if attr.LogTrueSize != 0 {
+			// Kernel supports LogTrueSize and previous iteration undershot the buffer
+			// size. Try again with the given true size.
+			attr.LogSize = attr.LogTrueSize
+			return true
+		}
+
+		// ENOSPC means we've loaded the program before and the log buffer was too
+		// small.
+		if attr.LogSize == maxVerifierLogSize {
```
brycekahle, line 519: I'm not sure you will ever reach this condition, because I don't think the kernel will return `ENOSPC` when using the max verifier log size. That said, it doesn't hurt to be defensive here.

brycekahle (approved):
I tested and did not receive the same infinite loop we were seeing before.

prog.go
```diff
@@ -409,59 +415,45 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+			attr.LogSize = internal.Between(opts.LogSizeStart, minVerifierLogSize, maxVerifierLogSize)
 		}
 
-		if attr.LogSize != 0 && !errors.Is(err, unix.ENOSPC) {
-			// Logging is enabled and the error is not ENOSPC, so we can infer
-			// that the log buffer is large enough.
-			break
-		}
+		attempts := 1
+		for {
```
lmb, line 437: Nit:
```suggestion
		for attempts := 0; attemps < maxVerifierAttempts ; attempts++ {
```
Putting the increment into the for loop has the benefit that it does the right thing when continuing, etc.

prog.go
```diff
@@ -511,6 +503,52 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+		// ENOSPC means we've loaded the program before and the log buffer was too
+		// small.
+		if attr.LogSize == maxVerifierLogSize {
+			// Maximum buffer size reached, don't grow or retry.
+			return false
+		}
+
+		// Ensure the size doesn't overflow.
+		const factor = 2
+		if attr.LogSize >= maxVerifierLogSize/factor {
```
lmb, line 526: I find this whole factor thing not very intuitive. This is to avoid overflow?

prog.go
```diff
@@ -511,6 +503,52 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
 	return nil, internal.ErrorWithLog("load program", err, logBuf)
 }
 
+func retryLogAttrs(attr *sys.ProgLoadAttr, startSize uint32, err error) bool {
+	// ENOSPC means the log was enabled on the previous iteration, so we only
+	// need to grow the buffer.
+	if errors.Is(err, unix.ENOSPC) {
+		if attr.LogTrueSize != 0 {
```
lmb, line 510: This doesn't preserve the behaviour that we don't retry if LogSize >= LogTrueSize. I'm also not sure why this only happens on ENOSPC? LogTrueSize should always be populated?

prog.go
```diff
@@ -409,59 +415,45 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions) (*Program, er
...
+		// Loading with logging disabled should never retry.
 		fd, err = sys.ProgLoad(attr)
 		if err == nil {
-			return &Program{unix.ByteSliceToString(logBuf), fd, spec.Name, "", spec.Type}, nil
-		}
-
-		if opts.LogDisabled {
-			break
+			return &Program{"", fd, spec.Name, "", spec.Type}, nil
 		}
```
lmb, line 426: You can return here, all of the other code only deals with the log being present. Then the else can go and we can unindent by one leve.

lmb (changes requested):

prog.go
brycekahle: LogTrueSize is not available on older kernels

brycekahle:
Where do we stand on this? Anything you need me to do? We'd like to update to get some other fixes/improvements, but not having this is a blocker.

ti-mo:
Just got back from PTO, I'll try to get this in tomorrow.

prog.go
ti-mo: Good point, let's not assume ENOSPC will be returned. If the max log size is hit, we should never retry.

prog.go
ti-mo: What do you mean? The code after the `else {}` deals with wrapping EPERM, EFAULT, etc.

prog.go
ti-mo: True in principle, but we want to make sure we hit the error condition at the end of the loop so we return an error instead of continuing execution if the limit is hit. The way I wrote it, we're much less likely to get off-by-one errors resulting in the error return being skipped.

prog.go
ti-mo: If the kernel returns ENOSPC, LogSize is smaller than LogTrueSize by definition, and we just need to allocate whatever LogTrueSize specified and retry. I don't think what I wrote here will result in a retry if LogSize > LogTrueSize. What am I missing?

prog.go
ti-mo: Do you have an alternative suggestion? The intention is for this to keep working if/when maxVerifierLogSize gets bumped to MaxUint32. Currently, we can't rely on a u32 overflow happening, detecting it, and clamping the value to maxVerifierLogSize. Either way, that's more code and more obtuse than this solution. I'll go ahead with the current version for now.

ti-mo:
> That tells me that the current approach just doesn't work very well and we should try to do better. I have a firm belief that making log size at all configurable was a mistake from the get go.

I think at the time, you just picked whatever libbpf did. Over the years, this piece of code grew more and more complex. We added (and then removed) VerifierError.Truncated. Then LogTrueSize came and made things worse, although improving things in the long term. I remember a conversation years ago when I migrated the Cilium agent over to the lib, where we concluded to kick the can down the road and let the agent deal with automatic buffer resizing. Then years later you suddenly changed your mind and added resizing to the lib anyway.

Meanwhile, we only made half-hearted attempts at reining in the complexity, so we kept shipping regressions. I don't think it's helpful blaming this on a single decision made half a decade ago when the landscape was very different. :) In the current proposal, the starting size of the buffer is plumbed through in a way that makes it easy to remove if we really want to later on. If anything, blame it on the kernel API. :wink: 

> So: how large is "often quite large"?

Large enough to matter. :smile: It doesn't really matter what the actual size is, this will vary per use case. The bigger the delta from minVerifierLogSize, the more retries have to be made. Think of the poor CPUs. :cry:

PR #1694, opened 2025-02-20
memory: fix off-by-one in bounds check

ti-mo:
It previously wasn't possible to access the last byte of the memory segment. Also add a u64 overflow check.

This wasn't caught in tests because this requires accessing a page-aligned variable in its entirety. Add a test with a page-aligned array Variable.

cc @mejedi


mejedi (approved):
Thank you for making a fix.

PR #1702, opened 2025-02-24
ebpf: clean up tests

lmb:
Introduce helpers for creating a map, program, collection from a spec. Additionally, add helpers which create very common types like arrays or socket filter programs.

This significantly cuts down on boiler plate in unit tests, and ensures that all resources backed by a file descriptor are cleaned up.

It will also allow adding compatibility code for running Linux unit tests on Windows in a central location.

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -5,6 +5,9 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/go-quicktest/qt"
+
+	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/btf"
 	"github.com/cilium/ebpf/internal/testutils"
 )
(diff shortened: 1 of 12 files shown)


helpers_test.go
```diff
@@ -35,3 +38,149 @@ func requireTestmod(tb testing.TB) {
...
+}
+
+func newProgram(tb testing.TB, spec *ProgramSpec, opts *ProgramOptions) (*Program, error) {
+	tb.Helper()
+
+	if opts == nil {
+		opts = new(ProgramOptions)
+	}
+
+	if spec.Type == SkLookup && spec.AttachType == AttachNone {
```
ti-mo, line 103: Assuming this is for `TestProgramRunEmptyData`, I'd make that test call `newProgram` instead so it can provide a spec with the right attach type. This seems a little heavy-handed.

ti-mo (commented):
Ack on the concept, left one comment.

PR #1709, opened 2025-03-03
windows: even more test fixes

lmb:
More prep work to make the Windows support easier to add.

collection: remove obsolete TestCollectionSpecLoadCopy

    The test was added in bf5a6cd6 ("collection: don't copy btf.Spec in
    CollectionSpec.Copy()"). We now copy the spec, so the test doesn't add
    anything meaningful.

collection: remove TestCollectionSpecLoadMutate

    Replace the test added in 278a9141 ("map: check MapReplacements 
    compatibility after applying BPF_F_MMAPABLE") with one that doesn't rely on
    the ELF reader.

collection: fix missing Map.Close in collectionLoader.loadMap

program: rework TestProgramRunEmptyData

    Remove the test in favour of a much shorter assert in TestProgramRun. Also
    remove the now unnecessary version skip, since we don't test against such
    old kernel versions anymore.

program: remove duplicate TestProgramVerifierOutput

    The test is a less exhaustive version of TestProgramVerifierLog, remove it.

testutils: add missing t.Helper() calls

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -61,8 +61,8 @@ func (cs *CollectionSpec) Copy() *CollectionSpec {
 	}
 
 	cpy := CollectionSpec{
-		Maps:      make(map[string]*MapSpec, len(cs.Maps)),
-		Programs:  make(map[string]*ProgramSpec, len(cs.Programs)),
+		Maps:      copyMapOfSpecs(cs.Maps),
+		Programs:  copyMapOfSpecs(cs.Programs),
 		Variables: make(map[string]*VariableSpec, len(cs.Variables)),
 		ByteOrder: cs.ByteOrder,
 		Types:     cs.Types.Copy(),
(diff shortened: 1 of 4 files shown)


collection.go
```diff
@@ -61,8 +61,8 @@ func (cs *CollectionSpec) Copy() *CollectionSpec {
 	}
 
 	cpy := CollectionSpec{
-		Maps:      make(map[string]*MapSpec, len(cs.Maps)),
-		Programs:  make(map[string]*ProgramSpec, len(cs.Programs)),
+		Maps:      copyMapOfSpecs(cs.Maps),
```
ti-mo, line 64: Did you forget to remove the two `for name, spec := range cs... {` loops below? I think this copies twice now.

lmb: Derp :cry:

PR #1715, opened 2025-03-08
map: fix problematic error returns

cuishuang:
We have already judged err before. Therefore, err here must be nil.

Here  `sysErr` should be returned instead of err.


florianl (approved):
Thanks for the fix!

lmb (approved):

PR #1720, opened 2025-03-11
improve error handling for einval with lpm trie.

tommyp1ckles:
In case of invalid key (i.e. not following lpm rules) the default error message is quite misleading:

`noPrealloc flag may be incompatible with map type`

In fact, lpm-trie requires no prealloc to be enabled so this error is likely to waste a confused programmers time.

Instead, we validate key params and return a specific error.

### DIFF
--- a/map.go
+++ b/map.go
@@ -556,7 +556,12 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 	if errors.Is(err, unix.EINVAL) && spec.Type == UnspecifiedMap {
 		return fmt.Errorf("map create: cannot use type %s", UnspecifiedMap)
 	}
-	if errors.Is(err, unix.EINVAL) && spec.Flags&sys.BPF_F_NO_PREALLOC > 0 {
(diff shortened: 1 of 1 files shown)


map.go
```diff
@@ -556,7 +556,12 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 	if errors.Is(err, unix.EINVAL) && spec.Type == UnspecifiedMap {
 		return fmt.Errorf("map create: cannot use type %s", UnspecifiedMap)
 	}
-	if errors.Is(err, unix.EINVAL) && spec.Flags&sys.BPF_F_NO_PREALLOC > 0 {
+	// LPM trie key size must be a multiple of 8 and not exceed 2048.
```
chenhengqi, line 559: This seems incorrect according to my experience and the doc `https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_LPM_TRIE/`.

ti-mo:
Continuing this in https://github.com/cilium/ebpf/pull/1776.

PR #1728, opened 2025-03-21
features: update map feature matrix for types added since 5.11

tklauser:
Add feature probes for map types `BloomFilter`, `UserRingbuf`, `CgroupStorage` and `Arena`.

### DIFF
--- a/features/map.go
+++ b/features/map.go
@@ -213,6 +213,47 @@ var haveMapTypeMatrix = internal.FeatureMatrix[ebpf.MapType]{
 		Version: "5.11",
 		Fn:      func() error { return probeStorageMap(sys.BPF_MAP_TYPE_TASK_STORAGE) },
 	},
+	ebpf.BloomFilter: {
+		Version: "5.16",
+		Fn: func() error {
+			return createMap(&sys.MapCreateAttr{
+				MapType:    sys.BPF_MAP_TYPE_BLOOM_FILTER,
+				KeySize:    0,
+				ValueSize:  4,
+				MaxEntries: 1,
+			})
+		},
+	},
+	ebpf.UserRingbuf: {
+		Version: "6.1",
+		Fn: func() error {
+			// keySize and valueSize need to be 0
+			// maxEntries needs to be power of 2 and PAGE_ALIGNED
+			return createMap(&sys.MapCreateAttr{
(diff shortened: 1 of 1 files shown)


features/prog.go
```diff
@@ -261,6 +270,8 @@ func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
 		spec.AttachType = ebpf.AttachSkLookup
 	case ebpf.Syscall:
 		spec.Flags = sys.BPF_F_SLEEPABLE
+	case ebpf.Netfilter:
```
lmb, line 273: Can you extend the unit tests to check at least one netfilter helper? We don't get coverage by default, I think.

rgo3 (approved):

features/prog.go
tklauser: @lmb sorry, I'm a bit lost here. Looking at the helpers  I don't see any netfilter specific ones (yet). In the kernel selftests I see e.g. `bpf_dynptr_*` helpers [being used](https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/bpf/progs/ip_check_defrag.c), but these helpers are used for other program types as well, so I'm not sure whether that's applicable.

ti-mo: Correct, I don't think there are any netfilter-specific helpers. The last one (cgrp_storage_delete) was merged in Oct '22, which is before Netfilter progs were merged (April '23). Pretty normal given the decision to introduce new helper functionality using kfuncs instead; there likely won't be any netfilter-specific helpers, ever. And Kfuncs can be probed using CO-RE within the bpf prog itself, so no need for a Go API to probe them.

ti-mo: @tklauser I think what Lorenz means is adding a test to `TestHaveProgramHelper` that exercises this `case ebpf.Netfilter`, since coverage here is not provided by the addition to `haveProgramTypeMatrix` above.

That said, it looks like we're missing tests for `CGroupSockopt`, `SkLookup` and `Syscall`.. Would you be able to pick those up in this PR as well? :pray:

tklauser: Done, thanks for the explanation. Pushed an additional commit adding coverage for `CGroupSockopt`, `SkLookup` and `Syscall`.

ti-mo:
Excellent, many thanks!

PR #1729, opened 2025-03-21
go.mod: bump golang.org/x/net to 0.36.0

tklauser:
This was generated by running:

    go get golang.org/x/net@v0.36.0
    go mod tidy
    go mod edit -toolchain=none

Replaces #1724

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,6 @@ require (
 	github.com/mdlayher/netlink v1.7.2 // indirect
 	github.com/mdlayher/socket v0.4.1 // indirect
 	github.com/rogpeppe/go-internal v1.12.0 // indirect
-	golang.org/x/net v0.33.0 // indirect
+	golang.org/x/net v0.36.0 // indirect
 	golang.org/x/sync v0.1.0 // indirect
 )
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cilium/ebpf
 
-go 1.23
+go 1.23.0
```
lmb, line 3: This runs into https://github.com/cilium/ebpf/commit/eaef6a7c6a620f7e673463e367c8e3ede8c4ca57 again. I'm really quite sick of the the whole add a .0 or don't add a .0 to the go line saga xD

lmb: I'd be grateful for some semi-official answer what the thing to do is for depending on a language version. I don't want to pin toolchain because I don't really understand why its necessary and it's another thing to manage.

ti-mo: I'm also equally confused about the meaning of the `toolchain` directive on projects like these. I guess it's the toolchain version used when running `go test`, etc. Maybe we can let this be managed by dependabot, because I don't see a way of configuring this in dependabot.yml. There's no ecosystem-specific configuration whatsoever, and we don't even have a gomod block in there, it's enabled through the GitHub UI.

ti-mo:
I force-pushed this branch to read `go 1.23`, but this somehow makes golangci-lint unable to run: https://github.com/cilium/ebpf/actions/runs/14054937043/job/39352226944?pr=1729

Reading https://github.com/cilium/ebpf/commit/eaef6a7c6a620f7e673463e367c8e3ede8c4ca57 again:

> On https://github.com/cilium/ebpf/commit/b3432fb1eca90dae072e53d428f248aecec65d18, the minimum supported Go version was updated to `1.21.0`. This changes the version to `1.21` so that we are using a language version. **If not using toolchain**, this seems to be the preferred convention in the ecosystem and golang/go.

Seems to me like `go mod tidy` behaviour changes depending on the used toolchain, so putting a `toolchain` directive in go.mod seems like the best way to keep this behaviour stable in the same sense that e.g. an LLVM builder image does.

Bottom line is https://github.com/cilium/ebpf/commit/eaef6a7c6a620f7e673463e367c8e3ede8c4ca57 may no longer be an issue if both `go 1.23.0` and `toolchain go1.24.1` are specified. I've run `go mod tidy` on the dependabot PR and force-pushed it; looks like it's working. Let's go with the initial PR.

Issue #1652, opened 2025-01-21
pin: WalkDir makes it too easy to leak objects


lmb:
> It's very easy to accidentally leak objects when using WalkDir, and kind of awkward to do the right thing (because obj can be nil?) This is why https://pkg.go.dev/github.com/cilium/ebpf/btf#HandleIterator.Take exists for handles. It might be good to do another pass over this API.

_Originally posted by @lmb in https://github.com/cilium/ebpf/issues/1651#issuecomment-2605231732_
            
More context from Mahe:
>Mmmh I'm slightly confused now, this it the typical use we have in Tetragon:
>
>    https://github.com/cilium/tetragon/blob/d72f98aefb8e88c06caec7cfbc687bbeabf97a5f/cmd/tetra/debug/progs.go#L408-L445
>    https://github.com/cilium/tetragon/blob/d72f98aefb8e88c06caec7cfbc687bbeabf97a5f/pkg/bugtool/maps.go#L148-L176
>
> Maybe having an example in the doc would make thing more clear?

As far as  I can tell both of these examples leak objects. Not a big deal in these cases since the processes are short lived (I think?) but an indication that the API is bug prone. My gut feeling is that this is because of wrapping `WalkDir` which in itself is a really clunky API.

Ideas:
1. Drop Walk outright and have callers do the call to Load themselves. That makes it clearer that they are reponsible for cleaning up.
2. Only invoke Walk for valid objects so that an unconditional `defer obj.Close()` at the top of the function works.

Since 1.24 is around the corner it might also be interesting to consider what it would look like to turn this API into an iterator.

ti-mo:
As far as I can tell, Go iterators are also callback-based and would require something similar to `HandleIterator.Take` if we decide to always Close() under the covers.

> Drop Walk outright and have callers do the call to Load themselves. That makes it clearer that they are reponsible for cleaning up.

The API aims to unburden the caller from this exact pain. The library already does the (runtime) work of opening the pin and determining its type. Closing it and requiring the caller to repeat the same work feels very inefficient; doubly so if these are short-lived objects.

I think this is a documentation issue; this isn't much different from any other omission of `Close()`. I'm of course open to other API forms if there are better suggestions, but iirc Go iterators lack the feedback mechanism that WalkDir provides, e.g. skipping directories to avoid unnecessary tree visits.

Haven't really heard about people taking issue with `filepath.WalkDir()`, I think it's just okay personally.

PR #1736, opened 2025-03-31
pin: make WalkDir harder to misuse and add Windows support

lmb:
The current WalkDir API requires the user to close each object that is iterated over. This isn't very intuitive, the two uses of the function in the tetragon code base end up leaking objects. It's also not straight forward to implement WalkDir on Windows, since BPFFS isn't really a thing on that platform.

Instead, turn WalkDir into an iterator which returns a new Pin object. By default, objects are closed once they have been iterated over. The user may call Pin.Take if they wish to retain the object. This is similar to the existing link.Iterator.Take.

One downside is that directories are not returned by the iteration anymore. The existing code in tetragon doesn't use them and is therefore not affected. There is also no more fine grained control over whether to descent into a directory. Aborting an iteration is still supported and as efficient as before.

Fixes: https://github.com/cilium/ebpf/issues/1652


lmb:
@mtardy could you take a look whether this still fits your usecase in tetragon?

pin/pin.go
```diff
@@ -0,0 +1,22 @@
+package pin
+
+// Pin represents an object and its pinned path.
+type Pin struct {
+	Path   string
+	Object Pinner
```
lmb, line 6: This could also be io.Closer instead.

mtardy:
> @mtardy could you take a look whether this still fits your usecase in tetragon?

will take a look shortly 👀 , was busy conferencing...

lmb:
Yeah those changes look good! You can pass `nil` options to WalkDir I think. I wonder whether I should drop the return value from `Take()` for now, not sure it's particularly useful.

mtardy:
> Yeah those changes look good! You can pass `nil` options to WalkDir I think.

ah yeah it seems `ReadOnly: true` doesn't work well with links

> I wonder whether I should drop the return value from `Take()` for now, not sure it's particularly useful.

ah yeah maybe indeed

mtardy (approved):
I like this new version when using the iter with existing for := range.

PR #1742, opened 2025-04-02
map: unmarshal values for partial batch

### LINKED ISSUE #1741: Value unmarshalling skipped for undersized last batch in BatchLookup
### Describe the bug

If [`batchLookupCmd` returns `ErrKeyNotExist` in `batchLookup`](https://github.com/cilium/ebpf/blob/e3234a15a707cc0b79b78acee40773895350d5d7/map.go#L1160), the [unmarshalling of `valuesOut`](https://github.com/cilium/ebpf/blob/e3234a15a707cc0b79b78acee40773895350d5d7/map.go) is skipped.

My expectation is that the unmarshalling of `valuesOut` is called for the last batch as well. Just like it happens for `keysOut`.

### How to reproduce

The batch size should be chosen in a way so the last batch is less than batch size and `ErrKeyNotExist` is returned.

```go
package main_test

import (
	"errors"
	"testing"

	"github.com/cilium/ebpf"
)

var (
	keysMockCalled   int
	valuesMockCalled int
)

type keysMock []any

func (u keysMock) UnmarshalBinary(data []byte) error {
	keysMockCalled++
	return nil
}

type valuesMock []any

func (u valuesMock) UnmarshalBinary(data []byte) error {
	valuesMockCalled++
	return nil
}

func TestLastBatchUnmarshalValuesNotCalled(t *testing.T) {
	// Choose batch size in a way that last batch returns ErrKeyNotExist.
	var maxEntries, batchSize uint32 = 4, 3

	m, err := ebpf.NewMap(&ebpf.MapSpec{
		Type:       ebpf.Array,
		MaxEntries: maxEntries,
		KeySize:    4,
		ValueSize:  4,
	})
	if err != nil {
		t.Fatalf("create map: %v", err)
	}

	var (
		cursor    ebpf.MapBatchCursor
		keysOut   = make(keysMock, batchSize)
		valuesOut = make(valuesMock, batchSize)
	)

	_, err = m.BatchLookup(&cursor, keysOut, valuesOut, nil)
	if err != nil {
		t.Fatalf("first batch lookup failed: %v", err)
	}

	_, err = m.BatchLookup(&cursor, keysOut, valuesOut, nil)
	if !errors.Is(err, ebpf.ErrKeyNotExist) {
		t.Fatalf("second batch lookup must return ErrKeyNotExist, got %v", err)
	}

	if keysMockCalled != 2 {
		t.Errorf("keys mock calls: got %d, want %d", keysMockCalled, 2)
	}

	// This fails unexpectedly. It should be 2 (called for both batches), but
	// is 1 (not called for last batch).
	if valuesMockCalled != 2 {
		t.Errorf("values mock calls: got %d, want %d", valuesMockCalled, 2)
	}
}
```

This fails with:

```
=== RUN   TestLastBatchUnmarshalValuesNotCalled
    main_test.go:66: values mock calls: got 1, want 2
--- FAIL: TestLastBatchUnmarshalValuesNotCalled (0.00s)
FAIL
```

### Version information

github.com/cilium/ebpf v0.18.0

aibor:
In case of a partial batch the values are not unmarshalled in BatchLookup and BatchLookupDelete for non-perCPU maps.

This commit fixes it by adapting the error handling from batchLookupPerCPU.

Fixes #1741

### DIFF
--- a/map.go
+++ b/map.go
@@ -1124,13 +1124,13 @@ func (m *Map) batchLookup(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, valuesOu
 
 	valueBuf := sysenc.SyscallOutput(valuesOut, count*int(m.fullValueSize))
 
-	n, err := m.batchLookupCmd(cmd, cursor, count, keysOut, valueBuf.Pointer(), opts)
-	if errors.Is(err, unix.ENOSPC) {
+	n, sysErr := m.batchLookupCmd(cmd, cursor, count, keysOut, valueBuf.Pointer(), opts)
+	if errors.Is(sysErr, unix.ENOSPC) {
 		// Hash tables return ENOSPC when the size of the batch is smaller than
 		// any bucket.
-		return n, fmt.Errorf("%w (batch size too small?)", err)
-	} else if err != nil {
-		return n, err
+		return n, fmt.Errorf("%w (batch size too small?)", sysErr)
+	} else if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
+		return 0, sysErr
 	}
 
 	err = valueBuf.Unmarshal(valuesOut)
@@ -1138,7 +1138,7 @@ func (m *Map) batchLookup(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, valuesOu
 		return 0, err
 	}
 
-	return n, nil
+	return n, sysErr
 }
 
 func (m *Map) batchLookupPerCPU(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, valuesOut interface{}, opts *BatchOptions) (int, error) {


lmb:
Oof, thanks. Do you think you can extend the unit tests to cover this case?

aibor:
I could add a test that uses a `binary.Unmarshaler` implementation of a slice type. Kinda like in my reproducer but without the global counter variable. Might be a bit hacky, but should be fine for a test, I guess?

Do you think this should be a new separate test? Or do you have an existing test function in mind that should be extended with a test case for this issue?

aibor:
I have added a new test function that tests key and value unmarshaling with a custom `binary.Unmarshaler`. Without the fix it fails the unmarshaled values assertion:

```
--- FAIL: TestMapBatchLookupCustomUnmarshaler (0.00s)
    map_test.go:1040:
        error:
          values are not deep equal
        diff (-want +got):
            []uint8{
            	0x03,
            	0x04,
          - 	0x05,
          + 	0x03,
            }
        got:
          []uint8("\x03\x04\x03")
        want:
          []uint8("\x03\x04\x05")
```

Issue #1750, opened 2025-04-13
Update `asm` package with load-aquire and store-release instructions


dylandreimerink:
Two new atomic instructs were recently added to the BPF ISA. Lets update our asm package to add the ability to decode them and emit them.

Links:
* https://github.com/torvalds/linux/commit/880442305a3908589bf4d6fc1d79edb577ee497c
* https://lore.kernel.org/all/cover.1741049567.git.yepeilin@google.com/
* https://github.com/llvm/llvm-project/pull/108636

PR #1751, opened 2025-04-13
asm: Add handling for atomic operations

dylandreimerink:
As it stands, we only had proper handling of atomic add operations. `lock *(u32 *)(r1 + 0x1) += w2` was the only instruction that was properly handled. Atomic operations have the same opcode, but and are differentiated by the imm value. So far we have not been looking at the imm value, so all atomic operations look like atomic adds to us.

This commit adds decoding for all current atomic operations. We handle them similarly to how we handle ISAv4 instructions which use the offset to further specify the instruction.

In https://github.com/cilium/ebpf/pull/1193 we expanded the opcode from a u8 to u16, which is bigger than the actual size. This allowed us to still represent functionally different opcodes in Go, even tough the kernel uses other bits of the instruction. So our opcode in Go is not identical to the opcode in the kernel. We translate during marshalling and unmarshaling.

So far, we have only needed a few additional bits, but the atomic ops need 9 bits of imm to fully encode all possibilities. Since 9 + 8 > 16 we have to grow the opcode to 32 bits.

During unmarshaling, we simply take the lower 9 bits of the imm shift it left by 16 bits and or it with the opcode. During marshaling this process is reversed.

Fixes: #1750

### DIFF
--- a/asm/load_store.go
+++ b/asm/load_store.go
@@ -41,24 +41,37 @@ const (
 
 	// AddAtomic - add src to memory address dst atomically
 	AddAtomic AtomicOp = AtomicOp(Add) << 8
+	// FetchAdd - add src to memory address dst atomically, store result in src
+	FetchAdd AtomicOp = AddAtomic | fetch
 	// AndAtomic - bitwise AND src with memory address at dst atomically
 	AndAtomic AtomicOp = AtomicOp(And) << 8
+	// FetchAnd - bitwise AND src with memory address at dst atomically, store result in src
+	FetchAnd AtomicOp = AndAtomic | fetch
 	// OrAtomic - bitwise OR src with memory address at dst atomically
 	OrAtomic AtomicOp = AtomicOp(Or) << 8
+	// FetchOr - bitwise OR src with memory address at dst atomically, store result in src
+	FetchOr AtomicOp = OrAtomic | fetch
 	// XorAtomic - bitwise XOR src with memory address at dst atomically
 	XorAtomic AtomicOp = AtomicOp(Xor) << 8
+	// FetchXor - bitwise XOR src with memory address at dst atomically, store result in src
+	FetchXor AtomicOp = XorAtomic | fetch
 
-	// xchgAtomic - atomically exchange the old value with the new value
-	xchgAtomic AtomicOp = 0x0000_e000
-	// cmpXchgAtomic - atomically compare and exchange the old value with the new value
-	cmpXchgAtomic AtomicOp = 0x0000_f000
+	// Xchg - atomically exchange the old value with the new value
+	//
+	// src gets populated with the old value of *(size *)(dst + offset).
+	Xchg AtomicOp = 0x0000_e000 | fetch
+	// CmpXchg - atomically compare and exchange the old value with the new value
+	//
+	// Compares R0 and *(size *)(dst + offset), writes src to *(size *)(dst + offset) on match.
+	// R0 gets populated with the old value of *(size *)(dst + offset), even if no exchange occurs.
+	CmpXchg AtomicOp = 0x0000_f000 | fetch
 
 	// fetch modifier for copy-modify-write atomics
 	fetch AtomicOp = 0x0000_0100
-	// loadAcquireAtomic - atomically load with acquire semantics
-	loadAcquireAtomic AtomicOp = 0x0001_0000
-	// storeReleaseAtomic - atomically store with release semantics
-	storeReleaseAtomic AtomicOp = 0x0001_1000
+	// loadAcquire - atomically load with acquire semantics
+	loadAcquire AtomicOp = 0x0001_0000
+	// storeRelease - atomically store with release semantics
+	storeRelease AtomicOp = 0x0001_1000
 )
 
 func (op AtomicOp) String() string {
@@ -66,15 +79,15 @@ func (op AtomicOp) String() string {
 	switch op {
 	case AddAtomic, AndAtomic, OrAtomic, XorAtomic:
 		name = ALUOp(op >> 8).String()
-	case AddAtomic | fetch, AndAtomic | fetch, OrAtomic | fetch, XorAtomic | fetch:
+	case FetchAdd, FetchAnd, FetchOr, FetchXor:
 		name = "Fetch" + ALUOp((op^fetch)>>8).String()
-	case xchgAtomic | fetch:
+	case Xchg:
 		name = "Xchg"
-	case cmpXchgAtomic | fetch:
+	case CmpXchg:
 		name = "CmpXchg"
-	case loadAcquireAtomic:
+	case loadAcquire:
 		name = "LdAcq"
-	case storeReleaseAtomic:
+	case storeRelease:
 		name = "StRel"
 	default:
 		name = fmt.Sprintf("AtomicOp(%#x)", uint32(op))
@@ -86,8 +99,8 @@ func (op AtomicOp) String() string {
 func (op AtomicOp) OpCode(size Size) OpCode {
 	switch op {
 	case AddAtomic, AndAtomic, OrAtomic, XorAtomic,
-		AddAtomic | fetch, AndAtomic | fetch, OrAtomic | fetch, XorAtomic | fetch,
-		xchgAtomic | fetch, cmpXchgAtomic | fetch:
+		FetchAdd, FetchAnd, FetchOr, FetchXor,
+		Xchg, CmpXchg:
 		switch size {
 		case Byte, Half:
 			// 8-bit and 16-bit atomic copy-modify-write atomics are not supported
@@ -100,17 +113,6 @@ func (op AtomicOp) OpCode(size Size) OpCode {
 
 // Mem emits `*(size *)(dst + offset) (op) src`.
 func (op AtomicOp) Mem(dst, src Register, size Size, offset int16) Instruction {
-	switch op {
-	case xchgAtomic, cmpXchgAtomic:
-		// XchgAtomic and CmpXchgAtomic always have fetch set, FetchMem must be used
-		return Instruction{
-			OpCode: InvalidOpCode,
-			Dst:    dst,
-			Src:    src,
-			Offset: offset,
-		}
-	}
-
 	return Instruction{
 		OpCode: op.OpCode(size),
 		Dst:    dst,
@@ -119,17 +121,10 @@ func (op AtomicOp) Mem(dst, src Register, size Size, offset int16) Instruction {
 	}
 }
 
-// FetchMem is like Mem but also stores the result in src.
-func (op AtomicOp) FetchMem(dst, src Register, size Size, offset int16) Instruction {
-	fetchOp := op | fetch
-	ins := fetchOp.Mem(src, dst, size, offset)
-	return ins
-}
-
 // Emits `lock-acquire dst = *(size *)(src + offset)`.
 func LoadAcquire(dst, src Register, size Size, offset int16) Instruction {
 	return Instruction{
-		OpCode: loadAcquireAtomic.OpCode(size),
+		OpCode: loadAcquire.OpCode(size),
 		Dst:    dst,
 		Src:    src,
 		Offset: offset,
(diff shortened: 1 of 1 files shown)


asm/load_store.go
```diff
@@ -223,3 +241,25 @@ func StoreXAdd(dst, src Register, size Size) Instruction {
...
+		OpCode:   StoreXAddOp(size), // We use Store here, for a Load, this is a quirk of the ISA
+		Dst:      dst,
+		Src:      src,
+		Offset:   offset,
+		Constant: int64(LoadAcquireAtomicFlag),
+	}
+}
+
+// StoreRelease emits an atomic store with release semantics.
+func StoreRelease(dst, src Register, offset int16, size Size) Instruction {
```
dylandreimerink, line 257: I kept the descriptions on these brief for now. Not sure how much context we need to give the user about what these instructions do.

asm/instruction.go
```diff
@@ -372,6 +372,31 @@ func (ins Instruction) Format(f fmt.State, c rune) {
...
 	case cls.isLoadOrStore():
+		// Special case StXClass | XAddMode + Constant > 0, where the actual op is determined by the
+		// imm value, and thus cannot be properly printed with the opcode alone.
+		if op.Class() == StXClass && op.Mode() == XAddMode && ins.Constant > 0 {
+			switch AtomicFlag(ins.Constant) {
+			case LoadAcquireAtomicFlag:
+				fmt.Fprint(f, "LdXMemAcq")
+			case StoreReleaseAtomicFlag:
+				fmt.Fprint(f, "StXMemRel")
+			}
```
lmb, line 383: What about the default case?

asm/load_store.go
```diff
@@ -70,6 +70,24 @@ func (s Size) Sizeof() int {
...
+// AtomicFlag is optionally set in the immediate field of StXClass | XAddMode instructions
+// to indicate further specialization of the instruction.
+type AtomicFlag uint32
+
+const (
+	// LoadAcquire indicates an atomic load with "acquire" memory ordering. It acts as a barrier to prevent
+	// reordering of memory operations. Reads and writes that happen after this instruction cannot be reordered
+	// to happen before it.
+	// All writes in other threads that release the same atomic variable are visible in the current thread.
+	LoadAcquireAtomicFlag AtomicFlag = 0x100
```
lmb, line 82: Are there other flags besides these? Why don't they start at 0 / 1?

asm/load_store.go
lmb: Being brief is fine!

asm/load_store.go
dylandreimerink: Not that I can find. I have no clue why this is. The discussion mailing list thread does not mention it at all, and the actual patch set says "As discussed in [1]" linking to the initial thread.

> As discussed in [1], introduce BPF instructions with load-acquire and
store-release semantics under -mcpu=v4. Define 2 new flags:
>
>BPF_LOAD_ACQ 0x100
>BPF_STORE_REL 0x110

Given the following snippet added to the kernel, there seem to be no other flags at the moment:
```c
/* Given a BPF_ATOMIC instruction @atomic_insn, return true if it is an
 * atomic load or store, and false if it is a read-modify-write instruction.
 */
static inline bool
bpf_atomic_is_load_store(const struct bpf_insn *atomic_insn)
{
	switch (atomic_insn->imm) {
	case BPF_LOAD_ACQ:
	case BPF_STORE_REL:
		return true;
	default:
		return false;
	}
}
```

asm/instruction.go
dylandreimerink: Good point, added a default case.

asm/load_store.go
dylandreimerink: Looking at the [LLVM patch](https://github.com/llvm/llvm-project/pull/108636/files), it seems like the intent is to have the 9th least significant byte be 1 when its not a read-modify-write instruction. Then have bits 8-4 hold an actual enum value, 0 and 1 in this case. So the next value would be `0x120`. Its an odd scheme.

lmb:
So for signed division we extended the opcode space to be 16 bits. Does it make sense to use some of that space to encode LoadAcquire / StoreRelease as actual opcodes as well? I'm a bit torn because of that whole scheme you pointed out. We might end up with 2^5 additional opcodes in the worst case? I'm guessing that not all combinations are valid though.

If we stick to the current approach, can we unexport AtomicFlags for now? That way we can punt on how they are encoded / whether we need further flags.

dylandreimerink:
> So for signed division we extended the opcode space to be 16 bits. Does it make sense to use some of that space to encode LoadAcquire / StoreRelease as actual opcodes as well?

Yes, I had forgotten we did that. I guess that makes more sense so we can avoid the special caseing in the string generation, which was the reason for doing this in the first place. I will give this a shot.

> I'm a bit torn because of that whole scheme you pointed out. We might end up with 2^5 additional opcodes in the worst case?

In theory. It seems that in practice there are 6 memory orderings [commonly defined in C/CPP](https://en.cppreference.com/w/cpp/atomic/memory_order), so with a bit of luck we will never come close to 2^5 additional ops.

What personally bothers me is that using odd schemes like offset and imm as part of the opcode is becoming more common. And while instructions are mostly still constant width, the encoding is moving away from the original and becoming less consistent.

asm/load_store.go
dylandreimerink: I just noticed https://elixir.bootlin.com/linux/v6.14.1/source/kernel/bpf/verifier.c#L575

```
static bool is_cmpxchg_insn(const struct bpf_insn *insn)
{
	return BPF_CLASS(insn->code) == BPF_STX &&
	       BPF_MODE(insn->code) == BPF_ATOMIC &&
	       insn->imm == BPF_CMPXCHG;
}
```

So, that is why the lower imm portion isn't used https://elixir.bootlin.com/linux/v6.14.1/source/include/uapi/linux/bpf.h#L49-L52
```
/* atomic op type fields (stored in immediate) */
#define BPF_FETCH	0x01	/* not an opcode on its own, used to build others */
#define BPF_XCHG	(0xe0 | BPF_FETCH)	/* atomic exchange */
#define BPF_CMPXCHG	(0xf0 | BPF_FETCH)	/* atomic compare-and-write */
```

asm/load_store.go
```diff
@@ -210,16 +244,109 @@ func StoreImm(dst Register, offset int16, value int64, size Size) Instruction {
...
+}
+
 // StoreXAddOp returns the OpCode to atomically add a register to a value in memory.
 func StoreXAddOp(size Size) OpCode {
-	return OpCode(StXClass).SetMode(XAddMode).SetSize(size)
+	return AtomicOp(size, AddAtomic, false)
 }
 
-// StoreXAdd atomically adds src to *dst.
+// AtomicAnd atomically adds src to *(dst + offset), storing the old value in src if fetch is true.
```
kevsecurity, line 269: s/AtomicAnd/AtomicAdd/

kevsecurity (approved):

asm/opcode.go
```diff
@@ -71,24 +71,29 @@ func (cls Class) isJumpOrALU() bool {
...
 //
-// For LdClass, LdXclass, StClass and StXClass:
+// For LdClass, LdXclass, StClass:
 //
-//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
-//	        0      | MDE |SIZ| CLS
+//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
+//	                       0                       | MDE |SIZ| CLS
+//
+// For StXClass:
```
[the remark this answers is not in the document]
dylandreimerink: Right, I will make an additional line.

lmb:
@dylandreimerink I pushed a commit which cuts down the repetition a bit. IIRC I added the free functions for Load, etc. because there wasn't that much variation. That isn't the case with the atomic ops so using a method on the type seems nicer?

```
asm.AddAtomic.Mem()
asm.AddAtomic.FetchMem()
```

read okay to me, but I'm not 100% sure they are really correct. LoadAcquire and StoreRelease are also annoying.

asm/load_store.go
```diff
@@ -26,10 +26,65 @@ const (
...
+	AddAtomic AtomicOp = AtomicOp(Add) << 8
+	// AndAtomic - bitwise AND src with memory address at dst atomically
+	AndAtomic AtomicOp = AtomicOp(And) << 8
+	// OrAtomic - bitwise OR src with memory address at dst atomically
+	OrAtomic AtomicOp = AtomicOp(Or) << 8
+	// XorAtomic - bitwise XOR src with memory address at dst atomically
+	XorAtomic AtomicOp = AtomicOp(Xor) << 8
+
+	// LdAcqAtomic - atomically load with acquire semantics
+	LdAcqAtomic AtomicOp = 0x0001_0000
```
lmb, line 49: Maybe we just go with `LoadAcquire` and `StoreRelease` since we don't need to distinguish from Add and Xor, etc?

asm/opcode.go
```diff
@@ -71,24 +71,29 @@ func (cls Class) isJumpOrALU() bool {
...
+//	              0              |    ATOMIC OP    | MDE |SIZ| CLS
 //
 // For JumpClass, Jump32Class:
 //
-//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
-//	        0      |  OPC  |S| CLS
-type OpCode uint16
+//	7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0 7 6 5 4 3 2 1 0
+//	                       0                       |  OPC  |S| CLS
+type OpCode uint32
```
lmb, line 96: Nit: we could stick to 16bit now? Though 32bit is also nice, because we will never run out of space with that I think.

lmb: Going to 32bit doesn't change size of Instruction, so this is probably fine.

dylandreimerink: We still need the 9th bit. For Atomic ops, we would use the lower 2 bytes, and the lowest bit of the upper 2 bytes.
So the mask is `0x0001FFFF`. 

We could likely avoid it if we take advantage of the fact that for atomics only 4 ALU instructions are allowed and renumber everything so it fits in in a single byte. But having the larger opcode is the least intrusive.

asm/load_store.go
dylandreimerink: I guess the main issue is the type. If we do not define them, then the user must. And not all ALU operations are supported. Seem more user friendly to have the `AtomicAdd`, `AtomicAnd`, `AtomicOr` and `AtomicXor`.

I agree we don't need the `Fetch` variants, they were mostly there for the string generation. But we can solve that with some logic.

dylandreimerink:
> I pushed a commit which cuts down the repetition a bit. IIRC I added the free functions for Load, etc. because there wasn't that much variation. That isn't the case with the atomic ops so using a method on the type seems nicer?

Ah, I think this is better. I had all of the different ops spelled out for the string generation. But that is something we can fix by checking for the `Fetch` bit manually.

> read okay to me, but I'm not 100% sure they are really correct. LoadAcquire and StoreRelease are also annoying.

This works for the copy-modify-write style of atomics. We need to add a guard for `XCHG` and `CMPXCHG` since they only have `Fetch` variants. 

The `LoadAcquire` and `StoreRelease` are essentially a different class of atomic op. So we could given them dedicated functions, or  only allow them to be used with `Mem` since they never have the `Fetch` bit set.

It technically works, its just that from a UX perspective they don't fit neatly in the `*(size *)(dst + offset) (op) src` description, they really are:

* Load - `dst = lock-acquire *(size *)(src + offset)`
* Store - `lock-release *(size *)(dst + offset) = src`

asm/load_store.go
```diff
@@ -26,10 +28,119 @@ const (
...
+
+type AtomicOp uint32
+
+const (
+	InvalidAtomic AtomicOp = 0xffff_ffff
+
+	// AddAtomic - add src to memory address dst atomically
+	AddAtomic AtomicOp = AtomicOp(Add) << 8
+	// FetchAdd - add src to memory address dst atomically, store result in src
+	FetchAdd AtomicOp = AddAtomic | fetch
```
lmb, line 45: @dylandreimerink I exported these. Then you can just write `asm.FetchAdd.Mem(...)` which is shorter than `asm.AddAtomic.Fetch(...)`. I suspect the Fetch variants might be more commonly used anyways. What do you think?

asm/load_store.go
```diff
@@ -26,10 +28,119 @@ const (
...
+	FetchOr AtomicOp = OrAtomic | fetch
+	// XorAtomic - bitwise XOR src with memory address at dst atomically
+	XorAtomic AtomicOp = AtomicOp(Xor) << 8
+	// FetchXor - bitwise XOR src with memory address at dst atomically, store result in src
+	FetchXor AtomicOp = XorAtomic | fetch
+
+	// Xchg - atomically exchange the old value with the new value
+	//
+	// src gets populated with the old value of *(size *)(dst + offset).
+	Xchg AtomicOp = 0x0000_e000 | fetch
```
lmb, line 62: Same here.

asm/load_store.go
dylandreimerink: Yes, this seems nice!

dylandreimerink:
I am happy this this if you are. I say squash+merge :rocket:

lmb:
Thanks for iterating with me!

PR #1755, opened 2025-04-16
[PoC] Introduce new flag `SpecCacheDisabled` & Parse only the requires BTF types

burak-ok:
Based on https://github.com/cilium/ebpf/pull/1589

This is a proof-of-concept. The code is not ready for merging but it shows it is possible to significantly reduce the memory consumption (by 27MB).

---

This PR aims to lower the memory footprint when using the `cilium/ebpf` library. This is achieved in two ways:
1. Reducing memory usage: Adding a new flag which disables caching of the BTF spec
2. Parsing only the needed BTF types

The tradeoff is for lowering the memory footprint is of course performance while loading eBPF programs etc...

In this PoC when using the new `SpecCacheDisabled` flag (1) it will also automatically check which BTF types are needed and load/parse only those (2)

---

Benchmarks:
1. `ParseVmlinux` is always the base for both tests
2. The `new.txt` Inspektor gadget run is with the types from https://github.com/cilium/ebpf/pull/1755/#issuecomment-2820919961
3. The `new.txt` `ParseVmlinux` run is called `BenchmarkParseVmlinuxWithoutFilter` in this PR
```
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz
                     │   old.txt   │               new.txt                │
                     │   sec/op    │   sec/op     vs base                 │
ParseVmlinux-8         29.28m ± 3%   60.62m ± 8%  +107.06% (p=0.000 n=10)
InspektorGadget-8      29.28m ± 3%   12.94m ± 3%   -55.79% (p=0.000 n=10)
geomean                29.28m        28.01m         -4.32%

                     │   old.txt    │                new.txt                │
                     │     B/op     │     B/op      vs base                 │
ParseVmlinux-8         24.43Mi ± 0%   53.58Mi ± 0%  +119.37% (p=0.000 n=10)
InspektorGadget-8      24.43Mi ± 0%   10.89Mi ± 0%   -55.40% (p=0.000 n=10)
geomean                24.43Mi        24.16Mi         -1.08%

ParseVmlinux-8         271.9k ± 0%   365.1k ± 0%  +34.30% (p=0.000 n=10)
InspektorGadget-8      271.9k ± 0%   110.1k ± 0%  -59.51% (p=0.000 n=10)
geomean                271.9k        200.5k       -26.26%
```

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -179,6 +179,13 @@ type Spec struct {
 	strings *stringTable
 }
 
+// SpecOptions control loading a BTF spec
+type SpecOptions struct {
+	// Set of type names to include in the spec. Dependencies will be included.
+	// If empty, all types are included.
+	TypeNames map[string]struct{}
+}
+
 // LoadSpec opens file and calls LoadSpecFromReader on it.
 func LoadSpec(file string) (*Spec, error) {
 	fh, err := os.Open(file)
@@ -195,16 +202,24 @@ func LoadSpec(file string) (*Spec, error) {
 // Returns ErrNotFound if reading from an ELF which contains no BTF. ExtInfos
 // may be nil.
 func LoadSpecFromReader(rd io.ReaderAt) (*Spec, error) {
+	return LoadSpecFromReaderWithOptions(rd, nil)
+}
+
+// LoadSpecFromReader reads from an ELF or a raw BTF blob.
+//
+// Returns ErrNotFound if reading from an ELF which contains no BTF. ExtInfos
+// may be nil.
+func LoadSpecFromReaderWithOptions(rd io.ReaderAt, opts *SpecOptions) (*Spec, error) {
 	file, err := internal.NewSafeELFFile(rd)
 	if err != nil {
 		if bo := guessRawBTFByteOrder(rd); bo != nil {
-			return loadRawSpec(io.NewSectionReader(rd, 0, math.MaxInt64), bo, nil)
+			return loadRawSpec(io.NewSectionReader(rd, 0, math.MaxInt64), bo, nil, nil)
 		}
 
 		return nil, err
 	}
 
-	return loadSpecFromELF(file)
+	return loadSpecFromELF(file, opts)
 }
 
 // LoadSpecAndExtInfosFromReader reads from an ELF.
@@ -217,7 +232,7 @@ func LoadSpecAndExtInfosFromReader(rd io.ReaderAt) (*Spec, *ExtInfos, error) {
 		return nil, nil, err
 	}
 
-	spec, err := loadSpecFromELF(file)
+	spec, err := loadSpecFromELF(file, nil)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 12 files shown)


burak-ok:
> Seems like vmlinux spec keeps being a problem! Just checking: when you say memory usage you mean heap at idle? Calling https://pkg.go.dev/github.com/cilium/ebpf/btf#FlushKernelSpec does not help?

Yes, that would help for having a lower heap usage after starting the program. But if one sets a low memory limit in a pod spec, one also needs to avoid high memory while initializing -> while loading every program.

Furthermore with `FlushKernelSpec` we need to find and identify every possible call into the `ebpf` library which might parse and save the while KernelSpec. With a flag we can set the flag and forget about flushing the kernel spec.

> How would you determine which types to parse from vmlinux?

For that we are reading the all `relo.TypeNames` out of the program where the relocations are getting applied: https://github.com/cilium/ebpf/pull/1755/files#diff-981ef293a9c93614e843135eb5b207f951babe5604aee33c72f66567ccaa01de

ti-mo:
@lmb To me it sounds like lazy-decoding could've been a better avenue to explore after all?

Not sure how (in)feasible it is today, but iirc we had btf.Spec.Add() back in the day which was a blocker. Now we have btf.Builder, we could technically, hypothetically, make btf.Spec a querying layer over an encoded btf blob and only inflate what's queried, and cache the results to enable type comparisons. Or, implement comparers on all types if we don't want to cache anything. Seems like some (most?) users care more about keeping both resident and peak memory usage low rather than speed.

lmb:
My main concern is / was complexity of a lazy decoder. The whole "fixups" concept needs to be redone... I think you are right that peak usage seems more important.

I see two avenues: there is some perf to be gained by not unmarshaling into an interface for rawType, I think. Two is the lazy decode you mentioned. I still have some old proof of concepts lying around, I'll push those somewhere.

lmb:
@burak-ok could you come up with a list of types which you most frequently need from vmlinux? That way we can add a benchmark we can start optimising against. Right now that benchmark is decoding all of vmlinux which isn't useful.

lmb:
@burak-ok can you take a look at https://github.com/cilium/ebpf/pull/1763? Doesn't address FlushKernelSpec, but I think that might be better done by relying on `weak` in Go 1.24.

burak-ok:
> @burak-ok can you take a look at #1763?

I added roughly the same benchmarks that you posted for this PR in the top post.
Comparing these results your PR saves roughly the same amount of memory and allocations for the `InspektorGadget` benchmark.
But the `ParseVmlinux` Benchmark shows that my Draft PR has more disadvantages for the general usecase of reading and parsing every type.

I'll take a deeper look into your PR some time later, thanks for opening it.

> Doesn't address FlushKernelSpec, but I think that might be better done by relying on `weak` in Go 1.24.

With having memory limits I think this would be the best case scenario for us.

lmb:
The new lazy BTF code is in. @burak-ok could you try the code and report back how much of a difference it makes?

PR #1761, opened 2025-04-21
btf: use append-style encoding when marshaling

lmb:
The Builder currently uses binary.Write with a shared bytes.Buffer. Replace this with binary.Append and a byte slice.

This drops the number of allocations by 99% and gives a small speed up in benchmarks. The overall amount of memory allocated increases a little bit, but I can't figure out why.

The []byte is threaded through the encoder.deflate* methods because assigning to a shared buffer in encoder incurs write barrier overheads.

    core: 1
    goos: linux
    goarch: amd64
                 │  base.txt   │            append3.txt            │
                 │   sec/op    │   sec/op     vs base              │
    Marshaler      11.31m ± 1%   11.41m ± 2%  +0.93% (p=0.041 n=6)
    BuildVmlinux   140.7m ± 4%   145.6m ± 1%  +3.50% (p=0.002 n=6)
    geomean        39.89m        40.77m       +2.21%

                 │   base.txt   │             append3.txt             │
                 │     B/op     │     B/op      vs base               │
    Marshaler      3.286Mi ± 0%   3.469Mi ± 0%   +5.58% (p=0.002 n=6)
    BuildVmlinux   37.31Mi ± 0%   42.22Mi ± 0%  +13.16% (p=0.002 n=6)
    geomean        11.07Mi        12.10Mi        +9.31%

                 │   base.txt    │            append3.txt             │
                 │   allocs/op   │  allocs/op   vs base               │
    Marshaler       12627.0 ± 0%    208.0 ± 0%  -98.35% (p=0.002 n=6)
    BuildVmlinux   165.689k ± 0%   1.508k ± 0%  -99.09% (p=0.002 n=6)
    geomean          45.74k         560.1       -98.78%


lmb:
Overall I'm a bit confused by the outcome of this PR. I'd have thought that dropping allocations this much would give a good boost to runtime, but it almost has a negative effect. Can't figure it out. Maybe one of you has an idea?

btf/btf_types.go
```diff
@@ -268,30 +269,14 @@ func (bt *btfType) SetSize(size uint32) {
...
-	}
-
-	return binary.Write(w, bo, rt.data)
+func (bt *btfType) Append(buf []byte, bo binary.ByteOrder) []byte {
+	buf = slices.Grow(buf, int(unsafe.Sizeof(*bt)))
+	pos := len(buf)
+	h := buf[pos : pos+int(unsafe.Sizeof(*bt))]
+	bo.PutUint32(h[0:], bt.NameOff)
+	bo.PutUint32(h[4:], bt.Info)
+	bo.PutUint32(h[8:], bt.SizeType)
```
dylandreimerink, line 278: I am a little confused by this. We grow the capacity of the slice, and then add the header in the added capacity, but we never increase the length of the slice. So we are hiding it away, next time an `append` happens on `buf` that header is overwritten. https://go.dev/play/p/_VD2CgylFGm


This makes more sense:
```go
func (bt *btfType) Append(buf []byte, bo binary.ByteOrder) []byte {
	buf = append(buf, make([]byte, int(unsafe.Sizeof(*bt)))...)
	pos := len(buf)
	h := buf[pos-int(unsafe.Sizeof(*bt)) : pos]
	bo.PutUint32(h[0:], bt.NameOff)
	bo.PutUint32(h[4:], bt.Info)
	bo.PutUint32(h[8:], bt.SizeType)
	return buf
}
```

Then at the usage site you can go:
```go
buf = raw.Append(buf, e.Order)
```

Instead of:
```go
header := buf[start : start+int(unsafe.Sizeof(raw))]
_ = raw.Append(header[:0], e.Order)
```

dylandreimerink: Ah, I see now why this is the way it is, due to the having to reserve the room for the header, then writing to it later.

So, can't we simply make it:
```go
func (bt *btfType) Write(buf []byte, bo binary.ByteOrder) {
	bo.PutUint32(h[0:], bt.NameOff)
	bo.PutUint32(h[4:], bt.Info)
	bo.PutUint32(h[8:], bt.SizeType)
}
```
Since we are not appending, simply writing to a slice. Then on the call site do?:
```go
raw.Write(buf[start : start+int(unsafe.Sizeof(raw))], e.Order)
```

dylandreimerink (changes requested):

btf/btf_types.go
lmb: Thanks! Went with `Encode` instead, due to https://pkg.go.dev/encoding/binary#Encode

dylandreimerink (approved):

Issue #739, opened 2022-07-19
btf: libbpf doesn't update instructions when deduplicating types


lmb:
```
--- FAIL: TestLibBPFCompat (0.01s)
    --- FAIL: TestLibBPFCompat/netif_receive_skb.linked3.o (0.01s)
        elf_reader_test.go:665: Error during loading: program trace_netif_receive_skb: apply CO-RE relocations: apply fixup target_type_id=67->16253: invalid immediate 73, expected 67 (fixup: target_type_id=67->16253)
```

`target_type_id=67->16253` means we expect to find type id `67` in `Instruction.Constant`. Instead we find `invalid immediate 73`.

We can dump the BTF to figure out what types this refers to:

```
$ bpftool btf dump file netif_receive_skb.linked3.o format raw | grep -F '[67]'
[67] STRUCT 'bpf_insn' size=8 vlen=5
$ bpftool btf dump file netif_receive_skb.linked3.o format raw | grep -F '[73]'
[73] STRUCT 'btf_enum' size=8 vlen=2
```

We think the instruction should reference a `struct bpf_insn` but the compiler encoded `struct btf_enum`. After adding some advanced printf debugging:

```
$ ./run-tests.sh 5.17 -run TestLibBPFCompat/netif_receive_skb.linked3.o . | grep -A1 -B1 302
offset 18 target_type_id 19 (from wire)
offset 302 target_type_id 67 (from wire)
offset 359 target_type_id 2 (from wire)
--
offset 18 target_type_id of Struct:"sk_buff" (0) local type id 19 (from ext info)
offset 302 target_type_id of Struct:"bpf_insn" (0) local type id 67 (from ext info)
offset 359 target_type_id of Int:"int" (0) local type id 2 (from ext info)
--
offset 18 LdImmDW dst: r2 imm: 19 target_type_id of Struct:"sk_buff" (0) (into metadata)
offset 302 LdImmDW dst: r3 imm: 73 target_type_id of Struct:"bpf_insn" (0) (into metadata)
offset 359 LdImmDW dst: r3 imm: 2 target_type_id of Int:"int" (0) (into metadata)
--
offset 18 LdImmDW dst: r2 imm: 19 target_type_id of Struct:"sk_buff" (0) (from metadata)
offset 302 LdImmDW dst: r3 imm: 73 target_type_id of Struct:"bpf_insn" (0) (from metadata)
[...]

PR #1762, opened 2025-04-21
bpf2go: support multiple source files

### LINKED ISSUE #1758: bpf2go: support multiple source files
## Background and Motivation

Currently, the bpf2go tool only supports compiling a single source file at a time. This limits our ability to create modular eBPF programs composed of multiple source files. Supporting multiple source files would enable better code organization, reusability, and testability for eBPF programs -- improving the developer experience.

## Use Cases

1. **Modular Components**: Develop reusable eBPF components that can be shared across multiple programs. Complex eBPF programs could be broken into logical components with separate source files
2. **Testability**: Create test harnesses for eBPF code without including test code in production builds

## Concrete Example - RMID Allocation in Unvariance Collector

In the [Unvariance Collector project](https://github.com/unvariance/collector), we need to allocate Resource Monitoring IDs (RMIDs) to processes. These RMIDs come from a limited space (typically 0 to ~460 on certain processors). We want to:

1. Create a library for RMID allocation/deallocation with a free list
2. Track metadata for each RMID and maintain state consistency
3. Test this functionality in isolation
4. Include this code in the main collector without the test harness

Ideally, we would:
- Develop the RMID allocation as a self-contained package with its own header file
- Create a test harness in a separate file that includes and tests this functionality
- Link both during testing
- Link just the RMID allocation code (without test harness) with the main collector

[unvariance/collector#109](https://github.com/unvariance/collector/issues/106?issue=unvariance%7Ccollector%7C109) for more info

## Proposed Implementation

After examining the codebase (particularly in cmd/bpf2go), I suggest the following changes:

1. **Modify the `b2g` struct** in main.go to accept multiple source files:
   - Change `SourceFile string` to `SourceFiles []string`
   - Update the command-line parsing to collect multiple positional arguments

2. **Refactor the convert() function**:
   - Extract the compilation logic in `convert` function into `compileOne`
   - When multiple source files are provided, compile each one in sequence
   - This separation allows compiling multiple source files before linking

3. **Add a linking step** when multiple source files are provided:
   - Create a new function in the gen package (e.g., `Link`) similar to existing `Compile`
   - The function would invoke `bpftool gen object` to combine multiple object files into one
   - Add the linking step when multiple source files are provided

4. **Handle dependencies correctly** when multiple files are used:
   - Parse dependencies for each source file (already present)
   - Merge the dependencies into a single set for the final object file (new functionality)
   - Write dependencies (already present)

## Backward Compatibility

When only one source file is provided, the tool should behave exactly as it does now. This ensures backward compatibility with existing workflows.

## Related Issues and Discussions

- [[Discussion #1142](https://github.com/cilium/ebpf/discussions/1142)](https://github.com/cilium/ebpf/discussions/1142): This discussion considered generating Go code from a given object file, which would support this use case by allowing users to create multiple object files out-of-band and use bpftool to link them together. However, as mentioned in that discussion, this approach would require users to understand internal plumbing and handle target-specific details manually. Our proposal solves this by using bpf2go to build all object files automatically without exposing internal complexity.

- [[Issue #466](https://github.com/cilium/ebpf/issues/466)](https://github.com/cilium/ebpf/issues/466): This issue calls for support of BPF object linking and explicitly mentions bpftool gen object, which we intend to use. However, it focuses specifically on handling weak symbols. While this remains an open issue, users who don't use weak symbols shouldn't encounter these problems with our proposed multi-file support.

I'd be happy to work on implementing this feature. Looking forward to your feedback!

yonch:
See issue #1758

Has been used successfully in unvariance/collector#109, PR unvariance/collector#110.

Tested in a dev container on MacOs and on a Linux bare metal box.

Closes #1758

### DIFF
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -21,13 +21,14 @@ import (
 	"github.com/cilium/ebpf/cmd/bpf2go/gen"
 )
 
-const helpText = `Usage: %[1]s [options] <ident> <source file> [-- <C flags>]
+const helpText = `Usage: %[1]s [options] <ident> <source files...> [-- <C flags>]
 
 ident is used as the stem of all generated Go types and functions, and
 must be a valid Go identifier.
 
-source is a single C file that is compiled using the specified compiler
-(usually some version of clang).
+source files are C files that are compiled using the specified compiler
+(usually some version of clang) and linked together into a single
+BPF program.
 
 You can pass options to the compiler by appending them after a '--' argument
 or by supplying -cflags. Flags passed as arguments take precedence
@@ -60,8 +61,8 @@ func run(stdout io.Writer, args []string) (err error) {
 type bpf2go struct {
 	stdout  io.Writer
 	verbose bool
-	// Absolute path to a .c file.
-	sourceFile string
+	// Absolute paths to .c files.
+	sourceFiles []string
 	// Absolute path to a directory where .go are written
 	outputDir string
 	// Alternative output stem. If empty, identStem is used.
@@ -186,13 +187,18 @@ func newB2G(stdout io.Writer, args []string) (*bpf2go, error) {
 
 	b2g.identStem = args[0]
 
-	sourceFile, err := filepath.Abs(args[1])
-	if err != nil {
-		return nil, err
+	sourceFiles := args[1:]
+	b2g.sourceFiles = make([]string, len(sourceFiles))
+	for i, source := range sourceFiles {
+		absPath, err := filepath.Abs(source)
+		if err != nil {
+			return nil, fmt.Errorf("convert source file to absolute path: %w", err)
+		}
+		b2g.sourceFiles[i] = absPath
 	}
-	b2g.sourceFile = sourceFile
 
 	if b2g.makeBase != "" {
+		var err error
 		b2g.makeBase, err = filepath.Abs(b2g.makeBase)
 		if err != nil {
 			return nil, err
@@ -298,10 +304,13 @@ func getBool(key string, defaultVal bool) bool {
 }
 
 func (b2g *bpf2go) convertAll() (err error) {
-	if _, err := os.Stat(b2g.sourceFile); os.IsNotExist(err) {
-		return fmt.Errorf("file %s doesn't exist", b2g.sourceFile)
-	} else if err != nil {
-		return err
+	// Check all source files exist
+	for _, source := range b2g.sourceFiles {
+		if _, err := os.Stat(source); os.IsNotExist(err) {
+			return fmt.Errorf("file %s doesn't exist", source)
+		} else if err != nil {
+			return err
+		}
 	}
 
 	if !b2g.disableStripping {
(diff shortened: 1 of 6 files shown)


ti-mo (commented):
This looks good overall, thank you! Left a few nits.

cmd/bpf2go/main_test.go
```diff
@@ -406,6 +407,98 @@ func TestParseArgs(t *testing.T) {
 	})
 }
 
+func TestMultipleSourceFiles(t *testing.T) {
+	modRoot, err := filepath.Abs("../..")
+	if err != nil {
+		t.Fatal("Can't get module root:", err)
+	}
+
+	if _, err := os.Stat(filepath.Join(modRoot, "go.mod")); os.IsNotExist(err) {
```
lmb, line 416: Can you move creating the temporary module into a helper instead of copying the code?

cmd/bpf2go/makedep_test.go
```diff
@@ -59,3 +60,34 @@ nothing:
...
+		return merged[i].file < merged[j].file
+	})
+
+	want := []dependency{
+		{"/foo/main.c", []string{"/foo/bar.h", "/foo/baz.h", "/foo/qux.h"}},
+		{"/foo/other.c", []string{"/foo/bar.h"}},
+		{"/foo/third.c", []string{"/foo/bar.h"}},
+	}
+
+	if !reflect.DeepEqual(merged, want) {
```
lmb, line 88:
```suggestion
	qt.Assert(t, qt.DeepEquals(merged, want))
```

cmd/bpf2go/makedep.go
```diff
@@ -106,3 +106,48 @@ func parseDependencies(baseDir string, in io.Reader) ([]dependency, error) {
...
+	// Map to track merged dependencies by file
+	merged := make(map[string][]string)
+
+	// Process each slice of dependencies
+	for _, deps := range depsSlices {
+		for _, dep := range deps {
+			// If we've seen this file before, merge prerequisites
+			if existing, ok := merged[dep.file]; ok {
+				// Combine prerequisites, avoiding duplicates
+				prereqs := make(map[string]struct{})
```
lmb, line 122: I think you might be able to simplify this by appending to a slice. At the end, sort the slice and use https://pkg.go.dev/slices#Compact in a second pass.

cmd/bpf2go/makedep_test.go
```diff
@@ -59,3 +60,34 @@ nothing:
...
+
+	merged := mergeDependencies(deps1, deps2)
+
+	// Sort merged dependencies for stable comparison
+	sort.Slice(merged, func(i, j int) bool {
+		return merged[i].file < merged[j].file
+	})
+
+	want := []dependency{
+		{"/foo/main.c", []string{"/foo/bar.h", "/foo/baz.h", "/foo/qux.h"}},
```
lmb, line 83: How can we get different dependencies for two compilations of the same file? Wouldn't it be more appropriate to check that both sets of dependencies are identical? Or that the sets are disjoint?

cmd/bpf2go/main.go
```diff
@@ -186,13 +187,18 @@ func newB2G(stdout io.Writer, args []string) (*bpf2go, error) {
...
-	if err != nil {
-		return nil, err
+	sourceFiles := args[1:]
+	b2g.sourceFiles = make([]string, len(sourceFiles))
+	for i, source := range sourceFiles {
+		absPath, err := filepath.Abs(source)
+		if err != nil {
+			return nil, fmt.Errorf("convert source file to absolute path: %w", err)
+		}
+		b2g.sourceFiles[i] = absPath
```
lmb, line 197: Nit: just appending to b2g.sourceFiles without pre-allocating the slice is probably fine / easier to reason about.


```suggestion
		b2g.sourceFiles = append(b2g.sourceFiles, absPath)
```

lmb (changes requested):
Agree with Timo, this is nice :) Be aware that libbpf linking may produce linked object files which we don't parse: https://github.com/cilium/ebpf/issues/739 I don't think it's a pre-requisite to get this fixed, but something to be aware of which might make this less useful to you.

ti-mo:
@yonch I'm closing some stale PRs, please ping if/when you want to pick this up again.

PR #1764, opened 2025-04-29
info: add memlock field to ProgramInfo

haozhangami:
This patch adds a ProgramInfo.Memlock() method, the memlock field is obtained from fdinfo.

### DIFF
--- a/info.go
+++ b/info.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"reflect"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/cilium/ebpf/asm"
@@ -258,16 +257,17 @@ type ProgramInfo struct {
 	numLineInfos uint32
 	funcInfos    []byte
 	numFuncInfos uint32
+
+	memlock uint64
 }
 
 func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 	var info sys.ProgInfo
-	err := sys.ObjInfo(fd, &info)
-	if errors.Is(err, syscall.EINVAL) {
-		return newProgramInfoFromProc(fd)
-	}
-	if err != nil {
-		return nil, err
+	err1 := sys.ObjInfo(fd, &info)
+	// EINVAL means the kernel doesn't support BPF_OBJ_GET_INFO_BY_FD. Continue
+	// with fdinfo if that's the case.
+	if err1 != nil && !errors.Is(err1, unix.EINVAL) {
+		return nil, fmt.Errorf("getting object info: %w", err1)
 	}
 
 	typ, err := ProgramTypeForPlatform(platform.Native, info.Type)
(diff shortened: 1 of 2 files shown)


info.go
```diff
@@ -388,29 +396,23 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 	return &pi, nil
 }
 
-func newProgramInfoFromProc(fd *sys.FD) (*ProgramInfo, error) {
-	var info ProgramInfo
-	var progType uint32
+func newProgramInfoFromProc(fd *sys.FD, pi *ProgramInfo) error {
```
ti-mo, line 399: As in https://github.com/cilium/ebpf/commit/21be59580a51a37397d2041c46d33c7c529b35b8, this should be named `readProgramInfoFromProc()`.

info.go
```diff
@@ -388,29 +396,23 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
...
+		return &internal.UnsupportedFeatureError{
 			Name:           "reading program info from /proc/self/fdinfo",
 			MinimumVersion: internal.Version{4, 10, 0},
 		}
 	}
 	if err != nil {
-		return nil, err
-	}
-
-	info.Type, err = ProgramTypeForPlatform(platform.Linux, progType)
```
ti-mo, line 408: This cannot just be removed.

info.go
```diff
@@ -289,6 +289,14 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 		jitedSize:            info.JitedProgLen,
 		loadTime:             time.Duration(info.LoadTime),
 		verifiedInstructions: info.VerifiedInsns,
+		memlock:              0,
+	}
+
+	// Supplement OBJ_INFO with data from /proc/self/fdinfo. It contains fields
+	// like memlock that is not present in OBJ_INFO.
+	err2 := newProgramInfoFromProc(fd, &pi)
```
ti-mo, line 297: As in https://github.com/cilium/ebpf/commit/21be59580a51a37397d2041c46d33c7c529b35b8, the reason for err1 and err2 existing is to be able to error if both approaches fail, but you added no such check at the bottom of the function. The result of `makeSecondCall` should go into `err1`.

info_test.go
```diff
@@ -160,8 +160,11 @@ func TestProgramInfo(t *testing.T) {
 func TestProgramInfoProc(t *testing.T) {
 	spec := fixupProgramSpec(basicProgramSpec)
 	prog := mustNewProgram(t, spec, nil)
+	info, err := prog.Info()
```
ti-mo, line 163: The point of this test is to test procinfo in isolation, since it's supported on older kernels, and objinfo isn't. Adding prog.Info() and the subsequent skip will reduce coverage on those kernels.

Just instantiate a new ProgramInfo here and populate it using `readProgramInfoFromProc`.

Also, validateProgInfo needs to be extended to check for memlock, otherwise we're not testing the new field anywhere.

info.go
```diff
@@ -289,6 +289,14 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 		jitedSize:            info.JitedProgLen,
 		loadTime:             time.Duration(info.LoadTime),
 		verifiedInstructions: info.VerifiedInsns,
+		memlock:              0,
```
ti-mo, line 292: Why initialize this field?

ti-mo (changes requested):
Thanks for picking this up! This is good to go with a few changes.

info.go
haozhangami: Done

info.go
haozhangami: Done

info_test.go
haozhangami: Done, use a new ProgramInfo and add check for memelock field

ti-mo (approved):
Thank you!

PR #1769, opened 2025-04-30
bpf2go, btf: emit structs.HostLayout in GoFormatter

### LINKED ISSUE #1686: bpf2go: emit `structs.HostLayout` into generated structs
We recently made a change to make sure our syscall wrappers continue to work on future versions of go, see #1681. Timo points out that we should do the same for bpf2go:

              Looks good, thanks! We should probably also emit these in bpf2go types.

_Originally posted by @ti-mo in https://github.com/cilium/ebpf/pull/1681#pullrequestreview-2620432502_
          
I think it might make sense to just update `GoFormatter` to always include `HostLayout` in structs instead of the hack I did for `gentypes`.

lmb:
Types described by BTF essentially follow C rules for laying out a struct. Use the newly introduced structs.HostLayout to hint at the compiler that reordering is not permissible.

Fixes #1686

### DIFF
--- a/cmd/bpf2go/gen/output.go
+++ b/cmd/bpf2go/gen/output.go
@@ -161,20 +161,31 @@ func Generate(args GenerateArgs) error {
 		Identifier: args.Identifier,
 	}
 
+	var typeDecls []string
+	needsStructsPkg := false
+	for _, typ := range types {
+		name := typeNames[typ]
+		decl, err := gf.TypeDeclaration(name, typ)
+		if err != nil {
+			return fmt.Errorf("generating %s: %w", name, err)
+		}
+		_, ok := typ.(*btf.Struct)
+		needsStructsPkg = needsStructsPkg || ok
+		typeDecls = append(typeDecls, decl)
+	}
+
 	ctx := struct {
-		*btf.GoFormatter
-		Module      string
-		Package     string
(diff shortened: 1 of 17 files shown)


cmd/bpf2go/gen/output.go
```diff
@@ -161,30 +161,40 @@ func Generate(args GenerateArgs) error {
...
 
+	var typeDecls []string
+	needsStructsPkg := false
+	for _, typ := range types {
+		name := typeNames[typ]
+		decl, err := gf.TypeDeclaration(name, typ)
+		if err != nil {
+			return fmt.Errorf("generating %s: %w", name, err)
+		}
+		needsStructsPkg = needsStructsPkg || strings.Contains(decl, " _ structs.HostLayout; ")
```
ti-mo, line 172: This can't quietly cause a regression, but I'd prefer a type assertion on `*btf.Struct` instead. Matching on a string including whitespace looks a little fragile in any case.

ti-mo (approved):
Nice! One nit, and you forgot to update the test battery in `TestGoTypeDeclaration`.

cmd/bpf2go/gen/output.go
lmb: Great idea, thanks!

PR #1774, opened 2025-05-06
kfunc: poison all weak kfunc calls if the kernel lacks BTF

bogushevich:
KFuncs fixing up loads only system BTF, so using custom (non-system) BTF specification with KFuncs leads to errors.

### DIFF
--- a/linker.go
+++ b/linker.go
@@ -300,9 +300,15 @@ func fixupKfuncs(insns asm.Instructions) (_ handles, err error) {
 	return nil, nil
 
 fixups:
-	// only load the kernel spec if we found at least one kfunc call
+	// Only load kernel BTF if we found at least one kfunc call. kernelSpec can be
+	// nil if the kernel does not have BTF, in which case we poison all kfunc
+	// calls.
 	kernelSpec, err := btf.LoadKernelSpec()
-	if err != nil {
+	// ErrNotSupportedOnOS wraps ErrNotSupported, check for it first.
+	if errors.Is(err, internal.ErrNotSupportedOnOS) {
+		return nil, fmt.Errorf("kfuncs are not supported on this platform: %w", err)
+	}
+	if err != nil && !errors.Is(err, ErrNotSupported) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


ti-mo (approved):
Thanks!

linker.go
```diff
@@ -301,9 +301,10 @@ func fixupKfuncs(insns asm.Instructions) (_ handles, err error) {
 
 fixups:
 	// only load the kernel spec if we found at least one kfunc call
-	kernelSpec, err := btf.LoadKernelSpec()
-	if err != nil {
-		return nil, err
+	if kernelSpec == nil {
+		if kernelSpec, err = btf.LoadKernelSpec(); err != nil {
```
chenhengqi, line 305: The CI is unhappy with this change. This should take module BTF into account.

ti-mo:
> Second, is we could allow users to provide a set of module BTFs which we would consult instead of the module BTFs from the kernel itself (if any exist). It would be up to the user to ensure they are generated with the provided vmlinux BTF as base. This can be sometime we add later as addition to the first approach.

To add a data point to this discussion, BTFHub ships a single blob per kernel and seems to disregard kernel modules entirely. (I could be wrong, of course!) libbpf also accepts only a single blob in `.btf_custom_path`: https://www.mankier.com/8/bpftool-gen#Examples-min_core_btf:

```
LIBBPF_OPTS(bpf_object_open_opts, opts, .btf_custom_path = "5.4.0-smaller.btf");
struct bpf_object *obj;

obj = bpf_object__open_file("one.bpf.o", &opts);
```

Purely for doing CO-RE, all kmods could technically be merged into a single blob, but it doesn't seem like the btfhub tooling currently supports that. And since these kernels aren't getting any younger, I doubt there's much incentive to do so.

Can't find the source, but if `.btf_custom_path` is set, libbpf ignores all other sources of BTF (like kmods). I think that only makes sense.

lmb:
I don't think it's possible to use external kmod BTF for kfuncs in the first place. See the code which adds the module handle to `fdArray`: that handle needs to point at a kmod BTF, you can't just load external BTF into the kernel.

I forget how kfuncs targeting the kernel work exactly. Is it even possible without having BTF built in? AKA, not sure this change makes sense at all.

ti-mo:
> I don't think it's possible to use external kmod BTF for kfuncs in the first place. See the code which adds the module handle to `fdArray`: that handle needs to point at a kmod BTF, you can't just load external BTF into the kernel.
> 
> I forget how kfuncs targeting the kernel work exactly. Is it even possible without having BTF built in? AKA, not sure this change makes sense at all.

Good points! I assume it may be possible to invoke vmlinux kfuncs using external BTF if a distro/admin decided to ship BTF separately instead of in /sys/kernel/btf as a 'security' measure? Iirc kfuncs are referred to by btf id. We probably shouldn't take kmod BTF into account during resolving in any case.

lmb:
> Iirc kfuncs are referred to by btf id.

Exactly, so how does the kernel figure out what to attach to if it doesn't have BTF included? It also needs to do the id -> kfunc lookup, no? And if it needs BTF included, why is it not available in /sys/kernel/btf?

@bogushevich can you tell us a bit more when / how this is useful / necessary?

dylandreimerink:
> Exactly, so how does the kernel figure out what to attach to if it doesn't have BTF included? It also needs to do the id -> kfunc lookup, no? And if it needs BTF included, why is it not available in /sys/kernel/btf?

ACK. From what I understand from the kconfigs and header files is that when compiling with BTF, the kfunc BTF ids are encoded in ".BTF_ids" ELF sections of the kernel, which are then read at runtime to know which ids are kfuncs with what flags. If you don't compile the kernel with BTF, these are not included, thus no kfuncs. Even if you were to create BTF from the DWARF like those on BTFHub.

lmb:
@bogushevich how about we document this on fixupKfuncs so that future generations don't go down the same rabbit hole?

linker.go
```diff
@@ -301,9 +301,10 @@ func fixupKfuncs(insns asm.Instructions) (_ handles, err error) {
...
+			return nil, err
```
ti-mo, line 306: @lmb I think adding some context to the error string would be more helpful to both devs and users:

```suggestion
			return nil, fmt.Errorf("kfuncs require the kernel to be built with BTF: %w", err)
```

bogushevich:
Sorry, it was my mistake. You are right, after this commit my program is loaded successfully, but doesn't work properly.
My code is like this:
```c
volatile const __u8 conntrack_enabled = 0;

extern struct nf_conn *bpf_skb_ct_lookup(
    struct __sk_buff *,
    struct bpf_sock_tuple *, u32,
    struct bpf_ct_opts___local *, u32
) __ksym __weak;

if (conntrack_enabled) {
   struct nf_conn *ct = bpf_skb_ct_lookup(/*...*/)
   if (ct) {/*...*/}
}
```
So if flag **conntrack_enabled** is not set from user space, program is successfully loaded even on kernels that don’t support necessary kfuncs. But if there is no system BTF, program fallback to custom BTF and loading is failed on fixupKfuncs step regardless of flag value.
Unfortunately some modern Linux distributions still don’t enable BTF for some reason. I just dont’t want to have two similar BPF program: one with conntrack support and one without.

ti-mo:
> So if flag **conntrack_enabled** is not set from user space, program is successfully loaded even on kernels that don’t support necessary kfuncs. But if there is no system BTF, program fallback to custom BTF and loading is failed on fixupKfuncs step regardless of flag value. Unfortunately some modern Linux distributions still don’t enable BTF for some reason. I just dont’t want to have two similar BPF program: one with conntrack support and one without.

I think this makes a lot of sense to support. @dylandreimerink @lmb Should we treat an error from `btf.LoadKernelSpec` in fixupKfuncs as recoverable and poison all weak kfunc calls? IMO that's minimally invasive.

PR #1778, opened 2025-05-07
collection: share kernel BTF when loading a collection

lmb:
We currently create a new copy of kernel BTF when loading each program in a 
    collection, which is quite expensive.

    Introduce a new type btf.Cache which allows re-using kernel BTF as-needed. 
    As an optimisation, the Cache is opportunistically populated from the global 
    cache on creation.

        core: 1
       goos: linux
       goarch: amd64
       pkg: github.com/cilium/ebpf
                           │   base.txt   │             cache.txt              │
                           │    sec/op    │   sec/op     vs base               │
       NewCollectionManyProgs   347.84m ± 1%   14.54m ± 2%  -95.82% (p=0.002
    n=6)

                            │    base.txt    │              cache.txt           
      │
                           │      B/op      │     B/op      vs base             
     │
       NewCollectionManyProgs   108.056Mi ± 0%   3.841Mi ± 0%  -96.44% (p=0.002
    n=6)

                            │   base.txt    │             cache.txt             
    │
                           │   allocs/op   │  allocs/op   vs base              
    │
       NewCollectionManyProgs   1026.83k ± 0%   38.04k ± 0%  -96.30% (p=0.002
    n=6)

### DIFF
--- a/btf/kernel.go
+++ b/btf/kernel.go
@@ -12,7 +12,8 @@ import (
 	"github.com/cilium/ebpf/internal/platform"
 )
 
-var kernelBTF = struct {
+// globalCache amortises decoding BTF across all users of the library.
+var globalCache = struct {
 	sync.RWMutex
 	kernel  *Spec
 	modules map[string]*Spec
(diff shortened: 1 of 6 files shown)


lmb:
@patrickpichler could you give this branch a try and let me know what you think?

patrickpichler:
Sorry for the super late response. I did run the benchmark again and here are the results:
Without your change
BenchmarkNewCollectionManyProgs-4   	     15	 75442629 ns/op	74477811 B/op	 809763 allocs/op

With your change
BenchmarkNewCollectionManyProgs-4   	    272	  4193884 ns/op	2734829 B/op	  30806 allocs/op

With the optimization I did in the draft PR (#1690)
BenchmarkNewCollectionManyProgs-4 184 5807742 ns/op 4134444 B/op 17325 allocs/op

When performing a single run, the total number of allocated memory according to pprof went down from `~660MB` to `~26MB`, so this is definitely a win. 

I guess there are a few more ways on how to reduce the number of allocations, but for my use case, this should be good enough. 

Thanks a lot!

dylandreimerink (approved):
This seems sane to me :rocket:

PR #1779, opened 2025-05-07
ci: use real staticcheck

lmb:
golangci-lint uses a copy of staticcheck rules, which leads to funky differences between what staticcheck reports locally and what golangci-lint reports. This is annoying, and the real staticcheck is of higher quality than golangci-lint. Use it in CI instead of the copy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,12 @@ jobs:
         with:
           go-version: '${{ env.go_version }}'
 
+      - name: Run staticcheck
+        uses: dominikh/staticcheck-action@v1
+        with:
+          version: "latest"
+          install-go: false
+
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v7.0.0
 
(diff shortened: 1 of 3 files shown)


ti-mo (approved):

lmb:
@ti-mo staticcheck is kind of slow, 20s for every run. How about I drop the darwin cross build? I don't think it adds much since we now do the wasm build.

ti-mo:
> @ti-mo staticcheck is kind of slow, 20s for every run. How about I drop the darwin cross build? I don't think it adds much since we now do the wasm build.

I'm not sure if that's necessarily true. Take an old run, for example: https://github.com/cilium/ebpf/actions/runs/14728073480/job/41335459778?pr=1764. golangci-lint takes 12s there, while it takes 4s in this PR. Clearly, running staticcheck warms up a cache somewhere.

Since `Build and Lint` is in the critical path, I'd suggest we move the cross-builds out into a separate job that kicks off in parallel together with the `Run tests` jobs. That way, the test matrix won't take as long to start, and we'll save more time overall.

PR #1785, opened 2025-05-20
kallsyms: optimize things a bit

lmb:
Two small optimisations which drop CPU time by 25% on my machine, plus a lot less allocations.

### DIFF
--- a/internal/kallsyms/kallsyms.go
+++ b/internal/kallsyms/kallsyms.go
@@ -1,13 +1,13 @@
 package kallsyms
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"slices"
 	"strconv"
-	"strings"
 
 	"github.com/cilium/ebpf/internal"
 	"github.com/cilium/ebpf/internal/platform"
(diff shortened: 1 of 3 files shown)


internal/kallsyms/kallsyms.go
```diff
@@ -221,70 +219,76 @@ func assignAddresses(f io.Reader, symbols map[string]uint64) error {
...
-func parseSymbol(r *reader, types []rune) (s ksym, err error, skip bool) {
-	for i := 0; r.Word(); i++ {
-		switch i {
-		// Address of the symbol.
-		case 0:
-			s.addr, err = strconv.ParseUint(r.Text(), 16, 64)
-			if err != nil {
-				return s, fmt.Errorf("parsing address: %w", err), false
+// Example line: `ffffffffc1682010 T nf_nat_init\t[nf_nat]`
+func parseSymbol(r *reader, types []rune) iter.Seq2[ksym, error] {
```
ti-mo, line 253: Since it's now an iterator:
```suggestion
func symbols(r *reader, types []rune) iter.Seq2[ksym, error] {
```

internal/kallsyms/kallsyms.go
```diff
@@ -221,70 +219,76 @@ func assignAddresses(f io.Reader, symbols map[string]uint64) error {
...
+			var s ksym
+			var err error
+			for i := 0; r.Word(); i++ {
+				switch i {
+				// Address of the symbol.
+				case 0:
+					s.addr, err = strconv.ParseUint(r.Text(), 16, 64)
+					if err != nil {
+						yield(ksym{}, fmt.Errorf("parsing address: %w", err))
+						return
```
ti-mo, line 266: Stray thought: doesn't `yield` return/consume?

ti-mo (approved):
Thanks! Making `parseSymbol` return an iterator makes its implementation a little awkward since it itself contains a loop over an iterator type. I'd probably drop that change, I think the optimization stands well on its own.

internal/kallsyms/kallsyms.go
lmb: `yield` returns the value to the caller, `return` aborts the iteration so that the for loop exits. I guess I could skip the return and instead process the next line. Not sure that makes sense though? Don't really have a use case for it.

PR #1798, opened 2025-06-07
bpf2go: import package structs when all structs are typedefs

RonFed:
The current code checks for the existence of any `btf.Struct` when deciding to import the "structs" package.
Change this to check the underlying type is struct. This handles cases where all the structs are using `typedef`.


lmb:
Thanks!

PR #1800, opened 2025-06-10
ci: upgrade LLVM to 20.1, drop LLVM 11

ti-mo:
LLVM 11 is getting a bit long in the tooth and doesn't support decl tags.

Target 14, 17 and the newly-released LLVM 20.


dylandreimerink:
> unknown insn class 6

That can't be good :face_with_peeking_eye:

ti-mo:
> > unknown insn class 6
> 
> That can't be good 🫣

Looking into it, I think LLVM20 may be using ISAv3 by default.

ti-mo:
Found it: https://github.com/llvm/llvm-project/commit/7852ebc088b925ef1c1940cbd56a93d9f8e3e330. -mcpu=v3 is now the default.

@yonghong-song This is slightly surprising given this breaks existing tools on kernels <5.1 if they don't specify -mcpu=v1/v2 explicitly. The kernel docs (https://docs.kernel.org/bpf/bpf_devel_QA.html#q-new-bpf-instruction-for-kernel-and-llvm) need to be updated as well:
> By default the `generic` processor target is used, which is the base instruction set (v1) of BPF.

I'll set -mcpu=v2 here until we drop <5.1 in CI.

yonghong-song:
> Found it: [llvm/llvm-project@7852ebc](https://github.com/llvm/llvm-project/commit/7852ebc088b925ef1c1940cbd56a93d9f8e3e330). -mcpu=v3 is now the default.
> 
> @yonghong-song This is slightly surprising given this breaks existing tools on kernels <5.1 if they don't specify -mcpu=v1/v2 explicitly. The kernel docs (https://docs.kernel.org/bpf/bpf_devel_QA.html#q-new-bpf-instruction-for-kernel-and-llvm) need to be updated as well:
> 
> > By default the `generic` processor target is used, which is the base instruction set (v1) of BPF.
> 
> I'll set -mcpu=v2 here until we drop <5.1 in CI.

Yes, in llvm20 we move -mcpu=v3 as default to handle some atomic insns (esp. for arm64). Please to use -mcpu=v1/v2 as appropriate. Thanks.

yonghong-song:
> Found it: [llvm/llvm-project@7852ebc](https://github.com/llvm/llvm-project/commit/7852ebc088b925ef1c1940cbd56a93d9f8e3e330). -mcpu=v3 is now the default.
> 
> @yonghong-song This is slightly surprising given this breaks existing tools on kernels <5.1 if they don't specify -mcpu=v1/v2 explicitly. The kernel docs (https://docs.kernel.org/bpf/bpf_devel_QA.html#q-new-bpf-instruction-for-kernel-and-llvm) need to be updated as well:

Will update the doc soon. thanks.

PR #1794, opened 2025-06-04
program: perform CO-RE against all kmod

lmb:
This fixes a long standing issue with our CO-RE implementation. Please see the commit message for details.

PR #1802, opened 2025-06-17
kallsyms: add KallsymsPreLoad to warm the kallsyms cache

alban:
When a process uses several CollectionSpecs, `/proc/kallsyms` is read separately for each collection, which is inefficient. This PR adds a way to warm the kallsyms cache for all CollectionSpecs.

```
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf
cpu: Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
                       │ base.bench  │              opt.bench              │
AssignCollectionMany-4   806.0m ± 5%   267.9m ± 3%  -66.77% (p=0.000 n=10)

                       │  base.bench  │              opt.bench               │
AssignCollectionMany-4   25.15Ki ± 1%   16.26Ki ± 1%  -35.36% (p=0.000 n=10)

                       │ base.bench │             opt.bench             │
                       │ allocs/op  │ allocs/op   vs base               │
AssignCollectionMany-4   169.0 ± 1%   156.0 ± 1%  -7.69% (p=0.000 n=10)
```

I would like to use this for https://github.com/inspektor-gadget/inspektor-gadget/pull/4625


lmb:
@alban https://github.com/cilium/ebpf/pull/1794 removes the dependency on kallsyms for kprobes. Can you try that PR and let me know whether it helps as well?

alban:
> @alban #1794 removes the dependency on kallsyms for kprobes. Can you try that PR and let me know whether it helps as well?

I tested with and without #1794 (https://github.com/alban/inspektor-gadget/commit/a349756fbc7ab3ca1978c47cde8204bcc36ac6fd). It gives some speed-up but not as much as this PR:

```
# Without #1794
$ hyperfine -w 2 --runs 10 "expect -c 'spawn sudo ig run trace_dns ; expect \"Gadget started successfully\" { puts \"\n--- OK ---\" } timeout { puts \"\n--- Timeout ---\"; exit 1 } eof { puts \"\n--- EOF ---\"; exit 1 }; exit 0'"
Benchmark 1: expect -c 'spawn sudo ig run trace_dns ; expect "Gadget started successfully" { puts "\n--- OK ---" } timeout { puts "\n--- Timeout ---"; exit 1 } eof { puts "\n--- EOF ---"; exit 1 }; exit 0'
  Time (mean ± σ):      3.311 s ±  0.177 s    [User: 0.003 s, System: 0.002 s]
  Range (min … max):    3.157 s …  3.654 s    10 runs

# With #1794 
$ hyperfine -w 2 --runs 10 "expect -c 'spawn sudo ig run trace_dns ; expect \"Gadget started successfully\" { puts \"\n--- OK ---\" } timeout { puts \"\n--- Timeout ---\"; exit 1 } eof { puts \"\n--- EOF ---\"; exit 1 }; exit 0'"
Benchmark 1: expect -c 'spawn sudo ig run trace_dns ; expect "Gadget started successfully" { puts "\n--- OK ---" } timeout { puts "\n--- Timeout ---"; exit 1 } eof { puts "\n--- EOF ---"; exit 1 }; exit 0'
  Time (mean ± σ):      2.991 s ±  0.044 s    [User: 0.002 s, System: 0.003 s]
  Range (min … max):    2.934 s …  3.089 s    10 runs
```
Speedup with #1794: 0.32s (10%)
Speedup with this PR: 1.123s (27%)

ti-mo:
I think this is somewhat far-fetched. The main factor behind the kallsyms rework was to trade off memory for CPU usage, since the latter is often (more) abundant. It's already locally (within a single CollectionSpec) optimal, but I assume that doesn't work for you.

In inspektor-gadget, how will you guarantee you have the correct list of ksyms up front? If there's one missing, you incur another unintended kallsyms scan, even with a warm cache. Do you have tests to make sure this won't happen? Or do you collect these by iterating Programs[].AttachTo?

Have you considered building all your programs into a single ELF? :smile: More seriously, I've considered exporting the kallsyms package somewhere, iirc @brycekahle or someone else from ddog asked about it a while ago. Then we don't need to add this funky kind of API.

alban:
> I think this is somewhat far-fetched. The main factor behind the kallsyms rework was to trade off memory for CPU usage, since the latter is often (more) abundant. It's already locally (within a single CollectionSpec) optimal, but I assume that doesn't work for you.

Correct, the previous improvements for symbols located in a single CollectionSpec already improved performance, but we use several CollectionSpecs (a different set can be used depending on the configuration, so it is not practical to put all the programs in a single ELF), so the performance is not as good as it could be.

> In inspektor-gadget, how will you guarantee you have the correct list of ksyms up front? If there's one missing, you incur another unintended kallsyms scan, even with a warm cache. Do you have tests to make sure this won't happen? Or do you collect these by iterating Programs[].AttachTo?

Previously, my initial PoC from https://github.com/inspektor-gadget/inspektor-gadget/pull/4625 was just hardcoding the list, which is indeed not clean and has maintenance problems.

I tried your suggestion of iterating Programs[].AttachTo and it seems better.

> Have you considered building all your programs into a single ELF? 😄 More seriously, I've considered exporting the kallsyms package somewhere, iirc @brycekahle or someone else from ddog asked about it a while ago. Then we don't need to add this funky kind of API.

I agree my suggested API "KallsymsPreLoad" is funky or clunky. I didn't have better ideas how to make the API cleaner. If the kallsyms package was exported, that would be ideal.

Meanwhile, I think this PR can be closed. I am trying an alternative approach in https://github.com/inspektor-gadget/inspektor-gadget/pull/4653 to iterate on Programs[].AttachTo to collect the symbols and then load a dummy CollectionSpec without any BPF bytecode. The loading will fail (with the error "instructions cannot be empty") but cilium/ebpf will still warm the kallsyms cache with the correct symbols, so I get the performance I wanted.

lmb:
Okay, let's consider making the package public.

PR #1803, opened 2025-06-18
prog: get context out from syscall program

mejedi:
Linux syscall program errors on non-nil ctxOut, uses ctxIn for both input and output [1]. We have separate Context and ContextOut fields in RunOptions. It should be possible to capture output from syscall programs which is currently not: non-nil ContextOut triggers EINVAL, while Context is (expectedly) not updated.

Implement the following semantics: either Context, or ContextOut or both can be non-nil. Map to ctxIn internally, complain if Context and ContextOut lengths are different.

[1] https://elixir.bootlin.com/linux/v6.15.2/source/net/bpf/test_run.c#L1530

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -884,6 +884,16 @@ func (p *Program) run(opts *RunOptions) (uint32, time.Duration, error) {
 		Cpu:         opts.CPU,
 	}
 
+	if p.Type() == Syscall && ctxIn != nil && ctxOut != nil {
+		// Linux syscall program errors on non-nil ctxOut, uses ctxIn
+		// for both input and output. Shield the user from this wart.
+		if len(ctxIn) != len(ctxOut) {
+			return 0, 0, errors.New("length mismatch: Context and ContextOut")
+		}
+		attr.CtxOut, attr.CtxSizeOut = sys.TypedPointer[uint8]{}, 0
+		ctxOut = ctxIn
+	}
+
 retry:
 	for {
 		err := sys.ProgRun(&attr)
--- a/prog_linux_test.go
+++ b/prog_linux_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
 	"github.com/cilium/ebpf/internal/testutils"
 	"github.com/cilium/ebpf/internal/unix"
 )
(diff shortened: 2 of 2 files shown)


mejedi:
I would like to add tests in `prog_test.go`. I see that most tests use a synthetic program spawned via `createProgram`. I need something more complex to verify that ctxOut properly roundtrips. Is it OK to load a program from an object file in these tests?

prog.go
```diff
@@ -869,6 +869,18 @@ func (p *Program) run(opts *RunOptions) (uint32, time.Duration, error) {
...
+	sysCtxOut := ctxOut
+	if ctxOut != nil && p.Type() == Syscall {
+		// Linux syscall program errors on non-nil ctxOut, uses ctxIn
+		// for both input and output. Shield the user from this wart.
+		if ctxIn == nil {
+			ctxIn = ctxOut
+		} else if len(ctxIn) != len(ctxOut) {
+			return 0, 0, errors.New("Context and ContextOut length mismatch")
+		}
+		sysCtxOut = ctxIn
```
mejedi, line 881:
```suggestion
		sysCtxOut = nil
```

prog.go
```diff
@@ -869,6 +869,20 @@ func (p *Program) run(opts *RunOptions) (uint32, time.Duration, error) {
 		ctxOut = make([]byte, binary.Size(opts.ContextOut))
 	}
 
+	sysCtxOut := ctxOut
+	if ctxOut != nil && p.Type() == Syscall {
```
lmb: FWIW I'd prefer sticking as close to upstream semantics as possible. They might allow ctx out at some point after all.

Just to make sure I understand correctly: would it be enough to do roughly:

```
if p.Type() == Syscall {
   // unmarshal from ctxIn
}
```

aftert the syscall has returned?

lmb (commented):
Could you add a test which executes a syscall program or are they too complicated to set up?

prog.go
mejedi: Yes, adding unmarshal from ctxIn would be enough.

Re sticking to upstream semantics: libbpf apparently took this route [1]. `struct` [`bpf_test_run_opts`](https://elixir.bootlin.com/linux/v6.12.6/C/ident/bpf_test_run_opts) defines `ctx_in` as a pointer to `const`:
```c
const void *ctx_in;
void *ctx_out;
```
This is quite unfortunate since syscall prog actually writes there.

Likewise, unmarshaling  from `ctxIn` to `Context` has implications: 
 * only slices and pointers are accepted for `Context`, unlike other prog types;
 * `Context` is updated, even if we are only interested in passing info one way.

Overall, it feels rather quirky. Please let me know if you disagree.

[1] https://elixir.bootlin.com/linux/v6.12.6/source/tools/lib/bpf/bpf.c#L978

prog_linux_test.go
```diff
@@ -135,3 +136,45 @@ func TestProgramVerifierLogLinux(t *testing.T) {
 	})
 	qt.Assert(t, qt.IsTrue(len(prog.VerifierLog) > minVerifierLogSize))
 }
+
+func TestProgramTestRunSyscall(t *testing.T) {
```
mejedi, line 140: @lmb added a test

prog.go
```diff
@@ -869,6 +869,20 @@ func (p *Program) run(opts *RunOptions) (uint32, time.Duration, error) {
...
+		// Linux syscall program errors on non-nil ctxOut, uses ctxIn
+		// for both input and output. Shield the user from this wart.
+		sysCtxOut = nil
+		if ctxIn == nil {
```
lmb: Why does this set ctxOut?

mejedi: `ctxOut` is unmarshaled from later.

We need it to be the same as the syscall input (`ctxIn == ctxOut`). We can't `ctxIn = ctxOut` since it has been possibly marshaled to, therefore `ctxOut = ctxIn`.

lmb: I think that this should just fail if no input is provided, since that is what upstream does as well. We don't support `in == nil && out != nil` for other types, no?

mejedi: > We don't support in == nil && out != nil for other types, no?

It looks like it is supposed to work — see for instance [bpf_ctx_init](https://elixir.bootlin.com/linux/v6.15.3/source/net/bpf/test_run.c#L808) — but it doesn't, the below program fails with `EINVAL` (arguments to `bpf` syscall look reasonable).

I still feel that we shouldn't repurpose `Context` for output. I ~~will~~ DID simplify by dropping support for `in == nil && out != nil`. 

```go
package main

import (
	"log"

	"github.com/cilium/ebpf"
	"github.com/cilium/ebpf/asm"
)

func main() {
	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
		Type: ebpf.XDP,
		Instructions: []asm.Instruction{
			asm.Mov.Imm(asm.R2, -8),
			asm.FnXdpAdjustMeta.Call(),
			asm.Return(),
		},
	})
	if err != nil {
		log.Fatal(err)
	}

	data := make([]byte, 128)
	dataOut := make([]byte, 1024)
	ctx := struct{ Data, DataEnd, DataMeta, I1, I2, I3 uint32 }{DataEnd: 128}

	rc, err := prog.Run(&ebpf.RunOptions{Data: data, DataOut: dataOut, ContextOut: &ctx})
	log.Print(rc, err, ctx)
}
```

lmb (approved):
Thanks!

Issue #1797, opened 2025-06-05
`call -1` instructions no longer load without error


brycekahle:
### Describe the bug

https://github.com/cilium/ebpf/commit/f086705551436bf5bd27d6dc0a22df8974b87ef2 breaks loading of programs with error:
`decoding instructions for section <sectionname>: offset <offset>: invalid constant 0xffffffff`

The actual instruction looks like `85 00 00 00 ff ff ff ff	call -1`

This is happening because we have `call -1` instructions that we use as sentinels for patching before loading into the kernel. Since `0xffff` is a valid 32 bit value, it shouldn't be considered an invalid constant or instruction. I'm not sure why it is being sign-extended to 64 bits in the error message.

### How to reproduce

Write an ebpf program with the following:
```
#define PATCH_TARGET -1
static void *(*bpf_patch)(unsigned long, ...) = (void *)PATCH_TARGET;
```
and in your program use it
```
bpf_patch(0, 0);
```
then try to load the program using `cilium/ebpf`

### Version information

n/a

lmb:
@dylandreimerink could you take a look? I'll be out for the next week and a half.

dylandreimerink:
Yes, will take a look

dylandreimerink:
Hmm. I am able to repo this and found the cause. In https://github.com/cilium/ebpf/commit/f086705551436bf5bd27d6dc0a22df8974b87ef2 I moved this line https://github.com/cilium/ebpf/blob/df9ebe841740387ce8bcd7f72220fc1ef1b21ed8/asm/instruction.go#L67 from below the if statement to above it.

Before that change `ins.Constant` is always `0` when passed to `BuiltinFuncForPlatform`. So I think that logic didn't work as expected before it was accidentally "fixed" by this move.

`BuiltinFuncForPlatform` applies a tag in the upper 4 bits which is required for Windows support. The code we have asserts that no more than the lower 28 bits of the constant are used, but -1 of course uses all 32 bits due to the way its encoded.

https://github.com/cilium/ebpf/commit/6bba483eceea01da4f61c05858ae3345b7b414ab made it so `85 00 00 00 ff ff ff ff	call -1` cannot be decoded anymore, however the new logic was never properly invoked.

Option one is to not throw an error anymore and overwrite the upper 4 bits. Side effect is that `85 00 00 00 ff ff ff ff` would be decoded as `call 268435455` (the sign bit zeroed out) which would likely also mess up your placeholder code.

[...]

PR #1806, opened 2025-06-19
asm: allow negative constants for builtin function calls

lmb:
The work to encode a platform into various constant types made it so that decoding a call to a builtin helper with a negative value fails with

    decoding instructions for section <sectionname>:
    offset <offset>: invalid constant 0xffffffff

for a BPF instruction of "call -1". This is because we can't represent -1 as a tagged platform constant.

Allow negative constants by not transforming them into a platform constant at all. Adjust the platform tag size so that we never generate a platform constant with the high bit set. This avoids confusing it with a negative number when reinterpreting it as a signed number and ensures that trying to marshal such an instruction gives an error.

Fixes: https://github.com/cilium/ebpf/issues/1797

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -67,11 +67,17 @@ func (ins *Instruction) Unmarshal(r io.Reader, bo binary.ByteOrder, platform str
 	ins.Constant = int64(int32(bo.Uint32(data[4:8])))
 
 	if ins.IsBuiltinCall() {
-		fn, err := BuiltinFuncForPlatform(platform, uint32(ins.Constant))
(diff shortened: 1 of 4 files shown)


lmb:
@dylandreimerink @brycekahle PTAL

dylandreimerink (approved):
That is an easier fix than I thought. The builtin function call numbers can never get that high anyway.

Seems like we do have to adjust `TestConstant` so it iterates only over 0 to int32_max instead of 0 to uint32_max

ti-mo (approved):

PR #1807, opened 2025-06-19
elf_reader: do not panic in `loadMaps`if ELF file has no BTF

paulcacheux:
3f1a755f1451d058e8b56b516c7a64ff6f3a33e3 introduced the ability to read btf_decl_tag attributes set on maps but `loadMaps` is crashing if the ELF file has no BTF.

Fix this by checking if the BTF is there before querying it.


paulcacheux:
Definitely not something we ship, just something that we run in our CI and that it seems is using some outdated build commands (that I will update)

ti-mo:
Thanks! I've created a follow-up to this: https://github.com/cilium/ebpf/issues/1810.

PR #1812, opened 2025-06-24
btf: mmap vmlinux if possible

lmb:
See commit messages. Merge once 6.15 is released.


btf/kernel.go
florianl, line 1: I do have a similar approach in progress for kallsyms. :sweat_smile: Happy to see the use of mmap also here! :+1:

lmb: Nice! I wasn't aware that /proc/kallsyms can be mmapped?

PR #1815, opened 2025-06-26
collection: Support omitempty tag

jrife:
WIP idea. Triggering CI.


ti-mo:
Let's go with passing `[]any` to LoadAndAssign instead, that way we gain some composability without bringing out the hammer that's `omitempty`. :wink:

jrife:
@ti-mo I totally forgot about this PR, but yes I agree let's abandon this.

PR #1819, opened 2025-07-07
feat : Add `LoadSplitSpec` function

tdaudi:
Similarly to LoadSpec, LoadSplitSpec could make an easier use of split-BTF files

### DIFF
--- a/btf/btf_test.go
+++ b/btf/btf_test.go
@@ -433,16 +433,10 @@ func TestTypesIterator(t *testing.T) {
 	qt.Assert(t, qt.DeepEquals(have, types))
 }
 
-func TestLoadSplitSpecFromReader(t *testing.T) {
+func TestLoadSplitSpec(t *testing.T) {
 	spec := vmlinuxTestdataSpec(t)
 
-	f, err := os.Open("testdata/btf_testmod.btf")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer f.Close()
-
-	splitSpec, err := LoadSplitSpecFromReader(f, spec)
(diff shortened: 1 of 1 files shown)


btf/btf.go
```diff
@@ -500,6 +500,17 @@ func (s *Spec) TypeByName(name string, typ interface{}) error {
 	return nil
 }
 
+// LoadSplitSpec opens file and calls LoadSplitSpecFromReader on it.
```
ti-mo: Nit: the callout to `LoadSplitSpecFromReader` isn't so relevant.

Let's reuse the docstring:

```
// LoadSplitSpec loads split BTF from the given file.
//
// Types from base are used to resolve references in the split BTF.
// The returned Spec only contains types from the split BTF, not from the base.
```

ti-mo (changes requested):
Thank you! Could you also change `TestLoadSplitSpecFromReader` to use `LoadSplitSpec` instead, and rename it to `TestLoadSplitSpec`?

btf/btf.go
tdaudi: Sorry for this.
Thank you for pointing this out, I changed it

tdaudi:
> Thank you! Could you also change `TestLoadSplitSpecFromReader` to use `LoadSplitSpec` instead, and rename it to `TestLoadSplitSpec`?

Sure ! I added another commit for this.
I can squash the commits if you prefer

tdaudi:
Sorry for the noise, I pushed the commits with the wrong email

ti-mo (approved):
Thanks!

Issue #1818, opened 2025-07-06
OpenWRT 24.10.2 failed to load BTF for kmod nls_cp437


Testeera:
### Describe the bug

After updating from 0.15.0 to 0.19.0(https://github.com/daeuniverse/dae/pull/857), dae cannot run
[Jul 06 08:43:51]  INFO Include config files: [/etc/dae/a.dae]
[Jul 06 08:43:51]  INFO Waiting for network...
[Jul 06 08:43:51]  INFO Network online.
[Jul 06 08:43:51]  WARN No node found.
[Jul 06 08:43:51]  WARN No interface to bind.
[Jul 06 08:43:51]  INFO Loading eBPF programs and maps into the kernel...
[Jul 06 08:43:51]  INFO The loading process takes about 120MB free memory, which will be released after loading. Insufficient memory will cause loading failure.
Variable PARAM: PARAM (type=Var:"PARAM"[global], map=.rodata, offset=0, size=24)
[Jul 06 08:43:51]  INFO Incompatible new map format with existing map routing_tuples_map detected; removed the old one.
Variable PARAM: PARAM (type=Var:"PARAM"[global], map=.rodata, offset=0, size=24)
[Jul 06 08:43:51] FATAL load eBPF objects: field LocalTcpSockops: program local_tcp_sockops: apply CO-RE relocations: load BTF for kmod nls_cp437: read string section: string table is empty

### How to reproduce

Compile daeuniverse/dae with pull request 857
dae run -c /etc/dae/a.dae
`global{}
routing{}`

### Version information

github.com/cilium/ebpf v0.19.0

Testeera:
Related to https://github.com/cilium/ebpf/commit/9958a4f58fb4e533201e0a76d6d82eeeaa91434b

Testeera:
How to skip empty @lmb

Testeera:
https://github.com/Testeera/ebpf/commit/62bc71746ae94e47a18f0f6b861d818a7f0df78c Is this OK? It definitely works.

ti-mo:
@Testeera Thanks for the report, that's a little suspicious. Could you upload a copy of `/sys/kernel/btf/vmlinux` and `/sys/kernel/btf/nls_cp437` here, please? Also, what's the value of the `CONFIG_NLS_CODEPAGE_437` kconfig of your kernel?

Testeera:
[...]

PR #1820, opened 2025-07-07
btf: tolerate empty btf files

Testeera:
See issue: https://github.com/cilium/ebpf/issues/1818
I encountered this problem in https://github.com/daeuniverse/dae/pull/857
Tested by myself.


lmb:
@Testeera could you add a test which decodes an empty BTF? You can use `Builder` while setting PreventNoTypeFound to false.

Testeera:
@lmb could you help because i'm not good at it, besides i'm working away from home now.

PR #1821, opened 2025-07-10
collection: drop unnecessary kallsyms module lookup

lmb:
collection: drop unnecessary kallsyms module lookup

    The library doesn't rely on knowing symbol module names anymore. Drop some
    dead code which was missed during that refactoring.

internal/kallsyms: drop Module, AssignModules

    The library doesn't use these routines anymore. Splitting module lookup from
    address lookup also doesn't really make sense because the really expensive
    part is parsing /proc/kallsyms, not storing a bit of extra metadata.


lmb:
@ti-mo not sure how attached you are to the modules code, I think it's easy to bring back if we do end up needing it.

PR #1822, opened 2025-07-16
prog: speed up opening pinned Programs

aibor:
With commit df9ebe8 the opening of programs by fd was improved. The LoadPinnedProgram still uses to full ProgramInfo instead of the sufficient minimal one.

This commit improves the load of pinned Programs as well by using the central helper for creating programs from a fd. This also makes it more consistent with the LoadPinnedMap function as well.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -978,20 +978,16 @@ func LoadPinnedProgram(fileName string, opts *LoadPinOptions) (*Program, error) 
 		return nil, fmt.Errorf("%s is not a Program", fileName)
 	}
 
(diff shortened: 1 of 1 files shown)


ti-mo (approved):
Thanks Tobias! Not sure how I missed this one. Have you done a broader check for similar cases? Clearly I didn't find them all, and I'm currently on leave. 😅

aibor:
Yes, I have checked for other uses of the functions providing the full info. They are only used for the actual `Info()` methods and tests now.

florianl (approved):

PR #1823, opened 2025-07-16
map: avoid allocations during PerCPUMap batch lookups

aibor:
Extend the optimization for batch lookups done in b89a4cb9 to PerCPUMaps as well.

It adapts the flow of the single value maps. For passed values objects that can not be used as backing memory, the behavior is the same as before.

```
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf
cpu: AMD Ryzen 5 5600G with Radeon Graphics
Iterate/PerCPUHash/BatchLookup-8            551.3µ ± 1%   129.7µ ± 1%  -76.47% (p=0.000 n=10)
Iterate/PerCPUHash/BatchLookupAndDelete-8   568.2µ ± 1%   148.4µ ± 1%  -73.88% (p=0.000 n=10)
geomean                                     559.7µ        138.7µ       -75.21%

                                          │   old.txt    │              new.txt               │
                                          │     B/op     │    B/op     vs base                │
Iterate/PerCPUHash/BatchLookup-8            89709.0 ± 0%   136.0 ± 0%  -99.85% (p=0.000 n=10)
Iterate/PerCPUHash/BatchLookupAndDelete-8   89799.0 ± 0%   224.0 ± 1%  -99.75% (p=0.000 n=10)
geomean                                     87.65Ki        174.5       -99.81%

                                          │    old.txt    │              new.txt               │
                                          │   allocs/op   │ allocs/op   vs base                │
Iterate/PerCPUHash/BatchLookup-8            1006.000 ± 0%   5.000 ± 0%  -99.50% (p=0.000 n=10)
Iterate/PerCPUHash/BatchLookupAndDelete-8   1006.000 ± 0%   5.000 ± 0%  -99.50% (p=0.000 n=10)
geomean                                       1.006k        5.000       -99.50%
```

### DIFF
--- a/map.go
+++ b/map.go
@@ -1220,17 +1220,18 @@ func (m *Map) batchLookupPerCPU(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, va
 		return 0, fmt.Errorf("keys: %w", err)
 	}
 
-	valueBuf := make([]byte, count*int(m.fullValueSize))
-	valuePtr := sys.UnsafeSlicePointer(valueBuf)
+	valueBuf := sysenc.SyscallOutput(valuesOut, count*int(m.fullValueSize))
 
-	n, sysErr := m.batchLookupCmd(cmd, cursor, count, keysOut, valuePtr, opts)
+	n, sysErr := m.batchLookupCmd(cmd, cursor, count, keysOut, valueBuf.Pointer(), opts)
 	if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
 		return 0, sysErr
 	}
 
-	err = unmarshalBatchPerCPUValue(valuesOut, count, int(m.valueSize), valueBuf)
-	if err != nil {
-		return 0, err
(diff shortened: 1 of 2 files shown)


lmb:
IIRC this will only work for types which do not have any implicit padding like uint64.

```
struct {
  a uint64
  b uint32
}
```

will still go via the slow path because there is a 4 byte trailer. This is because we don't want to allow reading into memory that is not "visible" in the go type system. Users can work around this with explicit padding.

```
struct {
  a uint64
  b uint32
  _ uint32
}
```

Do you still want to add this in that case? If yes please update the commit message to explain this and probably drop a comment in the source code as well.

aibor:
Yes, I am aware of the behavior and would like to have the zero-allocations behavior for types satisfying the requirements as described by you.

I have updated the commit message and referenced the original commit that has a very detailed description of the behavior.

How would you like to have it documented in the code? As far as I can see the behavior is not really documented for the Map methods directly yet, or have I missed it?

lmb:
Maybe having it in the commit message is fine then!

PR #1826, opened 2025-07-18
ci: print kernel version on arm64

lmb:


lmb:
See https://github.com/cilium/ebpf/pull/1827

PR #1828, opened 2025-07-19
btf: accept empty string table

Testeera:


lmb:
@Testeera thanks for the PR, you need to add a Signed-off-by: trailer to your commit.

btf/strings_test.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/go-quicktest/qt"
 )
 
+type mockSizedReader struct {
```
lmb, line 11: This is not used?

PR #1829, opened 2025-07-22
bpf2go: Allow multiple commands inside `BPF2GO_CC`

jrife:
Allow multiple commands to be specified inside `BPF2GO_CC`. Example:

```
BPF2GO_CC="ccache clang" go generate ../pkg/a/b
```

I'm working on https://github.com/cilium/cilium/pull/38693 where one of the checks uses `ccache clang` instead of just `clang` as the compile command substitution. Currently, this doesn't work with `bpf2go`. This change makes it possible to use the same compile command as the rest of the build in cases like this.

### DIFF
--- a/cmd/bpf2go/gen/compile.go
+++ b/cmd/bpf2go/gen/compile.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 )
 
 type CompileArgs struct {
@@ -14,7 +15,7 @@ type CompileArgs struct {
 	CC string
 	// Command used to strip DWARF from the ELF.
 	Strip string
-	// Flags to pass to the compiler.
+	// Flags to pass to the compiler. This may contain positional arguments as well.
 	Flags []string
 	// Absolute working directory
 	Workdir string
@@ -27,9 +28,8 @@ type CompileArgs struct {
 	DisableStripping bool
 }
 
-// Compile C to a BPF ELF file.
-func Compile(args CompileArgs) error {
-	// Default cflags that can be overridden by args.cFlags
+func insertDefaultFlags(flags []string) []string {
+	// Default cflags that can be overridden by the user.
 	overrideFlags := []string{
 		// Code needs to be optimized, otherwise the verifier will often fail
 		// to understand it.
@@ -40,7 +40,26 @@ func Compile(args CompileArgs) error {
 		"-mcpu=v1",
 	}
 
-	cmd := exec.Command(args.CC, append(overrideFlags, args.Flags...)...)
+	insert := 0
+
+	// Find the first non-positional argument to support CC commands with
+	// multiple components. E.g.: BPF2GO_CC="ccache clang" ...
+	for ; insert < len(flags); insert++ {
+		if strings.HasPrefix(flags[insert], "-") {
+			break
+		}
+	}
+
+	result := append([]string(nil), flags[:insert]...)
+	result = append(result, overrideFlags...)
+	result = append(result, flags[insert:]...)
+
+	return result
+}
+
+// Compile C to a BPF ELF file.
+func Compile(args CompileArgs) error {
+	cmd := exec.Command(args.CC, insertDefaultFlags(args.Flags)...)
 	cmd.Stderr = os.Stderr
 
 	inputDir := filepath.Dir(args.Source)
--- a/cmd/bpf2go/main.go
+++ b/cmd/bpf2go/main.go
@@ -155,9 +155,12 @@ func newB2G(stdout io.Writer, args []string) (*bpf2go, error) {
 		return nil, errors.New("missing package, you should either set the go-package flag or the GOPACKAGE env")
 	}
 
-	if b2g.cc == "" {
+	// Allow CC like "ccache clang" to work.
+	ccParts := strings.Fields(b2g.cc)
+	if len(ccParts) == 0 {
 		return nil, errors.New("no compiler specified")
 	}
+	b2g.cc = ccParts[0]
 
 	args, cFlags := splitCFlagsFromArgs(fs.Args())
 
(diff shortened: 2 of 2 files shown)


cmd/bpf2go/gen/compile.go
```diff
@@ -40,7 +41,8 @@ func Compile(args CompileArgs) error {
 		"-mcpu=v1",
 	}
 
-	cmd := exec.Command(args.CC, append(overrideFlags, args.Flags...)...)
+	ccParts := strings.Fields(args.CC)
```
lmb: Doing this here means that `CC` is kind of like an implicit additional `Flags` argument, without any documentation. I think it makes more sense to do this in `main()` where we pull CC out of flags. There we can split it and prepend it to CompileArgs.Flags.

lmb (changes requested):
Out of curiosity, does the following work?

```
bpf2go -cc ccache ... -- clang
```

Hunch: this will fail because of overrideFlags being passed to ccache before clang. Not sure that ccache likes that. If that's the case we can be a bit smarter in `Compile` and insert overrideFlags at the position where the first flags starts with `-`.

jrife:
> Hunch: this will fail because of overrideFlags being passed to ccache before clang. Not sure that ccache likes that. If that's the case we can be a bit smarter in Compile and insert overrideFlags at the position where the first flags starts with -

Yeah, it won't work currently. And yes, you could reposition `overrideFlags` to make this possible:

```
bpf2go -cc ccache ... -- clang
```

However, from the perspective of using bpf2go in the same Makefile as a bunch of other build commands, it would be inconvenient to need to do some special handling around its invocation to, e.g., split apart my `CC` variable. I think it makes a lot of sense to allow this sort of thing to be passed through via `BPF2GO_CC`.

cmd/bpf2go/gen/compile.go
jrife: Sure, I can move the split earlier in the sequence.

jrife: Done.

jrife:
@lmb Could you take another look at this when you get a chance? This is a blocker for me on https://github.com/cilium/cilium/pull/38693

cmd/bpf2go/main.go
```diff
@@ -375,11 +375,14 @@ func (b2g *bpf2go) convert(tgt gen.Target, goarches gen.GoArches) (err error) {
 		)
 	}
 
+	// Allow CC like "ccache clang" to work.
+	ccParts := strings.Split(b2g.cc, " ")
```
lmb: Switched this to `Split` since that guarantees returning an element, which Fields does not.

lmb (approved):

lmb:
@jrife DCO is failing, please take a look at https://github.com/cilium/ebpf/pull/1829/checks?check_run_id=46950786710

cmd/bpf2go/main.go
jrife: `Split()` won't work if there's multiple spaces or other whitespace characters for some reason. For example `" clang"` would yield `["", "clang"]`. `Fields` is a little more forgiving.

https://github.com/cilium/ebpf/blob/2b837f5d2d1574f2a855325a8b255914fd6841ad/cmd/bpf2go/main.go#L158-L160

There's already some validation to make sure `b2g.cc` contains *something*, but in the event that it only contains whitespce for some reason, we can add a check here for an empty slice and error out.

jrife:
@lmb 

> DCO is failing, please take a look at

Fixed. Author didn't match the `Signed-off-by` tag.

Also, switched back to `Fields` for the reasons I listed above and added a check to error out if `ccParts` is empty. Example:

```
BPF2GO_CC="   " BPF2GO_CFLAGS="-DSKIP_DEBUG=1 -DENABLE_IPV4=1 -DENABLE_IPV6=1 -DENABLE_ROUTING=1 -DPOLICY_VERDICT_NOTIFY=1 -DALLOW_ICMP_FRAG_NEEDED=1 -DENABLE_IDENTITY_MARK=1 -DMONITOR_AGGREGATION=3 -DCT_REPORT_FLAGS=0x0002 -DENABLE_HOST_FIREWALL=1 -DENABLE_ICMP_RULE=1 -DENABLE_CUSTOM_CALLS=1 -DENABLE_SRV6=1 -DENABLE_L7_LB=1 -DENABLE_MASQUERADE_IPV4=1 -DENABLE_IP_MASQ_AGENT_IPV4=1 -DENABLE_MASQUERADE_IPV6=1 -DENABLE_IP_MASQ_AGENT_IPV6=1 -DENABLE_SRC_RANGE_CHECK=1 -DENABLE_NODEPORT=1 -DENABLE_NODEPORT_ACCELERATION=1 -DENABLE_SESSION_AFFINITY=1 -DENABLE_DSR_ICMP_ERRORS=1 -DENABLE_DSR=1 -DENABLE_DSR_HYBRID=1 -DENABLE_IPV4_FRAGMENTS=1 -DENABLE_BANDWIDTH_MANAGER=1 -DENABLE_EGRESS_GATEWAY=1 -DENABLE_VTEP=1 -DENABLE_TPROXY=1 -DENABLE_HOST_ROUTING=1 -DENCAP_IFINDEX=1 -DTUNNEL_MODE=1 -DENABLE_MULTICAST=1 -DLB_SELECTION_PER_SERVICE=1 -DLB_SELECTION_MAGLEV=2 -DLB_SELECTION_RANDOM=1 -DENABLE_IPSEC=1 -I/go/src/github.com/cilium/cilium/bpf -I/go/src/github.com/cilium/cilium/bpf/include -O2 -g --target=bpf -std=gnu99 -nostdinc -ftrap-function=__undefined_trap -Wall -Wextra -Werror -Wshadow -Wno-address-of-packed-member -Wno-unknown-warning-option -Wno-gnu-variable-sized-type-not-at-end -Wimplicit-int-conversion -Wenum-conversion -Wimplicit-fallthrough -mcpu=v3" go generate ../pkg/datapath/loader
Error: no compiler specified
```

cmd/bpf2go/main.go
jrife: As an alternative, we could just do all validation inside `newB2G` and call `Fields` there. That would avoid the deferred validation inside `convert`.

lmb: Yes that's better.

jrife: Alright, I went with `Field` and moved all the logic into `newB2G` now. For good measure, I tested manually again to make sure `BPF2GO_CC="     "` results in a `no compiler specified` error and that positional args are properly passed through.

PR #1837, opened 2025-07-29
btf: don't try to load strings when file is empty

therealak12:
I opened this PR to address [issue#1818](https://github.com/cilium/ebpf/issues/1818).

In my experience, the problem has been occurring since version v0.19.0 (it does not appear in v0.18.0), specifically for the `aesni_intel` kernel module. The error encountered is:

```text
apply CO-RE relocations: load BTF for kmod aesni_intel: read string section: string table is empty
exit status 1
```

I've checked `stat /sys/kernel/btf/aesni_intel`. It shows 24 bytes, which is for the BTF header, but it has no content.

### DIFF
--- a/btf/btf_test.go
+++ b/btf/btf_test.go
@@ -2,6 +2,7 @@ package btf
 
 import (
 	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io/fs"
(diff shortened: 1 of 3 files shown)


dylandreimerink:
This seems to be a duplicate / alternative PR for https://github.com/cilium/ebpf/pull/1828/files

therealak12:
> This seems to be a duplicate / alternative PR for https://github.com/cilium/ebpf/pull/1828/files

That one returns a non-empty list while the file is empty. IMHO, this is a better solution.

lmb:
Thanks for adding the test, that was useful. Rebased this on main and changed it to use the approach where newStringTable deals with emptiness.

lmb (approved):

therealak12:
I assumed introducing an outer check on `StringLen` could help avoid unnecessary function and system calls.

lmb:
There are no system calls performed, and function calls are not really measurable at this level. Maybe you mean because the original code did some io.Read? That is gone now.

PR #1844, opened 2025-08-10
link: allow attaching with ProgramFd (needed for struct_ops)

shun159:
This PR extends RawLinkOptions with an optional ProgramFd so callers can attach link types whose target isn't a bpf_prog (e.g. struct_ops maps).

See: https://github.com/cilium/ebpf/discussions/1502

### DIFF
--- a/link/link_test.go
+++ b/link/link_test.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/go-quicktest/qt"
@@ -69,6 +70,41 @@ func TestRawLinkLoadPinnedWithOptions(t *testing.T) {
 	})
 	if !errors.Is(err, unix.EINVAL) {
 		t.Fatal("Invalid flags don't trigger an error:", err)
+	}
+}
+
+func TestRawLinkWithProgramFd(t *testing.T) {
+	cg, prog := mustCgroupFixtures(t)
+	t.Cleanup(func() {
+		_ = prog.Close()
+		_ = cg.Close()
+	})
+
+	ln, err := AttachRawLink(RawLinkOptions{
+		Target:    int(cg.Fd()),
+		Program:   nil,
+		ProgramFd: prog.FD(),
+		Attach:    ebpf.AttachCGroupInetEgress,
+	})
+	testutils.SkipIfNotSupported(t, err)
+	if err != nil {
+		t.Fatalf("AttachRawLink with ProgramFd failed: %v", err)
+	}
+	t.Cleanup(func() { _ = ln.Close() })
+}
+
+func TestRawLinkWithInvalidProgramFd(t *testing.T) {
+	cg := testutils.CreateCgroup(t)
+	t.Cleanup(func() { _ = cg.Close() })
+
+	_, err := AttachRawLink(RawLinkOptions{
+		Target:    int(cg.Fd()),
+		Program:   nil,
+		ProgramFd: -1,
+		Attach:    ebpf.AttachCGroupInetEgress,
+	})
+	if err == nil || !strings.Contains(err.Error(), "invalid program") {
+		t.Fatalf("expected invalid program error, got %v", err)
 	}
 }
 
--- a/link/link.go
+++ b/link/link.go
@@ -108,6 +108,9 @@ type RawLinkOptions struct {
 	BTF btf.TypeID
 	// Flags control the attach behaviour.
 	Flags uint32
+	// ProgramFd allows attaching with an arbitrary FD instead of Program.
+	// Useful for link types whose target is not a bpf_prog (e.g. struct_ops maps).
+	ProgramFd int
 }
 
 // Info contains metadata on a link.
--- a/link/link_other.go
+++ b/link/link_other.go
@@ -37,7 +37,10 @@ func AttachRawLink(opts RawLinkOptions) (*RawLink, error) {
 		return nil, fmt.Errorf("invalid target: %s", sys.ErrClosedFd)
 	}
 
-	progFd := opts.Program.FD()
+	progFd := opts.ProgramFd
+	if opts.Program != nil {
+		progFd = opts.Program.FD()
+	}
 	if progFd < 0 {
 		return nil, fmt.Errorf("invalid program: %s", sys.ErrClosedFd)
 	}


link/link_test.go
```diff
@@ -72,6 +73,42 @@ func TestRawLinkLoadPinnedWithOptions(t *testing.T) {
...
 }
 
+func TestRawLinkWithProgramFd(t *testing.T) {
+	cg, prog := mustCgroupFixtures(t)
+	t.Cleanup(func() {
+		_ = prog.Close()
+		_ = cg.Close()
+	})
+
+	ln, err := AttachRawLink(RawLinkOptions{
```
lmb: How would you use this with a struct ops map? `Target` is the map and ProgramFd is what exactly?

shun159: @lmb For struct_ops we follow libbpf’s semantics:

https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L13163
```c
struct bpf_link *bpf_map__attach_struct_ops(const struct bpf_map *map)
{
	struct bpf_link_struct_ops *link;
	__u32 zero = 0;
	int err, fd;
.
.
.
	fd = bpf_link_create(map->fd, 0, BPF_STRUCT_OPS, NULL);
	if (fd < 0) {
		free(link);
		return libbpf_err_ptr(fd);
	}
}
bpf_link_create(/* prog_fd = */ map_fd, /* target_fd = */ 0, BPF_STRUCT_OPS, …)
```

So the intent is to pass the struct_ops map FD as ProgramFd and not use Target (i.e., target_fd = 0). Usage would look like:

```go
m := objs.AStructOpsMap
ln, err := link.AttachRawLink(link.RawLinkOptions{
    ProgramFd: int(m.FD()),    // map FD goes into prog_fd
    Attach:    ebpf.AttachStructOps,
    // Target is unused (0)
})
```

lmb: In that case it seems better to just add a new, custom API specifically targeted at StructOps. Then the argument can simply be a `*Map`. Taking an int like this is confusing, but also unsafe: `int(m.Fd())` doesn't protect m from being closed by the GC.

```
func AttachStructOps(opts StructOpsOptions) (*RawLink, error)
```

shun159: @lmb sounds good. let me try this in this PR. how do you see this shape?
```go
type StructOpsOptions struct {
    Map   *ebpf.Map // required
    Flags uint32    // optional link_create flags
}

func AttachStructOps(opts StructOpsOptions) (*RawLink, error)
```

shun159: let me revert the changes to main

lmb: I'm guessing there must be some way to specify more parameters to the attachment though? What if there are multiple instances of the same struct ops? (I guess what I'm asking: given multiple sched_ext maps, which one ends up being active? Can you only create a single link?)

shun159: @lmb 

> What if there are multiple instances of the same struct ops?

let's say we try to load multiple sched_ext maps, in that case, it should be failed with EBUSY (in practical cases, for each struct_ops can have only one instance for struct_ops map). in other words, it should be selected by like "first load wins"

btw, but (technically) one ELF file can have 2 or more different struct_ops instances, 
for example:
```c
__section(".struct_ops.link") struct bpf_testmod_ops2 dummy_ops_1 = {
	.test_1 = (void *)dummy_test_1,
};
__section(".struct_ops.link") struct bpf_testmod_ops3 dummy_ops_2 = {
	.test_1 = (void *)dummy_test_2,
	.test_2 = (void *)dummy_test_3,
};
```

even in that case, because one program can belong to only one struct_ops instance,
this case, should be failed
```c
__section("struct_ops/dummy_test_1") int dummy_test_1(void) {
	return 0;
}

__section("struct_ops/dummy_test_1") int dummy_test_2(void) {
	return 0;
}

__section(".struct_ops.link") struct bpf_testmod_ops2 dummy_ops_1 = {
	.test_1 = (void *)dummy_test_1,
};
__section(".struct_ops.link") struct bpf_testmod_ops3 dummy_ops_2 = {
	.test_1 = (void *)dummy_test_1,
	.test_2 = (void *)dummy_test_2,
};
```

shun159:
depends on https://github.com/cilium/ebpf/pull/1845. we can proceed with this once the MR get merged.

shun159:
@lmb I changed this MR to draft since it depends on #1869 and #1845.

link/struct_ops.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, fmt.Errorf("map cannot be nil")
+	}
+
+	if t := m.Type(); t != ebpf.StructOpsMap {
+		return nil, fmt.Errorf("invalid map type %s, expected struct_ops", t)
+	}
+
+	mapFD := m.FD()
+	if mapFD <= 0 {
+		return nil, fmt.Errorf("invalid map: %s (was it created?)", sys.ErrClosedFd)
```
lmb: Nit: I think the extra message is a bit confusing: you can only get a `*Map` by creating it.

```suggestion
		return nil, fmt.Errorf("invalid map: %s", sys.ErrClosedFd)
```

link/struct_ops.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	// ".struct_ops.link" requires the map to be created with BPF_F_LINK.
+	if (int(m.Flags()) & sys.BPF_F_LINK) != sys.BPF_F_LINK {
```
lmb: I'm guessing that the kernel does this check already? I'd probably drop this, since it doesn't matter from the library side whether the flag is set (it does matter that we don't create ones without, but that is already taken care of.)

link/struct_ops.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, fmt.Errorf("map is missing BPF_F_LINK flag: %w", ErrNotSupported)
+	}
+
+	fd, err := sys.LinkCreate(&sys.LinkCreateAttr{
+		// struct_ops expects target_fd = map FD
```
lmb: This comment confuses me because below target_fd = 0 and we set prog_fd instead.

link/struct_ops_test.go
```diff
@@ -0,0 +1,42 @@
+package link
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+func TestStructOps(t *testing.T) {
```
lmb, line 17: Mark this file as `go:build !windows`.

lmb (changes requested):

link/struct_ops.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+type structOpsLink struct {
+	*RawLink
+}
+
+// AttachStructOps attaches a struct_ops map (created from a ".struct_ops.link"
+// section) to its kernel subsystem via a BPF link.
+func AttachStructOps(m *ebpf.Map) (Link, error) {
```
lmb: Is there a chance we'll ever have to support more arguments? If yes it might make sense to introduce StructOpsOptions from the start.

link/struct_ops.go
shun159: > I'm guessing that the kernel does this check already?

yes, you're right.  
I think we don't need to do that here. I will drop.

link/struct_ops.go
shun159: I aligned the implementation to be closer to how XDPOptions is used.

link/struct_ops.go
shun159: done

link/struct_ops.go
shun159: fixed

link/struct_ops.go
shun159: yes....thx. fixed.

link/struct_ops_test.go
shun159: thx. I've fixed but still failing with `Error: Build-x64-Debug artifact not found in the workflow run
`

link/struct_ops_test.go
```diff
@@ -0,0 +1,48 @@
...
+	testutils.SkipOnOldKernel(t, "6.12", "bpf_testmod_ops")
+
+	coll, err := mustStructOpsFixtures(t)
+	if errors.Is(err, ErrNotSupported) {
+		t.Skipf("bpf_testmod_ops: %s", err)
+	}
+	if err != nil && strings.Contains(err.Error(), "unknown struct_ops target") {
```
shun159: I don't think this is good...

shun159:
it looks like the tests are also failing on arm64.
Has there been any update to the VM image within the last few hours?

lmb:
Windows has fixed itself, arm64 hasn't been changed. Made an issue: https://github.com/cilium/ebpf/issues/1888

link/struct_ops_test.go
lmb: What kernel is this required for? Would it make sense to fold this into `ErrNotSupported`?

link/struct_ops_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	l, err := AttachStructOps(StructOpsOptions{Map: m})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer l.Close()
+
+	// Close the program and map on test teardown.
+	t.Cleanup(func() {
```
lmb: Nit: add a cleanup for `Collection.Close` in `mustStructOpsFixtures` instead.

shun159: thx. fixed.

link/struct_ops_test.go
shun159: it requires a kernel where bpf_testmod_ops is available. I've updated this to wrap "unknown struct_ops target" as ErrNotSupported.

shun159:
> Windows has fixed itself, arm64 hasn't been changed. Made an issue: https://github.com/cilium/ebpf/issues/1888

got it :saluting_face:

link/link_other_test.go
```diff
@@ -112,6 +115,96 @@ func mustCgroupFixtures(t *testing.T) (*os.File, *ebpf.Program) {
...
+					asm.Return(),
+				},
+			},
+		},
+		Variables: map[string]*ebpf.VariableSpec{},
+	}
+
+	coll, err := ebpf.NewCollection(spec)
+	if err != nil {
+		if strings.Contains(err.Error(), "unknown struct_ops target") {
```
shun159: @lmb 
> Would it make sense to fold this into ErrNotSupported?

is this align with your thought?

shun159:
@lmb I’m really happy that I was able to make this contribution with you. Thank you so much!

PR #1845, opened 2025-08-10
struct_ops: add structOpsMeta to carry BTF hints for StructOpsMap creation

shun159:
1. Verify creating a StructOpsMap from a hand-crafted MapSpec with metadata.
2. Resolve attr.BtfVmlinuxValueTypeId from vmlinux BTF before MapCreate.
3. Value population is deferred to follow-up PRs.

see: https://github.com/cilium/ebpf/discussions/1502

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -44,6 +44,36 @@ func requireTestmod(tb testing.TB) {
 	}
 }
 
+var haveStructOpsDummy = sync.OnceValues(func() (bool, error) {
+	if platform.IsWindows {
+		return false, nil
+	}
+	kspec, err := btf.LoadKernelSpec()
+	if err != nil {
+		return false, nil
+	}
+	_, err = kspec.AnyTypeByName("bpf_struct_ops_bpf_dummy_ops")
+	if err == nil {
+		return true, nil
+	}
+	if errors.Is(err, btf.ErrNotFound) {
+		return false, nil
+	}
+	return false, err
+})
+
+func requireStructOpsDummy(tb testing.TB) {
+	tb.Helper()
+
+	ok, err := haveStructOpsDummy()
+	if err != nil {
+		tb.Fatal(err)
+	}
+	if !ok {
+		tb.Skip("struct_ops dummy_ops wrapper type not present in vmlinux BTF")
+	}
+}
+
 func newMap(tb testing.TB, spec *MapSpec, opts *MapOptions) (*Map, error) {
 	tb.Helper()
 
--- a/map.go
+++ b/map.go
@@ -556,6 +556,47 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
 		}
 	}
 
+	if spec.Type == StructOpsMap {
+		meta, err := extractStructOpsMeta(spec.Contents)
+		if err != nil {
+			return nil, err
+		}
+
+		// we need drop meta entry here
+		if len(spec.Contents) > 0 {
+			spec.Contents = spec.Contents[1:]
+		}
+
+		var b btf.Builder
+		h, err := btf.NewHandle(&b)
+		if err != nil {
+			return nil, err
+		}
+		defer h.Close()
+
+		s, err := btf.LoadKernelSpec()
+		if err != nil {
+			return nil, fmt.Errorf("open vmlinux BTF: %w", err)
+		}
+
+		typeSpec, err := s.AnyTypeByName(meta.kernTypeName)
+		if errors.Is(err, btf.ErrNotFound) {
+			return nil, fmt.Errorf("struct_ops kernel type %q: %w", meta.kernTypeName, ErrNotSupported)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("lookup kernel type %q: %w", meta.kernTypeName, err)
+		}
+
+		btfValueTypeId, err := s.TypeID(typeSpec)
+		if err != nil {
+			return nil, fmt.Errorf("lookup type_id of %s: %w", typeSpec.TypeName(), err)
+		}
+
+		attr.ValueSize = spec.ValueSize
+		attr.BtfVmlinuxValueTypeId = btfValueTypeId
+		attr.BtfFd = uint32(h.FD())
+	}
+
 	fd, err := sys.MapCreate(&attr)
 
 	// Some map types don't support BTF k/v in earlier kernel versions.
--- a/struct_ops.go
+++ b/struct_ops.go
@@ -1,0 +1,42 @@
+package ebpf
+
+import "fmt"
+
+// structOpsMeta is a placeholder object inserted into MapSpec.Contents
+// so that later stages (loader, ELF parser) can recognise this map as
+// a struct‑ops map without adding public fields yet.
+type structOpsMeta struct {
+	userTypeName string
+	kernTypeName string
+	members      []struct {
+		name     string
+		userOfs  uint32
+		size     uint32
+		kind     uint8
+		progName string
+	}
+	initUserBlob []byte
+}
+
+// extractStructOpsMeta returns the *structops.Meta embedded in a MapSpec’s Contents
+// according to the struct-ops convention:
+//
+//	contents[0].Key   == uint32(0)
+//	contents[0].Value == *structopsMeta
+func extractStructOpsMeta(contents []MapKV) (*structOpsMeta, error) {
+	if len(contents) == 0 {
+		return nil, fmt.Errorf("struct_ops: missing meta at Contents[0]")
+	}
+
+	k, ok := contents[0].Key.(uint32)
+	if !ok || k != 0 {
+		return nil, fmt.Errorf("struct_ops: meta key must be 0")
+	}
+
+	meta, ok := contents[0].Value.(structOpsMeta)
+	if !ok {
+		return nil, fmt.Errorf("struct_ops: meta value must be structOpsMeta")
+	}
+
+	return &meta, nil
+}
--- a/struct_ops_test.go
+++ b/struct_ops_test.go
@@ -1,0 +1,35 @@
+package ebpf
+
+import (
+	"testing"
+
+	"github.com/cilium/ebpf/internal/testutils"
+)
+
+func TestCreateStructOpsMapSpecSimple(t *testing.T) {
+	requireStructOpsDummy(t)
+
+	ms := &MapSpec{
+		Name:       "dummy_ops",
+		Type:       StructOpsMap,
+		KeySize:    4,
+		ValueSize:  128,
+		MaxEntries: 1,
+		Contents: []MapKV{
+			{
+				Key: uint32(0),
+				Value: structOpsMeta{
+					userTypeName: "bpf_dummy_ops",
+					kernTypeName: "bpf_struct_ops_bpf_dummy_ops",
+				},
+			},
+		},
+	}
+
+	m, err := NewMap(ms)
+	testutils.SkipIfNotSupported(t, err)
+	if err != nil {
+		t.Fatalf("creating struct_ops map failed: %v", err)
+	}
+	t.Cleanup(func() { _ = m.Close() })
+}


helpers_test.go
```diff
@@ -44,6 +44,36 @@ func requireTestmod(tb testing.TB) {
 	}
 }
 
+var haveStructOpsDummy = sync.OnceValues(func() (bool, error) {
```
lmb: This should probably be a feature test. See `haveFoo()` functions in the codebase.

map.go
```diff
@@ -556,6 +556,47 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+			return nil, err
+		}
+		defer h.Close()
+
+		s, err := btf.LoadKernelSpec()
+		if err != nil {
+			return nil, fmt.Errorf("open vmlinux BTF: %w", err)
+		}
+
+		typeSpec, err := s.AnyTypeByName(meta.kernTypeName)
```
lmb: This doesn't have to be a `Struct`?

map.go
```diff
@@ -556,6 +556,47 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
 		}
 	}
 
+	if spec.Type == StructOpsMap {
```
lmb: I think this case needs to be in the `spec.Key != nil || spec.Value != nil` block that handles map BTF. This is just a special case of that.

collection.go
```diff
@@ -268,6 +268,11 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 	}
 	defer loader.close()
 
+	// initialize struct_ops maps
```
lmb: It should be possible to load a struct ops map via NewCollections, without LoadAndAssign. Putting code here makes me thing that currently doesnt work?

collection.go
```diff
@@ -303,6 +308,11 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
 		return err
 	}
 
+	// load programs to struct_ops maps
+	if err := loader.stOps.postLoad(loader.maps); err != nil {
```
lmb: Why can't this be part of populateDeferredMaps()?

collection.go
```diff
@@ -413,6 +423,7 @@ type collectionLoader struct {
 	programs map[string]*Program
 	vars     map[string]*Variable
 	types    *btf.Cache
+	stOps    *structOpsLoader
```
lmb: Why do you put this on a separate type? My instinct would be to just integrate with the collectionLoader.

collection_test.go
```diff
@@ -771,3 +771,40 @@ func ExampleCollectionSpec_LoadAndAssign() {
...
+		Maps: map[string]*MapSpec{
+			"dummy_ops": {
+				Name:       "dummy_ops",
+				Type:       StructOpsMap,
+				KeySize:    4,
+				ValueSize:  128,
+				MaxEntries: 1,
+				Contents: []MapKV{
+					{Key: uint32(0), Value: structOpsMeta{
+						userTypeName: "bpf_dummy_ops",
```
lmb: I think the API shouldn't rely on referring to the user type as a string. Instead we should use MapSpec.Value to store the user type.

```
MapSpec.Value = &btf.Struct{Name: "bpf_dummy_ops", ...}
```

I suspect that we also don't need the kernTypeName, since that can just be MapSpec.Value.Name.

shun159:
> This PR is missing the part where program fds are inserted into the struct ops map. Without that it's hard to tell what is going on.

@lmb yeah, agreed.
I was trying to keep the PR small, but I see how it’s unclear without kern_vdata population.
then, let's proceed with including the changes so that make it clear.

collection_test.go
shun159: @lmb 

Let me confirm whether my understanding is correct.
If that’s the case, `structOpsMeta` should be sufficient with contents at about this level.

```go
type structOpsFunc struct {
	member   string <------ member name of the struct_ops
	progName string <------ program name which is stored in collection spec
}

// structOpsMeta is a placeholder object inserted into MapSpec.Contents
// so that later stages (loader, ELF parser) can recognise this map as
// a struct-ops map without adding public fields yet.
type structOpsMeta struct {
	funcs []structOpsFunc
	data  []byte <----- use to store values other than funcpointers
}
```

Overall, it would look like this:
```go
ms := &MapSpec{
	Name:       "dummy_ops",
	Type:       StructOpsMap,
	KeySize:    4,
	ValueSize:  128,
	MaxEntries: 1,
	// we use `Value` to specify a user struct type as BTF
	Value: &btf.Struct{Name: "bpf_dummy_ops"},
                                   ^^^^^^^^^^^^ 
        Since the user struct name can be derived from the struct name in Value, 
        this should be sufficient.
	Contents: []MapKV{
		{
			Key: uint32(0),
			Value: structOpsMeta{
				data: []byte{},
				funcs: []structOpsFunc{
                    Since the func offset can also be derived from the member name,
                    this should be sufficient.
					{"test_1", "dummy_test_1"},
					{"test_2", "dummy_test_2"},
				},
			},
		},
	},
}
```

collection.go
shun159: let's make it be a part of populateDeferredMaps

helpers_test.go
shun159: @lmb 

Understood. However, feature testing with dummy_ops has limitations (since dummy_ops isn’t linkable). How about switching to a test that uses bpf_testmod_ops, which is linkable? The downside is that struct_ops would also need mod_btf support, so the diff will get larger.

map.go
shun159: We should check if it is a `Struct`. fixed.

map.go
shun159: agreed. In this case, we should put them in the block. fixed.

collection_test.go
shun159: For now, I'm proceeding as above. Since dummy_ops doesn't allow linking and programs need an AttachType specified before loading, I'm first testing whether a map can be created from the MapSpec. At a later step, I'll switch the tests to use testmod_ops.

helpers_test.go
shun159: @lmb this is not needed. removed for now.

collection_test.go
lmb: Follow up question: what does the value BTF look like for a struct ops map defined in C? I found https://github.com/torvalds/linux/blob/0cc53520e68bea7fb80fdc6bdf8d226d1b6a98d9/tools/testing/selftests/sched_ext/maybe_null.bpf.c is that the common use case?

I think that I misunderstood how struct ops maps are created: they don't really appear in the source code, instead we need the magic .struct_ops.link sections? Those then get turned into maps behind the scenes.

shun159:
vimto tests on 6.1 and 6.6 has passed on my setup, so I feel this is a temporary error.

collection.go
```diff
@@ -697,6 +757,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		if ms.Type != StructOpsMap {
+			continue
+		}
+
+		userSt := ms.Value
+		if userSt == nil {
+			return fmt.Errorf("user struct type should be specified as Value")
+		}
+
+		userStructType, ok := ms.Value.(*btf.Struct)
```
lmb: Maybe `btf.As`?

collection.go
```diff
@@ -697,6 +757,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		if !ok {
+			return fmt.Errorf("user struct type should be a Struct")
+		}
+
+		// resolve kernel-side types (target + wrapper)
+		kernTypes, err := findStructOpsKernTypes(userStructType)
+		if err != nil {
+			return fmt.Errorf("find kern_type: %w", err)
+		}
+		ms.Value = kernTypes.typ
```
lmb: MapSpec in a collection must never be modified since they may be loaded multiple times. If necessary you can clone the spec.

map.go
```diff
@@ -541,18 +541,75 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
 
 	if spec.Key != nil || spec.Value != nil {
-		handle, keyTypeID, valueTypeID, err := btf.MarshalMapKV(spec.Key, spec.Value)
-		if err != nil && !errors.Is(err, btf.ErrNotSupported) {
-			return nil, fmt.Errorf("load BTF: %w", err)
-		}
+		if spec.Type == StructOpsMap {
+			// we need drop meta entry here
+			if len(spec.Contents) > 0 {
+				spec.Contents = spec.Contents[1:]
```
lmb: Should this be an error instead?

map.go
```diff
@@ -541,18 +541,75 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+			}
+
+			var b btf.Builder
+			h, err := btf.NewHandle(&b)
```
lmb: Why create an empty BTF blob here?

map.go
```diff
@@ -541,18 +541,75 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+				// BPF_F_VTYPE_BTF_OBJ_FD is required if the type comes from a module
+				attr.MapFlags |= sys.BPF_F_VTYPE_BTF_OBJ_FD
+				// resolve a module handler for the kernel model
+				modH, err := btf.NewHandleFromID(btf.ID(modBtfObjId))
+				if err != nil {
+					return nil, fmt.Errorf("open module BTF (id=%d): %w", modBtfObjId, err)
+				}
+				// set FD for the kernel module
+				attr.ValueTypeBtfObjFd = int32(modH.FD())
+				defer modH.Close()
```
lmb: Nit: please move this after the `if err` check.

collection.go
```diff
@@ -697,6 +757,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		}
+
+		// where the kernel expects the prog FD to be written
+		kernFuncOff := kern.dataMember.Offset/8 + kern.typ.Members[kernMemberIdx].Offset/8
+		structOps.kernFuncOff[idx] = int(kernFuncOff)
+		structOps.progAttachType[ps.Name] = attachType
+		cl.stOpsProgsToMap[ps.Name] = ms.Name
+
+		// seed program with attach metadata (checked at load time)
+		ps.Instructions[0].Metadata.Set(structOpsProgMetaKey{}, &structOpsProgMeta{
```
lmb: Again, can't modify ps.Instructions here, similar to MapSpec.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+//
+// It expects val to be the user-visible struct type (e.g. "bpf_dummy_ops")
+// and looks up the kernel-side wrapper name:
+//
+//	"bpf_dummy_ops" -> "bpf_struct_ops_bpf_dummy_ops"
+//
+// Returns the matching *btf.Struct, the *btf.Spec it was found in (either the
+// base vmlinux spec or a module spec), and the module BTF ID (0 for vmlinux).
+// See doFindStructTypeByName for resolution details and error behavior.
+func findStructByNameWithPrefix(s *btf.Spec, val *btf.Struct) (*btf.Struct, *btf.Spec, uint32, error) {
```
lmb: Please don't add very short functions like this unless absolutely necessary.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+
+// skipModsAndTypedefs resolves a single layer of BTF indirection/qualification
+// for the given type within the provided *btf.Spec.
+//
+// Behavior:
+//   - Uses s.TypeID(typ) and s.TypeByID(id) to canonicalize the type within s.
+//   - If the resolved type is a Typedef or C qualifier (Const/Volatile/Restrict),
+//     returns its immediate underlying type (one level).
+//   - Otherwise returns the resolved type as-is.
+func skipModsAndTypedefs(s *btf.Spec, typ btf.Type) (btf.Type, error) {
```
lmb: Why have this instead of `UnderlyingType`?

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+		dataMember:  dataMem,
+		modBtfObjId: uint32(modID),
+	}, nil
+}
+
+// skipModsAndTypedefs resolves a single layer of BTF indirection/qualification
+// for the given type within the provided *btf.Spec.
+//
+// Behavior:
+//   - Uses s.TypeID(typ) and s.TypeByID(id) to canonicalize the type within s.
```
lmb: What does canonicalize type mean? IIRC doing this is pure overhead.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+// Search order and behavior:
+//  1. vmlinux (base spec): try AnyTypeByName(name). If it exists and is a
+//     *btf.Struct, return it immediately with moduleID=0.
+//     - If AnyTypeByName returns a non-notfound error, the error is propagated.
+//     - If a type is found but is not a *btf.Struct, we fall back to modules.
+//  2. modules: see findStructTypeByNameFromModule.
+//
+// Returns (*btf.Struct, *btf.Spec, moduleID, nil) on success, or btf.ErrNotFound
+// if no matching struct is present in vmlinux or any module.
+func doFindStructTypeByName(s *btf.Spec, name string) (*btf.Struct, *btf.Spec, uint32, error) {
```
lmb: Why doesn't this reuse `findTargetInKernel`?

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// findStructTypeByNameFromModule scans all loaded kernel modules and tries
+// to resolve a struct type with the exact name. The iteration uses the base
+// vmlinux spec for string/ID interning as required by btf.Handle.Spec(base).
+//
+// For the first module where AnyTypeByName(name) returns a *btf.Struct,
+// the function returns that struct, the module's *btf.Spec, and its BTF ID.
+// If the type is not found in any module, btf.ErrNotFound is returned.
+func findStructTypeByNameFromModule(base *btf.Spec, name string) (*btf.Struct, *btf.Spec, uint32, error) {
```
lmb: Dito, not neccessary?

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// findStructTypeByName resolves the exact BTF struct type that corresponds
+// to typ.TypeName() by searching first in vmlinux and then across all loaded
+// kernel modules.
+//
+// Returns the first *btf.Struct that matches the name verbatim, the *btf.Spec
+// where it was found, and the module BTF ID (0 if found in vmlinux).
+// If no matching struct exists anywhere, btf.ErrNotFound is returned.
+func findStructTypeByName(s *btf.Spec, typ *btf.Struct) (*btf.Struct, *btf.Spec, uint32, error) {
```
lmb: Dito, see above.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+// findByTypeFromStruct searches for the first member of a struct whose
+// resolved BTF type ID matches the given typ.
+//
+// It resolves the BTF type ID of typ and compares it against each
+// member’s TypeID in st.Members. If a match is found, the corresponding
+// *btf.Member is returned.
+//
+// Returns an error if typ cannot be resolved, if any member type
+// resolution fails, or if no member with the requested type exists.
+func findByTypeFromStruct(spec *btf.Spec, st *btf.Struct, typ btf.Type) (*btf.Member, error) {
```
lmb: This is a bit surprising to me. So far all other equality comparisons we have are based on concrete type and name. Why compare by ID here all of a sudden?

collection_test.go
```diff
@@ -771,3 +772,62 @@ func ExampleCollectionSpec_LoadAndAssign() {
...
+		},
+		Maps: map[string]*MapSpec{
+			"testmod_ops": {
+				Name:       "testmod_ops",
+				Type:       StructOpsMap,
+				Flags:      sys.BPF_F_LINK,
+				KeySize:    4,
+				ValueSize:  448,
+				MaxEntries: 1,
+				Value:      &btf.Struct{Name: "bpf_testmod_ops"},
```
lmb: I expec this struct to contain members which match the actual type we will get from the ELF. So roughly https://elixir.bootlin.com/linux/v6.16.6/source/tools/testing/selftests/bpf/test_kmods/bpf_testmod.h#L35 For the purpose of this test we can only specify the first couple of function pointers.

This means that more of the metadata we need is embedded in MapSpec.Value instead of structOpsMeta.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+	 *   struct bpf_struct_ops_tcp_congestion_ops {
+	 *       [... kernel internal fields ...]
+	 *       struct tcp_congestion_ops data;
+	 *   }
+	 *
+	 * In this case, len(kernVData) == sizeof(struct bpf_struct_ops_tcp_congestion_ops).
+	 * copyDataMember() will copy user-supplied data
+	 * into kernVData, which is then pushed into the map.
+	 */
+	kernVData []byte
```
lmb: This can't be in a structure referenced from MapSpec, since that needs to be immutable.

collection.go
lmb: Also, this requires map loads to happen before program loads. Seems kind of brittle.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+		// target kernel struct type (e.g. tcp_congestion_ops).
+	}
+	return btf.Member{}, fmt.Errorf("member %s not found in struct %s", name, s.Name)
+}
+
+// getStructMemberIndexOf returns the index of `member` within struct `s` by
+// comparing the member’s bit-offset.
+func getStructMemberIndexOf(s *btf.Struct, member btf.Member) int {
+	for idx, m := range s.Members {
+		if m.Offset == member.Offset {
```
lmb: Again, why not compare by name?

prog.go
```diff
@@ -387,6 +387,26 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
 		}
 	}
 
+	if spec.Type == StructOps {
```
lmb: This should not be a separate if statement, but be part of the `spec.AttachTo != ""` statement above. Add a case for `match{StructOps, ...}` to `findProgramTargetInKernel`.

```
} else if spec.AttachTo != "" {
	attachTo := spec.AttachTo
	var targetMember string
	if spec.Type == StructOps {
		attachTo, targetMember, ok := strings.Cut(attachTo, ":")
		if !ok {
			return error
		}
	}

	// NB: new target return value
	module, target, id, err := findProgramTargetInKernel(...)

	if targetMember != "" {
		// find member idx in target
		attr.ExpectedAttachType = idx
	}
}
```

collection_test.go
```diff
@@ -771,3 +772,62 @@ func ExampleCollectionSpec_LoadAndAssign() {
...
+	spec := &CollectionSpec{
+		Programs: map[string]*ProgramSpec{
+			"test_func_1": {
+				Name:    "test_func_1",
+				Type:    StructOps,
+				License: "GPL",
+				Instructions: asm.Instructions{
+					asm.Mov.Imm(asm.R0, 0),
+					asm.Return(),
+				},
```
lmb, line 778: We should encode the attach target here instead of relying on `structOpsProgMeta`. `AttachTo: "bpf_testmod_ops:test_1",` here since that allows simplifying the logic to retrieve the attach target in prog.go.

struct_ops.go
```diff
@@ -0,0 +1,343 @@
...
+
+	"github.com/cilium/ebpf/btf"
+	"github.com/cilium/ebpf/internal/sys"
+)
+
+const structOpsValuePrefix = "bpf_struct_ops_"
+
+// TODO: Doc
+type structOpsProgMetaKey struct{}
+type structOpsProgMeta struct {
```
lmb: Not necessary anymore.

collection_test.go
```diff
@@ -771,3 +772,62 @@ func ExampleCollectionSpec_LoadAndAssign() {
...
+			"testmod_ops": {
+				Name:       "testmod_ops",
+				Type:       StructOpsMap,
+				Flags:      sys.BPF_F_LINK,
+				KeySize:    4,
+				ValueSize:  448,
+				MaxEntries: 1,
+				Value:      &btf.Struct{Name: "bpf_testmod_ops"},
+				Contents: []MapKV{
+					{Key: uint32(0), Value: structOpsMeta{
```
lmb: Get rid of `structOpsMeta` and just store the pure data from the ELF here. For now `{Key: uint32(0), Value: make([]byte, 448)}` will suffice.

collection.go
```diff
@@ -697,6 +757,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		return fmt.Errorf("size mismatch for member %s: %d != %d (kernel)", memberName, memberSize, kernMemberSize)
+	}
+	copy(kernMemberData[:memberSize], memberData[:memberSize])
+
+	return nil
+}
+
+// initKernStructOps indexes struct_ops maps: resolve kernel types, allocate per-map state,
+// and stage copy/attach metadata. No kernel objects are created here.
+func (cl *collectionLoader) initKernStructOps() error {
```
lmb: The metadata shouldn't be computed up front, but instead be retrieved when populateDeferredMaps is invoked. This was only necessary because you had to populate the program instruction metadata, which we can fix by using ProgramSpec.AttachTo.

collection.go
```diff
@@ -697,6 +757,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
 	return nil
 }
 
+// copyDataMember copies one field from the user struct into the kernel buffer.
+// If the field is a function pointer, record attach metadata instead.
+func (cl *collectionLoader) copyDataMember(
```
lmb: This function has the wrong scope and does too many things at once. Please split it like so:

```
// fills in all fields in from that exist in to with the contents of fromData.
func translateStructData(from *btf.Struct, fromData []byte, to *btf.Struct) (toData []byte, _ error) 
func populateFuncPtr(typ *btf.Struct, data []byte, programs map[string]*Program) error
```

The resulting code (invoked from populateDeferredMaps) will look something like:

```
kernData, err := translateStructData(userType, ms.Contents[0].Value.([]byte), kernType)
err := populateFuncPtr(kernType, kernData)
ms.Contents = []MapKV{{uint32(0), kernData}}
```

collection_test.go
```diff
@@ -771,3 +772,62 @@ func ExampleCollectionSpec_LoadAndAssign() {
...
 	defer objs.Map.Close()
 }
+
+func TestStructOpsMapSpecSimpleLoadAndAssign(t *testing.T) {
+	requireTestmod(t)
+
+	spec := &CollectionSpec{
+		Programs: map[string]*ProgramSpec{
+			"test_func_1": {
+				Name:    "test_func_1",
```
lmb:
```suggestion
				Name:    "test_1",
```

lmb:
@shun159 please ping me explicitly when you need another review.

shun159:
> @shun159 please ping me explicitly when you need another review.

Hi @lmb I believe I've addressed most of your feedbacks earlier. 
However, after switching to use AttachTo, it looks like elf_reader_test is now failing. 
My assumption is that if the AttachTo string doesn't match the expected format introduced by this PR, we could simply ignore it, in which case the test itself should pass but I would appreciate your thought how we should proceed with this?

https://github.com/cilium/ebpf/actions/runs/17712680837/job/50333512472?pr=1845

struct_ops_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}
+
+	s, err := btf.LoadKernelSpec()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	target := btf.Type((*btf.Struct)(nil))
+	_, module, err := findTargetInKernel(s, "bpf_struct_ops_bpf_testmod_ops", &target)
```
lmb: TODO: Why is this necessary?

types.go
```diff
@@ -144,7 +144,7 @@ func (mt MapType) hasPerCPUValue() bool {
 // canStoreMapOrProgram returns true if the Map stores references to another Map
 // or Program.
 func (mt MapType) canStoreMapOrProgram() bool {
-	return mt.canStoreMap() || mt.canStoreProgram()
+	return mt.canStoreMap() || mt.canStoreProgram() || mt.canStoreSturct()
```
lmb: Nit: typo. Since this is the only callsite of canStoreStruct I'd opt for just inlining `|| mt == StructOps` here.
```suggestion
	return mt.canStoreMap() || mt.canStoreProgram() || mt == StructOps
```

prog.go
```diff
@@ -381,15 +382,63 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
+			}
+
+			target := btf.Type((*btf.Struct)(nil))
+			s, module, err := findTargetInKernel(s, attachTo, &target)
+			if err != nil {
+				return nil, fmt.Errorf("lookup struct_ops kern type %q: %w", attachTo, err)
+			}
+			defer module.Close()
+
+			kType, ok := btf.As[*btf.Struct](target)
```
lmb: You can just use a plain type assertion here since we initialize target just above.

prog.go
```diff
@@ -381,15 +382,63 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
+				return nil, fmt.Errorf("conversion target %s -> Struct for prog", attachTo)
+			}
+
+			tid, err := s.TypeID(kType)
+			if err != nil {
+				return nil, fmt.Errorf("type id for %s: %w", kType.TypeName(), err)
+			}
+			attr.AttachBtfId = sys.TypeID(tid)
+
+			idx := getStructMemberIndexByName(kType, targetMember)
```
lmb: I think you can replace this with `slices.IndexFunc`.

prog.go
```diff
@@ -381,15 +382,63 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
+			if idx < 0 {
+				return nil, fmt.Errorf("member %q not found in %s", targetMember, kType.Name)
+			}
+
+			attr.ExpectedAttachType = sys.AttachType(idx)
```
lmb, line 426: Please add a comment explaining this StructOps quirk.

prog.go
```diff
@@ -381,15 +382,63 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
+
+			if module != nil {
```
lmb: Nit: why duplicate this if block? Just move the whole `if Type == StructOps` block after this?

map.go
lmb: This question is still unanswered.

map.go
```diff
@@ -547,18 +547,71 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+			}
+			defer h.Close()
+
+			s, err := btf.LoadKernelSpec()
+			if err != nil {
+				return nil, fmt.Errorf("open vmlinux BTF: %w", err)
+			}
+
+			if spec.Value == nil {
+				return nil, fmt.Errorf("struct type is not specified as Value")
```
lmb: I think this error message is a bit confusing.
```suggestion
				return nil, fmt.Errorf("struct_ops map: missing value type information")
```

map.go
```diff
@@ -547,18 +547,71 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+			// struct_ops: resolve value type ("bpf_struct_ops_<name>") and
+			// record kernel-specific BTF IDs / FDs needed for map creation.
+			target := btf.Type((*btf.Struct)(nil))
+			s, module, err := findTargetInKernel(s, valueType.Name, &target)
+			if err != nil {
+				return nil, fmt.Errorf("lookup value type %q: %w", valueType.Name, err)
+			}
+			defer module.Close()
+
+			vType, ok := btf.As[*btf.Struct](target)
```
lmb: Plain assert is fine.

map.go
```diff
@@ -547,18 +547,71 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+				return nil, err
+			}
+			defer h.Close()
+
+			s, err := btf.LoadKernelSpec()
+			if err != nil {
+				return nil, fmt.Errorf("open vmlinux BTF: %w", err)
+			}
+
+			if spec.Value == nil {
```
lmb: What about `spec.Key`? We should make sure that has a sensible value as well.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+// translateStructData fills in all fields in `from` that exist in `to` with contents of fromData.
+func translateStructData(from *btf.Struct, fromData []byte, to *btf.Struct) ([]byte, error) {
+	innerName := strings.TrimPrefix(to.Name, structOpsValuePrefix)
+	toData := make([]byte, int(to.Size))
+
+	var inner *btf.Struct
+	var innerOff int
+
+	for _, m := range to.Members {
+		st, ok := btf.As[*btf.Struct](btf.UnderlyingType(m.Type))
```
lmb: `As` and `UnderlyingType` both get rid of qualifiers. You don't need both here.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+			inner = st
+			innerOff = int(m.Offset / 8)
+		}
+	}
+
+	if inner == nil {
+		return nil, fmt.Errorf("member %s not found in %s", innerName, to.Name)
+	}
+
+	kernIndexByName := make(map[string]btf.Member, len(inner.Members))
```
lmb: This should probably a pointer, otherwise member is copied a lot.
```suggestion
	kernIndexByName := make(map[string]*btf.Member, len(inner.Members))
```

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+
+	for _, m := range to.Members {
+		st, ok := btf.As[*btf.Struct](btf.UnderlyingType(m.Type))
+		if ok && st.Name == innerName {
+			inner = st
+			innerOff = int(m.Offset / 8)
+		}
+	}
+
+	if inner == nil {
```
lmb: I think this lookup of the inner type should be a separate function with a doc comment which explains what it does. Right now its kind of hard to figure out.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+	toData := make([]byte, int(to.Size))
+
+	var inner *btf.Struct
+	var innerOff int
+
+	for _, m := range to.Members {
+		st, ok := btf.As[*btf.Struct](btf.UnderlyingType(m.Type))
+		if ok && st.Name == innerName {
+			inner = st
+			innerOff = int(m.Offset / 8)
```
lmb:
```suggestion
			innerOff = m.Offset.Bytes()
```

collection.go
lmb: Nit: probably cleaner to call this `toMembers` or something like that.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+	}
+
+	for _, m := range from.Members {
+		if m.BitfieldSize > 0 {
+			return nil, fmt.Errorf("bitfield %s not supported", m.Name)
+		}
+
+		kernMember, ok := kernIndexByName[m.Name]
+		if !ok {
+			continue
```
lmb, line 760: Is this what libbpf does as well? Seems a bit prone to breakage.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		}
+
+		if kernMember.BitfieldSize > 0 {
```
lmb: Move this to where kernIndexByName is built?

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve size of %s: %w", m.Name, err)
+		}
+
+		kernSz, err := btf.Sizeof(kernMember.Type)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve size of %s: %w", kernMember.Name, err)
+		}
+
+		if sz != kernSz {
```
lmb: This check is pretty basic. Is that all that libbpf does as well?

Might make sense to call `btf.CheckTypeCompatibility` if libbpf does something similar.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+			return nil, fmt.Errorf("size mismatch for %s: %d != %d", m.Name, sz, kernSz)
+		}
+
+		src := int(m.Offset / 8)
```
lmb: .Bytes()

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		dst := innerOff + int(kernMember.Offset/8)
+		copy(toData[dst:dst+int(sz)], fromData[src:src+int(sz)])
```
lmb: Please bounds check src and dst sizes.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+
+	for _, m := range vType.Members {
+		st, ok := btf.As[*btf.Struct](btf.UnderlyingType(m.Type))
+		if ok && st.Name == innerName {
+			inner = st
+			innerOff = int(m.Offset / 8)
+		}
+	}
+
+	if inner == nil {
```
lmb: See above, this should be calculated once, in the caller.

collection.go
lmb: Otherwise, I think it might make sense to restrict this to work on `*btf.Int` only for now.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		if _, isPtr := btf.As[*btf.Pointer](m.Type); !isPtr {
+			continue
+		}
+
+		p, ok := programs[m.Name]
+		if !ok || p == nil {
+			continue
+		}
+
+		off := innerOff + int(m.Offset/8)
```
lmb: .BYtes()

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		src := int(m.Offset / 8)
+		dst := innerOff + int(kernMember.Offset/8)
+		copy(toData[dst:dst+int(sz)], fromData[src:src+int(sz)])
+	}
+
+	return toData, nil
+}
+
+// populateFuncPtr writes progFD into `data` at the offset
+func populateFuncPtr(vType *btf.Struct, data []byte, programs map[string]*Program) error {
```
lmb: Nit: maybe `populateStructOpsPrograms`? More descriptive and I doubt that we'll reuse this logic.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+
+		src := int(m.Offset / 8)
+		dst := innerOff + int(kernMember.Offset/8)
+		copy(toData[dst:dst+int(sz)], fromData[src:src+int(sz)])
+	}
+
+	return toData, nil
+}
+
+// populateFuncPtr writes progFD into `data` at the offset
```
lmb: This doesn't describe accurately what the function does.

collection.go
lmb: This whole function plus setStructOpsProgAttachTo should be deleted.

collection.go
```diff
@@ -697,6 +741,165 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+		member := m.Name
+		ps, ok := cl.coll.Programs[member]
+		if !ok || ps == nil || ps.Type != StructOps {
+			continue
+		}
+
+		// TODO: Do we need to check if the AttachedTo property is already configured?
+		// For example, let it fail if it is already configured,
+		// since it will attach to another struct_ops.
+		ps.AttachTo = fmt.Sprintf("%s:%s", kernType.Name, member)
```
lmb: Again: you cannot modify specs from the loader. Changing AttachTo must happen in the ELF reader. The collection loader must operate on the assumption that AttachTo is in the correct format.

collection.go
```diff
@@ -688,6 +694,44 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+			defer module.Close()
+
+			vType, _ := btf.As[*btf.Struct](target)
+
+			kernVData, err := translateStructData(valueType, userData, vType)
+			if err != nil {
+				return err
+			}
+
+			if err := populateFuncPtr(vType, kernVData, cl.programs); err != nil {
```
lmb: This function is kind of dangerous because it converts fds to raw data. The GC may close the fds prematurely if we aren't careful. Please add a comment to that effect and then add runtime.KeepAlive as suggested.

collection.go
```diff
@@ -688,6 +694,44 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+				return err
+			}
```
lmb, line 698:
```suggestion
			}
			defer runtime.KeepAlive(cl.programs)
```

collection_test.go
lmb: Not actioned.

lmb (commented):
Looks pretty good! There are a couple of comments which you didn't address. Please go through the conversations and resolve them if you've actioned them.

To fix the CI breakage: for now it's ok to set `AttachTo = ""` if were dealing with a struct ops program in TestLibBPFCompat.

lmb (changes requested):
Looks pretty good! There are a couple of comments which you didn't address. Please go through the conversations and resolve them if you've actioned them.

To fix the CI breakage: for now it's ok to set `AttachTo = ""` if were dealing with a struct ops program in TestLibBPFCompat.

lmb:
Please rebase on top of https://github.com/cilium/ebpf/pull/1865. The end result will be that you shouldn't have to call `btf.LoadKernelSpec` explicitly anymore.

shun159:
> Please rebase on top of #1865. The end result will be that you shouldn't have to call `btf.LoadKernelSpec` explicitly anymore.

done

shun159:
@lmb Since the comment disappeared, I’ll respond by quoting below.

```
> +		return nil, fmt.Errorf("member %s not found in %s", innerName, to.Name)
+	}
+
+	kernIndexByName := make(map[string]btf.Member, len(inner.Members))
+	for _, m := range inner.Members {
+		kernIndexByName[m.Name] = m
+	}
+
+	for _, m := range from.Members {
+		if m.BitfieldSize > 0 {
+			return nil, fmt.Errorf("bitfield %s not supported", m.Name)
+		}
+
+		kernMember, ok := kernIndexByName[m.Name]
+		if !ok {
+			continue
Is this what libbpf does as well? Seems a bit prone to breakage.
```

```
> +
+		if kernMember.BitfieldSize > 0 {
+			return nil, fmt.Errorf("bitfield %s not supported in kern struct", kernMember.Name)
+		}
+
+		sz, err := btf.Sizeof(m.Type)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve size of %s: %w", m.Name, err)
+		}
+
+		kernSz, err := btf.Sizeof(kernMember.Type)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve size of %s: %w", kernMember.Name, err)
+		}
+
+		if sz != kernSz {
This check is pretty basic. Is that all that libbpf does as well?

Might make sense to call btf.CheckTypeCompatibility if libbpf does something similar.
```

From my understanding, libbpf seems to handle it as follows:

1. if a field in the “user struct” is not found in the kernel struct and its value is non-zero, it results in an error.
2. explicitly checks for matching type kinds.

It may be need to keep the parsed result of the ELF somewhere.

shun159:
At the moment, we only have a test that writes progFD into kern_vdata, but please let me also add some test to check whether scalar values can be written.

shun159:
> @lmb Since the comment disappeared, I’ll respond by quoting below.
> 
> ```
> > +		return nil, fmt.Errorf("member %s not found in %s", innerName, to.Name)
> +	}
> +
> +	kernIndexByName := make(map[string]btf.Member, len(inner.Members))
> +	for _, m := range inner.Members {
> +		kernIndexByName[m.Name] = m
> +	}
> +
> +	for _, m := range from.Members {
> +		if m.BitfieldSize > 0 {
> +			return nil, fmt.Errorf("bitfield %s not supported", m.Name)
> +		}
> +
> +		kernMember, ok := kernIndexByName[m.Name]
> +		if !ok {
> +			continue
> Is this what libbpf does as well? Seems a bit prone to breakage.
> ```
> 
> ```
> > +
> +		if kernMember.BitfieldSize > 0 {
> +			return nil, fmt.Errorf("bitfield %s not supported in kern struct", kernMember.Name)
> +		}
> +
> +		sz, err := btf.Sizeof(m.Type)
> +		if err != nil {
> +			return nil, fmt.Errorf("failed to resolve size of %s: %w", m.Name, err)
> +		}
> +
> +		kernSz, err := btf.Sizeof(kernMember.Type)
> +		if err != nil {
> +			return nil, fmt.Errorf("failed to resolve size of %s: %w", kernMember.Name, err)
> +		}
> +
> +		if sz != kernSz {
> This check is pretty basic. Is that all that libbpf does as well?
> 
> Might make sense to call btf.CheckTypeCompatibility if libbpf does something similar.
> ```
> 
> From my understanding, libbpf seems to handle it as follows:
> 
> 1. if a field in the “user struct” is not found in the kernel struct and its value is non-zero, it results in an error.
> 2. explicitly checks for matching type kinds.
> 
> It may be need to keep the parsed result of the ELF somewhere.

@lmb 
From the VAR section, we can obtain information about each field of the “user struct” as shown below. By passing such section to CollectionSpec or similar, it seems possible to achieve the same functionality. I’d appreciate your thoughts on this.


(It can be store this in the MapKV of each structops map)

```
[10] VAR 'null_sched_init.____fmt' type_id=9, linkage=static
[11] STRUCT 'sched_ext_ops' size=424 vlen=39
        'select_cpu' type_id=12 bits_offset=0
        'enqueue' type_id=21 bits_offset=64
        'dequeue' type_id=21 bits_offset=128
        'dispatch' type_id=23 bits_offset=192
        'tick' type_id=25 bits_offset=256
        'runnable' type_id=21 bits_offset=320
        'running' type_id=25 bits_offset=384
        'stopping' type_id=27 bits_offset=448
        'quiescent' type_id=21 bits_offset=512
        'yield' type_id=29 bits_offset=576
        'core_sched_before' type_id=29 bits_offset=640
        'set_weight' type_id=31 bits_offset=704
        'set_cpumask' type_id=36 bits_offset=768
        'update_idle' type_id=41 bits_offset=832
        'cpu_acquire' type_id=43 bits_offset=896
        'cpu_release' type_id=47 bits_offset=960
        'init_task' type_id=51 bits_offset=1024
        'exit_task' type_id=55 bits_offset=1088
        'enable' type_id=25 bits_offset=1152
        'disable' type_id=25 bits_offset=1216
        'dump' type_id=59 bits_offset=1280
        'dump_cpu' type_id=63 bits_offset=1344
        'dump_task' type_id=65 bits_offset=1408
        'cgroup_init' type_id=67 bits_offset=1472
        'cgroup_exit' type_id=73 bits_offset=1536
        'cgroup_prep_move' type_id=75 bits_offset=1600
        'cgroup_move' type_id=77 bits_offset=1664
        'cgroup_cancel_move' type_id=77 bits_offset=1728
        'cgroup_set_weight' type_id=79 bits_offset=1792
        'cpu_online' type_id=81 bits_offset=1856
        'cpu_offline' type_id=81 bits_offset=1920
        'init' type_id=83 bits_offset=1984
        'exit' type_id=85 bits_offset=2048
        'dispatch_max_batch' type_id=33 bits_offset=2112
        'flags' type_id=18 bits_offset=2176
        'timeout_ms' type_id=33 bits_offset=2240
        'exit_dump_len' type_id=33 bits_offset=2272
        'hotplug_seq' type_id=18 bits_offset=2304
        'name' type_id=89 bits_offset=2368
```

lmb:
My idea is to place `STRUCT 'sched_ext_ops'` into `MapSpec.Value`. Would that not work?

shun159:
> My idea is to place `STRUCT 'sched_ext_ops'` into `MapSpec.Value`. Would that not work?

@lmb  Ah! I see, does something like the following match what you have in mind?
(This is an scx example; apologies if any details aren’t exact)

```go
ms := &MapSpec{
    Name:  "scx_null",
    Type:  StructOpsMap,
    Value: &btf.Struct{
        Name: "sched_ext_ops",
        // ... parsed from the ELF reader
        Members: []Member{
            {Name: "init", Type: ..., Offset: ...., }
            .
            .
        },
    },
}
```

I think this would work, however, in practice the map data will be a bytes of the value type, so in this change I used the value type name for Value. In the example above we’d be putting the user struct there, which I think could be a bit confusing.

shun159:
From  [bpf_map__init_kern_struct_ops](https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L1135), it appears that libbpf uses the BTF generated from the ELF for type validation and for offsets when copying, etc. For example, for type-size validation, it seems to read and check against the BTF corresponding to what loadSpecFromELF(file) would produce as btf.Spec. 

Therefore, I think we need a way to pass the BTF generated from the ELF into NewCollectionWithOptions(). It’s probably not sufficient to pass struct members. 
I’d appreciate your thoughts.

shun159:
> My idea is to place `STRUCT 'sched_ext_ops'` into `MapSpec.Value`. Would that not work?

@lmb done. it works.

shun159:
@lmb Please review it when you have time.

shun159:
> From [bpf_map__init_kern_struct_ops](https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L1135), it appears that libbpf uses the BTF generated from the ELF for type validation and for offsets when copying, etc. For example, for type-size validation, it seems to read and check against the BTF corresponding to what loadSpecFromELF(file) would produce as btf.Spec.
> 
> Therefore, I think we need a way to pass the BTF generated from the ELF into NewCollectionWithOptions(). It’s probably not sufficient to pass struct members. I’d appreciate your thoughts.

Correction: this is not needed. copying userStruct into ms.Value will be sufficient.

prog.go
```diff
@@ -381,15 +383,48 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
+		}
+
+		module, targetID, err := findProgramTargetInKernel(attachTo, spec.Type, spec.AttachType, c)
 		if err != nil && !errors.Is(err, errUnrecognizedAttachType) {
 			// We ignore errUnrecognizedAttachType since AttachTo may be non-empty
 			// for programs that don't attach anywhere.
 			return nil, fmt.Errorf("attach %s/%s: %w", spec.Type, spec.AttachType, err)
 		}
 
+		if spec.Type == StructOps && targetMember != "" {
```
lmb: Why check that targetMember != ""? I'd prefer this to be an error instead.

map.go
```diff
@@ -560,6 +560,41 @@ func (spec *MapSpec) createMap(inner *sys.FD) (_ *Map, err error) {
...
+			attr.BtfVmlinuxValueTypeId = targetID
+
+			if module != nil {
+				// BPF_F_VTYPE_BTF_OBJ_FD is required if the type comes from a module
+				attr.MapFlags |= sys.BPF_F_VTYPE_BTF_OBJ_FD
+				// set FD for the kernel module
+				attr.ValueTypeBtfObjFd = int32(module.FD())
+			}
+
+			// TODO: This is required to make the map load, but doesn't make
```
lmb: @shun159 this needs investigating please

lmb:
CI failures are probably because the PR hasn't been rebased, not exactly sure.

lmb:
Squashed and rebased.

map.go
shun159: I've confirmed the code comment added by @lmb .

lmb (approved):

lmb:
I need to rebase this once https://github.com/cilium/ebpf/pull/1879 is in so that I can test that it doesn't break windows.

collection.go
```diff
@@ -697,6 +707,208 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+				}
+			}
+			continue
+		}
+
+		km := kType.Members[i]
+
+		switch btf.UnderlyingType(m.Type).(type) {
+		case *btf.Struct, *btf.Union:
+			return fmt.Errorf("nested struct or union not supported")
```
dylandreimerink, line 767: We should not error here, just ignore it.

There are struct ops which have nested structs such as [`struct tcp_congestion_ops`](https://elixir.bootlin.com/linux/v6.17.1/source/include/net/tcp.h#L1262). When working with a definition copied from sources or vmlinux.h, you hit this.

prog.go
```diff
@@ -381,15 +383,51 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
-		module, targetID, err := findProgramTargetInKernel(spec.AttachTo, spec.Type, spec.AttachType, c)
+		var targetMember string
+		attachTo := spec.AttachTo
+
+		if spec.Type == StructOps {
+			attachTo, targetMember, _ = strings.Cut(attachTo, ":")
+			if targetMember == "" {
+				return nil, fmt.Errorf("struct_ops: AttachTo must be '<ops>:<member>' (got %s)", spec.AttachTo)
+			}
+		}
```
dylandreimerink, line 394: This is not consistent with the standard that libbpf sets out. In this implementation a user is required to make a section like so:
```
SEC("struct_ops/tcp_congestion_ops:undo_cwnd")
```

But the libbpf standard is that the name after the `struct_ops/` does not matter. Instead, the structure and member are based on how a function is used in the map value:

```
SEC("struct_ops/write_sk_pacing_undo_cwnd")
__u32 BPF_PROG(write_sk_pacing_undo_cwnd, struct sock *sk)
{
	return tcp_sk(sk)->snd_cwnd;
}

SEC(".struct_ops")
struct tcp_congestion_ops write_sk_pacing = {
	.undo_cwnd = (void *)write_sk_pacing_undo_cwnd,
	.name = "bpf_w_sk_pacing",
};
```

So we to should use the map value BTF to figure out this info so that we are ELF compatible with libbpf.

collection.go
```diff
@@ -697,6 +707,208 @@ func (cl *collectionLoader) populateDeferredMaps() error {
...
+	}
+
+	// Populate the map explicitly and keep a reference on cl.programs.
+	// This is necessary because we may inline fds into kernVData which
+	// may become invalid if the GC frees them.
+	if err := m.Put(uint32(0), kernVData); err != nil {
+		return err
+	}
+	mapSpec.Contents = nil
+	runtime.KeepAlive(cl.programs)
```
dylandreimerink, line 799: This creates really unintuitive/unexpected behavior. The act of writing a value to a struct_ops_map that has not been created with the `F_LINK` flag attaches it. So the moment you run `ebpf.NewCollection(spec)`, you have attached.

Even worse, when attached the refcount of the map is incremented. When you call `coll.Close()` the map value is not deleted, and thus the struct ops programs currently stay attached.

That is not expected by users, since for all other program types you need to explicitly attach using the link package.

I do not think this is necessarily a bad approach, but only if we **always** load the struct ops map with `F_LINK`. Loading with `F_LINK` makes it so writing to the map does not do anything, and attaching is managed via an actual BPF link.

Caveats are that `F_LINK` was added to the kernel later on, so we would limit the minimum kernel version requirements. And also normally this is user controllable by specifying `SEC("struct_ops")` vs `SEC("struct_ops.link")`.

collection.go
lmb: I added this in the hope that we could get away with not dealing with this. The correct fix is to make this function recursive.

prog.go
lmb: - Discussed this with @shun159 before: the ELF reader will fudge up `AttachTo` to conform to the format we expect based on the struct definition. We'll throw away the section information.
- The map value BTF is not enough to capture this, the information is only in relocation entries.

The design constraint here is that we need to know the struct ops type name and member name when loading the program (sigh), so this info **has** to be in ProgramSpec.

collection.go
lmb: Great catch! I think for now we should force BPF_F_LINK by only allowing the ELF reader to consume .link sections. Tests should also only do F_LINK. I don't see how to make the non-link semantics work, they are just too broken / special. Maybe people need to "manually" create such maps if they need further backwards compat.

collection.go
lmb: IMO we can fix this in a follow up PR.

collection.go
lmb: Unit tests already use BPF_F_LINK, so no change needed here.

prog.go
dylandreimerink: Ack. I guess those changes are just not https://github.com/cilium/ebpf/pull/1869 yet which is what I was testing with. It currently just passes the section straight through. If we do want to keep this as two separate PRs we should ensure that gets fixed there.

lmb: @shun159 please update the ELF reader PR based on this.

collection.go
dylandreimerink: > I think for now we should force BPF_F_LINK by only allowing the ELF reader to consume .link sections. Tests should also only do F_LINK. I don't see how to make the non-link semantics work, they are just too broken / special. Maybe people need to "manually" create such maps if they need further backwards compat.

Yes, agreed. Lets throw an error if we do not see `BPF_F_LINK` when loading the map. Worst case people will need the add the flag to the spec manually in Go.

If we do go that route, we are missing the logic to create the link. We can add that in a followup PR. So this PR adds the logic to load a spec. https://github.com/cilium/ebpf/pull/1869 has the logic to go from ELF to spec. And some future PR will add the link logic to attach the loaded program/map. Sound good?

lmb: Yup that is the plan.

prog.go
shun159: @lmb ack.

shun159:
thx @lmb and @dylandreimerink for your kind support!

PR #1849, opened 2025-08-15
ringbuf: Windows support

lmb:


lmb:
@florianl please let me know if you want to review this!

ringbuf/reader_windows.go
```diff
@@ -0,0 +1,105 @@
...
+	"os"
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/sys/windows"
+
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/efw"
+)
+
```
florianl, line 15: To guarantee compatibility between windows and others, maybe we should have an interface for `poller` and make sure the respective implementations are compatible.

ringbuf/helper_windows_test.go
```diff
@@ -0,0 +1,75 @@
+package ringbuf
```
florianl, line 1: For consistency, maybe add a build constraint
```suggestion
//go:build windows

package ringbuf
```

ringbuf/reader_test.go
```diff
@@ -279,8 +182,8 @@ func TestReaderNoWakeup(t *testing.T) {
 
 	prog, events := mustOutputSamplesProg(t,
 		sampleMessage{size: 5, flags: sys.BPF_RB_NO_WAKEUP}, // Read after timeout
-		sampleMessage{size: 6, flags: sys.BPF_RB_NO_WAKEUP}, // Discard
```
florianl, line 282: I'm wondering, where the discard case was dropped.

ringbuf/reader_windows.go
```diff
@@ -0,0 +1,105 @@
+package ringbuf
```
florianl, line 1: For consistency, maybe add a build constraint

```suggestion
//go:build windows

package ringbuf
```

ringbuf/ring_other.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"runtime"
+	"unsafe"
+
+	"github.com/cilium/ebpf/internal/unix"
+)
+
```
florianl, line 14: Similar to poller, maybe there should also be an interface for `ringbufEventRing` so we can make sure all implementations fit here.

ringbuf/helper_windows_test.go
lmb: That is redundant with the file name, no?

ringbuf/reader_test.go
lmb: Discard is currently unsupported on windows, so the helper issues a skip. It’s important to test the wake-up behaviour on windows, and the discard here doesn’t change the observed behaviour at all. So dropping it is the easiest solution.

lmb:
Thanks! Made an issue for the interface changes, since I don’t have access to a laptop at the moment.

PR #1850, opened 2025-08-15
go 1.24

lmb:
With the release of Go 1.25 we will require 1.24 as the minimum version.


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/cilium/ebpf
 
-go 1.23.0
-
-toolchain go1.24.1
+go 1.24
```
lmb, line 3: This is what the toolchain now does when doing `go mod edit -go=1.24` (modulo the toolchain line). I hope this spells the end of the go version mess.

florianl (dismissed):

PR #1851, opened 2025-08-15
use runtime.AddCleanup

lmb:


internal/epoll/poller.go
```diff
@@ -75,7 +76,10 @@ func New() (_ *Poller, err error) {
 		return nil, fmt.Errorf("add flush eventfd: %w", err)
 	}
 
-	runtime.SetFinalizer(p, (*Poller).Close)
+	p.cleanup = runtime.AddCleanup(p, func(epollFd int) {
+		_ = unix.Close(epollFd)
```
ti-mo: Wouldn't this omit a bunch of other behaviours? `Poller.Close()` wakes wait, takes out two mutexes and closes two channels. It looks like we need to take a different approach here. I also liked the fact that the finalizer and `Close()` were the same thing. By passing ad-hoc closures to AddCleanup, we're inviting bugs due to updating one but not the other.

One argument in its favor, though: technically, the cleanup only runs if *Poller is unreachable, meaning there can't be any more callers to Wait() or Add(), so the synchronization behaviour should no longer be needed at that point. We just need to make sure to never return `p.closeEvent` or `p.flushEvent` to the caller. We'll also rely on the cleanups of the eventFd's underlying os.File, since we're no longer calling eventFd.close() explicitly when the Poller's cleanup runs.

@brycekahle Curious to hear your take on this.

PR #1855, opened 2025-08-20
link: allow '-' in tracepoint group names

### LINKED ISSUE #1854: link: link.Tracepoint fails for tracepoint groups containing '-' (hyphen)
### Describe the bug

When trying to attach an eBPF program to a tracepoint whose group contains a hyphen (e.g. `xhci-hcd`), `link.Tracepoint` fails with:
```
opening tracepoint: invalid tracefs group: "xhci-hcd"
```
This happens because the `validIdentifier` function currently rejects hyphens (-), even though they are valid in tracefs event groups (for example, USB tracepoints use `xhci-hcd`).
This prevents attaching to valid tracepoints that exist in the kernel.

### Expected behavior
The tracepoint should attach successfully, since `xhci-hcd` is a valid group under `/sys/kernel/tracing/events/`.

### How to reproduce

Build a simple Go program that tries to attach to a tracepoint with a `-` in the group:
```
tp, err := link.Tracepoint("xhci-hcd", "xhci_setup_device", objs.XhciSetupDevice, nil)
if err != nil {
	log.Fatalf("opening tracepoint: %s", err)
}
defer tp.Close()
```
and run the program.

### Version information

github.com/cilium/ebpf v0.19.0

AryanBakliwal:
This PR fixes an issue where attaching eBPF programs to tracepoints with group names containing `-` (e.g., `xhci-hcd`) would fail with:
```
opening tracepoint: invalid tracefs group: "xhci-hcd"
```
The root cause is that the `validIdentifier` function in `tracefs` rejected `-` characters, even though they are valid in tracepoint group names.

Fixes #1854


internal/tracefs/kprobe.go
```diff
@@ -86,6 +86,7 @@ func validIdentifier(s string) bool {
 		case c >= 'a' && c <= 'z':
 		case c >= 'A' && c <= 'Z':
 		case c == '_':
+		case c == '-':
```
mmat11: If the group name can start with a dash, I think the regex up in the comment should be `"^[a-zA-Z_-][0-9a-zA-Z_-]*$"`, else we should check the index like we do for numbers

mmat11:
I think this test will fail https://github.com/cilium/ebpf/blob/main/internal/tracefs/perf_event_test.go#L34

internal/tracefs/kprobe.go
AryanBakliwal: Updated to check index for dash, same as numbers

mmat11 (dismissed):
lgtm

Issue #669, opened 2022-05-10
Add support for MapExtra and bloom filter maps


ti-mo:
Kernel 5.16 saw the introduction of the bloom filter map type: https://lore.kernel.org/all/20211027234504.30744-1-joannekoong@fb.com/.

This also included new `map_extra` field for maps that need more configuration than the existing map properties.

In https://github.com/cilium/ebpf/pull/668, selftests relying on MapExtra are ignored for now.

Maybe we should not expose the field from MapSpec immediately, until we figure out how to expose it as a user-friendly API.

PR #1856, opened 2025-08-20
map: add MapSpec.MapExtra field

NoNickeD:
## Summary

This PR adds support for the MapExtra field and bloom filter maps, addressing issue #669.

## Changes Made

- Add MapExtra field to MapSpec struct for map-specific configuration
- Update map creation to pass MapExtra to kernel via sys.MapCreateAttr
- Handle map_extra field in BTF map definitions (elf_reader.go)
- Add NewBloomFilter helper function for creating bloom filter maps
- Add tests for bloom filter functionality

## Notes

- Bloom filters available from Linux 5.16+
- Uses lower 4 bits of MapExtra to specify 1-15 hash functions (default: 5)
- Backward compatible - existing code unaffected


map.go
```diff
@@ -1809,3 +1816,26 @@ func sliceLen(slice any) (int, error) {
...
+
+// NewBloomFilter creates a new bloom filter map.
+//
+// valueSize is the size of the values to be hashed.
+// maxEntries is used to approximate the size of the bloom filter bitmap.
+// numHashes specifies the number of hash functions to use (1-15).
+// If numHashes is 0, the default of 5 hash functions will be used.
+//
+// The bloom filter map type is available from Linux 5.16.
+func NewBloomFilter(name string, valueSize, maxEntries uint32, numHashes uint8) (*MapSpec, error) {
```
dylandreimerink: We do not add constructors for specific map types in cilium/ebpf, its additional API surface area which is frankly not required since all fields on MapSpec are public.

Please do away with this bit.

dylandreimerink (changes requested):
Thanks for the PR! However, I have some changes to request.

I think the change to parse the MapExtra from ELF makes sense. The addition of the constructor does not.

All of the test logic added tests this constructor. I suggest you remove it, and instead extend `TestLoadCollectionSpec` to test parsing of map extra from ELF. You would do so my adding a new map definition to `testdata/loader.c` and then adding a new expected MapSpec.

NoNickeD:
Hello @dylandreimerink,

Thanks for the feedback! I have removed the NewBloomFilter constructor and all associated tests as requested. The MapExtra parsing from ELF was already implemented in the existing code, so bloom filters can still be created by setting the MapSpec fields directly. 

The changes have been pushed and the PR is ready for another review.

dylandreimerink:
> The MapExtra parsing from ELF was already implemented in the existing code

Yea, passing from MapSpec was already there, but this addition is needed to make ELF -> MapSpec work properly.

Thank you for the changes. Please don't forget about extending the tests to cover this new behavior https://github.com/cilium/ebpf/blob/main/elf_reader_test.go#L48 + https://github.com/cilium/ebpf/blob/main/testdata/loader.c

That way we can assert this stays working if we ever have to make modifications in the future.

ti-mo:
@NoNickeD Thanks for picking this up! I ended up moving things around a bit and added separate testdata for arena maps since loader.c needs to work on all kernels in the matrix. I'm planning on doing more arena-related work, so it made sense long-term.

Arenas also need the `__ulong` macro for map_extra since the kernel's field is a 64-bit value, added support for that as well. Bloom filters should work just fine, though. Didn't end up needing extra testing there.

NoNickeD:
@ti-mo Thanks for merging and improving this! The separate arena testdata and __ulong macro support for 64-bit map_extra make sense. Learned a lot from your restructuring.

Thanks for the guidance!

PR #1858, opened 2025-08-27
info: Return dedicated error when program info is restricted

dylandreimerink:
On systems where the `kernel.kptr_restrict` and
`net.core.bpf_jit_harden` sysctls are enabled, certain fields in the program info may be restricted. When this is the case xlated and jitted instruction, line info, and function info are unavailable.

When such fields are unavailable, we do get ..len fields set to non-zero values, but the kernel will have not written any data to the pointers provided.

At present when this happens we don't recognize this and attempt to parse the empty buffers resulting in a `parse func info: offset 0: type ID 0 is a *btf.Void, but expected a Func` error.

The only clue we have that this is happening is that the kernel will zero out the insns pointer. This commit adds logic to detect when this happens and then returns a dedicated `ErrInfoRestricted` error so this case can be handled gracefully.


florianl (approved):

info.go
```diff
@@ -556,6 +566,10 @@ func (pi *ProgramInfo) btfSpec() (*btf.Spec, error) {
 	return spec, nil
 }
 
+// ErrInfoRestricted when kernel.kptr_restrict and net.core.bpf_jit_harden are set, the kernel may restrict access
+// to certain bits of the program info.
+var ErrInfoRestricted = errors.New("program info restricted by combination of kernel.kptr_restrict and net.core.bpf_jit_harden")
```
ti-mo: Gut feel: we may need this in more places down the line, and we should do a bit of exploration in the kernel source to figure out which other APIs are affected by kptr_restrict and bpf_jit_harden. (kallsyms comes to mind)

I like the name of the sentinel, but the string is slightly too specific IMO; the context around program info being restricted should be wrapped into the error string by the methods on ProgramInfo.

dylandreimerink: > kallsyms comes to mind

Yes, correct. When reading kallsyms all addresses are 0x00000000, so features like ksyms will not work. This happens based on `kptr_restrict` settings.

The `kptr_restrict` has some effects on the instructions read back, such as addresses of maps, helper functions, ect. Everything that gets converted to a kernel pointer by the verifier.

I will dig a bit further into the effects of `bpf_jit_harden` as well.

> I like the name of the sentinel, but the string is slightly too specific IMO; the context around program info being restricted should be wrapped into the error string by the methods on ProgramInfo.

Yea, fair. I will make the sentinel a bit more generic and wrap it with this more specific error case.

ti-mo: I did some digging.

Looks like `kptr_restrict` is only really used in `lib/vsprintf.c` and `kernel/ksyms_common.c`, both of which are concerned with printing kernel pointers through procfs (and dmesg entries?).

`bpf_jit_harden` is only used to set `prog->blinding_requested`, used in dumping prog insns.

tl;dr: looks like kallsyms and prog insns are the only places we need to account for at the moment.

Issue #1857, opened 2025-08-26
Panic from perf ring buffer reader


damemi:
### Describe the bug

We are seeing a panic in the perf ring buffer reader, after recently bumping our dep to commit `ae226118949d`

```
unexpected fault address 0x7fae3afff000
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7fae3afff000 pc=0x1e86b75]

goroutine 340 gp=0xc0004cb880 m=66 mp=0xc004189808 [running]:
runtime.throw({0x2867897?, 0x0?})
	/usr/local/go/src/runtime/panic.go:1101 +0x48 fp=0xc001f43d48 sp=0xc001f43d18 pc=0x47b728
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:939 +0x26c fp=0xc001f43da8 sp=0xc001f43d48 pc=0x47daec
github.com/cilium/ebpf/ringbuf.(*ringReader).readRecord(0xc00062ac38?, 0x997?)
	/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250818092626-ae226118949d/ringbuf/ring.go:46 +0x15 fp=0xc001f43e18 sp=0xc001f43da8 pc=0x1e86b75
github.com/cilium/ebpf/ringbuf.(*Reader).ReadInto(0xc00062ac30, 0xc001f43f40)
	/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250818092626-ae226118949d/ringbuf/reader.go:167 +0x318 fp=0xc001f43ec0 sp=0xc001f43e18 pc=0x1e86818
github.com/cilium/ebpf/ringbuf.(*Reader).Read(...)
```

Could this possibly be related to https://github.com/cilium/ebpf/pull/1843, which seems to have recently changed those lines?

### How to reproduce

No consistent reproducer at the moment, sorry

### Version information

v0.19.1-0.20250818092626-ae226118949d

PR #1859, opened 2025-08-27
ringbuf: fix read after close panic

RonFed:
Fix the issue describe in #1857.
Following https://github.com/cilium/ebpf/pull/1849, the `Close` function was changed to not set the ring to nil.
The panic happens when `r.haveData` is true, a `Close` happens and then another `Read`.
Added a test that reproduces the panic.


florianl (approved):
Thanks for the quick fix!

PR #1860, opened 2025-09-02
bpf2go: Generate constant names for maps, programs, and variables

mattijons:
Idea https://github.com/cilium/ebpf/discussions/1833.

Add bpf2go functionality that generates constant string names for maps, programs, and variables. Useful when working with `ebpf.Collection` and `ebpf.CollectionSpec`.

Example output from `TestObjects`:
```go
// Names of all BPF objects in the ELF.
//
// Used for safe lookups in a Collection or CollectionSpec.
const (
	barMapMap1      = "map1"
	barProgProgFoo1 = "prog_foo_1"
	barVarVar1      = "var_1"
)
```

### DIFF
--- a/cmd/bpf2go/gen/output.tpl
+++ b/cmd/bpf2go/gen/output.tpl
@@ -22,6 +22,23 @@ import (
 {{ end }}
 {{- end }}
 
+{{- if or .Maps (or .Variables .Programs) }}
+// Names of all BPF objects in the ELF.
+//
+// Used for safe lookups in a Collection or CollectionSpec.
+const (
+{{- range $name, $id := .Maps }}
+	{{ $.Name }}Map{{ $id }} = "{{ $name }}"
+{{- end }}
+{{- range $name, $id := .Programs }}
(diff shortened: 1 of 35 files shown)


lmb:
What is the benefit of turning strings into constants? Some type of compile time error I guess?

My gut feeling is against adding this: usually names for maps and programs don't change that often, and in my experience are used only in a few places.

Maybe you can explain why you want to use bpf2go while still going via collection and collectionspec instead of the generated types?

ti-mo:
> Maybe you can explain why you want to use bpf2go while still going via collection and collectionspec instead of the generated types?

IMO precisely to get the best of both worlds: compile-time codegen and ELF embedding + 'fearless' access of e.g. `spec.Programs[SomeProgram]` without really requiring a nil check, for maximum flexibility. If the prog gets moved or renamed, that code will break at compile time instead of at runtime.

ti-mo:
@mattijons Thanks for working on this, I've removed most of the fluff around the generated code and shortened the intermediate 'prefixes' to Map/Prog/Var.

PR #1869, opened 2025-09-29
elf_reader: add struct_ops support

shun159:
This commit adds struct_ops support to the ELF reader: it classifies non-executable PROGBITS sections, parses their BTF Datasec to build MapSpecs, associates relocs with func-pointer members to set ps.AttachTo, and adds TestStructOps.

Related: #1845


elf_reader.go
```diff
@@ -1379,6 +1415,173 @@ func (ec *elfCode) loadKsymsSection() error {
...
+			return fmt.Errorf("datasec %s: %w", sec.Name, err)
+		}
+
+		dataSec, ok := btf.As[*btf.Datasec](dataType)
+		if !ok {
+			return fmt.Errorf("%s BTF is not a Datasec", sec.Name)
+		}
+
+		for _, vsi := range dataSec.Vars {
+			varType, ok := btf.As[*btf.Var](vsi.Type)
```
lmb: Why mix btf.As and btf.UnderlyingType here?

elf_reader.go
```diff
@@ -1379,6 +1415,173 @@ func (ec *elfCode) loadKsymsSection() error {
...
+			mapName := varType.Name
+
+			userType, ok := btf.UnderlyingType(varType.Type).(*btf.Struct)
+			if !ok {
+				return fmt.Errorf("var %s: expect struct, got %T", varType.Name, varType.Type)
+			}
+
+			// Retrieve raw data from the ELF section.
+			// This data contains the initial values for the struct_ops map.
+			userData, err := sec.Data()
```
lmb: Why do this for each iteration? Can be done once.

elf_reader.go
```diff
@@ -1379,6 +1415,173 @@ func (ec *elfCode) loadKsymsSection() error {
...
+					secIdx:   secIdx,
+					userOff:  userOff,
+					userSize: userSize,
+				}
+
+			ec.maps[mapName] =
+				&MapSpec{
+					Name:       mapName,
+					Type:       StructOpsMap,
+					Key:        &btf.Int{Size: 4},
```
shun159: `KeySize` should be specified as 4.

elf_reader.go
```diff
@@ -116,8 +126,15 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
 		case sec.Type == elf.SHT_REL:
 			// Store relocations under the section index of the target
 			relSections[elf.SectionIndex(sec.Info)] = sec
-		case sec.Type == elf.SHT_PROGBITS && (sec.Flags&elf.SHF_EXECINSTR) != 0 && sec.Size > 0:
-			sections[idx] = newElfSection(sec, programSection)
+		case sec.Type == elf.SHT_PROGBITS:
```
lmb: Keep `sec.Size > 0` in the case?
```suggestion
		case sec.Type == elf.SHT_PROGBITS && sec.Size > 0:
```

elf_reader.go
```diff
@@ -116,8 +126,15 @@ func LoadCollectionSpecFromReader(rd io.ReaderAt) (*CollectionSpec, error) {
...
+			if (sec.Flags&elf.SHF_EXECINSTR) != 0 && sec.Size > 0 {
+				sections[idx] = newElfSection(sec, programSection)
+			} else if sec.Name == ".struct_ops" || sec.Name == ".struct_ops.link" {
```
lmb: Refuse `".struct_ops"` here.

elf_reader.go
```diff
@@ -349,6 +381,10 @@ func (ec *elfCode) loadProgramSections() (map[string]*ProgramSpec, error) {
 			continue
 		}
 
+		if !(sec.Type == elf.SHT_PROGBITS && (sec.Flags&elf.SHF_EXECINSTR) != 0) {
```
lmb: Why is this necessary? if `kind == programSection` then this should always be true.

elf_reader.go
```diff
@@ -1379,6 +1415,174 @@ func (ec *elfCode) loadKsymsSection() error {
...
+// loadStructOpsMapsFromSections creates StructOps MapSpecs from DataSec sections
+// ".struct_ops" and ".struct_ops.link" found in the object BTF.
+func (ec *elfCode) loadStructOpsMaps() error {
+	for secIdx, sec := range ec.sections {
+		if sec.kind != structOpsSection {
+			continue
+		}
+
+		// Process the struct_ops section to create the map
+		dataType, err := ec.btf.AnyTypeByName(sec.Name)
```
lmb: This should use TypeByName.

elf_reader.go
```diff
@@ -1379,6 +1415,174 @@ func (ec *elfCode) loadKsymsSection() error {
...
+					userSize: userSize,
+				}
+
+			ec.maps[mapName] =
+				&MapSpec{
+					Name:       mapName,
+					Type:       StructOpsMap,
+					Key:        &btf.Int{Size: 4},
+					KeySize:    4,
+					Value:      userType,
```
lmb: Maybe also set ValueSize for completeness sake?

elf_reader.go
```diff
@@ -41,20 +41,30 @@ type ksymMeta struct {
 	Name    string
 }
 
+type structOpsSpec struct {
```
lmb: I don't think this is necessary. `elfSection.relocations` already carries this information in the form of an `elf.Symbol`.

elf_reader.go
```diff
@@ -1379,6 +1415,174 @@ func (ec *elfCode) loadKsymsSection() error {
 	return nil
 }
 
+// loadStructOpsMapsFromSections creates StructOps MapSpecs from DataSec sections
+// ".struct_ops" and ".struct_ops.link" found in the object BTF.
+func (ec *elfCode) loadStructOpsMaps() error {
```
lmb: This should be merged with associateStructOpsRelocs unless there is a reason why they need to be separate which I don't understand.

elf_reader.go
```diff
@@ -1379,6 +1415,174 @@ func (ec *elfCode) loadKsymsSection() error {
...
+}
+
+// associateStructOpsRelocs handles `.struct_ops(.link)`
+// and associates the target function with the correct struct member in the map.
+func (ec *elfCode) associateStructOpsRelocs(
+	progs map[string]*ProgramSpec,
+	relSecs map[elf.SectionIndex]*elf.Section,
+	symbols []elf.Symbol,
+) error {
+	for _, sec := range relSecs {
```
lmb: This just duplicates the loop in loadRelocations? You can iterate sec.relocations instead.

lmb (changes requested):
Please refuse loading bare `.struct_ops` and re-enable the `LibBPFCompat` test of StructOps.

elf_reader.go
dylandreimerink: Would it make sense to still allow `.struct_ops` parsing to Spec, but then not allow loading unless you manually add `BPF_F_LINK` to the map spec?

If you block it in ELF parsing, then you force users to modify the section, and take away the ability to patch this in userspace.

lmb: The reason for not using BPF_F_LINK is to get compat with older kernels, no? Adding BPF_F_LINK manually then kind of defeats the purpose?

I agree that it's probably nicer to refuse loading a StructOpsMap without BPF_F_LINK instead, then we can test more of the libbpf ELFs.

dylandreimerink: > The reason for not using BPF_F_LINK is to get compat with older kernels, no? Adding BPF_F_LINK manually then kind of defeats the purpose?

Yea, `BPF_F_LINK` got added in 6.4, and struct ops originally got added in 5.6. In the end, its all about the attachment, there is not difference in the how you write the programs. Its only really tcp_congestion_ops that existed before `BPF_F_LINK` so its  examples and selftests use `.struct_ops`. Schedext and HIDBPF both have macros for defining the struct which do use `.struct_ops.link`, since those features came after 6.4.

shun159: ack. let's refuse `".struct_ops"` here.

elf_reader.go
shun159: @lmb At this point, the size of the value type is not yet determined, so my understanding is that it’s difficult to specify this field for now.

shun159:
> re-enable the LibBPFCompat test of StructOps

@lmb @dylandreimerink 

It seems that in the test cases for programs that include "autocreate", the program fails when load because it cannot determine which map it belongs to.
I believe at this point such cases should be treated as `NotSupoorted`?

test case: https://github.com/cilium/ebpf/actions/runs/18612775255/job/53073227432?pr=1869#step:8:1224
test data : https://elixir.bootlin.com/linux/v6.17.1/source/tools/testing/selftests/bpf/progs/struct_ops_autocreate2.c#L25

elf_reader.go
shun159: outdated, but I will take a look this.

elf_reader.go
```diff
@@ -1379,6 +1402,141 @@ func (ec *elfCode) loadKsymsSection() error {
 	return nil
 }
 
+// associateStructOpsRelocs handles `.struct_ops.link`
```
shun159, line 1397: TODO: need to be refactored.

lmb:
> I believe at this point such cases should be treated as `NotSupoorted`?

That failure just tells me that AttachTo is not populated. Can you explain why that is?

elf_reader.go
lmb: Size should be `btf.Sizeof(userType)` no?

shun159: @lmb this is the size of the section data of the user-side struct, which is different from the size of the kernel-side value type.

shun159:
> That failure just tells me that AttachTo is not populated. Can you explain why that is?

@lmb when given the code below, (for now) we cannot derive which struct_ops map the standalone program `foo` belongs to, so handling `AttachTo` for foo fails in this case.

```c
#include "common.h"

char _license[] __section("license") = "GPL";

struct bpf_testmod_ops {
	int (*test_1)(void);
	int data;
};

__section("?struct_ops/test_1") int foo(void) {
	return 0;
}

__section("?struct_ops/test_1") int bar(void) {
	return 0;
}

__section(".struct_ops.link") struct bpf_testmod_ops testmod_ops = {
	.test_1 = (void *)bar,
};
```

lmb:
> (for now) we cannot derive which struct_ops map the standalone program `foo` belongs to, so handling `AttachTo` for foo fails in this case.

How will we derive that in the future then? IMO the correct semantics here is to set `AttachTo == ""` in the ELF reader if a structops program is not referenced by a structops section.

shun159:
rebased to `main`

elf_reader.go
```diff
@@ -1379,6 +1402,135 @@ func (ec *elfCode) loadKsymsSection() error {
...
+			userSize uint64
+		}
+
+		ofsSizes := make(map[string]structOpsMapOfsSize)
+
+		for _, vsi := range ds.Vars {
+			varType := btf.UnderlyingType(vsi.Type).(*btf.Var)
+			mapName := varType.Name
+
+			userType, ok := btf.UnderlyingType(varType.Type).(*btf.Struct)
```
lmb: Nit: `btf.As`

elf_reader.go
```diff
@@ -1379,6 +1402,135 @@ func (ec *elfCode) loadKsymsSection() error {
...
+				return fmt.Errorf("%s exceeds section", mapName)
+			}
+
+			// Register the MapSpec for this struct_ops instance.
+			ec.maps[mapName] = &MapSpec{
+				Name:       mapName,
+				Type:       StructOpsMap,
+				Key:        &btf.Int{Size: 4},
+				KeySize:    4,
+				Value:      userType,
```
lmb: Why not `ValueSize`? We know `len(userData[...])`

elf_reader.go
```diff
@@ -1379,6 +1402,135 @@ func (ec *elfCode) loadKsymsSection() error {
...
+			}
+
+			userSize := uint64(userType.Size)
+			userOff := uint64(vsi.Offset)
+			if userOff+userSize > uint64(len(userData)) {
+				return fmt.Errorf("%s exceeds section", mapName)
+			}
+
+			// Register the MapSpec for this struct_ops instance.
+			ec.maps[mapName] = &MapSpec{
```
lmb: This should check `ec.maps` to ensure we aren't overriding an existing mapSpec.

elf_reader.go
```diff
@@ -1379,6 +1402,135 @@ func (ec *elfCode) loadKsymsSection() error {
...
+						Value: append([]byte(nil), userData[userOff:userOff+userSize]...),
+					},
+				},
+			}
+			ofsSizes[mapName] =
+				structOpsMapOfsSize{userOff, userSize}
+		}
+
+		// Process relo sections that target this struct_ops section.
+		for relSecIdx, relSec := range relSecs {
```
lmb: Why duplicate work and not use sec.relocations? You should be able to iterate userType.Members and find the relocation: `sec.relocations[userOff+member.Offset]`.

struct_ops.go
```diff
@@ -137,3 +143,20 @@ func structOpsIsMemZeroed(data []byte) bool {
...
 }
+
+// funcPtrMemberAtOffset returns the member name at bit offset `moff`
+// if the member is a pointer to a FuncProto. Otherwise returns an empty string.
+func structOpsFuncPtrMemberAtOffset(userSt *btf.Struct, moff btf.Bits) (string, bool) {
+	for _, m := range userSt.Members {
+		if m.Offset != moff {
+			continue
+		}
+		mt := btf.UnderlyingType(m.Type)
```
lmb: @shun159: this keeps cropping up. `UnderlyingType` and `As` do the same thing, you should not need to use both of them. I've raised this multiple times in several separate PRs. Why do you keep writing the same code? Is something not clear?

elf_sections.go
```diff
@@ -104,8 +104,8 @@ var elfSectionDefs = []libbpfElfSectionDef{
 	{"cgroup/getsockopt", sys.BPF_PROG_TYPE_CGROUP_SOCKOPT, sys.BPF_CGROUP_GETSOCKOPT, _SEC_ATTACHABLE},
 	{"cgroup/setsockopt", sys.BPF_PROG_TYPE_CGROUP_SOCKOPT, sys.BPF_CGROUP_SETSOCKOPT, _SEC_ATTACHABLE},
 	{"cgroup/dev", sys.BPF_PROG_TYPE_CGROUP_DEVICE, sys.BPF_CGROUP_DEVICE, _SEC_ATTACHABLE_OPT},
-	{"struct_ops+", sys.BPF_PROG_TYPE_STRUCT_OPS, 0, _SEC_NONE},
```
lmb, line 107: This file is meant to be auto generated. This looks like a manual change.

I suspect you use this to clear the setting of `AttachTo`.  Either change `init()` in `elf_reader.go` to add this flag after the fact or change gensections.awk to include this flag.

elf_reader.go
```diff
@@ -1379,6 +1402,135 @@ func (ec *elfCode) loadKsymsSection() error {
...
 }
 
+// associateStructOpsRelocs handles `.struct_ops.link`
+// and associates the target function with the correct struct member in the map.
+func (ec *elfCode) associateStructOpsRelocs(
+	progs map[string]*ProgramSpec,
+	relSecs map[elf.SectionIndex]*elf.Section,
+	symbols []elf.Symbol,
+) error {
+	willAttachToMap := make(map[string]bool)
```
lmb: Seems unused besides setting a value to true?

elf_reader.go
```diff
@@ -1379,6 +1402,135 @@ func (ec *elfCode) loadKsymsSection() error {
...
+		if sec.Name == structOpsLinkSec {
+			flags = sys.BPF_F_LINK
+		}
+
+		type structOpsMapOfsSize struct {
+			userOff  uint64
+			userSize uint64
+		}
+
+		ofsSizes := make(map[string]structOpsMapOfsSize)
```
lmb: Gut feeling: you won't need this after switching to sec.relocations.

elf_reader.go
lmb: @shun159 still TODO.

shun159: @lmb yes, right. it is unnecessary guard.

elf_reader.go
shun159: @lmb yes, this is no longer used.

elf_reader.go
shun159: @lmb let me take a look later

elf_reader.go
shun159: yes, we can use it as the ValueSize

elf_reader.go
shun159: let me take a look later.

struct_ops.go
shun159: sorry, this is the old code. I understand that.

elf_reader.go
shun159: (TODO: I will address soon)

elf_reader.go
shun159: (TODO: I will address soon)

elf_reader.go
shun159: guard added. for now, if there already is the mapName in ec.maps, let it fail.

elf_sections.go
shun159: got it. I will take a look this one.

elf_reader.go
shun159: this was addressed

elf_reader.go
shun159: we specified the length of the user-type as the value type now

elf_reader.go
shun159: Merged

elf_sections.go
shun159: fixed `init()` to add the flag for the section.

elf_reader.go
shun159: refactored to stop scanning all relocs and ofs ranges.
instead, simply we can use member's offset to solve the reloc

struct_ops.go
shun159: removed. we don't need this for now.

elf_reader.go
shun159: removed. we don't need this for now.

elf_reader.go
shun159: refactored

elf_reader.go
shun159: we can use sec.relocations. fixed now.

elf_reader.go
shun159: right. this was removed.

elf_reader.go
shun159: fixed

elf_reader.go
shun159: fixed

elf_reader.go
shun159: removed such code now.

shun159:
@lmb could you please review at your leisure?

elf_reader.go
```diff
@@ -1419,6 +1525,8 @@ func init() {
 		// This has been in the library since the beginning of time. Not sure
 		// where it came from.
 		{"seccomp", sys.BPF_PROG_TYPE_SOCKET_FILTER, 0, _SEC_NONE},
+		{"struct_ops+", sys.BPF_PROG_TYPE_STRUCT_OPS, 0, _SEC_NONE | ignoreExtra},
```
lmb, line 1523: Took me a bit to figure out why this works. Lets add a comment?
```suggestion
		// Override libbpf definition because we want ignoreExtra.
		{"struct_ops+", sys.BPF_PROG_TYPE_STRUCT_OPS, 0, _SEC_NONE | ignoreExtra},
```

elf_reader.go
```diff
@@ -1379,6 +1394,97 @@ func (ec *elfCode) loadKsymsSection() error {
 	return nil
 }
 
+// associateStructOpsRelocs handles `.struct_ops.link`
+// and associates the target function with the correct struct member in the map.
+func (ec *elfCode) associateStructOpsRelocs(
+	progs map[string]*ProgramSpec,
```
lmb: Nit: we don't usually use this style for function argument lists. Would you mind just putting the args inline?

elf_reader.go
```diff
@@ -1379,6 +1394,97 @@ func (ec *elfCode) loadKsymsSection() error {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// createStructOpsMap() creates and registers a MapSpec for a struct_ops
+func (ec *elfCode) createStructOpsMap(
+	vsi btf.VarSecinfo,
```
lmb: Nit: see above.

lmb (commented):
Some nits, I think this is good to go. We're going to cut a release and merge this afterwards so that we get some more time to iterate.

elf_reader.go
shun159: fixed

elf_reader.go
shun159: fixed

shun159:
@lmb I've addressed your feedback! :smile:

shun159:
@lmb thx for merging this, and #1844 is the last PR for StructOps feature.

PR #1875, opened 2025-10-13
*: use `go tool` instead of `go run` in `go:generate` statements

ti-mo:
A change to `stringer` broke main, so it's time to use `go tool` for all go:generate statements. Use it for internal tools too, and suggest it as a best practice in documentation and help text.


dylandreimerink:
Changes look good. But, I guess we need to figure out a way to let staticcheck ignore generated files :roll_eyes:

ti-mo:
@lmb Could use some input here on why the Windows tests could be failing. x/sys got an upgrade in this PR, could be the culprit?

PR #1878, opened 2025-10-14
map: add the ability to pin maps in different paths

aureliar8:
This PR allows to pin maps of a program in different directories. 

This is useful for the following usecase:
- I want a CNI-like structure where I attach an ebpf program to different virtual interfaces on a host 
- I want all maps used in this program to be pinned
- I want some maps to be shared across instances of this program and some maps to be isolated so that each program only interact with their own data

Without this PR all map pinning had to happen in the same directory. Now each map can be configured to be pinned in a specific location, which can be in a different directory.

### DIFF
--- a/map.go
+++ b/map.go
@@ -41,7 +41,12 @@ type MapOptions struct {
 	// The base path to pin maps in if requested via PinByName.
 	// Existing maps will be re-used if they are compatible, otherwise an
 	// error is returned.
-	PinPath        string
+	PinPath string
+	// The full path to use for each map with PinByPath
+	// Existing maps will be re-used if they are compatible, otherwise an
+	// error is returned.
+	PinPathMapping map[string]string
+
 	LoadPinOptions LoadPinOptions
 }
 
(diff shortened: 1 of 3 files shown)


ti-mo:
Hi @aureliar8, thanks for the patch! I've also had this particular itch before in Cilium, and this proposal is indeed one way to solve it. However, I don't think this necessarily has its place in the library. We've been adding more and more things to CollectionOptions lately (though on the ProgramOptions side) that are starting to weigh on the loader's complexity.

Also, what you're suggesting here (PinByPath) is not a part of libbpf, which is what the MapSpec's pinning flag is supposed to reflect. (see `enum libbpf_pin_type`) We'll always be hesitant to add our own values and advertise them as public API, since that means we're explicitly no longer compatible with libbpf.

I'd like to suggest a few alternative ways of implementing what you're describing:

1. We've recently added `MapSpec.Tags`, so you could tag your map definition with e.g. `__attribute__((btf_decl_tag("pin_path:some/other/path")))`, prefix it with your bpffs mount point of choice and append the map name. Then, try to open the pin before loading the CollectionSpec and pass the Map in `CollectionOptions.MapReplacements`. If it doesn't exist yet, pin it explicitly afterwards.

1. Slightly simpler, like what we do in Cilium: maps shared across the whole application go in a central location (`tc/globals` for legacy reasons) and use PinByName. All others, like per-endpoint/interface maps are opened/pinned manually since their names and paths are completely dynamic, so they can't really be specified in the ELF anyway, not even with decl tags.

lmb:
Little known fact: a single map can be pinned in multiple locations if that is useful to you. `m.Clone()` followed by `m.Pin()`.

aureliar8:
Okay, thank you for your advice ! I'll use the Cilium approach  and manually open/pin my maps

PR #1879, opened 2025-10-14
windows: fix CI issues

lmb:
Fix Windows CI breaking due to https://github.com/microsoft/ebpf-for-windows/pull/4662 by switching to a different program type which is distributed as part of sample_ebpf_ext. Unfortunately Windows stricter requirements on BPF_PROG_RUN for this program type. Instead of adding lots of Windows specific code to tests we break `Program.Benchmark` and `Program.Test` and add some compat code in there.

See the individual commit messages for more context.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -346,12 +346,14 @@ jobs:
           Invoke-WebRequest -Uri "${{ steps.determine-url.outputs.download_url }}" -OutFile "$env:TEMP\efw.zip"
           Expand-Archive -Path "$env:TEMP\efw.zip" -DestinationPath "$env:TEMP"
           Expand-Archive -Path "$env:TEMP\build-Debug.zip" -DestinationPath "$env:TEMP\ebpf"
-          $setupScript = Get-ChildItem -Path "$env:TEMP\ebpf" -Filter "setup-ebpf.ps1" -Recurse | Select-Object -First 1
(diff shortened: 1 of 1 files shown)


prog.go
```diff
@@ -707,83 +700,107 @@ type RunOptions struct {
...
 	// Note not all program types support this field.
 	CPU uint32
-	// Called whenever the syscall is interrupted, and should be set to testing.B.ResetTimer
-	// or similar. Typically used during benchmarking. Optional field.
-	//
-	// Deprecated: use [testing.B.ReportMetric] with unit "ns/op" instead.
-	Reset func()
+}
+
+type testingTB interface {
```
florianl: nit: maybe add a comment that this interface intends to mimic the interface of the package testing without importing the package.

prog.go
```diff
@@ -707,83 +700,107 @@ type RunOptions struct {
...
-	}
+	opts = p.fillRunOptionsForTesting(opts)
+	opts.Repeat = 1
 
-	ret, _, err := p.run(&opts)
-	if err != nil {
-		return ret, nil, fmt.Errorf("test program: %w", err)
+	ret, _, err := p.run(opts, nil)
+	if errors.Is(err, ErrNotSupported) {
+		t.Skipf("Execute %s program: %v", p.typ, err)
```
florianl: Calling t.Skipf for the ErrNotSupported case can be tricky, I think. The skip information is usually only printed if tests run with verbose enabled. So this can lead to the situation, that `Test()` is called on a ebpf program type, e.g. perf, that is not implemented. Without closer look and in verbose more, this could lead to false expectations.

PR #1881, opened 2025-10-21
fix(link/perf_event): ensure both perf link and perf event FDs are always closed

maheshbhatiya73:
### Summary

`perfEventLink.Close()` previously returned early if closing the link FD failed,
skipping cleanup of the underlying perf event (`pl.pe`). This caused:

- Perf event file descriptor leaks
- Stale tracefs events
- Potential resource exhaustion in long-running programs

### Changes

This patch ensures that both resources are always closed, collecting all
errors using Go 1.20+ `errors.Join`:

```go
func (pl *perfEventLink) Close() error {
    var errs []error
    if err := pl.fd.Close(); err != nil {
        errs = append(errs, fmt.Errorf("perf link close: %w", err))
    }
    if pl.pe != nil {
        if err := pl.pe.Close(); err != nil {
            errs = append(errs, fmt.Errorf("perf event close: %w", err))
        }
    }
    return errors.Join(errs...)
}

### DIFF
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -97,20 +97,24 @@ type perfEventLink struct {
 func (pl *perfEventLink) isLink() {}
 
 func (pl *perfEventLink) Close() error {
+	var errs []error
+
+	// Always attempt to close the link FD first.
 	if err := pl.fd.Close(); err != nil {
-		return fmt.Errorf("perf link close: %w", err)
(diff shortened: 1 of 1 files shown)


lmb:
Couple of questions:
1. Have you seen this in problem in production?
2. Did you make this PR using an AI coding tool?

Issue #1868, opened 2025-09-24
Allow callers to create VariableSpecs


ti-mo:
For testing purposes, it would be really convenient if the caller could instantiate their own VariableSpecs. Cilium currently uses something like this:
```go
var _ VariableSpec = (*ebpf.VariableSpec)(nil)

type VariableSpec interface {
	MapName() string
	Offset() uint64
	Size() uint64
	Get(out any) error
	Constant() bool
}
```

This requires converting ebpf.VariableSpecs to our Cilium-internal VariableSpec since that's what our internal functions take to allow tests to provide their own.

This issue is for adding a `NewVariableSpec` or similar with some sanity checks to allow users to create their own. As long as the accesses performed by the VariableSpec are valid, I guess this isn't something we should really prevent. Pay attention to CollectionSpec/VariableSpec.Copy() behaviour.

lmb:
Not sure where we had the discussion, but I think the answer is to export the fields somehow. That would make it in line with our other specs as well.

PR #1882, opened 2025-10-22
Allow callers to create VariableSpecs

lmb:
See the commit messages for details.

Fixes: https://github.com/cilium/ebpf/issues/1868

### DIFF
--- a/collection.go
+++ b/collection.go
@@ -615,9 +615,9 @@ func (cl *collectionLoader) populateStructOps(m *Map, mapSpec *MapSpec) error {
 		return fmt.Errorf("value should be a *Struct")
 	}
 
-	userData, ok := mapSpec.Contents[0].Value.([]byte)
-	if !ok {
-		return fmt.Errorf("value should be an array of byte")
+	userData, err := mapSpec.dataSection()
+	if err != nil {
+		return fmt.Errorf("getting data section: %w", err)
 	}
 	if len(userData) < int(userType.Size) {
 		return fmt.Errorf("user data too short: have %d, need at least %d", len(userData), userType.Size)
--- a/map.go
+++ b/map.go
@@ -29,7 +29,6 @@ var (
 	ErrKeyExist         = errors.New("key already exists")
 	ErrIterationAborted = errors.New("iteration aborted")
 	ErrMapIncompatible  = errors.New("map spec is incompatible with existing map")
-	errMapNoBTFValue    = errors.New("map spec does not contain a BTF Value")
 
 	// pre-allocating these errors here since they may get called in hot code paths
 	// and cause unnecessary memory allocations
@@ -187,32 +186,23 @@ func (spec *MapSpec) fixupMagicFields() (*MapSpec, error) {
 	return spec, nil
 }
 
-// dataSection returns the contents and BTF Datasec descriptor of the spec.
-func (ms *MapSpec) dataSection() ([]byte, *btf.Datasec, error) {
-	if ms.Value == nil {
-		return nil, nil, errMapNoBTFValue
-	}
-
-	ds, ok := ms.Value.(*btf.Datasec)
-	if !ok {
-		return nil, nil, fmt.Errorf("map value BTF is a %T, not a *btf.Datasec", ms.Value)
-	}
-
+// dataSection returns the contents of a datasec if the MapSpec represents one.
+func (ms *MapSpec) dataSection() ([]byte, error) {
 	if n := len(ms.Contents); n != 1 {
-		return nil, nil, fmt.Errorf("expected one key, found %d", n)
+		return nil, fmt.Errorf("expected one key, found %d", n)
 	}
 
 	kv := ms.Contents[0]
 	if key, ok := ms.Contents[0].Key.(uint32); !ok || key != 0 {
-		return nil, nil, fmt.Errorf("expected contents to have key 0")
+		return nil, fmt.Errorf("expected contents to have key 0")
 	}
 
 	value, ok := kv.Value.([]byte)
 	if !ok {
-		return nil, nil, fmt.Errorf("value at first map key is %T, not []byte", kv.Value)
+		return nil, fmt.Errorf("value at first map key is %T, not []byte", kv.Value)
 	}
 
-	return value, ds, nil
+	return value, nil
 }
 
 // updateDataSection copies the values of variables into MapSpec.Contents[0].Value.
@@ -232,7 +222,7 @@ func (ms *MapSpec) updateDataSection(vars map[string]*VariableSpec, sectionName 
 		return nil
 	}
 
-	data, _, err := ms.dataSection()
+	data, err := ms.dataSection()
 	if err != nil {
 		return err
 	}


lmb:
@ti-mo can you give this a review please?

map.go
```diff
@@ -188,27 +188,61 @@ func (spec *MapSpec) fixupMagicFields() (*MapSpec, error) {
 }
 
 // dataSection returns the contents and BTF Datasec descriptor of the spec.
```
ti-mo:
```suggestion
// dataSection returns the contents of a datasec if the MapSpec represents one.
```

map.go
```diff
@@ -188,27 +188,61 @@ func (spec *MapSpec) fixupMagicFields() (*MapSpec, error) {
...
 
-	if n := len(ms.Contents); n != 1 {
-		return nil, nil, fmt.Errorf("expected one key, found %d", n)
+	return value, nil
+}
+
+// updateDataSection copies the values of vars into MapSpec.Contents[0].Value.
+//
+// vars is sorted in place by offset.
+func (ms *MapSpec) updateDataSection(vars []*VariableSpec) error {
```
ti-mo: Nit:

```suggestion
func (ms *MapSpec) mergeVariables(vars []*VariableSpec) error {
```

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
-	size   uint64
-
-	m *MapSpec
-	t *btf.Var
+	Name string
+	// Name of the map this variable belongs to.
+	MapName string
+	// Offset of the variable within the map.
+	Offset int
+	// Byte representation of the variables's value.
```
ti-mo:
```suggestion
	// Byte representation of the variable's value.
```

cmd/bpf2go/gen/types.go
```diff
@@ -44,12 +44,7 @@ func collectMapTypes(types []btf.Type, maps map[string]*ebpf.MapSpec) []btf.Type
 // collectVariableTypes collects all types used by VariableSpecs.
 func collectVariableTypes(types []btf.Type, vars map[string]*ebpf.VariableSpec) []btf.Type {
 	for _, vs := range vars {
-		v := vs.Type()
```
ti-mo, line 47: What's the rationale for removing the nil check? bpf2go always emits BTF-enabled ELFs?

collection.go
```diff
@@ -170,7 +170,7 @@ func (cs *CollectionSpec) RewriteConstants(consts map[string]interface{}) error
 			continue
 		}
 
-		if !v.Constant() {
+		if !strings.HasPrefix(v.MapName, ".rodata") {
```
ti-mo: We have a helper for this: `isConstantDataSection()`.

collection.go
```diff
@@ -514,6 +514,19 @@ func (cl *collectionLoader) loadMap(mapName string) (*Map, error) {
 		return m, nil
 	}
 
+	var vars []*VariableSpec
```
ti-mo: Maybe lift this into `updateDataSection()` and make it take a `map[string]*VariableSpec`?

elf_reader.go
```diff
@@ -1259,11 +1262,19 @@ func (ec *elfCode) loadDataSections() error {
 				continue
 			}
 
+			if off+sym.Size > uint64(len(data)) {
```
ti-mo, line 1280: Have you run into these cases in practice?

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
 type VariableSpec struct {
-	name   string
-	offset uint64
-	size   uint64
-
-	m *MapSpec
-	t *btf.Var
+	Name string
+	// Name of the map this variable belongs to.
+	MapName string
```
ti-mo: Since we're breaking API, I'm on the fence about calling this `SectionName` instead to match `ProgramSpec.SectionName`. WDYT?

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
+	// Offset of the variable within the map.
+	Offset int
+	// Byte representation of the variables's value.
+	Value []byte
```
ti-mo, line 26: This should prob remain unexported for now, but that wouldn't work with the current implementation of Get() and Set().

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
 // host's native endianness.
 type VariableSpec struct {
-	name   string
-	offset uint64
-	size   uint64
-
-	m *MapSpec
-	t *btf.Var
+	Name string
+	// Name of the map this variable belongs to.
```
ti-mo:
```suggestion
	// Name of the datasec this variable was allocated in.
```

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
+	MapName string
+	// Offset of the variable within the map.
```
ti-mo:
```suggestion
	// Offset of the variable within the datasec.
```

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
+	Offset int
```
ti-mo: I'd rather this just be a uint64 like it was before. (`elf.SectionHeader.Offset` and `.Size` are u64) We have no input validation on negative ints, so the first error the user gets when specifying a negative offset is `variable %s overlaps with previous variable`, which points in the right direction but is still inaccurate.

Sticking to uints sidesteps this issue.

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
 // Set sets the value of the VariableSpec to the provided input using the host's
 // native endianness.
 func (s *VariableSpec) Set(in any) error {
-	buf, err := sysenc.Marshal(in, int(s.size))
-	if err != nil {
-		return fmt.Errorf("marshaling value %s: %w", s.name, err)
-	}
-
-	b, _, err := s.m.dataSection()
+	buf, err := sysenc.Marshal(in, len(s.Value))
```
ti-mo: We should try to account for nil `s.Value` here. Allocating new VariableSpecs and calling .Set() looks pretty annoying if the caller needs to correctly size the underlying buffer beforehand. (imagine creating a `map[name]*VariableSpec` using a composite literal)

There may be a few options here:

1. Add a VariableSpec.Size field and allocate the buffer automatically in Set()
1. `make([]byte, binary.Size(in))` or its reflect equivalent if nil, but potentially comes up with a different size than `unsafeBackingMemory()`
1. `NewVariableSpec()` that takes all values and calls VariableSpec.Set()

Option 1 looks the most appealing to me since it captures the caller's intent well (this variable is of size x at offset y and will be of value z), without cluttering comp lits with `make([]byte, ...)`. Potential downside is multiple sources of truth regarding variable size, which I guess is why you opted for the current proposal. If Value is unexported, this is less of an issue, we just need to take into account the caller could change s.Size after calling s.Set(). We'd always have to consider `len(s.Value)` to be authoritative.

Thoughts?

variable.go
```diff
@@ -15,118 +16,48 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("getting data section of map %s: %w", s.m.Name, err)
-	}
-
-	if int(s.offset+s.size) > len(b) {
-		return fmt.Errorf("offset %d(+%d) for variable %s is out of bounds", s.offset, s.size, s.name)
-	}
-
-	if err := sysenc.Unmarshal(out, b[s.offset:s.offset+s.size]); err != nil {
+	if err := sysenc.Unmarshal(out, s.Value); err != nil {
```
ti-mo, line 65: The result of calling Get() with a nil s.Value should be well-defined and documented.

cmd/bpf2go/gen/types.go
lmb: Yes it does, not exactly sure why I removed this though.

variable.go
lmb: Interesting point, I need to check what the semantics are. It's a bit confusing with now with names having some characters stripped, so SectionName might be better indeed.

elf_reader.go
lmb: Nope, just defense in depth to return an error instead of panicking.

variable.go
lmb: I half agree, in the sense that maybe int isn't the right unit. But uint64 also feels wrong, because most of the other lengths we use are uint32 or int. Maybe uint32 is better?

variable.go
lmb: I had code for this but ripped it out since I wasn't sure about it. I'd probably do the following:

- if Type != nil use binary.Size()
- otherwise just use whatever the first marshaling gives us

I'd prefer to keep .Value: I think it's good that all fields of *Spec are public.

variable.go
lmb: I agree about well defined, do we need to document it? Seems straightforward that it should return an error or some sort, but does it matter which?

variable.go
ti-mo: Not sure if I follow the other lengths. Why does it feel wrong? Variable.offset and .size are also uint64. They're encoded this way in the ELF.

variable.go
ti-mo: It doesn't matter which, but I'd call it out and maybe add a `len(s.Value) == 0` check at the top to avoid any arcane errors from surfacing.

variable.go
lmb: Public fields in *Spec which specify sized. MapSpec.KeySize, ValueSize for example. Yes they are 64 bit in the ELF, but we can't ever create a map > 32 bit so having offset and size larger than that doesn't make much sense.

In general, having to deal with mismatched size units is a bit of a pain, so having less is better.

variable.go
ti-mo: > I'd prefer to keep .Value: I think it's good that all fields of *Spec are public.

Okay, I can see the appeal.

> otherwise just use whatever the first marshaling gives us

We'll need to make Marshal work with a zero `size`, though. Even in the case of `binary.Append`, len(buf) needs to match size. Should we disable checking size if it's zero?

variable.go
ti-mo: Okay, whatever you choose, please make it consistent across Variable and its exported API. Those are all 64 bits.

cmd/bpf2go/gen/types.go
lmb: The nil check ends up not being necessary because of how addType works.

map.go
```diff
@@ -187,28 +187,71 @@ func (spec *MapSpec) fixupMagicFields() (*MapSpec, error) {
...
+	data, err := ms.dataSection()
+	if err != nil {
+		return err
 	}
 
-	return value, ds, nil
+	// Do not modify the original data slice, ms.Contents is a shallow copy.
+	data = slices.Clone(data)
+
+	sort.Slice(specs, func(i, j int) bool {
```
ti-mo: slices.Sorted()?

ti-mo:
@lmb Please check my feedback commit and squash if okay. We need Size() and Constant() in Cilium, `len(vs.Value)` is annoying to use, and it's pretty nice if we can fall back to BTF in case it's not initialized yet.

gnurizen:
FYI, this commit appears to break the opentelemetry-ebpf-profiler, it appears we just call `VariableSpec.Set` and load the maps our selves with `NewMap` and bypass the collection loader which I guess copies the `VariableSpec`s over now in `updateDataSection` instead of updating the underlying spec.   Apparently we aren't using the collection loader because we have to do a bunch of map resizings and loader chicanery to support using the same programs/maps across kprobes and perf probes.  Seems like a missing release note at minimum, not sure if it qualifies as a regression.   Maybe a new `(*MapSpec) ApplyVariables()` or something would be nice but I think we can just copy the variables into the mapSpec ourselves for the time being.  Any guidance would be appreciated!

gnurizen:
Nevermind, apparently I'm late to the game and the otel profiler has already upgraded and addressed this.

PR #1883, opened 2025-10-27
struct_ops: skip copy for zeroed nested structs and fail on non-zeroed

### LINKED ISSUE #1880: struct_ops: value type should allow nested structs and unions
We should not error here, just ignore it.

There are struct ops which have nested structs such as [`struct tcp_congestion_ops`](https://elixir.bootlin.com/linux/v6.17.1/source/include/net/tcp.h#L1262). When working with a definition copied from sources or vmlinux.h, you hit this.

_Originally posted by @dylandreimerink in https://github.com/cilium/ebpf/pull/1845#discussion_r2435812714_

The fix is to make the copy function recursive.

shun159:
resolves #1880


lmb:
Thanks!

PR #1886, opened 2025-11-03
staticcheck: run main, remove staticcheck.conf

ti-mo:
Pick up dominikh/go-tools@8273271 and drop the config to override enabled checks. Adding the config was needed to make staticcheck pass after stringer made a change violating SA4003. Generated files are now ignored for this check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Run staticcheck
         uses: dominikh/staticcheck-action@v1
         with:
-          version: "latest"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Run staticcheck
         uses: dominikh/staticcheck-action@v1
         with:
-          version: "latest"
+          version: "main"
```
dylandreimerink: Seems like this should be "master" not "main"

PR #1889, opened 2025-11-04
feat: read full program name from function info

secDre4mer:
While the program name in the kernel is limited to 16 bytes (including a terminating zero byte), the BTF information may contain the full name.

bpftool uses this information to extract the full program name, if possible, see https://github.com/libbpf/bpftool/blob/b799c6f513400009d56cfc11b8e3a8fcbd2b7c14/src/common.c#L413

Adapt this logic to do the same.

This also exposed a bug where some information was missing due to strange (arguably buggy) kernel behaviour. Add a workaround for this as well.

### DIFF
--- a/info.go
+++ b/info.go
@@ -305,6 +305,7 @@ type ProgramInfo struct {
 
 	maps                 []MapID
 	insns                []byte
+	numInsns             uint32
 	jitedSize            uint32
 	verifiedInstructions uint32
 
@@ -372,6 +373,7 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 		jitedSize:            info.JitedProgLen,
 		loadTime:             time.Duration(info.LoadTime),
 		verifiedInstructions: info.VerifiedInsns,
+		numInsns:             info.XlatedProgLen,
 	}
 
 	// Supplement OBJ_INFO with data from /proc/self/fdinfo. It contains fields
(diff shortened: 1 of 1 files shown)


info.go
```diff
@@ -416,9 +416,22 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
 
 	if info.XlatedProgLen > 0 {
 		pi.insns = make([]byte, info.XlatedProgLen)
+		var info2 sys.ProgInfo
```
dylandreimerink: Call this `info3`, so we do not shadow the `info2` that is declared outside of the scope of this if statement. It avoids confusion.

info.go
```diff
@@ -416,9 +416,22 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
...
 		info2.XlatedProgLen = info.XlatedProgLen
 		info2.XlatedProgInsns = sys.SlicePointer(pi.insns)
-		makeSecondCall = true
+
+		// Linux kernel has a bug(?): if xlated_prog_len is specified and access is denied
+		// by kernel.kptr_restrict or net.core.bpf_jit_harden sysctls, the syscall doesn't only
+		// skip the xlated program's bytes, but other information as well.
+		// See handling of xlated_prog_insns in bpf_prog_get_info_by_fd in kernel/bpf/syscall.c.
+		// To avoid skipping other information, we specify xlated_prog_len in a separate syscall.
```
dylandreimerink: Lets make this comment slightly more concise. Include as much detail in the git commit to add context to this.

```suggestion
		// When kernel.kptr_restrict and net.core.bpf_jit_harden are both set, it causes the
		// syscall to abort when trying to readback xlated instructions, skipping other info 
		// as well. So request xlated instructions seperately.
```

info.go
```diff
@@ -499,9 +499,38 @@ func newProgramInfoFromFd(fd *sys.FD) (*ProgramInfo, error) {
...
+	funcInfos, err := btf.LoadFuncInfos(
+		bytes.NewReader(pi.funcInfos),
+		internal.NativeEndian,
+		1, // Only the first info
+		spec,
+	)
+	if err != nil {
+		return "", err
+	}
+	return funcInfos[0].Func.Name, nil
```
dylandreimerink: This code a bad assumption. That the first function info in the BTF is the func info for the entrypoint, but this is not guaranteed.

Additionally, the code currently assumes `funcInfos` always has at least 1 element, because `pi.numFuncInfos` != 0. But if something when silently wrong somewhere, `funcInfos` could have 0 elements, and the unchecked access at index 0 causes a panic. 

The proper way to do this is to request all func infos (`pi.numFuncInfos`), so that we read back all func infos from the kernel. Then loop over the funcInfos until you find the function with `Offset` 0. It should be the first entry, since `btf.LoadFuncInfos` sorts by `Offset`, but we don't have to depend on that behavior. Bonus points, the loop also means that if we have 0 elements, we don't panic.

secDre4mer: Thanks, good catch. Adjusted the function as you suggested.

info.go
secDre4mer: Done.

PR #1894, opened 2025-11-05
Deprecate All The Things

ti-mo:
This PR removes a bunch of interfaces we've marked as deprecated over the years, but never followed through on removing them. All of these are at least a few releases old, some dating back to 0.9.0. This should've given users ample time to migrate.

The addition of alternative, improved APIs are typically what causes deprecation notices to be given, so there should be a clear path forward for all things removed here.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -240,22 +240,6 @@ func (ins *Instruction) encodeMapFD(fd int) {
 	ins.Constant = int64(offset | rawFd)
 }
 
-// MapPtr returns the map fd for this instruction.
-//
-// The result is undefined if the instruction is not a load from a map,
-// see IsLoadFromMap.
-//
-// Deprecated: use Map() instead.
-func (ins *Instruction) MapPtr() int {
-	// If there is a map associated with the instruction, return its FD.
-	if fd := ins.Metadata.Get(mapMeta{}); fd != nil {
-		return fd.(FDer).FD()
-	}
-
-	// Fall back to the fd stored in the Constant field
-	return ins.mapFd()
-}
-
 // mapFd returns the map file descriptor stored in the 32 least significant
 // bits of ins' Constant field.
 func (ins *Instruction) mapFd() int {


collection.go
```diff
@@ -95,97 +94,6 @@ func copyMapOfSpecs[T interface{ Copy() T }](m map[string]T) map[string]T {
...
-// RewriteMaps replaces all references to specific maps.
-//
-// Use this function to use pre-existing maps instead of creating new ones
-// when calling NewCollection. Any named maps are removed from CollectionSpec.Maps.
-//
-// Returns an error if a named map isn't used in at least one program.
-//
-// Deprecated: Pass CollectionOptions.MapReplacements when loading the Collection
-// instead.
-func (cs *CollectionSpec) RewriteMaps(maps map[string]*Map) error {
```
florianl, line 107: This will be a hard breaking change for Open Telemetry.

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/1e5516f97d8b8dea21ad0d9d3f6e126647013fcf/tracer/tracer.go#L325

rgo3 (approved):
Was the retry logic always only intended for the SIGPROF case? Masking SIGPROF has been around for a while, wondering why we only remove the retry logic now?

Other deprecations LGTM!

Giga nit: There's a small typo in that last commit message, s/endlessely/endlessly

collection.go
lmb: I’d be fine un-deprecating that. It’s fine to have one declarative and one imperative way of doing things.

lmb:
> Was the retry logic always only intended for the SIGPROF case?

I think it also applies in the case of asynchronous preemption, which uses signals on Linux. Might have to double check whether that has special casing for threads in a syscall to not signal them (aka those are considered preempted).

I would remove reset but keep the retry behaviour unchanged. I’ve got a branch which does that, I should probably dust it off.

More generally, EINTR means try this operation again. Signals used to kill a process won’t return EINTR. So it’s not a clear cut thing to say that EINTR should be returned. The fact that nobody has complained so far tells me that the current code is at least not detrimental. Or that not many people benchmark programs like this (I found very few callers in open source software).

collection.go
ti-mo: I see, otel-ebpf-profiler doesn't use NewCollection, full stop.. I don't see an easy way out of that, but the way the code currently works, I think you're locked out of using bpf2bpf at least? Maybe more features down the line. We don't really intend for ProgramSpecs to be pulled out of a CollectionSpec and loaded independently, and it's not something we really test for iirc.

We deprecated this because plumbing live bpf resources (Program/Map) into *Spec types is somewhat of a hazard, and we'd prefer not to recommend it to users. At least with FDer involved, it'll resolve to a -1 fd if the underlying map gets closed in the meantime, so there's no risk of reusing a wrong fd.

This kind of conditional loading mechanism can be made much more elegant by supporting an `[]any` argument to LoadAndAssign (we need this for Cilium too). You assemble a slice of output objects based on the features enabled, you provide your pre-created maps, and lazy loading does the rest. This will still require major refactoring for you, but the outcome should be much more maintainable.

I see you've `//nolint:staticcheck`'ed the call to `coll.RewriteMaps`. 🙂 This has been marked deprecated for years at this point. I'm willing to postpone the removal, but IMO ample notice was given.

collection.go
```diff
@@ -95,97 +94,6 @@ func copyMapOfSpecs[T interface{ Copy() T }](m map[string]T) map[string]T {
...
-//
-// Replacement values must be of the same length as the C sizeof(type).
-// If necessary, they are marshalled according to the same rules as
-// map values.
-//
-// From Linux 5.5 the verifier will use constants to eliminate dead code.
-//
-// Returns an error wrapping [MissingConstantsError] if a constant doesn't exist.
-//
-// Deprecated: Use [CollectionSpec.Variables] to interact with constants instead.
```
mmat11, line 162: I think most of the times these checks need to be made anyway when setting constants -- I understand this is just an utility wrapper but it's practically what most projects end up rewriting:

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/main/pkg/internal/ebpf/convenience/convenience.go#L41

https://github.com/cilium/tetragon/blob/main/pkg/sensors/program/loader_linux.go#L863

https://github.com/elastic/ebpfevents/blob/main/loader.go#L90

https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/utils/ebpf/ebpf.go#L24

maybe we can consider undeprecating it?

ti-mo: Thanks for digging into this! The thing is; the damage for these particular projects (and likely others as well) has already been done. These projects already moved over and figured the existing implementation of `RewriteConstants` suited their needs just fine, so they copied it since they likely didn't want to spend time on it (understandable) and to minimize the risk of regressions. Not counting inspektor-gadget here, since they opted for a different implementation.

Same thing as with RewriteMaps above, we've made the choice of deprecating it because we found shortcomings and/or footguns in the existing API. In the case of RewriteConstants, as the name suggests, it only allows accessing a subset of global vars (consts), and MissingConstantsError was introduced for the niche use case of checking for the existence of vars in a trial-and-error manner for inspektor-gadget. [They didn't even end up using it eventually.](https://github.com/inspektor-gadget/inspektor-gadget/pull/1201#discussion_r1095999193)

I think it's best to leave it up to the user to decide how they want to go about reading/writing values, checking membership, and how to handle errors. CollectionSpec.Variables was inspired by libbpf skeleton, so on the off chance folks are porting over from libbpf, the new API may also feel more straightforward. Also, there should be one-- and preferably only one --obvious way to do it. ™

mmat11: I understand, thanks for the explanation

collection.go
florianl: I have opened https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/954 to bring this update forward.

ti-mo: @florianl If you need some extra time to refactor, you can also just copy the current implementation of `RewriteMaps` as-is, no pressure. It's only using exported API.

florianl: I think, that will be the short term solution until a major refactory of this part of the code.

ti-mo:
@lmb I've gone ahead and backed out of the Reset() removal entirely. Let's go ahead with converting Program.Benchmark() to take a `testing.B` instead, since it feels pointless to break API twice if we're planning on doing that anyway. When we make this conversion, we can hide Reset() from the caller and report our own metric, e.g. `retries` to both make sure retries don't go unnoticed and don't need to be handled explicitly by the user.

Then the `retries` argument also becomes unnecessary as we can take `b.N` as the value to `RunOptions.Repeat`.

PR #1895, opened 2025-11-05
examples: Add a minimal sched_ext example

shun159:
Add a minimal sched_ext example to showcase StructOps support


shun159:
thx! @lmb

PR #1896, opened 2025-11-07
feat: more link inspection support

secDre4mer:
Add support for more link inspection types, e.g. to query a uprobe's symbol.

### DIFF
--- a/link/link_other.go
+++ b/link/link_other.go
@@ -111,9 +111,9 @@ func wrapRawLink(raw *RawLink) (_ Link, err error) {
 }
 
 type TracingInfo struct {
-	AttachType  sys.AttachType
-	TargetObjId uint32
-	TargetBtfId sys.TypeID
+	AttachType     sys.AttachType
+	TargetObjectId uint32
+	TargetBtfId    sys.TypeID
 }
 
 type CgroupInfo struct {
@@ -123,7 +123,7 @@ type CgroupInfo struct {
 }
 
 type NetNsInfo struct {
-	NetnsIno   uint32
+	NetnsInode uint32
 	AttachType sys.AttachType
 }
 
@@ -137,10 +137,10 @@ type XDPInfo struct {
 }
 
 type NetfilterInfo struct {
-	Pf       NetfilterProtocolFamily
-	Hooknum  NetfilterInetHook
-	Priority int32
-	Flags    uint32
+	ProtocolFamily NetfilterProtocolFamily
+	Hook           NetfilterInetHook
+	Priority       int32
+	Flags          uint32
 }
 
 type NetkitInfo struct {
@@ -276,10 +276,10 @@ type UprobeInfo struct {
 	// However, due to various circumstances (differing mount namespaces,
 	// file replacement, ...), this path may not point to the same binary
 	// the uprobe was originally attached to.
-	File         string
-	Offset       uint32
-	Cookie       uint64
-	RefCtrOffset uint64
+	File                 string
+	Offset               uint32
+	Cookie               uint64
+	OffsetReferenceCount uint64
 }
 
 type TracepointInfo struct {
--- a/link/netfilter.go
+++ b/link/netfilter.go
@@ -41,7 +41,7 @@ type NetfilterOptions struct {
 	// The protocol family.
 	ProtocolFamily NetfilterProtocolFamily
 	// The netfilter hook to attach to.
-	HookNumber NetfilterInetHook
+	Hook NetfilterInetHook
 	// Priority within hook
 	Priority int32
 	// Extra link flags
@@ -74,7 +74,7 @@ func AttachNetfilter(opts NetfilterOptions) (Link, error) {
 		AttachType:     sys.BPF_NETFILTER,
 		Flags:          opts.Flags,
 		Pf:             opts.ProtocolFamily,
-		Hooknum:        opts.HookNumber,
+		Hooknum:        opts.Hook,
 		Priority:       opts.Priority,
 		NetfilterFlags: uint32(opts.NetfilterFlags),
 	}
@@ -97,10 +97,10 @@ func (nf *netfilterLink) Info() (*Info, error) {
 		return nil, fmt.Errorf("netfilter link info: %s", err)
 	}
 	extra := &NetfilterInfo{
-		Pf:       info.Pf,
-		Hooknum:  info.Hooknum,
-		Priority: info.Priority,
-		Flags:    info.Flags,
+		ProtocolFamily: info.Pf,
+		Hook:           info.Hooknum,
+		Priority:       info.Priority,
+		Flags:          info.Flags,
 	}
 
 	return &Info{
--- a/link/perf_event.go
+++ b/link/perf_event.go
@@ -185,10 +185,10 @@ func (pl *perfEventLink) Info() (*Info, error) {
 			return nil, fmt.Errorf("uprobe link info: %s", err)
 		}
 		extra2 = &UprobeInfo{
-			Offset:       uprobeInfo.Offset,
-			Cookie:       uprobeInfo.Cookie,
-			RefCtrOffset: uprobeInfo.RefCtrOffset,
-			File:         fileName,
+			Offset:               uprobeInfo.Offset,
+			Cookie:               uprobeInfo.Cookie,
+			OffsetReferenceCount: uprobeInfo.RefCtrOffset,
+			File:                 fileName,
 		}
 	case PerfEventTracepoint:
 		var tracepointInfo sys.TracepointLinkInfo
--- a/link/netfilter_test.go
+++ b/link/netfilter_test.go
@@ -19,7 +19,7 @@ func TestAttachNetfilter(t *testing.T) {
 	l, err := AttachNetfilter(NetfilterOptions{
 		Program:        prog,
 		ProtocolFamily: NetfilterProtoIPv4,
-		HookNumber:     NetfilterInetLocalOut,
+		Hook:           NetfilterInetLocalOut,
 		Priority:       -128,
 	})
 	if err != nil {
@@ -31,8 +31,8 @@ func TestAttachNetfilter(t *testing.T) {
 		t.Fatal(err)
 	}
 	nfInfo := info.Netfilter()
-	qt.Assert(t, qt.Equals(nfInfo.Pf, NetfilterProtoIPv4))
-	qt.Assert(t, qt.Equals(nfInfo.Hooknum, NetfilterInetLocalOut))
+	qt.Assert(t, qt.Equals(nfInfo.ProtocolFamily, NetfilterProtoIPv4))
+	qt.Assert(t, qt.Equals(nfInfo.Hook, NetfilterInetLocalOut))
 	qt.Assert(t, qt.Equals(nfInfo.Priority, -128))
 
 	testLink(t, l, prog)
--- a/link/netns.go
+++ b/link/netns.go
@@ -44,7 +44,7 @@ func (ns *NetNsLink) Info() (*Info, error) {
 		return nil, fmt.Errorf("netns link info: %s", err)
 	}
 	extra := &NetNsInfo{
-		NetnsIno:   info.NetnsIno,
+		NetnsInode: info.NetnsIno,
 		AttachType: info.AttachType,
 	}
 
--- a/link/tracing.go
+++ b/link/tracing.go
@@ -26,9 +26,9 @@ func (f *tracing) Info() (*Info, error) {
 		return nil, fmt.Errorf("tracing link info: %s", err)
 	}
 	extra := &TracingInfo{
-		TargetObjId: info.TargetObjId,
-		TargetBtfId: info.TargetBtfId,
-		AttachType:  info.AttachType,
+		TargetObjectId: info.TargetObjId,
+		TargetBtfId:    info.TargetBtfId,
+		AttachType:     info.AttachType,
 	}
 
 	return &Info{


link/kprobe_multi.go
```diff
@@ -159,10 +159,28 @@ func (kml *kprobeMultiLink) Info() (*Info, error) {
...
 		return nil, fmt.Errorf("kprobe multi link info: %s", err)
 	}
+	var addrs = make([]uint64, info.Count)
+	var cookies = make([]uint64, info.Count)
+	info = sys.KprobeMultiLinkInfo{
+		Addrs:   sys.SlicePointer(addrs),
+		Cookies: sys.SlicePointer(cookies),
+		Count:   uint32(len(addrs)),
+	}
+	if err := sys.ObjInfo(kml.fd, &info); err != nil {
```
mmat11, line 168: is it necessary to call obj_info twice?

link/link_other.go
```diff
@@ -167,6 +201,118 @@ func (kpm *KprobeMultiInfo) Missed() (uint64, bool) {
...
+	}
+	return adresses, true
+}
+
+// Symbols returns the symbols that the uprobe was attached to.
+func (umi *UprobeMultiInfo) Symbols() ([]UprobeSymbol, error) {
+	if umi.offsets == nil {
+		return nil, fmt.Errorf("no offsets available")
+	}
+	ex, err := OpenExecutable(umi.path)
```
mmat11: hmm I wonder if this is safe to do; should we just return addresses like for kprobeMulti?

link/link_other.go
```diff
@@ -177,9 +323,26 @@ func (r *PerfEventInfo) Kprobe() *KprobeInfo {
...
+	e, _ := r.extra.(*EventInfo)
+	return e
+}
+
 type KprobeInfo struct {
-	address uint64
-	missed  uint64
+	address  uint64
+	missed   uint64
+	Function string
```
mmat11, line 268: should we have these unexported with a getter?

link/link_other.go
```diff
@@ -167,6 +201,118 @@ func (kpm *KprobeMultiInfo) Missed() (uint64, bool) {
...
+}
+
+// Pid returns the process ID that this uprobe is attached to.
+//
+// If it does not exist, the uprobe will trigger for all processes.
+func (umi *UprobeMultiInfo) Pid() (uint32, bool) {
+	return umi.pid, umi.pid > 0
+}
+
+const (
```
mmat11, line 230: do these aliases need to be exported?

link/uprobe_multi.go
```diff
@@ -176,6 +177,58 @@ func (kml *uprobeMultiLink) Update(_ *ebpf.Program) error {
...
+		refCtrOffsets = nil
+	}
+	extra := &UprobeMultiInfo{
+		count:         info.Count,
+		flags:         info.Flags,
+		pid:           info.Pid,
+		offsets:       addrs,
+		cookies:       cookies,
+		refCtrOffsets: refCtrOffsets,
+		path:          string(bytes.TrimRight(path, "\x00")),
```
mmat11: nit: use [unix.ByteSliceToString()](https://pkg.go.dev/golang.org/x/sys/unix#ByteSliceToString)

secDre4mer:
> looks mostly good to me, left a few comments; can this be split in multiple commits?

Sure, I'll try. Any ideas on how the commits should be structured? E.g. one focused on uprobe, one focused on tracepoint, ...?

link/kprobe_multi.go
secDre4mer: Yes; we first need to query the address count, then allocate the address slice to a fitting size, and then query again so that the kernel can fill in the address slice.

link/link_other.go
secDre4mer: I'm not sure, that's why there are the different `Offsets` and `Symbols` functions. `Offsets` returns the raw offsets, and only `Symbols` tries to resolve them.

link/link_other.go
secDre4mer: I'm ambivalent to that; based on the kernel code I read, they should always be filled in (unlike address, which may be missing), so a getter doesn't provide more information. But I can understand wanting to keep it contiguous in a single struct.

Which way would you prefer?

link/link_other.go
secDre4mer: I think so. `sys` is an internal package, so users can't access the constants there; and they need some way to differentiate `PerfEventInfo.Type` so they know which subfunction to call.

link/link_other.go
mmat11: I think it's fine keeping as is as long as it's always filled

mmat11:
> > looks mostly good to me, left a few comments; can this be split in multiple commits?
> 
> Sure, I'll try. Any ideas on how the commits should be structured? E.g. one focused on uprobe, one focused on tracepoint, ...?

I guess you could split code generation and the different link types in different commits

link/uprobe_multi.go
secDre4mer: Done.

secDre4mer:
Split is done.

mmat11:
@secDre4mer sorry for the delay, I was on vacation; I ran CI and some tests are failing, could you please take a look?

secDre4mer:
> @secDre4mer sorry for the delay, I was on vacation; I ran CI and some tests are failing, could you please take a look?

Don't worry about the delay, I hope your vacation was nice!

The CI complains that some generated files are out of date, see [here](https://github.com/cilium/ebpf/actions/runs/19225109327/job/56240468961?pr=1896#step:6:314); however, running `make` locally, as suggested, does not change any generated files. Maybe it's due to differing kernel versions? I'm running 6.17.8 locally.

mmat11:
> > @secDre4mer sorry for the delay, I was on vacation; I ran CI and some tests are failing, could you please take a look?
> 
> Don't worry about the delay, I hope your vacation was nice!
> 
> The CI complains that some generated files are out of date, see [here](https://github.com/cilium/ebpf/actions/runs/19225109327/job/56240468961?pr=1896#step:6:314); however, running `make` locally, as suggested, does not change any generated files. Maybe it's due to differing kernel versions? I'm running 6.17.8 locally.

hmm I think you would need to update `CI_MAX_KERNEL_VERSION` https://github.com/cilium/ebpf/blob/main/.github/workflows/ci.yml#L10 and run `./scripts/update-kernel-deps.sh`

the script has been added after last time I tried to update kernel types so I am not sure this is the correct thing to do -- /cc @lmb

secDre4mer:
@mmat11 Thanks for the hints, the solution turned out to be somewhat easier than expected; main was already on 6.16, so I only had to rebase the branch and re-run `make update-external-deps`.

Now there are some issues with connecting to git.kernel.org, but those look unrelated to the changes I've made.

dylandreimerink:
Re-triggered the tests, seems like we got some legit failures now

secDre4mer:
I'll take a look.

secDre4mer:
@dylandreimerink Do you have any idea why the error with Netfilter / XDP occurs here? I'm a little stumped.
The direct source is clear: The kernel uses EBUSY to indicate that another eBPF program is already attached.
However, I don't get why this is the case. While both `TestAttachNetfilter` and `TestNetfilterInfo` use the same hook / priority, they are executed sequentially and each should remove their program / link at the end.
Also, this seems to happen quite reliably in the CI on kernel 6.12; but when I run 6.12 locally, I can't reproduce it.

ti-mo:
@secDre4mer Sorry for letting this languish, would you be able to rebase and resolve conflicts? Can't see the failures you were mentioning.

@mmat11 We're a little short on review cycles over here (what's new) so I'd appreciate if could take another look. Thanks! :pray:

secDre4mer:
@ti-mo sure, done.

mmat11:
> @mmat11 We're a little short on review cycles over here (what's new) so I'd appreciate if could take another look. Thanks! 🙏

I restarted the failing tests and the CI looks ok now, the code lgtm!

mmat11 (dismissed):

link/perf_event.go
```diff
@@ -132,6 +132,22 @@ func (pl *perfEventLink) PerfEvent() (*os.File, error) {
 	return fd.File("perf-event")
 }
 
+func queryInfoWithString(fd *sys.FD, info sys.Info, stringField *sys.TypedPointer[byte], stringLengthField *uint32) (string, error) {
+	// Query info to get the length
+	if err := sys.ObjInfo(fd, info); err != nil {
+		return "", err
+	}
+	var stringData = make([]byte, *stringLengthField)
```
dylandreimerink, line 148: Why is `stringLengthField` a `*uint32`, the only thing we do is dereference it and use it in `make`. And on the callsite you have to make a pointer as well. Just make this a `uint32`.

link/link_other.go
```diff
@@ -132,10 +134,32 @@ type TCXInfo struct {
 	AttachType sys.AttachType
 }
 
+func (t TCXInfo) Interface() (net.Interface, error) {
+	return resolveIfindex(t.Ifindex)
+}
```
dylandreimerink: I think we should leave this bit of API out. Nowhere in the library do we provide utility to go from ifindex to `net.Interface`, just let the user do this on their own.

dylandreimerink (changes requested):
Sorry for letting this linger, like Timo mentioned, review cycles have been short. But better late than never.

Overall this looks good, 2 minor points of interest.

Thanks for your work on this.

link/perf_event.go
secDre4mer: This needs to be a `*uint32` (though the reason is a bit intransparent, this should probably be documented better): `stringLengthField` points into the `info` object that was passed to `queryInfoWithString`, and its pointee is filled by the first `ObjInfo` call. Before that call, it's uninitialized and zero.

The pointee value is then used to create the byte slice, `stringField`'s pointee (which is similarly part of `info`) is initialized, and `ObjInfo` is called again, now with the string length and pointer set.

link/link_other.go
secDre4mer: Done.

link/perf_event.go
secDre4mer: Added a few more comments to `queryInfoWithString` to better explain what parameters are expected, and how the string length / field pointers are handled.

dylandreimerink: Right. I see now.

link/netfilter.go
```diff
@@ -13,13 +13,30 @@ const NetfilterIPDefrag NetfilterAttachFlags = 0 // Enable IP packet defragmenta
 
 type NetfilterAttachFlags uint32
 
+type NetfilterHook uint32
```
ti-mo: These constants already exist in package `x/sys/unix`. These are pretty tedious to maintain by hand, so let's not do that. Unfortunately, `nf_inet_hooks` in particular has a wrong identifier (0x5 should be NF_INET_INGRESS, but it only appears as NF_INET_NUMHOOKS). Go's tooling to generate these isn't perfect and the uapi conventions on this are kind of all over the place. Fortunately, these enums both appear in kernel BTF, so our `gentypes` tool can generate Go enums from them, and we can re-export their values through the link package, similar to what you've added here.

```go
type NetfilterInetHook uint32

const (
	NetfilterInetPreRouting NetfilterInetHook = sys.NF_INET_PRE_ROUTING
	NetfilterInetLocalIn NetfilterInetHook = sys.NF_INET_LOCAL_IN
    ...
)
```

For `nf_inet_hooks`, the case is straightforward: add a record to the `enums` slice in gentypes, since it's a named enum: https://github.com/cilium/ebpf/blob/61c4efeaf4c2993d0e74300003e29a3164b60ac9/internal/cmd/gentypes/main.go#L224

The `NFPROTO_*` ones are more tricky since they're in an anonymous enum. It would be nice if we could add a small feature to gentypes that collects all enum values matching a prefix and emits them as a Go 'enum', e.g.:

`{"NetfilterProtocolFamily", "NFPROTO_"},`

would match and collect these values:

```c
enum {
	NFPROTO_UNSPEC =  0,
	NFPROTO_INET   =  1,
	NFPROTO_IPV4   =  2,
	NFPROTO_ARP    =  3,
	NFPROTO_NETDEV =  5,
	NFPROTO_BRIDGE =  7,
	NFPROTO_IPV6   = 10,
	NFPROTO_DECNET = 12,
	NFPROTO_NUMPROTO,
};
```

and output the following definition:

```go
type NetfilterProtocolFamily uint32

const (
    NFPROTO_UNSPEC NetfilterProtocolFamily = 0
    NFPROTO_INET NetfilterProtocolFamily = 1
    ...
)
```

Alternatively, we could add an extra condition to the `strings.HasPrefix(..., "BPF_")` filter at the top, but it would be nice to have something proper as it lowers the cost of future additions.

secDre4mer: Added the collection of unnamed enums to `gentypes` and used the new types / constants for the user-facing ones.

link/link_other.go
```diff
@@ -167,6 +169,31 @@ func (kpm *KprobeMultiInfo) Missed() (uint64, bool) {
 	return kpm.missed, kpm.count > 0
 }
 
+type KprobeMultiAddress struct {
+	Address uint64
+	Cookie  uint64
+}
+
+// Addresses are the addresses hooked by the kprobe.
```
ti-mo, line 169: (curious as I'm not familiar with the kernel API yet) What's the reason for having Addresses() return both addresses and cookies and providing a separate Cookies() method in addition? I think the doc comments should do a better job at describing the API and how it's intended to be used.

From a distance, Cookies() doesn't look very useful on its own since there's no guarantee about the ordering of the output (I think?), and there's no address or symbol name to associate the values with.

link/link_other.go
ti-mo: Sorry to walk this back, but consistency is key. We've ended up with a random mix of methods and exported fields, adding to user confusion.

I've also never been a fan of those methods that only end up doing `return kp.address, kp.address > 0`, yeah no shit the field is zero. 😛 We initially went this way to leave the door open for real feature probes (e.g. bool true == kernel supports the field, even if it's zero), but given the amount of fields and permutations of features, this clearly never happened. I think we're getting more use out of them in ProgInfo, though, so we'll keep those around.

So, as a guide: if it's a simple field, let's just export it. If it's an array that needs more finessing or validation, let's put it behind a method. Same thing for when the kernel outputs a pointer and we need a type assertion (like Info.Tracing() and friends below), obviously we can't leave this up to the user to figure out.

@secDre4mer I like what you've done with kmi.Addresses() above, that's a good example of where providing an accessor is useful. Would you be up for removing `kp.Address()` and `kp.Missed()` below and exporting the fields instead? Now would be a good time to change this as we're touching surrounding functionality.

link/perf_event.go
```diff
@@ -140,14 +166,50 @@ func (pl *perfEventLink) Info() (*Info, error) {
...
+		}
+		extra2 = &TracepointInfo{
+			Tracepoint: tpName,
+			Cookie:     tracepointInfo.Cookie,
+		}
+	case PerfEventEvent:
+		var eventInfo sys.EventLinkInfo
+		err := sys.ObjInfo(pl.fd, &eventInfo)
+		if err != nil {
+			return nil, fmt.Errorf("trace point link info: %s", err)
```
ti-mo, line 207: `perf event link info`?

link/netfilter.go
```diff
@@ -13,13 +13,35 @@ const NetfilterIPDefrag NetfilterAttachFlags = 0 // Enable IP packet defragmenta
 
 type NetfilterAttachFlags uint32
 
+type NetfilterHook = sys.NetfilterInetHook
```
ti-mo: Nit: `type NetfilterInetHook`?

link/netfilter.go
```diff
@@ -13,13 +13,35 @@ const NetfilterIPDefrag NetfilterAttachFlags = 0 // Enable IP packet defragmenta
...
+	NetfilterProtoIPv6   = sys.NFPROTO_IPV6
+)
+
 type NetfilterOptions struct {
 	// Program must be a netfilter BPF program.
 	Program *ebpf.Program
 	// The protocol family.
-	ProtocolFamily uint32
+	ProtocolFamily NetfilterProtocolFamily
 	// The number of the hook you are interested in.
```
ti-mo: This can be made a bit less-specific now: `// The netfilter hook to attach to.`

link/link_other.go
```diff
@@ -198,6 +198,63 @@ func (kpm *KprobeMultiInfo) Cookies() ([]uint64, bool) {
 	return kpm.cookies, kpm.cookies != nil
 }
 
+type UprobeMultiInfo struct {
```
ti-mo, line 184: Same as below, let's just export the fields that make sense to access directly.

link/link_other.go
```diff
@@ -198,6 +198,63 @@ func (kpm *KprobeMultiInfo) Cookies() ([]uint64, bool) {
...
+}
+
+func (umi *UprobeMultiInfo) Missed() (uint64, bool) {
+	return umi.missed, umi.missed > 0
+}
+
+type UprobeMultiOffset struct {
+	Offset uint64
+	Cookie uint64
+	RefCtr uint64
```
ti-mo: I prefer Go-like field names like `References` or `ReferenceCount` in user-facing structs.

link/uprobe_multi_test.go
```diff
@@ -40,6 +40,31 @@ func TestUprobeMulti(t *testing.T) {
...
+	}
+	defer um.Close()
+
+	linkInfo, err := um.Info()
+	if err != nil {
+		t.Fatal(err)
+	}
+	qt.Assert(t, qt.Equals(linkInfo.Type, UprobeMultiType))
+	uprobeDetails := linkInfo.UprobeMulti()
+	qt.Assert(t, qt.StringContains(uprobeDetails.Path(), bashEx.path)) // On some platforms, /bin/bash may point to /usr/bin/bash, thus only a contains and no equals check
```
ti-mo: Nit: put long comment above the statement.

link/uprobe.go
```diff
@@ -196,12 +209,30 @@ func (ex *Executable) address(symbol string, address, offset uint64) (uint64, er
...
+	return symaddress.addr + offset, nil
+}
+
+// Symbol calculates the symbol and offset for an address in the executable.
+func (ex *Executable) Symbol(address uint64) (UprobeSymbol, error) {
+	if err := ex.lazyLoadSymbols(); err != nil {
+		return UprobeSymbol{}, fmt.Errorf("lazy load symbols: %w", err)
+	}
+
+	for name, symaddress := range ex.cachedAddresses {
```
ti-mo: `symaddress` -> `sym`? Then accessing the field looks like `sym.addr`, much easier to read and doesn't stutter.

link/link_other.go
ti-mo: I think this method needs better documentation (something more than a one-liner is often useful for public API). If a process iterates all links on a system, this won't work if the calling process is in another mount namespace (e.g. a container) from the process that initially created the link, since that's typically when the path string is copied to kernel memory.

To me, this method falls more into convenience territory, since the user can accomplish the same using exported APIs. We tend to not include stuff like this, it goes a bit beyond the information returned by the kernel.

@secDre4mer I presume you have an application that does this kind of symbol resolution? Would you be willing to carry this feature locally?

link/uprobe.go
```diff
@@ -181,11 +182,23 @@ func (ex *Executable) address(symbol string, address, offset uint64) (uint64, er
...
+		return address + offset, nil
+	}
+
+	err := ex.lazyLoadSymbols()
 	if err != nil {
 		return 0, fmt.Errorf("lazy load symbols: %w", err)
 	}
 
-	address, ok := ex.cachedAddresses[symbol]
+	symaddress, ok := ex.cachedAddresses[symbol]
```
ti-mo: `symaddress` -> `sym`

link/uprobe.go
```diff
@@ -39,11 +39,16 @@ type Executable struct {
 	// Path of the executable on the filesystem.
 	path string
 	// Parsed ELF and dynamic symbols' cachedAddresses.
-	cachedAddresses map[string]uint64
+	cachedAddresses map[string]symbolAddress
```
ti-mo: I'd call this `cachedSymbols` and `cachedSymbolsOnce`.

link/uprobe.go
```diff
@@ -39,11 +39,16 @@ type Executable struct {
...
 	// Keep track of symbol table lazy load.
 	cacheAddressesOnce sync.Once
 }
 
+type symbolAddress struct {
```
ti-mo: I'd call this `symbol`.

link/uprobe.go
```diff
@@ -196,12 +209,30 @@ func (ex *Executable) address(symbol string, address, offset uint64) (uint64, er
...
+}
+
+// Symbol calculates the symbol and offset for an address in the executable.
+func (ex *Executable) Symbol(address uint64) (UprobeSymbol, error) {
+	if err := ex.lazyLoadSymbols(); err != nil {
+		return UprobeSymbol{}, fmt.Errorf("lazy load symbols: %w", err)
+	}
+
+	for name, symaddress := range ex.cachedAddresses {
+		if symaddress.addr <= address && symaddress.addr+symaddress.size > address {
```
ti-mo: This calculation would be a good candidate to put in its own helper on `symbol`:

```go
// for returns true if addr falls within the range covered by the symbol.
func (s symbol) for(addr uint32) bool { ... }
```

link/link_other.go
```diff
@@ -312,6 +332,19 @@ type UprobeInfo struct {
 	RefCtrOffset uint64
 }
 
+type UprobeSymbol struct {
```
ti-mo: This isn't necessarily specific to uprobe functionality since this is returned from Executable.Symbol() as well. I'd simply name this `Symbol`.

Also, this could use a docstring.

link/link_other.go
```diff
@@ -312,6 +332,19 @@ type UprobeInfo struct {
...
+	Symbol string
+	Offset uint64
+}
+
+func (u *UprobeInfo) Symbol() (UprobeSymbol, error) {
```
ti-mo: This could use a docstring and will fail from mount namespaces other than the one where the link was created.

link/uprobe.go
```diff
@@ -196,12 +209,30 @@ func (ex *Executable) address(symbol string, address, offset uint64) (uint64, er
...
+			return UprobeSymbol{name, address - symaddress.addr}, nil
```
ti-mo: Shouldn't this be `name, sym.address`? A Symbol should always point at the start of a symbol, at least if this function wants to honor its description. If the user wants to attach to an offset within the function, they need to specify that during link creation.

link/uprobe.go
```diff
@@ -196,12 +209,30 @@ func (ex *Executable) address(symbol string, address, offset uint64) (uint64, er
...
 
-	return address + offset, nil
+	if offset >= symaddress.size {
+		return 0, fmt.Errorf("offset %d is out of range of symbol %s", offset, symbol)
+	}
+
+	return symaddress.addr + offset, nil
+}
+
+// Symbol calculates the symbol and offset for an address in the executable.
```
ti-mo: Prefer describing what the method provides to the user. When I read this, I think "ok, it calculates some values, but what's the output?".

Example:

```
// Symbol returns the Symbol the given address points to.
//
// Any address within the range covered by the symbol's instruction
// text will return the same Symbol.
```

secDre4mer:
> Would you be up for removing kp.Address() and kp.Missed() below and exporting the fields instead? Now would be a good time to change this as we're touching surrounding functionality.
@ti-mo Gladly; I noticed that it was inconsistent when implementing the PR and wasn't sure how to work with it.

link/link_other.go
secDre4mer: Arguably, `UprobeSymbol` / `Symbol` is incorrect since this does not describe a symbol, but an offset within a symbol. Do you have a suggestion on how to name this?

link/uprobe.go
secDre4mer: There's a different thread open regarding the naming of `Symbol`; let's discuss this there.

secDre4mer:
> To me, this method falls more into convenience territory, since the user can accomplish the same using exported APIs. We tend to not include stuff like this, it goes a bit beyond the information returned by the kernel.
> I presume you have an application that does this kind of symbol resolution? Would you be willing to carry this feature locally?

Certainly. I understand your point that the `Symbol()`  methods on uprobe / uprobemulti infos are possibly beyond the scope of this library, especially with the caveats (mount namespaces, potentially deleted / replaced files, ...)
Ultimately, it's your call on whether you want to include this. We could also separate this into a different merge request to keep the changes simpler / easier to discuss.
If it's not included, I'd like to keep the "look up a symbol for an offset" logic in `Executable`, though, since it is necessary to do this locally.

link/link_other.go
secDre4mer: Done.

link/uprobe.go
secDre4mer: Done.

link/uprobe.go
secDre4mer: Done.

link/uprobe.go
secDre4mer: Done.

link/uprobe.go
secDre4mer: Done.

link/uprobe.go
secDre4mer: Done.

link/uprobe.go
secDre4mer: Done.

link/uprobe_multi_test.go
secDre4mer: Done.

link/link_other.go
secDre4mer: Done.

link/netfilter.go
secDre4mer: Done.

link/netfilter.go
secDre4mer: Done.

link/perf_event.go
secDre4mer: Done.

link/link_other.go
secDre4mer: No reason; this was a remnant of an older version where adresses had its own, separate method that did not return the cookies. Removed Cookies().

link/link_other.go
secDre4mer: Changed this one; but I think there are other cases that could also benefit from better names. I'll take a look later across the new fields to search for them.

secDre4mer: There are cases, e.g. `Pf` or `Hooknum`, but they already existed before this merge request. Should I rename them as well (at risk of breaking code that uses them)?

ti-mo: Yeah, `Pf` isn't great no matter how you spin it. Let's make those `ProtocolFamily` and `Hook` respectively?

link/link_other.go
ti-mo: Okay, now you've added docstrings it's clear to me what it's supposed to represent. Since it represents an offset within a symbol: `SymbolOffset`? It doesn't necessarily need to be uprobe-specific.

ti-mo:
> Ultimately, it's your call on whether you want to include this. [...] If it's not included, I'd like to keep the "look up a symbol for an offset" logic in `Executable`, though, since it is necessary to do this locally.

I've mulled it over for a bit, I'd prefer not to keep the convenience methods and go for only `Executable.Symbol()`; I think it's more universally useful.

Thanks!

link/link_other.go
secDre4mer: Sounds good, and matches the existing ones in the attach options for `Netfilter`.

link/link_other.go
secDre4mer: Done.

secDre4mer:
> I've mulled it over for a bit, I'd prefer not to keep the convenience methods and go for only `Executable.Symbol()`; I think it's more universally useful.
> 
> Thanks!

Removed the convenience methods.

secDre4mer:
> Looks good, thanks! One of the tests started failing, though.

Thanks, I removed one "is ok" check in the test and that caused older kernels to fail. Added an explicit skip on kernels <6.6 with a comment.

ti-mo (approved):
Thanks!

PR #1898, opened 2025-11-07
asm: add support for the may_goto insn and JCOND opcode

mtardy:
The `may_goto` instruction is a bit special and relatively new, from March 2024, added by 011832b97b31 ("bpf: Introduce may_goto instruction")[^1]. It uses the `BPF_JCOND` opcode `0xe0` with the source register being `BPF_MAY_GOTO = 0`. Note that this instruction is not documented in the latest instruction set kernel documentation[^2].

As I understand, the goal of this instruction is to be injected in loops (also potentially by LLVM) to offload the verifier responsability of loop verification to a runtime check. The runtime check being decrementing a counter so far.

[^1]: https://lore.kernel.org/bpf/20240306031929.42666-2-alexei.starovoitov@gmail.com.
[^2]: https://www.kernel.org/doc/html/v6.17/bpf/standardization/instruction-set.html#jump-instructions.

I realized when writing this PR that I had never used the `asm` package of c/ebpf and that I didn't have the time right now specifically to dive into this. So I'll take that later, or if you want to guide me on what would be the best strategy for you for this "a bit weird" instruction.

### DIFF
--- a/asm/instruction.go
+++ b/asm/instruction.go
@@ -396,9 +396,9 @@ func (ins Instruction) Format(f fmt.State, c rune) {
 		}
 
 	case cls.IsJump():
-		fmt.Fprintf(f, "%v ", op)
 		switch jop := op.JumpOp(); jop {
 		case Call:
+			fmt.Fprintf(f, "%v ", op)
 			switch ins.Src {
 			case PseudoCall:
 				// bpf-to-bpf call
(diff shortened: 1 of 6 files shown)


mtardy:
cc @dylandreimerink

ti-mo:
@mtardy @dylandreimerink What's required to get this merged? Is there a compiler that emits this kind of instruction yet?

dylandreimerink:
I think we just have to double check that tests have been updated to our liking. Any clang version can emit the instruction with some inline asm magic. For some reason the macros have not been added to libbpf and are only in the in-kernel experimental header https://elixir.bootlin.com/linux/v6.19.10/source/tools/testing/selftests/bpf/bpf_experimental.h#L435

But it can already be used. Primary example would be sched_ext, which has its own library with in-BPF datastructure implementations that use this any time they loop. https://github.com/sched-ext/scx/blob/d4dee55a1bb07db6ae74d7ef0cd2bd4338e6b55f/lib/alloc/stack.bpf.c#L155
https://github.com/sched-ext/scx/blob/d4dee55a1bb07db6ae74d7ef0cd2bd4338e6b55f/scheds/include/bpf_arena_common.bpf.h#L97

ti-mo (approved):

PR #1899, opened 2025-11-10
ci: use tagged efW releases

lmb:


lmb:
cc @cilium/ebpf-go-windows-reviewers

PR #1900, opened 2025-11-10
link: Fix misleading error message for fexit/fentry attach

SRodi:
## Problem

When `link.AttachTracing` failed for fexit/fentry programs, users would see misleading error messages:

```
Failed to attach fexit program: create raw tracepoint: device or resource busy
```

This was confusing because:
- The program type is **fexit**, not raw tracepoint
- Users waste time debugging the wrong attach mechanism
- The actual issue (e.g., ftrace disabled) wasn't clear

## Solution

Modified `link/tracing.go` to prevent fexit/fentry/fmod_ret/lsm programs from falling through to the RawTracepointOpen path when LinkCreateTracing fails. 

Now the error clearly shows the actual attach type:

```
Failed to attach fexit program: create TraceFExit tracing link: device or resource busy
```

## Changes

- **link/tracing.go**: Added check to return early for non-raw-tracepoint attach types with proper error message
- **examples/fexit/**: Added minimal example demonstrating the fix and testing error messages

## Testing

The example can be tested on kernels with/without fexit support:

```bash
cd examples/fexit
go build
sudo ./fexit
```

Expected output:
- **With fexit support**: Successfully attaches

```sh
❯ sudo ./fexit
2025/11/10 11:51:47 Successfully attached fexit program to vfs_read
2025/11/10 11:51:47 Monitoring file reads (Ctrl+C to exit)...
2025/11/10 11:51:47 To see output: sudo cat /sys/kernel/debug/tracing/trace_pipe
2025/11/10 11:51:47 
2025/11/10 11:51:47 Try reading a file in another terminal to trigger the fexit program:
2025/11/10 11:51:47   $ cat /etc/hostname
```
- **Without fexit/ftrace**: Clear error showing `TraceFExit tracing link` instead of misleading `raw tracepoint`

```sh
❯ sudo ./fexit
2025/11/10 11:44:34 Failed to attach fexit program: create TraceFExit tracing link: device or resource busy
```

## Impact

- Improves debugging experience for fexit/fentry users
- Makes error messages accurate and actionable
- Backward compatible - no API changes


SRodi:
Closing this PR as the improvement to error messages is negligible. When the caller doesn't explicitly set `AttachType` (e.g., for an fexit program), attachment failures still produce a misleading error message: "create raw tracepoint..".

This happens because `AttachType` is optional. When not provided, it defaults to `ebpf.AttachNone`, which causes the code to fall through to the `RawTracepointOpen` path, resulting in the confusing "raw tracepoint" error message even for non-raw-tracepoint programs.

A better approach might be to dynamically retrieve the program's actual attach type from its metadata, but this would require further discussion to ensure it aligns with the project's design philosophy.

Another option is to preserve and surface the original errno with context from both attachment attempts (BPF link and raw tracepoint fallback). This would:

* Provide clearer diagnostics about which attachment method failed and why
* Allow consumers to make informed decisions in their error handling (e.g., fall back to kprobe, retry with different flags, or choose a different BTF symbol)

PR #1903, opened 2025-11-21
btf: Add type deduplication

dylandreimerink:
This PR introduces a deduplication mechanism for BTF types.

Most of the time BTF types originate from a single spec where a compiler or other external tool has already ensured that types are unique. In such cases we can simply rely on pointer equality to determine if two types are the same.

When dealing with manually generated BTF or merging multiple BTF specs, duplicate types are common. Meaning we have multiple different go objects which represent the same underlying BTF type.

It is useful to be able to deduplicate these types, both to reduce the size of the resulting BTF, as well as allowing name based lookups after combining multiple specs with duplicate types.

The deduplication algorithm is loosely based on the one used in libbpf. This version for example does not do FWD type resolution, as that is only needed when combining BTF from multiple compilation units, which is something typically not seen in eBPF use-cases (only pahole). In the libbpf implementation the first step is string deduplication, however, we do this step during marshaling, and thus we do not deduplicate strings in the Go representation.

When a type is deduplicated, we try to deduplicate not just that root type, but the full subtree of types reachable from that type. We start by traversing all types in post-order, and any time there is an edge we try to replace that child with an equal type we have already seen.

Comparing every type with those seen before would be very expensive. So what we do is we compute a hash of each type. The hash is an approximation of all properties of the type, including recursively hashing child types. When using this hash as key in a map we end up with a set of candidate types which might be equal to the type we are currently deduplicating. We still need to do a full equality check to be sure two types are equal, both to avoid hash collisions as well as to compare properties which are not included in the hash (recursion limit). In practically every case a hash narrows down to 0 or 1 candidate types.

Once we have narrowed down to candidate types, we do a full equality check in which we walk the two types to be compared together in depth first manner, and bail out as soon as we find a difference. Since types can form cycles with pointers, we keep track of already visited types in the current equality check, and assume types are equal.

This deduplication mechanism can be used via a standalone function, but is also integrated in the BTF spec builder via a new method to add and deduplicate a types.

The implementation as it stands performs reasonable for its intended application:
```
goos: linux
goarch: amd64
BenchmarkDeduplicateSKBuff-16     	     100	  11320693 ns/op	 2288514 B/op	     295 allocs/op
BenchmarkDeduplicateVMLinux-16    	       9	 118353647 ns/op	23789357 B/op	   12823 allocs/op
PASS
```
So ~10ms for `struct sk_buff` which has ~5800 types, and ~100ms for vmlinux which has ~93000 types. For smaller types such as map key and value types the cost is obviously way cheaper.

### DIFF
--- a/btf/dedup.go
+++ b/btf/dedup.go
@@ -1,0 +1,633 @@
+package btf
+
+import (
+	"errors"
+	"fmt"
+	"hash/maphash"
+	"slices"
+)
+
+// deduper deduplicates BTF types by finding all types in a Type graph that are
+// Equivalent and replaces them with a single instance.
+//
+// See doc comments in types.go to understand the various ways in which Types
+// can relate to each other and how they are compared for equality. We separate
+// Identity (same memory location), Equivalence (same shape/layout), and
+// Compatibility (CO-RE compatible) to be explicit about intent.
+//
+// This deduper opportunistically uses a combination of Identity and Equivalence
+// to find types that can be deduplicated.
+type deduper struct {
+	visited   map[Type]struct{}
+	hashCache map[hashCacheKey]uint64
+
+	// Set of types that have been deduplicated.
+	done map[Type]Type
+
+	// Map of hash to types with that hash.
+	hashed  map[uint64][]Type
+	eqCache map[typKey]bool
+
+	seed maphash.Seed
+}
+
+func newDeduper() *deduper {
+	return &deduper{
+		make(map[Type]struct{}),
+		make(map[hashCacheKey]uint64),
+		make(map[Type]Type),
+		make(map[uint64][]Type),
+		make(map[typKey]bool),
+		maphash.MakeSeed(),
+	}
+}
+
+func (d *deduper) deduplicate(t Type) (Type, error) {
+	// If we have already attempted to deduplicate this exact type, return the
+	// result.
+	if done, ok := d.done[t]; ok {
+		return done, nil
+	}
+
+	// Visit the subtree, if a type has children, attempt to replace it with a
+	// deduplicated version of those children.
+	for t := range postorder(t, d.visited) {
+		for c := range children(t) {
+			var err error
+			*c, err = d.hashInsert(*c)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	// Finally, deduplicate the root type itself.
+	return d.hashInsert(t)
+}
+
+// hashInsert attempts to deduplicate t by hashing it and comparing against
+// other types with the same hash. Returns the Type to be used as the common
+// substitute at this position in the graph.
+func (d *deduper) hashInsert(t Type) (Type, error) {
+	// If we have deduplicated this type before, return the result of that
+	// deduplication.
+	if done, ok := d.done[t]; ok {
+		return done, nil
+	}
+
+	// Compute the hash of this type. Types with the same hash are candidates for
+	// deduplication.
+	hash, err := d.hash(t, -1)
+	if err != nil {
+		return nil, err
+	}
+
+	// A hash collision is possible, so we need to compare against all candidates
+	// with the same hash.
+	for _, candidate := range d.hashed[hash] {
+		// Pre-size the visited slice, experimentation on VMLinux shows a capacity
+		// of 16 to give the best performance.
+		const visitedCapacity = 16
+		err := d.typesEquivalent(candidate, t, make([]Type, 0, visitedCapacity))
+		if errors.Is(err, errNotEquivalent) {
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		// Found a Type that's both Equivalent and hashes to the same value, choose
+		// it as the deduplicated version.
+		d.done[t] = candidate
+
+		return candidate, nil
+	}
+
+	d.hashed[hash] = append(d.hashed[hash], t)
+
+	return t, nil
+}
+
+// The hash of a Type is the same given its pointer and depth budget.
+type hashCacheKey struct {
+	t           Type
+	depthBudget int
+}
+
+// hash computes a hash for t. The produced hash is the same for Types which
+// are similar. The hash can collide such that two different Types may produce
+// the same hash, so equivalence must be checked explicitly. It will recurse
+// into children. The initial call should use a depthBudget of -1.
+func (d *deduper) hash(t Type, depthBudget int) (uint64, error) {
+	if depthBudget == 0 {
+		return 0, nil
+	}
+
+	h := &maphash.Hash{}
+	h.SetSeed(d.seed)
+
+	switch t := t.(type) {
+	case *Void:
+		maphash.WriteComparable(h, kindUnknown)
+
+	case *Int:
+		maphash.WriteComparable(h, kindInt)
+		maphash.WriteComparable(h, *t)
+
+	case *Pointer:
+		maphash.WriteComparable(h, kindPointer)
+		// If the depth budget is positive, decrement it every time we follow a
+		// pointer.
+		if depthBudget > 0 {
+			depthBudget--
+		}
+
+		// If this is the first time we are following a pointer, set the depth
+		// budget. This limits amount of recursion we do when hashing pointers that
+		// form cycles. This is cheaper than tracking visited types and works
+		// because hash collisions are allowed.
+		if depthBudget < 0 {
+			depthBudget = 1
+
+			// Double pointers are common in C. However, with a depth budget of 1, all
+			// double pointers would hash the same, causing a performance issue when
+			// checking equivalence. So we give double pointers a bit more budget.
+			if _, ok := t.Target.(*Pointer); ok {
+				depthBudget = 2
+			}
+		}
+		sub, err := d.hash(t.Target, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Array:
+		maphash.WriteComparable(h, kindArray)
+		maphash.WriteComparable(h, t.Nelems)
+		sub, err := d.hash(t.Index, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+		_, err = d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Struct, *Union:
+		// Check the cache to avoid recomputing the hash for this type and depth
+		// budget.
+		key := hashCacheKey{t, depthBudget}
+		if cached, ok := d.hashCache[key]; ok {
+			return cached, nil
+		}
+
+		var members []Member
+		switch t := t.(type) {
+		case *Struct:
+			maphash.WriteComparable(h, kindStruct)
+			maphash.WriteComparable(h, t.Name)
+			maphash.WriteComparable(h, t.Size)
+			members = t.Members
+
+		case *Union:
+			maphash.WriteComparable(h, kindUnion)
+			maphash.WriteComparable(h, t.Name)
+			maphash.WriteComparable(h, t.Size)
+			members = t.Members
+		}
+
+		maphash.WriteComparable(h, len(members))
+		for _, m := range members {
+			maphash.WriteComparable(h, m.Name)
+			maphash.WriteComparable(h, m.Offset)
+			sub, err := d.hash(m.Type, depthBudget)
+			if err != nil {
+				return 0, err
+			}
+			maphash.WriteComparable(h, sub)
+		}
+
+		sum := h.Sum64()
+		d.hashCache[key] = sum
+		return sum, nil
+
+	case *Enum:
+		maphash.WriteComparable(h, kindEnum)
+		maphash.WriteComparable(h, t.Name)
+		maphash.WriteComparable(h, t.Size)
+		maphash.WriteComparable(h, t.Signed)
+		for _, v := range t.Values {
+			maphash.WriteComparable(h, v)
+		}
+
+	case *Fwd:
+		maphash.WriteComparable(h, kindForward)
+		maphash.WriteComparable(h, *t)
+
+	case *Typedef:
+		maphash.WriteComparable(h, kindTypedef)
+		maphash.WriteComparable(h, t.Name)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Volatile:
+		maphash.WriteComparable(h, kindVolatile)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Const:
+		maphash.WriteComparable(h, kindConst)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Restrict:
+		maphash.WriteComparable(h, kindRestrict)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Func:
+		maphash.WriteComparable(h, kindFunc)
+		maphash.WriteComparable(h, t.Name)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *FuncProto:
+		// It turns out that pointers to function prototypes are common in C code,
+		// function pointers. Function prototypes frequently have similar patterns
+		// of [ptr, ptr] -> int, or [ptr, ptr, ptr] -> int. Causing frequent hash
+		// collisions, for the default depth budget of 1. So allow one additional
+		// level of pointers when we encounter a function prototype.
+		if depthBudget >= 0 {
+			depthBudget++
+		}
+
+		maphash.WriteComparable(h, kindFuncProto)
+		for _, p := range t.Params {
+			maphash.WriteComparable(h, p.Name)
+			sub, err := d.hash(p.Type, depthBudget)
+			if err != nil {
+				return 0, err
+			}
+			maphash.WriteComparable(h, sub)
+		}
+		sub, err := d.hash(t.Return, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Var:
+		maphash.WriteComparable(h, kindVar)
+		maphash.WriteComparable(h, t.Name)
+		maphash.WriteComparable(h, t.Linkage)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Datasec:
+		maphash.WriteComparable(h, kindDatasec)
+		maphash.WriteComparable(h, t.Name)
+		for _, v := range t.Vars {
+			maphash.WriteComparable(h, v.Offset)
+			maphash.WriteComparable(h, v.Size)
+			sub, err := d.hash(v.Type, depthBudget)
+			if err != nil {
+				return 0, err
+			}
+			maphash.WriteComparable(h, sub)
+		}
+
+	case *declTag:
+		maphash.WriteComparable(h, kindDeclTag)
+		maphash.WriteComparable(h, t.Value)
+		maphash.WriteComparable(h, t.Index)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *TypeTag:
+		maphash.WriteComparable(h, kindTypeTag)
+		maphash.WriteComparable(h, t.Value)
+		sub, err := d.hash(t.Type, depthBudget)
+		if err != nil {
+			return 0, err
+		}
+		maphash.WriteComparable(h, sub)
+
+	case *Float:
+		maphash.WriteComparable(h, kindFloat)
+		maphash.WriteComparable(h, *t)
+
+	default:
+		return 0, fmt.Errorf("unsupported type for hashing: %T", t)
+	}
+
+	return h.Sum64(), nil
+}
+
+type typKey struct {
+	a Type
+	b Type
+}
+
+var errNotEquivalent = errors.New("types are not equivalent")
+
+// typesEquivalent checks if two types are Equivalent.
+func (d *deduper) typesEquivalent(ta, tb Type, visited []Type) error {
+	// Fast path: if Types are Identical, they are also Equivalent.
+	if ta == tb {
+		return nil
+	}
+
+	switch a := ta.(type) {
+	case *Void:
+		if _, ok := tb.(*Void); ok {
+			return nil
+		}
+		return errNotEquivalent
+
+	case *Int:
+		b, ok := tb.(*Int)
+		if !ok {
+			return errNotEquivalent
+		}
+		if a.Name != b.Name || a.Size != b.Size || a.Encoding != b.Encoding {
+			return errNotEquivalent
+		}
+		return nil
+
+	case *Enum:
+		b, ok := tb.(*Enum)
+		if !ok {
+			return errNotEquivalent
+		}
+		if a.Name != b.Name || len(a.Values) != len(b.Values) {
+			return errNotEquivalent
+		}
+		for i := range a.Values {
+			if a.Values[i].Name != b.Values[i].Name || a.Values[i].Value != b.Values[i].Value {
+				return errNotEquivalent
+			}
+		}
+		return nil
+
+	case *Fwd:
+		b, ok := tb.(*Fwd)
+		if !ok {
+			return errNotEquivalent
+		}
+		if a.Name != b.Name || a.Kind != b.Kind {
+			return errNotEquivalent
+		}
+		return nil
+
+	case *Float:
+		b, ok := tb.(*Float)
+		if !ok {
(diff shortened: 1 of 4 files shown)


btf/dedup.go
```diff
@@ -0,0 +1,457 @@
+package btf
+
+import (
+	"fmt"
+	"hash/maphash"
+	"slices"
+)
+
+func Deduplicate(roots []Type) ([]Type, error) {
```
lmb: Please add a doc comment.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+		}
+	}
+
+	dm.candidates[hash] = append(dm.candidates[hash], t)
+	return t
+}
+
+// maphash.Hash by default uses a random seed per Hash instance.
+// To ensure consistent hashing across multiple calls, we use a fixed seed for the lifetime of the program.
+var seed = maphash.MakeSeed()
```
lmb: I think we should store `maphash.Hash` and `Seed` in deduper instead. That avoids temporarily allocating `Hash` (which is big-ish) and gets rid of global state, which is always a bit of a smell.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+	t           Type
+	depthBudget int
+}
+
+// typeHash computes a hash for `t`. The produced hash is the same for types which are similar.
+// The hash can collide such that two different types may produce the same hash, so equality must be checked separately.
+// It will recursively call itself to hash child types. The initial call should use a depthBudget of -1.
+func typeHash(t Type, depthBudget int, cache map[hashCacheKey]uint64) (sum uint64) {
+	if depthBudget == 0 {
+		return 0
```
lmb: So this means that if a type is nested too deeply we use 0 as the hash? Seems unintuitive to give 0 that meaning, especially because it means that the 0 value then potentially has a large number of collisions in a pathological case.

Could this return an error instead?

If this needs to stay then the doc comment should mention this.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+// To ensure consistent hashing across multiple calls, we use a fixed seed for the lifetime of the program.
+var seed = maphash.MakeSeed()
+
+// The hash of a type is the same given the type and depth budget, and thus is the key for the cache.
+type hashCacheKey struct {
+	t           Type
+	depthBudget int
+}
+
+// typeHash computes a hash for `t`. The produced hash is the same for types which are similar.
```
lmb: What does "similar" mean here? equivalent?

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+		return 0
+	}
+
+	// Check the cache to avoid recomputing the hash for this type and depth budget.
+	key := hashCacheKey{t, depthBudget}
+	if cached, ok := cache[key]; ok {
+		return cached
+	}
+	defer func() {
+		cache[key] = sum
```
lmb: Nit: neat but brittle due to variable shadowing. I'd probably just do this explicitly at the bottom of the method, since there is only a single return statement.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+	}()
+
+	var hash maphash.Hash
+	h := &hash
+	h.SetSeed(seed)
+
+	switch t := t.(type) {
```
lmb, line 129: One problem with this approach is that it is very easy to forget to implement this for new types. Could we instead do something like the following?

```go
type Type interface{
	// ...

	// hash all comparable fields of the type, excluding any [Type]s.
	hash(*maphash.Hash)
}

func hashType(h *Hash, t Type, budget int) {
	if budget < 0 {
		return
	}

	// Write all non-Type fields to h. Since we don't deal with Type there is no recursion.
	t.hash(h)

	for _, c := range children(t) {
		hashType(h, *c, budget-1)
	}
}

func typeHash(h *Hash, s *Seed t *Type) (uint64) {
	h.SetSeed(s)
	hashType(h, t, N)
	return h.Sum64()
}
```

(Assumes that caching of typeHash happens in the caller.)

Another upside is that this could let us add a test in `testType` or whatever that is called which asserts some behaviour of `hash`.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+	case *Array:
+		maphash.WriteComparable(h, kindArray)
+		maphash.WriteComparable(h, t.Nelems)
+		maphash.WriteComparable(h, typeHash(t.Index, depthBudget, cache))
+		maphash.WriteComparable(h, typeHash(t.Type, depthBudget, cache))
+	case *Struct:
+		maphash.WriteComparable(h, kindStruct)
+		maphash.WriteComparable(h, t.Name)
+		maphash.WriteComparable(h, t.Size)
+		for _, m := range t.Members {
```
lmb: Does it make sense to include `len(t.Members)` in the hash?

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+	return h.Sum64()
+}
+
+type typKey struct {
+	a Type
+	b Type
+}
+
+// typesEqual checks if two types are equal, recursively checking child types as needed.
+func typesEqual(aTyp, b Type, visited []Type, cache map[typKey]bool) bool {
```
lmb: In the code base "equal" so far means Go equality rules. Can we call this `typesEquivalent` or similar instead?

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+	case *Enum:
+		maphash.WriteComparable(h, kindEnum)
+		maphash.WriteComparable(h, t.Name)
+		for _, v := range t.Values {
+			maphash.WriteComparable(h, v.Name)
+			maphash.WriteComparable(h, v.Value)
+		}
+	case *Fwd:
+		maphash.WriteComparable(h, kindForward)
+		maphash.WriteComparable(h, t.Name)
```
lmb: This is equivalent and probably faster:

```suggestion
		maphash.WriteComparable(h, t)
```

This might even mean that `WriteComparable(h, kindForward)` is not necessary. Same for Void, Int, Float, etc.

btf/marshal.go
```diff
@@ -90,6 +92,7 @@ func NewBuilder(types []Type) (*Builder, error) {
 		make([]Type, 0, len(types)),
 		make(map[Type]TypeID, len(types)),
 		nil,
+		newDeduper(),
```
lmb: Should this be nil since AddUnique checks for nilness?

btf/marshal.go
```diff
@@ -144,6 +147,22 @@ func (b *Builder) Add(typ Type) (TypeID, error) {
 	return id, nil
 }
 
+// AddUnique a Type, allocate a stable ID for it, and deduplicate it.
+//
+// Added types and all types reachable from it are deduplicated based on
+// structural equality, not identity.
+//
+// Adding the equal Type multiple times is valid and will return the same ID.
+func (b *Builder) AddUnique(typ Type) (TypeID, error) {
```
lmb: How does AddUnique and Add interact? For example:

```
var a, b, c Type
b.Target = &a
builder.Add(a)
builder.AddUnique(b)
builder.Add(c)
```

My gut feeling is that it's quite complicated to explain what this does. Would it make sense to turn this into a toggle on the builder instead? `Builder.EnableDeduplication()` or similar, which has to be called before the first add.

btf/dedup.go
lmb: Is this even used at the moment? Maybe just remove it outright if the use case is via the builder.

btf/dedup.go
lmb: P.S. It took me a long time to figure out what this really does: produce a hash of the first couple of levels of indirection, as an approximation of equivalence. This is a really neat idea, I wonder whether we can simplify it even more.

Right now depth is changed dynamically based on what kind of type we are looking at. Sometimes we use depth 1, sometimes 2. This trades off how much work we do. At the same time it avoids infinite recursion for self referential types. I think instead of having some dynamic level we could always use a fixed budget. Basically make this the hash of the first N levels of indirection. If a type is self referential we just end up hashing the same data up to N times, which isn't a problem as long as N is something reasonably small.

This has the nice side effect that hash(T) now doesn't change anymore, and the cache can become `map[Type]uint64`. We might also be able to move the cache into the caller. Plus the code is a bit easier to understand.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+// typeHash computes a hash for `t`. The produced hash is the same for types which are similar.
+// The hash can collide such that two different types may produce the same hash, so equality must be checked separately.
+// It will recursively call itself to hash child types. The initial call should use a depthBudget of -1.
+func typeHash(t Type, depthBudget int, cache map[hashCacheKey]uint64) (sum uint64) {
+	if depthBudget == 0 {
+		return 0
+	}
+
+	// Check the cache to avoid recomputing the hash for this type and depth budget.
+	key := hashCacheKey{t, depthBudget}
```
lmb: Does it really pay off to put all types into the cache? Intuitively I'd expected we'd only cache composite types.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+		for c := range children(t) {
+			*c = dm.deduplicateSingle(*c)
+		}
+	}
+
+	// Finally, deduplicate the root type itself.
+	return dm.deduplicateSingle(t)
+}
+
+func (dm *deduper) deduplicateSingle(t Type) Type {
```
lmb: Nit: I think you can merge Deduplicate and deduplicateSingle?

btf/dedup.go
lmb: And please update the doc comment on `Type` to explain what the term that you settle on means.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+}
+
+// typesEqual checks if two types are equal, recursively checking child types as needed.
+func typesEqual(aTyp, b Type, visited []Type, cache map[typKey]bool) bool {
+	// Fast path: do a pointer comparison, if we have the same object, they are equal.
+	if aTyp == b {
+		return true
+	}
+
+	switch a := aTyp.(type) {
```
lmb: Similar feelings to this switch as the one above, it's very easy to forget to implement things. Not sure how to fix this though. The core code looks veeery similar, I'll look at that and maybe I can find a common abstraction.

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+			typesEqual(a.Type, b.Type, visited, cache)
+	case *Pointer:
+		b, ok := b.(*Pointer)
+		if !ok {
+			return false
+		}
+
+		// Detect cycles by tracking visited types. Assume types are equal if we have already
+		// visited this type in the current equality check.
+		if slices.Contains(visited, aTyp) {
```
lmb: This is going to get quite expensive pretty fast, no?

btf/dedup.go
```diff
@@ -0,0 +1,457 @@
...
+			typesEqual(a.Type, b.Type, visited, cache)
+	case *TypeTag:
+		b, ok := b.(*TypeTag)
+		if !ok {
+			return false
+		}
+		return a.Value == b.Value &&
+			typesEqual(a.Type, b.Type, visited, cache)
+	default:
+		panic(fmt.Sprintf("unsupported type for equality: %T", a))
```
lmb: Not sure how I feel about this panic as well.

btf/dedup.go
dylandreimerink: Yes, as you might imagine this is where I started, a static budget. The issue is that in select cases, specifically double points and function prototypes you often end up with hash collisions since the first levels of these are often similar, causing a lot of work in equality checks. Increasing the budget overall fixes that, but it also makes the hash way more expensive, canceling out the benefit. Which is why I settled on the current state, adding depth where it reduces collisions going from ~100 (in double pointers and func protos) to <10 collisions on all types in vmlinux.

So we can simplify at the cost of some performance, I can run the numbers so we can quantify.

btf/dedup.go
dylandreimerink: I think we can get rid of it, minimize the API surface. Its was useful while testings, but we can just call the unexported function to do that.

btf/marshal.go
dylandreimerink: Yea, its tricky. Only types presented to the deduper are known. So if you `.Add(typeA)` and then do `.AddUnique(typeB)` where `typeB` contains something equivalent to `typeA`, you will still end up with 2 equivalent types.

In practice there would be no reason to mix them. I did not want to add options to `NewBuilder` since it would break the API. But doing it via a method like this seems reasonable.

btf/dedup.go
dylandreimerink: Yes, I believe it does, I recall testing just composites and this was better. I will double check and post the numbers.

btf/dedup.go
dylandreimerink: Its really depends. Turns out CPUs are quite fast at checking this when its in contiguous memory, since its a predictable access pattern and cache lines ect. So for small-ish slices, its faster than a map, both the lookup and writing. Doing this only for pointers and not all types really helps. I can also double check and re-measure.

btf/dedup.go
dylandreimerink: Here are the numbers when we swap out the dynamic depth limit for a static limit of 1, 2 or 3:
```
goos: linux
goarch: amd64
                      │ dynamic-depth.txt │             depth-1.txt              │             depth-2.txt              │             depth-3.txt              │
                      │      sec/op       │    sec/op     vs base                │    sec/op     vs base                │    sec/op     vs base                │
DeduplicateSKBuff-20          8.290m ± 4%   10.035m ± 5%  +21.04% (p=0.000 n=10)   9.624m ± 10%  +16.08% (p=0.000 n=10)   12.523m ± 5%  +51.05% (p=0.000 n=10)
DeduplicateVMLinux-20         95.49m ± 4%   123.51m ± 3%  +29.35% (p=0.000 n=10)   96.37m ± 10%        ~ (p=0.315 n=10)    98.56m ± 2%   +3.22% (p=0.004 n=10)
geomean                       28.14m         35.21m       +25.12%                  30.45m         +8.24%                   35.13m       +24.86%

                      │ dynamic-depth.txt │             depth-1.txt              │              depth-2.txt              │              depth-3.txt              │
                      │       B/op        │     B/op       vs base               │     B/op       vs base                │     B/op       vs base                │
DeduplicateSKBuff-20        2.505Mi ±  4%   2.559Mi ± 17%  +2.17% (p=0.014 n=10)   2.804Mi ± 19%  +11.92% (p=0.000 n=10)   3.770Mi ± 17%  +50.48% (p=0.000 n=10)
DeduplicateVMLinux-20       29.06Mi ± 10%   29.74Mi ±  0%       ~ (p=0.089 n=10)   30.99Mi ±  7%   +6.63% (p=0.014 n=10)   31.99Mi ±  3%  +10.09% (p=0.007 n=10)
geomean                     8.533Mi         8.724Mi        +2.24%                  9.321Mi         +9.24%                  10.98Mi        +28.71%

                      │ dynamic-depth.txt │             depth-1.txt              │             depth-2.txt              │             depth-3.txt             │
                      │     allocs/op     │  allocs/op    vs base                │  allocs/op    vs base                │  allocs/op   vs base                │
DeduplicateSKBuff-20          298.0 ±  3%     320.0 ± 3%   +7.38% (p=0.000 n=10)    325.0 ± 12%   +9.06% (p=0.000 n=10)    387.0 ± 8%  +29.87% (p=0.000 n=10)
DeduplicateVMLinux-20        7.081k ± 10%   12.128k ± 0%  +71.28% (p=0.000 n=10)   8.315k ±  7%  +17.43% (p=0.000 n=10)   7.968k ± 4%  +12.53% (p=0.001 n=10)
geomean                      1.453k          1.970k       +35.62%                  1.644k        +13.17%                  1.756k       +20.89%
```

It seems a static depth of 2 is the best alternative at ~10% slower.

btf/dedup.go
dylandreimerink: Turns out just caching composites is better, so will do that instead:
```
goos: linux
goarch: amd64
                      │ cache-all.txt │        cache-composites.txt         │
                      │    sec/op     │   sec/op     vs base                │
DeduplicateSKBuff-20      8.322m ± 1%   8.045m ± 0%   -3.33% (p=0.000 n=30)
DeduplicateVMLinux-20     98.95m ± 2%   85.39m ± 1%  -13.71% (p=0.000 n=30)
geomean                   28.70m        26.21m        -8.67%

                      │ cache-all.txt │         cache-composites.txt         │
                      │     B/op      │     B/op      vs base                │
DeduplicateSKBuff-20     2.673Mi ± 1%   1.725Mi ± 1%  -35.46% (p=0.000 n=30)
DeduplicateVMLinux-20    30.98Mi ± 6%   22.27Mi ± 3%  -28.12% (p=0.000 n=30)
geomean                  9.100Mi        6.198Mi       -31.89%

                      │ cache-all.txt │        cache-composites.txt         │
                      │   allocs/op   │  allocs/op   vs base                │
DeduplicateSKBuff-20       311.0 ± 0%    260.0 ± 1%  -16.40% (p=0.000 n=30)
DeduplicateVMLinux-20     8.314k ± 6%   8.097k ± 3%        ~ (p=0.059 n=30)
geomean                   1.608k        1.451k        -9.77%
```

btf/dedup.go
dylandreimerink: I have attempted this change, but it significantly impacts the amount of allocations and thus performance, even after reusing the `maphash.Hash`. So I would prefer to keep it as is, even if that means being more careful when adding new `Type` implementations in the future.

```
goos: linux
goarch: amd64
                      │ before.txt  │              after.txt               │
                      │   sec/op    │    sec/op     vs base                │
DeduplicateSKBuff-20    8.786m ± 4%    9.761m ± 1%  +11.09% (p=0.000 n=30)
DeduplicateVMLinux-20   96.55m ± 6%   106.38m ± 3%  +10.18% (p=0.000 n=30)
geomean                 29.13m         32.22m       +10.64%

                      │  before.txt  │               after.txt               │
                      │     B/op     │     B/op       vs base                │
DeduplicateSKBuff-20    1.783Mi ± 3%   2.312Mi ±  0%  +29.67% (p=0.000 n=30)
DeduplicateVMLinux-20   24.07Mi ± 4%   31.54Mi ± 14%  +31.02% (p=0.000 n=30)
geomean                 6.551Mi        8.539Mi        +30.35%

                      │ before.txt  │                after.txt                 │
                      │  allocs/op  │   allocs/op    vs base                   │
DeduplicateSKBuff-20     265.5 ± 2%    37996.5 ± 0%  +14211.30% (p=0.000 n=30)
DeduplicateVMLinux-20   8.949k ± 8%   473.048k ± 1%   +5186.04% (p=0.000 n=30)
geomean                 1.541k          134.1k        +8597.71%
```

[hash-per-type.patch](https://github.com/user-attachments/files/23853215/hash-per-type.patch)

btf/dedup.go
dylandreimerink: I made the seed part of the deduper. It turns out that having the `maphash.Hash` where it is is actually beneficial. It allocated on the stack since its never passed, which means that its cheap to allocate and dispose of. So I kept it as is for now.

btf/dedup.go
dylandreimerink: Measured it, slices are way better in this instance:

```
goos: linux
goarch: amd64
                      │ before.txt  │              after.txt               │
                      │   sec/op    │    sec/op     vs base                │
DeduplicateSKBuff-20    8.178m ± 1%   10.184m ± 4%  +24.53% (p=0.000 n=10)
DeduplicateVMLinux-20   88.88m ± 2%   119.56m ± 2%  +34.53% (p=0.000 n=10)
geomean                 26.96m         34.90m       +29.43%

                      │  before.txt  │               after.txt                │
                      │     B/op     │     B/op       vs base                 │
DeduplicateSKBuff-20    1.692Mi ± 1%    7.364Mi ± 4%  +335.17% (p=0.000 n=10)
DeduplicateVMLinux-20   22.93Mi ± 6%   104.92Mi ± 4%  +357.48% (p=0.000 n=10)
geomean                 6.230Mi         27.80Mi       +346.18%

                      │ before.txt  │                after.txt                │
                      │  allocs/op  │   allocs/op    vs base                  │
DeduplicateSKBuff-20     257.0 ± 1%    17597.0 ± 0%  +6747.08% (p=0.000 n=10)
DeduplicateVMLinux-20   8.360k ± 6%   277.394k ± 0%  +3218.11% (p=0.000 n=10)
geomean                 1.466k          69.87k       +4666.48%
```

btf/dedup.go
dylandreimerink: Yea, agreed. I think equivalence/equivalent is a good term to avoid confusion. What i settled on is "Identical" to describe two `Type`s at the same memory address. They get the same ID when marshaled, and would be checked with Go pointer equality.

When two `Type`s have the same shape, but are distinct objects we call them "equivalent", having the "equivalence" property.

And I also decided to mention "compatibility" in the CO-RE sense, to make it clear that this is a different property altogether. Please double check to see if you are happy with the Doc changes.

btf/dedup.go
dylandreimerink: Not really, `Deduplicate` does the traversal, and `deduplicateSingle` does not recurse. And since we have two callsites I can't neatly inline it without duplicating, this is nicer I think.

btf/types.go
```diff
@@ -21,15 +21,25 @@ type TypeID = sys.TypeID
...
 //
 //	a := &Int{Size: 1}
 //	b := &Int{Size: 1}
 //	a != b
 //
+// Equivalence: two Types are considered equivalent if they have the same shape
+// and thus are functionally interchangeable, even if they are located at different
+// memory addresses. The above two Int types are equivalent.
+//
+// Compatibility: two Types are considered compatible if code compiled against one
```
lmb: "code compiled against" is a bit vague. Maybe just

```suggestion
// Compatibility: two Types are considered compatible according to the rules of CO-RE.
```

lmb (changes requested):
LGTM thanks!

btf/marshal.go
```diff
@@ -107,6 +110,19 @@ func (b *Builder) Empty() bool {
 	return len(b.types) == 0 && (b.strings == nil || b.strings.Length() == 0)
 }
 
+func (b *Builder) EnableDeduplication() error {
```
ti-mo, line 116: This is missing a doc comment!

btf/dedup.go
dylandreimerink: Yea, there is two sides of it. First, if we ever hit this panic, then its definitely unexpected and a bug in the library. The idea is that test coverage _should_ catch it if it ever were to happen.

On the other hand, if it ever does slip through CI, then throwing a panic as library is not cool.

What we could do is to change the return type of `typesEquivalent` to be an err, where nil means equivalent, we have known error types for not equivalent, and return a different error if we hit the default here. And we can bubble the error up to `Deduplicate` and `Builder.Add`. This would result in a failure more that a user can handle if need be.

We also have a panic in `children` if we forget to add a new type there, so its not without precedent.

btf/dedup.go
```diff
@@ -0,0 +1,439 @@
...
+		make(map[Type]struct{}),
+		make(map[hashCacheKey]uint64),
+		make(map[Type]Type),
+		make(map[uint64][]Type),
+		make(map[typKey]bool),
+		maphash.MakeSeed(),
+	}
+}
+
+func (dm *deduper) Deduplicate(t Type) Type {
```
ti-mo: Curious, why is this method exported? I don't think it needs to be, maybe left over from a previous iteration.

btf/dedup.go
ti-mo: > What we could do is to change the return type of typesEquivalent to be an err, where nil means equivalent, we have known error types for not equivalent, and return a different error if we hit the default here.

That approach makes sense to me.

> We also have a panic in `children` if we forget to add a new type there, so its not without precedent.

Indeed, and that function also potentially takes arbitrary user input. It returns an iterator, so there's nothing we can do but panic to signal errors.

Users can't implement their own types, so there's little to gain from turning these panics into errors. (aside from code that's harder to read) I'd say this is fine.

ti-mo (approved):
Thanks for the hard work! This is awesome.

dylandreimerink:
Did another pass. The deduplicate method is now unexported. And missing types in switch statements now bubble up errors instead of panic'ing.

ti-mo:
I'll push up some changes for bikeshedding and readability, but I also found a large perf regression in the bench code:
```diff
@@ -131,8 +130,9 @@ func BenchmarkDeduplicateVMLinux(b *testing.B) {

        for i := 0; i < b.N; i++ {
                for _, typ := range types[i] {
-                       _, err = dedup.deduplicate(typ)
-                       qt.Assert(b, qt.IsNil(err))
+                       if _, err := dedup.deduplicate(typ); err != nil {
+                               b.Fatal(err)
+                       }
                }
        }
 }
```

```
goos: linux
goarch: amd64
                      │   sec/op    │    sec/op     vs base               │
DeduplicateSKBuff-16    10.58m ± 3%   10.11m ±  6%        ~ (p=0.065 n=6)
DeduplicateVMLinux-16   181.6m ± 3%   115.7m ± 21%  -36.26% (p=0.002 n=6)
geomean                 43.84m        34.21m        -21.97%

                      │   old.txt    │               new.txt                │
                      │     B/op     │     B/op       vs base               │
DeduplicateSKBuff-16    2.200Mi ± 0%   2.200Mi ± 13%   -0.01% (p=0.002 n=6)
DeduplicateVMLinux-16   43.28Mi ± 0%   20.66Mi ± 27%  -52.28% (p=0.002 n=6)
geomean                 9.758Mi        6.741Mi        -30.92%

                      │   old.txt    │               new.txt               │
                      │  allocs/op   │  allocs/op    vs base               │
DeduplicateSKBuff-16      301.0 ± 0%    296.0 ±  9%   -1.66% (p=0.002 n=6)
DeduplicateVMLinux-16   495.67k ± 0%   11.98k ± 11%  -97.58% (p=0.002 n=6)
geomean                  12.21k        1.883k        -84.59%

```

qt.Assert always allocates, so let's not use it in the hot path. I haven't optimized the deduper itself due to time constraints, there are likely some more gains on the table wrt. allocations since there are so many layers of caches, but this is a nice baseline. vmlinux dedupes in 180ms on my machine, that's nice!

PR #1906, opened 2025-11-24
features: export link features

ajwerner:
The use of multi-attach for uprobes/kprobes requires both loading the program with the appropriate attach type, and then using the correct call to do the attach. Sometimes the program loading is not tightly coupled to the attach so it's painful to have to fail later and then re-load the program. More generally, forcing consumers of this package to track their own feature availability on top of which things return unsupported errors is bad -- we should make most feature probing available and exported to the extent possible.

This PR exports from `features`:
 * `HaveBPFLinkUprobeMulti`
 * `HaveBPFLinkKprobeMulti`
 * `HaveBPFLinkKprobeSession`

### DIFF
--- a/features/link.go
+++ b/features/link.go
@@ -1,0 +1,157 @@
+package features
+
+import (
+	"errors"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/internal"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+// HaveBPFLinkUprobeMulti probes the running kernel if uprobe_multi link is supported.
+//
+// See the package documentation for the meaning of the error return value.
+func HaveBPFLinkUprobeMulti() error {
+	return haveBPFLinkUprobeMulti()
+}
+
+var haveBPFLinkUprobeMulti = internal.NewFeatureTest("bpf_link_uprobe_multi", func() error {
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Name: "probe_upm_link",
+		Type: ebpf.Kprobe,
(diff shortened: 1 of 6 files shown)


ajwerner:
@mmat11 done

features/link.go
```diff
@@ -0,0 +1,157 @@
...
+		Instructions: asm.Instructions{
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+		AttachType: ebpf.AttachTraceUprobeMulti,
+		License:    "MIT",
+	})
+	if errors.Is(err, unix.E2BIG) {
+		// Kernel doesn't support AttachType field.
+		return internal.ErrNotSupported
```
mmat11: just noticed these functions return `internal.ErrNotSupported` -- I'm not 100% sure, they should probably return `ebpf.ErrNotSupported` /cc @rgo3

ajwerner: it's the same thing, right?

ajwerner: To make that extra concrete: https://github.com/cilium/ebpf/blob/v0.20.0/prog.go#L26

ajwerner: Updated to use `ebpf.ErrNotSupported` so its symmetric with the rest of the package.

mmat11 (approved):
lgtm!

ti-mo:
Thanks for the patch!

> More generally, forcing consumers of this package to track their own feature availability on top of which things return unsupported errors is bad -- we should make most feature probing available and exported to the extent possible.

@ajwerner If you're talking about the feature probes in the root `ebpf` package, those cannot be moved over to `features` due to dependency cycles. (they all load progs or maps using package `ebpf`) The ones you moved in this PR were in package link, so no problem. The ones from the root package would need to be duplicated, which we'd rather avoid.

ti-mo (approved):

PR #1913, opened 2025-11-27
test: Separate the netns for netkitanchor

### LINKED ISSUE #1577: Run netkit test in separate network namespace
This blew up on my machine today:
```
--- FAIL: TestNetkitAnchor (0.00s)
    quicktest.go:12:
        error:
          got non-nil value
        got:
          e"netlink receive: file exists"
        stack:
          /home/timo/github/cilium/ebpf/link/netkit_test.go:95
            qt.Assert(tb, qt.IsNil(err))
          /home/timo/github/cilium/ebpf/link/netkit_test.go:32
            linkA, ifIndex := mustAttachNetkit(t, a, ebpf.AttachNetkitPrimary)

FAIL
```

This test should be running in a new netns so it doesn't interfere with the host's network stack.

Jack-R-lantern:
Separated the network namespace for netkitanchor

close #1577

### DIFF
--- a/link/netkit_test.go
+++ b/link/netkit_test.go
@@ -19,8 +19,10 @@ import (
 func TestAttachNetkit(t *testing.T) {
 	testutils.SkipOnOldKernel(t, "6.7", "Netkit Device")
 
+	ns := testutils.NewNetNS(t)
+
 	prog := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
-	link, _ := mustAttachNetkit(t, prog, ebpf.AttachNetkitPrimary)
+	link, _ := mustAttachNetkit(t, prog, ebpf.AttachNetkitPrimary, ns)
 
 	testLink(t, link, prog)
 }
@@ -31,7 +33,9 @@ func TestNetkitAnchor(t *testing.T) {
 	a := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
 	b := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
 
-	linkA, ifIndex := mustAttachNetkit(t, a, ebpf.AttachNetkitPrimary)
+	ns := testutils.NewNetNS(t)
+
+	linkA, ifIndex := mustAttachNetkit(t, a, ebpf.AttachNetkitPrimary, ns)
 
 	programInfo, err := a.Info()
 	qt.Assert(t, qt.IsNil(err))
@@ -50,13 +54,16 @@ func TestNetkitAnchor(t *testing.T) {
 		AfterLinkByID(linkID),
 	} {
 		t.Run(fmt.Sprintf("%T", anchor), func(t *testing.T) {
-			linkB, err := AttachNetkit(NetkitOptions{
-				Program:   b,
-				Attach:    ebpf.AttachNetkitPrimary,
-				Interface: ifIndex,
-				Anchor:    anchor,
-			})
-			qt.Assert(t, qt.IsNil(err))
+			var linkB Link
+			qt.Assert(t, qt.IsNil(ns.Do(func() (err error) {
+				linkB, err = AttachNetkit(NetkitOptions{
+					Program:   b,
+					Attach:    ebpf.AttachNetkitPrimary,
+					Interface: ifIndex,
+					Anchor:    anchor,
+				})
+				return err
+			})))
 			qt.Assert(t, qt.IsNil(linkB.Close()))
 		})
 	}
@@ -67,10 +74,12 @@ var prevIfindex atomic.Uint32
 
 func init() { prevIfindex.Store(1000 - 1) }
 
-func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachType) (Link, int) {
-	var err error
-	conn, err := rtnetlink.Dial(nil)
-	qt.Assert(tb, qt.IsNil(err))
+func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachType, ns *testutils.NetNS) (Link, int) {
+	var conn *rtnetlink.Conn
+	qt.Assert(tb, qt.IsNil(ns.Do(func() (err error) {
+		conn, err = rtnetlink.Dial(nil)
+		return err
+	})))
 	tb.Cleanup(func() {
 		qt.Assert(tb, qt.IsNil(conn.Close()))
 	})
@@ -79,7 +88,7 @@ func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachT
 
 	layer2 := driver.NetkitModeL2
 	blackhole := driver.NetkitPolicyDrop
-	err = conn.Link.New(&rtnetlink.LinkMessage{
+	qt.Assert(tb, qt.IsNil(conn.Link.New(&rtnetlink.LinkMessage{
 		Family: unix.AF_UNSPEC,
 		Index:  ifIndex,
 		Flags:  unix.IFF_UP,
@@ -93,18 +102,20 @@ func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachT
 				},
 			},
 		},
-	})
-	qt.Assert(tb, qt.IsNil(err))
+	})))
 	tb.Cleanup(func() {
 		qt.Assert(tb, qt.IsNil(conn.Link.Delete(uint32(ifIndex))))
 	})
 
-	link, err := AttachNetkit(NetkitOptions{
-		Program:   prog,
-		Attach:    attachType,
-		Interface: int(ifIndex),
-	})
-	qt.Assert(tb, qt.IsNil(err))
+	var link Link
+	qt.Assert(tb, qt.IsNil(ns.Do(func() (err error) {
+		link, err = AttachNetkit(NetkitOptions{
+			Program:   prog,
+			Attach:    attachType,
+			Interface: int(ifIndex),
+		})
+		return err
+	})))
 	tb.Cleanup(func() {
 		qt.Assert(tb, qt.IsNil(link.Close()))
 	})


link/netkit_test.go
```diff
@@ -4,12 +4,14 @@ package link
...
 import (
 	"fmt"
+	"runtime"
 	"sync/atomic"
 	"testing"
 
 	"github.com/go-quicktest/qt"
 	"github.com/jsimonetti/rtnetlink/v2"
 	"github.com/jsimonetti/rtnetlink/v2/driver"
+	"github.com/vishvananda/netns"
```
florianl: Instead of introducing a new dependency, I would prefer to follow the approach as in https://github.com/mdlayher/netlink/pull/110 - this would result in a bit more code, but I think it is worth it.

Jack-R-lantern: Could you share what additional benefits this approach provides beyond removing the dependency on the package?

My initial intention was to minimize the amount of utility code needed for testing, so I opted to use a new package.
Based on your feedback, I’ve reconsidered this, and I agree that if we plan to isolate network namespaces for all tests in the future, having our own utility functions could offer significant long-term advantages.

link/netkit_test.go
```diff
@@ -31,6 +33,26 @@ func TestNetkitAnchor(t *testing.T) {
 	a := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
 	b := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
 
+	runtime.LockOSThread()
```
ti-mo: This is an awful amount of boilerplate for a test, and quite fragile. Flo gave you a good suggestion, but the mdl/netlink netns code is also 7 years old by this point. 🙂

Please copy a minimal subset of the approach I added in https://github.com/cilium/cilium/blob/5bb0c048cf8a00b110f283f110bd432cfb88b0f4/pkg/netns/netns_linux.go#L154 and https://github.com/cilium/cilium/blob/5bb0c048cf8a00b110f283f110bd432cfb88b0f4/pkg/testutils/netns/netns.go. The only exposed API surface should be `testutils.NetNS` (`NetNS.Do()`) and `testutils.NewNetNS`.

Jack-R-lantern: Thank you for the review.
I gained a lot of insight from how Cilium manages network namespaces.

btf/testdata/btf_testmod.btf
florianl, line 1: Is this changed file some left over from some other change?

link/netkit_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build !windows
+//go:build linux
```
florianl: For consistency with other files, it should stay as `!windows`:
```suggestion
//go:build !windows
```

btf/testdata/vmlinux.btf.gz
florianl, line 1: Similar to the comment on `btf/testdata/btf_testmod.btf`. Can this change be reverted?

btf/testdata/btf_testmod.btf
Jack-R-lantern: The CI step “Generate and format code” failed.

According to the logs, the suggested action is to run make.
When I run make, it modifies btf_testmod.btf and vmlinux.btf.gz.

Is there a way to get the CI to pass without changing those two files?

For now, I’ll revert the changes to those two files temporarily.

link/netkit_test.go
Jack-R-lantern: `netns` are supported only on Linux, so build errors occurred during cross build (e.g. darwin). I’ve made changes.

If to keep the current `go build directives`, I think we would need to leave stub implementations for operating systems other than Linux.

Do you have any recommended approach?

internal/testutils/netns.go
```diff
@@ -0,0 +1,204 @@
...
+	return os.OpenFile(path, unix.O_RDONLY|unix.O_CLOEXEC, 0)
+}
+
+// getFromThread gets a file descriptor to the network namespace of a given pid
+// and tid.
+func getFromThread(pid, tid int) (*os.File, error) {
+	return getFromPath(fmt.Sprintf("/proc/%d/task/%d/ns/net", pid, tid))
+}
+
+func WithNetNameSpace(ns *NetNS, fn func(t *testing.T)) func(t *testing.T) {
```
ti-mo: This looks unused.

link/netkit_test.go
ti-mo: Please stay consistent with the existing approach and conventions. Stubs are fine. 🙂

link/netkit_test.go
```diff
@@ -31,7 +31,14 @@ func TestNetkitAnchor(t *testing.T) {
 	a := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
 	b := mustLoadProgram(t, ebpf.SchedCLS, ebpf.AttachNetkitPrimary, "")
 
-	linkA, ifIndex := mustAttachNetkit(t, a, ebpf.AttachNetkitPrimary)
+	ns := testutils.NewNetNS(t)
+
+	var linkA Link
+	var ifIndex int
+	ns.Do(func() error {
+		linkA, ifIndex = mustAttachNetkit(t, a, ebpf.AttachNetkitPrimary)
```
ti-mo: IIRC calling t.Fatal in a goroutine other than the one the test is running in will cause a panic. `mustAttachNetkit` should take a NetNS and call only `rtnetlink.Dial()` and `AttachNetkit` in the new netns. Asserts should all be done outside of ns.Do().

internal/testutils/netns_linux.go
```diff
@@ -0,0 +1,204 @@
+//go:build linux
+
```
ti-mo, line 2:
```suggestion
// The netns implementation in this file was taken from cilium/cilium.
```

florianl (dismissed):

internal/testutils/netns_other.go
```diff
@@ -0,0 +1,18 @@
...
+type NetNS struct {
+}
+
+func NewNetNS(tb testing.TB) *NetNS {
+	tb.Helper()
+	return &NetNS{}
+}
+
+func (h *NetNS) Do(f func() error) error {
+	return f()
```
ti-mo: This shouldn't silently execute the function anyway, this does _not_ do what's expected. I'd return `ErrNotSupportedOnOS` here.

internal/testutils/netns_other.go
```diff
@@ -0,0 +1,18 @@
...
+package testutils
+
+import "testing"
+
+type NetNS struct {
+}
+
+func NewNetNS(tb testing.TB) *NetNS {
+	tb.Helper()
+	return &NetNS{}
```
ti-mo: The function body should be:
```suggestion
	return nil
```

link/netkit_test.go
```diff
@@ -50,14 +56,18 @@ func TestNetkitAnchor(t *testing.T) {
...
-				Interface: ifIndex,
-				Anchor:    anchor,
+			ns.Do(func() error {
+				linkB, err := AttachNetkit(NetkitOptions{
+					Program:   b,
+					Attach:    ebpf.AttachNetkitPrimary,
+					Interface: ifIndex,
+					Anchor:    anchor,
+				})
+				qt.Assert(t, qt.IsNil(err))
```
ti-mo: Same as https://github.com/cilium/ebpf/pull/1913#discussion_r2665251916, this can call `t.Fatal()` on a testing.T bound to another (parent) goroutine, which will panic. Declare `linkB` and `err` outside of ns.Do() and lift the asserts out of it as well.

link/netkit_test.go
```diff
@@ -67,9 +77,13 @@ var prevIfindex atomic.Uint32
 
 func init() { prevIfindex.Store(1000 - 1) }
 
-func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachType) (Link, int) {
+func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachType, ns *testutils.NetNS) (Link, int) {
 	var err error
-	conn, err := rtnetlink.Dial(nil)
+	var conn *rtnetlink.Conn
+	err = ns.Do(func() error {
+		conn, err = rtnetlink.Dial(nil)
```
ti-mo, line 80: This first assigns directly from the closure to `err` in the outer scope, then assigns it again using the `err` returned from `ns.Do`. That's not ideal.

I'd do this:
```go
	qt.Assert(tb, qt.IsNil(ns.Do(func() (err error) {
		conn, err = rtnetlink.Dial(nil)
		return err
	})
```

And no `var err error` declaration in the outer scope.

link/netkit_test.go
```diff
@@ -99,10 +113,14 @@ func mustAttachNetkit(tb testing.TB, prog *ebpf.Program, attachType ebpf.AttachT
 		qt.Assert(tb, qt.IsNil(conn.Link.Delete(uint32(ifIndex))))
 	})
 
-	link, err := AttachNetkit(NetkitOptions{
-		Program:   prog,
-		Attach:    attachType,
-		Interface: int(ifIndex),
+	var link Link
+	err = ns.Do(func() error {
```
ti-mo: Same as above.

internal/testutils/netns_other.go
Jack-R-lantern: If return `nil`, then every test that creates a namespace via NewNetNS will need to add additional checks to handle the case where ns is nil.

I think this increases boilerplate.
Even so, do you think changing it to return nil is the right approach?

ti-mo: > If return `nil`, then every test that creates a namespace via NewNetNS will need to add additional checks to handle the case where ns is nil.

Not the case for two reasons:
- The test should be skipped on unsupported platforms, so the code wouldn't execute anyway
- Methods can be called on typed nil pointers as long as the receiver isn't accessed in said method

(see https://go.dev/play/p/nno5uE7FNWI for example)

Jack-R-lantern: Oh, I didn’t realize it worked that way. I’ll change it to return `nil`.

Jack-R-lantern:
I’m not sure why the tag value in TestProgramInfo differs from the expected value.
Is there something additional that I need to do?

ti-mo:
@Jack-R-lantern That's due to ongoing https://github.com/cilium/ebpf/pull/1932 and https://github.com/cilium/ebpf/pull/1933. We'll try to get those merged over the next few days. I've pushed a fixup and will merge this to get it over with. Thanks for bearing with me!

Jack-R-lantern:
@ti-mo 

No problem at all.
I learned a lot from your kind review.
Thanks again!

PR #1914, opened 2025-12-01
feat: support program run batching with new BatchSize option

takehaya:
## Summary
Adds BatchSize to RunOptions and plumbs it through to sys.ProgRunAttr.BatchSize, enabling multiple program executions per single `BPF_PROG_RUN` syscall.

## Motivation
In cases where the call overhead is high, such as with`BPF_F_TEST_XDP_LIVE_FRAMES`, this option becomes desirable.
I am developing my own packet generator, and I found that I need this feature for optimization.
I would be very happy if this could be accepted.


prog.go
```diff
@@ -745,6 +745,8 @@ type RunOptions struct {
 	// CPU to run Program on. Optional field.
 	// Note not all program types support this field.
 	CPU uint32
+	// BatchSize is the number of executions per syscall.
```
ti-mo: I had to look up what this is actually for, and I think the niche-ness of it warrants some elaboration. If I understand the manual correctly, this doesn't affect the amount of syscalls made by the library (that's what the `Repeat` option does).

```suggestion
	// BatchSize (default 64) affects the kernel's packet buffer allocation behaviour when running
	// programs with BPF_F_TEST_XDP_LIVE_FRAMES and a non-zero [RunOptions.Repeat] value.
	// For more details, see the kernel documentation on BPF_PROG_RUN:
	// https://docs.kernel.org/bpf/bpf_prog_run.html#running-xdp-programs-in-live-frame-mode
```

I'm not sure if it makes sense to test this flag explicitly, we'd mostly be exercising the kernel itself. Getting a new entry for this in `examples/` would be really nice, though!

ti-mo (changes requested):

prog.go
dylandreimerink: > If I understand the manual correctly, this doesn't affect the amount of syscalls made by the library (that's what the Repeat option does).

So `Repeat` also does not change the amount of syscalls made. `Repeat` does what it sounds like, for every syscall, the kernel will invoke the program `Repeat` amount of times.

The `BatchSize` is specific to the packet generation with XDP scenario. Its the batch size of the amount of packets to push down to the network stack at once. So lets say you set `Repeat` to 10.000, and `BatchSize` to 64. You will get 1 syscall, it will execute the XDP program 10.000 times, and it will "flush" the resulting packets every 64 invocations to the network stack. Currently `BatchSize` can go up to 256, which gives a performance boost over the default of 64.

takehaya: Sorry, I had written an explanation that was based on a misunderstanding. Thanks for the suggestion.

I can add an example if needed.

Of course, I have confirmed that this actually works. While developing a tool called xdperf, I made some changes and ran experiments, and I was able to confirm a performance improvement.
https://github.com/takehaya/xdperf/pull/53

takehaya:
@ti-mo san

I added an example of the behavior!
This shows that arbitrary packets can be sent :)

[886a64a](https://github.com/cilium/ebpf/pull/1914/commits/886a64aef1b5736f9ca236c536d2a76471b39811)
```shell
sudo ./xdp_live_frame ens4 1000 10 "192.168.1.1" "192.168.1.2" "<dst mac>"
```
スクリーンショット 2025-12-02 22 51 23

takehaya:
It feels like this failing test has nothing to do with mine, and it just looks like a very quirky one....

> === FAIL: asm TestGetSetJumpOp (unknown)

dylandreimerink:
> It feels like this failing test has nothing to do with mine, and it just looks like a very quirky one....

No, not your fault. Its a persistent flake we have not been able to figure out yet.

takehaya:
That is a bit of a problem....😫
Thanks for kicking the CI again. Please continue the review 🙏

examples/xdp_live_frame/xdp.c
```diff
@@ -0,0 +1,24 @@
...
+int xdp_prog_tx(struct xdp_md *ctx) {
+	return XDP_TX;
+}
+
+// xdp_prog_pass is attached to the interface to enable XDP_TX.
+// XDP_TX requires an XDP program to be attached to the target interface.
+SEC("xdp")
+int xdp_prog_pass(struct xdp_md *ctx) {
+	return XDP_PASS;
+}
```
ti-mo: Nit: missing newline at end of file. :)

takehaya: I overlooked that. I've fixed it here!
[a3db07c](https://github.com/cilium/ebpf/pull/1914/commits/a3db07c3697a1f43a420546c0143811eef753be8)

takehaya:
@ti-mo san

Thanks for the review. I squashed this commit too. Ready to merge now, please check! 🙏

ti-mo:
@takehaya Thanks, merged!

PR #1915, opened 2025-12-01
ringbuf: add zero-copy consumer APIs

jschwinger233:
The current ringbuf consumer APIs Read/ReadInto always copy samples into a user-provided buffer, see https://github.com/cilium/ebpf/blob/v0.20.0/ringbuf/ring.go#L96. For callers that consume data synchronizely, this extra copy is unnecessary.

This change adds two zero-copy consumer APIs:

- Reader.Records: yields records with RawSample as a zero‑copy view of the mmap’d ring buffer.
- Reader.RecordsInto: same iterator but reuses a caller‑supplied Record to avoid allocations.

Existing Read/ReadInto semantics are unchanged and continue to work as before.

A preliminary microbenchmark [1] shows zero-copy advantage grows with larger records because copy throughput falls while view throughput stays roughly flat.

Single-core run on CPU 2, ring size 1 GiB, Go 1.24.4. Throughput is in million events/second (Mev/s); "speedup" is zero-copy / copy.

| event-size (B) | events/run | copy (Mev/s) | zero-copy (Mev/s) | speedup |
| --- | --- | --- | --- | --- |
| 128 | 7,895,160 | 48.77 | 42.96 | 0.88x |
| 512 | 2,064,888 | 32.16 | 40.52 | 1.26x |
| 1024 | 1,040,447 | 10.57 | 38.45 | 3.64x |
| 2048 | 522,247 | 5.49 | 33.38 | 6.08x |

Small records can run slower on new APIs, because extra call layers add instructions even though IPC is higher:

```shell
$ taskset -c 2 perf stat -d -- ./bpf_ringbuf_zc_benchmark -mode copy -event-size 64
     9,939,969,508      cpu_core/instructions/           #    2.44  insn per cycle  

$ taskset -c 2 perf stat -d -- ./bpf_ringbuf_zc_benchmark -mode view -event-size 64
    11,223,740,646      cpu_core/instructions/           #    2.71  insn per cycle 
```

[1] https://github.com/jschwinger233/bpf_ringbuf_zc_benchmark


ringbuf/reader.go
```diff
@@ -144,6 +153,32 @@ func (r *Reader) Read() (Record, error) {
...
+// PeekInto is like Peek but stores the sample in view for reuse.
+func (r *Reader) PeekInto(view *View) error {
+	return r.readLocked(func() error {
+		return r.ring.readView(view)
+	})
+}
+
+// Consume advances the consumer position after a successful Peek/PeekInto.
+func (r *Reader) Consume(view *View) {
+	atomic.StoreUintptr(r.ring.cons_pos, view.nextCons)
```
florianl: It feels like this change introduces a race condition for the following call chain:
- Peek() or PeekInto()
- Read() or ReadInto() 
- Consume()

A similar undefined behaviour might also be possible, if different Go routines call Peek() or PeekInto() and Consume() in various orders.

jschwinger233: Thanks for pointing that out. I have to confess, it's by design to only support a single consumer with peek-only APIs. 😅

My first version was implemented with a boolean `viewInFlight` in the ring buffer to take care of multiple calls to Peek, Read after Peek, etc. However, it turns out the additional synchronization inside the ring buffer (mutex or atomic for `viewInFlight`) jeopardizes performance — the advantages of zero-copy are dragged down. So I re-designed the APIs as they are.

(BTW, I was wondering if I could get some advice on fixing the CI failures. 🙏 The new test cases are commented out — it seems like my changes regarding Read/ReadInto are problematic? Which test case in particular failed? I don't understand the error log: https://github.com/cilium/ebpf/actions/runs/19843613354/job/56856983722?pr=1915 😬)

ti-mo: That error would be due to https://github.com/lmb/vimto/issues/29. A re-run should fix that, the patch was already reverted in vimto.

I feel like it's always going to be difficult to marry Read(Into) and Peek/Consume when they're implemented onto a single reader and are thus sharing state. Users of a Reader expect some form of concurrency safety. Interfaces with different safety guarantees should probably be separated. Or, instead of Peek/Consume, design a different interface that e.g. takes a caller-provided function that allows the caller to inspect (parts of) a message to decide if they'd be interested in receiving a copy (kind of like [pin.Pin.Take()](https://pkg.go.dev/github.com/cilium/ebpf/pin)), or where they can unmarshal (part of) a message and send it to a channel. After invoking the callback, you move the consumer position. That saves one copy compared to Read(Into).

Also, a 13% raw throughput improvement sounds rather.. disappointing? I think allocation statistics are missing from this picture. Have you profiled it to find other potential gains? Would it be possible to write this as a plain Go benchmark? (might be difficult due to needing to spin up a producer somewhere, etc.)

jschwinger233: Hi Timo, long time no see. 

I did another round of benchmark, the latest results are updated in PR description. It seems throughput improvement grows with larger records, zero-copy API can be 4x faster when record size is 1024 bytes. I think it's valuable to switch to zero-copy APIs when ringbuf events are of big size.

jschwinger233:
Thank you for the inputs :pray: 

- API: Zero-copy APIs are intentionally single-consumer and mirror existing Read/ReadInto semantics. I’m open to reshaping this (separate interfaces, callback-based API, etc.) if you prefer another direction.
- Performance: Benchmarks in the PR description are updated; for 1024-byte records the zero-copy path is up to ~4× faster.
- CI: a re-run is now green.

From my side this is ready for review. Please feel free to request changes if any part of the design or implementation doesn’t make sense.

lmb:
I agree with Timo and Florian: `Reader` needs to be concurrency safe. As Timo mentioned, doing a callback based design is one option. Another one would be an iterator. This has the benefit of being a bit more ergonomic for callers because its easier to break out of a for loop than a closure.

```Go
// Records iterates over records in the reader until [Reader.Close] is called.
//
// Record.Sample is only valid until the next call to the iterator.
func (*Reader) Records() iter.Seq2[*Record, error]
```

You might have to adjust the method to return `Record` instead to avoid allocations, depends a bit on how smart the Go compiler is. Internally the iterator would lock the ring, construct a Record from the ring contents (be careful to re-slice the capacity of `Sample` to prevent OOB appends to the buffer) and yield the record. Once the yield returns it would update the ring consumer offset and do another round. The tricky part will be figuring out when to take / drop locks here.

Also note that this means that `View` is not necessary anymore. The consumer offset state is stored in the iterator instead.

jschwinger233:
I re-implemented Records iterator API as Lorenz suggests 🙏

ringbuf/reader.go
```diff
@@ -171,9 +232,7 @@ func (r *Reader) ReadInto(rec *Record) error {
 		}
 
 		for {
-			err := r.ring.readRecord(rec)
-			// Not using errors.Is which is quite a bit slower
```
lmb, line 175: Nit: please keep the comment.

ringbuf/reader.go
```diff
@@ -144,6 +152,59 @@ func (r *Reader) Read() (Record, error) {
...
+//
+// This convenience wrapper allocates a single Record once. To fully avoid
+// allocations, use [Reader.RecordsInto] and pass in a reusable Record.
+func (r *Reader) Records() iter.Seq2[*Record, error] {
+	rec := Record{}
+	return r.RecordsInto(&rec)
+}
+
+// RecordsInto is like Records but allows reusing the Record and associated buffers.
+func (r *Reader) RecordsInto(rec *Record) iter.Seq2[*Record, error] {
```
lmb, line 173: If `Records()` can get away with allocating Record once when creating the iterator there isn't much point in this function IMO.

ringbuf/reader.go
```diff
@@ -144,6 +152,59 @@ func (r *Reader) Read() (Record, error) {
...
+				yield(nil, err)
+				return
+			}
+
+			// Limit cap to len so append can't write past the record and corrupt the ring.
+			rec.RawSample = sample[:len(sample):len(sample)]
+			rec.Remaining = remaining
+
+			if !yield(rec, nil) {
+				atomic.StoreUintptr(r.ring.cons_pos, nextCons)
```
lmb, line 197: I don't think this is safe to do. Consider two concurrent callers of Records():
- First caller reads record A, offset 10
- Second caller reads record B, offset 20
- Second caller finishes the loop quicker than the first, offset 20 is written.
- Data of the first caller is corrupted.

I think you need to hold the lock across the yield. Maybe its as simple as moving it all into readLocked?

ringbuf/ring.go
```diff
@@ -87,15 +88,26 @@ func (rr *ringReader) readRecord(rec *Record) error {
 			continue
 		}
 
-		if n := header.dataLen(); cap(rec.RawSample) < n {
-			rec.RawSample = make([]byte, n)
-		} else {
-			rec.RawSample = rec.RawSample[:n]
-		}
+		end := int(start) + header.dataLen()
+		return rr.ring[start:end], int(prod - cons), cons, nil
```
lmb, line 92: The reslicing to reduce cap could already happen here.

ringbuf/reader.go
```diff
@@ -144,6 +152,59 @@ func (r *Reader) Read() (Record, error) {
...
+				return
+			}
+
+			atomic.StoreUintptr(r.ring.cons_pos, nextCons)
+		}
+	}
+}
+
+// readLocked drives the polling / data-availability loop shared by Record reads.
+func (r *Reader) readLocked(read func() error) error {
```
lmb, line 207: having read be a closure over Reader state is a bit black magic. How about
```suggestion
func (r *Reader) readLocked(handle func(sample []byte, remaining int) error) error {
```

readLocked would be responsible for calling readSample and advancing the consumer position. `readRecord` would just be `func(*Record, sample []byte, remaining) error`.

Probably also easier for the compiler to turn into good code since handle doesn't have to be a closure.

ringbuf/reader_test.go
```diff
@@ -384,3 +384,34 @@ func BenchmarkReadInto(b *testing.B) {
...
+	mustRun(t, prog)
+
+	rd, err := NewReader(events)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer rd.Close()
+
+	var seen [][]byte
+	for rec, err := range rd.Records() {
```
lmb, line 404: This needs a test that Reader.Close() stops the iterator without yielding io.ErrClosed. Using iter.Pull might be the easiest way to achieve that.

ringbuf/reader.go
```diff
@@ -144,6 +152,59 @@ func (r *Reader) Read() (Record, error) {
...
 // ReadInto is like Read except that it allows reusing Record and associated buffers.
 func (r *Reader) ReadInto(rec *Record) error {
+	return r.readLocked(func() error {
+		return r.ring.readRecord(rec)
+	})
+}
+
+// Records iterates over records in the reader until [Reader.Close] is called.
+//
+// Record.RawSample is only valid until the next call to the iterator. Callers
```
lmb, line 162: Even more explicit: both Record and Record.RawSample are only valid until the next call.

lmb (commented):
Left some thoughts. In general we tried to keep perf and ringbuffer API the same-ish, so we might ask you to tackle perf as well (unless there is a reason why this might not work?) @florianl what do you think?

ringbuf/reader.go
jschwinger233: I am aware of this concurrent read issue, my first version was holding the lock during the iterator. However, that leads to the problem of dead lock when calling reader.SetDeadline:
```go
for rec, err := range reader.Records() {
    reader.SetDeadline() // dead lock
}
```
reader.Close() also tries to acquire the reader.mu, it will become impossible to call Close() during iteration :grimacing:

jschwinger233: I drafted a PR on my fork for your preview: https://github.com/jschwinger233/ebpf/pull/5

btf/testdata/btf_testmod.btf
florianl, line 1: What is the reason for the change on this file?

btf/testdata/vmlinux.btf.gz
florianl, line 1: What is the reason for the change on this file?

florianl (commented):
> In general we tried to keep perf and ringbuffer API the same-ish, so we might ask you to tackle perf as well [..]

I agree with @lmb on that.

btf/testdata/vmlinux.btf.gz
jschwinger233: CI told me to run `make` locally, otherwise it turns red :grimacing: then those two files are added...

ringbuf/reader.go
lmb: Not being able to call `Close()` from inside the iterator seems fine to me. Are you saying it's not possible to Close() while we are stuck in Wait()? Sounds like a bug to me.

SetDeadline is a problem, that should work. This might mean you need to split the lock somehow. Take a look at the perf reader, this has more complex locking due to pauseMu. IIRC it also drops the lock before waiting, and then acquires it after. Maybe you can use a similar strategy here (which could resolve Close() during Wait()).

btf/testdata/vmlinux.btf.gz
aviadhahami: poking here -
since these are generated locally - they are bound to your kernel version, ~v6.x, which will render the repo's tests bound to it as well, which is generally not good :)

ringbuf/reader.go
jschwinger233: > it also drops the lock before waiting, and then acquires it after.

Let me make it clear, so we are talking something like

```go
for rec, err := range reader.Records() {
    // read.mu is released 
    reader.SetDeadline() // no dead lock
}
```

This strategy doesn't fix the two concurrent callers of Records(), right? 

> 1.  First caller reads record A, offset 10
> 2.  Second caller reads record B, offset 20
> 3.  Second caller finishes the loop quicker than the first, offset 20 is written.
> 4.  Data of the first caller is corrupted.

Consider 1st reader:

```go
for rec, err := range reader.Records() {
    // read.mu is released
    time.Sleep() // for some reason, let's sleep
    handle_rec()
}
```

2nd reader:
```go
for rec, err := range reader.Records() {
    // read.mu is released, so no blocking when 1st reader is sleeping
    handle_rec()
}
```

2nd reader will move forward the cosumer_pointer, leaving slower 1st reader's rec invalid.

jschwinger233:
I drafted a new PR to give a try of callback API: https://github.com/cilium/ebpf/pull/1930, I hope it can address the concurrency issue :pray:

lmb:
@jschwinger233 I stepped down as a maintainer, please coordinate with Timo or Dylan!

btf/testdata/vmlinux.btf.gz
ti-mo: > poking here - since these are generated locally - they are bound to your kernel version, ~v6.x, which will render the repo's tests bound to it as well, which is generally not good :)

Not really, we tend to be quite anal about reproducible builds and tests. This file is pulled from an image in cilium/ci-kernels, which saw some changes recently. A rebase should fix this.

ti-mo:
@jschwinger233 Another zero-copy proposal popped up, and this one is the least likely to get merged, so I'll close it for now.

PR #1919, opened 2025-12-20
fix(elf): set attach_type to BPF_XDP for XDP section

chantra:
Summary: Since
4540aed51b12 ("bpf: Enforce expected_attach_type for tailcall compatibility")

expected_attach_type is enforced for tailcall:
https://lore.kernel.org/bpf/20250926171201.188490-1-daniel@iogearbox.net/

When an XDP program is loaded though cillium/ebpf, its `AttachType` is set to `AttachNone`.
When chaining programs using a `BPF_MAP_TYPE_PROG_ARRAY` of XDP programs, it works fine as long as all XDP program are loaded using `cillium/ebpf`.

Programs based off libbpf will have their `expected_attach_type` set to XDP: https://github.com/libbpf/libbpf/blob/afb8b17bc50b0b7606ad4ea468cbc9f5aede8dae/src/libbpf.c#L9868

Until 4540aed51b12 it was fine, but with 4540aed51b12, the second program to register to the PROG_ARRAY map would fail with:
```
libbpf: prog 'demo_chainer': BPF program load failed: Invalid argument
libbpf: prog 'demo_chainer': -- BEGIN PROG LOAD LOG --
processed 235 insns (limit 1000000) max_states_per_insn 1 total_states 19 peak_states 19 mark_read 7
-- END PROG LOAD LOG --
libbpf: prog 'demo_chainer': failed to load: -22
libbpf: failed to load object 'demo_chainer_bpf'
libbpf: failed to load BPF skeleton 'demo_chainer_bpf': -22
```

Test Plan:
* Updated unittests
* Tested on a custom deployment where both a cillium/bpf based program and a libbpf_rs one call into each others.

### DIFF
--- a/elf_reader.go
+++ b/elf_reader.go
@@ -1556,14 +1556,6 @@ func getProgType(sectionName string) (ProgramType, AttachType, uint32, string) {
 		if t.flags&_SEC_XDP_FRAGS > 0 {
 			flags |= sys.BPF_F_XDP_HAS_FRAGS
 		}
-		if t.flags&_SEC_EXP_ATTACH_OPT > 0 {
-			if programType == XDP {
-				// The library doesn't yet have code to fallback to not specifying
-				// attach type. Only do this for XDP since we've enforced correct
-				// attach type for all other program types.
-				attachType = AttachNone
-			}
-		}
 		if t.flags&ignoreExtra > 0 {
 			extra = ""
 		}
--- a/elf_reader_test.go
+++ b/elf_reader_test.go
@@ -180,6 +180,7 @@ func TestLoadCollectionSpec(t *testing.T) {
 				Name:        "xdp_prog",
 				Type:        XDP,
 				SectionName: "xdp",
+				AttachType:  AttachXDP,
 				License:     "MIT",
 			},
 			"no_relocation": {
@@ -1296,7 +1297,7 @@ func TestELFSectionProgramTypes(t *testing.T) {
 		{"xdp_cpumap/", XDP, AttachXDPCPUMap, 0, ""},
 		// Used incorrect attach type.
 		{"xdp.frags/foo", XDP, AttachXDP, sys.BPF_F_XDP_HAS_FRAGS, ""},
-		{"xdp/foo", XDP, AttachNone, 0, ""},
+		{"xdp/foo", XDP, AttachXDP, 0, ""},
 		{"perf_event", PerfEvent, AttachNone, 0, ""},
 		{"lwt_in", LWTIn, AttachNone, 0, ""},
 		{"lwt_out", LWTOut, AttachNone, 0, ""},


lmb:
Seems to me like this will break the following:

- program creates and populates prog array with old version and pins it.
- program is updated to include your fix. Starts up, opens pinned map, tries to insert program with BPF_XDP.
- This fails due to the upstream change.

Is that correct?

chantra:
> Seems to me like this will break the following:
> 
> * program creates and populates prog array with old version and pins it.
> * program is updated to include your fix. Starts up, opens pinned map, tries to insert program with BPF_XDP.
> * This fails due to the upstream change.
> 
> Is that correct?

AFAIUI, I think this is correct depending how you look at it, and the sooner it is fixed (before the upstream change gets back ported) the less likely to cause breakages as the kernel will ignore the attach type for PROG_ARRAY until the upstream patch rolls in. Is this correct @borkmann ?

here are the 2 ways to look at it that I can think of.  Sorry, this gets a bit lengthy.

1) In a context where *ALL* cooperating BPF programs are all using this library, your statement is correct but only once people run them on a kernel that has been updated to contain the upstream fix. I believe it is making it down into the  various LTS kernels. Until the programs run on those kernels, the attach type value was being ignored so you could update, use the new attach type and still attach to the old prog_array map.

2) in a context where some of the loaders are using cillium/ebpf, some are backed by libbpf. The apps using libbpf are currently setting attach type to BPF_XDP. Before this PR and the upstream change. Both the cillium/ebpf and libbpf apps can create/populate the same PROG_ARRAY and work together. After the upstream change, it does not work anymore unless I force the libbpf app to change the attach type to None after opening. With upstream change and this PR, both apps can run together happily.

I did not manage to set the attach type from the current API when using the files generated by bpf2go, but given that the canonical loader defaults to BPF_XDP type, I feel that while being able to change the type after opening the BPF program can be useful, defaulting to BPF_XDP type seems sensible.

chantra:
Hi folks,

I am thawing from the holiday break and checking back here.

@borkmann is my current understanding of the problem correct?

@lmb what are your thoughts? Anything you would like me to address? I understand this could be a breaking change and chances of breaking increases as time pass.

We can probably provide a way to set the attach type through [ProgramOptions](https://pkg.go.dev/github.com/cilium/ebpf#ProgramOptions) so it gets set during `LoadAndAssign`. But this still leave the default behaviour of this library somewhat incompatible to libraries using libbpf.

lmb:
@chantra im winding down my involvement with the library and will defer to @ti-mo and @dylandreimerink .

dylandreimerink:
Hey @chantra. Thank you for your work on this. I will take over for Lorenz.

The logic in question only effect the ELF parsing. Users of the library can still overwrite the defaults in the CollectionSpec before loading. So even for older versions of the library there is an escape hatch for those running into this.

Ultimately I do this this is the right way to go. This is an exception we have kept in place for backwards compatibility if I understand the description of https://github.com/cilium/ebpf/commit/0acd95c6c83cabeff32e94594cd4b7de6e7df4a2 correctly. But we ultimately want to follow libbpf's lead here.

There is indeed a risk that a library update might actually trigger this error on v6.18+ kernels and LTS kernels with backports when attach types are mixed, but I don't really see a good way to avoid that, so lets move forward and make sure to put this in the release notes. Users will need to explicitly set the attach type for XDP programs to `AttachNone` if they need to preserve old behavior or recreate the prog array map and reload all programs with `AttachXDP`.

dylandreimerink (changes requested):
Changes look good in general. I see you updated the `btf/testdata` files, I assume due to the failing CI we had for a bit. This should be fixed already by #1928, could you rebase / drop the changes to the testdata?

chantra:
Thanks @dylandreimerink !
I reverted the testdata change and rebased. 
```
VIMTO_DISABLE_KVM=true vimto -kernel :mainline -- go test ./...
```
pass on my dev box, hoping for CI to do as well :D

> Users of the library can still overwrite the defaults in the CollectionSpec before loading.

yes, this is my understanding. I did not dig too much as to how exactly this would be done but I am under the impression that one would basically need to reimplement the helper:

```
func loadXdp_rootObjects(obj interface{}, opts *ebpf.CollectionOptions) error {
	spec, err := loadXdp_root()
	if err != nil {
		return err
	}

	return spec.LoadAndAssign(obj, opts)
}
```
with something like:
```
func myLoadXdp_rootObjects(obj interface{}, opts *ebpf.CollectionOptions) error {
	spec, err := loadXdp_root()
        spec.Programs[$notsurewhatkeyhere].AttachType = AttachXDP
	if err != nil {
		return err
	}

	return spec.LoadAndAssign(obj, opts)
}
```

Is this correct? If so, may I suggest a quick hint in the release note?

dylandreimerink:
> Is this correct?

Yes. That is correct, and `$notsurewhatkeyhere` would be the name of the BPF program / C function, in case your collection has multiple.

We will be releasing a new version of the library somewhere in the near future. I will be sure to mention this in the release notes we make for it.

chantra:
@dylandreimerink is there anything you want me to address or the PR is good as it is?

ti-mo:
@chantra I'm taking a look and adding a few commits on top that take care of cleaning up the legacy non-libbpf XDP sections we've supported for a long time. Since we're merging a potentially-breaking change, now would be the time to get it all over with. :) Then we can draw attention to all of it in the release notes.

ti-mo:
Sorry, I'll actually make a separate PR for that so it's a little more obvious.

ti-mo:
@chantra Thanks for driving this!

Issue #1921, opened 2025-12-23
Adding support for privilege delegation using BPF Token


anryko:
I'm trying to load ebpf program in unprivileged container with bpffs privilege delegation enabled. It works with `bpftool` but fails while using `cilium/ebpf`. Is this functionality supported?

BPFFS mount:
```sh
none on /sys/fs/bpf type bpf (rw,relatime,delegate_cmds=any,delegate_maps=any,delegate_progs=any,delegate_attachs=any)
```

Successful map creation while using `bpftool`:
```sh
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=860, max_entries=1, map_flags=BPF_F_RDONLY_PROG|0x10000, inner_map_fd=0, map_name="", map_ifindex=0, btf_fd=3, btf_key_type_id=0, btf_value_type_id=0, btf_vmlinux_value_type_id=0, map_extra=0, ...}, 92) = 13
```

Failure loading the same object file using `cilium/ebpf`:
```sh
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_ARRAY, key_size=4, value_size=860, max_entries=1, map_flags=BPF_F_RDONLY_PROG, inner_map_fd=0, map_name="", map_ifindex=0, btf_fd=0, btf_key_type_id=0, btf_value_type_id=0, btf_vmlinux_value_type_id=0, map_extra=0}, 80) = -1 EPERM (Operation not permitted)
```

dylandreimerink:
> Is this functionality supported?

Not yet. Though I see you are actively working on this (thank you).

> Please help out and advice on how to proceed with moving forward with this feature.
> - #1926

We should probably think about the abstraction level at which we want to implement it and the architecture for this more broadly. What you have done so far in the POC is to essentially do it at the syscall level and to make a "token per process" model.

I took a quick peek at how libbpf approached this, and I think they have got it by the right end in this case. They went with a token per Collection model. They allow a user to pass the path of the bpffs they would like to use. If no path is specified they attempt to use the default `/sys/fs/bpf`. They also have a environment variable as path input which I think we can do away with. For us that would be a string field in the collection spec.

[...]

PR #1926, opened 2026-01-02
feat: poc of bpf token delegation

anryko:
A minimal implementation of [privilege delegation via BPF token](https://docs.ebpf.io/linux/concepts/token/) feature. I only implemented the functionality which was required to test my specific usecase. I'm not sure if this is a good idea to implement it as I did, via a fallback action to try on permission error. But I did not want to introduce API changes while testing basic functionality.

Please help out and advice on how to proceed with moving forward with this feature.

Relates to: #1921

### DIFF
--- a/bpffs/bpffs.go
+++ b/bpffs/bpffs.go
@@ -1,4 +1,4 @@
-package ebpf
+package bpffs
 
 import (
 	"errors"
@@ -58,7 +58,7 @@ func (bf *BPFFS) Close() error {
 	return errors.Join(errs...)
 }
 
-func (bf *BPFFS) token() (*sys.FD, error) {
+func (bf *BPFFS) Token() (*sys.FD, error) {
 	if bf.tokenFd != nil {
 		return bf.tokenFd.Dup()
 	}
--- a/btf/btf_test.go
+++ b/btf/btf_test.go
@@ -18,6 +18,8 @@ import (
 	"github.com/cilium/ebpf/internal/testutils"
 )
 
+const NotBPFTokenFd int32 = -1
+
 func vmlinuxSpec(tb testing.TB) *Spec {
 	tb.Helper()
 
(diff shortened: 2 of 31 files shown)


anryko:
**Summary**

- The `FeatureTestFn` signature was changed to support configurable `NewFeatureTest()`, required to setup BPF Token FD and MountPath.

- BPFFS token is created by `Collection` based on `CollectionOptions` configuration and added as `obj.BpffsTokenFD` to the Map and Program Specs. On Map and Program creation it is duplicated and stored in `obj.tokenFd` to make `obj.Copy()` work. The `LoadPinOptions` has `BpffsTokenFd` as well. Token has to be provided in unprivileged mode, to work with the objects loaded from FD.

- To the entities that make a copy and hold on to `tokenFd` it is passed as `*sys.FD`. To everything that just uses the token and doesn't manage its lifetime it's passed as `int32`. Usually the token is needed only for feature detections, as those are attempting to load the dummy program or create a dummy map and check the result. BPF token related documentation will have to inform the user that BPFFS `delegate_progs` list has to include  the `socket_filter` and `kprobe`, otherwise feature detection will not work. Maybe we could consider allowing user to provide their own dummy BPF program to be used for feature detection to avoid mandatory allowing `socket_filter` and `kprobe`.

Please let me know if these changes make sense and I'm looking forward to your feedback.

javiercardona-work:
Hi @anryko

It looks like we are both interested in getting ebpf to work with BPF token.  I have #1948 out which I believe implements what you already have, plus a test program to confirm that probes and ebpf operations work with dropped privileges.  I would really appreciate your comments on it and whether that would meet your needs.

Cheers,
Javier

ti-mo:
@anryko @javiercardona-work Thanks for your work on this, but we'll go forward with Dylan's proposal (https://github.com/cilium/ebpf/pull/1953). It should get merged in the next few weeks.

Issue #1916, opened 2025-12-04
ProgramSpec's Tag() broken in kernel 6.18+


lmtcf:
### Describe the bug

Users of the ebpf cillium library can use the `Tag()` method on a ProgramSpec (https://github.com/cilium/ebpf/blob/3d4ca808756e3953ced47867419e5bb5f3b50477/asm/instruction.go#L748) to check if the current program to be loaded matches the currently loaded program in the kernel. An example in given in the `How to reproduce` section.

This is useful because it means the check itself doesn't need any special capabilities - it doesn't use `BPF_PROG_GET_FD_BY_ID`, thus no need for `CAP_SYS_ADMIN`, nor CAP_BPF given it doesn't need to create/load any program/maps.

However, from kernel 6.18 the tag now uses sha256 [1], instead of sha1, meaning the `Tag()` method referenced above stops working.

I'm opening this issue for visibility on the above, but also as a discussion point. Is there a better way to do this than to duplicate/mimic kernel code?

Our use case is simple - we simply want to check if the program we have (ProgramSpec) is the same as the one that's loaded in the kernel - without requiring any special capabilities. So we don't even need to use the kernel tag - some unique number like the commit version, which we could store in a global variable in the program, would also work.

However, from what I've looked at, even if we'd use a global variable in `.rodata`, like in [2] - where `.rodata` would still be kept around even if not referenced by the main program - my understanding is that we'd still need to have special capabilities to be able to read the loaded program's `.rodata` afterwards.

[1]  https://lore.kernel.org/all/20250914215141.15144-2-kpsingh@kernel.org/#t
[2] https://lore.kernel.org/netdev/20200915234543.3220146-1-sdf@google.com/

Kudos to @jsitnicki / @arges for finding the issue and root cause.

### How to reproduce

```
func compatible(newProgSpec *ebpf.Program) error {
    newTag, err := newProgSpec.Tag()
    if err != nil {
        return fmt.Errorf("calculate tag: %s", err)
    }

    // Get old program info by calling (*Program).Info()

    if newTag != oldProgInfo.Tag {
        return fmt.Errorf("new program and old program tags differ"
    }
}
```

### Version information

all versions

lmb:
I think the answer is to add `func ProgramSpec.Compatible(ProgramInfo)` or similar which simply tries both algorithms. And then deprecate `ProgramSpec.Tag()` since we can't reliable calculate that value without looking at the kernel (which is a big pain and not very reliable).

PR #1932, opened 2026-01-09
Deprecate `ProgramSpec.Tag`, and introduce `ProgramSpec.Compatible`

dylandreimerink:
In kernel v6.18 the hashing algorithm used by the kernel to calculate program hashes was changed from SHA-1 to SHA-256. The `ProgramSpec.Tag` method calculates the SHA-1 hash of the instructions in a `ProgramSpec` by implementing a version of the kernels algorithm in Go.

However, now that there are two algorithms in use depending on kernel version, it becomes impossible for us to return a correct hash without knowing the kernel version, which we do not want to do.

The primary use case for knowing the Tag of a ProgramSpec is to check if the instructions in the spec are equal to an already loaded program. Which would allow a user to avoid reloading the program if it is already loaded and has not changed.

This commit introduces a new method on ProgramSpec called `Compatible` which compares the program info of a loaded program with the current ProgramSpec. It compares the loaded tag with both SHA-1 and SHA-256 hashes of the ProgramSpec to ensure compatibility on all kernels.

`ProgramSpec.Tag` is now deprecated, in favor of `ProgramSpec.Compatible`.

Fixes: #1916

### DIFF
--- a/variable_test.go
+++ b/variable_test.go
@@ -378,5 +378,5 @@ func TestVariablePointerGC(t *testing.T) {
 	runtime.GC()
 
 	// Wait for backing array to be finalized.
-	testutils.WaitChan(t, mgc, time.Second)
+	testutils.WaitChan(t, mgc, time.Second*5)
 }


prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
 }
 
+// ErrProgIncompatible is returned when a loaded program is incompatible with a given spec.
+var ErrProgIncompatible = errors.New("program is incompatible")
+
+// Compatible checks if a loaded program is of the same program type and has
+// matching instruction hashes. Returns nil on match, and error otherwise.
+func (ps *ProgramSpec) Compatible(info *ProgramInfo) error {
+	if platform.IsWindows {
+		return fmt.Errorf("%w: Windows does not support tag read back from kernel", internal.ErrNotSupportedOnOS)
```
ti-mo:
```suggestion
		return fmt.Errorf("%w: Windows does not support tag readback from kernel", internal.ErrNotSupportedOnOS)
```

prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
+	}
+
+	if info.Type != ps.Type {
+		return fmt.Errorf("%w: program type mismatch", ErrProgIncompatible)
```
ti-mo: It only takes a little effort on our part to be more descriptive here:

```suggestion
		return fmt.Errorf("%w: ProgramSpec type %s does not match Program type %s", ErrProgIncompatible, ps.Type, info.Type)
```

prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
+var ErrProgIncompatible = errors.New("program is incompatible")
+
+// Compatible checks if a loaded program is of the same program type and has
+// matching instruction hashes. Returns nil on match, and error otherwise.
+func (ps *ProgramSpec) Compatible(info *ProgramInfo) error {
+	if platform.IsWindows {
+		return fmt.Errorf("%w: Windows does not support tag read back from kernel", internal.ErrNotSupportedOnOS)
+	}
+
+	if info.Type != ps.Type {
```
ti-mo: OCD ordering
```suggestion
	if ps.Type != info.Type {
```

prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
+		return fmt.Errorf("%w: program type mismatch", ErrProgIncompatible)
+	}
+
+	insnCompatible, err := ps.Instructions.TagsMatch(info.Tag, internal.NativeEndian)
```
ti-mo: For these short-lived vars, I prefer short names:
```suggestion
	ok, err := ps.Instructions.TagsMatch(info.Tag, internal.NativeEndian)
```

prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
+	if err != nil {
+		return err
+	}
+
+	if !insnCompatible {
+		return fmt.Errorf("%w: instruction hashes mismatch", ErrProgIncompatible)
```
ti-mo: I don't think we speak of 'instruction hashes' in the API, let's stick to tag:

```suggestion
		return fmt.Errorf("%w: ProgramSpec and Program tags do not match", ErrProgIncompatible)
```

prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
+// Deprecated: The value produced by this method no longer matches tags produced
+// by the kernel since v6.18. Use [ProgramSpec.Compatible] instead.
 func (ps *ProgramSpec) Tag() (string, error) {
 	return ps.Instructions.Tag(internal.NativeEndian)
 }
 
+// ErrProgIncompatible is returned when a loaded program is incompatible with a given spec.
+var ErrProgIncompatible = errors.New("program is incompatible")
+
+// Compatible checks if a loaded program is of the same program type and has
```
ti-mo: Do we need to introduce the check on the program type? Do we care? This may be overly restrictive in some cases and we didn't have a similar constraint before.

asm/instruction.go
```diff
@@ -745,18 +747,72 @@ func (insns Instructions) Marshal(w io.Writer, bo binary.ByteOrder) error {
...
+
+// TagsMatch checks if the given tag (truncated hash of instructions) matches
+// the tag of this instruction stream.
+func (insns Instructions) TagsMatch(tag string, bo binary.ByteOrder) (bool, error) {
+	sha1Tag, err := insns.tagSha1(bo)
+	if err != nil {
+		return false, err
+	}
+
+	sha256Tag, err := insns.tagSha256(bo)
```
ti-mo: I'd try sha256 first since it's the happy path going forward, and add an `err == nil` short-circuit to avoid always having to calculate both (and incurring allocs). Over time, this will automatically do the right thing as older kernels are phased out.

prog.go
```diff
@@ -172,10 +172,39 @@ func (ps *ProgramSpec) Copy() *ProgramSpec {
...
+// matching instruction hashes. Returns nil on match, and error otherwise.
```
ti-mo:
```suggestion
// matching program tags. Returns nil on match, and error otherwise.
```

asm/instruction.go
```diff
@@ -745,18 +747,72 @@ func (insns Instructions) Marshal(w io.Writer, bo binary.ByteOrder) error {
...
+// by the kernel since v6.18. Use [Instructions.TagsMatch] instead.
 func (insns Instructions) Tag(bo binary.ByteOrder) (string, error) {
+	// We cannot determine which hashing function to use without probing the kernel.
+	// So use the legacy SHA-1 implementation and deprecate this method.
+	return insns.tagSha1(bo)
+}
+
+// TagsMatch checks if the given tag (truncated hash of instructions) matches
+// the tag of this instruction stream.
+func (insns Instructions) TagsMatch(tag string, bo binary.ByteOrder) (bool, error) {
```
ti-mo: WDYT about calling this `Compatible()` as well? Or `HasTag()`? Open to other suggestions. :bikeshed:

prog.go
dylandreimerink: We started with the name `Compatible`, so what does it mean for two programs to be "compatible"? I added it because it would make sense for two programs of different program types to not be compatible.

Lets remove it for now. We can always put if back if we feel like we need it.

asm/instruction.go
dylandreimerink: I like `HasTag` better. `Compatible` would require further definition, and this method is specifically related to the tags.

dylandreimerink:
So. The new mainline and stable kernels we test on are 6.18 and above, which we need to validate these changes. To be able to use the `IsVersionLessThan` utility in tests we also need to update the `CI_MAX_KERNEL_VERSION` to 6.18 and thus our other kernel deps. Did that, but the update breaks BTF tests. Working on both the CI kernel version update and the fix for the tests in https://github.com/cilium/ebpf/pull/1933. This PR is blocked until that is resolved.

PR #1937, opened 2026-01-16
ringbuf: context support

hown3d:
Support canceling ringbuf epoll using a context.

### DIFF
--- a/ringbuf/reader_windows.go
+++ b/ringbuf/reader_windows.go
@@ -52,21 +52,9 @@ func newPoller(fd int) (*poller, error) {
 // Returns [ErrFlushed] if the ring buffer was flushed manually.
 // Returns [os.ErrClosed] if the poller was closed.
 func (p *poller) Wait(ctx context.Context, deadline time.Time) error {
-	errChan := make(chan error)
-
-	go func() {
-		errChan <- p.wait(deadline)
-		close(errChan)
-	}()
-	select {
-	case <-ctx.Done():
-		if err := p.Close(); err != nil {
-			return err
-		}
-		return ctx.Err()
-	case err := <-errChan:
-		return err
-	}
+	return waitWithContext(ctx, func() error {
+		return p.wait(deadline)
+	}, p)
 }
 
 func (p *poller) wait(deadline time.Time) error {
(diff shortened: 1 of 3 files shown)


ringbuf/reader_windows.go
```diff
@@ -50,7 +51,13 @@ func newPoller(fd int) (*poller, error) {
 // Returns [os.ErrDeadlineExceeded] if a deadline was set and no wakeup was received.
 // Returns [ErrFlushed] if the ring buffer was flushed manually.
 // Returns [os.ErrClosed] if the poller was closed.
-func (p *poller) Wait(deadline time.Time) error {
```
florianl, line 53: Can you elaborate more a bit on why this change is needed for your case?

As you show, with `SetDeadline()` in on of the tests, a similar behaviour can be achieved.

hown3d: My use case would be to stop blocking if a context, e.g. from a http request, is canceled.

I found the `SetDeadline()` function a bit misleading for that use case, because then the application would need to handle the context cancel correctly and use the function with `time.Now()`. 
Passing contexts down to subfunction calls felt easier instead of handling it on the application side. 

Maybe there is a better take to handle this, I am open for discussion :)

ti-mo: > Passing contexts down to subfunction calls felt easier instead of handling it on the application side.

Sure, but how much easier, really? :slightly_smiling_face: After this patch, every read spawns a goroutine and incurs a channel read + write, without a way to opt out. On very busy rings, this will make a difference. I tried running the benchmarks in package `ringbuf`, but they seem to be broken at the moment. No bandwidth on my side to fix, unfortunately.

I feel like this has come up before, but I can't find anything relevant. I'm not totally opposed to the idea, but there are a few criteria that should be met before we can consider this:
1. it should be completely optional without the risk of performance regressions
2. in order to guarantee 1., the benchmarks should be fixed first
3. context should not be introduced in a viral way, meaning:
    * some internal refactoring is likely in order to make this fit
    * it should probably take a `cancel func() <-chan struct{}` so `Context.Done` can be passed in instead of context.Context wholesale
4. package `perf` has a near-identical API, so the same work would need to be done over there

hown3d: I agree with you on all of these aspects, will see how we can achieve this.

ti-mo:
@hown3d I'm closing this for now, please ping if you have an updated proposal and we can reopen.

PR #1938, opened 2026-01-18
uprobe: only verify ELF if required

RonFed:
This PR removes duplicate ELF check performed while creating a uprobe.
* If the caller provides an address and offset, the ELF verification is not required.
* If symbol is provided, the address resolving is done once and same for the ELF verification (which is done twice today).

This change also allows to attach uprobes to non ELF files, assuming the caller provides the address and offset.

This behavior is also compatible with libbpf.

### DIFF
--- a/link/uprobe.go
+++ b/link/uprobe.go
@@ -6,6 +6,7 @@ import (
 	"debug/elf"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"sync"
 
@@ -33,6 +34,8 @@ var (
 	// in the ELF symbols table.
 	ErrNoSymbol = errors.New("not found")
 )
+
+const permExec fs.FileMode = 0111
 
 // Executable defines an executable program on the filesystem.
 type Executable struct {
@@ -96,6 +99,15 @@ func (uo *UprobeOptions) cookie() uint64 {
 func OpenExecutable(path string) (*Executable, error) {
 	if path == "" {
 		return nil, fmt.Errorf("path cannot be empty")
+	}
+
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("stat executable: %w", err)
+	}
+
+	if info.Mode()&permExec == 0 {
+		return nil, fmt.Errorf("file %s is not executable", path)
 	}
 
 	return &Executable{
--- a/link/uprobe_test.go
+++ b/link/uprobe_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/go-quicktest/qt"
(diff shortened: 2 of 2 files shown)


mmat11:
> If symbol is provided, the address resolving is done once and same for the ELF verification (which is done twice today).

Correct me if I'm wrong, from what I see the check is only performed once as of now, but the ELF open is done twice even if it's potentially not needed

> This change also allows to attach uprobes to non ELF files, assuming the caller provides the address and offset.

If it's not an ELF do we need some kind of check/verification on the target?

RonFed:
> Correct me if I'm wrong, from what I see the check is only performed once as of now, but the ELF open is done twice even if it's potentially not needed

Right, elf.Open is called twice.

> If it's not an ELF do we need some kind of check/verification on the target?

Well, technically, it could be any executable file, there isn't really a requirement for a format (assuming the caller provide the offset in the file, and no symbol resolving is required).

mmat11:
> > If it's not an ELF do we need some kind of check/verification on the target?
> 
> Well, technically, it could be any executable file, there isn't really a requirement for a format (assuming the caller provide the offset in the file, and no symbol resolving is required).

right now you could do `exe, err := link.OpenExecutable("blabla")` and it would not return an error, I wonder if we should at least check if it's an executable file regardless of the format -- and to get some more context, is this coming from a performance issue or you need to open an uprobe on not an ELF file? if it's the latter, could we have a test case?

mmat11 (approved):
lgtm!

RonFed:
Hey @ti-mo, can you please have a look if this is ok to merge?

link/uprobe.go
```diff
@@ -98,15 +98,14 @@ func OpenExecutable(path string) (*Executable, error) {
...
-		return nil, fmt.Errorf("parse ELF file: %w", err)
+		return nil, fmt.Errorf("stat executable: %w", err)
 	}
-	defer f.Close()
 
-	if f.Type != elf.ET_EXEC && f.Type != elf.ET_DYN {
-		// ELF is not an executable or a shared object.
-		return nil, errors.New("the given file is not an executable or a shared object")
+	mode := info.Mode()
+	if mode&0111 == 0 {
```
ti-mo: No magic numbers, please. I'd probably write something like:

```go
const permExec fs.FileMode = 0111 // somewhere at the top of the file

if info.Mode() & permExec == 0 { ... }
```

link/uprobe_test.go
```diff
@@ -30,6 +31,35 @@ func TestExecutable(t *testing.T) {
 		t.Fatal("create executable: expected error on empty path")
 	}
 
+	_, err = OpenExecutable("/non/existent/path")
```
ti-mo, line 34: Check for `os.PathError` here or drop the statement, not sure what else it could return when specifying a bogus path.

RonFed:
Thanks @ti-mo, addressed your comments

Issue #1810, opened 2025-06-23
Ensure the ELF reader works for objects built without BTF


ti-mo:
Follow-up to https://github.com/cilium/ebpf/pull/1807.

Loading BPF objects without BTF info has always been supported, and although there's little reason to disable compiling with BTF nowadays, we should have at least some tests that somewhat verify this doesn't regress. Some users may not need BTF or don't want their users to be able to inspect it. Either we decide to always require BTF (feels heavy-handed and unnecessary), or we ensure it keeps working.

Some suggestions:
- Build loader.c without BTF on the latest toolchain. Most map defs are BTF, but there are still some legacy ones. Add a test that loads it and compares the CollectionSpec to a well-known definition like `TestLoadCollectionSpec`.
- Separate .c file with a subset of loader.c to lower the footprint of the test.
- Run some discovery in the code base to find code paths that require BTF and make sure proper errors are returned and there are no panics.

7h3-3mp7y-m4n:
can I work on this ?

ti-mo:
@7h3-3mp7y-m4n Sure! Assigned the issue to you.

Al-Masih-ad-Dajjal:
Hey! I also wanted to work on this. @7h3-3mp7y-m4n, are you busy? If so, can I pick this up?

7h3-3mp7y-m4n:
If you are that eager, then you can pick this up!
Good luck @Al-Masih-ad-Dajjal

Al-Masih-ad-Dajjal:
hey @ti-mo , can you please assign this to me ?

gajjuu:
Hi, I’d like to work on this issue. Could you assign me if it’s still available?

ti-mo:
Hi, I appreciate the interest, but given there's little movement here I'll remove all current assignments and leave it that way. If someone ends up finding some improvements in this area, please just send a PR. Thank you!

PR #1939, opened 2026-01-19
elf_reader: add tests for ELF objects without BTF

puwun:
Adds infrastructure to test loading of ELF objects compiled without BTF information, verifying the fix from #1807 

**Changes:**
- Added `testdata/loader_nobtf.c` using legacy map definitions.
- Updated `Makefile` to compile this new test data with `-g0` (no BTF).
- Added `TestLoadCollectionSpecWithoutBTF` to verify successful parsing when `spec.Types` is nil.

Fixes: #1810


ti-mo:
@puwun Thanks for picking this up! After looking at this for a bit, I think it would be preferable to reuse the existing loader.c as much as possible to increase coverage. I suggest the following:

- move BTF map definitions to a new `loader_maps_btf.h`
- create a new `loader_maps_bpf.h` with equivalent bpf_map_def-based definitions, ignoring `{bpf,btf}_decl_map` since their purpose is testing decl tags on maps, which doesn't exist pre-BTF
- `#ifndef __NOBTF__, #include "loader_maps_btf.h", #else, #include "loader_maps_bpf.h"`
- build loader.c twice, additionally specifying `-D__NOBTF__` along with the `-g0` flag
- Add a subtest to TestLoadCollectionSpec for the no-btf version. You'll want to do a similar thing with the CollectionSpec under test where you populate a part of CollectionSpec.Maps in the BTF-enabled subtest, since some maps will only be present when BTF is enabled.

Try to make only the strictly-necessary changes to keep things straightforward to review. Thank you!

puwun:
@ti-mo thanks, updated 👍

PR #1941, opened 2026-01-21
btf: support aya-style BTF map wrappers

littlejo:
Aya wraps BTF map definitions in two struct layers (`__0` → `value`) before the actual map definition. Unwrap these layers until a canonical BTF map definition is reached.


ti-mo:
Hi @littlejo, thanks for the patch. Interesting (and unfortunate :slightly_smiling_face:) that Aya takes a different approach to encoding map definitions.

Unfortunately, as you may expect, there are a few problems with this patch. To get the technical side out of the way first: we can't accept opaque non-reproducible testdata. It puts us in a difficult spot when changes need to be made down the line, and a blob like this turns into bitrot second it gets merged. It also doesn't respond to changes in tooling, which I assume are exceedingly common in the Rust ecosystem.

This brings us to the second problem: making this reproducible involves introducing the Rust toolchain to our CI and build infrastructure. Which in turn brings us to the third problem: this will require much more than a casual contribution to get done.

What you're actually proposing here is adding support for a new build target, which is a decision that can't be taken lightly and will carry significant burden for us maintainers going forward. Speaking of, we're down one maintainer as of January, so our appetite for this sort of endeavour is relatively low at the moment.

To be clear, I'm not against this kind of code landing in ebpf-go, but we'll need to have some conversations and concrete long-term commitments from your side (or from whomever else is interested in making this happen) before we can start considering this. As a minimum requirement, someone would need to liase with the upstream Aya project proper (not an issue, some Aya contributors are ex-Cilium committers) as we'll invariably hit bugs caused by upstream Rust that will need to get fixed rather than worked around downstream. In the same vein, this project has had strong ties to the Linux kernel community since its inception and bugfixes frequently flow back upstream, whether through patches or reports/reproducers.

My initial suggestion here would be to make Aya emit the same BTF as libbpf+clang. We've always aimed for compatibility with libbpf for any features we implement, so there would need to be strong reasons to deviate from that. If Aya would adopt the same compatibility stance, the code in this PR wouldn't be necessary.

Please let me know what your overall long-term plans are with this work, if any, so we can evaluate how we'll proceed. Thanks!

littlejo:
@ti-mo Thanks a lot for the detailed and thoughtful feedback, this all makes perfect sense.

I completely understand the concerns around opaque testdata and the maintenance burden of introducing a Rust toolchain into ebpf-go’s CI. That was not my intention, and I agree this would be a significant long-term commitment rather than a small incremental change.

My main goal with this work was to understand why aya-generated BTF map definitions fail to load with ebpf-go, and to explore whether this was something that could or should be addressed on the consumer side. Your feedback makes it clear that aligning Aya’s BTF output with libbpf is the preferred direction, and I agree that this is the right long-term solution.

Given that, I think the next step on my side is to engage with the Aya project to discuss emitting libbpf-compatible BTF for map definitions, or at least to understand whether this is feasible and acceptable upstream. If progress can be made there, the need for changes in ebpf-go would naturally disappear.

For now, I’m happy to drop this PR and treat it as an exploratory effort rather than something that needs to land here. Thanks again for taking the time to explain the broader context and constraints.

ti-mo:
@littlejo Thanks for doing the effort to take this upstream, looks like it set something in motion. :slightly_smiling_face: We'll close this one for now. :+1:

Issue #1078, opened 2023-06-26
BatchLookupAndDelete is much slower than iterator


msherif1234:
**Describe the bug**
It was expected to see much better performance when using BatchLookupAndDelete compared to regular iterate and delete loop

```sh
uname -r
6.1.5-100.fc36.x86_64
```
**To Reproduce**
https://github.com/msherif1234/ebpf/tree/batch_performance
```sh
go test -exec sudo -bench=BenchmarkHashDelete -count 5 -run=^#
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf
cpu: Intel Core Processor (Skylake, IBRS)
BenchmarkHashDelete/MapIteratorDelete-12                  787828              1506 ns/op              80 B/op          4 allocs/op
BenchmarkHashDelete/MapIteratorDelete-12                  795927              1500 ns/op              80 B/op          4 allocs/op
BenchmarkHashDelete/MapIteratorDelete-12                  703017              1512 ns/op              80 B/op          4 allocs/op
BenchmarkHashDelete/MapIteratorDelete-12                  663597              1559 ns/op              80 B/op          4 allocs/op
BenchmarkHashDelete/MapIteratorDelete-12                  628033              1665 ns/op              80 B/op          4 allocs/op
BenchmarkHashDelete/MapBatchLookupAndDelete-12             78336             16656 ns/op           32846 B/op          8 allocs/op
BenchmarkHashDelete/MapBatchLookupAndDelete-12             67497             16563 ns/op           32846 B/op          8 allocs/op
BenchmarkHashDelete/MapBatchLookupAndDelete-12             72202             17565 ns/op           32846 B/op          8 allocs/op
BenchmarkHashDelete/MapBatchLookupAndDelete-12             66657             16030 ns/op           32846 B/op          8 allocs/op
BenchmarkHashDelete/MapBatchLookupAndDelete-12             77671             15628 ns/op           32846 B/op          8 allocs/op
PASS
ok      github.com/cilium/ebpf  18.168s
```

**Expected behavior**
I was expecting much better cpu and memory when using Batch apis

brycekahle:
Can you add `b.ResetTimer()` right before your loops? I wonder if the slice allocation is influencing the timing?

msherif1234:
```sh
BenchmarkHashDelete/MapIteratorDelete-12                  405836              3537 ns/op              80 B/op          4 allocs/op
[...]

PR #1943, opened 2026-01-24
fix: don't unmarshal unnecessary items on batch lookup

### LINKED ISSUE #1080: Batch: only unmarshal as much as needed
See https://github.com/cilium/ebpf/issues/1078#issuecomment-1609818046.

`unmarshalBytes()` tries to unmarshal all of `buf` into `data`, but can't be told about the amount of data in `buf`. 

The way it's used in `Map.batchLookup()`, `keyBuf` and `valueBuf` could be sliced to the length of `attr.Count` but then `binary.Read()` complains about an EOF. (the inputs are shorter than `keysOut` and `valuesOut` after all.

I'm not sure if this can be solved in `binary.Read()` specifically, but we don't really care about the size of `out` parameters, as long as they're not too small.

Arpit529Srivastava:
previously, `BatchLookup` and `BatchLookupAndDelete` unmarshaled the full key/value buffers based on the requested batch size, even when the kernel returned fewer items. this caused wasted cpu cycles and extra memory allocations for empty entries.

this change updates `batchLookupCmd`, `batchLookup` and `batchLookupPerCPU` to use the actual count returned by the kernel and slice the buffers accordingly before unmarshaling, so only valid results are processed.

fixes: https://github.com/cilium/ebpf/issues/1080

### DIFF
--- a/map.go
+++ b/map.go
@@ -1290,7 +1290,25 @@ func (m *Map) batchLookup(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, valuesOu
 		return 0, sysErr
 	}
 
-	err = valueBuf.Unmarshal(valuesOut)
+	if n < count {
+		// Only unmarshal the elements that were actually returned by the kernel.
+		// Use reflection to slice the output slice to the correct length.
+		vOut := reflect.ValueOf(valuesOut).Slice(0, n).Interface()
+
+		// valueBuf should be sliced to match the number of elements.
+		// If the buffer was created with SyscallOutput(..., size), it has the
+		// full size.
+		if b := valueBuf.Bytes(); b != nil {
+			err = sysenc.Unmarshal(vOut, b[:n*int(m.fullValueSize)])
+		} else {
+			// Zero-copy: valuesOut already contains the data.
+			// Because vOut is a slice of valuesOut, it's safe to use Unmarshal
+			// which will just be a no-op but consistency check.
+			err = valueBuf.Unmarshal(vOut)
+		}
+	} else {
+		err = valueBuf.Unmarshal(valuesOut)
+	}
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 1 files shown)


Arpit529Srivastava:
@ti-mo PTAL, let me know if you need any changes.
thanks

Arpit529Srivastava:
@ti-mo a gentle ping, thanks

dylandreimerink (changes requested):
Sorry for taking so long to review this PR, our review queues are a bit on the long side at the moment.

To me the code seems correct. But I question if it makes sense to have it.

In the end this is an optimization, and any optimization should be supported by a benchmark to show the before and after. The issue linked to this comment https://github.com/cilium/ebpf/issues/1078#issuecomment-1609818046. When I run that benchmark locally before and after I essentially don't see any change.

Now, to my understanding this is because our existing iteration benchmarks use the zero copy path which essentially does not benefit from this optimization.

So, this potentially only optimizes the non-zero-copy path. And only when on the last batch read from a map if the map size is not a multiple of the batch size. A very specific set of circumstances.  Which begs the question if adding all of this code is worth the gain? 

So I think we should have a benchmark at the very least to show the improvement.

In addition, I ran tests locally with test coverage, and it seems like the non-zero-copy case is not covered by existing tests. So I would also like to see a test added or exciting test modified to cover non-zero-copy batch lookups to assert this new logic works and keeps working.

Arpit529Srivastava:
hi @dylandreimerink, thanks for the detailed review.

yeah that makes sense , i’ll add a benchmark that focuses on the non-zero-copy case with partial batches, using `customTestUnmarshaler` so i actually exercise real unmarshaling work. i’ll also add test coverage for the partial batch scenario to make sure the new logic is properly tested. 

since I'm on macOS, I can't run the benchmarks locally. would you be able to run them, once i pushes the code?

dylandreimerink:
> since I'm on macOS, I can't run the benchmarks locally. would you be able to run them, once i pushes the code?

Yes of course, no problem.

dylandreimerink:
#2052 made another attempt at this with a benchmark. I think the conclusion is that it does not make sense to add additional logic, the code is good enough for now. So closing this PR and the original issue.

PR #1949, opened 2026-02-03
Add `ErrNotPermitted` error for permission denied in feature probes                                                                                                                                                                          

javiercardona-work:
Add `ErrNotPermitted` to distinguish permission denied (`EPERM`) from feature not supported (`EINVAL`) in feature probes.

This informs the caller that the feature may work with proper permissions (`CAP_BPF` or a `BPF` token) instead of having to upgrade to a newer kernel.                                                                                                                                                              
                                                                                                                                                                                                                                             
## Problem                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                             
  Previously, feature probes treated EPERM and EINVAL identically, returning ErrNotSupported for both. This was imprecise:                                                                                                                                                                                    
                                                                                                                                                                                                                                       
  - EINVAL: kernel doesn't recognize the request (feature not supported)                                                                                                                                                                     
  - EPERM: kernel recognized the request but denied permission (feature exists)                                                                                                                                                              
                                                                                                                                                                                                                                             
## Solution                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
  Feature probes now return:                                                                                                                                                                                                                 
  - `nil`: feature is supported and accessible                                                                                                                                                                                               
  - `ErrNotSupported`: kernel doesn't have the feature                                                                                                                                                                                       
  - `ErrNotPermitted`: feature exists but permission denied                                                                                                                                                                                  
                                                                                                                                                                                                                                             
  ErrNotPermitted wraps the original EPERM error so callers can unwrap if needed.                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                             
  ## Usage                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                             
  ```go                                                                                                                                                                                                                                      
  err := features.HaveProgramType(ebpf.XDP)                                                                                                                                                                                                  
  switch {                                                                                                                                                                                                                                   
  case err == nil:                                                                                                                                                                                                                           
      // feature works                                                                                                                                                                                                                       
  case errors.Is(err, ebpf.ErrNotSupported):                                                                                                                                                                                                 
      // kernel doesn't have it                                                                                                                                                                                                              
  case errors.Is(err, ebpf.ErrNotPermitted):                                                                                                                                                                                                 
      // permission denied (but feature exists)                                                                                                                                                                                              
  }                                                                                                                                                                                                                                          
```
                                                                                                                                                                                                                                             
## Breaking Change                                                                                                                                                                                                                            
                                                                                                                                                                                                                                             
  Code checking only for `ErrNotSupported` will no longer catch permission errors. To maintain previous behavior:                                                                                                                                                                                          

```go                                                                                                                                                                                                                                             
  // Before:                                                                                                                                                                                                                                 
  if errors.Is(err, ebpf.ErrNotSupported) { ... }                                                                                                                                                                                            
                                                                                                                                                                                                                                             
  // After:                                                                                                                                                                                                                                  
  if errors.Is(err, ebpf.ErrNotSupported) || errors.Is(err, ebpf.ErrNotPermitted) { ... }                                                                                                                                                    
                                                                                                                                                                                                                                             
  // Or simply:                                                                                                                                                                                                                              
  if err != nil { ... }                                                           
  ```

### DIFF
--- a/btf/feature.go
+++ b/btf/feature.go
@@ -15,13 +15,14 @@ import (
 var haveBTF = internal.NewFeatureTest("BTF", func() error {
 	// 0-length anonymous integer
 	err := probeBTF(&Int{})
-	if errors.Is(err, unix.EINVAL) {
+	switch {
+	case errors.Is(err, unix.EINVAL):
 		return internal.ErrNotSupported
-	}
-	if errors.Is(err, unix.EPERM) {
+	case errors.Is(err, unix.EPERM):
 		return fmt.Errorf("%w: %w", internal.ErrNotPermitted, err)
+	default:
+		return err
 	}
-	return err
 }, "4.18")
 
 // haveMapBTF attempts to load a minimal BTF blob containing a Var. It is

### CODE REVIEW

File: btf/feature.go [hunk 1]
```diff
@@ -14,9 +15,12 @@ import (
...
 	// 0-length anonymous integer
 	err := probeBTF(&Int{})
-	if errors.Is(err, unix.EINVAL) || errors.Is(err, unix.EPERM) {
+	if errors.Is(err, unix.EINVAL) {
 		return internal.ErrNotSupported
 	}
+	if errors.Is(err, unix.EPERM) {
+		return fmt.Errorf("%w: %w", internal.ErrNotPermitted, err)
+	}
 	return err
```
florianl: just a minor style nit: we could turn it into a switch statement.
```suggestion
	switch {
	case errors.Is(err, unix.EINVAL):
		return internal.ErrNotSupported
	case errors.Is(err, unix.EPERM):
		return fmt.Errorf("%w: %w", internal.ErrNotPermitted, err)
	default:
		return err
	}
```

florianl (approved):
Just a comment on the style, that a switch statement could be used in favor of multiple if statements.
Otherwise, I'm in favor for this differentiation.

javiercardona-work:
> [ci / Run tests (Linux) (stable) (pull_request)](https://github.com/cilium/ebpf/actions/runs/21952351876/job/63410469409?pr=1949)Failing after 1m

This looks like a pre-existing flaky test related to GC finalization timing, not something caused by the changes in this PR.  Can we re-test?

btf/feature.go
```diff
@@ -34,11 +39,12 @@ var haveMapBTF = internal.NewFeatureTest("Map BTF (Var/Datasec)", func() error {
 	}
 
 	err := probeBTF(v)
-	if errors.Is(err, unix.EINVAL) || errors.Is(err, unix.EPERM) {
-		// Treat both EINVAL and EPERM as not supported: creating the map may still
-		// succeed without Btf* attrs.
+	if errors.Is(err, unix.EINVAL) {
```
florianl, line 42: For consistency, the style as mentioned in https://github.com/cilium/ebpf/pull/1949#discussion_r2793730744 should also be applied here.

btf/feature.go
```diff
@@ -56,9 +62,12 @@ var haveProgBTF = internal.NewFeatureTest("Program BTF (func/line_info)", func()
 	}
 
 	err := probeBTF(fn)
-	if errors.Is(err, unix.EINVAL) || errors.Is(err, unix.EPERM) {
+	if errors.Is(err, unix.EINVAL) {
```
florianl, line 65: For consistency, the style as mentioned in https://github.com/cilium/ebpf/pull/1949#discussion_r2793730744 should also be applied here.

link/syscalls.go
```diff
@@ -105,6 +106,11 @@ var haveProgQuery = internal.NewFeatureTest("BPF_PROG_QUERY", func() error {
 	if errors.Is(err, unix.EBADF) {
 		return nil
 	}
+	// EPERM means the kernel recognized the syscall but denied permission.
```
florianl, line 109: For consistency, the style as mentioned in https://github.com/cilium/ebpf/pull/1949#discussion_r2793730744 should also be applied here.

internal/feature_test.go
```diff
@@ -88,3 +89,14 @@ func TestFeatureTestNotSupportedOnOS(t *testing.T) {
 		qt.Assert(t, qt.ErrorIs(NewFeatureTest("foo", fn, "1.0")(), sentinel))
 	}
 }
+
+func TestErrNotPermitted(t *testing.T) {
```
rgo3, line 93: This test feels like we are testing Go standard library behavior instead of our own code, so I don't see a lot of value in this test.

ti-mo:
@javiercardona-work I think it may be better to drop this change altogether. I assume this PR came to be after a frustrating trial-and-error session for figuring out the minimum set of capabilities the Cilium agent needed in order to run correctly. Was there anything else on your side that motivated this change?

#1953 introduces accurate reporting on exactly which permission is missing when tokens are enabled, which would eliminate this guessing game. Probes will also start failing with an exact error instead of a simple EPERM. As Robin indicated, given the inconsistent nature of the timing/location of permission checks in the kernel, I don't think we can draw probe conclusions based on EPERM alone.

If a probe fails with EPERM, in the majority of cases this should be due to running in a userns/unpriv environment where tokens are used to enforce the principle of least privilege. Returning a sentinel in this case doesn't feel super useful, since the user will need to modify the sandbox configuration to provide the necessary caps if they want a conclusive probe result. There's not much the program itself can do to remediate this situation. If it wants to fall back to a different behaviour, there's already the EPERM signal.

Have you found specific cases in the Cilium agent where this type of fallback behaviour is necessary for running in userns?

ti-mo:
@javiercardona-work Please check what else is needed with https://github.com/cilium/ebpf/pull/1953 in the picture. If EPERM happens when tokens are in use, the error will hint at it. As of 6.17, the exact missing permission will be mentioned in the error message. That should make permission errors a lot more straightforward to figure out.

Closing this for now.

Issue #1929, opened 2026-01-07
modules BTF loading fails when CAP_SYS_ADMIN is missing


Andreagit97:
### Describe the bug

If during CO-RE relocations the library is not able to find the target type in the running kernel BTF, it searches for it in the modules' BTF. The downside of this approach is that loading modules' BTF requires CAP_SYS_ADMIN, and this prevents BPF applications from running with more granular capabilities (e.g. CAP_BPF) when they don't need to search types into modules' BTF. 
It's valid to reference non-existent types when their use is guarded by something like `if core_type_exists(...).`

This is the same issue I faced in the past with libbpf https://github.com/libbpf/libbpf/commit/3f33f9a6b8f77364ea410cf90df0349fa679826b. The fix there was to skip the modules' BTF loading phase in case of `EPERM`.

### How to reproduce

This is a repro of the issue. 

C part

```c
// go:build ignore
#include "vmlinux.h"
#include <bpf/bpf_core_read.h>
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_tracing.h>

char __license[] SEC("license") = "Dual MIT/GPL";

// this kfunc doesn't exist, it is a mocked name
extern int bpf_my_kfunc(int example) __ksym __weak;

SEC("fexit/do_mkdirat")
int BPF_PROG(cgroup_catcher, int dfd, struct filename *name, umode_t mode) {
	if(bpf_ksym_exists(bpf_my_kfunc)) {
		bpf_printk("bpf_my_kfunc is available");
	} else {
		bpf_printk("bpf_my_kfunc is not available");
	}
	return 0;
}
```

The userspace part

```go
package main

import (
	"log"

	"github.com/cilium/ebpf/rlimit"
)

//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -tags linux bpf ./bpf/main.c -- -I/usr/include/ -I../

func main() {
	if err := rlimit.RemoveMemlock(); err != nil {
		log.Fatal(err)
	}

	// Load pre-compiled programs and maps into the kernel.
	objs := bpfObjects{}
	if err := loadBpfObjects(&objs, nil); err != nil {
		log.Fatalf("loading objects: %v", err)
	}

	objs.Close()

	log.Println("Ebpf configuration completed.")
}
```

If you compile the above program and use the setcap binary, you will see the following

```bash
sudo setcap cap_bpf+ep ./main
```

error:

```txt
[...]

PR #1950, opened 2026-02-04
Allow loading weak kfuncs without `CAP_SYS_ADMIN`

dylandreimerink:
Currently when users attempt to load a program with a weak kfunc on a kernel that does not have that kfunc while lacking `CAP_SYS_ADMIN`, loading fails with the following error.

> fixing up kfuncs: finding kfunc in kernel: find target in modules:
> iterate modules: get next BTF ID: operation not permitted`

This happens because when we are unable to find the kfunc in the kernel BTF, we attempt to look for it in kernel modules. Iterating over kernel modules requires `CAP_SYS_ADMIN`, and we throw the error we get back.

This PR changes `findTargetInKernel` to wrap the `EPERM` error we get back when lacking `CAP_SYS_ADMIN` with `btf.ErrNotFound`. This allows the kfunc resolution logic to treat it the same way as if the kfunc was not found. For non weak kfuncs, this will still cause loading to fail, but for weak kfuncs, loading will now succeed.Currently 

The PR also includes a new test helper which allows us to drop capabilities to a specific set while running a callback so we can write tests that assert certain behavior works with reduced capabilities.

Fixes: #1929

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -1186,6 +1186,12 @@ func findTargetInKernel(typeName string, target *btf.Type, cache *btf.Cache) (*b
 	if errors.Is(err, btf.ErrNotFound) {
 		spec, module, err := findTargetInModule(typeName, target, cache)
 		if err != nil {
+			// EPERM may be returned when we do not have CAP_SYS_ADMIN.
+			// Wrap error with btf.ErrNotFound so callers can handle it accordingly.
+			if errors.Is(err, unix.EPERM) {
+				return spec, nil, fmt.Errorf("find target in modules: %w (%w)", btf.ErrNotFound, err)
+			}
+
 			return nil, nil, fmt.Errorf("find target in modules: %w", err)
 		}
 		return spec, module, nil


internal/testutils/cap.go
```diff
@@ -0,0 +1,50 @@
...
+
+	if !platform.IsLinux {
+		fn()
+		return
+	}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	originalCapset, err := unix.Capget()
```
ti-mo: I know this may seem insignificant, but could we use short identifiers wherever possible? At least for me, it significantly eases reading and mental overhead, and a lot of Go code is written this way, so familiarity and consistency goes a long way. We're currently in a function that deals with capabilities, so calling this `orig` would be sufficient.

internal/testutils/cap.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		t.Fatal("Can't get capabilities:", err)
+	}
+
+	var newCapset unix.CapUserData
```
ti-mo:
```suggestion
	var set unix.CapUserData
```

internal/testutils/cap.go
```diff
@@ -0,0 +1,50 @@
...
+	for _, cap := range caps {
+		newCapset.Effective |= 1 << uint(cap)
+	}
+	newCapset.Permitted = originalCapset.Permitted
+
+	if err := unix.Capset(newCapset); err != nil {
+		t.Fatal("Can't set capabilities:", err)
+	}
+
+	defer func() {
```
ti-mo: I'd omit the `defer` here, since `fn` isn't fallible.

internal/testutils/cap.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/cilium/ebpf/internal/platform"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+// WithCapabilities runs `fn` with only the given capabilities
+// in the effective set. This allows us to assert that certain operations
+// only require specific capabilities.
+//
```
ti-mo, line 25: Similar warning applies as in `NetNS.Do()`:

```
// The code in f and any code called by f must NOT call [runtime.LockOSThread],
// as this could leave the current goroutine permanently pinned to an OS thread.
// It must also not create any goroutines of its own, as that will result in a new
// OS thread being created that may or may not inherit the new capabilities of its
// parent, and will later be released into the schedulable pool of threads available
// for goroutine scheduling.
```

internal/testutils/cap.go
```diff
@@ -0,0 +1,50 @@
...
+// Warning: on non-linux platforms, this function calls through to `fn` without
+// side effects.
+func WithCapabilities(t testing.TB, caps []int, fn func()) {
```
ti-mo:
```suggestion
func WithCapabilities(tb testing.TB, caps []Capability, f func()) {
```

- The var is typically called `tb`
- Let's make this take a custom enum-like type for discoverability
- f for consistency with the NetNS.Do()

internal/testutils/cap.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/cilium/ebpf/internal/platform"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+// WithCapabilities runs `fn` with only the given capabilities
+// in the effective set. This allows us to assert that certain operations
+// only require specific capabilities.
+//
+// Warning: on non-linux platforms, this function calls through to `fn` without
```
ti-mo: Is there a strong reason for this behaviour? What's the intended behaviour on Windows? (no idea if this test is currently being run) I think this should probably issue a skip on other platforms.

prog.go
```diff
@@ -1186,6 +1186,12 @@ func findTargetInKernel(typeName string, target *btf.Type, cache *btf.Cache) (*b
 	if errors.Is(err, btf.ErrNotFound) {
 		spec, module, err := findTargetInModule(typeName, target, cache)
 		if err != nil {
+			// EPERM may be returned when we do not have CAP_SYS_ADMIN.
+			// Wrap error with btf.ErrNotFound so callers can handle it accordingly.
+			if errors.Is(err, unix.EPERM) {
+				return spec, nil, fmt.Errorf("find target in modules: %w: %w", btf.ErrNotFound, err)
```
ti-mo: If memory serves me right, we typically add extra context between parentheses:


```suggestion
				return spec, nil, fmt.Errorf("find target in modules: %w (%w)", btf.ErrNotFound, err)
```

elf_reader_test.go
```diff
@@ -711,24 +712,31 @@ func TestKfunc(t *testing.T) {
...
-	spec, err := LoadCollectionSpec(file)
-	if err != nil {
-		t.Fatal(err)
-	}
 
-	var obj struct {
-		Missing *Program `ebpf:"weak_kfunc_missing"`
-		Calling *Program `ebpf:"call_weak_kfunc"`
-	}
+	// CAP_SYS_ADMIN is required to load kfuncs implemented in kernel modules.
```
ti-mo, line 715: Does it make sense to test both with and without capabilities? I'm struggling to come up with fun ways we could let a regression slip if we don't, but I'm sure there's a case.

internal/unix/cap_other.go
```diff
@@ -0,0 +1,29 @@
...
+package unix
+
+const (
+	LINUX_CAPABILITY_VERSION_3 = 0
+	CAP_SYS_ADMIN
+	CAP_BPF
+	CAP_PERFMON
+)
+
+type CapUserHeader struct {
```
ti-mo: This is unused in the non-Linux stub.

internal/unix/cap_linux.go
ti-mo, line 1: - Could you elaborate a bit on where this code comes from? The `x/sys/unix` types all seem to have 32-bit fields, I guess maybe v3 has a different/extended ABI?
- This package should only contain type aliases and minimal wrappers to the stdlib functions. If we need more, let's put that closer to the consuming code. In this case, maybe x/sys could actually warrant some extension, not sure.
- Having multiple type and function definitions like `CapUserData` and `Capget()` that are similar-but-not-quite-the-same as x/sys is pretty confusing. I'd name these differently (and unexport them, signaling that they're internal types) and move them to testutils.

dylandreimerink: > Could you elaborate a bit on where this code comes from?

The code I wrote myself. I used the go libcap library as reference https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap

>  The x/sys/unix types all seem to have 32-bit fields, I guess maybe v3 has a different/extended ABI?

Yes. So, originally the `capset/capget` syscall used 32-bit numbers, but these got exhausted. So they made v2 and v3, you pass in the magic number as secret handshake, and now the kernel starts writing two `unix.CapUserData`'s back for a total of 64 bits. But they are interleaved, so this wrapper mostly reconstructs actual 64-bit numbers so its easier to work with.

This is particularly relevant for us since CAP_BPF is bit 39, and thus can only be worked with if you have the full 64 bits.

So I mostly did it this way because to my mind, the original syscall was not useful without this additional work, at least not for our purposes.

> This package should only contain type aliases and minimal wrappers to the stdlib functions. If we need more, let's put that closer to the consuming code.

Yea, that is fair, for the sake of clarity and consistency.

> Having multiple type and function definitions like CapUserData and Capget() that are similar-but-not-quite-the-same as x/sys is pretty confusing. I'd name these differently (and unexport them, signaling that they're internal types) and move them to testutils.

I can do that.

internal/unix/cap_other.go
dylandreimerink: Stupid leftover from before I removed the header parameter from the wrapper. It will need to stay if we are moving back to giving the wrapper the exact same signature as the stdlib function.

elf_reader_test.go
dylandreimerink: We have `TestKfuncKmod`, which runs with `CAP_SYS_ADMIN`, and covers the bit of code we skip by not doing module lookup for weak kfuncs.

In general, we should probably identify methods / features of the library that we want to ensure work under reduced capabilities and ensure this by testing them without those.

internal/testutils/cap.go
dylandreimerink: The reason for this is that I am secretly expecting to use this util in more tests in the future, since it makes sense to me that we scope out which parts of our API require which capabilities and then ensure no regressions by asserting it in tests.

The way we did the Windows integration is that we swap out implementation at the BPF syscall layer. So I was imagining that there may be operations that work on both Linux and Windows, that require certain capabilities on Linux which we would like to assert, but we would also still like to run the tests on Windows if possible.

internal/testutils/cap.go
dylandreimerink: I feel like its a matter of taste and context. 

I personally am not that big of a fan of things like `r io.Reader`, especially when a function has 3 readers at the same time. `i` and `n` for a loop iterator or number of bytes returned from a reader are more acceptable because they are short lived (hopefully). But my general rule is that the longer the lifetime of a variable, the more descriptive its name so you don't have to go to definition every time. And as a non native English speaker I sometimes stumble over compacted names like `xfrm`, `insn` and `fmt`, at least the first view times.

Anyway, end of my manifesto.

In this case, the function fits on my screen, so I am inclined to agree. But I don't agree with "shorter whenever possible" as general guidance everywhere.

ti-mo: > But my general rule is that the longer the lifetime of a variable, the more descriptive its name [...] I don't agree with "shorter whenever possible" as general guidance everywhere

Totally! That's not what I was trying to say. I read this in some early slides on Go conventions by the Go team, and still see it get cited often. It was probably Rob Pike: https://doc.cat-v.org/bell_labs/pikestyle.

I think we're on the same page, thanks.

PR #1951, opened 2026-02-14
Update getting started document

shibukawa:
Thank you for your grate product and effort.

* apt Package:

  https://packages.debian.org/search?keywords=llvm-strip&searchon=names&suite=all&section=all
  
  https://packages.ubuntu.com/search?keywords=llvm-strip&searchon=names&suite=all&section=all
  
  Both Debian and Ubuntu current versions don't have llvm-strip package and llvm seems to ship llvm-strip command.

* link option:

  I tried to run on UTM (macOS's virtual machine) with Ubuntu 25.10, I need to add flag to run.

### DIFF
--- a/docs/ebpf/guides/getting-started.md
+++ b/docs/ebpf/guides/getting-started.md
@@ -214,6 +214,14 @@ standard library here.
    `CountPackets`) eBPF program with `eth0`. This returns a {{
    godoc('link/Link') }} abstraction.
 
+   Note: On virtualized environments or network interfaces whose drivers do not
+   support native XDP, attaching may fail with "operation not supported".
+   In such cases, you can use generic mode instead:
+
+       Flags: link.XDPGenericMode
+
+Generic mode has lower performance but works on a wider range of interfaces.
+
 1. Close the file descriptor of the Program-to-interface association. Note that
    this will stop the Program from executing on incoming packets if the Link was
    not {{ godoc('link/Link.Pin') }}ed to the bpf file system.
--- a/docs/ebpf/guides/getting-started.md
+++ b/docs/ebpf/guides/getting-started.md
@@ -220,7 +220,7 @@ standard library here.
 
        Flags: link.XDPGenericMode
 
-Generic mode has lower performance but works on a wider range of interfaces.
+   Generic mode has lower performance but works on a wider range of interfaces.
 
 1. Close the file descriptor of the Program-to-interface association. Note that
    this will stop the Program from executing on incoming packets if the Link was
--- a/docs/ebpf/guides/getting-started.md
+++ b/docs/ebpf/guides/getting-started.md
@@ -19,7 +19,7 @@ more.
     To follow along with the example, you'll need:
 
     * Linux kernel version 5.7 or later, for bpf_link support
-    * LLVM 11 or later [^1] (`clang` and `llvm`)
+    * LLVM 11 or later [^1] (`clang` and `llvm-strip`)
     * libbpf headers [^2]
     * Linux kernel headers [^3]
     * Go compiler version supported by {{ proj }}'s Go module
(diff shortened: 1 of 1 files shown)


docs/ebpf/guides/getting-started.md
```diff
@@ -19,7 +19,7 @@ more.
     To follow along with the example, you'll need:
 
     * Linux kernel version 5.7 or later, for bpf_link support
-    * LLVM 11 or later [^1] (`clang` and `llvm-strip`)
```
dylandreimerink, line 22: `llvm-strip` is the actual utility we use in our makefile hence why it is listed here. Its not a mistake. https://github.com/cilium/ebpf/blob/2e4cf1cc3c8ff7447d399c4fc0ec797dcf73b661/Makefile#L5

docs/ebpf/guides/getting-started.md
```diff
@@ -214,6 +214,14 @@ standard library here.
    `CountPackets`) eBPF program with `eth0`. This returns a {{
    godoc('link/Link') }} abstraction.
 
+   Note: On virtualized environments or network interfaces whose drivers do not
+   support native XDP, attaching may fail with "operation not supported".
+   In such cases, you can use generic mode instead:
+
+       Flags: link.XDPGenericMode
+
+   Generic mode has lower performance but works on a wider range of interfaces.
```
dylandreimerink, line 224: I am not entirely sure if we want to encourage the usage of generic mode. XDP driver support for Virtio support exists since v4.10, and veth support since v4.19. A lot of more specialized virtual NICs have gained support in later kernels https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_XDP/#driver-support.

So for the purposes of testing a user can always use a veth pair (via a container).

Generic mode has poor test coverage. So in production, when native XDP is not available, users should fallback to `tc` / `BPF_PROG_TYPE_SCHED_CLS` programs instead, which have broader capabilities, are better tested, and have comparable performance.

Long story short. I don't think this is advice we want to give people, especially not those that don't know any better.

dylandreimerink (changes requested):

docs/ebpf/guides/getting-started.md
shibukawa: OK. I my intention was apt package name, not program name. I will fix it.

docs/ebpf/guides/getting-started.md
```diff
@@ -28,7 +28,8 @@ more.
     Use `clang --version` to check which version of LLVM you have installed.
     Refer to your distribution's package index to finding the right packages to
     install, as this tends to vary wildly across distributions. Some
-    distributions ship `clang` and `llvm-strip` in separate packages.
+    distributions ship `clang` and `llvm-strip` in separate packages
```
ti-mo, line 31: Thanks for the pointer re: debian's llvm package. I just checked Fedora 44, and they also seem to ship llvm-strip in the `llvm` package (it's just a symlink to llvm-objcopy anyway, which is usually included). Let's remove the `Some distributions ship...` sentence altogether then, it'll just confuse new users.

docs/ebpf/guides/getting-started.md
ti-mo: Completely agree in principle, but with proper warnings attached, the user can make this decision on their own. The text you proposed is in a tooltip, though, so empty newlines break that. I'd make this an actual note below the `Save this file as ...` under the snippet.

```
!!! tip ""
    Some network interface drivers don't support native XDP, making this
    example fail with `operation not supported`. Try using generic mode by
    setting:
    
    ``` go
    Flags: link.XDPGenericMode
    ```
    
    Note that generic mode performs poorly and should not be used in production deployments.
```

ti-mo (changes requested):

ti-mo:
Closing some stale PRs, ping here if you want to pick this up again.

PR #1952, opened 2026-02-16
elf: add regression test for loading objects without BTF

### LINKED ISSUE #1810: Ensure the ELF reader works for objects built without BTF

sfmqrb:
This adds a small regression test for loading ELF objects that don't contain BTF.

Context: issue #1810 points out that loading BPF objects without BTF has historically worked, and we should keep coverage for that behavior.

### What this changes
- Adds two no-BTF test fixtures:
  - `testdata/iproute2_map_compat-nobtf-el.elf`
  - `testdata/iproute2_map_compat-nobtf-eb.elf`
- Adds `TestLoadCollectionSpecWithoutBTF` in `elf_reader_test.go`.

The test verifies that:
- `LoadCollectionSpec()` succeeds for no-BTF objects,
- `spec.Types` is nil,
- expected map metadata (`hash_map`) is still parsed.

### Why this fixture
I used the `iproute2_map_compat` object as a stable legacy-style map definition source, then removed `.BTF` and `.BTF.ext` sections with `llvm-objcopy`.

This keeps the test focused on ELF-reader behavior without requiring BTF map definitions.

Fixes #1810.


dylandreimerink:
Additionally. PR #1939 is also trying to do exactly this. Perhaps it makes sense to see if these changes are still needed if it merges.

ti-mo:
Closing as duplicate of https://github.com/cilium/ebpf/pull/1939.

PR #1948, opened 2026-02-02
Add BPF token support for program and map loading                                                                                                  

javiercardona-work:
Add support for loading BPF programs and creating maps using BPF tokens,
  enabling unprivileged BPF operations in user namespaces with delegated
  capabilities.   
                                                                                                                                       
  Changes:                                                                                                                                                                                                                                                                     
                                                                   
   - Add TokenFD field to ProgramOptions and MapOptions                                                                                                                                                                                                                           
   - When set, pass token to kernel with BPF_F_TOKEN_FD flag during syscalls                                                                                                                                                                                                      
   - Add internal/token package for global token storage
   - Add `features.SetGlobalToken()`/`GetGlobalToken()` for feature probe support
                                                                   
  Since the original PR, the following commits were added:
                                                                                                                                       
  - Thread BPF token through feature probes: Pass the global token to
  all feature probes (program type, map type, BTF, link, and program
  helper probes) so that capability detection works correctly inside user
  namespaces with delegated BPF permissions.                                                                                           
  - Add BPF token integration test program: Standalone test
  (`examples/token_test/`) that creates a BPF token, optionally drops                                                                                                                                                                                                            
  `CAP_BPF`/`CAP_SYS_ADMIN`/`CAP_PERFMON`, and exercises feature probes,                                                                     
  program loading, and map creation. Validated on a container with
  bpf_token delegation (kernel 6.16.1, `unprivileged_bpf_disabled=1`).            
  
  This is the output of the test program when run on a container with delegated BPF token enabled:
  
  ```
$ /tmp/token-test --drop-caps /run/tw/bpf
Creating BPF token from /run/tw/bpf...
BPF token created: fd=4

Dropping CAP_BPF, CAP_SYS_ADMIN, CAP_PERFMON...
Capabilities dropped. BPF access now depends solely on token.

=== Feature Probes (with token) ===
  prog SocketFilter         supported
  prog Kprobe               supported
  prog SchedCLS             supported
  prog XDP                  supported
  prog TracePoint           supported
  prog CGroupSKB            supported
  map  Hash                 supported
  map  Array                supported
  map  PerfEventArray       supported
  map  LRUHash              supported
  map  RingBuf              supported

=== Program Load (with token) ===
  program load: OK (fd=3)

=== Map Create (with token) ===
  map create: OK (fd=3)

=== Without Token (uncached probe) ===
  prog PerfEvent (no token):     detect support for PerfEvent: load program: operation not permitted
[...]

PR #1953, opened 2026-02-17
BPF token support and testing infrastructure 

dylandreimerink:
This PR introduces test utilities that allows us to write tests for BPF tokens. This `RunWithToken` utility function is largely based on the [test code from the kernel selftests](https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/prog_tests/token.c) but modified to work with Go and its runtime.

The individual commit messages contain more details. The high level is that we can write tests (and subtests) where we start by calling `RunWithToken` with a set of delegated permissions. The rest of the test body will be executed in an environment with a BPFFS mounted with those permissions.

This PR also includes BPF token support. This takes the "ambient token" approach, where we do not add any new API surface, but we automatically use tokens when they are available. This allows us to experiment with tokens without committing to an API. We can always add more controls in the future.

Instead of relying on hardcoding search paths or environment variables, this implementation scans the `/proc/self/mountinfo` procfile to find all BPFFS instances mounted in the processes mount namespace. This means we are able to use any BPFFS path, and even multiple BPFFS'es, for example if one FS has no permissions but holds pinned objects, and a secondary BPFFS is used to the delegated tokens. However, there is no method to select a preferred one, if multiple BPFFS'es with delegated privileges exist (which should be uncommon).

A nice side effect of reading the mount info is that we also know which permissions are delegated which is used in this implementation to enhance error messages when using tokens but the token lacks certain permissions.

The main purpose of this PR was to implement the testing utilities, and the PR is structured such that we can drop the actual token implementation if any of the other proposals is more appealing then this one.

Note: This PR currently has cherry-picked a commit from #1950 for its capability syscalls. This PR should be rebased once #1950 is merged.


javiercardona-work:
Thanks for doing this.  If you end up landing this, I will withdraw #1948.

I tested this PR in our environment, and it works as documented.  In case it can be useful, I'm attaching the test program we used: [ambient_token_test.go.zip](https://github.com/user-attachments/files/25375252/ambient_token_test.go.zip)

This runs in our namespaced container with `CAP_BPF` permissions and a delegated BPF mountpoint (`/run/tw/bpf`).  The mountpoint is automatically detected:

```
$ /tmp/ambient-token-test            
ambient-token-test v7

=== Diagnostics ===
  user namespace: user:[4026535428]
  unprivileged_bpf_disabled: 1

  Capabilities before:
  CapInh:       0000000000000000
  CapPrm:       000001c7a9ac7dff
  CapEff:       000001c7a9ac7dff
  CapBnd:       000001c7a9ac7dff
  CapAmb:       0000000000000000

=== Token Create ===
  token.Create(): OK (fd=6, mount=/run/tw/bpf)

=== Feature Probes (ambient token) ===
  prog SocketFilter         supported
  prog Kprobe               supported
  prog SchedCLS             supported
  prog XDP                  supported
  prog TracePoint           supported
  prog CGroupSKB            supported
  map  Hash                 supported
  map  Array                supported
  map  PerfEventArray       supported
  map  LRUHash              supported
  map  RingBuf              supported

=== Map Create (ambient token) ===
  map create: OK (fd=3)

=== Program Load (ambient token) ===
  program load: OK (fd=3)

Done.
```

When running it without `CAP_BPF`, token creation and all subsequent bpf operations fail, as expected:

```
$ /tmp/ambient-token-test --drop-caps
ambient-token-test v7

=== Diagnostics ===
  user namespace: user:[4026535428]
  unprivileged_bpf_disabled: 1

  Capabilities before:
  CapInh:       0000000000000000
  CapPrm:       000001c7a9ac7dff
  CapEff:       000001c7a9ac7dff
  CapBnd:       000001c7a9ac7dff
  CapAmb:       0000000000000000

  Dropping CAP_BPF, CAP_SYS_ADMIN, CAP_PERFMON...
  Capabilities after:
  CapInh:       0000000000000000
  CapPrm:       00000107a98c7dff
  CapEff:       00000107a98c7dff
  CapBnd:       00000107a98c7dff
  CapAmb:       0000000000000000

=== Token Create ===
  token.Create(): BPF Token is not available

=== Feature Probes (ambient token) ===
  prog SocketFilter         detect support for SocketFilter: load program: operation not permitted
  prog Kprobe               detect support for Kprobe: load program: operation not permitted
  prog SchedCLS             detect support for SchedCLS: load program: operation not permitted
  prog XDP                  detect support for XDP: load program: operation not permitted
  prog TracePoint           detect support for TracePoint: load program: operation not permitted
  prog CGroupSKB            detect support for CGroupSKB: load program: operation not permitted
  map  Hash                 detect support for Hash: operation not permitted
  map  Array                detect support for Array: operation not permitted
  map  PerfEventArray       detect support for PerfEventArray: operation not permitted
  map  LRUHash              detect support for LRUHash: operation not permitted
  map  RingBuf              detect support for RingBuf: operation not permitted

=== Map Create (ambient token) ===
  map create: FAILED (creating map: map create: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock))

=== Program Load (ambient token) ===
  program load: FAILED (load program: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock))

Done.
```

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,308 @@
...
+		Inheritable: caps,
+	})
+	if err != nil {
+		panic(fmt.Sprintf("capset: %v", err))
+	}
+
+	// Mark next exec as stage 2, to actually execute the tests.
+	env := os.Environ()
+	i := slices.IndexFunc(env, func(s string) bool { return strings.HasPrefix(s, TOKEN_SUBTEST+"=") })
+	env = slices.Replace(env, i, i, TOKEN_SUBTEST+"=2")
```
anryko:
```suggestion
	env = slices.Replace(env, i, i+1, TOKEN_SUBTEST+"=2")
```

map.go
```diff
@@ -670,6 +685,15 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 	}
 
 	if errors.Is(err, unix.EPERM) {
+		if tok != nil {
+			if !slices.Contains(tok.Mount.Cmds, sys.BPF_MAP_CREATE) {
+				return fmt.Errorf("map create: %w (BPF token does not allow BPF command BPF_MAP_CREATE)", err)
+			}
+			if !slices.Contains(tok.Mount.Maps, attr.MapType) {
+				return fmt.Errorf("map create: %w (BPF token does not allow map type)", err)
```
anryko: Add `EncodeConstant` of `attr.MapType` to the error message to make it more actionable.

prog.go
```diff
@@ -514,6 +529,19 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
 	switch {
 	case errors.Is(err, unix.EPERM):
 		if len(logBuf) > 0 && logBuf[0] == 0 {
+			// When a token was used, check if the failure could have been caused by missing permissions on the token.
+			if tok != nil {
+				if !slices.Contains(tok.Mount.Cmds, sys.BPF_PROG_LOAD) {
+					return nil, fmt.Errorf("load program: %w (BPF token does not allow BPF command BPF_PROG_LOAD)", err)
+				}
+				if !slices.Contains(tok.Mount.Progs, attr.ProgType) {
+					return nil, fmt.Errorf("load program: %w (BPF token does not allow BPF program type)", err)
```
anryko: Add `EncodeConstant` of `attr.ProgType` to the error message.

prog.go
```diff
@@ -514,6 +529,19 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
...
+				}
+				if !slices.Contains(tok.Mount.AttachTypes, attr.ExpectedAttachType) {
+					return nil, fmt.Errorf("load program: %w (BPF token does not allow BPF attach type)", err)
```
anryko: Add `EncodeConstant` of `attr.ExpectedAttachType` to the error message.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,308 @@
...
+	// Inherit the test's standard output and error so that we can see the subtest's logs and errors in real time.
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+
+	// Create a socket pair for communication between the parent and child process.
+	pair, err := unix.Socketpair(unix.AF_UNIX, unix.SOCK_STREAM, 0)
+	if err != nil {
+		tb.Fatal(err)
+	}
+	parent, child := pair[0], pair[1]
```
anryko, line 99:
```suggestion
	parent, child := pair[0], pair[1]
	defer unix.Close(parent)
```

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,308 @@
...
+		AmbientCaps: []uintptr{unix.CAP_SYS_ADMIN, unix.CAP_DAC_READ_SEARCH, unix.CAP_BPF},
+		UidMappings: uidMappings,
+		GidMappings: gidMappings,
+	}
+
+	// Start the child process in the background.
+	err = cmd.Start()
+	if err != nil {
+		tb.Fatal(err)
+	}
```
anryko, line 128:
```suggestion
    if err := cmd.Start(); err != nil {
        unix.Close(child)
        tb.Fatal(err)
    }
    unix.Close(child)
```

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,308 @@
...
+
+	return true
+}
+
+func parseUIDsGIDs() (uids, gids []syscall.SysProcIDMap, err error) {
+	// Files can only be read if the owning user and group are mapped into the user namespace.
+	// So find out all uids and gids of users on the system and map them into the user namespace of the child process.
```
anryko, line 158: Could you clarify why this is needed. Wouldn't we need just the UID and GID of the current user executing the test?

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,308 @@
...
+	env = slices.Replace(env, i, i, TOKEN_SUBTEST+"=2")
+
+	// Re-exec ourselves so that its truly as if all of the setup was done before the test started.
+	err = syscall.Exec(os.Args[0], os.Args, env)
+	if err != nil {
+		panic(fmt.Sprintf("exec: %v", err))
+	}
+}
+
+// recvFDs receives file descriptors from the provided socket. It expects to receive `fdRecvCnt“ fds.
```
anryko, line 277: Is there a usecase for receiving more than 1 FD? If not, code could be simplified.

internal/token/token.go
```diff
@@ -0,0 +1,249 @@
...
+type Token struct {
+	*sys.FD
+	Mount BPFFSMount
+}
+
+// Create attempts to create a BPF token, it does so by finding all mounted BPFFS filesystems, and trying to create a
+// token on each of them until it finds one that works.
+//
+// If a token cannot be created, it returns [ErrTokenNotAvailable]. If the OS is not Linux, it returns
+// [internal.ErrNotSupportedOnOS].
```
anryko: If we have a hypothetical situation with two bpffs mounts. One with delegation and one without. And assume that we have all the necessary CAPs. This approach will always succeed and return a token. Then, because of how the rest of the code is written, we always rely on that token (instead of fallback to it on perm issue). So we scope by least privileged access method in this case instead of the most privileged. I did not actually test this... but it seems to me that this is how this works.

internal/testutils/token_linux.go
dylandreimerink: No I do not, but `ParseSocketControlMessage` and `ParseUnixRights` both return slices. So it just makes sense to do it this way. Otherwise I would just be adding more logic to artificially cripple the function.

internal/testutils/token_linux.go
dylandreimerink: Using the UID/GID of the current user was also my initial idea. However, we use [vimto](https://github.com/lmb/vimto) both in local development and CI to run our tests with arbitrary kernels. The host filesystem is mounted via p9 which preserves the UIDs/GIDs of the actual host, but inside the VM the user is root (0 / 0). So can't simply check the current UID/GID.

Ultimately, the only reason we need it is to read the files that are part of this repository (and perhaps sources of the stdlib) for stack traces and testdata. So perhaps we could scan the UIDs/GIDs used in the current working directory instead.

map.go
dylandreimerink: Yea, I did think about it. Using `EncodeConstant` isn't even nessesery, its a compatibility thing for windows. We can just print the integer value of the enum, but that does not help most people.

I will look into hooking up the enums to the stringer tool so we can call `.String()` and use that instead.

internal/token/token.go
dylandreimerink: Yes, effectively. Token creation would fail for the BPFFS without delegated permissions, and succeed for the one with delegated permissions. So we would have a token, and use it. When a token is given, the kernel will ignore any capabilities the process may have, at least as far as I understand it.

So yes, with this logic, we would not fallback to using the native capabilities when a token is present. But then again, having CAPs and tokens is fundamentally a broken setup, at least from a security perspective.

map.go
dylandreimerink: nevermind, the `MapType` in the root package already has a stringer, and it has the same enum value, so we can just cast to get a string.

dylandreimerink:
Rebased on main and updated to match changes made in https://github.com/cilium/ebpf/pull/1950. Marking ready for review.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+func parseUIDsGIDs() (uids, gids []syscall.SysProcIDMap, err error) {
+	// Files can only be read if the owning user and group are mapped into the user namespace.
+	// So find out all uids and gids of users on the system and map them into the user namespace of the child process.
+	passwdFile, err := os.ReadFile("/etc/passwd")
+	if err != nil {
+		return nil, nil, fmt.Errorf("read /etc/passwd: %v", err)
+	}
+
+	for line := range strings.Lines(string(passwdFile)) {
+		cols := strings.Split(line, ":")
```
javiercardona-work, line 165: malformed passwd file might have lines with len(cols) < 4.  Maybe check?
```suggestion
		cols := strings.Split(line, ":")
		if len(cols) < 4 {
		    continue
		}
```

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+		gid, err := strconv.Atoi(cols[3])
+		if err != nil {
+			return nil, nil, fmt.Errorf("parse gid: %v", err)
+		}
+		gids = append(gids, syscall.SysProcIDMap{
+			ContainerID: gid,
+			HostID:      gid,
+			Size:        1,
+		})
+	}
```
javiercardona-work, line 194: As multiple users may have the same GID, the gid mapping constructed here might have multiple entries for the same GID.  I believe the kernel will consider these overlapping entries and reject the mounts.  Maybe dedup as...

```suggestion
		gid, err := strconv.Atoi(cols[3])
		seen := map[int]bool{}
		if err != nil {
			return nil, nil, fmt.Errorf("parse gid: %v", err)
		}
		if !seen[gid] {                                                                                                                                                                                                                                                              
                    seen[gid] = true  
		    gids = append(gids, syscall.SysProcIDMap{
			    ContainerID: gid,
			    HostID:      gid,
			    Size:        1,
		    })
		}
```

internal/testutils/token_linux.go
anryko: Then root will be able to read files, should be no need for the mappings. Unless, you want to log the actual file owner name and group...

internal/testutils/token_linux.go
anryko: Thanks! You can resolve this.

internal/token/token.go
anryko: Got it. If we are able to acquire the token, we always assume that this the intended use. In that case I would suggest to always check the default path first (`/sys/fs/bpf`) and only after fallback to parsing `/proc/self/mountinfo`. Also it would make sense to cache the discovered token.

dylandreimerink: >  In that case I would suggest to always check the default path first (`/sys/fs/bpf`) and only after fallback to parsing `/proc/self/mountinfo`.

I am not sure if we need to be solving for having 2 bpffs'es with delegated permissions just yet. 

> Also it would make sense to cache the discovered token.

Yes. I have been thinking about this. In the current solution, we do 2 syscalls to get a token every time we load a program / map, so that could add up over time.

We could move the token creation to the collection loader. But we have a few places such as feature probing where no collections are involved.

We could do some sort of straight up cache with a global variable, but it means we hold onto a token file descriptor forever. We usually are very strict about that sort of thing in the library. I had been thinking that perhaps we can use a [weak](https://pkg.go.dev/weak) pointer, combined with the normal sys.FD cleanup logic. That way we hold onto tokens in-between GC cycles, but not forever. 

In the end I decided to leave that for another day, perhaps I am overthinking it.

internal/testutils/token_linux.go
dylandreimerink: > Then root will be able to read files, should be no need for the mappings.

You would think so, I tried this. But without the mapping, even root with `CAP_DAC_READ_SEARCH` gets a `-EPERM` when it attempts to read the files.

I have not dug into the kernel logic, but to me it seems like there is some sort of protection to do with user namespaces that makes it so we cannot access files owned by users outside of the user namespace.

internal/testutils/token_linux.go
dylandreimerink: The man page `man 5 passwd` states:
> /etc/passwd contains one line for each user account, with seven fields delimited by colons (“:”).

So we can expect up to 7 fields. We only need access to UID and GID, which are the third and forth fields (index 2 and 3). So I don't think we can change this check. If we don't have at least 4 fields, we can't handle it.

internal/token/token.go
anryko: It should be safe to assume that if `/proc/self/mountinfo` file was not changed, the token is valid; and reparse and recreate the token otherwise.

internal/testutils/token_linux.go
anryko: Got it. Then I think it's fine as is for the testutils. Walking the paths will result in less mappings but heavier and more fragile. Thank you for the explanations. You can resolve this thread.

internal/testutils/token_linux.go
javiercardona-work: I was referring to "malformed" files (i.e. due to human error, disk corruption, or malicious activity).  But I suppose if that happens, we have other things to worry about.

javiercardona-work:
@rgo3 I there something I can do to help merge this PR?  Happy to run additional tests on my environment.

dylandreimerink:
> I there something I can do to help merge this PR?

We are waiting for @ti-mo to review, since I made the PR, and he is the only other project maintainer. But Timo is a busy man, with a very long todo list, so this just takes some time.

All reviews are welcome, but Timo's approval determines merge-ability of the PR.

javiercardona-work:
Hi @ti-mo,

Could you provide a rough estimate on when this might be ready for merge? I’m curious if there are any outstanding design concerns we should address, or if it’s primarily a matter of finding an opening in your schedule. Our team is excited to move away from our internal fork and adopt @dylandreimerink better implementation for this feature, so we’re just looking to coordinate our timeline.  

Thanks!
Javier

ti-mo:
Hi Javier,

This PR is the result of internal discussions, so don't expect any departures from the overall architecture, if that makes things a bit more predictable for you.

For the initial implementation, we want there to not be any API changes, which should make this branch easy to carry in your Cilium fork since there should be no token-related Cilium changes required. I'll start working on this on Monday and it will be merged when it's done.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+// RunWithToken sets up an environment with a BPFFS with the provided delegated permissions.
+//
+// When called, the calling test method is re-executed in a subprocess with the environment set up. RunWithToken
+// returns true in the parent process, and false in the child process, so that the test can distinguish between them and
+// only run the actual test logic in the child process.
+func RunWithToken(tb testing.TB, delegated Delegated) bool {
+	tb.Helper()
+
+	if !platform.IsLinux {
+		tb.Skip("Tokens only work on linux")
```
ti-mo: Nit: `BPF tokens only work on Linux`

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+		"-test.run=^" + tb.Name() + "$",
+	}
+
+	// If we are running in verbose mode, pass the flag to the subtest as well.
+	if testing.Verbose() {
+		args = append(args, "-test.v")
+	}
+
+	// Call the current test binary (os.Args[0])
+	cmd := exec.CommandContext(tb.Context(), os.Args[0], args...)
```
ti-mo: Nit: I think using `os.Executable` should be more robust, at least it promises an absolute path. IIRC args[0] can be e.g. set to a symlink in a local directory or a shell alias, which may not be visible from the context of the calling program. (thinking about our NixOS friends here)

internal/cmd/gentypes/main.go
```diff
@@ -233,6 +234,12 @@ import (
 	enumTypes := make(map[string]btf.Type)
 
 	outputEnum := func(goType string, t *btf.Enum) error {
+		for i, value := range t.Values {
+			if strings.HasPrefix(value.Name, "__MAX") {
```
ti-mo: `strings.Replace(value.Name, "__MAX", "MAX", 1))`? And maybe add a small comment about why it's necessary.

internal/cmd/gentypes/main.go
```diff
@@ -249,6 +256,20 @@ import (
 		w.WriteString(decl)
 		w.WriteRune('\n')
 
+		w.WriteString(fmt.Sprintf("func %sFromString(s string) (%s, error) {\n", goType, goType))
+		w.WriteString("\tswitch s {\n")
+		for _, value := range t.Values {
+			if t.Signed {
+				fmt.Fprintf(w, "\tcase %q:\n\t\treturn %s(%d), nil\n", value.Name, goType, int64(value.Value))
```
ti-mo: Maybe clearer to return the constant itself instead of a typed integer? e.g. `return BPF_ADJ_ROOM_NET, nil`. If not desirable, let's document the reason.

internal/cmd/gentypes/main.go
```diff
@@ -249,6 +256,20 @@ import (
 		w.WriteString(decl)
 		w.WriteRune('\n')
 
+		w.WriteString(fmt.Sprintf("func %sFromString(s string) (%s, error) {\n", goType, goType))
```
ti-mo: Make this its own function, please.

internal/cmd/gentypes/main.go
```diff
@@ -249,6 +256,20 @@ import (
...
+		w.WriteString("\tswitch s {\n")
```
ti-mo: The tabs in these strings make this quite hard to read. Note the call to `WriteFormatted` at the end, let that take care of formatting. Only some `\n`s may be needed here and there. For sequential statements, a single Fprintf usually does, fmt will split it.

btf/handle.go
```diff
@@ -99,6 +116,12 @@ func NewHandleFromRawBTF(btf []byte) (*Handle, error) {
 		attr.BtfLogLevel = 1
 	}
 
+	if errors.Is(err, unix.EPERM) {
+		if tok != nil && !slices.Contains(tok.Mount.Cmds, sys.BPF_BTF_LOAD) {
```
ti-mo: I'd implement permission queries against Token directly, e.g. `func Token.AllowedCmd(cmd sys.Cmd) error`, `Token.AllowedMap() error`, .. These methods could tolerate nil receivers to make the repeated `tok != nil` checks unnecessary.

Return value semantics:
- nil == nil Token, or Cmd/Map/Prog type is allowed
- error == the open-coded error string you're formatting on the subsequent line (`BPF token does not allow BPF command BPF_BTF_LOAD`). I wouldn't make this wrap a sentinel for now.

Then the call site looks something like:
```go
if errors.Is(err, unix.EPERM) {
	if cerr := tok.AllowedCmd(sys.BPF_BTF_LOAD); cerr != nil {
		return nil, fmt.Errorf("%w: %w", cerr, err)
	}
}
```

We could wrap both errors with %w so we don't have to update any call sites if we want to add a sentinel later. I think we typically put the underlying error last in the stack/string in this scenario. I wouldn't put the reason between brackets since at this point we're 100% certain the syscall will fail due to missing caps; it's not just a hint.

internal/token/token.go
```diff
@@ -0,0 +1,249 @@
...
+			Mount: mount,
+		}, nil
+	}
+
+	return nil, fmt.Errorf("%w: %w", ErrTokenNotAvailable, errs)
+}
+
+// BPFFSMount represents a mounted BPF filesystem. It contains information about the mount point and the permissions
+// that have been delegated to it.
+type BPFFSMount struct {
```
ti-mo: Unexport this type and its fields if possible.

internal/testutils/token_linux.go
ti-mo: This a testutil, so we can make more assumptions about the environment than we can for production code and we don't need to be as defensive. If passwd is borked, there's not much we can do indeed. It wouldn't hurt to throw a bounds check in there, though, if only to ease debugging.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+
+func init() {
+	if stage, ok := os.LookupEnv(TOKEN_SUBTEST); ok {
+		// If this is stage 2 of the subtest, we let the tests run.
+		if stage != "1" {
+			return
+		}
+
+		// If this is stage 1 of the subtest, we setup the environment and then re-exec ourselves.
+		setupSubtest()
```
ti-mo: The logic here is confusingly inverted. Why not `if stage == "1" { setupSubtest() } return`?

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+	"strings"
+	"syscall"
+	"testing"
+
+	"github.com/cilium/ebpf/internal/platform"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+func init() {
```
ti-mo, line 30: This could use some documentation about how the subprocess mechanism works at a high level and what the various stages mean.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+		if stage != "1" {
+			return
+		}
+
+		// If this is stage 1 of the subtest, we setup the environment and then re-exec ourselves.
+		setupSubtest()
+	}
+}
+
+// RunWithToken sets up an environment with a BPFFS with the provided delegated permissions.
```
ti-mo: Here and everywhere else: document what it does for the user, not what it does behind the scenes. The implementation is subject to change and does not typically need to be known by the user.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+//
+// When called, the calling test method is re-executed in a subprocess with the environment set up. RunWithToken
+// returns true in the parent process, and false in the child process, so that the test can distinguish between them and
```
ti-mo: I can't infer from the function signature how it should be used. A small code snippet in the docstring on how to handle the return value would be nice.

rlimit/rlimit_linux.go
```diff
@@ -59,6 +62,21 @@ func detectMemcgAccounting() error {
 		MaxEntries: 1,
 	}
 
+	var tok *token.Token
+	if platform.IsLinux {
+		var err error
+		tok, err = token.Create()
+		if err != nil && !errors.Is(err, token.ErrTokenNotAvailable) {
```
ti-mo: Since we're checking for `tok != nil` in all cases, what's the benefit of having `token.ErrTokenNotAvailable`? As long as `token.Create()` is allowed to return a nil token, it should only return an error if it's actionable/fatal.

internal/token/token.go
ti-mo: > In the current solution, we do 2 syscalls to get a token every time we load a program / map, so that could add up over time.

You bet, cilium_lxc in its base configuration has 31 tail call progs and an entrypoint, ignoring probes executed on agent start and by ebpf-go itself.

> We could do some sort of straight up cache with a global variable, but it means we hold onto a token file descriptor forever. We usually are very strict about that sort of thing in the library.

Given this would be a singleton fd held by a sync.OnceValues and only enabled when a bpffs with caps is present in the mount namespace, I'd say this more than compensates for 60-odd extra fds being allocated and closed in a short timespan for each endpoint creation.

Over on the libbpf side, a user would also typically request a single token and reuse it across all bpf() calls for the duration of the process. Sure they'd have control over its lifecycle, but I don't see one fd as cost-prohibitive for this kind of feature.

internal/token/token.go
```diff
@@ -0,0 +1,249 @@
...
+			errs = errors.Join(errs, fmt.Errorf("open bpffs mount %q: %w", mount.Path, err))
+			continue
+		}
+		defer unix.Close(bpffsfd)
+
+		token, err := sys.TokenCreate(&sys.TokenCreateAttr{
+			BpffsFd: uint32(bpffsfd),
+		})
+		if err != nil {
+			if errors.Is(err, unix.EINVAL) {
```
ti-mo: Execute these before the `if err != nil` clause so we end up with less indentation. `err != nil` is typically the catch-all block.

internal/token/token.go
```diff
@@ -0,0 +1,249 @@
...
+	Path        string
+	Cmds        []sys.Cmd
+	Maps        []sys.MapType
+	Progs       []sys.ProgType
+	AttachTypes []sys.AttachType
+
+	mountID int
+}
+
+var readBPFFSMounts = sync.OnceValues(func() ([]BPFFSMount, error) {
```
ti-mo: This needs refactoring (into functions) and we should aim to for close to 100% test coverage.

Also maybe benchmarks, though I'll add those when I convert this and the kconfig reader to `kallsyms.reader`; that one needs to be exported first and made configurable. Not so important for now since it's a cached function.

internal/token/token.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/cilium/ebpf/internal/platform"
+	"github.com/cilium/ebpf/internal/sys"
+	"github.com/cilium/ebpf/internal/unix"
+)
+
+var (
+	ErrTokenNotAvailable = errors.New("BPF Token is not available")
+)
+
+type Token struct {
```
ti-mo: This is missing a docstring, and fields should be unexported. We typically expose only `func (p *Program) FD() int` etc. I know it's an internal dependency, but it keeps concerns separated and discourages open-coding against on Token.FD in business logic.

internal/token/token.go
```diff
@@ -0,0 +1,249 @@
...
+	*sys.FD
+	Mount BPFFSMount
+}
+
+// Create attempts to create a BPF token, it does so by finding all mounted BPFFS filesystems, and trying to create a
+// token on each of them until it finds one that works.
+//
+// If a token cannot be created, it returns [ErrTokenNotAvailable]. If the OS is not Linux, it returns
+// [internal.ErrNotSupportedOnOS].
+func Create() (*Token, error) {
```
ti-mo: As mentioned above, this should probably go behind a `sync.OnceValues`.

internal/testutils/token_linux.go
```diff
@@ -0,0 +1,324 @@
...
+// only run the actual test logic in the child process.
+func RunWithToken(tb testing.TB, delegated Delegated) bool {
```
ti-mo: I forgot what your early attempts at this function looked like. Could we make this take a subtest closure instead? I find the `... }) return }` really awkward and most of the current tests are written as subtests anyway. Down the line, we can add a `BenchWithToken()` if we really need to, but for now testing.T is fine.

internal/sys/types.go
ti-mo, line 1: This ended up pooping out _a lot_ of code. Could we write something similar to stringer, but in reverse? All identifiers could go into a single string and the lookup table could be a map indexed by subslices of the single string.

ti-mo (commented):
Thanks! This is all I got for now, will do a 2nd round when these are discussed/addressed.

internal/token/token.go
dylandreimerink: Yea, libbpf does a token per collection. Given the current assumptions with the ambient token approach I think its reasonably alright to have a cache for the life of the process. We can adjust this later on if we feel the need for invalidation.

internal/sys/types.go
dylandreimerink: Using the token info means we do not have to parse strings from the mounts. Much cleaner now.

dylandreimerink:
I figured out that you can request object info on tokens as well, which returns the delegated permissions as numbers, which means we don't have to have all of the from string and parsing logic. I still need to address a few of the outstanding comments with regards to changing the API of the tokens testutils. Just pushing what I have so far.

internal/testutils/token_linux.go
dylandreimerink: I added an additional commit to convert them to a subtest style, its easier to use now I feel like.

rlimit/rlimit_linux.go
dylandreimerink: I suppose that is true. In my head it just makes more sense to say `tok == nil + err != nil` or `tok != nil + err == nil`. Having a sentinel for this sort of thing makes sense to me, but its not a hill I will die on.

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+	// {id} {parent id} {major:minor} {root} {mount point} {mount options} [optional fields...] - {filesystem type} {source} {superblock options}
+	for scanner.Scan() {
+		line := scanner.Text()
+		// Ignore the trailing newline
+		if line == "" {
+			continue
+		}
+
+		// Split on the dash, since we don't know the amount of optional fields, which can throw of indexes.
+		firstHalfStr, secondHalfStr, ok := strings.Cut(line, "-")
```
ti-mo: Directory names frequently contain dashes. This needs to split on `" - "` instead of just `"-"`. Please include this case in `ParseTestMounts` as well.

Less of a problem, but there are also superblock options with dashes (like `remount-ro`):
```
48 30 259:1 / /boot rw,relatime shared:84 - vfat /dev/nvme0n1p1 rw,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,utf8,errors=remount-ro
```

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+		idStr := firstHalf[0]
+		id, err := strconv.Atoi(idStr)
+		if err != nil {
+			return nil, fmt.Errorf("atoi mount id: %w", err)
+		}
+
+		mountPoint := firstHalf[4]
+
+		// Remove any BPFFS mounts on the same mount point with a lower mount ID, as they would be hidden by the newer mount.
+		mounts = slices.DeleteFunc(mounts, func(mnt bpffsMount) bool {
```
ti-mo: Couldn't something like this be run once, outside of the loop? This only works if the mountinfo entries appear in strictly ascending order, which I'm not sure we should rely on because it's not documented in `man 5 proc_pid_mountinfo`.

To keep things simple, we could use a `map[string]bpffsMount` and only insert if the incoming ID is larger than the existing one. (don't use a pointer as the value type, so `mountID` will be 0 on lookup miss)

---

Edit: Leaving my original comment since I don't think it's bad advice. We don't need to care about the mount ID anyway now we no longer need to parse capabilities out of the mountinfo string. Just collect mounts into a `map[string]struct{}` and return a []string, get rid of `bpffsMount`.

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+
+var (
+	ErrTokenNotAvailable = errors.New("BPF Token is not available")
+)
+
+// Token is BPF token, which can be used to create privileged BPF object that normally require CAP_SYS_ADMIN or
+// CAP_NET_ADMIN. With a token, the same BPF objects can be created with CAP_BPF.
+type Token struct {
+	fd    *sys.FD
+	_info *sys.TokenInfo
```
ti-mo: Nit: make this `info` and export `info()`? The underscore prefix looks strange. Info() returning a `sys` type is fine for now.

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Int returns the raw file descriptor of the token.
+func (t *Token) Int() int {
```
ti-mo: We typically use FD() for fd.Int() proxies:

```suggestion
func (t *Token) FD() int {
```

like
```go
func (p *Program) FD() int {
	return p.fd.Int()
}
```

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+})
+
+// create attempts to create a BPF token, it does so by finding all mounted BPFFS filesystems, and trying to create a
+// token on each of them until it finds one that works.
+//
+// If a token cannot be created, it returns [ErrTokenNotAvailable]. If the OS is not Linux, it returns
+// [internal.ErrNotSupportedOnOS].
+func create() (*Token, error) {
+	if !platform.IsLinux {
+		return nil, internal.ErrNotSupportedOnOS
```
ti-mo:
```suggestion
		return nil,  fmt.Errorf("%w: %w", ErrTokenNotAvailable, internal.ErrNotSupportedOnOS)
```

Would this be a terrible idea? Making ErrTokenNotAvailable a superset of ENSOnOS simplifies the `token.Get()` call sites quite a bit:

```go
var tok *token.Token
if platform.IsLinux {
	var err error
	tok, err = token.Get()
	if err != nil && !errors.Is(err, token.ErrTokenNotAvailable) {
		return nil, fmt.Errorf("get BPF token: %w", err)
	}

	if tok != nil {
		attr.BtfTokenFd = int32(tok.Int())
		attr.BtfFlags |= sys.BPF_F_TOKEN_FD
	}
}
```

becomes:

```go
tok, err := token.Get()
if err != nil && !errors.Is(err, token.ErrTokenNotAvailable) {
	return nil, fmt.Errorf("get BPF token: %w", err)
}
if tok != nil {
	attr.BtfTokenFd = int32(tok.Int())
	attr.BtfFlags |= sys.BPF_F_TOKEN_FD
}
```

Ask for forgiveness, not permission. 🙂

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+	if err := sys.ObjInfo(t.fd, &tokenInfo); err != nil {
+		return sys.TokenInfo{}, err
+	}
+
+	t._info = &tokenInfo
+	return tokenInfo, nil
+}
+
+// AllowedCmd checks if the token allows usage of the given BPF syscall command.
+func (t *Token) AllowedCmd(cmd sys.Cmd) error {
```
ti-mo: Ultra OCD nit: I'd invert the names since they'd flow nicer as questions: `CmdAllowed(foo)`(?). I'd expect a func called `AllowedCmd` to return a list of some sort.

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+	mounts, err := readBPFFSMounts()
+	if err != nil {
+		return nil, fmt.Errorf("get bpffs mounts: %w", err)
+	}
+
+	var errs error
+	for _, mount := range mounts {
+		bpffsfd, err := unix.Open(mount.path, unix.O_DIRECTORY|unix.O_RDONLY, 0)
+		if err != nil {
+			errs = errors.Join(errs, fmt.Errorf("open bpffs mount %q: %w", mount.path, err))
```
ti-mo: I'm not convinced we should continue if we fail to open one of the paths. If there are multiple, and let's say the first one fails, the second one may succeed and yield a valid token, and the user will never see the error generated by the first.

If there's a specific reason we tolerate failures here, let's document it.

We probably should document this fact on the subsequent `errors.Is` cases as well. If there are multiple mounts, e.g. one plain /sys/fs/bpf mounted from the system and one specifically for providing permissions, we should ignore errors from the first and continue with the second.

rlimit/rlimit_linux.go
ti-mo: Let's go with the sentinel, but let's not gate on `platform.IsLinux` everywhere. See my comment above.

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+
+		// Split on the dash, since we don't know the amount of optional fields, which can throw of indexes.
+		firstHalfStr, secondHalfStr, ok := strings.Cut(line, "-")
+		if !ok {
+			return nil, fmt.Errorf("invalid mountinfo line, missing dash: %q", line)
+		}
+
+		secondHalf := strings.Fields(strings.TrimSpace(secondHalfStr))
+		if len(secondHalf) < 1 {
```
ti-mo: Let's make this `== 0` for clarity.

internal/token/token.go
```diff
@@ -0,0 +1,272 @@
...
+			return nil, fmt.Errorf("invalid mountinfo line, too few fields after dash: %q", line)
+		}
+
+		fstype := secondHalf[0]
+		if fstype != "bpf" {
+			continue
+		}
+
+		firstHalf := strings.Fields(strings.TrimSpace(firstHalfStr))
+		if len(firstHalf) < 5 {
```
ti-mo: According to `man 5 proc_pid_mountinfo`, the first half has at least 6 mandatory fields, not 5. The 7th field is optional.

map.go
```diff
@@ -670,6 +684,13 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 	}
 
 	if errors.Is(err, unix.EPERM) {
+		if terr := tok.AllowedCmd(sys.BPF_MAP_CREATE); terr != nil {
```
ti-mo: If possible, I'd prefer if we can do this wrapping in a single spot in `sys.BPF()` right before `return r1, err`. That way, all callers get to benefit from potentially enriched errors. Not sure if you've evaluated this approach before or if there were any issues with it.

syscalls.go
```diff
@@ -71,12 +72,42 @@ func progLoad(insns asm.Instructions, typ ProgramType, license string) (*sys.FD,
...
+		if tok != nil {
+			attr.ProgTokenFd = int32(tok.Int())
+			attr.ProgFlags |= sys.BPF_F_TOKEN_FD
+		}
+	}
+
+	return sys.ProgLoad(attr)
+}
+
+func mapLoad(attr *sys.MapCreateAttr) (*sys.FD, error) {
```
ti-mo:
```suggestion
func mapCreate(attr *sys.MapCreateAttr) (*sys.FD, error) {
```

internal/token/token.go
dylandreimerink: I think you are right, in the case of a `/sys/fs/bpf` which does not have tokens, I expect the error to occur as result of the `sys.TokenCreate`. An error while trying to open the dir would mean there is some unexpected situation, a bug in the mount parsing or something truly unexpected.

map.go
dylandreimerink: We cannot do it at the `sys.BPF` level because each syscall command has the token in a different field in its attributes.

We could add it to `sys.ProgLoad`, `sys.MapCreate`, and `sys.BtfLoad` respectively. Take the fd from the passed attributes, construct a `Token` from it, then query it. I thought it would be odd if we did token based error handling in the `sys` package, and the rest outside of it.

Also, we would get an import cycle between `internal/token` and `internal/sys`, so it would mean moving the the `Token` implementation into `internal/sys` instead of having a separate package, if we are fine with that.

Edit: `types.go` where  `sys.ProgLoad`, `sys.MapCreate`, and `sys.BtfLoad`  are defined is currently generated. So we would have to special case the generation logic or pull these into manually maintained files instead. Overall a lot of churn, not sure if its worth it at this stage.

ti-mo:
@dylandreimerink Played with this a bit after getting back from break.

tl;dr: `ErrTokenNotAvailable` was removed since it didn't have much purpose if returned `*Token`s can be nil. I've completely pushed token handling into `func BPF()` and added `ErrTokenCapabilities` to communicate to the caller that tokens are in use and EPERM was encountered. This sentinel is always returned with surrounding context about missing caps. (caveat: wherever possible) This removed a lot of the boilerplate for token plumbing in callers of BPF() and centralized all handling so we don't accidentally miss any call sites.

In response to https://github.com/cilium/ebpf/pull/1953#discussion_r3022175184:

I ended up moving `Token` to package `sys` for this reason. I've also discovered that obj_info for tokens was only added in 6.17, included in the 6.18 LTS release, so the 6.12 LTS won't have the accurate errors we envisioned. I've changed it so that if `tok != nil && EPERM && tok.info == nil`, we always return `ErrTokenCapabilities`, even if the token may actually have the right caps, but the kernel throws EPERM somewhere else. I think that's a worthwhile trade-off over re-evaluating the meaning of EPERM across the library; the latter wouldn't be feasible IMO.

I've left the testing harness untouched, but I think we should make some improvements to the subtest output streaming because, at least locally, `gotestsum` breaks if any of the subtests return errors. We may have to do some stdio parsing to properly pipe the subtest's output through to the parent and to set SKIP/FAIL properly if detected within the subtest. Not sure if we already do something along those lines.

Let's run through the changes offline tomorrow.

PR #1956, opened 2026-02-24
docs: add maps documentation

gajjuu:
Adds documentation for declaring and using eBPF maps, including examples for
lookup, update, iteration, and pinning.

The structure and style follow the existing documentation pages.

### DIFF
--- a/docs/ebpf/concepts/maps.md
+++ b/docs/ebpf/concepts/maps.md
@@ -1,0 +1,124 @@
+# Maps
+
+eBPF maps are kernel data structures used to exchange data between eBPF programs
+and user space, or between multiple eBPF programs. The library provides helpers
+to load, access, and manipulate maps defined in ELF objects.
+
+This section describes how to declare maps in C, load them using `bpf2go`, and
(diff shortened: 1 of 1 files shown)


gajjuu:
It looks like the Netlify deploy preview failed due to an environment issue
during the MkDocs build (multiprocessing semaphore error). Please let me know
if there’s anything I should adjust on my side.

ti-mo:
Ignore the netlify errors, looks like they made some breaking changes that broke the build (again). We're going to migrate off of Netlify, this happens every 2 months on average.

Thanks for the patch, but what does this really add to the existing documentation? This looks like a copy of the getting started guide (https://ebpf-go.dev/guides/getting-started), and the Common Pitfalls section doesn't add anything substantial either. Was this generated by an LLM?

gajjuu:
Thanks for the feedback.
You're right that this overlaps too much with the Getting Started guide and doesn't add enough new value. I'll rework it to avoid duplication and focus on more reference-style content specific to maps and their usage in this library.

I'll update the PR accordingly.

Issue #1957, opened 2026-02-28
support StructOps declarative initialization in bpf2go


shun159:
This proposal introduces StructOps Initialization Structs in bpf2go. This allows users to define the initial state of a struct_ops map (currently, I consider scalar values such as flags) using standard Go struct literals, which are then synchronized with the MapSpec during the LoadAndAssign process.

Initialization Struct Generation: bpf2go generates Go structs that mirror the BTF definition of struct_ops members. These structs are used as templates for map creation.

Pre-load Synchronization: It patches the MapSpec.Contents with the provided values before the map is created in the kernel.

example:

generated code
```go
// bpfStructOpsMinimalSched is a struct type for the struct_ops map.
type bpfStructOpsMinimalSched struct {
	_         structs.HostLayout
	Init      *ebpf.Program `ebpf:"init"`
	Flags     uint64        `ebpf:"flags"`
	TimeoutMs uint32        `ebpf:"timeout_ms"`
	Name      [128]int8     `ebpf:"name"`
}
```

user code
```go
func main() {
	if err := rlimit.RemoveMemlock(); err != nil {
		log.Fatal(err)
	}

	objs := bpfObjects{}
	objs.bpfStructOps.StructOpsMinimalSched.Flags = ScxOpsSwitchPartial
	objs.bpfStructOps.StructOpsMinimalSched.TimeoutMs = 1000
	if err := loadBpfObjects(&objs, nil); err != nil {
		log.Fatalf("loading objects: %v", err)
	}
	defer objs.Close()

	m := objs.MinimalSched
	l, err := link.AttachStructOps(link.StructOpsOptions{Map: m})
	if err != nil {
		log.Fatalf("failed to attach sched_ext: %s", err)
	}
	defer l.Close()

	stopper := make(chan os.Signal, 1)
	signal.Notify(stopper, os.Interrupt, syscall.SIGTERM)

	<-stopper

	log.Print("quit sched_ext")
}
```

PR #1973, opened 2026-03-29
[POC] bpf2go/struct_ops: implement StructOps support

shun159:
Implement code generation and library support for declarative initialization of struct_ops maps. This allows users to initialize struct_ops member, using standard Go structs during the LoadAndAssign process.

bpf2go changes:

- Scans for ebpf.StructOpsMap in the CollectionSpec and generates corresponding Go struct types.
- Automatically maps kernel function pointers to *ebpf.Program and scalar members to appropriate Go types.

Library changes:

- Modify assignValues and LoadAndAssign to pass the reflect.Value of target fields, enabling the library to read pre-instantiated data from the user-provided objects struct.
- Introduce loadStructOps to synchronize Go struct fields with the MapSpec before the map is created in the kernel.
- Implement a "partial override" strategy in structOpsPatchValue: only non-zero fields in the Go struct are patched into the ELF section data, preserving ELF-defined defaults for uninitialized fields.
- Support early relocation of struct_ops function pointers by injecting ebpf.Program FDs into the section data buffer during the patch process.

Ref: https://github.com/cilium/ebpf/issues/1957


collection.go
```diff
@@ -183,7 +183,19 @@ func (cs *CollectionSpec) LoadAndAssign(to interface{}, opts *CollectionOptions)
...
+		if ms != nil && ms.Type == StructOpsMap {
+			vType := typ
+			if vType.Kind() == reflect.Ptr {
+				vType = vType.Elem()
+			}
+
+			if vType.Kind() == reflect.Struct && typ != reflect.TypeOf((*Map)(nil)) {
+				return loader.loadStructOps(name, val)
+			}
+		}
```
dylandreimerink, line 197: This code should go into `loadMap`, as its a special case of loading a map.

struct_ops.go
```diff
@@ -107,6 +108,7 @@ func structOpsCopyMember(m, km btf.Member, data []byte, kernVData []byte) error
 	dstOff := int(km.Offset.Bytes())
 
 	if srcOff < 0 || srcOff+mSize > len(data) {
+		fmt.Println(srcOff, srcOff+mSize, len(data))
```
dylandreimerink, line 111: I think you forgot to remove this after debugging.

struct_ops.go
```diff
@@ -145,16 +147,54 @@ func structOpsIsMemZeroed(data []byte) bool {
...
 }
 
+// structOpsCopyMemberFromStructValue updates a single member in the ELF section data
+// using a value from a generated Go struct.
+//
+// It implements a "partial override" strategy: if the Go field is a zero value,
+// the operation is skipped to preserve the original default value defined in the ELF.
+// Currently, nested structs or unions are not supported for non-zero updates to
+// avoid complex recursive patching.
+func structOpsCopyMemberFromStructValue(fieldVal reflect.Value, m btf.Member, dstOff int, secData []byte) error {
```
dylandreimerink, line 157: That is a very verbose function name. Perhaps just `structOpsCopyMember`? That is about as long as I would go.

struct_ops.go
```diff
@@ -145,16 +147,54 @@ func structOpsIsMemZeroed(data []byte) bool {
...
+	if srcSize != kSize {
+		return fmt.Errorf("size mismatch (Go:%d, Kernel:%d)", srcSize, kSize)
+	}
+
+	if dstOff+kSize > len(secData) {
+		return fmt.Errorf("kernel buffer overflow: dstOff: %v kSize: %v, kernVdata: %v", dstOff, kSize, len(secData))
+	}
+
+	srcPtr := unsafe.Pointer(fieldVal.UnsafeAddr())
+	srcData := unsafe.Slice((*byte)(srcPtr), kSize)
```
dylandreimerink, line 173: We shouldn't be unsafe casting like this. Use the `sysenc.Marshal` utility instead. It internally has optimizations to do an unsafe cast as well, but only when it determines it is safe to do so, and will fallback to slower but safe logic otherwise.

struct_ops.go
```diff
@@ -173,3 +213,52 @@ func structOpsSetAttachTo(
...
+		}
+
+		if err := structOpsCopyMemberFromStructValue(fieldVal, *m, dstOff, secData); err != nil {
+			return fmt.Errorf("member %s: %w", tag, err)
+		}
+	}
+	return nil
+}
+
+func structOpsFindMemberByName(st *btf.Struct, name string) (*btf.Member, error) {
```
dylandreimerink, line 257: Also very verbose name for what it does. 

I would just inline this. This function is not reused outside of its parent. You are creating an error just to then `continue` in its parent function.

It can be something like:
```
var m btf.Member
if idx := slices.IndexFunc(st.Members, func(m btf.Member) bool {
    return m.Name == name
} {
    m = st.Members[idx]
} else {
    continue
}
```

struct_ops.go
```diff
@@ -173,3 +213,52 @@ func structOpsSetAttachTo(
...
+		dstOff := int(m.Offset.Bytes())
+		fieldVal := v.Field(i)
+
+		if prog, ok := fieldVal.Interface().(*Program); ok {
+			if prog != nil {
+				if err := structOpsPopulateValue(*m, secData, prog); err != nil {
+					return fmt.Errorf("member %s: %w", tag, err)
+				}
+			}
+			continue
```
dylandreimerink, line 247:
```suggestion
			if prog == nil {
			    continue
			}
	
			if err := structOpsPopulateValue(*m, secData, prog); err != nil {
				return fmt.Errorf("member %s: %w", tag, err)
			}
```

collection.go
```diff
@@ -544,6 +556,38 @@ func (cl *collectionLoader) loadVariable(varName string) (*Variable, error) {
...
+		return nil, fmt.Errorf("map %s: not found in loader", name)
+	}
+
+	if field.IsValid() && field.Kind() == reflect.Struct {
+		userType, ok := btf.As[*btf.Struct](ms.Value)
+		if !ok {
+			return nil, fmt.Errorf("map %s: value type is not a Struct", name)
+		}
+
+		rawData, ok := ms.Contents[0].Value.([]byte)
```
dylandreimerink, line 578: Use `ms.dataSection` instead, it has guards in place for unexpected map values. It is also what is used in `populateStructOps`.

collection.go
```diff
@@ -544,6 +556,38 @@ func (cl *collectionLoader) loadVariable(varName string) (*Variable, error) {
...
+// into the MapSpec's raw data buffer. This allows setting initial flags,
+// parameters, and function pointers before the kernel performs
+// struct_ops validation.
+func (cl *collectionLoader) loadStructOps(name string, field reflect.Value) (interface{}, error) {
+	ms := cl.coll.Maps[name]
+	if ms == nil {
+		return nil, fmt.Errorf("map %s: not found in loader", name)
+	}
+
+	if field.IsValid() && field.Kind() == reflect.Struct {
```
dylandreimerink, line 572: If you invert this there is far less code indented.

```suggestion
if !field.IsValid() || field.Kind() != reflect.Struct {
    return field.Interface(), nil
}

// [...]
```

cmd/bpf2go/gen/output.tpl
```diff
@@ -94,6 +94,14 @@ type {{ .Name.Objects }} struct {
...
 	{{ .Name.Maps }}
 	{{ .Name.Variables }}
+	{{ .Name.StructOps }}
+}
+
+// {{ .Name.StructOps }} contains all struct_ops types.
+type {{ .Name.StructOps }} struct {
+{{- range $name, $id := .StructOps }}
+	{{ $id }} {{ $.Name }}{{ $id }} `ebpf:"{{ $name }}"`
+{{- end }}
```
dylandreimerink, line 104: I would consider surrounding this with an `{{ if .StructOps }}`. Programs, maps, and variables are expected to be in every generated file. But is odd to have empty struct ops related types, even when no struct ops are in use, which will be the majority of cases.

cmd/bpf2go/gen/output.go
```diff
@@ -104,6 +108,62 @@ type GenerateArgs struct {
...
+		if fieldName == "" {
+			continue
+		}
+
+		var fieldType string
+		if ptr, ok := btf.As[*btf.Pointer](btf.UnderlyingType(m.Type)); ok {
+			if _, ok := btf.As[*btf.FuncProto](btf.UnderlyingType(ptr.Target)); ok {
+				fieldType = "*ebpf.Program"
+			}
+		}
```
dylandreimerink, line 146: This logic of emitting a `*ebpf.Program` for pointers to functions seems to be the only reason for having this logic here. I think it would be better to adjust the `GoFormatter` in btf/format.go to do this instead, so you can reuse the rest of the existing logic for writing go types.

cmd/bpf2go/gen/output.go
```diff
@@ -124,7 +124,7 @@ func generateStructOpsShadowType(goTypeName string, st *btf.Struct, gf *btf.GoFo
 
 	sb.WriteString(fmt.Sprintf("// %s is a struct type for the struct_ops map.\n", goTypeName))
 	sb.WriteString(fmt.Sprintf("type %s struct {\n", goTypeName))
-	sb.WriteString(fmt.Sprintf("\t_ structs.HostLayout\n"))
```
dylandreimerink, line 127: This commit can be squashed into the previous one.

cmd/bpf2go/test/test_bpfeb.go
```diff
@@ -118,6 +118,11 @@ type testObjects struct {
 	testPrograms
 	testMaps
 	testVariables
+	testStructOps
+}
+
+// testStructOps contains all struct_ops types.
+type testStructOps struct {
```
dylandreimerink, line 125: All of this churn in these files can be avoided if we only emit these when there is at least one struct_ops type.

cmd/bpf2go/gen/output.go
```diff
@@ -139,19 +139,33 @@ func generateStructOpsShadowType(goTypeName string, st *btf.Struct, gf *btf.GoFo
...
 		}
 
+		if fieldType == "" {
+			if _, ok := btf.As[*btf.Struct](underlying); ok {
+				size, _ := btf.Sizeof(m.Type)
+				fieldType = fmt.Sprintf("[%d]byte", size)
+			} else if _, ok := btf.As[*btf.Union](underlying); ok {
+				size, _ := btf.Sizeof(m.Type)
+				fieldType = fmt.Sprintf("[%d]byte", size)
+			}
```
dylandreimerink, line 157: This also deviates from the `GoFormatter` behavior, which would make nested structs here.

So question, did we do this because recursive emitting is difficult or is there a technical reason? If its just difficult, usage of the `GoFormatter` should solve it. Otherwise, I would suggest modifying the BTF before its passed to the formatter to get a similar end result.

dylandreimerink (changes requested):
Thanks again for working on this. I flagged a few things for you to look at.

sadisticbrew:
Hi @shun159,
you still planning to continue with this PR? I noticed it's been quiet since Dylan's review in March. If you'd rather move on, I'd be happy to pick it up and address @dylandreimerink 's feedback as my first contibution.
Also, since main has gained quite a bit of struct_ops support since then, I'd rebase and rework the runtime parts to build on what's already there.

PR #1978, opened 2026-04-03
map: avoid misleading error message for storage maps

venk8:
### Description
This PR improves error handling in handleMapCreateError when creating BPF maps with zero MaxEntries.

It fixes a misleading error message when creating maps that require MaxEntries to be zero (such as SkStorage, InodeStorage, TaskStorage, CgroupStorage, CGroupStorage, and PerCPUCGroupStorage). 

Previously, if map creation failed with EINVAL (e.g., due to missing BTF metadata) while MaxEntries was correctly set to 0, the error handler would wrongly suggest that MaxEntries may be incorrectly set to zero.

### Changes
* Added requiresZeroMaxEntries() to MapType in types.go to identify map types that require zero max entries (supporting CgroupStorage, CGroupStorage, PerCPUCGroupStorage, InodeStorage, TaskStorage, and SkStorage).
*Updated handleMapCreateError in map.go to exempt these map types from the MaxEntries == 0 check.

### DIFF
--- a/types.go
+++ b/types.go
@@ -188,6 +188,16 @@ func (mt MapType) mustHaveNoPrealloc() bool {
 	return false
 }
 
+// mustHaveZeroMaxEntries returns true if the map type requires MaxEntries to be zero.
+func (mt MapType) mustHaveZeroMaxEntries() bool {
+	switch mt {
+	case CgroupStorage, CGroupStorage, PerCPUCGroupStorage, InodeStorage, TaskStorage, SkStorage:
+		return true
+	}
+
+	return false
+}
+
 // ProgramType of the eBPF program
 type ProgramType uint32
 
--- a/map.go
+++ b/map.go
@@ -673,7 +673,7 @@ func handleMapCreateError(attr sys.MapCreateAttr, spec *MapSpec, err error) erro
 		return fmt.Errorf("map create: %w (MEMLOCK may be too low, consider rlimit.RemoveMemlock)", err)
 	}
 	if errors.Is(err, unix.EINVAL) {
-		if spec.MaxEntries == 0 {
+		if spec.MaxEntries == 0 && !spec.Type.mustHaveZeroMaxEntries() {
 			return fmt.Errorf("map create: %w (MaxEntries may be incorrectly set to zero)", err)
 		}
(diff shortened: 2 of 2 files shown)


dylandreimerink:
These changes seem sensible to me.  It might also make sense to check the opposite, throwing an more explicit error when entries is 0 for maps that require them to be non-zero.  I would squash the second commit into the first one.

venk8:
@dylandreimerink  Thanks for review! I have addressed the comments. Please take a look!

map_test.go
```diff
@@ -2323,3 +2323,67 @@ func ExampleMap_Iterate_nestedMapsAndProgramArrays() {
 		panic(fmt.Sprint("Iterator encountered an error:", err))
 	}
 }
+
+func TestMapCreateStorageMisleadingError(t *testing.T) {
```
ti-mo: Please drop both of these tests, they mostly exercise the kernel itself, and we don't typically assert error strings since they add maintenance pain.

This PR adds nicer errors for the user; if machines should be able to act on the info, we should return a sentinel instead. (not useful for now IMO)

types.go
```diff
@@ -188,6 +188,16 @@ func (mt MapType) mustHaveNoPrealloc() bool {
 	return false
 }
 
+// requiresZeroMaxEntries returns true if the map type requires MaxEntries to be zero.
+func (mt MapType) requiresZeroMaxEntries() bool {
```
ti-mo: Consistency nit:

```suggestion
func (mt MapType) mustHaveZeroMaxEntries() bool {
```

ti-mo:
> These changes seem sensible to me. It might also make sense to check the opposite, throwing an more explicit error when entries is 0 for maps that require them to be non-zero. I would squash the second commit into the first one.

@dylandreimerink CPUMap and PerfEventArray off the top of my head, but those need nCPU detection at runtime. There's no sane value that can be given at compile time, so they're encouraged to be left to 0. I think except for the list introduced in this PR, all maps need a non-zero max_entries.

ti-mo (changes requested):
Thanks! Left some nits.

venk8:
@ti-mo  @dylandreimerink  Addressed the comments! PTAL

PR #1979, opened 2026-04-06
ringbuf: add UnsafeReader as zero-copy ringbuf reader

jschwinger233:
ringbuf.Reader.ReadInto always copies sample bytes into user-owned memory. That is safe and convenient, but it becomes a visible bottleneck as sample size grows.

This PR introduces that path as a separate type (UnsafeReader) so the tradeoff is explicit: performance vs safety/concurrency guarantees.

# APIs

### Read
```
// Zero-copy Read — slice valid until Commit.
rec, err := unsafeReader.Read()
// Batch commit after processing N records.
unsafeReader.Commit()
```

### ReadInto
```
// Zero-copy ReadInto — slice valid until Commit.
err := unsafeReader.ReadInto(&rec)
// Batch commit after processing N records.
unsafeReader.Commit()
```

### ReadFunc
```
// Callback mode — sample is valid only during callback.
// Record is committed implicitly after callback returns.
rem, err := unsafeReader.ReadFunc(func(sample []byte, rem int) error {})
```

 ## Benchmark Results

Benchmark is done by https://github.com/jschwinger233/bpf_ringbuf_zc_benchmark

```
  for i in {1..20}; do
    echo $((i * 128))
    taskset -c 2 ./bpf_ringbuf_zc_benchmark -event-size $((i * 128))
  done
```

  Observed throughput (UnsafeReader.ReadInto vs Reader.ReadInto):

  | Event size | Speedup |
  |---:|---:|
  | 128B  | 1.18x |
  | 1024B | 3.61x |
  | 2048B | 6.02x |
  | 2432B | 6.39x |

  Trend: small records show modest gains; larger records show strong scaling, peaking at ~6.4x.

# Related
#1968
#1930 
#1915

### DIFF
--- a/ringbuf/doc.go
+++ b/ringbuf/doc.go
@@ -3,4 +3,7 @@
 // BPF allows submitting custom events to a BPF ring buffer map set up
 // by userspace. This is very useful to push things like packet samples
 // from BPF to a daemon running in user space.
+//
+// UnsafeReader provides zero-copy access to samples and is not safe for
+// concurrent use.
 package ringbuf
--- a/ringbuf/read_internal.go
+++ b/ringbuf/read_internal.go
@@ -1,0 +1,48 @@
+package ringbuf
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"time"
+)
+
+func readWithPoll(poller poller, ring eventRing, deadline time.Time, haveData *bool, pendingErr *error, read func() error) error {
+	if ring == nil {
+		return fmt.Errorf("ringbuffer: %w", ErrClosed)
+	}
+
+	for {
+		if !*haveData {
+			if pe := *pendingErr; pe != nil {
+				*pendingErr = nil
+				return pe
+			}
+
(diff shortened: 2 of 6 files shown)


jschwinger233:
@ti-mo I drafted this PR to address the comments and move the discussion forward: https://github.com/cilium/ebpf/pull/1968#issuecomment-4168358156

orishuss:
Thanks for pushing this @jschwinger233!
I wrote #1981 which has a lot of similarities, and also has "commit tokens" to allow concurrency when reading from the unsafe reader. It has a small cpu & memory overhead. You're welcome to look and decide if you like the idea.

ti-mo:
I've spent considerable time on a proposal taking into account both the zero-copy and commit batching use cases: https://github.com/cilium/ebpf/pull/2037. @jschwinger233 and @orishuss, please give the new proposal a spin in a real-world use case, curious to hear your thoughts.

yossilevi-web:
hi, any news about it?

PR #1981, opened 2026-04-12
ringbuf: add UnsafeReader for zero-copy reads with deferred commit

orishuss:
## Summary

Supersedes #1968. Implements the separate `UnsafeReader` type as recommended by @ti-mo and @florianl in the review of #1968, rather than adding zero-copy methods directly to `Reader`.

## Changes

- **`UnsafeReader`** — new type for zero-copy ring buffer reads. Returns `UnsafeRecord` with `RawSample` pointing directly into the mmap'd ring buffer region. Separate type from `Reader` so the two APIs can't be mixed accidentally.
- **`UnsafeRecord`** — wraps `Record` via embedding. Compile-time type safety: you can't pass `*UnsafeRecord` to `Reader.ReadInto` or `*Record` to `UnsafeReader.Read`.
- **`CommitToken`** — opaque token (unexported `consPos` field) returned by `Read`. User can't inspect or forge it.
- **`Commit(token)`** — per-token commit with ordered advancement. Consumer position only advances when all preceding tokens are also committed (e.g. commit B,C then A → advances past C).
- **`CommitAll()`** — batch commit, advances consumer to the latest read position.
- **`ReadFunc(func([]byte))`** — callback-based read with implicit auto-commit after the callback returns.

## Architecture

Addresses @ti-mo and @florianl's feedback from #1968:

1. **Separate type** — `UnsafeReader` is distinct from `Reader`. No mixing of safe and unsafe APIs on the same type.
2. **Shared internals, no duplication** — both embed `readerBase` which provides Close, SetDeadline, BufferSize, Flush, AvailableBytes, and the poll loop (`readWaitLocked`).
3. **Clean `eventRing` interface** — only exposes primitives both readers use: `readRecordUnsafe`, `advanceTo`, `pendingPosition`. No `readRecord` on the ring — that logic lives in `safeRing`, a small wrapper used only by `Reader`.
4. **`safeRing`** — wraps `eventRing` to provide `readRecord` (readRecordUnsafe + copy + advanceTo). Keeps `Reader`'s API and behavior identical to today.
5. **Callback support** — `ReadFunc` receives `[]byte` (not `Record`-driven), auto-commits after the callback.
6. **Concurrency documented** — Read/Commit/CommitAll are goroutine-safe (mutex). `RawSample` validity after commit is the caller's responsibility.

## File layout

| File | Contents |
|---|---|
| `ring.go` | `eventRing` interface, `ringbufHeader`, `ringReader` (readRecordUnsafe, advanceTo, pendingPosition) |
| `reader_base.go` | errors, `poller`, `Record`, `readerBase` + shared methods, `initReaderBase` |
| `reader.go` | `safeRing`, `Reader`, NewReader, Read, ReadInto |
| `unsafe_reader.go` | `UnsafeRecord`, `CommitToken`, `UnsafeReader`, NewUnsafeReader, Read, ReadFunc, Commit, CommitAll |
| `unsafe_reader_test.go` | 10 tests + 2 benchmarks |

No changes to `ring_other.go`, `ring_windows.go`, `reader_test.go`, or platform files.

### DIFF
--- a/ringbuf/reader.go
+++ b/ringbuf/reader.go
@@ -1,145 +1,52 @@
 package ringbuf
 
 import (
-	"errors"
 	"fmt"
-	"os"
-	"sync"
-	"time"
-	"unsafe"
 
 	"github.com/cilium/ebpf"
-	"github.com/cilium/ebpf/internal/platform"
-	"github.com/cilium/ebpf/internal/sys"
 )
 
-var (
-	ErrClosed = os.ErrClosed
-	errEOR    = errors.New("end of ring")
-	errBusy   = errors.New("sample not committed yet")
-)
-
-// poller abstracts platform-specific event notification.
-type poller interface {
-	Wait(deadline time.Time) error
-	Flush() error
-	Close() error
+// Wraps an eventRing to provide a copy-and-advance read for Reader.
+type safeRing struct {
+	eventRing
 }
 
-// eventRing abstracts platform-specific ring buffer memory access.
-type eventRing interface {
-	size() int
-	AvailableBytes() uint64
-	readRecord(rec *Record) error
-	Close() error
-}
+func (sr *safeRing) readRecord(rec *Record) error {
+	buf := rec.RawSample
 
-// ringbufHeader from 'struct bpf_ringbuf_hdr' in kernel/bpf/ringbuf.c
-type ringbufHeader struct {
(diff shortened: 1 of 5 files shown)


ti-mo:
Closing as duplicate of https://github.com/cilium/ebpf/pull/1979.

PR #1987, opened 2026-04-20
bpf2go: add compdb export

stepbrobd:
i was trying to use bpf2go in my project where the dev is happening in a nixos machine

the library paths are at non-conventional locations and my editor/lsp wasn't able to find them and not really providing useful suggestions

i temporarily worked around this by adding a similar compdb wrapper around bpf2go (https://github.com/stepbrobd/rfm/blob/3c86b1ec512c83f46f7d7ad7ee88fa1f64efce44/cmd/bpfgen/main.go#L31) to generate compile_commands.json in the project root and i think this could be useful to someone else

### DIFF
--- a/cmd/bpf2go/README.md
+++ b/cmd/bpf2go/README.md
@@ -38,6 +38,13 @@ up-to-date list.
 disable this behaviour using `-no-global-types`. You can add to the set of
 types by specifying `-type foo` for each type you'd like to generate.
 
+## Editor integration
+
+Pass `-compdb path/to/compile_commands.json` (or set `BPF2GO_COMPDB`) to upsert
+[JSON Compilation Database](https://clang.llvm.org/docs/JSONCompilationDatabase.html)
+entry for each source. The file is written before compilation, so the entry is
+kept up to date even when the build step itself is unable to run.
+
 ## Examples
 
 See [examples/kprobe](../../examples/kprobe/main.go) for a fully worked out example.
--- a/cmd/bpf2go/compdb.go
+++ b/cmd/bpf2go/compdb.go
@@ -1,0 +1,82 @@
+//go:build !windows
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"golang.org/x/sys/unix"
+)
+
+// A single JSON Compilation Database entry.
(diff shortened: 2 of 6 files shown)


ti-mo:
Hi, could you elaborate on the issue with `the library paths`? Which library are we talking about?

stepbrobd:
for example a specific version of vmlinux headers, or libbpf and xdp headers that look like this

`/nix/store/0y6byys2fd1xzq66vld217g7y69kvqa1-libbpf-1.6.3/include`

if i edit the files directly with nvim, clangd can't find the symbols and just generating massive list of errors, and this compdb change basically will create compile_commands.json entries at user specified location on `go generate` call

for example (i copied from my source and i'm still working on the upstream patch, the final outcome will be more like `//go:generate go tool bpf2go -tags linux counter counter.c -compdb ../compile_commands.json`):

```go
package probe

//go:generate go tool bpfgen --ident rfm --output-dir . --pkg-config libbpf --compdb ../compile_commands.json ../bpf/rfm_tc.c
```

```console
$ cat compile_commands.json
[
  {
    "directory": "/home/ysun/Workspace/rfm/bpf",
    "file": "/home/ysun/Workspace/rfm/bpf/rfm_tc.c",
    "arguments": [
      "clang",
      "--target=bpf",
      "-O2",
      "-g",
      "-Wall",
      "-Wno-missing-declarations",
      "-I/home/ysun/Workspace/rfm/bpf/include",
      "-I/home/ysun/Workspace/rfm/bpf/include/vmlinux/x86",
      "-D__TARGET_ARCH_x86",
      "-I/nix/store/0y6byys2fd1xzq66vld217g7y69kvqa1-libbpf-1.6.3/include",
      "-c",
      "/home/ysun/Workspace/rfm/bpf/rfm_tc.c"
    ]
  }
]
```

ti-mo:
Closing for now due to inactivity, ping here if you want to pick this up again. To provide reproducible BPF C builds (including in CI) and detect drift to checked-in ELFs, you kind of want to run all builds through an LLVM toolchain container like we do in the lib. That could be another possible solution and isn't tied to bpf2go.

stepbrobd:
thanks for the nudge, had a few deadlines in the past few weeks

i’ll try to pick this up in the coming months

to clarify a little more, the point of this PR is to provide better experience for developers so that the imported headers can be referenced in the source file (after running go generate) thru clangd compdb support (e.g. doc references, go to definition, etc)

PR #1988, opened 2026-04-21
btf: replace global BTF cache with user provided cache

matthyx:
## Problem

`globalCache` in `btf/kernel.go` holds strong `*Spec` pointers to the kernel and module BTF specs. Once loaded, these ~20 MiB of type information are never released, even after all eBPF programs have been loaded and no `CollectionSpec` retains a reference to them.

`FlushKernelSpec()` exists as a manual escape hatch but requires callers to know when it is safe to call — which is inherently racy in a library.

## Solution

Change `globalCache.kernel` and `globalCache.modules` from strong `*Spec` pointers to `weak.Pointer[Spec]` (Go 1.24+). The GC will collect the cached specs automatically once no `CollectionSpec` (or other caller) holds a strong reference to them.

Key behaviours preserved:
- **Amortisation**: The cache still avoids re-parsing `/sys/kernel/btf/vmlinux` across multiple `LoadKernelSpec` calls that overlap in time.
- **Lazy reload**: If the GC collects the cached spec, the next `LoadKernelSpec` call transparently re-parses it.
- **`FlushKernelSpec()`**: Still works; zeroing a `weak.Pointer` is equivalent to the previous `= nil`. Now deprecated since explicit flushing is no longer needed.
- **`NewCache()`**: Handles the case where the weak pointer has already been collected, falling back to an empty `Cache` (which then loads on demand via `Kernel()`).

## `Cache` is now self-contained

Previously, `Cache.Module` re-used the global module cache and then rebased the result against the Cache's own kernel decoder. That created a hidden invariant: the Cache's kernel decoder had to share its raw mmap'd buffer with whatever kernel spec the global cache currently held — otherwise `rebaseDecoder` would fail with `"raw BTF differs"`.

Under strong references this invariant held permanently. Under weak references it could break: if the global cached kernel spec were GC'd between `Cache.Kernel()` and `Cache.Module()`, the next module load would parse a fresh kernel mmap and the rebase would blow up.

`Cache.Module` now loads module BTF directly against the Cache's own kernel via `loadKernelModuleSpec(name, c.kernelTypes)`. The Cache no longer depends on the global cache's state for correctness, and no `kernelSource` anchor field is needed. Memory reclaim is unchanged: the mmap'd vmlinux buffer is freed (`munmap` cleanup on `sharedBuf`) once the last `Cache` (and any other strong holder) is dropped.

The trade-off is that a module loaded via `Cache.Module` is no longer re-cached in the global module cache, so a subsequent `LoadKernelModuleSpec` for the same module reparses sysfs. Module BTF is small (kilobytes) and this only matters if `Cache` and the package-level loaders are mixed for the same modules.

## Memory impact

In a long-running process that loads a handful of eBPF programs at startup and then runs indefinitely, this frees ~20 MiB of Go heap automatically — without any caller changes.

Measured in a Kubernetes node-agent workload (inspektor-gadget) where kernel BTF was the dominant post-load heap consumer.

## Go version requirement

`weak.Pointer` was stabilised in Go 1.24. The module already requires Go 1.23; this PR bumps the minimum to Go 1.24.

## Testing

Existing `TestLoadKernelSpec` and `TestCache` pass. The cache correctness is unchanged — the weak pointer is semantically transparent to callers as long as they hold a strong reference (via a returned `*Spec`) across the critical section.

### DIFF
--- a/btf/kernel.go
+++ b/btf/kernel.go
@@ -16,24 +16,6 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
 
-// globalCache amortises decoding BTF across all users of the library.
-var globalCache = struct {
-	sync.RWMutex
-	kernel  *Spec
-	modules map[string]*Spec
-}{
-	modules: make(map[string]*Spec),
-}
-
-// FlushKernelSpec removes any cached kernel type information.
-func FlushKernelSpec() {
-	globalCache.Lock()
-	defer globalCache.Unlock()
-
-	globalCache.kernel = nil
-	globalCache.modules = make(map[string]*Spec)
-}
-
 // LoadKernelSpec returns the current kernel's BTF information.
 //
 // Defaults to /sys/kernel/btf/vmlinux and falls back to scanning the file
@@ -42,88 +24,6 @@ func FlushKernelSpec() {
 //
 // Consider using [Cache] instead.
 func LoadKernelSpec() (*Spec, error) {
-	spec, err := loadCachedKernelSpec()
-	return spec.Copy(), err
-}
-
-// load (and cache) the kernel spec.
-//
-// Does not copy Spec.
-func loadCachedKernelSpec() (*Spec, error) {
-	globalCache.RLock()
-	spec := globalCache.kernel
-	globalCache.RUnlock()
-
-	if spec != nil {
-		return spec, nil
-	}
-
-	globalCache.Lock()
-	defer globalCache.Unlock()
-
-	// check again, to prevent race between multiple callers
-	if globalCache.kernel != nil {
-		return globalCache.kernel, nil
-	}
-
-	spec, err := loadKernelSpec()
-	if err != nil {
-		return nil, err
-	}
-
-	globalCache.kernel = spec
-	return spec, nil
-}
-
-// LoadKernelModuleSpec returns the BTF information for the named kernel module.
-//
-// Using [Cache.Module] is faster when loading BTF for more than one module.
-//
-// Defaults to /sys/kernel/btf/<module>.
-// Returns an error wrapping ErrNotSupported if BTF is not enabled.
-// Returns an error wrapping fs.ErrNotExist if BTF for the specific module doesn't exist.
-func LoadKernelModuleSpec(module string) (*Spec, error) {
-	spec, err := loadCachedKernelModuleSpec(module)
-	return spec.Copy(), err
-}
-
-// load (and cache) a module spec.
-//
-// Does not copy Spec.
-func loadCachedKernelModuleSpec(module string) (*Spec, error) {
-	globalCache.RLock()
-	spec := globalCache.modules[module]
-	globalCache.RUnlock()
-
-	if spec != nil {
-		return spec, nil
-	}
-
-	base, err := loadCachedKernelSpec()
-	if err != nil {
-		return nil, err
-	}
-
-	// NB: This only allows a single module to be parsed at a time. Not sure
-	// it makes a difference.
-	globalCache.Lock()
-	defer globalCache.Unlock()
-
-	// check again, to prevent race between multiple callers
-	if spec := globalCache.modules[module]; spec != nil {
-		return spec, nil
-	}
-
-	spec, err = loadKernelModuleSpec(module, base)
-	if err != nil {
-		return nil, err
-	}
-
-	globalCache.modules[module] = spec
-	return spec, nil
-}
-
-func loadKernelSpec() (*Spec, error) {
 	if platform.IsWindows {
 		return nil, internal.ErrNotSupportedOnOS
 	}
@@ -167,6 +67,21 @@ func loadKernelSpec() (*Spec, error) {
 
 	spec, err := LoadSpecFromReader(file)
 	return spec, err
+}
+
+// LoadKernelModuleSpec returns the BTF information for the named kernel module.
+//
+// Using [Cache.Module] is faster when loading BTF for more than one module.
+//
+// Defaults to /sys/kernel/btf/<module>.
+// Returns an error wrapping ErrNotSupported if BTF is not enabled.
+// Returns an error wrapping fs.ErrNotExist if BTF for the specific module doesn't exist.
+func LoadKernelModuleSpec(module string) (*Spec, error) {
+	base, err := LoadKernelSpec()
+	if err != nil {
+		return nil, err
+	}
+	return loadKernelModuleSpec(module, base)
 }
 
 func loadKernelModuleSpec(module string, base *Spec) (*Spec, error) {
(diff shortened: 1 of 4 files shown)


btf/kernel.go
```diff
@@ -17,21 +18,25 @@ import (
...
-	kernel  *Spec
-	modules map[string]*Spec
+	kernel  weak.Pointer[Spec]
+	modules map[string]weak.Pointer[Spec]
 }{
-	modules: make(map[string]*Spec),
+	modules: make(map[string]weak.Pointer[Spec]),
 }
 
 // FlushKernelSpec removes any cached kernel type information.
```
ti-mo, line 32: I'd move to formally deprecate `FlushKernelSpec()` (again) since it's a bit of a footgun.

```suggestion
// FlushKernelSpec removes any cached kernel type information.
//
// Deprecated: the global BTF cache now uses weak pointers, making explicit flushing redundant.
```

btf/kernel.go
```diff
@@ -17,21 +18,25 @@ import (
 )
 
 // globalCache amortises decoding BTF across all users of the library.
+//
+// Weak pointers allow the cached specs to be GC'd when no CollectionSpec
```
ti-mo: I find the second half of the sentence to be slightly inaccurate, I'd change it to:

```suggestion
// Weak pointers allow the cached Specs to be GC'd when no callers hold
// any references to them.
```

Reasons being 1. I initially parsed 'kernel BTF memory' as 'kernel memory', which it isn't, 2. the 20MiB number will change over time, and 3. references may exist for reasons other than CollectionSpec loading. We don't need to be too specific here.

btf/kernel.go
```diff
@@ -61,16 +66,19 @@ func loadCachedKernelSpec() (*Spec, error) {
...
+		return spec, nil
 	}
 
 	spec, err := loadKernelSpec()
 	if err != nil {
 		return nil, err
 	}
 
-	globalCache.kernel = spec
+	// Module specs were rebased against the old (now GC'd) kernel spec.
```
ti-mo, line 79: Please squash this into a single commit, there's no need to separate these changes.

ti-mo (changes requested):
Thanks! Left some comments. Looks like there's still a failure in the selftests.

matthyx:
@ti-mo looks like the CI passes now

btf/kernel.go
```diff
@@ -226,6 +235,7 @@ func findVMLinux() (*os.File, error) {
 // It is not safe for concurrent use.
 type Cache struct {
 	kernelTypes   *Spec
+	kernelSource  *Spec // keeps the original cached spec alive so module specs remain valid
```
ti-mo: I think adding this field obscures a latent problem that only emerged by making the global cache use weak pointers. A `Cache` should never be able to be invalidated by external factors. Could you investigate?

matthyx: very good point, see what we propose now

matthyx:
@ti-mo can you check my proposal on the module BTF?

ti-mo:
@matthyx I've given this a lot of thought, but something tells me this isn't the right approach. Turning these into weak pointers completely defeats their purpose as a long-lived cache, since there's no guarantee whatsoever that the global cache will outlive two sequential calls to LoadCollectionSpec, at which point we may as well remove it since btf.Cache fills intra-load caching now. I understand the difficulty in finding a good point to call Flush(); I assume you do this after every collection load to keep consumption down.

Though I have one question: what does that 20MiB of heap contain? We have lazy decoding now, so we should only (mostly) inflate types that get returned to the caller. I don't know your use case, but 20MiB feels like a lot. Are you counting heap or RSS?

It's worth noting that a few things changed since introducing the global cache. Originally, it intended to amortize the cost of parsing BTF across multiple programs within a single LoadCollectionSpec, but the fact that it persisted across multiple collection loads was a nice bonus. Now that decoding is lazy + mmap-backed, a lot of that cost is gone. Conversely, I find the lower-level logic around how the caches interact almost impossible to reason about today.

With btf.Cache in the picture, it may be time to sunset the global cache altogether and allow the caller to pass one in through `CollectionOptions`. This way, the caller can decide for themselves when and how to manage a cache and we sidestep this issue altogether.

matthyx:
Thanks, this makes sense. You're right that weak pointers turn the global cache from a deterministic long-lived cache into a best-effort memoization layer, which is probably the wrong trade-off here.

Given that we already use a btf.Cache during collection load, surfacing that through CollectionOptions seems like the cleaner direction. That lets callers choose the cache lifetime explicitly instead of the library trying to guess a safe flush 
point.

The memory saving is heap, but more 17 than 20, you can see some info here: https://github.com/kubescape/node-agent/pull/786#issue-4302065223

Should I close this?

ti-mo:
Hi, are you generating these replies using an LLM? If you're up for making the CollectionOptions change, feel free to reuse this PR for that.

matthyx:
@ti-mo yes I use one to redact sentences from bullet points, as I'm not native speaker in English. Sorry if that offended you.

matthyx:
@ti-mo something like that?

dylandreimerink (dismissed):
I think this updated PR is nice, having less memory overhead by default, and allowing power users to enable caching between collections.

btf/kernel.go
```diff
@@ -8,31 +8,18 @@ import (
...
-	globalCache.Lock()
-	defer globalCache.Unlock()
-
-	globalCache.kernel = nil
-	globalCache.modules = make(map[string]*Spec)
-}
+//
+// Deprecated: kernel BTF is no longer cached at the package level.
+// Discard your [Cache] instances to release cached BTF instead.
+func FlushKernelSpec() {}
```
ti-mo: We should simply remove this as a call to action to anyone that depended on this behaviour. For anyone relying on package-level caching, Flush() becoming a no-op is already sort-of a breaking change. We may as well make it explicit.

btf/kernel.go
```diff
@@ -83,44 +40,11 @@ func loadCachedKernelSpec() (*Spec, error) {
...
 	if err != nil {
 		return nil, err
 	}
-
-	globalCache.modules[module] = spec
-	return spec, nil
+	return loadKernelModuleSpec(module, base)
 }
 
 func loadKernelSpec() (*Spec, error) {
```
ti-mo: Nit: move this above LoadKernelModuleSpec and export instead?

collection.go
```diff
@@ -36,6 +36,14 @@ type CollectionOptions struct {
 	// The given Maps are Clone()d before being used in the Collection, so the
 	// caller can Close() them freely when they are no longer needed.
 	MapReplacements map[string]*Map
+
+	// Cache amortises the cost of decoding kernel BTF across multiple
+	// collection loads. Callers loading more than one collection should
```
ti-mo: Nit:
```suggestion
	// Collection loads. Callers loading more than one Collection should
```

btf/kernel.go
```diff
@@ -231,33 +155,12 @@ type Cache struct {
 	loadedModules []string
 }
 
-// NewCache creates a new Cache.
+// NewCache returns an empty Cache.
 //
-// Opportunistically reuses a global cache if possible.
+// Pass it to [CollectionOptions] to share kernel BTF across multiple
+// collection loads.
```
ti-mo: Nit:
```suggestion
// Collection loads.
```

collection.go
```diff
@@ -36,6 +36,14 @@ type CollectionOptions struct {
...
+	// allocate a Cache via [btf.NewCache] and share it across loads.
+	// If nil, a fresh cache is allocated per load and discarded.
+	//
+	// The same Cache must not be used concurrently.
```
ti-mo: This looks straightforward enough to address, it would be a rather inconvenient limitation. Could you pick this up as well?

ti-mo (commented):
Thanks! This cleans up nicely. Left a few nits, and we should prob make btf.Cache concurrency-safe.

matthyx:
@ti-mo updated

btf/kernel.go
```diff
@@ -16,24 +16,6 @@ import (
 	"github.com/cilium/ebpf/internal/unix"
 )
```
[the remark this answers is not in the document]
dylandreimerink: We know, we can ignore this

btf/kernel.go
```diff
@@ -169,6 +69,21 @@ func loadKernelSpec() (*Spec, error) {
...
+//
+// Defaults to /sys/kernel/btf/<module>.
+// Returns an error wrapping ErrNotSupported if BTF is not enabled.
+// Returns an error wrapping fs.ErrNotExist if BTF for the specific module doesn't exist.
+func LoadKernelModuleSpec(module string) (*Spec, error) {
+	base, err := LoadKernelSpec()
+	if err != nil {
+		return nil, err
+	}
+	return loadKernelModuleSpec(module, base)
```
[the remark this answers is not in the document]
dylandreimerink: Yes, we diverged from the original description, also ignoring.

dylandreimerink (approved):
Timo is currently unavailable so I will take over. Existing feedback is fixed, I did not find anything additional. AI also does not have useful comments, approved.

Thank you, sorry for the wait.

matthyx:
thanks @dylandreimerink

PR #1990, opened 2026-04-22
linker: check ErrNotSupported too

andrewstrohman:
When we fixup kfuncs, we only check hard-coded paths for the BTF file. If we cannot find the BTF file in these paths, then we should poison the kfunc calls in order to allow loading the collection.

The fixup kfunc logic does not respect the KernelTypes *btf.Spec passed with ProgramOptions because kfunc support depends on CONFIG_DEBUG_INFO_BTF, which exposes /sys/kernel/btf/vmlinux. As such, we expect that file to exist and be available.

But if it's not available, findTargetInKernel will return ErrNotSupported, and we shouldn't refuse to load the collection.

### DIFF
--- a/linker.go
+++ b/linker.go
@@ -305,10 +305,12 @@ fixups:
 			return nil, fmt.Errorf("kfuncMetaKey doesn't contain kfuncMeta")
 		}
 
-		// findTargetInKernel returns btf.ErrNotFound if the input btf.Spec is nil.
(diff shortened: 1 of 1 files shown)


ti-mo:
@andrewstrohman Thanks for the PR! If you don't mind, we'll go with https://github.com/cilium/ebpf/pull/1991 instead, since there are more callers to `findTargetInKernel` that rely on it only returning ErrNotFound. We should probably restore the API contract instead of papering over it in the callers until we make bigger changes.

PR #1993, opened 2026-04-29
bpf2go: improve duplicate type name error message

### LINKED ISSUE #1638: bpf2go: bogus "type name is used multiple times" error
### Describe the bug

When (redundant) `-type T` option refers to a type that is already exposed because a variable of `T` exists, a bogus "type name is used multiple times" error is reported.

### How to reproduce

```c
//go:build ignore
// x.c
struct S {} s;
``` 
```
$ bpf2go --type S --go-package eee x x.c
Error: can't write /home/nickz/cilium-ebpf.git/x_bpfel.go: type name "xS" is used multiple times
```
The error occurs because variable's type is a copy [produced](https://github.com/cilium/ebpf/blob/v0.17.1/elf_reader.go#L1256) in `loadDataSections`. We end up with two distinct objects.

### Version information

v0.17.1

wucm667:
When --type T is used for a type that is already exported via a
variable or map k/v, bpf2go reports a confusing error about the
type name being used multiple times.

Update the error message to guide users toward the fix:
remove the redundant --type flag.

Fixes #1638

### DIFF
--- a/cmd/bpf2go/gen/output.go
+++ b/cmd/bpf2go/gen/output.go
@@ -221,7 +221,7 @@ func sortTypes(typeNames map[btf.Type]string) ([]btf.Type, error) {
 		}
 
 		if names[i] == name {
-			return nil, fmt.Errorf("type name %q is used multiple times", name)
+			return nil, fmt.Errorf("type name %q is used multiple times (remove its --type flag?)", name)
 		}
 
 		types = append(types[:i], append([]btf.Type{typ}, types[i:]...)...)


ti-mo:
@wucm667 Thanks for the patch, but I don't really agree that the behaviour should be changed. Maybe the error message is a little obtuse at the moment, that could likely be improved.

If there's already a variable or map k/v of type T, specifying it additionally using `--type T` will effectively make that flag dead code. It's also not ideal to couple the contents of `go:generate` statements to type names contained in C files. Including more than a handful of types using CLI flags sounds really cumbersome, especially when the list of types changes frequently.

For example, in Cilium we use this for exporting enums to Go using a bpf2go-like tool:

```c
#define EXPORT_TYPE(type) __expand_type(type, __COUNTER__)
#define __expand_type(type, n) ___expand_type(type, n)
#define ___expand_type(type, n) type __dpexp_ ## n

...
enum identity {
       UNKNOWN_ID = 0,
       HOST_ID = 1,
       WORLD_ID = 2,
};
EXPORT_TYPE(enum identity);
```

I'd probably just update the error string in `sortTypes` to read something like `"type name %q is used multiple times (remove its --type flag?)"`.

wucm667:
@ti-mo Good point — makes sense. Updated to just improve the error message as you suggested. The `--type` flag is indeed dead code in this case, so guiding the user to remove it is the right approach.

PR #1995, opened 2026-04-30
fix(link): add missing BPF_F_REPLACE flag for RawAttachProgram

### LINKED ISSUE #1994: RawAttachProgram with ReplaceProgram anchor silently fails for BPF_PROG_ATTACH-based cgroup attachment
### Describe the bug

Using RawAttachProgram with Anchor: link.ReplaceProgram(oldProg) on a BPF_CGROUP_DEVICE cgroup doesn't actually replace the program. The new program gets added alongside the old one. No error is returned — it looks like it worked, but BPF_PROG_QUERY shows two programs attached instead of one.

In commit 285069d the Replace field was removed in favor of the Anchor API. The new code handles BPF_F_REPLACE differently from other anchor flags - https://github.com/cilium/ebpf/blob/main/link/program.go#L53-L59
So BPF_F_REPLACE is never added to AttachFlags. 
But as per https://github.com/torvalds/linux/commit/7dd68b3279f1792103d12e69933db3128c6d416e
kernel needs BPF_F_REPLACE in attach_flags to actually perform the replacement — without it, replace_bpf_fd is ignored and the call is treated as a regular append

do I follow it correctly that it's missing the `attr.AttachFlags |= sys.BPF_F_REPLACE`?

### How to reproduce

tested on kernel 5.14 (RHEL 9)
- attach a BPF_CGROUP_DEVICE program with BPF_F_ALLOW_MULTI — 1 program attached
- call RawAttachProgram with Anchor: link.ReplaceProgram(oldProg) — returns no error
- query: 2 programs attached. Old one still there.

### Version information

github.com/cilium/ebpf v0.17.3

wucm667:
Fixes #1994

## Summary

When using `RawAttachProgram` with `Anchor: link.ReplaceProgram(oldProg)`, the `BPF_F_REPLACE` flag was not being set in `AttachFlags`. This caused the kernel to silently ignore the replacement and treat the call as a regular append, resulting in two programs being attached instead of one being replaced.

## Root Cause

In `link/program.go`, when `flags == sys.BPF_F_REPLACE`:
- `attr.ReplaceBpfFd` was set correctly
- But `attr.AttachFlags` was **not** updated with `sys.BPF_F_REPLACE`

Per the Linux kernel ([commit 7dd68b3](https://github.com/torvalds/linux/commit/7dd68b3279f1792103d12e69933db3128c6d416e)), the kernel requires `BPF_F_REPLACE` in `attach_flags` to perform the replacement — without it, `replace_bpf_fd` is ignored.

## Fix

Add `attr.AttachFlags |= sys.BPF_F_REPLACE` in the `BPF_F_REPLACE` branch of `RawAttachProgram`, ensuring the kernel actually performs the replacement operation.

### DIFF
--- a/link/program.go
+++ b/link/program.go
@@ -50,12 +50,11 @@ func RawAttachProgram(opts RawAttachProgramOptions) error {
 			return fmt.Errorf("attach program: %w", err)
 		}
 
-		if flags == sys.BPF_F_REPLACE {
-			// Ensure that replacing a program works on old kernels.
+		attr.AttachFlags |= flags
+		if flags&sys.BPF_F_REPLACE != 0 {
 			attr.ReplaceBpfFd = fdOrID
 		} else {
 			attr.RelativeFdOrId = fdOrID
-			attr.AttachFlags |= flags
 		}
 	}
 
--- a/link/program_test.go
+++ b/link/program_test.go
@@ -5,6 +5,7 @@ package link
 import (
 	"fmt"
 	"net"
+	"os"
 	"testing"
 
 	"github.com/cilium/ebpf"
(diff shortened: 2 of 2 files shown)


link/program.go
```diff
@@ -50,9 +50,10 @@ func RawAttachProgram(opts RawAttachProgramOptions) error {
 			return fmt.Errorf("attach program: %w", err)
 		}
 
-		if flags == sys.BPF_F_REPLACE {
+		if flags&sys.BPF_F_REPLACE != 0 {
 			// Ensure that replacing a program works on old kernels.
```
michalskrivanek: I'm not sure the "on old kernels" is accurete. IIUC the behavior was like that since the BPF_F_REPLACE had been introduced.

michalskrivanek (commented):
Thank you for the quick fix!
I can confirm the patch fixes the problem.

wucm667:
@michalskrivanek Good catch — you're right, that comment was misleading. Updated it to clarify that the kernel requires BPF_F_REPLACE in attach_flags for the replace operation to take effect, not just on old kernels.

wucm667:
@michalskrivanek @mmat11 Heads up on the CI — the Build and Lint failures (3 govet inline warnings in  and ) are pre-existing and unrelated to my change in .

They appear to be triggered by a new inline check in the latest golangci-lint version. Happy to help fix those if needed, or we can handle them separately.

link/program.go
```diff
@@ -50,9 +50,11 @@ func RawAttachProgram(opts RawAttachProgramOptions) error {
 			return fmt.Errorf("attach program: %w", err)
 		}
 
-		if flags == sys.BPF_F_REPLACE {
-			// Ensure that replacing a program works on old kernels.
+		if flags&sys.BPF_F_REPLACE != 0 {
+			// The kernel requires BPF_F_REPLACE in attach_flags for the replace operation to take effect.
+			// Without it, the kernel ignores replace_bpf_fd and treats the call as an append.
 			attr.ReplaceBpfFd = fdOrID
+			attr.AttachFlags |= sys.BPF_F_REPLACE
```
mmat11: is it possible to have additional flags other than `BPF_F_REPLACE` in this branch? if so, I think flags should be assigned as they are:
```
attr.AttachFlags |= flags
if flags&sys.BPF_F_REPLACE != 0 {
    // The kernel requires BPF_F_REPLACE in attach_flags for the replace operation to take effect.
    // Without it, the kernel ignores replace_bpf_fd and treats the call as an append.
    attr.ReplaceBpfFd = fdOrID
} else {
    attr.RelativeFdOrId = fdOrID
}
```

mmat11:
would it be possible to add a test that exercises this code path?

wucm667:
@mmat11 Done both:

1. **Refactored `AttachFlags` assignment** — set `attr.AttachFlags |= flags` unconditionally before the if/else, which now only routes `fdOrID` to `ReplaceBpfFd` or `RelativeFdOrId`.
2. **Added `TestRawAttachProgramReplace`** — attaches progA, replaces it with progB using `ReplaceProgram(progA)`, then queries attached programs to verify only progB remains (not both).

All changes compile cleanly and `go vet` passes. Runtime tests require CAP_BPF so they'll run in CI.

link/program_test.go
```diff
@@ -135,3 +135,59 @@ func TestRawAttachProgramAnchor(t *testing.T) {
...
+	// Attach progA.
+	err := RawAttachProgram(RawAttachProgramOptions{
+		Target:  int(cgroup.Fd()),
+		Program: progA,
+		Attach:  ebpf.AttachCGroupInetEgress,
+		Flags:   uint32(flagAllowMulti),
+	})
+	testutils.SkipIfNotSupported(t, err)
+	qt.Assert(t, qt.IsNil(err))
+	defer RawDetachProgram(RawDetachProgramOptions{
```
ti-mo: Detach failures will be silently ignored this way, put these into `t.Cleanup()`s with the usual asserts.

ti-mo (changes requested):
Thanks! This needs a rebase to pass CI, and detach errors shouldn't be ignored. Please squash this into a single commit.

Also, for awareness, our GenAI policy is about to be finalized and will go into effect soon: https://github.com/cilium/community/pull/408. I'm calling this out because your replies feel very generated, and this won't be allowed under the new policy. Please don't be afraid of making mistakes! Most of us aren't native speakers either. We'd rather talk to a real human, we have agents too, just like everyone else. :)

wucm667:
done, rebased and squashed. moved the detach into t.Cleanup with assert

ti-mo (approved):

PR #1998, opened 2026-05-05
Go 1.25, run `go fix` in CI, fix nil derefs in reflect usage

ti-mo:
go vet started complaining on main about reflect.Ptr not being inlined. Turns out it's been replaced by reflect.Pointer. Run `go fix` on the codebase to automatically fix this, along with a few other deprecations we hadn't noticed before.

This turned into a yak shave because `go fix` received new CLI flags we needed (disabling `reflecttypefor`) in 1.25 and I noticed we hadn't upgraded yet. Upgrading the Go version in the tooling container image exposed a breaking change in the LLVM apt repositories, since clang-format now implicitly depends on a specific package (`clang-format-22` in this case) whose repo wasn't in apt.list yet. Then, due to the `go fix` churn, Copilot started looking at some of the reflect usage and pointed out nil derefs, which we really should fix.


mmat11 (dismissed):
lgtm!

PR #1999, opened 2026-05-07
Makefile: use go tool for linting

florianl:
Replace the GH actions by moving the tooling to use go tool. This also allows to run these tools locally in the same version/setup as in CI more easily.


Makefile
```diff
@@ -101,6 +101,12 @@ clean:
 format:
 	find . -type f -name "*.c" | xargs clang-format -i
 
+lint-golangci:
+	go tool golangci-lint run
+
+lint-staticcheck:
+	go tool staticcheck -tests  ./...
```
ti-mo: `-tests` is true by default, no need to include it.

Makefile
```diff
@@ -101,6 +101,12 @@ clean:
 format:
 	find . -type f -name "*.c" | xargs clang-format -i
 
+lint-golangci:
+	go tool golangci-lint run
```
ti-mo: Apparently not technically required, but I always run `golangci-lint run ./...` out of habit of other tooling.

go.mod
```diff
@@ -6,43 +6,243 @@ require (
 	github.com/go-quicktest/qt v1.101.1-0.20240301121107-c6c8733fa1e6
 	github.com/google/go-cmp v0.7.0
 	github.com/jsimonetti/rtnetlink/v2 v2.0.1
-	golang.org/x/sync v0.17.0
-	golang.org/x/sys v0.37.0
+	golang.org/x/sync v0.20.0
```
ti-mo, line 9: Please bump in a separate commit.

.github/workflows/ci.yml
```diff
@@ -42,13 +42,12 @@ jobs:
           fi
 
       - name: Run staticcheck
-        uses: dominikh/staticcheck-action@v1
-        with:
-          version: "master"
-          install-go: false
+        run: |
+          make lint-staticcheck
```
ti-mo: I don't think the make targets add much and I find them harder to remember than invoking the tools directly.

ti-mo (changes requested):
Hi, Flo. Technically a no-brainer, but I think there are some downsides.

- The golangci-lint GH action pulls a binary version of the tool and doesn't need to compile, which does seem to add up (34s up from 5s). I see the `go fix` I added recently also added quite some time to the critical path, I'm curious about the impact of moving this to a parallel job. Maybe warming the cache will move to `Run staticcheck` then.
- I haven't actually looked into this, but how does tool version pinning work? Looks like external tools are added as an indirect dependency, but who's responsible for bumping them? At least dependabot automatically bumps community workflows, but with tools we need to do it manually.
- It adds a _lot_ of indirect dependencies, which I think all of our direct consumers will inherit.

I'd probably keep the golangci-lint action, but make staticcheck a `go tool`. I've heard the upstream go toolchain will absorbing some or all of the staticcheck functionality in the future. (can't remember source)

go.mod
florianl: I think this did conflict with golangci-lint and was updated because of that. Will revert it with the removal of golangci-lint from go tools.

florianl:
Setting the PR to draft for a moment to land https://github.com/cilium/ebpf/pull/2001 first.

https://github.com/cilium/ebpf/pull/2001 is required for the most recent version of staticcheck (v0.7.0).

The current version in the PR, v0.7.0-dev, is a bit older and does not contain the following commit https://github.com/dominikh/go-tools/commit/8273271481d0419ff572b946dd3dc3c8044e4cb9. For that reason, CI is currently failing with errors like these:
```
Error: asm/opcode_string.go:27:5: no value of type uint8 is less than 0 (SA4003)
Error: btf/btf_types_string.go:79:5: no value of type uint8 is less than 0 (SA4003)
Error: internal/sys/types_string.go:59:5: no value of type uint32 is less than 0 (SA4003)
Error: internal/sys/types_string.go:113:5: no value of type uint32 is less than 0 (SA4003)
Error: internal/sys/types_string.go:164:5: no value of type uint32 is less than 0 (SA4003)
Error: internal/sys/types_string.go:240:5: no value of type uint32 is less than 0 (SA4003)
Error: internal/tracefs/probetype_string.go:21:5: no value of type uint8 is less than 0 (SA4003)
Error: types_string.go:166:5: no value of type uint32 is less than 0 (SA4003)
```

As these SA4003 errors happen in auto-generated code and will be fixed with a more recent version, that includes https://github.com/dominikh/go-tools/commit/8273271481d0419ff572b946dd3dc3c8044e4cb9, I'm setting this PR to draft for the moment.

florianl:
> I haven't actually looked into this, but how does tool version pinning work? Looks like external tools are added as an indirect dependency, but who's responsible for bumping them? At least dependabot automatically bumps community workflows, but with tools we need to do it manually.

I don't think external tools, like dependabot or renovatebot, will update staticcheck or other tools available via go tool. As there aren't that much tagged versions of staticcheck per year, I think the manual overhead updating it should be managable.

ti-mo (approved):
Thanks!

PR #2004, opened 2026-05-08
tracefs: discover mount via /proc/self/mountinfo

### LINKED ISSUE #2000: link: configurable tracefs path for /host-style container layouts
## Context

Like #981, this is about making the tracefs mount point that `internal/tracefs.getTracefsPath` probes overridable. That issue was closed because the requestor's container runtime didn't expose `/sys` at all and they ended up bind-mounting `/sys/`. Our case is different and (I think) deserves another look.

## The use case

We run a security agent in Kubernetes. Each pod already bind-mounts the host filesystem at `/host` because the agent walks `/host/proc`, reads container runtime sockets, and inspects various host paths. To use `link.Tracepoint`/`link.Kprobe`, we *additionally* have to bind-mount:

- `/sys/kernel/tracing` → `/sys/kernel/tracing`
- `/sys/kernel/debug` → `/sys/kernel/debug`

…purely because `getTracefsPath` only probes those two canonical locations. Tracefs is already reachable inside the pod at `/host/sys/kernel/tracing` via the existing `/host` mount; we just can't tell cilium/ebpf to look there.

This isn't about a broken container runtime — it's about removing redundant Helm mounts from a pod that's already mounting the host root.

## What I'd like

A small API to point the tracefs probe at any real tracefs/debugfs mount, validated via `statfs`:

```go
// link package
func SetTracefsPath(path string) error
```

Auto-detection stays the default; the override only activates when explicitly set, and `SetTracefsPath("")` clears it. Internally backed by an atomic override checked before the existing `sync.OnceValues` auto-detect, so probe attaches before/after a `SetTracefsPath` call do the right thing.

## Why this and not "fix your runtime"

In `/host`-style deployments, the host filesystem *is* exposed — just not at the kernel-canonical paths. Forcing an extra two bind mounts (`/sys/kernel/tracing`, `/sys/kernel/debug`) for a path-resolution preference is fixable in ~20 lines without changing default behavior or affecting any existing user.

## Working implementation

I have a fork at https://github.com/yoav-orca/ebpf/tree/feat/tracefs-override with the patch and tests; happy to open a PR if the approach is acceptable. The change is isolated to `internal/tracefs/kprobe.go` and one wrapper in `link/link.go`. Tests cover override-honored, override-cleared, and rejection of non-tracefs paths.

I'm open to alternative shapes (env var, options struct, etc.) — wanted to surface the use case and proposal before sending a PR.

yoav-orca:
Lift the BPFFS mountinfo parser out of internal/sys into a new internal/mountinfo package, parameterized by filesystem type, and reuse it in internal/tracefs.

Tracefs auto-detection now consults /proc/self/mountinfo first: any tracefs mount is accepted regardless of its path, and a debugfs mount with an existing tracing/ subdirectory is used as a fallback. The hardcoded /sys/kernel/{tracing,debug/tracing} probe via statfs remains as a last-resort fallback for environments where mountinfo is unavailable, and to keep behavior identical on kernels that predate tracefs being lifted out of debugfs (Linux 4.1).

This removes the need for callers in containerized environments to bind-mount tracefs at the kernel-canonical paths just to make cilium/ebpf find it.

Closes #2000


internal/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// ParseEntries parses mountinfo data read from r. Exposed for tests; most
+// callers want [Read] or [FindByFSType].
+//
+// Uses [bufio.Reader] rather than [bufio.Scanner] because mountinfo lines
+// can exceed Scanner's default 64 KiB token limit — overlay filesystems
+// in containers commonly produce single mount entries with many KB of
+// lowerdir options.
+func ParseEntries(r io.Reader) ([]Entry, error) {
```
dylandreimerink: This function isn't used outside of the package and can thus be unexported.

internal/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,159 @@
...
+// On non-Linux, returns an error wrapping [internal.ErrNotSupportedOnOS].
+func Read() ([]Entry, error) {
+	return readOnce()
+}
+
+// FindByFSType returns the mount points for filesystems of the given type,
+// in order of appearance, with duplicates removed. A single mount point
+// may legitimately appear multiple times (for example, a bpffs mounted
+// twice with different delegation options) — callers typically only care
+// about the path, so duplicates are filtered.
```
dylandreimerink: This is more verbose than needed, lets keep the explanation of why deduplicate for the commit message

```suggestion
// FindByFSType returns the mount points for filesystems of the given type,
// in order of appearance, with duplicates removed.
```

internal/mountinfo/mountinfo.go
dylandreimerink: Second part of the comment is a design discussion, should be in the commit message not the doc comment.

```suggestion
// ParseEntries parses mountinfo data read from r. Exposed for tests; most
// callers want [Read] or [FindByFSType].
func ParseEntries(r io.Reader) ([]Entry, error) {
```

internal/tracefs/kprobe.go
```diff
@@ -135,6 +152,32 @@ var getTracefsPath = sync.OnceValues(func() (string, error) {
...
 })
 
+// findTracefsInEntries scans a list of mount entries for tracefs, then for
+// debugfs with an existing tracing/ subdirectory. Subdirectory bind mounts
+// (entries whose Root field is not "/") are skipped, since their mount
+// point doesn't expose the tracefs root files (kprobe_events,
+// uprobe_events, events/) that callers need.
+//
+// Returns the empty string (and a nil error) when no usable mount is
+// found, leaving the caller to fall back to canonical-path probing.
```
dylandreimerink: Comments on functions like this should be aimed at its user, not a description of what the function does.


```suggestion
// findTracefsInEntries returns the first occurrence of a tracefs in the
// given `entries`.
//
// Returns a empty string when no usable mount is found.
```

internal/tracefs/kprobe.go
```diff
@@ -135,6 +152,32 @@ var getTracefsPath = sync.OnceValues(func() (string, error) {
...
+		if e.FSType != "debugfs" || e.Root != "/" {
+			continue
+		}
+		tracing := filepath.Join(e.MountPoint, "tracing")
+		if info, err := os.Stat(tracing); err == nil && info.IsDir() {
+			return tracing, nil
+		}
+	}
+	return "", nil
+}
```
dylandreimerink: This function never returns an error value. You can change it to just return a string.

dylandreimerink (changes requested):
Overall direction of this PR makes sense. Just a few things to change.

General rules of thumb for comments:
* Doc comments (comments on a function / type that end up in generated documentation) should be user facing, so they should talk about what a function does, its inputs, and its outputs. But not the why of things, that is for git.
* Inline comments should not explain what code does, but why it does it that way when doing something unintuitive or unconventional.
* Comments should be proportional to the code, no 10 line comment for a 3 line function.

LLMs have a tendency to be overly verbose with comments, I suspect due to the way they do "reasoning" about code, so sometimes you need to strip some of that verbosity.

internal/tracefs/kprobe.go
yoav-orca: Trimmed

yoav-orca:
@dylandreimerink I fixed all the issues you highlighted

dylandreimerink (approved):
Thank you for the changes!

internal/mountinfo/mountinfo_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	entries, err := parseEntries(strings.NewReader(mountinfo))
+	qt.Assert(t, qt.IsNil(err))
+	qt.Assert(t, qt.HasLen(entries, 2))
+	qt.Assert(t, qt.Equals(entries[0].Root, "/"))
+	qt.Assert(t, qt.Equals(entries[1].Root, "/events"))
+}
+
+func TestParseEntriesLongLine(t *testing.T) {
+	// Overlay lowerdir lists in containers commonly exceed bufio.Scanner's
```
ti-mo, line 26: Why are we exercising whether the Go stdlib behaves as expected? Do long lines require different treatment by the caller somehow?

internal/mountinfo/mountinfo_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/go-quicktest/qt"
+)
+
+func TestParseEntriesPreservesRoot(t *testing.T) {
+	// Subdirectory bind mounts retain their original fstype but expose a non-/
```
ti-mo, line 11: Why is this its own test? Did the original code parse the root field incorrectly?

internal/tracefs/kprobe.go
```diff
@@ -135,6 +152,28 @@ var getTracefsPath = sync.OnceValues(func() (string, error) {
 	return "", errors.New("neither debugfs nor tracefs are mounted")
 })
 
+// findTracefsInEntries returns the first occurrence of a tracefs in the
+// given entries.
+//
+// Returns an empty string when no usable mount is found.
+func findTracefsInEntries(entries []mountinfo.Entry) string {
+	for _, e := range entries {
+		if e.FSType == "tracefs" && e.Root == "/" {
```
ti-mo, line 161: Is there a reason to explicitly reject bind-mounted tracefs mounts? Why can't we use the first-found entry?

ti-mo (commented):
I have some leftover questions, please take a look.

PR #2006, opened 2026-05-10
internal/kconfig: reject bool arrays for string values

immanuwell:
Fixes a tiny `.kconfig` type check footgun.

`putValueString` said string configs need a one-byte non-bool int array, but the check used `&&`. So `_Bool CONFIG_DEFAULT_HOSTNAME[1] __kconfig` could accept a string value and copy bytes into a bool array. Thats not a valid target type.

Repro:
1. Add a quoted value to the existing bool-array case in `TestPutValue`.
2. Run `go test ./internal/kconfig -run TestPutValue -count=1`.
3. Before this patch it fails because `PutValue` returns nil. After this patch it rejects it.

Checked:
- `go test ./internal/kconfig -count=1`
- `go test ./internal ./internal/linux ./internal/sysenc ./internal/platform ./internal/unix -count=1`

### DIFF
--- a/internal/kconfig/kconfig.go
+++ b/internal/kconfig/kconfig.go
@@ -183,8 +183,8 @@ func putValueString(data []byte, typ btf.Type, value string) error {
 
 	// Any Int, which is not bool, of one byte could be used to store char:
 	// https://github.com/torvalds/linux/blob/1a5304fecee5/tools/lib/bpf/libbpf.c#L3637-L3638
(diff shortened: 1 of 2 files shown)


dylandreimerink (approved):
Nice find. Changes make sense.

florianl (approved):

dylandreimerink:
Windows CI is failing due to errors in ebpf-for-windows main, ignoring.

PR #2011, opened 2026-05-15
prog.go: cache BTF for CAP_SYS_ADMIN-less freplace

mejedi:
Before the change, performing freplace with `cilium/ebpf` required `CAP_SYS_ADMIN`.

The subprogram to replace is specified via `AttachTarget` (`*Program`) and `AttachTo` (`string`) in the replacement `ProgramSpec`. Since the kernel expects a BTF type ID of the subprogram to replace, the library has to resolve the `AttachTo` string. It needs BTF of the `AttachTarget`. To obtain BTF from the kernel `CAP_SYS_ADMIN` is needed.

Extend Program with `btf` (`*btf.Handle`). It is populated for `Program`-s loaded by the library. Program's `.Handle()` leverages the cached BTF transparaently, enabling `CAP_SYS_ADMIN`-less freplace.

Add `SetHandle` method to `Program` to cover program references obtained via `NewProgramFromFD()`, and `LoadPinnedProgram()`. The intended usage is as follows:

 * a `CAP_SYS_ADMIN`-less user obtains a program reference and the corresponding BTF (e.g. both from a pin);

 * calls `.SetHandle()` on the `Program`;

 * freplace succeeds with the `Program` as `AttachTarget`.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -242,6 +242,8 @@ type Program struct {
 	name       string
 	pinnedPath string
 	typ        ProgramType
+
+	btf *btf.Handle
 }
 
 // NewProgram creates a new Program.
@@ -282,7 +284,7 @@ var (
 	kfuncBadCall = fmt.Appendf(nil, "invalid func unknown#%d\n", kfuncCallPoisonBase)
 )
 
-func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache) (*Program, error) {
+func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache) (result *Program, _ error) {
 	if len(spec.Instructions) == 0 {
 		return nil, errors.New("instructions cannot be empty")
 	}
@@ -356,12 +358,18 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
 		attr.LineInfo = sys.SlicePointer(lib)
 	}
 
+	var handle *btf.Handle
 	if !b.Empty() {
-		handle, err := btf.NewHandle(&b)
+		var err error
+		handle, err = btf.NewHandle(&b)
 		if err != nil {
 			return nil, fmt.Errorf("load BTF: %w", err)
 		}
-		defer handle.Close()
+		defer func() {
+			if result == nil || result.btf != handle {
+				handle.Close()
+			}
+		}()
 
 		attr.ProgBtfFd = uint32(handle.FD())
 	}
@@ -471,7 +479,7 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
 		// Loading with logging disabled should never retry.
 		fd, err = sys.ProgLoad(attr)
 		if err == nil {
-			return &Program{"", fd, spec.Name, "", spec.Type}, nil
+			return &Program{"", fd, spec.Name, "", spec.Type, handle}, nil
 		}
 	} else {
 		// Only specify log size if log level is also specified. Setting size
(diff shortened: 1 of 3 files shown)


prog.go
```diff
@@ -684,6 +699,44 @@ func (p *Program) Handle() (*btf.Handle, error) {
...
+// Type info is auto-cached for Programs loaded by the library itself.
+// Use SetHandle() with program references obtained via NewProgramFromFD
+// and LoadPinnedProgram.
+func (p *Program) SetHandle(h *btf.Handle) error {
+	info, err := p.Info()
+	if err != nil {
+		return err
+	}
+
+	handleInfo, err := h.Info()
```
dylandreimerink, line 719: If someone were to pass in `nil` this panics. An explicit nil check can't hurt, so we always return a nice error.

prog.go
```diff
@@ -242,6 +242,11 @@ type Program struct {
 	name       string
 	pinnedPath string
 	typ        ProgramType
+
+	// Program's BTF (if known). Querying BTF from the kernel requires
+	// CAP_SYS_ADMIN; keeping a copy available makes it possible to
+	// avoid granting this capability.
```
dylandreimerink: I think this comment is not needed, should be evident in a git blame + commit message.

dylandreimerink (changes requested):
I like this PR, all makes sense to me. Just got two minor comments.

prog.go
mejedi: Hi @dylandreimerink ! Do you feel strongly about adding a nil check?
 * Personally I feel that panic is perfectly fine. I am happy to make the change though since I am a guest and you are a host :)
 * However, it doesn't look like defensive nil checks are an established practice in the library: see  `btf.NewHandle`. I DO see that `btf.Builder.Add(typ btf.Type)` checks for `typ` being `nil` or a `nil` interface value. I do not believe in one-size-fits-all-s; the builder is probably a special case. It is complex; panics generated deep inside the marshaller's guts are hard to troubleshoot. Here we have something extremely simple.
 * Where to draw the line? The pointer the method is called on (`prog *Program`) can be nil as well.

So as I said, happy to make the change. Will appreciate you elaborating a bit.

dylandreimerink: Fair questions.  

> Personally I feel that panic is perfectly fine.

My general view is libraries should avoid panics as much as possible. We only throw panics in rare cases where errors returns are too costly, and only to indicate bugs in the library.
Panics may cause the whole application to abort when uncaught and thus should usually only happen when a situation is truly unrecoverable.

Explicit checks and error returns also makes our expectations of the input clear. In a different programming language we may have used debug mode asserts for this. That way anyone reading this code does not have to wonder "could this pointer be nil?", doing so with a check over a comment is nice.

> it doesn't look like defensive nil checks are an established practice in the library

> Where to draw the line? The pointer the method is called on (`prog *Program`) can be nil as well.

We likely do to little nil checking. Go not distinguishing between a pointer that is nil-able and one we always expect to have a valid value is a bit frustrating sometimes. So I would say we should look at what is reasonable which is subjective (sorry).

I feel like most Go programmers understand that calling methods on nil pointers is not valid in 95% of cases. And thus its reasonable to place a nil-check burden on them to avoid panics. But for arguments to a function or method its a tossup, sometimes you can (`btf.Handle.Spec(nil)` or `btf.Builder.Marshal(nil, nil)` for example) and sometimes you can not. So we rely on the user reading the comments or the implementation to discover if nil pointers are or are not allowed.

In this case the method is called `SetHandler`. I have seen similar APIs where you can un-set/delete references by passing in nil.

So, I would personally say:
* Methods on pointer values, no nil check (unless specifically desirable `CollectionSpec.Copy`).
* Arguments of exported functions/methods, yes
* Arguments of internal functions/methods, probably no nil check (let the user facing functions do that).

> Will appreciate you elaborating a bit.

I hope that helps explain my viewpoint.

mejedi: Thank you for detailed response. I made requested changes.

dylandreimerink (approved):
Thank you for the changes, this seems good to me!

PR #2017, opened 2026-05-26
fix(map): defer haveBatchAPI check in BatchLookupCmd until syscall fails

### LINKED ISSUE #2013: Improve `haveBatchAPI()` feature probing in `BatchLookup()`
### Describe the bug

`BatchLookup()` calls into [batchLookupCmd()](https://github.com/cilium/ebpf/blob/v0.21.0/map.go#L1324), which in turn calls [`haveBatchAPI()`](https://github.com/cilium/ebpf/blob/v0.21.0/syscalls.go#L246) indiscriminately, caching the result to avoid future calls. The problem here is that `haveBatchAPI()` creates a small temporary ebpf map which it then batch updates to verify that the system supports batching. However, the creation of the map implies `CAP_BPF` capabiliries.

[`BatchDelete`](https://github.com/cilium/ebpf/blob/v0.21.0/map.go#L1469) (and `BatchUpdate()`, for that matter) was improved in https://github.com/cilium/ebpf/pull/1071 to have the `haveBatchAPI()` being called only [after](https://github.com/cilium/ebpf/blob/v0.21.0/map.go#L1469) a failure is observed while calling the underlying BPF syscall, to check if the cause was lack of batching support.

I'm suggesting the same for the same to be done for `BatchLookup()`, otherwise this can lead to awkward to debug errors like 
```
batch lookup: map batch lookup: map batch api not supported (requires >= v5.6)".
```

Our binary is split in two - A and B,  where A creates the maps and B reads and performs several read operations. B runs with minimal permissions, so suddenly having to require CAP_BPF is surprising and hard to justify.

### How to reproduce

// Bellow BatchUpdate and BatchDelete work, but BatchLookup doesn't.
```
fun() {
    m, err := ebpf.LoadPinnedMap(path, nil)
    if err != nil {
        fmt.Fprintln(os.Stderr, "[run] LoadPinnedMap:", err)
        os.Exit(1)
    }
    defer m.Close() 

    keys := []uint32{1, 2}
    vals := []uint32{11, 22}

    // BatchUpdate first: populates entries we can later delete/lookup.
    n, err := m.BatchUpdate(keys, vals, nil)
    fmt.Printf("[run] BatchUpdate n=%d err=%v\n", n, err)

    // BatchLookup: this is the call cilium gates with the probe.
    var cursor ebpf.MapBatchCursor
    outKeys := make([]uint32, 2)
    outVals := make([]uint32, 2)
    n, err = m.BatchLookup(&cursor, outKeys, outVals, nil)
    fmt.Printf("[run] BatchLookup n=%d err=%v keys=%v vals=%v\n", n, err, outKeys, outVals)
    if errors.Is(err, ebpf.ErrKeyNotExist) {
        fmt.Println("[run] (ErrKeyNotExist is the normal end-of-iteration signal)")
    }

    // BatchDelete: should succeed even without caps on the happy path.
    n, err = m.BatchDelete(keys, nil)
    fmt.Printf("[run] BatchDelete n=%d err=%v\n", n, err)
}
```

### Version information

All versions

s9kt:
fix(map): defer `haveBatchAPI()` check in `BatchLookupCmd()` until syscall fails

`BatchLookup()` called `haveBatchAPI()` indiscriminately, which creates a temporary BPF map implying support for CAP_BPF. `BatchDelete()` and `BatchUpdate()` were fixed in #1071 to only check for the batch API after a syscall failure, same pattern applied here.

Fixes #2013

### DIFF
--- a/map.go
+++ b/map.go
@@ -1355,10 +1355,6 @@ func (m *Map) batchLookupCmd(cmd sys.Cmd, cursor *MapBatchCursor, count int, key
 		return 0, errors.New("a cursor may not be reused across maps")
 	}
 
-	if err := haveBatchAPI(); err != nil {
-		return 0, err
-	}
-
 	keyBuf := sysenc.SyscallOutput(keysOut, count*int(m.keySize))
 
 	attr := sys.MapLookupBatchAttr{
@@ -1378,6 +1374,9 @@ func (m *Map) batchLookupCmd(cmd sys.Cmd, cursor *MapBatchCursor, count int, key
 	_, sysErr := sys.BPF(cmd, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
 	sysErr = wrapMapError(sysErr)
 	if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
+		if err := haveBatchAPI(); err != nil {
+			return 0, err
+		}
 		return 0, sysErr
 	}
 
--- a/map.go
+++ b/map.go
@@ -1374,8 +1374,8 @@ func (m *Map) batchLookupCmd(cmd sys.Cmd, cursor *MapBatchCursor, count int, key
 	_, sysErr := sys.BPF(cmd, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
 	sysErr = wrapMapError(sysErr)
 	if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
-		if err := haveBatchAPI(); err != nil {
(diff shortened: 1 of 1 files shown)


dylandreimerink:
Hey @s9kt, thank you for opening the PR and taking a stab at this. In the end I decided to go with https://github.com/cilium/ebpf/pull/2018 which came in just after due to its addition of the tests.

s9kt:
No problem, thanks @dylandreimerink!

PR #2018, opened 2026-05-26
ebpf: defer batch API probe for batch lookup

### LINKED ISSUE #2013: Improve `haveBatchAPI()` feature probing in `BatchLookup()`

arunsingh:
Fixes #2013.

`BatchLookup()` and `BatchLookupAndDelete()` previously called `haveBatchAPI()` before attempting the actual batch lookup syscall. The feature probe creates a temporary map, which can require `CAP_BPF`, so a process that only needs to read from an already-open map could fail before trying the operation that would otherwise succeed.

This defers the batch API feature probe until after a non-`ENOENT` lookup syscall failure, matching the existing `BatchUpdate()` and `BatchDelete()` behavior. The normal end-of-iteration `ENOENT` path remains unchanged.

The regression test creates maps while privileged, drops all effective capabilities, then verifies batch lookup, update, and delete through the existing map FDs.

Tested:

- `go test -exec sudo -run '^(TestHaveBatchAPI|TestMapBatch|TestMapBatchUnprivileged|TestMapBatchCursorReuse)$' .`
- `git diff --check -- map.go map_test.go`

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -202,6 +202,35 @@ func TestMapBatch(t *testing.T) {
 	}
 }
 
+func TestMapBatchUnprivileged(t *testing.T) {
+	arr := createMap(t, Array, 4)
+	hash := createMap(t, Hash, 4)
+
+	keys := []uint32{0, 1}
+	values := []uint32{42, 4242}
+	qt.Assert(t, qt.IsNil(hash.Update(keys[0], values[0], UpdateAny)))
+	qt.Assert(t, qt.IsNil(hash.Update(keys[1], values[1], UpdateAny)))
+
+	testutils.WithCapabilities(t, nil, func() {
+		tmp := make([]uint32, 2)
+
+		var cursor MapBatchCursor
+		_, err := arr.BatchLookup(&cursor, tmp, tmp, nil)
+		testutils.SkipIfNotSupported(t, err)
+		qt.Assert(t, qt.IsNil(err))
+
+		count, err := arr.BatchUpdate(keys, values, nil)
+		testutils.SkipIfNotSupported(t, err)
+		qt.Assert(t, qt.IsNil(err))
+		qt.Assert(t, qt.Equals(count, len(keys)))
+
+		count, err = hash.BatchDelete(keys, nil)
+		testutils.SkipIfNotSupported(t, err)
+		qt.Assert(t, qt.IsNil(err))
+		qt.Assert(t, qt.Equals(count, len(keys)))
+	})
+}
+
 func TestMapBatchCursorReuse(t *testing.T) {
 	arr1 := createMap(t, Array, 4)
 	arr2 := createMap(t, Array, 4)
--- a/map.go
+++ b/map.go
@@ -1355,10 +1355,6 @@ func (m *Map) batchLookupCmd(cmd sys.Cmd, cursor *MapBatchCursor, count int, key
 		return 0, errors.New("a cursor may not be reused across maps")
 	}
 
-	if err := haveBatchAPI(); err != nil {
-		return 0, err
-	}
-
 	keyBuf := sysenc.SyscallOutput(keysOut, count*int(m.keySize))
 
 	attr := sys.MapLookupBatchAttr{
@@ -1378,6 +1374,9 @@ func (m *Map) batchLookupCmd(cmd sys.Cmd, cursor *MapBatchCursor, count int, key
 	_, sysErr := sys.BPF(cmd, unsafe.Pointer(&attr), unsafe.Sizeof(attr))
 	sysErr = wrapMapError(sysErr)
 	if sysErr != nil && !errors.Is(sysErr, unix.ENOENT) {
+		if haveFeatErr := haveBatchAPI(); haveFeatErr != nil {
+			return 0, haveFeatErr
+		}
 		return 0, sysErr
 	}
 


arunsingh:
Ready for review. I kept the change limited to BatchLookup/BatchLookupAndDelete via batchLookupCmd(), with ENOENT behavior unchanged. The focused regression and root package tests pass on a Linux VM; details are in the PR description.

map_test.go
```diff
@@ -202,6 +202,35 @@ func TestMapBatch(t *testing.T) {
...
+	qt.Assert(t, qt.ErrorIs(err, ErrKeyNotExist))
+	qt.Assert(t, qt.Equals(count, 2))
+	qt.Assert(t, qt.DeepEquals(map[uint32]uint32{
+		keys[0]: values[0],
+		keys[1]: values[1],
+	}, map[uint32]uint32{
+		1: 11,
+		2: 22,
+	}))
+}
```
dylandreimerink: This does not really test the issue, the issue is that batch lookup did not work for unprivileged processes. So we should test it like:

```
func TestMapBatchUnprivileged(t *testing.T) {
	arr := createMap(t, Array, 4)
	testutils.WithCapabilities(t, nil, func() {
		tmp := make([]uint32, 2)

		var cursor MapBatchCursor
		_, err := arr.BatchLookup(&cursor, tmp, tmp, nil)
		testutils.SkipIfNotSupported(t, err)
		qt.Assert(t, qt.IsNil(err))
	})
}
``` 

Create a map, then drop all capabilities, including `CAP_BPF` needed for map creation, then do the batch lookup. This should also be extended to batch update and delete, so we covered all three of them.

dylandreimerink (changes requested):
Thank you! Change seems good to me, just want to change the regression test.

map_test.go
arunsingh: @dylandreimerink I have updated the regression to drop capabilities and cover batch lookup, update, and delete.

dylandreimerink (approved):
Changes look good to me. 

CI failure turns out to be caused by a bug in `WithCapabilities`, on 5.4 we trigger the skip, which breaks the capabilities of other tests that run after `TestMapBatchUnprivileged`.

Fix is underway in https://github.com/cilium/ebpf/pull/2023, so you will have to rebase to pull it in once merged.

arunsingh:
> Changes look good to me.
> 
> CI failure turns out to be caused by a bug in `WithCapabilities`, on 5.4 we trigger the skip, which breaks the capabilities of other tests that run after `TestMapBatchUnprivileged`.
> 
> Fix is underway in #2023, so you will have to rebase to pull it in once merged.

Thanks @dylandreimerink . As you pointed out that #2023 is merged now, so I rebased this branch on current main and pushed the updated commit. The focused batch regression still passes locally. I’ll keep an eye on CI from here.

PR #2020, opened 2026-05-27
chore: improve ebpf maintenance path

otjdiepluong:
Summary:
- Add or tighten focused edge-case tests or type assertions in fuzz_test.go, info_test.go related to CI, Docker, Kubernetes, build tooling, release automation; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


ti-mo:
Hi @otjdiepluong, what benefit does this provide in practice? I don't see how the cases you added exercise functionally different code paths from a case with a missing field.

PR #2021, opened 2026-05-27
BTF: fixed panics during parsing of malformed input

### LINKED ISSUE #2019: BTF string offset boundary check can panic parser on malformed ELF/BTF input
### Describe the bug

`github.com/cilium/ebpf` can panic while parsing malformed BTF / BTF.ext metadata through public parser APIs such as `LoadCollectionSpec` and `LoadCollectionSpecFromReader`.

The issue is in shared BTF string-table lookup logic. A malformed `.BTF` or `.BTF.ext` record can set a non-zero string offset equal to `BTF StringLen`. This boundary value is incorrectly treated as valid, even though it points to the end of the string table. The parser then searches for a NUL byte in an empty slice, receives `-1`, and reaches an invalid slice bound, causing a Go panic instead of returning a parse error.

Expected behavior:

Malformed BTF / BTF.ext input should return an error.

Actual behavior:

The parser panics. In a process that parses less-trusted eBPF ELF/BTF artifacts, this can cause process-level denial of service.

I validated the same boundary issue through multiple fields:

- `.BTF` `btf_type.NameOff`
- `.BTF.ext` `func_info` `SecNameOff`
- `.BTF.ext` `line_info` `SecNameOff`
- `.BTF.ext` `line_info` `FileNameOff`
- `.BTF.ext` `line_info` `LineOff`

The demonstrated impact is denial of service only.

### How to reproduce

I prepared a compact reproduction package with PoC source, commands, observed results, and evidence mapping:

https://gist.github.com/thesmartshadow/256bff0f8042c584f993ace89074a815

From the root of a `github.com/cilium/ebpf` checkout:

```sh
chmod +x ghsa-btf-string-offset-gist/reproduce.sh
./ghsa-btf-string-offset-gist/reproduce.sh
````

Manual steps:

```sh
go run ghsa-btf-string-offset-gist/mutate_elf.go \
  -in cmd/bpf2go/testdata/minimal-el.elf \
  -out /tmp/cilium-ebpf-btf-nameoff-eq-stringlen.elf

go run ghsa-btf-string-offset-gist/worker_dos_poc.go \
  -input cmd/bpf2go/testdata/minimal-el.elf

go run ghsa-btf-string-offset-gist/variants_check.go \
  cmd/bpf2go/testdata/minimal-el.elf
```

Observed behavior:

* the unmodified ELF fixture parses successfully
* the mutated ELF differs by one 32-bit BTF string offset field
* the malformed input causes a panic instead of a returned error
* the worker PoC exits after parsing the malformed artifact
* a follow-up benign artifact cannot be processed because the worker process is already gone

Key observed output includes:

```text
control: one-shot scanner on benign input
exit_code: 0
stdout:
  ACCEPT 01-benign.elf

control: one-shot scanner on malformed ELF should reject, but currently panics
exit_code: 2
stderr:
  panic: runtime error: slice bounds out of range [376:375]

impact: long-running artifact scanner worker
worker_response: ACCEPT 01-benign.elf
worker_exit_code_after_malicious: 2
follow_up_write_error: write |1: file already closed
worker_process_state_exited: true

dylandreimerink:
Bug reports and fuzzing found a few cases where our BTF parsing logic would panic instead of returning errors when presented with malformed input. This PR fixes the reported issue as well as the two, after these fixes no further issue were found with an hour of fuzzing.

Fixes: #2019

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -214,7 +214,7 @@ func loadRawSpec(btf []byte, base *Spec) (*Spec, error) {
 	}
 	btf = btf[header.HdrLen:]
 
-	if int(header.StringOff+header.StringLen) > len(btf) {
+	if uint64(header.StringOff)+uint64(header.StringLen) > uint64(len(btf)) {
 		return nil, fmt.Errorf("string table is out of bounds")
 	}
 	stringsSection := btf[header.StringOff : header.StringOff+header.StringLen]
@@ -224,7 +224,7 @@ func loadRawSpec(btf []byte, base *Spec) (*Spec, error) {
 		return nil, fmt.Errorf("read string section: %w", err)
 	}
 
-	if int(header.TypeOff+header.TypeLen) > len(btf) {
+	if uint64(header.TypeOff)+uint64(header.TypeLen) > uint64(len(btf)) {
 		return nil, fmt.Errorf("types section is out of bounds")
 	}
 	typesSection := btf[header.TypeOff : header.TypeOff+header.TypeLen]
--- a/btf/btf_types.go
+++ b/btf/btf_types.go
@@ -120,6 +120,10 @@ func parseBTFHeader(buf []byte) (*btfHeader, binary.ByteOrder, error) {
 		return nil, nil, errors.New("header length shorter than btfHeader size")
 	}
 
+	if len(buf) < int(remainder) {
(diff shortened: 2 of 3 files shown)


florianl (approved):
great catch!

thesmartshadow:
Thanks

PR #2025, opened 2026-05-29
fix: prevent uint32 overflow in struct_ops bounds check

### LINKED ISSUE #2024: elf: malformed struct_ops DATASEC offset can panic LoadCollectionSpecFromReader
### bug description

`LoadCollectionSpecFromReader` can panic while parsing malformed struct_ops metadata from an ELF object.

The issue is in the `.struct_ops` DATASEC handling. A malformed object can make the DATASEC variable offset resolve to a large value through `.rel.BTF`. `createStructOpsMap` checks `baseOff + userSize` against the section length using `uint32` arithmetic. That addition can overflow, causing the bounds check to pass before the code slices the `.struct_ops.link` section with the original large offset.

Expected behavior:

Malformed struct_ops metadata should return an error.

Actual behavior:

The parser panics with a slice bounds violation.

This is separate from #2019 and #2022. Those are BTF string/header parser issues; this one is in the ELF struct_ops map construction path.

### reproducing the bug

I reproduced this on upstream `main` at commit `533dfc8`.

Public API regression test:

```go
package ebpf_test

import (
	"bytes"
	"encoding/binary"
	"math"
	"os"
	"testing"

	"github.com/cilium/ebpf"
)

func TestStructOpsDatasecOffsetOverflowRepro(t *testing.T) {
	elf, err := os.ReadFile("testdata/struct_ops-el.elf")
	if err != nil {
		t.Fatal(err)
	}

	// In this fixture, symtab entry 4 is testmod_ops and st_value is at 0x308.
	// The .rel.BTF relocation copies that value into the .struct_ops.link DATASEC offset.
	binary.LittleEndian.PutUint64(elf[0x308:], uint64(math.MaxUint32-3))

	_, err = ebpf.LoadCollectionSpecFromReader(bytes.NewReader(elf))
	if err == nil {
		t.Fatal("expected error")
	}
}
```

Run:

```bash
go test . -run TestStructOpsDatasecOffsetOverflowRepro -count=1
```

Observed failure:

```text
panic: runtime error: slice bounds out of range [4294967292:20]
github.com/cilium/ebpf.(*elfCode).createStructOpsMap
elf_reader.go:1554
github.com/cilium/ebpf.(*elfCode).associateStructOpsRelocs
elf_reader.go:1503
github.com/cilium/ebpf.LoadCollectionSpecFromReader
elf_reader.go:201
```

The relevant code is:

```go
userSize := userSt.Size
baseOff := vsi.Offset
if baseOff+userSize > uint32(len(userData)) {
	return nil, 0, fmt.Errorf("%s exceeds section", mapName)
}

Value: append([]byte(nil), userData[baseOff:baseOff+userSize]...),
```

### Version information

```text
upstream main: 533dfc8
go version go1.26.3 linux/amd64
Linux amd64
```

### patch solution

Use overflow-safe arithmetic before slicing, matching the recent raw BTF bounds-check pattern:

```go
if uint64(baseOff)+uint64(userSize) > uint64(len(userData)) {
	return nil, 0, fmt.Errorf("%s exceeds section", mapName)
}
```

Then add a regression test that verifies malformed struct_ops DATASEC offsets return an error instead of panicking.

DARSHANR007:
### Problem
`LoadCollectionSpecFromReader` could panic while parsing malformed struct_ops metadata from an ELF object.

A malformed object can provide a very large offset that bypasses the bounds check due to an integer overflow. The code then accesses memory outside the section bounds, leading to a panic.
### Solution
Convert `baseOff` and `userSize` to `uint64` before addition to prevent overflow:

```go
// Before (vulnerable)
if baseOff+userSize > uint32(len(userData)) {
    return nil, 0, fmt.Errorf("%s exceeds section", mapName)
}

// After (fixed)
if uint64(baseOff)+uint64(userSize) > uint64(len(userData)) {
    return nil, 0, fmt.Errorf("%s exceeds section", mapName)
}
```

Fixes #2024


struct_ops_overflow_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestStructOpsDatasecOffsetOverflowRepro(t *testing.T) {
+	elf, err := os.ReadFile("testdata/struct_ops-el.elf")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// In this fixture, symtab entry 4 is testmod_ops and st_value is at 0x308.
+	// The .rel.BTF relocation copies that value into the .struct_ops.link DATASEC offset.
+	binary.LittleEndian.PutUint64(elf[0x308:], uint64(math.MaxUint32-3))
```
dylandreimerink: This test is not going to be stable. It breaks as soon as the base ELF changes. In this case I will take the fix without the test.

dylandreimerink (changes requested):
Fix looks good. Lets drop the test, its not stable like this.

struct_ops_overflow_test.go
DARSHANR007: Hey Appreciate the review , ive removed the test .

Thx

dylandreimerink (approved):

PR #2026, opened 2026-05-29
Improve resilience of ELF parser to malformed input

dylandreimerink:
Inspired by https://github.com/cilium/ebpf/issues/2024 I decided to run our existing `FuzzLoadCollectionSpec` fuzz test until it didn't return panics for an hour.

This surfaced quite a number of places where "user input" in the form of an ELF file would cause a panic. These were out of bounds reads from slices, nil pointer dereferenceing, and out of memory exceptions. The out of bounds reads were often caused by integer overflows in bounds checks.

All in all, this should make the library panic less often and throw normal errors instead when faced with bad or unexpected input.


ti-mo:
Great, thanks. :heart:

PR #2033, opened 2026-06-06
link: expose syscall wrapper symbol helper

### LINKED ISSUE #1891: kprobe_multi should automatically add syscall wrapper prefix if necessary
In https://github.com/cilium/ebpf/pull/1890 we worked around the fact that somehow the syscall name changed on arm64. @paulcacheux pointed out that for a plain Kprobe we have logic which automatically adds the syscall wrapper if necessary: https://github.com/cilium/ebpf/blob/e2c357d39ef6498d552d2e0776da3f8415e4bd1f/link/kprobe.go#L176-L198

I think it could make sense to have the same logic for kprobe_multi. Unfortunately this involves making a copy of the slice.

It's unfortunate that the logic currently means that new kernels are penalised due to having to retry creating the kprobe. Maybe it's possible to reverse this.

arunsingh:
Fixes #1891

I have exposed a small helper for deriving platform-specific syscall wrapper symbols, then use it from the kprobe_multi syscall attach test.

Kprobe/Kretprobe can retry syscall symbols with the platform prefix, but doing the same automatically for kprobe_multi is ambiguous because a multi-attach list can contain both syscall entrypoints and ordinary kernel functions. The helper lets callers opt in per symbol without changing KprobeMulti attach semantics.

Tests:

- [x] go test ./link -run '^TestSyscallWrapper$' -count=1
- [x] go test -exec sudo ./link -run '^TestKprobeMultiInput$' -count=1
- [x] go test -exec sudo ./link -run '^(TestKprobeMultiProgramCall|TestKprobeMulti|TestKprobeMultiErrors|TestKprobeMultiCookie)$' -count=1 -v
- [x] git diff --check -- link/kprobe.go link/kprobe_test.go link/kprobe_multi_test.go

Also attempted `go test -exec sudo ./...` in the Ubuntu VM. The focused link tests passed; the broad run hit unrelated `cmd/bpf2go` module-cache setup with `GOPROXY=off` and an unrelated `features/LircMode2` kernel capability failure.

### DIFF
--- a/link/kprobe_test.go
+++ b/link/kprobe_test.go
@@ -27,6 +27,10 @@ var symTests = []string{
 	"unregister_kprobes.part.0", // function body that was split and partially inlined
 }
 
+func TestSyscallWrapper(t *testing.T) {
+	qt.Assert(t, qt.Not(qt.Equals(SyscallWrapper("sys_open"), "sys_open")))
+}
+
 func TestKprobe(t *testing.T) {
 	prog := mustLoadProgram(t, ebpf.Kprobe, 0, "")
 
--- a/link/kprobe.go
+++ b/link/kprobe.go
@@ -50,6 +50,20 @@ func (ko *KprobeOptions) cookie() uint64 {
 	return ko.Cookie
 }
 
+// SyscallWrapper returns the platform-specific kernel wrapper symbol for
+// syscall. If syscall already has the current platform's prefix, it's returned
+// unchanged.
+//
+// For example, on amd64, an input of "sys_open" will return "__x64_sys_open".
+func SyscallWrapper(syscall string) string {
+	prefix := linux.PlatformPrefix()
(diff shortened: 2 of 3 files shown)


link/kprobe_test.go
```diff
@@ -27,6 +28,17 @@ var symTests = []string{
 	"unregister_kprobes.part.0", // function body that was split and partially inlined
 }
 
+func TestSyscallWrapper(t *testing.T) {
```
ti-mo, line 30: This is basically a copy of the helper logic itself, that makes this test a little unnecessary. It's already being exercised in `TestKprobeMultiProgramCall`, that's fine.

ti-mo (changes requested):
Thanks, one nit.

ti-mo:
@arunsingh This also needs a rebase to pick up a fix to `Generate and Fix`.

mmat11 (dismissed):
lgtm

PR #2037, opened 2026-06-11
ringbuf: add RawReader for zero-copy reading and commit batching

ti-mo:
This PR introduces a lower-level API, `ringbuf.RawReader`, to solve two main use cases: batching commits and zero-copy reading. The existing interfaces, Read() and ReadInto(), were rebased on top of the new RawReader for dogfooding and example purposes.

A few of the prior proposals that inspired this work:
- https://github.com/cilium/ebpf/pull/1979
- https://github.com/cilium/ebpf/pull/1981
- https://github.com/cilium/ebpf/pull/1968

### DIFF
--- a/ringbuf/raw_reader.go
+++ b/ringbuf/raw_reader.go
@@ -1,0 +1,215 @@
+package ringbuf
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"runtime"
+	"sync/atomic"
+	"time"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/internal"
+)
+
+// RawReader reads raw samples submitted to a BPF ringbuf.
+//
+// RawReader exposes samples as slices backed by shared ring memory. Access to
+// that memory is scoped to a [Lease] obtained via [RawReader.WithLease].
+// Returned samples remain valid until the next call to [Lease.Commit] or
+// until the session ends.
+//
+// RawReader must not be used concurrently by multiple goroutines, except that
+// [RawReader.Close] and [RawReader.Flush] may be called from another goroutine
+// to interrupt a blocked read.
+//
+// Use [Reader] instead for a thread-safe alternative.
+type RawReader struct {
+	poller poller
+
+	ring       eventRing
+	deadline   time.Time
+	bufferSize int
+	drainErr   error
+
+	inflight, closed atomic.Bool
+}
+
+var errConcurrent = errors.New("concurrent use of RawReader")
+
+// enter marks the start of an in-flight lease. It returns an error if the
+// reader is already in-flight or closed. Every successful call to enter must be
+// paired with a call to [RawReader.leave].
+func (rr *RawReader) enter() error {
+	if !rr.inflight.CompareAndSwap(false, true) {
+		return errConcurrent
+	}
+
+	if rr.closed.Load() {
+		rr.leave()
+		return fmt.Errorf("ringbuf: %w", ErrClosed)
+	}
+
+	return nil
+}
+
+// leave resets the reader's in-flight state.
+func (rr *RawReader) leave() {
+	rr.inflight.Store(false)
+}
+
+// NewRawReader creates a new raw BPF ringbuf reader. The given Map must be of
+// type [ebpf.RingBuf] or [ebpf.WindowsRingBuf].
+func NewRawReader(m *ebpf.Map) (*RawReader, error) {
+	if m.Type() != ebpf.RingBuf && m.Type() != ebpf.WindowsRingBuf {
+		return nil, fmt.Errorf("invalid Map type: %s", m.Type())
+	}
+
+	maxEntries := int(m.MaxEntries())
+	if maxEntries == 0 || !internal.IsPow(maxEntries) {
+		return nil, fmt.Errorf("ringbuffer map size %d is zero or not a power of two", maxEntries)
+	}
+
+	poller, err := newPoller(m.FD())
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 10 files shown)


jschwinger233:
I like it, especially for the "force atomic access" and "remove internal retry". Haven't tried yet, but given current shape of "zero callbacks" (comparing to my PR that passes in anonymous function everywhere), I am sort confident in its performance boost.

ti-mo:
@jschwinger233 Thanks! Also see the (now fixed) benchmarks to see how the different APIs stack up against each other and what the performance cliffs are.

@orishuss Try the benchmarks of package ringbuf to get an idea of the performance characteristics. On my Zen 2/DDR4 machine, pulling the writer position and pushing the reader position take about 30ns each, so if your sample filtering logic is fast enough, you could potentially make some gains by batching commits. Otherwise the cost becomes less relevant. Personally I find the gap between RawReader and ReadInto relatively small, so make sure you carefully consider whether using RawReader is actually worth it. There are some traps using it, take a look at ReadInto to make sure you use it correctly.

ringbuf/ring.go
```diff
@@ -6,96 +6,116 @@ import (
...
+	// See https://github.com/torvalds/linux/blob/v6.8/kernel/bpf/ringbuf.c#L484
+	start := cons & rr.mask
+	header := ringbufHeader{}
+
+retry:
+	header.Len = atomic.LoadUint32((*uint32)(unsafe.Pointer(&rr.data[start])))
+	if header.isBusy() {
+		// Sample has not been committed by the bpf program yet but should be
+		// soon. Busypoll without advancing the consumer position.
+		goto retry
```
florianl, line 102: This loop could cause a busy loop, if the bpf program side stalls. Should we add a `runtime.Gosched()`?

ringbuf/raw_reader.go
```diff
@@ -0,0 +1,215 @@
...
+func (rr *RawReader) SetDeadline(t time.Time) {
+	rr.deadline = t
+}
+
+// Lease provides exclusive access to a [RawReader] for the duration of a
+// [RawReader.WithLease] callback.
+//
+// A Lease is only valid during that callback and must not be used outside of
+// the scope of [RawReader.WithLease].
+type Lease struct {
```
florianl, line 139: I think there is a chance, that users read stale entries. We could avoid this by adding a `atomic.Unint64` in `RawReader`, increment it in `(*RawReader)WithLease()`, setting it to the current value of `RawReader` and add a check in `(*Lease)checkValid()` if the counter in `Lease` is the same as in `RawReader`.

florianl (commented):
I have some reservations about the tight busy loop in `(*ringReader) readSample()`. That said, it’s still the best approach compared to the alternatives we've seen so far.

And sorry for the delay, it's been a busy few weeks!

ringbuf/ring.go
ti-mo: I initially went back and forth with Copilot a few times and eventually decided to leave it out, since the busy bit means the sample was reserved but not yet committed. The verifier rejects reservations without a pairing discard or commit in the same program, so a stall would mean a bpf program getting preempted or a sleepable program entering a long-running task.

For correctness sake, I'd also prefer a Gosched in there to prevent runaway CPU usage under stress. It would also stay closer to the previous implementation, since the errBusy return would typically cause the Go scheduler to be invoked.

ringbuf/raw_reader.go
ti-mo: Could you elaborate on the sequence of events that would cause a stale entry to be read? `RawReader.enter()` already has a CAS to detect concurrent leases being taken.

florianl: I think, `RawReder.enter()`'s CAS guards against two callers both reaching `WithLease()` at the same time. It might not guard against a `Lease` value that was issued during a previous `WithLease()` call, so one that has already gone through `leave()` and being used while a new `WithLease()` is active. At that point `inflight` is true for the new lease, not the stale one, and `ReadSample()` has no way to tell the difference.

  1. First `WithLease()` starts → `inflight = true`
  2. Callback receives `lease1` and stores it somewhere (nothing prevents this)
  3. Callback returns → `defer leave()` → `inflight = false`
  4. Second `WithLease()` starts → CAS succeeds (`inflight` was `false`) → `inflight = true`
  5. User calls `lease1.ReadSample()` with the first lease, `inflight` is `true` again (due to step 4), so there's no signal that `lease1` is expired and it reads ring memory concurrently with the second lease holder

ringbuf/ring.go
florianl: Good point on the verifier. Though even with preemption or sleepable programs holding the busy bit, there is a second concern: while spinning inside `readSample()`, `inflight` stays `true`, so a concurrent `Close()` will also spin waiting for it to clear. `runtime.Gosched()` could handle both cases.

ringbuf/raw_reader.go
ti-mo: Isn't that what the `// A Lease is only valid during that callback and must not be used outside of the scope of [RawReader.WithLease].` comment is for? :wink: Lease is passed primarily by-value for performance reasons, but it's also a decent signal that it shouldn't be copied out of the closure.

ringbuf/ring.go
ti-mo: Gosched() will only yield the OS thread, it won't cause `inflight` to flip to false. With an additional Gosched(), Close() will technically block for even longer. We're talking microseconds here, I don't think this is worth considering.

I'll add it for now so we can move forward with this, we should have some time to polish if needed since we just cut 0.22.

PR #2038, opened 2026-06-12
Preserve verifier logs on EFAULT

### LINKED ISSUE #2003: EFAULT might have useful log that is not displayed
### Describe the bug

In some cases verifier rejects a program and instead logs with useful info shows this message:

```
load program: bad address (hit verifier bug, increase LogSizeStart to fit the log and check dmesg)
```

However, logs says this:

```
...
2192: (56) if w0 != 0x0 goto pc+373
The sequence of 8193 jumps is too complex.
verification time 850563 usec
stack depth 64+8+0+8+8+64+8+0+8+0+40+24+280+112+80+40+8+16+0+0+0+0+16+8+8+24+0+8+0+16+0+0+0+0+0+0
processed 336006 insns (limit 1000000) max_states_per_insn 19 total_states 13435 peak_states 2210 mark_read 177
```

Here is a simple but improper fix:

https://github.com/cilium/ebpf/blob/18b5c5b0444aa99d3a836b294d0d988d4c42883f/prog.go#L526-L530

```diff
 		// EFAULT is returned when the kernel hits a verifier bug, and always
 		// overrides ENOSPC, defeating the buffer growth strategy. Warn the user
 		// that they may need to increase the buffer size manually.
-		return nil, fmt.Errorf("load program: %w (hit verifier bug, increase LogSizeStart to fit the log and check dmesg)", err)
+		return nil, internal.ErrorWithLog("load program (hit verifier bug, increase LogSizeStart to fit the log and check dmesg)", err, logBuf)
 
 	case errors.Is(err, unix.EINVAL):
 		if bytes.Contains(tail, coreBadCall) {
```

It took several weeks of blind debugging when we could use at least some info with the logs

### How to reproduce

-

### Version information

main branch

immanuwell:
Fixes #2003

When program load fails with `EFAULT`, we currently drop verifier output and only return the generic hint.
That is kinda rough on older kernels, because the partial log can still show the real reject reason.

This keeps the hint, but if the verifier already wrote something we return it as `VerifierError` instead of hiding it.

How to repro

1. Use a kernel where verifier log growth gets short-circuited by `EFAULT`. The report in #2003 hit this on kernels starting at 5.4.
2. Load a program that trips this path. In #2003 the log already had `The sequence of 8193 jumps is too complex.`
3. Current output is only `load program: bad address (hit verifier bug, increase LogSizeStart to fit the log and check dmesg)`
4. With this patch you still get that hint, plus the captured verifier lines, so debugging is not blind anymore.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -517,6 +517,10 @@ func newProgramWithOptions(spec *ProgramSpec, opts ProgramOptions, c *btf.Cache)
 		return nil, fmt.Errorf("load program: %w", err)
 	}
 
+	return nil, programLoadError(err, logBuf, spec)
+}
+
+func programLoadError(err error, logBuf []byte, spec *ProgramSpec) error {
 	end := bytes.IndexByte(logBuf, 0)
 	if end < 0 {
 		end = len(logBuf)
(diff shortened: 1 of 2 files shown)


ti-mo:
@immanuwell Thanks, I moved the function back up to get a usable diff. Also removed the conditions around returning the original error or EFAULT; I don't think we need to overthink this. VerifierError should already take care of formatting itself correctly when the log is empty. I'll merge this in its current state, please give it a try.

ti-mo (approved):

PR #2040, opened 2026-06-20
link: add Info() support for Iter link type

### LINKED ISSUE #506: link: LoadPinnedLink doesn't expose link information

yifeng2019uwb:
Implements Info() on the Iter link type, returning IterInfo with
the iterator's TargetName (e.g. "bpf_map", "task").

Also adds IterLinkInfo.info() to internal/sys to satisfy the
sys.Info interface required by queryInfoWithString.

Related to #506

### DIFF
--- a/link/iter_test.go
+++ b/link/iter_test.go
@@ -6,9 +6,27 @@ import (
 	"io"
 	"testing"
 
+	"github.com/go-quicktest/qt"
+
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal/testutils"
 )
+
+func TestIterInfo(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.9", "bpf_map iter")
+
+	prog := mustLoadProgram(t, ebpf.Tracing, ebpf.AttachTraceIter, "bpf_map")
+
+	it, err := AttachIter(IterOptions{Program: prog})
+	qt.Assert(t, qt.IsNil(err))
+	defer it.Close()
+
+	info, err := it.Info()
+	qt.Assert(t, qt.IsNil(err))
+	qt.Assert(t, qt.Equals(IterType, info.Type))
+	iterInfo := info.Iter()
+	qt.Assert(t, qt.Equals(iterInfo.TargetName, "bpf_map"))
+}
 
 func TestIter(t *testing.T) {
 	testutils.SkipOnOldKernel(t, "5.9", "bpf_map iter")
--- a/internal/sys/syscall.go
+++ b/internal/sys/syscall.go
@@ -81,6 +81,10 @@ func (i *RawTracepointLinkInfo) info() (unsafe.Pointer, uint32) {
 	return unsafe.Pointer(i), uint32(unsafe.Sizeof(*i))
 }
 
+func (i *IterLinkInfo) info() (unsafe.Pointer, uint32) {
+	return unsafe.Pointer(i), uint32(unsafe.Sizeof(*i))
+}
+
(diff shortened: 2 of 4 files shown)


yifeng2019uwb:
Hi, I am new to ebpf and just learnt your AI policy. So I need clarify:
I used the Claude Code to help understand existing code base patterns, discuss the solution. But I did all code review, test. I tested the code manually in my DigitalOcean VM(since i use maxOS, tried test with lima but failed). I am also working on my personal ebpf project for a couple month(public repo can prove my work). So I am fully understand how is the issue and my solution mostly refer to the solution for the same issue with rawtracepoint: https://github.com/cilium/ebpf/commit/6cf7086b91cca859914f07c6c24832af66453e50

link/iter_test.go
```diff
@@ -8,8 +8,30 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/internal/testutils"
+
+	"github.com/go-quicktest/qt"
```
rgo3: Nit: imports of external dependencies go above the import block for internal dependencies (github.com/cilium/ebpf/*)

rgo3 (dismissed):
Thanks for the contribution!

Please squash the two commits into one, no need for a separate commit to address the linter issue. Otherwise LGTM!

link/iter_test.go
```diff
@@ -8,8 +8,30 @@ import (
...
 )
 
+func TestIterInfo(t *testing.T) {
+	testutils.SkipOnOldKernel(t, "5.9", "bpf_map iter")
+
+	prog := mustLoadProgram(t, ebpf.Tracing, ebpf.AttachTraceIter, "bpf_map")
+
+	it, err := AttachIter(IterOptions{Program: prog})
+	if err != nil {
```
ti-mo: Nit here and below:
```suggestion
	qt.Assert(t, qt.IsNil(err))
```

ti-mo (commented):
Thanks! Also left a few nits.

link/iter_test.go
yifeng2019uwb: Thank you for catch this! I've fixed this in the latest commit and now the imports followed the order: system/standard, external, internal.

link/iter_test.go
yifeng2019uwb: Already fixed! All error handling in iter_test.go now uses qt.Assert consistently - throughout all tests in the file, not just the new ones.

rgo3: I think this went a bit too far :sweat_smile:. As general feedback I would try to address feedback in the scope of your proposed changes. If there are other instances that can be addressed they potentially warrant a future PR. Although in this case I think the proposed fixes to the error handling look a bit too eager, as they remove context for the developers that run into potential failures of these tests.

yifeng2019uwb: Thank you for the feedback! You're absolutely right. I've reverted the changes to only the new TestIterInfo unit test uses qt.Assert, old tests keep original error handling.
I'll strictly keep changes within PR scope. Refactoring should be separate PRs. Appreciate the guidance!

rgo3 (approved):

ti-mo (approved):

PR #2045, opened 2026-06-29
feat: add support for loading ebpf programs with typed ksyms

usamasaqib:
This PR builds upon #1578 to add support for typed ksyms in bpf programs.
https://lore.kernel.org/bpf/20200929235049.2533242-1-haoluo@google.com/


linker.go
```diff
@@ -505,13 +520,71 @@ func resolveKsymReferences(insns asm.Instructions) error {
...
+		return cleanup, fmt.Errorf("unable to get list of all loaded modules in btf cache: %w", err)
+	}
+
+	for _, fixup := range typedFixups {
+		var kernelVar *btf.Var
+
+		foundSpec := btfSpec
+		foundMod := "vmlinux"
+		btfVar := fixup.ksymMeta.BtfVar
+		if err := btfSpec.TypeByName(btfVar.Name, &kernelVar); err != nil {
```
brycekahle: I think you might want to pass a copy of `fixup.ksymMeta.BtfVar.Type` here (and for the module `TypeByName`) as the second argument. It helps narrow down matches if there are multiple with the same name.

usamasaqib: The kernel expects the btf id to be for a `KIND_VAR`. This is also what libbpf does when looking up the btf id, i.e., specifically limit it only to `KIND_VAR` https://elixir.bootlin.com/linux/v7.1/source/tools/lib/bpf/libbpf.c#L8634

Using `fixup.ksymMeta.BtfVar.Type` directly here would either have the type be `extern` or the underlying type (if we use `btf.UnderlyingType`) both of which would trip the verifier [here](https://elixir.bootlin.com/linux/v7.1/source/kernel/bpf/verifier.c#L17959) since it explicitly checks for `KIND_VAR`

dylandreimerink:
Just adding some context here to help with review, mostly for myself since I had to read up on the mechanism. Please correct me if any of this is wrong.

Starting with typeless ksyms (on which this builds). A user writes a program (this is the simplified selftest):

```c
__u64 out__bpf_link_fops = -1;

extern const void bpf_link_fops __ksym;

SEC("raw_tp/sys_enter")
int handler(const void *ctx)
{
    out__bpf_link_fops = (__u64)&bpf_link_fops;
}
```

The `__ksym` annotation puts these variables in the ".ksyms" section, which compilers do not actually emit. But they do emit a BTF datasec containing type info for each variable declared. These are read by `elfCode.loadKsymsSection` and put into `ec.ksyms`. Currently this map has `struct{}` as value, which this PR changes to store the BTF as given to us in the ELF.

The ksyms are referenced by relocation entries for instructions. In this case a load instruction created for `&bpf_link_fops`. When we conclude that the instruction is a LDIMM64 and we have a ksym matching the relocation name, we set `ksymMeta` metadata on the instruction.

During loading (in `resolveKsymReferences`) we use kallsyms to find the kernel symbol by name and its associated address, we then patch the instructions imm / constant value with this memory address. The kernel sees this as just a scalar value. Note: during collection loading we also do a kallsyms lookup to prepopulate a cache to speed and deduplicate effort for individual programs.

---

Now for typed ksyms the process is similar and different. The selftest looks like this:

```c
#include "vmlinux.h"

__u64 out__runqueues_addr = -1;
__u64 out__bpf_prog_active_addr = -1;

__u32 out__rq_cpu = -1; /* percpu struct fields */
int out__bpf_prog_active = -1; /* percpu int */

__u32 out__this_rq_cpu = -1;
int out__this_bpf_prog_active = -1;

__u32 out__cpu_0_rq_cpu = -1; /* cpu_rq(0)->cpu */

extern const struct rq runqueues __ksym; /* struct type global var. */
extern const int bpf_prog_active __ksym; /* int type global var. */

SEC("raw_tp/sys_enter")
int handler(const void *ctx)
{
	struct rq *rq;
	int *active;
	__u32 cpu;

	out__runqueues_addr = (__u64)&runqueues;
	out__bpf_prog_active_addr = (__u64)&bpf_prog_active;

	cpu = bpf_get_smp_processor_id();

	/* test bpf_per_cpu_ptr() */
	rq = (struct rq *)bpf_per_cpu_ptr(&runqueues, cpu);
	if (rq)
		out__rq_cpu = rq->cpu;
	active = (int *)bpf_per_cpu_ptr(&bpf_prog_active, cpu);
	if (active)
		out__bpf_prog_active = *active;

	rq = (struct rq *)bpf_per_cpu_ptr(&runqueues, 0);
	if (rq) /* should always be valid, but we can't spare the check. */
		out__cpu_0_rq_cpu = rq->cpu;

	/* test bpf_this_cpu_ptr */
	rq = (struct rq *)bpf_this_cpu_ptr(&runqueues);
	out__this_rq_cpu = rq->cpu;
	active = (int *)bpf_this_cpu_ptr(&bpf_prog_active);
	out__this_bpf_prog_active = *active;

	return 0;
}
```

In this self test the addresses are to per-CPU variables, which with the help of the `bpf_per_cpu_ptr` kfunc can be turned into valid pointers the value for a given CPU. The return value also have a BTF type the verifier understands and allows for further dereferencing such as `rq->cpu`.

In order for the verifier to validate this it needs the BTF type and it needs to validate that the address is valid. The way this is implemented is that we do not do the address resolution in userspace but this kernel does this for us. What we need to do is find the kernel BTF type ID of the global variable and put this type ID in the imm / constant value of the instruction + a pseudo value marker in the src register field.

So fundamentally most of this is the same accept for plumbing, then at the linking/loading stage we do a kfunc like BTF type resolution instead of doing a kallsyms lookup.

asm/register.go
```diff
@@ -41,6 +41,7 @@ const (
 	PseudoFunc      = R4 // BPF_PSEUDO_FUNC
 	PseudoKfuncCall = R2 // BPF_PSEUDO_KFUNC_CALL
 	PseudoMayGoto   = R0 // BPF_MAY_GOTO
+	PseudoBtfId     = R3 // BPF_PSEUDO_BTF_ID
 )
```
dylandreimerink, line 45: Lets put the map_fd, map_value, and btf_id consts together like in the kernel sources.

```c
const (
	PseudoMapFD     = R1 // BPF_PSEUDO_MAP_FD
	PseudoMapValue  = R2 // BPF_PSEUDO_MAP_VALUE
	PseudoBTFID     = R3 // BPF_PSEUDO_BTF_ID
	PseudoCall      = R1 // BPF_PSEUDO_CALL
	PseudoFunc      = R4 // BPF_PSEUDO_FUNC
	PseudoKfuncCall = R2 // BPF_PSEUDO_KFUNC_CALL
	PseudoMayGoto   = R0 // BPF_MAY_GOTO
)
```

linker.go
```diff
@@ -505,13 +520,71 @@ func resolveKsymReferences(insns asm.Instructions) error {
...
+					return i.IsModule() && i.Name == foundMod
+				})
+				if err != nil {
+					return cleanup, fmt.Errorf("unable to get btf handle for module %s: %w", foundMod, err)
+				}
+
+				keepAlive[foundMod] = btfHandle
+			}
+			btfObjFD = int64(btfHandle.FD())
+		}
```
dylandreimerink: I think this can mostly be replaced by a call to `findTargetInKernel`, see `fixupKfuncs` for example usage. It does a lookup in vmlimux and modules, returns the handle when found in a module ect.

linker.go
```diff
@@ -452,15 +452,22 @@ func resolveKconfigReferences(insns asm.Instructions) (_ *Map, err error) {
...
-	var fixups []fixup
+	var untypedFixups []fixup
+	var typedFixups []fixup
 
+	keepAlive := make(map[string]*btf.Handle)
+	cleanup := func() {
+		for _, h := range keepAlive {
+			h.Close()
+		}
+	}
```
dylandreimerink: Returning this `cleanup` function isn't nice. I believe we can defer this just like we do in `fixupKfuncs` to close module handles.

testdata/ksym.c
```diff
@@ -31,9 +31,31 @@ __section("socket") int ksym_test() {
...
-__section("socket") int ksym_missing_test() {
+__section("socket") int untyped_ksym_missing_test() {
 	if (&non_existing_symbol == 0) {
 		return 1;
 	}
 	return 0;
 }
+
+extern const int bpf_prog_active __ksym __weak;
+
```
dylandreimerink, line 39: I would also like to see the struct type ksyms like in the kernel selftest.

dylandreimerink (changes requested):
Nice work so far, I think this is close. Left some changes requests.

testdata/ksym.c
usamasaqib: Done. Picked `softnet_data` since it was already present when the support for typed ksyms was added in 5.10

linker.go
usamasaqib: Done

linker.go
```diff
@@ -507,11 +516,44 @@ func resolveKsymReferences(insns asm.Instructions) error {
...
+			return nil, fmt.Errorf("unable to find type id for variable %s in the kernel btf: %w", btfVar.Name, err)
+		}
+
+		btfObjFD := int64(0)
+		if module != nil {
+			btfObjFD = int64(module.FD())
+		}
+
+		fixup.Instruction.Src = asm.PseudoBtfId
+		fixup.Instruction.Constant = int64(uint32(typeId)) | (btfObjFD << 32)
```
dylandreimerink: Module support is neat, could we also add a test case for that?

Is see there is a test per_cpu var in the bpf test module. https://elixir.bootlin.com/linux/v7.1.2/source/tools/testing/selftests/bpf/progs/test_ksyms_module.c#L20

testdata/ksym.c
```diff
@@ -31,9 +31,40 @@ __section("socket") int ksym_test() {
...
+	bpf_map_update_elem(&array_map, &i, &val, 0);
+
+    i = 1;
+    val = (uint64_t)&softnet_data;
+    bpf_map_update_elem(&array_map, &i, &val, 0);
+
+	return 0;
+}
+
+extern const struct file_operations__local non_existing_typed_ksym __ksym __weak;
```
dylandreimerink: We seem to be missing the declaration for this type, how does that work?

dylandreimerink (changes requested):
Thanks for addressing the feedback, one more round

testdata/ksym.c
usamasaqib: Ah this is a leftover from a previous iteration, I forgot to fix the types. 
It works because the test its used in expect a failure for a non existing variable.

linker.go
usamasaqib: The reason I didn't add module support is because I wasn't sure whats available in the test environment. I was not aware of the bpf test module. Can we expect this to be loaded in the VMs?

testdata/ksym.c
usamasaqib: I guess this is not a very useful test for typed ksyms. Should I remove it altogether?

linker.go
dylandreimerink: Yes, we also use it when testing kfuncs see `testdata/kfunc-mod.c`.

testdata/ksym.c
dylandreimerink: I don't think we care about asserting that a missing ksym will result in an error. I think we can leave it out.

linker.go
usamasaqib: Added tests for modules but the `Run tests on arm64` job does not have `bpf_testmod` loaded. I see it loaded into the `vimto` kernels but arm64 seems to run directly on the host.

dylandreimerink: Ah, didn't realize. You can add a call to `requireTestmod` at the start of the test, this will skip it when the testmod is not loaded.

usamasaqib: Thanks, done.

dylandreimerink:
I think we are almost there. There is still the matter of the failing CI, first part of that the formatting of the C files, this should be fixable by running `make container-all` which will spin up the same docker container and toolchain version as used in CI and format your C code for you.

---

Secondly are the failures for "Run tests on previous stable Go", a bit misleading since its also the only place where we pass in all that is needed to run our `TestLibBPFCompat` test. This test attempts to load all the BPF programs of the BPF selftests (with some exceptions).

You can locally repro the failure by running: `CI_KERNEL_SELFTESTS=/usr/src/linux/tools/testing/selftests/bpf vimto -kernel :stable-selftests -- go test -run "TestLibBPFCompat"`

---

For `test_ksyms_weak.bpf.o`, we used to get this error: ` elf_reader_test.go:1115: file /usr/src/linux/tools/testing/selftests/bpf/test_ksyms_weak.bpf.o: load virtual .ksyms section: data section .ksyms: expected runqueues to be *Void, not *btf.Var: not supported` which was tolerated due to the `ErrNotSupported` sentinel. We should now support this, so it continues and errors out with: `elf_reader_test.go:1144: Error during loading: program pass_handler: resolve .ksyms: unable to find btf type for variable bpf_link_fops2 in vmlinux and loaded modules: find target in modules: not found`

In the selftest code its written like:

```
/* typed symbols, default to zero. */
extern const int bpf_link_fops2 __ksym __weak;

...

/* tests non-existent symbols. */
	out__non_existent_typed = (__u64)&bpf_link_fops2;
```

So it seems that libbpf and by extension the selftest code expects us to handle weak typed ksyms without throwing an error. Would be good to tackle this now.

---

In the case of `linked_vars1.bpf.o` and `linked_vars2.bpf.o`, similar story, we used to throw an `ErrNotSupported` due to the typed ksyms in these files. Now that we support typed ksyms, they fail because they contain unlinked symbols.

In the kernel selftests these two object files are linked together with bpftool then turned into a skeleton (header file) which embeds the linked ELF. These unlinked ELF files are not supposed to be loadable so we need to add an exception for them in the `TestLibBPFCompat` test. So add `"linked_vars1", "linked_vars2"` to this case https://github.com/cilium/ebpf/blob/main/elf_reader_test.go#L1105.

dylandreimerink (dismissed):
I am happy with the changes. It couldn't hurt if @ti-mo also takes a look, he usually spots different things.

After that I say we squash the commits to keep the history clean and merge.

usamasaqib:
> I am happy with the changes. It couldn't hurt if @ti-mo also takes a look, he usually spots different things.
> 
> After that I say we squash the commits to keep the history clean and merge.

Thanks for all the feedback! I'll wait for another set of eyes and then merge.

ti-mo:
@usamasaqib Thanks for the patch! I played around with this implementation a bit to get a better understanding of it. I've ended up refactoring `resolveKsymReferences` into `applyUntypedKsymFixups` and `applyTypedKsymFixups` because it was getting a bit hard to follow. The asm.Instruction operations were moved into a new helper, `AssociateBTFID`, this is something we commonly do if mask/shift operations need to be done to encode some bits.

I also added symbol caching to the BTF lookup, since every instruction referencing a var would incur a BTF lookup and (most importantly) a btf.Handle/fd creation. It's still not ideal since btf.Cache doesn't cache handles, maybe this can be improved later on. I think we used to cache handles, but then ripped it out because shared state, and now btf.Cache is no longer shared across lib consumers, so we could potentially add it again.

I also consolidated the tests and added non-weak missing syms. Having typed/untyped subtests looks fairly tidy, I'm happy with the current state.

ti-mo (approved):

dylandreimerink (approved):

Issue #1683, opened 2025-02-14
TestStats: expected a runtime other than 0ns (arm64)


lmb:
### Describe the bug

We get this flake every once in a while:

```
 === FAIL: . TestStats (0.00s)
    info_test.go:362: expected a runtime other than 0ns
```

That number comes directly from the kernel, so something weird is going on. Maybe round to zero or similar?

### How to reproduce

Run the arm64 CI a couple of times.

### Version information

main

Issue #1460, opened 2024-05-10
ci: arm64 tests fail


lmb:
### Describe the bug

Tests on arm64 fail since the new kernel doesn't have the necessary config for kprobe_multi.

```
=== Failed
=== FAIL: link TestKprobeMulti (0.00s)
    kprobe_multi_test.go:16: Feature 'bpf_link_kprobe_multi' isn't supported even though kernel is newer than v5.18

=== FAIL: link TestKprobeMultiErrors (0.00s)
    kprobe_multi_test.go:59: Feature 'bpf_link_kprobe_multi' isn't supported even though kernel is newer than v5.18

=== FAIL: link TestKprobeMultiCookie (0.00s)
    kprobe_multi_test.go:79: Feature 'bpf_link_kprobe_multi' isn't supported even though kernel is newer than v5.18

=== FAIL: link TestKprobeMultiProgramCall (0.00s)
    kprobe_multi_test.go:94: Feature 'bpf_link_kprobe_multi' isn't supported even though kernel is newer than v5.18

=== FAIL: link TestHaveBPFLinkKprobeMulti (0.00s)
    feature.go:30: Feature 'bpf_link_kprobe_multi' isn't supported even though kernel is newer than v5.18
```

### How to reproduce

Run the arm64 ci task.

### Version information

main

lmb:
Seems pretty clear that `CONFIG_FPROBE` is needed: https://elixir.bootlin.com/linux/latest/source/kernel/trace/bpf_trace.c#L3090 Further debugging in https://github.com/cilium/ebpf/pull/1462

alexellis:
Hi, I added everything you asked for and redeployed to all CNCF servers:

https://github.com/cilium/ebpf/pull/1245#issuecomment-2102431551

If you could run grep /boot/config-6.9.10 you should see `CONFIG_FPROBE` etc as per above.

lmb:
Here is the equivalent of the grep you mentioned: https://github.com/cilium/ebpf/actions/runs/9031025998/job/24816484288?pr=1462#step:5:275

Maybe you are missing some of the dependencies? https://cateee.net/lkddb/web-lkddb/FPROBE.html

alexellis:
I've added these dependencies, but I'm not sure if this is supported for 64-bit Arm, because olddefconfig removes them when I run it. This typically means a dependent CONFIG_ is missing or the feature is not available.

dylandreimerink:
[...]

PR #2046, opened 2026-06-30
fix: avoid arm64 flake in TestProgramStats

immanuwell:
arm64 still flakes here.

`#1684` skipped the first arm64 runtime check, but `testStats()` still expects `Runtime != 0` after one run. `3ece212` moved that assertion into the helper, so the same edge case is still there.

This bumps the first stats-enabled run to `Repeat: 1024`. Same test intent, just a bit more signal so `0ns` stops sneaking through.

How to repro
1. Run arm64 CI a few times.
2. Or on an arm64 box run `go test -exec sudo . -run TestProgramStats -count=100`.
3. Watch for `expected runtime to be at least 1ns after first invocation`.

Related: #1683, #1460

### DIFF
--- a/info_test.go
+++ b/info_test.go
@@ -458,6 +458,7 @@ func testStats(tb testing.TB, prog *Program) error {
 	tb.Helper()
 
 	in := internal.EmptyBPFContext
+	const minRepeatForRuntimeStats = 1024
 
 	stats, err := EnableStats(uint32(sys.BPF_STATS_RUN_TIME))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ti-mo:
@immanuwell Thanks for the submission! This inspired me to do some research about why this could be happening. Turns out timers on arm64 have slightly different guarantees than on x86, their accesses can be reordered more aggressively, and they can be 'smoothed' (interpolated) by the hypervisor based on a lower-resolution clock. On the x86 side, TSC is typically sub-nanosecond and dynamic in update frequency according to the CPU's clock speed. By doing this time assertion in tests, we're essentially measuring time in CI, which is never a good idea.

I've raised https://github.com/cilium/ebpf/pull/2047 that contains a similar change with fewer repeats (only 10). This behaves fine after 10 CI runs, please try it locally to see if it still blows up. There was some more arm64 stuff I needed to clean up. Closing this one for now, please flag if the issue persists.

PR #2052, opened 2026-07-01
fix: unmarshal only returned batch lookup entries

### LINKED ISSUE #1080: Batch: only unmarshal as much as needed

immanuwell:
last partial batches still get unmarshaled like they were full size. this narrows batch lookup decode to the count actually returned by the kernel, including the partial final batch case.

the test coverage focuses on the non-zero-copy partial batch path in `TestMapBatchLookupCustomUnmarshaler` and verifies that keys and values are only decoded for the entries returned by the kernel.

repro:
`go test -exec sudo . -run TestMapBatchLookupCustomUnmarshaler`

Fixes #1080
Related: #1741

### DIFF
--- a/map.go
+++ b/map.go
@@ -1301,7 +1301,7 @@ func (m *Map) batchLookup(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, valuesOu
 		return 0, sysErr
 	}
 
-	err = valueBuf.Unmarshal(valuesOut)
+	err = unmarshalBatchSliceOutput(valuesOut, n, int(m.fullValueSize), valueBuf)
 	if err != nil {
 		return 0, err
 	}
@@ -1323,7 +1323,17 @@ func (m *Map) batchLookupPerCPU(cmd sys.Cmd, cursor *MapBatchCursor, keysOut, va
 	}
 
 	if bytesBuf := valueBuf.Bytes(); bytesBuf != nil {
-		err = unmarshalBatchPerCPUValue(valuesOut, count, int(m.valueSize), bytesBuf)
+		if n < count {
+			possibleCPUs, err := PossibleCPU()
+			if err != nil {
+				return 0, err
+			}
+
+			valuesOut = reflect.ValueOf(valuesOut).Slice(0, n*possibleCPUs).Interface()
+			bytesBuf = bytesBuf[:n*int(m.fullValueSize)]
+		}
+
+		err = unmarshalBatchPerCPUValue(valuesOut, n, int(m.valueSize), bytesBuf)
 		if err != nil {
 			return 0, err
 		}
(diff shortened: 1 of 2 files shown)


dylandreimerink:
Thank you for taking the time to look into this. However, when I run your benchmark locally before and after your changes, I see no meaningful improvement in performance.

```
$ benchstat /tmp/before.txt /tmp/after.txt
goos: linux
goarch: amd64
pkg: github.com/cilium/ebpf
                                                   │ /tmp/before.txt │        /tmp/after.txt         │
                                                   │     sec/op      │   sec/op     vs base          │
BatchLookupCustomUnmarshaler/PartialSecondBatch-20       93.49µ ± 1%   96.73µ ± 4%  ~ (p=0.341 n=20)

                                                   │ /tmp/before.txt │           /tmp/after.txt            │
                                                   │      B/op       │     B/op      vs base               │
BatchLookupCustomUnmarshaler/PartialSecondBatch-20      18.19Ki ± 0%   18.24Ki ± 0%  +0.26% (p=0.000 n=20)

                                                   │ /tmp/before.txt │           /tmp/after.txt           │
                                                   │    allocs/op    │ allocs/op   vs base                │
BatchLookupCustomUnmarshaler/PartialSecondBatch-20        13.00 ± 0%   15.00 ± 0%  +15.38% (p=0.000 n=20)
```

In fact, all stats seem to have gotten worse. CPU 93.49 -> 96.73 (this is within margin of measurement, but no improvement). And memory usage increased by 2 allocs per operations and a small amount of bytes.

Looking at the pprof of the benchmark, 99% of the work happens in the syscall itself and the GC. So this is an attempt at optimizing something that isn't a performance issue.

Screenshot from 2026-07-06 11-50-59

I am sorry, but I do not think the added code complexity of this PR is worth it.

immanuwell:
@dylandreimerink I removed the benchmark and narrowed the PR to a correctness fix, WDYT?

dylandreimerink:
@immanuwell Thank you for giving this a shot, but I still do not think the additional code makes sense. There is nothing functionally broken, we just do a bit of extra work which apparently is not a performance issue.

The original issue suggested this change, but the question of if it makes sense should always be checked. I think we can conclude now that the answer to the original issue is "no, we don't care". 

With that I will close this PR, the earlier PR with an attempt and the issue.

Your benchmark and proposed change was critical in making that decision so please do not be disappointed by this result.

PR #2055, opened 2026-07-08
btf: unexport decoder.TypesByName

ti-mo:
It looks like this had been exported by mistake in v0.19.0 and nobody noticed. The same happened to decoder.TypeByID and decoder.TypeID, but those were shadowed by their Spec equivalents.

@dylandreimerink WDYT? This is still inconsistent, not sure about committing to making it a named field instead.

### DIFF
--- a/btf/btf.go
+++ b/btf/btf.go
@@ -2,6 +2,7 @@ package btf
 
 import (
 	"debug/elf"
+	"encoding/binary"
 	"errors"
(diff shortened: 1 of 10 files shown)


dylandreimerink (dismissed):
I don't imagine unexporting is going to be an issue, I don't expect anyone to use these since the decoder itself is also unexported.

PR #2058, opened 2026-07-15
kfunc: allow bpf_ksym_exists on weak symbols from kmods

### LINKED ISSUE #2056: Symbol existence checking of kfunc in module causes verifier error
### Describe the bug

When loading a program that uses `bpf_ksym_exists` on a kfunc that is defined in a kernel module the verifier rejects the program with a `BPF_LD_IMM64 uses reserved fields` error.

This happens because in `fixupKfuncs` we set the Offset field of instructions equal to the file descriptor of the module BTF when resolving a kfunc from a module. https://github.com/cilium/ebpf/blob/075392bc812bf4914118d2a45fb03ff99deaae90/linker.go#L356

This is correct behavior for CALL instructions, where its needed for the verifier to resolve it. But for LD_IMM64 instructions, such as the ones generated by `if (bpf_ksym_exists(...))`, its not allowed. The verifier rejects any load instruction where the offset field is non-zero.

So we should update the `fixupKfuncs` to only open a BTF handle and write its file descriptor to the ins.Offset when its a CALL instruction. And only populate the Constant field for non-call instructions.

### How to reproduce

Create a BPF program like such:
```
#include "common.h"

char __license[] __section("license") = "Dual MIT/GPL";

extern void bpf_testmod_test_mod_kfunc(int) __ksym __weak;

__section("tc") int kfunc_exists() {
	if (bpf_ksym_exists(bpf_testmod_test_mod_kfunc___weak))
		return 1;
	return 2;
}
```

And then try to load it into the kernel.

### Version information

github.com/cilium/ebpf v0.22.0

hown3d:
Only set offset in instruction for kfunc if instructions is `CALL`.

Fixes https://github.com/cilium/ebpf/issues/2056


linker.go
```diff
@@ -353,7 +353,11 @@ fixups:
 		}
 
 		ins.Constant = int64(id)
-		ins.Offset = int16(idx)
+		// kfunc that are part of the kernel create a CALL instruction while kfuncs from kernel module use LD_IMM64.
+		// It's only valid to set offset for CALL instructions
```
dylandreimerink: When a program queries the existence of a ksym is when the LD_IMM64 is created (`bpf_ksym_exists(bpf_xdp_ct_lookup)`). Its just that when its from a kernel module that `idx` != 0 and a non-zero `Offset` is not valid on `LD_IMM64` instructions.


```suggestion
		// It's only valid to set offset for CALL instructions
```

dylandreimerink (changes requested):
Just one very small change, but otherwise I think this is good!

dylandreimerink (dismissed):
Thank you, looks good to me!

ti-mo:
Thanks. I've cleaned up the tests a bit (using bpf_testmod) and moved them to kfunc-kmod.c. When we land https://github.com/cilium/ebpf/pull/2058, we can refactor these a bit to use flavors for weak and non-weak variants, I'll get on that next.

ti-mo:
@cilium/ebpf-go-windows-reviewers eFW tests are broken, might be a good opportunity to dive in and get involved a bit. Looks like the build or the installer is broken.

PR #2059, opened 2026-07-17
feat: add unmarshal util to facilitate unit test

jayl1e:
# Background
- hard to write unit test for bpf map data unmarshal, becuse sysenc.Unmarshal() is in internal package
- integration testing is heavy and relies on linux syscall

# How
- add a wrapper to export internal function

### DIFF
--- a/marshalers.go
+++ b/marshalers.go
@@ -208,3 +208,10 @@ func unmarshalBatchPerCPUValue(slice any, batchLen, elemLength int, buf []byte) 
 	}
 	return nil
 }
+
+// UnmarshalDataFromBytes unmarshals bytes to golang value
+// It is useful for testing
+// The original function is internal, which limits client to write unit test
+func UnmarshalDataFromBytes(data any, buf []byte) error {
+	return sysenc.Unmarshal(data, buf)
+}
--- a/marshalers_test.go
+++ b/marshalers_test.go
@@ -130,3 +130,16 @@ func makeFilledSlice(len int) []uint32 {
 	}
 	return slice
 }
+
+func TestUnmarshalDataFromBytes(t *testing.T) {
+	type Value struct {
+		X int16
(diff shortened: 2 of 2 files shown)


marshalers.go
florianl, line 1: Thanks for the contribution!

I don't think this is the right approach. Anything added to the public API we have to support indefinitely, and adding a function whose stated purpose is solely to help users write tests doesn't add to the functionality cilium/ebpf should provide.

For struct types, `sysenc.Unmarshal` delegates to `encoding/binary.Decode` with the native byte order anyway. Since we require Go 1.21+, `binary.NativeEndian` is available directly:

```
  import "encoding/binary"

  var got MyStruct
  if err := binary.Decode(buf, binary.NativeEndian, &got); err != nil {
      t.Fatal(err)
  }
```

Also blank (`_)` fields are handled as padding by `binary.Decode`, so your test case with `_ int16` would work correctly.

jayl1e: If I just copy the binary.Decode() implementation in sysenc.Unmarshal, unit test may be inconsistent if the internal sysenc.Unmarshal() implementation changed.

Actually I do not think sysenc.Unmarshal is really internal, it is private to Golang code, but it is public to ebpf map. The marshalled/unmarshalled result will be used to interact with ebpf code. We have to check generated bytes if we want to write unit-test (not integrated test).

ti-mo:
Hi, could you provide an example of the kind of tests you want to write for this? Assuming it's just for https://github.com/cilium/cilium/pull/47134, this is a big hammer just for debugging an endianness issue. Same goes for the `unmarshalFromBpfMapBytes` stuff you're proposing to add over on cilium/cilium.

ti-mo:
Closing as per https://github.com/cilium/cilium/pull/47134#issuecomment-5168423473.

Issue #1767, opened 2025-04-30
asm: StoreImm with asm.DWord doesn't return an error


lmb:
### Describe the bug

There is no way to encode a dword store in the ISA. We also don't return an error.

### How to reproduce

Write a program which includes StoreImm(asm.R0, 0, -1, asm.DWord).

### Version information

main

PR #2066, opened 2026-07-24
asm: Allow StoreImm with DWord size

arthurfabre:
Issue #1767 and commit 353dc6a "asm: return InvalidOpCode for StoreImm with DWord size" explicitly rejected StoreImm instructions with a DWord size, claiming there's no way to encode it in the ISA.

This isn't completely true: while there's no way to encode a StoreImm that stores a full 64 bit immediate, it's possible to use a 32 bit immediate, which is then sign extended to 64 bits.

https://docs.kernel.org/bpf/standardization/instruction-set.html#regular-load-and-store-operations says:

> Where ‘<size>’ is one of: B, H, W, or DW

The selftests also exercise this: https://elixir.bootlin.com/linux/v7.1.4/source/tools/testing/selftests/bpf/verifier/bpf_st_mem.c#L4

### DIFF
--- a/asm/load_store.go
+++ b/asm/load_store.go
@@ -313,7 +313,8 @@ func StoreImmOp(size Size) OpCode {
 
 // StoreImm emits `*(size *)(dst + offset) = value`.
 func StoreImm(dst Register, offset int16, value int64, size Size) Instruction {
-	if size == DWord {
+	// Immediate will be stored in 32 bits.
+	if int64(int32(value)) != value {
 		return Instruction{OpCode: InvalidOpCode}
 	}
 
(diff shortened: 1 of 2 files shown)


asm/load_store.go
```diff
@@ -313,10 +313,6 @@ func StoreImmOp(size Size) OpCode {
 
 // StoreImm emits `*(size *)(dst + offset) = value`.
 func StoreImm(dst Register, offset int16, value int64, size Size) Instruction {
-	if size == DWord {
```
dylandreimerink, line 316: I see your point. But then we should return `Instruction{OpCode: InvalidOpCode}` when value requires more than 32 bits. Since that is the core of the issue.

dylandreimerink (changes requested):

asm/load_store.go
arthurfabre: Good idea, done.

dylandreimerink (approved):

arthurfabre:
Thanks for the quick merge! :)

PR #2061, opened 2026-07-20
docs: add kfuncs documentation

k-yamamoto-dev:
## Summary
Add a dedicated kfuncs documentation page.
Related to https://github.com/cilium/ebpf/issues/1152.

## Changes
  - add a **Kfunc availability** section to the Portable eBPF guide
  - explain that ebpf-go resolves kfunc references automatically when loading programs
  - document the `ErrNotSupported` behavior for required kfuncs
  - show how to declare optional kfuncs using `__ksym` and `__weak`
  - add a C example using `bpf_ksym_exists()` with paired acquire and release kfuncs
  - link the existing runtime constants documentation to the new section

PR #2067, opened 2026-07-24
btf: fix exact name filtering

rgo3:
Filter all essential-name candidates without skipping shifted entries. Return ErrNotFound when no exact match remains, preventing AnyTypeByName from indexing an empty result.

Found this while taking a peek at #2061


dylandreimerink (approved):
Nice!

PR #2072, opened 2026-08-04
efw: account for layout change in efW MSI package

ti-mo:
This causes the very descriptive `MSI installation FAILED. Exit code: 1619.` https://github.com/cilium/ebpf/actions/runs/30909207595/job/91991681601


ti-mo:
Looks like this has resolved on its own.

PR #2075, opened 2026-08-06
tracefs: prefer canonical mounts during discovery

alban:
The mountinfo-based discovery introduced by c919805279ee can select a tracefs mount exposed through a Kubernetes pod volume when /var is bind-mounted into the application. For example, it may select
`/host/var/lib/kubelet/pods/<pod>/volumes/kubernetes.io~empty-dir/union-volume/sys/kernel/tracing` instead of the canonical tracefs mount. That mount belongs to another pod and disappears when the pod is deleted, leaving tracepoint attachment with a stale path.

Probe the canonical tracefs paths before consulting mountinfo. Check /sys/kernel/tracing directly, and open /sys/kernel/debug/tracing before checking its filesystem type so the Linux debugfs compatibility automount is instantiated. This is done before mountinfo because the mountinfo parser caches its first result for the lifetime of the process.

If neither canonical path is usable, retain mountinfo discovery for non-canonical full tracefs mounts and debugfs tracing directories, preserving support for container layouts where tracefs is exposed only through a bind mount.

The debugfs tracefs automount was introduced in Linux 4.1 by [f76180bc07ab (tracing: Automatically mount tracefs on debugfs/tracing)](https://github.com/torvalds/linux/commit/f76180bc07ab) and deprecated in Linux 6.18 by [9ba817fb7c6a (tracing: Deprecate auto-mounting tracefs in debugfs)](https://github.com/torvalds/linux/commit/9ba817fb7c6a).

Add regression coverage for preferring a canonical tracefs mount over a transient mount listed earlier in mountinfo.

Fixes: c919805279ee ("tracefs: discover mount via /proc/self/mountinfo")

cc @flyth

---

Marking the PR as draft because it is not yet tested.

### DIFF
--- a/internal/tracefs/kprobe.go
+++ b/internal/tracefs/kprobe.go
@@ -114,38 +114,29 @@ func sanitizeTracefsPath(path ...string) (string, error) {
 //
 // The discovery order is:
 //
-//  1. Any tracefs mount listed in /proc/self/mountinfo (kernel 4.1+).
-//     This works regardless of where the mount sits in the filesystem,
-//     so containers that bind-mount tracefs at a non-canonical path are
-//     supported automatically.
-//  2. A debugfs mount with a tracing/ subdirectory, for older systems
-//     where tracefs has not been lifted out of debugfs.
-//  3. As a final fallback, probe the canonical kernel paths
-//     (/sys/kernel/tracing, /sys/kernel/debug/tracing) directly with
-//     statfs. This catches edge cases where /proc/self/mountinfo is
-//     unavailable or doesn't report the mount.
+//  1. Probe the canonical paths directly. The debugfs compatibility path is
+//     opened before checking its filesystem type to trigger its tracefs
+//     automount when available.
+//  2. Root tracefs mounts listed in /proc/self/mountinfo. This retains support
+//     for containers that expose tracefs only at a non-canonical path.
+//  3. A root debugfs mount whose tracing/ directory can be opened.
+//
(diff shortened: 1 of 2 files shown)


ti-mo:
@alban Let's try improving the mount selection algorithm before we give up on this. Do you have a full example mountinfo from within one of those pods? It could be as simple as preferring the shortest mount path in the container. And how does one end up with a mount layout like this? Also, I'm a bit confused by:

> [...] can select a tracefs mount exposed through a Kubernetes pod volume when /var is bind-mounted into the **application**.

What does application mean in this case? Thanks.

alban:
@ti-mo I've written a minimal reproducer with two workloads:

1. daemonset-azuresecuritylinuxagent contains a shell script that bind-mounts `/host/sys/kernel/tracing` into `/union/sys/kernel/tracing` and unmounts it on termination. `/union` is a Kubernetes empty volume with `mountPropagation=Bidirectional`.
2. daemonset-bpf-tracer is a cilium/ebpf workload (using the main branch without any fixes) that attempts to use tracepoints.

Reproducer: https://gist.github.com/alban/854441b9973f03222a6540e257d5d77d

> What does application mean in this case?

I meant the workload that uses cilium/ebpf.

ti-mo:
@alban Please give this a try: https://github.com/cilium/ebpf/pull/2077.

alban:
> @alban Please give this a try: #2077.

It worked for me!

ti-mo:
Thanks, closing in favor of https://github.com/cilium/ebpf/pull/2077.

PR #2077, opened 2026-08-12
tracefs: prefer unpropagated mounts during mountinfo scanning

ti-mo:
Kubernetes allows mount propagation between pods/containers on a host, which can cause ebpf-go to end up selecting another container's bind-mounted tracefs instead of its own.

Make the selection algorithm a bit smarter to prevent this. Note that propagated mounts can still be selected, but at least there's a clear solution for the user: make sure that every container that needs a tracefs mounts its own.

AIL:3

Fixes: #2075


dylandreimerink (dismissed):

internal/tracefs/kprobe.go
```diff
@@ -174,6 +174,51 @@ func findTracefsInEntries(entries []mountinfo.Entry) string {
...
+	})
+	return candidates
+}
+
+// propagationClass ranks a mount by how likely it is to remain valid for the
+// lifetime of the process, based on its mount propagation optional fields.
+// Lower is better.
+func propagationClass(e mountinfo.Entry) int {
+	switch {
+	case len(e.OptionalFields) == 0:
```
alban:
```suggestion
	case !e.HasOptionalField("shared") && !e.HasOptionalField("master"):
```

If it has the optional field "unbindable", it will not participate in mount propagation.

In practice, nobody would use "unbindable" on tracefs, so it does not matter much.

florianl (dismissed):

PR #2082, opened 2026-08-14
btf: parse BTF types in a single pass during decoder initialization

### LINKED ISSUE #2081: btf: avoid scanning the BTF type buffer twice when building a decoder
#### Problem 
When initializing BTF specifications in `newDecoder()` ([`btf/unmarshal.go`](file:///d:/s/ebpf/btf/unmarshal.go#L71-L133)), the decoder iterates over `allBtfTypeOffsets(raw, bo, &header)` **twice** across the entire raw BTF byte slice:

1. **Pass 1 (Lines 71–85):** Scans the entire raw BTF buffer solely to count `numTypes`, `numDeclTags`, and `numNamedTypes` to size internal maps and slices.
2. **Pass 2 (Lines 103–132):** Scans the exact same raw BTF buffer a second time to unmarshal type offsets

For large BTF payloads (which spans several megabytes), unmarshaling type headers and calculating data lengths twice adds unnecessary CPU overhead to BTF spec loading 

####  My Suggestion
Because every BTF type header is at least 12 bytes (`btfTypeSize`), the maximum possible number of types in a raw buffer of length `len(raw)` cannot exceed `len(raw) / 12`.

We can estimate the maximum slice capacity upfront using `maxTypes := len(raw) / 12` and perform all unmarshaling in a **single pass**, eliminating the initial scan loop entirely.

**Note on Capacity Trade-off:** Using len(raw) / 12 might reserve slightly more memory than needed for large BTF types. But, guessing the size beforehand  removes the need for a first scanning pass, making the decoding process noticeably faster overall.
#### Code changes

	// Each BTF type header is at least 12 bytes, so maximum possible types is len(raw) / 12
	maxTypes := len(raw) / 12
	if maxTypes < 16 {
		maxTypes = 16
	}

	offsets := make([]int, 0, maxTypes)
	declTags := make(map[TypeID][]TypeID)
	namedTypes := newFuzzyStringIndex(0)

#### Benchmark Results ( ` func BenchmarkParseVmlinux`)

run ```go test -run=^$ -bench BenchmarkParseVmlinux -benchmem ./btf```

| Metric | Before (Two-Pass) | After (Single-Pass) |
| :--- | :--- | :--- |
| **Execution Time** | `9.11 ms/op` (`9,106,499 ns`) | `8.23 ms/op` (`8,227,959 ns`) |
| **Throughput (1s)** | `132 ops/sec` | `145 ops/sec` |
| **Memory Allocated** | `1,315,123 B/op` | `4,304,885 B/op` |
| **Allocations/op** | `246 allocs/op` | `282 allocs/op` |

DARSHANR007:
Fixes #2081

`newDecoder()` currently scans the raw BTF buffer twice: once to count the number of types and a second time to parse them into memory.

Since every BTF type header is at least 12 bytes (`btfTypeSize`), we can estimate the maximum possible number of types upfront (`len(raw) / btfTypeSize`) and parse all types in a single pass instead.

### DIFF
--- a/btf/unmarshal.go
+++ b/btf/unmarshal.go
@@ -66,33 +66,10 @@ func newDecoder(raw []byte, bo binary.ByteOrder, strings *stringTable, base *dec
 	}
 
 	var header btfType
-	var numTypes, numDeclTags, numNamedTypes int
 
-	for _, err := range allBtfTypeOffsets(raw, bo, &header) {
-		if err != nil {
-			return nil, err
-		}
-
-		numTypes++
-
-		if header.Kind() == kindDeclTag {
-			numDeclTags++
-		}
-
-		if header.NameOff != 0 {
-			numNamedTypes++
-		}
-	}
-
-	if firstTypeID == 0 {
-		// Allocate an extra slot for Void so we don't have to deal with
-		// constant off by one issues.
-		numTypes++
-	}
-
-	offsets := make([]int, 0, numTypes)
-	declTags := make(map[TypeID][]TypeID, numDeclTags)
-	namedTypes := newFuzzyStringIndex(numNamedTypes)
+	offsets := make([]int, 0, len(raw)/btfTypeSize)
+	declTags := make(map[TypeID][]TypeID)
+	namedTypes := newFuzzyStringIndex(0)
 
 	if firstTypeID == 0 {
 		// Add a sentinel for Void.
@@ -100,7 +77,11 @@ func newDecoder(raw []byte, bo binary.ByteOrder, strings *stringTable, base *dec
 	}
 
 	id := firstTypeID + TypeID(len(offsets))
-	for offset := range allBtfTypeOffsets(raw, bo, &header) {
+	for offset, err := range allBtfTypeOffsets(raw, bo, &header) {
+		if err != nil {
+			return nil, err
+		}
+
 		if id < firstTypeID {
 			return nil, fmt.Errorf("no more type IDs")
 		}


dylandreimerink:
I must say I am not super convinced. I ran the benchmark myself as well, getting the following result:

```
goos: linux
goarch: amd64
                │ before.txt  │             after.txt              │
ParseVmlinux-20   7.659m ± 1%   7.328m ± 4%  -4.33% (p=0.024 n=20)

                │  before.txt  │               after.txt               │
                │     B/op     │     B/op      vs base                 │
ParseVmlinux-20   1.254Mi ± 0%   4.105Mi ± 0%  +227.33% (p=0.000 n=20)

                │ before.txt │             after.txt              │
                │ allocs/op  │ allocs/op   vs base                │
ParseVmlinux-20   246.0 ± 0%   282.0 ± 0%  +14.63% (p=0.000 n=20)
```

So we speed up 4% on average, in exchange for a 227% increase in temporary memory usage. Likely due to now having to resize the fuzzy string index and decl tags index multiple times.

This method also grossly overestimates the amount of offsets, leading to `offsets` slice capacity being ~50% larger than what is actually filled (on the vmlinux), which is about 900KiB which gets retained for as long as the `btf.Spec` stays alive, which can be long when a `btf.Cache` is used.

I don't think this tradeoff is worth.

ti-mo:
Yep, quadrupling allocs for a low-single digit % gain in wall clock time is typically not what our users ask for. On the contrary, memory consumption gets much more heavily scrutinized as it sticks around for longer in caches and throughout GC cycles.

PR #2086, opened 2026-08-17
btf: fix essential-name index for leading triple-underscore names

### LINKED ISSUE #2085: fexit ___pskb_trim not supported
### Describe the bug

When I tried to trace all kernel functions that have skb argument with [bpfsnoop](https://github.com/bpfsnoop/bpfsnoop) on kernel 7.2.0-rc6-gd58d2345016c-dirty, it failed with error:

```
# ./bpfsnoop -k '*:skb:struct sk_buff *'
2026/08/17 14:52:35 bpfsnoop is tracing 1717 kernel functions/tracepoints, this may take a while
2026/08/17 14:53:11 Failed to trace: failed to trace targets: failed to create bpf collection for tracing ___pskb_trim: program bpfsnoop_fn: attach Tracing/TraceFExit: fexit ___pskb_trim not supported
```

After reading the code, it seems it is intended to skip part of the symbol: https://github.com/cilium/ebpf/blob/e3a865be0e4be4a0ec78979629c2f35f3ec37c18/btf/unmarshal.go#L121-L126

Can we find a way to add tracing `___` kernel functions support?

Here's the `___` kernel functions list:

```bash
# ./bpfsnoop -k '*___*' --show-func-proto
Kernel functions: (total 22)
void ____fput(struct callback_head *work); [traceable]
void ____ip_mc_inc_group(struct in_device *in_dev, __be32 addr, unsigned int mode, gfp_t gfp);
int ____netdev_has_upper_dev(struct net_device *upper_dev, struct netdev_nested_priv *priv);
int ____sys_recvmsg(struct socket *sock, struct msghdr *msg_sys, struct user_msghdr *msg, struct sockaddr *uaddr, unsigned int flags, int nosec);
int ____sys_sendmsg(struct socket *sock, struct msghdr *msg_sys, unsigned int flags, struct used_address *used_address, unsigned int allowed_msghdr_flags);
void ___d_drop(struct dentry *dentry);
void ___free_pages(struct page *page, unsigned int order, fpi_t fpi_flags);
void *___kmalloc_large_node(size_t size, gfp_t flags, int node);
void ___migrate_enable();
struct neighbour *___neigh_create(struct neigh_table *tbl, const void *pkey, struct net_device *dev, u32 flags, bool exempt_from_gc, bool want_ref);
void ___p4d_free_tlb(struct mmu_gather *tlb, p4d_t *p4d);
struct pcpu_freelist_node *___pcpu_freelist_pop(struct pcpu_freelist *s);
void ___perf_sw_event(u32 event_id, u64 nr, struct pt_regs *regs, u64 addr);
void ___pmd_free_tlb(struct mmu_gather *tlb, pmd_t *pmd);
int ___pskb_trim(struct sk_buff *skb, unsigned int len);
void ___pte_free_tlb(struct mmu_gather *tlb, struct page *pte);
void ___pud_free_tlb(struct mmu_gather *tlb, pud_t *pud);
int ___ratelimit(struct ratelimit_state *rs, const char *func);
void *___slab_alloc(struct kmem_cache *s, gfp_t gfpflags, int node, const struct slab_alloc_context *ac);
int ___sys_recvmsg(struct socket *sock, struct user_msghdr *msg, struct msghdr *msg_sys, unsigned int flags, int nosec);
int ___sys_sendmsg(struct socket *sock, struct user_msghdr *msg, struct msghdr *msg_sys, unsigned int flags, struct used_address *used_address, unsigned int allowed_msghdr_flags);
struct rq *___task_rq_lock(struct task_struct *p, struct rq_flags *rf);
```

### How to reproduce

```
git clone --recurse-submodules https://github.com/bpfsnoop/bpfsnoop.git
cd bpfsnoop
make -j$(nproc)
./bpfsnoop -k '*:skb:struct sk_buff *'
```

### Version information

github.com/cilium/ebpf v0.21.0 => github.com/bpfsnoop/ebpf v0.21.0-next

pujitha24:
Motivation:
Spec.TypeByName/AnyTypesByName could not find BTF types whose name
starts with a "___" flavour delimiter, such as the kernel function
___pskb_trim. This broke attaching fexit/fentry programs to such
functions via link package tracing helpers, which resolve the target
BTF type through spec.TypeByName.

BTF "flavours" let a type name like sk_buff___read_mostly be looked
up by its essential name sk_buff: newEssentialName strips everything
after the last "___", but only when that delimiter does not start the
name (guarded by `lastIdx > 0`). The index built while decoding BTF in
newDecoder (btf/unmarshal.go) did not have that guard: for a name like
___pskb_trim, bytes.LastIndex returns 0, so the old code stripped the
whole name down to an empty essential name and indexed it under that
empty hash. A later lookup for the literal name ___pskb_trim computes
its (unstripped) essential name via newEssentialName and hashes to a
different bucket, so decoder.typesByName never finds the entry and
TypeByName/AnyTypesByName return ErrNotFound.

Approach:
Add the same `i > 0` guard to the index-building code in
btf/unmarshal.go so the essential name it computes always matches
newEssentialName exactly.

Validation:
Added TestAnyTypesByNameLeadingUnderscores in btf/btf_test.go, which
builds a synthetic spec containing only a type named ___pskb_trim and
looks it up by that exact name. Confirmed with `git stash` that the
test fails with ErrNotFound before this fix and passes after it:

  go test ./btf/ -run TestAnyTypesByNameLeadingUnderscores -v

Also ran, all passing (two pre-existing failures in TestHandleIterator
and TestParseModuleSplitSpec are unrelated environment limitations of
this sandbox, which has no live kernel BTF to query on darwin, and
fail identically with or without this change):

  go build ./...
  go test ./btf/...
  go vet ./btf/...
  gofmt -l btf/unmarshal.go btf/btf_test.go
  golangci-lint run ./btf/...

This fix is verified against a targeted unit reproduction of the
essential-name mismatch, not against a live kernel fexit attach; the
reporter's original bpfsnoop repro was not re-run here.

Report: https://github.com/cilium/ebpf/issues/2085
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #2085


dylandreimerink (approved):
This makes sense, the same logic is already in `newEssentialName`, just not here.

PR #2093, opened 2026-08-24
perf: improve error handling

florianl:
Skip the `reflectlite.TypeOf(target).Comparable()` call that `errors.Is()` makes unconditionally on entry. As `forwardReader.Read` and `reverseReader.Read` return the `io.*` errors directly and never wrap them, avoid the costs of `reflectlite.TypeOf()`.

```
$ benchstat main.txt new.txt 
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
                 │  main.txt   │           new.txt           │
ReadRecordEOR-20   8.967n ± 0%   6.527n ± 0%  -27.21% (n=50)

                 │  main.txt  │            new.txt             │
                 │    B/op    │    B/op     vs base            │
ReadRecordEOR-20   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=50) ¹
¹ all samples are equal

                 │  main.txt  │            new.txt             │
                 │ allocs/op  │ allocs/op   vs base            │
ReadRecordEOR-20   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=50) ¹
¹ all samples are equal
```


ti-mo (commented):
Thanks! I'm not super convinced the benchmark has long-term value, since it's rather specific and there's nothing to alert us about regressions. Any other ideas?

(also please remove the merge commit)

florianl (commented):
Happy to remove the benchmark - it was just intended to demonstrate the improvement from this change, not to serve as a regression check. I'll also remove the merge commit.

dylandreimerink (approved):

PR #2095, opened 2026-08-30
elf: apply implicit addends to object relocations

### LINKED ISSUE #2083: STT_OBJECT Implicit Addend Handling Difference
## Summary

Cilium/ebpf, Aya, and libbpf produce different results for an `R_BPF_64_64` relocation against an `STT_OBJECT` symbol with a non-zero implicit addend, which is a ELF compatibility issue.

## Tested Environment

The differential was reproduced on :

| Loader | Commit |
| --- | --- |
| Cilium/ebpf | `e3a865be0e4be4a0ec78979629c2f35f3ec37c18` |
| Aya | `15a0de1b363e51d55d0fca4245df54a61e8d5521` |
| libbpf | `f7081a6baf3f54949aacb8c2fc11bb30783b83e9` |

The input is a minimal hand-built little-endian BPF ELF containing one socket-filter program, one `SHT_REL` relocation, and an eight-byte `.data` section.

The program performs the equivalent of:

```c
return *(uint32_t *)(data_pointer);
```

Its four BPF instructions are:

```text
0: LDIMM64 r1, 4       R_BPF_64_64 relocation against STT_OBJECT "first"
1: LDIMM64 second slot
2: LDXW r0, [r1 + 0]
3: EXIT
```

The relocation symbol `first` has `st_value=0`, while the original `LDIMM64` immediate contains the implicit addend `A=4`. The `.data` contents are:

```text
offset 0: uint32_t 11
offset 4: uint32_t 22
```

Correct `S+A` processing therefore makes the program read offset 4 and return 22. Ignoring `A` makes it read offset 0 and return 11.

## Observed Behavior

Results:

| Loader | Resolved Offset | Return Value |
| --- | ---: | ---: |
| Cilium/ebpf | 0 | 11 |
| Aya | 4 | 22 |
| libbpf | 4 | 22 |

All three resulting programs were accepted by the kernel verifier and `BPF_PROG_TEST_RUN`.

## Root Cause

BPF uses `SHT_REL`, so the addend is stored in the original `LDIMM64` immediate. The documented calculation for `R_BPF_64_64` is:

```text
result = S + A
```

Cilium/ebpf instead selects either the symbol value or the instruction addend based on symbol type:

```go
case elf.STT_SECTION:
    offset = uint32(uint64(ins.Constant)) // A
case elf.STT_OBJECT:
    offset = uint32(rel.Value)            // S
```

Consequently, a non-zero addend is ignored for `STT_OBJECT` relocations.

Aya preserves both components:

```rust
instructions[ins_index + 1].imm =
    instructions[ins_index].imm + sym.address as i32;
```

libbpf does the same:

```c
insn[1].imm = insn[0].imm + relo->sym_off;
```

Cilium's behavior matches common LLVM output, where global `STT_OBJECT` symbols normally have `A=0`, while static variables use `STT_SECTION` with the offset stored in `A`. The difference appears with valid but uncommon `STT_OBJECT` relocations containing a non-zero addend.

## Suggested Resolution

Compute the final offset as checked `S+A`, followed by 32-bit overflow and section-boundary validation. If non-zero addends for `STT_OBJECT` are intentionally unsupported, reject them explicitly instead of silently resolving a different offset.

## References

- [Linux BPF LLVM Relocations](https://docs.kernel.org/next/bpf/llvm_reloc.html)
- [LLVM LLD BPF implicit addend implementation](https://reviews.llvm.org/D101336)

raghav2005:
## Summary

- Resolve STT_OBJECT data-section relocations as S + A, preserving the SHT_REL implicit addend from the original LDIMM64 instruction.
- Reject signed arithmetic overflow or underflow, offsets that cannot fit the pseudo-map-value encoding, and offsets outside the target section.
- Add regression coverage for the minimal issue reproducer, the GCC values[8192] case, and arithmetic edge cases.

## Testing

- `go test . -run '^TestRelocateDataSectionObjectAddend$' -count=1`
- `CGO_ENABLED=1 go test -race . -run '^TestRelocateDataSectionObjectAddend$' -count=1`
- `go build ./...`
- `go vet ./...`
- `go tool staticcheck ./...`
- `golangci-lint run --timeout=5m`
- `./scripts/go-fix.sh`
- Cross-build and test compilation for Darwin, Android/arm64, Linux/armv6, and js/wasm

Fixes #2083


ti-mo:
https://github.com/cilium/ebpf#generative-ai-policy

Issue #2089, opened 2026-08-21
Deprecate `features.HaveProgramHelper`


dylandreimerink:
This issue came to my attention https://github.com/cilium/cilium/issues/48016. Looking into the details, this fundamentally originates in the way we do helper function probing. In the latest kernel there are quite a few helper functions that have special casing https://elixir.bootlin.com/linux/v7.2/source/kernel/bpf/verifier.c#L10274

I think we should deprecate this functionality, and remove it in 2 releases. Encourage people to use CO-RE instead, or write more targeted probes themself.

eraycep:
Hi, I’d like to work on this as my first contribution. Am I correct that the scope for this PR is to add a Go deprecation notice to features.HaveProgramHelper and update the feature-detection documentation, while retaining the implementation and tests until the planned removal release? Is there a preferred wording for the CO-RE/targeted-probe recommendation?

dylandreimerink:
@ti-mo and I have talked offline. Consensus seems to be that instead of removing probing functionality like this completely, we want to change it so its BTF based. Most feature probes such as those for program type, map type, melper functions ect. can be determined from vmlinux BTF just like how CO-RE code does it. We can't really modify the internals of the current functions to do that, so we will likely need to make a new API for that first.

We have not gone into the details of what that would look like.

eraycep:
Thanks for the response. A new BTF-based feature-detection API sounds broader than the original deprecation-only scope, especially since its design is still open. I’m interested in contributing, but as this would be my first contribution to ebpf-go, would it be useful for me to investigate the existing BTF/CO-RE support and propose an initial API design, or would you prefer to settle the design first?

brycekahle:
👋 While we prefer BTF/CO-RE where possible, we do still support customers where that isn't available. It'd be great if a non-BTF way was still available, but I realize the maintenance burden that may carry.

dylandreimerink:
I do not think doing that in the library itself is the way to go. To to it properly you essentially need to write a individual/custom probe for the thing you want to test, specifically for the helper functions. I highly recommend writing CO-RE style programs or if you need to probe for certain features in userspace to write probes yourself. The library provides all the public API needed to writes probes outside of the library. Which is what cilium/cilium also does for example https://github.com/cilium/cilium/tree/main/pkg/datapath/linux/probes

At this point it simply makes more sense to let users maintain their own probes than to try and maintain them in the library.

PR #2099, opened 2026-09-07
features: fix FnSetRetval probe failure on kernel 7.2+

maxweisspoker:
features: fix FnSetRetval probe failure on kernel 7.2+

Kernel 7.2 (commit "bpf: Add validation for bpf_set_retval argument" by Xu Kuohai) added explicit type checking for the first argument to bpf_set_retval in the BPF verifier. The helper now requires R1 to be SCALAR_VALUE in the range [-MAX_ERRNO, 0]. Previously, the argument type was ARG_ANYTHING with no validation.

The haveProgramHelper probe constructs a minimal BPF program that calls the helper as its first instruction, meaning R1 still holds PTR_TO_CTX from program entry. On kernel 7.2+, the verifier rejects this with EINVAL and the log message "R1 is not a scalar". This falls through the existing EINVAL handling (which only recognizes "invalid func" and "unknown func" as not-supported signals), causing features.HaveProgramHelper to return an unexpected error. In Cilium, this surfaces as a fatal startup failure when probing FnSetRetval for CGroupSock programs, blocking any configuration that depends on the probe (e.g. socketLB.hostNamespaceOnly=true for Kata Containers / gVisor compatibility).

This commit makes two changes: (1) Pre-load R1 with an immediate value of 0 before calling FnSetRetval in the probe program. The value 0 satisfies the new [-MAX_ERRNO, 0] range check on 7.2+ and is accepted as ARG_ANYTHING on older kernels, so the probe works on both. (2) In the EINVAL error handler, after exhausting the known "helper not found" patterns (invalid func, unknown func, wrong program type), check whether the helper's tag appears in the verifier log. If it does, the verifier reached and recognized the helper call — the EINVAL is from argument validation, not from the helper being unavailable. Treat this like EACCES (helper available, arguments wrong).

Refs: https://lore.kernel.org/bpf/20260605140243.664590-3-xukuohai@huaweicloud.com/
Refs: https://github.com/cilium/cilium/issues/48016

This is an AI contribution. The model that performed the change was opus-4.6, but I (the lowly human) reviewed and tested the change.

### DIFF
--- a/features/prog.go
+++ b/features/prog.go
@@ -246,14 +246,33 @@ func haveProgramHelper(pt ebpf.ProgramType, helper asm.BuiltinFunc) error {
 		return err
 	}
 
-	spec := &ebpf.ProgramSpec{
-		Type: pt,
-		Instructions: asm.Instructions{
+	insns := asm.Instructions{
+		helper.Call(),
+		asm.LoadImm(asm.R0, 0, asm.DWord),
+		asm.Return(),
+	}
+
+	// bpf_set_retval requires R1 to be a scalar value, not a context
+	// pointer. Since kernel 7.2 (commit "bpf: Add validation for
+	// bpf_set_retval argument"), the verifier rejects non-scalar R1
+	// with EINVAL. Pre-load R1 with an immediate so the probe tests
+	// helper availability rather than failing argument validation.
+	// The value 0 is within the allowed range [-MAX_ERRNO, 0] on 7.2+,
+	// and accepted as ARG_ANYTHING on older kernels.
+	switch helper {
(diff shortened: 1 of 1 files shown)


ti-mo:
Thank you, but before throwing your tokens at a problem description, please check if there's been any discussion around the topic: https://github.com/cilium/ebpf/pull/2089. The change you propose has the potential for breaking other probes which we can't realistically test for, nor do we want to maintain this sort of thing going forward.

maxweisspoker:
For the record, I _did_ do that, and that thread looked like it wasn't going anywhere...

PR #2106, opened 2026-09-16
elf: fix FUNC_INFO misassignment across .text alignment padding

### LINKED ISSUE #2103: `FUNC_INFO` misassignment across `.text` padding changes verifier behavior
## Summary

The valid Clang compiled `prog.o` produces opposite verifier results:

- libbpf preserves `victim(void *)`, the verifier rejects the scalar argument.
- cilium/ebpf incorrectly attaches `declared(long)` metadata to the `victim` body, 
the verifier accepts the argument and returns a program FD.

The cause is a `.text` alignment gap. cilium/ebpf advances the BTF.ext metadata
cursor only across function bodies, while the recorded `FUNC_INFO` offsets
include the gap. This shifts the preceding `declared(long)` prototype onto
`victim(void *)`.

## Reproducer

`prog.c`:

```c
// SPDX-License-Identifier: GPL-2.0
#include "vmlinux.h"
#include "bpf_helpers.h"

char LICENSE[] SEC("license") = "GPL";

__attribute__((noinline))
int alpha(int x)
{
	return 1;
}

__attribute__((noinline, aligned(32)))
int declared(long x)
{
	return 2;
}

__attribute__((noinline))
int victim(void *p)
{
	return (long)p;
}

SEC("xdp")
int entry(struct xdp_md *ctx)
{
	return victim((void *)(long)(ctx->rx_queue_index & 3))
		? XDP_PASS : XDP_DROP;
}
```

Compile with the following command:

```bash
clang -g -O2 -target bpf -mcpu=v3 -mlittle-endian \
  -D__TARGET_ARCH_x86 -I /path/to/headers -I /path/to/libbpf/src \
  -c prog.c -o prog.o
```

The final ELF has one XDP program and the `.text` layout:

```text
alpha       [ 0, 16)
padding     [16, 32)
declared    [32, 48)
victim      [48, 64)
```

 BTF.ext `FUNC_INFO` offsets are correctly encoded as raw instruction
offsets `0`, `4`, and `6`.

This makes the result easy to distinguish. `entry` derives a scalar from
`ctx->rx_queue_index` and passes it to `victim`:

```text
correct ELF metadata:       scalar -> victim(void *)  => reject
cilium/ebpf metadata shift: scalar -> victim(long)    => accept
```

## Results

Using the same `prog.o`:

| Loader | Result |
|---|---|
| libbpf | verifier rejects with `EINVAL` |
| cilium/ebpf | verifier accepts; program FD returned |

The libbpf verifier log explains the expected rejection:

```text
Caller passes invalid args into func#1 ('victim')
```

Both results reproduce with Linux 7.2.6 JIT enabled and disabled.

## Root Cause

The faulty path is in
[`loadFunctions()`](https://github.com/cilium/ebpf/blob/60e81073fdc61c98149c44896135441cbae583a4/elf_reader.go#L442-L507)
and
[`take()` / `assignMetadata()`](https://github.com/cilium/ebpf/blob/60e81073fdc61c98149c44896135441cbae583a4/elf_reader.go#L510-L533).

`parseFuncInfoRecords()` correctly converts the BTF.ext offsets to `0`, `4`,
and `6`. In `loadFunctions()`, however, line 498 passes a compact `raw`
counter to `assignMetadata()`. The counter advances only across decoded symbol
instructions and does not include the two-instruction padding gap, so function
starts are visited as `0`, `2`, and `4`.

`take()` at lines 512-523 checks only the first queued metadata record.
Consequently:

```text
declared: loader offset 2 != queued declared FUNC_INFO offset 4
victim:   loader offset 4 == queued declared FUNC_INFO offset 4
```

The `declared(long)` metadata is therefore attached to `victim`, while the real
`victim(void *)` record remains unused. The flattened program is then submitted
to the kernel with the wrong function prototype.

libbpf avoids this misassociation by retaining each function's original
section-relative offset.
[`bpf_object__add_programs()`](https://github.com/libbpf/libbpf/blob/f7081a6baf3f54949aacb8c2fc11bb30783b83e9/src/libbpf.c#L920-L968)
reads each function's absolute ELF `st_value`, and
[`bpf_object__init_prog()`](https://github.com/libbpf/libbpf/blob/f7081a6baf3f54949aacb8c2fc11bb30783b83e9/src/libbpf.c#L843-L861)
stores it as `sec_insn_off`. Padding is omitted when subprogram instructions
are appended, but
[`adjust_prog_btf_ext_info()`](https://github.com/libbpf/libbpf/blob/f7081a6baf3f54949aacb8c2fc11bb30783b83e9/src/libbpf.c#L6525-L6585)
selects `FUNC_INFO` using the original absolute range and rebases it to the
flattened position. Thus `victim` retains its own `void *` prototype, and the
verifier rejects the scalar argument.

## Impact

This is verifier metadata corruption. cilium/ebpf changes the function
prototype supplied to the kernel and obtains a program FD for an ELF that
libbpf correctly rejects.

## Suggested Fix

Associate metadata using each instruction's absolute section offset
(`sym.Value + offset within the symbol`), rather than a compact counter over
symbol bodies. Also reject stale or unconsumed BTF.ext records instead of
silently allowing them to shift onto later functions.

Add a regression test containing `.text` alignment padding and adjacent
functions with incompatible prototypes.

## Environment

- cilium/ebpf: `60e81073fdc61c98149c44896135441cbae583a4`
- libbpf: `f7081a6baf3f54949aacb8c2fc11bb30783b83e9`
- Linux: v7.2.6 (`500df175a7f9e6bc1a9c328590ca5150f84f9ff0`)

pujitha24:
Motivation:

loadFunctions() matches BTF.ext ExtInfo records (FUNC_INFO, LINE_INFO,
CO-RE relocations) against decoded instructions using a running counter
that only advanced across instructions it actually decoded. Any gap
between one symbol's end and the next symbol's start in the .text
section, e.g. compiler alignment padding from
__attribute__((aligned(N))), desynchronized that counter from the raw
section-relative instruction offsets the ExtInfo records actually
point at. Because ExtInfo consumption only checks the front of a
sorted-by-offset queue, that desync permanently shifted every FUNC_INFO
record after the gap onto the wrong function.

In the reported case, this attached a `declared(long)` prototype to a
`victim(void *)` function body, so the loader submitted the program
with the wrong function signature and the kernel verifier accepted an
argument that should have been rejected. libbpf doesn't have this bug
because it looks up FUNC_INFO by each function's absolute ELF section
offset.

Approach:

Compute each instruction's raw section-relative offset directly from
its symbol's absolute byte offset (sym.Value, already used earlier in
the same loop for relocation lookups) plus its offset within that
symbol, instead of accumulating a counter that skips gaps. This can't
drift regardless of any padding between symbols. Also reject symbols
whose offset isn't instruction-size aligned instead of silently
truncating it via integer division, matching the alignment checks
already used for ExtInfo offsets in btf/ext_info.go.

Validation:

Added TestLoadFunctionsPadding, which hand-builds an in-memory ELF
.text section (pure Go, no clang/BPF toolchain involved) mirroring the
issue's exact layout: three one-instruction-pair functions with a
16-byte padding gap before the second one, and FUNC_INFO records at
the correct raw offsets. It calls the unexported loadFunctions
directly and asserts each function keeps its own metadata.

  go test -run TestLoadFunctionsPadding -v .

This test fails on the pre-fix code (the padding-adjacent function's
own FUNC_INFO record is silently stolen by its neighbor, leaving it
with nil metadata) and passes after the fix.

Also ran, all passing/clean:

  go build ./...
  go vet ./...
  golangci-lint run ./...   (one pre-existing, unrelated finding:
                              internal/sys/signals.go sigsetAdd unused,
                              present identically without this change)
  GOOS=darwin go build ./... && go test -c -o /dev/null ./...
  GOOS=android GOARCH=arm64 CGO_ENABLED=0 go build/test -c ./...
  GOOS=linux GOARCH=arm GOARM=6 go build/test -c ./...
  GOOS=js GOARCH=wasm go build/test -c ./...

Could not run: `go tool staticcheck ./...` errors locally with an
export-data version mismatch (this sandbox's go1.27.1 toolchain vs.
the repo's CI-pinned ~1.26); this is an environment/toolchain issue,
not something this change can trip.

The full `go test .` suite has many pre-existing failures in this
sandbox unrelated to this change (no /sys/fs/bpf, no privileged BTF
syscalls, no real kernel -- this repo's own docs say most tests need
`sudo` or a Linux VM via vimto, neither available here). Diffing the
full sorted list of `--- FAIL` lines with and without this change
applied shows it is byte-for-byte identical, so this change introduces
no new failures anywhere in the package.

Report: https://github.com/cilium/ebpf/issues/2103
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #2103


ti-mo:
https://github.com/cilium/ebpf#generative-ai-policy

PR #2107, opened 2026-09-17
Reimplement program type, helper function and map type probing

### LINKED ISSUE #2089: Deprecate `features.HaveProgramHelper`

dylandreimerink:
So far we have mostly been doing feature probing by creating small programs of a given type, that use features we want to test. If we are able to load them without error it means we have support, and if they don't we try to figure out from verifier logs why not.

This method has a number of disadvantages, specifically for helper functions. In recent kernels the verifier has started to assert that arguments for certain helpers must be "valid" which caused our probes to fail. In addition, we used to check verifier logs to refine the error cause, changes in the verifier log format also broke this part.

Another disadvantage is that loading programs and making maps require permissions that may not be available, because we are running in a BPF token enabled environment or because LSM hooks are in place, certain capabilities are not given or because LSM hooks are in place.

This PR therefor changes our probing functions to now inspect the kernel BTF to see if it knows about certain program types, helper functions and map types. This is similar to how CO-RE queries would do it when writing CO-RE eBPF programs.

The existing functions have been re-implemented, a breaking change. They now require a `*btf.Cache`, which can be used to amortize BTF parsing cost across multiple probes and/or collection loads. The `HaveProgramHelper` semantics has also changed to now check if the kernel has support, but does not check if a specific program type can use that helper. Lastly the `HaveMapFlag` probe was removed since the old implementation was poorly maintained and of very limited use.

If users of the library need capabilities that have been removed, we encourage them to implement their own probes, which is completely doable with the external API that exists.

Fixes: #2089


dylandreimerink:
@ti-mo what do you think of this? Its not passing CI because `TestProgramVerifierOutputOnError` is broken and vmlinux not yet up to date.

We got 3 thing going on at the same time. This probes change, the `TestProgramVerifierOutputOnError` and vmlinux. We can merge fixes separately or I can pull all fixes into this PR.

ti-mo:
> pull all fixes into this PR

That's going to get messy, at least for the changes in this particular PR. We can batch the verifier output parsing and the vmlinux bump, they're kinda related (upgrade pain) anyway.

I'm working on some changes to ci-kernels that will sidestep this problem down the line and will make the tests more reproducible.

tl;dr: generate test matrix from ci-kernels/versions.json using a workflow, consume that as e.g. `uses: cilium/ci-kernels/matrix@7.3-rc4` and make dependabot open PRs here when versions.json is updated+tagged so we can address these issues in lockstep. The floating mainline/stable/longterm pins on the OCI images are not ideal to say the least.

internal/find/find.go
```diff
@@ -0,0 +1,84 @@
...
+	return kernelSpec, nil, err
+}
+
+// findTargetInModule attempts to find a named type in any loaded module.
+//
+// base must contain the kernel's types and is used to parse kmod BTF. Modules
+// are searched in the order they were loaded.
+//
+// Returns btf.ErrNotFound if the target can't be found in any module.
+func findTargetInModule[T btf.Type](typeName string, target *T, cache *btf.Cache) (*btf.Spec, *btf.Handle, error) {
```
ti-mo:
```suggestion
func targetInModule[T btf.Type](typeName string, target *T, cache *btf.Cache) (*btf.Spec, *btf.Handle, error) {
```
🤓

features/prog.go
```diff
@@ -231,102 +24,6 @@ var helperCache = internal.NewFeatureCache(func(key helperKey) *internal.Feature
...
-	})
-}
-
-func helperProbeNotImplemented(pt ebpf.ProgramType) bool {
-	switch pt {
-	case ebpf.Extension, ebpf.LSM, ebpf.StructOps, ebpf.Tracing:
-		return true
-	}
-	return false
+func HaveProgramHelper(cache *btf.Cache, helper asm.BuiltinFunc) error {
```
ti-mo: I'd put the cache as the last argument always.

Maybe we put it in a *ProbeOptions instead?

ti-mo (commented):
Thanks! Few nits.

features/prog.go
dylandreimerink: Yea, good idea, that make future changes easier to handle will do.

dylandreimerink:
> That's going to get messy, at least for the changes in this particular PR. We can batch the verifier output parsing and the vmlinux bump, they're kinda related (upgrade pain) anyway.

Yes, agreed. We will have to merge this or the other PR with broken CI, so some additional checks required, but not the end of the world.

> I'm working on some changes to ci-kernels that will sidestep this problem down the line and will make the tests more reproducible.
>
>tl;dr: generate test matrix from ci-kernels/versions.json using a workflow, consume that as e.g. uses: cilium/ci-kernels/matrix@7.3-rc4 and make dependabot open PRs here when versions.json is updated+tagged so we can address these issues in lockstep. The floating mainline/stable/longterm pins on the OCI images are not ideal to say the least.

Yes, sounds good. That is what I had been thinking of as well. That way we can upgrade the kernels we use and apply fixes in the same PR instead of having to fix CI in a hurry every time.

features/features.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"slices"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/btf"
+)
+
+type ProbeOptions struct {
+	BTFCache *btf.Cache
```
ti-mo, line 12: Thanks! This is called `Cache` in `CollectionOptions`.

I also just realized talking to Bryce on Slack that we need to make this take a `btf.Spec` for BTFHub/out-of-band BTF users to continue being able to use this.

dylandreimerink: Yea, agreed.

rgo3:
General comment that I discussed with Timo offline: I don't have an example but I suspect that the btf approach could yield false positives if some features are emitted to btf but compiled out based on kconfig? I'll try to find an example. 

Not as an objection to this approach, but if it's the cade we should document it.

